PR #530, opened 2021-01-25
Messaging advanced features/management

lucaslcode:


app/frontend/src/components/Dialog.tsx
```diff
@@ -36,7 +36,12 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export function Dialog(props: Omit<DialogProps, "className">) {
+export type AccessibleDialogProps = Required<
```
darrenvong: Is this trying to make `aria-labelledby` required? I think you can do something like this with an interface instead to make this less obscure to read:

```typescript
interface AccessibleDialogProps extends Omit<DialogProps, "className"> {
  "aria-labelledby": string;
}
```

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -18,30 +18,43 @@ import PageTitle from "../../../components/PageTitle";
...
+    menu: false,
+    invite: false,
+    members: false,
+    settings: false,
+    leave: false,
+  });
 
-  const handleClick = () => {
-    setMenuOpen(true);
+  const handleClick = (what: keyof typeof isOpen) => {
```
darrenvong:
```suggestion
  const handleClick = (item: keyof typeof isOpen) => {
```
Slightly clearer since the callback is receiving back which item on the menu it's getting back?

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -18,30 +18,43 @@ import PageTitle from "../../../components/PageTitle";
...
+    if (what !== "menu") {
+      setIsOpen({ ...isOpen, [what]: true, menu: false });
+    } else {
+      setIsOpen({ ...isOpen, [what]: true });
+    }
   };
 
-  const handleClose = () => {
-    setMenuOpen(false);
+  const handleClose = (what: keyof typeof isOpen) => {
```
darrenvong:
```suggestion
  const handleClose = (item: keyof typeof isOpen) => {
```
Same comment as above

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <DialogTitle id="members-dialog-title">Chat Members</DialogTitle>
+      <DialogContent>
+        {error && <Alert severity="error">{error}</Alert>}
+        <List>
+          {members.isLoading ? (
+            <CircularProgress />
+          ) : (
+            Array.from(
+              (members.data ?? new Map<undefined, undefined>()).values()
+            ).map((user) =>
```
darrenvong:
```suggestion
            Array.from(
              members.data?.values() ?? []
            ).map((user) =>
```
This looks a bit simpler to me than creating an empty map just so we can call `.values`

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <ListItemText
+        disableTypography
+        className={classes.memberListItemTextContainer}
+      >
+        <TextBody>
+          {member.name}
+          {memberIsAdmin && " (admin)"}
+        </TextBody>
+        {userIsAdmin &&
+          (memberIsAdmin ? (
```
darrenvong: These buttons don't line up properly on the smallest phone screen width (320px)
Manage members dialog

But maybe it doesn't matter so much for now in this iteration? It looks like in the latest hi-fi design, we're shifting to using icons instead?

darrenvong (changes requested):
Mostly minor comments, nice work on this!

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
lucaslcode: Oops

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
lucaslcode: That's because of the long name with no spaces to wrap. I added `noWrap` so names will be on one line now and truncated.

app/frontend/src/components/IconButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: IconButtonProps & { loading?: boolean }) {
+  const classes = useStyles();
+  const theme = useTheme();
+  return (
+    <MuiIconButton {...otherProps}>
+      {loading ? (
+        <CircularProgress
+          className={classes.circularProgress}
+          //stolen from source for MUI IconButton
+          size={theme.typography.pxToRem(18)}
```
darrenvong, line 34: Seems like a useful helper to use to "rem-ify" everything so we don't get hardcoded pixel values elsewhere?

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
```diff
@@ -87,29 +85,31 @@ function MemberListItem({
...
-            <Button loading={makeAdmin.isLoading} onClick={handleMakeAdmin}>
-              Make admin
-            </Button>
-          ))}
-      </ListItemText>
+    <ListItem dense className={classes.memberListItemContainer}>
+      {userIsAdmin &&
+        //TODO: Colours
+        (memberIsAdmin ? (
+          <IconButton
```
darrenvong:
```suggestion
          <IconButton
            aria-label="Remove as admin"
```
We should add the text we removed in favour of icon buttons as `aria-label` for these instead so they're still accessible to non-visual users.

app/frontend/src/features/messages/groupchats/MembersDialog.tsx
```diff
@@ -87,29 +85,31 @@ function MemberListItem({
...
+            size="small"
+            loading={removeAdmin.isLoading}
+            onClick={handleRemoveAdmin}
+          >
+            <CloseIcon />
+          </IconButton>
+        ) : (
+          <IconButton
```
darrenvong:
```suggestion
          <IconButton
            aria-label="Make admin"
```

darrenvong (changes requested):

lucaslcode:
Oh damn I forgot I was supposed to merge this first...

lucaslcode:
@darrenvong Made quite a lot of changes, I thought "I may as well go closer to the hifi dialogs" then "I may as well create a confirmation dialog component" etc

Might want to only look at the 1st of the 2 commits, the other is just the merge (it was horrible)

app/frontend/src/components/ConfirmationDialogWrapper.tsx
```diff
@@ -0,0 +1,50 @@
...
+import Button from "./Button";
+import {
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
+} from "./Dialog";
+
+interface ConfirmationDialogWrapperProps<P extends { onClick: () => void }> {
```
lucaslcode: I was hoping being clever like this would mean TS would check the child component has "onClick", but it's not that smart (even if it's in a prop and not children), so I did the assert below

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+        const previousGroupChat = queryClient.getQueryData<GroupChat.AsObject>([
+          "groupChat",
+          groupChatId,
+        ]);
+        const newAdminUserIdsList = previousGroupChat?.adminUserIdsList ?? [];
+        newAdminUserIdsList.push(member.userId);
+        queryClient.setQueryData(["groupChat", groupChatId], {
+          ...previousGroupChat,
+          adminUserIdsList: newAdminUserIdsList,
+        });
```
lucaslcode: Optimistic update. Otherwise it stops loading, but the user stays there until the invalidation is completed.

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+      onSuccess: () => {
+        const previousGroupChat = queryClient.getQueryData<GroupChat.AsObject>([
+          "groupChat",
+          groupChatId,
+        ]);
+        const newAdminUserIdsList = previousGroupChat?.adminUserIdsList ?? [];
+        newAdminUserIdsList.splice(
+          newAdminUserIdsList.indexOf(member.userId),
+          1
+        );
```
lucaslcode: This seems pretty old fashioned but any other way seemed convoluted.

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
darrenvong: I think it's better to do the optimistic update in `onMutate` so you can roll back to the previous state in `onError` before showing the error as shown [in the docs example](https://react-query.tanstack.com/guides/optimistic-updates#updating-a-list-of-todos-when-adding-a-new-todo)? Otherwise the user will be stuck with the faulty state and the error message at the same time which is a bit confusing.

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
darrenvong: Hmm not sure if this is gonna cause problem since `splice` mutate the list in place rather than doing it immutably (on a copy of a new list) like in most React code... maybe not since `setQueryData` synchronously updates the cache?

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+        queryClient.setQueryData(["groupChat", groupChatId], {
+          ...previousGroupChat,
+          adminUserIdsList: newAdminUserIdsList,
+        });
```
darrenvong, line 97: Since `splice` updates the list in place, and `newAdminUserIdsList` is another reference to `previousGroupChat.adminUserIdsList`, would it still work if you do:
```typescript
queryClient.setQueryData(["groupChat", groupChatId], previousGroupChat)
```
?

app/frontend/src/components/ConfirmationDialogWrapper.tsx
darrenvong: Is there other good reasons to keeping the generic props if it doesn't work?

app/frontend/src/components/ConfirmationDialogWrapper.tsx
```diff
@@ -0,0 +1,50 @@
...
+    );
+  }
+  const [isOpen, setIsOpen] = useState(false);
+  const child = cloneElement<P>(children, {
+    onClick: () => setIsOpen(true),
+  } as P);
+  const ariaLabel = `${title.replace(/\s+/g, "")}-confirmation-dialog`;
+  const handleConfirm = () => {
+    children.props.onClick();
+    setIsOpen(false);
```
darrenvong: This seems a bit unorthodox just to "hijack" the original children to get its `onClick` method, since the main plus of using the `children` prop is it's supposed to be generic/flexible.

I think it'd be easier and a bit clearer to pass through the confirmation action function as a prop like `onConfirm` and change `children` into a render prop function instead so `children` can remain generic while have access to the `setIsOpen` state setter.

app/frontend/src/components/ConfirmationDialogWrapper.tsx
```diff
@@ -0,0 +1,50 @@
...
+  };
+  return (
+    <>
+      {child}
```
darrenvong: If using `children` as a render prop, it'd become:
```typescript
{children(setIsOpen)}
```

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
```diff
@@ -0,0 +1,249 @@
...
+            >
+              <IconButton
+                aria-label="Remove as admin"
+                size="small"
+                loading={removeAdmin.isLoading}
+                onClick={handleRemoveAdmin}
+              >
+                <CloseIcon />
+              </IconButton>
+            </ConfirmationDialogWrapper>
```
darrenvong:
```suggestion
            <ConfirmationDialogWrapper
              title="Step down as admin?"
              message={
                `Are you sure you want to stop being an admin ` +
                `of this group chat? You will not be able to become ` +
                `an admin again unless another admin adds you.`
              }
              onConfirm={handleRemoveAdmin}
            >
              {(setIsOpen) => (
                <IconButton
                aria-label="Remove as admin"
                size="small"
                loading={removeAdmin.isLoading}
                onClick={() => setIsOpen(true)}
              >
                <CloseIcon />
              </IconButton>
              )}
            </ConfirmationDialogWrapper>
```
If using the render prop pattern I suggested in `ConfirmationDialogWrapper`

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -150,10 +152,19 @@ export default function GroupChatView() {
...
                   {isChatAdmin && (
-                    <MenuItem onClick={() => handleClick("settings")}>
-                      Chat settings
+                    <MenuItem onClick={() => handleClick("admins")}>
+                      Manage chat admins
                     </MenuItem>
                   )}
+
+                  {
+                    //<Menu> doesn't accept JSX Fragment children
```
darrenvong: Really? Seems odd that they have `children` typed as `React.ReactNode` which includes Fragment then 🤔

darrenvong (changes requested):

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
lucaslcode: Oh I was meant to do `Array.from` to copy it. I'd rather avoid mutating the original even if it works

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
lucaslcode: It won't be a problem because it only sets the data on success. It's more a "semi-optimistic" update. 

I could do it their way though, and it would look instant to the user. What do you think?

app/frontend/src/components/ConfirmationDialogWrapper.tsx
lucaslcode: In case it works in a future TS version? I think it will work when I change to render prop anyway

app/frontend/src/components/ConfirmationDialogWrapper.tsx
lucaslcode: Yeah that's a better idea

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
lucaslcode: It lints and works fine but gave console warnings

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
darrenvong: Ahh true since it's scoped to onSuccess 👍   Is the delay noticeable if you didn't go with this "semi-optimistic" route? I suspect so for people with slow connection?

lucaslcode: Even not a slow connection, because the loading spinner stops (because the mutation is complete), but there a short pause while the query refetches on invalidation, then the user jumps to the correct place.

darrenvong (approved):

app/frontend/src/features/messages/groupchats/AdminsDialog.tsx
darrenvong: Seems like a worthy tradeoff for better UX but without the faff of having to worry about rolling back overly optimistic updates, let's stick with it 👍

PR #532, opened 2021-01-26
Backend/feature/communities tests

aapeliv:
This is a mess

### DIFF
--- a/app/backend/src/tests/test_communities.py
+++ b/app/backend/src/tests/test_communities.py
@@ -15,7 +15,7 @@ from couchers.models import (
 from couchers.tasks import enforce_community_memberships
 from couchers.utils import create_coordinate, create_polygon_lat_lng, now, to_aware_datetime, to_multi
 from pb import communities_pb2, pages_pb2
-from tests.test_fixtures import communities_session, db, generate_user, pages_session, testconfig
+from tests.test_fixtures import communities_session, db_impl, generate_user, pages_session, testconfig
 
 
 @pytest.fixture(autouse=True)
@@ -169,8 +169,9 @@ def get_group_id(session, group_name):
     )
 
 
-@pytest.fixture(scope="class")
-def testing_communities(db):
+@pytest.fixture(params=["models", "migrations"], scope="class")
+def testing_communities(request):
+    db_impl(request.param)
     user1, token1 = generate_user(username="user1", geom=_create_1d_point(1), geom_radius=0.1)
     user2, token2 = generate_user(username="user2", geom=_create_1d_point(2), geom_radius=0.1)
     user3, token3 = generate_user(username="user3", geom=_create_1d_point(3), geom_radius=0.1)
@@ -206,8 +207,11 @@ def testing_communities(db):
     _create_place(token8, "World, Attraction", "Place content", "Somewhere in w", 51.5)
     _create_place(token6, "Country 2, Region 1, Attraction", "Place content", "Somewhere in c2r1", 59)
 
+    yield
+
 
 class TestCommunities:
+    @staticmethod
     def test_GetCommunity(testing_communities):
         with session_scope() as session:
             user2_id, token2 = get_user_id_and_token(session, "user2")
@@ -324,6 +328,7 @@ class TestCommunities:
             assert res.member_count == 2
             assert res.admin_count == 2
 
+    @staticmethod
     def test_ListCommunities(testing_communities):
         with session_scope() as session:
             user1_id, token1 = get_user_id_and_token(session, "user1")
@@ -339,6 +344,7 @@ class TestCommunities:
             )
             assert [c.community_id for c in res.communities] == [c1r1_id, c1r2_id]
 
+    @staticmethod
     def test_ListGroups(testing_communities):
         with session_scope() as session:
             user1_id, token1 = get_user_id_and_token(session, "user1")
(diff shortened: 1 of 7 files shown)


aapeliv:
@kalvdans: this takes forever to run. Do you know how to only run the `testing_communities` fixture once in the file and run all tests against it, without recreating everything every time?

kalvdans:
> @kalvdans: this takes forever to run. Do you know how to only run the `testing_communities` fixture once in the file and run all tests against it, without recreating everything every time?

Can we perhaps do a database dump the first time, and just reload the dump?

otherwise we just have to optimize, don't create stuff that's not needed by the active test.

aapeliv:
It seems that `scope="module"` should do precisely this. However, I'm running into some issues with PostGIS now, I think this is doing something unexpected.

kalvdans:
> It seems that `scope="module"` should do precisely this.

I have had a hard time understanding the magic what scope="module" does. If you want tests to be dependent on the leftover from an earlier test, use the `unittest` framework and its setUpClass function. https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUpClass

pytest will run tests written in the unittest framework as well.

aapeliv:
It just reuses the same fixture for all tests in the same module, without destroying it and recreating it, right?

I don't think we should mix in unittest, this seems to work fine?

I mean, it's not as nice as having completely isolated unit tests, but at the same time, these are read-only queries.

Another option would be to just put them all in the same test?

aapeliv:
@kalvdans: what would you want me to do about this? I guess the lame option is to just have these independent tests that are very slow. For that, I can revert the last commit. Another is to share the same `testing_communities` fixture as it is now. Up to you?

kalvdans:
> I guess the lame option is to just have these independent tests that are very slow. For that, I can revert the last commit. Another is to share the same `testing_communities` fixture as it is now. Up to you?

The issue with `scope="module"` is that if you have a test with the `db` fixture in between two tests with `testing_communities` fixture, it will destroy the database. It is better to use `scope="class"` and put all the tests that need to run after one another in that class. I'll try to see if I get something to work on top of your branch.

See https://github.com/Couchers-org/couchers/pull/554 for my take on the scoping.

aapeliv:
@kalvdans: that PR looks very good to me. Thanks for doing that. Are you happy for me to merge that into this and this into `develop`?

app/backend/src/tests/test_groups.py
```diff
@@ -0,0 +1,213 @@
...
+        assert res.parents[3].group.description == "Description for Country 2, Region 1, Foodies"
+        assert res.main_page.type == pages_pb2.PAGE_TYPE_MAIN_PAGE
+        assert res.main_page.slug == "main-page-for-the-country-2-region-1-foodies-community"
+        assert res.main_page.last_editor_user_id == 6
+        assert res.main_page.creator_user_id == 6
+        assert res.main_page.owner_user_id == 6
+        assert res.main_page.title == "Main page for the Country 2, Region 1, Foodies community"
+        assert res.main_page.content == "There is nothing here yet..."
+        # assert res.main_page.can_edit # TODO
+        assert lists_equal(res.main_page.editor_user_ids, [6])
```
kalvdans: ` assert res.main_page.editor_user_ids == [6]`

app/backend/src/tests/test_groups.py
```diff
@@ -0,0 +1,213 @@
...
+        hitchhikers_id = get_group_id(session, "Hitchhikers")
+        c1r2foodies_id = get_group_id(session, "Country 1, Region 2, Foodies")
+
+    with groups_session(token1) as api:
+        res = api.ListMembers(
+            groups_pb2.ListMembersReq(
+                group_id=hitchhikers_id,
+            )
+        )
+        assert lists_equal(res.member_user_ids, [user1_id, user2_id, user5_id, user8_id])
```
kalvdans: `assert set(res.member_user_ids) == set([user1_id, user2_id, user5_id, user8_id])`

app/backend/src/tests/test_fixtures.py
```diff
@@ -406,3 +431,8 @@ def fast_verify(hashed: bytes, password: bytes) -> bool:
     with patch("couchers.crypto.nacl.pwhash.verify", fast_verify):
         with patch("couchers.crypto.nacl.pwhash.str", fast_hash):
             yield
+
+
+def lists_equal(list1, list2):
+    """Compare equality of (hashable) elements in two lists, by computing size of symmetric set difference"""
+    return len(set(list1) ^ set(list2)) == 0
```
kalvdans: Remove function and inline into callers instead - I've commented on some examples.

kalvdans (changes requested):

app/backend/src/tests/test_communities.py
```diff
@@ -0,0 +1,467 @@
...
+
+            res = api.ListNearbyUsers(
+                communities_pb2.ListNearbyUsersReq(
+                    community_id=c1r1c2_id,
+                )
+            )
+            assert lists_equal(res.nearby_user_ids, [user4_id])
+
+
+# TODO: requires transferring of content
```
kalvdans, line 478: I'm usually against leaving TODO:s as comments in the code, I rather would like an issue opened. But I don't feel strongly about it so I'm ok with leaving it as-is.

aapeliv: Yes, I was going to work on those next! (Implementation then tests)

kalvdans (approved):
I get some spurios
```
psycopg2.errors.InternalError_: no spatial operator found for 'st_contains': opfamily 21450 type 21337
```
test failures when running the whole test suite locally. But if it doesn't occur on CI, I'm fine with how this branch looks now!

PR #533, opened 2021-01-26
Code coverage: Add __init__.py files

aapeliv:


aapeliv:
What's wrong with this @kalvdans? They're empty, they just declare the folders as packages. That makes the code coverage tool pick up files that are never tested (and hence imported). It's also good to have them as proper packages anyway, I believe it makes python resolve imports in a more correct way (gets put earlier in search order, or similar).

kalvdans:
At work we extensively use "PEP420 namespaces", without `__init__.py` file, and it works. We never had any problem with pytest-cov. Let me test it out tomorrow evening, there might be some flag we could add instead.

I'm fine adding them though, we can always revert if I come up with something else :)

aapeliv:
I'd like to add this so we can see full coverage. I'm also happy if you want to look into other ways of achieving this, we can revert this and go for something else in another PR?

kalvdans:
pytest --cov=couchers --cov-report=html works fine for me in `communities`:
image

Can you link to the CI coverage, I forgot where it is located..

aapeliv:
That is missing `couchers/servicers/sso.py` and `couchers/servicers/gis.py` that are not imported in any tests.

https://backend-bugfix-code-coverage--bcov.preview.coucher.org/

kalvdans (approved):
LGTM!

Issue #531, opened 2021-01-26
Error on login attempt with wrong password


aapeliv:
<img width="977" alt="Screen Shot 2021-01-25 at 7 02 39 pm" src="https://user-images.githubusercontent.com/2975660/105781132-f8ff6e80-5f3f-11eb-967a-77b7ffca80c9.png">

1. Try login with `test`
2. Enter wrong password.
3. You get an error: `Reads "can't access property "getResponseMessage", e is undefined".`

lucaslcode:
I suspect this is the same reason as #481, giving an "unauthenticated" error which triggers a log out, which fails because you aren't logged in. However I don't know why you then get that error, we don't call that function anywhere.

lucaslcode:
https://github.com/Couchers-org/couchers/blob/34cf62f4cd3d7dc75d0fd0717e78effe0d7b8d5d/app/backend/src/couchers/servicers/auth.py#L344 Errors in this function should be NOT_FOUND or something else

ShmuelTreiger:
When I try to login with test, I get invalid username/password. Could you add more steps to reproduce?

![image](https://user-images.githubusercontent.com/14131043/105946751-bb244800-601c-11eb-866b-3c909f3666f8.png)

ShmuelTreiger:
Brief googling tells me this is probably a frontend problem

https://www.google.com/search?q=can%27t+access+property+e+is+undefined&oq=can%27t+access+property+e+is+undefined&aqs=chrome..69i57j0i22i30l4.8719j1j7&sourceid=chrome&ie=UTF-8

lucaslcode:
Yes and no, there's two issues.

The first is the frontend treats the UNAUTHENTICATED error code to mean a user has tried to access something without being logged in. Since the login session uses secure cookies which can't be access in JS, we just have a variable `isAuthenticated` which is set to true on login and false on log out. However the cookie might expire, or the backend might log us out, so next time we try to make an API call, we get UNAUTHENTICATED, so we know to set `isAuthenticated` to false. This happens in an API interceptor. .

So that error code is not really relevant for an incorrect password, because it's a kind of failure rather than trying to access something without being logged in, if that makes sense. 

The other problem is what the error message says, which is a frontend problem. But I think it's a gRPC library issue or related to how interceptors, because our code doesn't call `getResponseMethod` anywhere. And changing the above issue should fix this bug anyway.

aapeliv:
[...]

PR #534, opened 2021-01-27
[MRG] Changed error message from unauthenticated to not found

### LINKED ISSUE #451: Backend gives unauthenticated error instead of permission denied for wrong passwords/login tokens

### LINKED ISSUE #481: Invalid tokens should not give UNAUTHENTICATED error
Should be reserved for not logged-in users trying to do something that requires login. Perhaps NOT_FOUND is more appropriate.

ShmuelTreiger:


app/backend/src/couchers/servicers/auth.py
```diff
@@ -346,7 +346,7 @@ def Authenticate(self, request, context):
                 logger.debug(f"Didn't find user")
                 # do about as much work as if the user was found, reduces timing based username enumeration attacks
                 hash_password(request.password)
-                context.abort(grpc.StatusCode.UNAUTHENTICATED, errors.INVALID_USERNAME_OR_PASSWORD)
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_USERNAME_OR_PASSWORD)
```
kalvdans, line 349: Remove the time delay as that is useless now when we return a different error anyway.

Also change the error message sent to the client to "User not found" or similar.

lucaslcode: I think the error message and time delay is good, we should just make the other error above also NOT_FOUND (this is needed anyway to fix #531). So no matter if the user doesn't exist or the password is incorrect, you get NOT_FOUND "Invalid username or password"

(Although an attacker can easily know if a user exists with the other login function which tell you how exactly that.)

ShmuelTreiger: Changed second error message. This should be good now

Out of curiosity, where's the time delay? I'm not noticing it at all

Edit: I suppose we should make tests for this. Not quite done.

aapeliv: @ShmuelTreiger: the `hash_password(...)` call takes some time (not so noticeable as a user, but can be timed easily if you look at how long the request takes). If it always takes longer when the user exists, then it's possible to find usernames by just checking whether the rpc returned quickly or took some time. It's good to not leak that kind of info all over the place unknowingly.

lucaslcode:
@aapeliv it is still returning UNAUTHENTICATED on the preview despite the latest commit, is it related to CI?

@ShmuelTreiger Do you want to do the same thing for the token login functions? They are also giving UNAUTHENTICATED atm. If that's annoying we can merge this first, once it works.

aapeliv:
> @aapeliv it is still returning UNAUTHENTICATED on the preview despite the latest commit, is it related to CI?

The CI doesn't deploy the backend, only the frontends (which hit `develop`). So you can't test it without doing it locally, if that's what you mean?

lucaslcode (dismissed):
Up to you if you want to merge this now, or change the token errors too and / or the frontend line which I mentioned (pasting my comment in the issue below)

> On [this line](https://github.com/Couchers-org/couchers/blob/69c27faeea54f6e6084ac8bea67d34f1cdb0f580/app/frontend/src/features/auth/useAuthStore.ts#L134), I forgot to await the service call, which is asynchronous.
>
> You can put that in your PR too if you want, or I can do it later. Just change that line to
await service.user.logout();
>
> With that fix, instead what happens is the message "You were logged out." is displayed at the start login page, which is what is supposed to happen when the server logs you out

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: OK, I understand the logic, but I don't see where that happens?

ShmuelTreiger:
> Do you want to do the same thing for the token login functions? They are also giving UNAUTHENTICATED atm. If that's annoying we can merge this first, once it works.

@lucaslcode do you mean here?

https://github.com/Couchers-org/couchers/blob/69c27faeea54f6e6084ac8bea67d34f1cdb0f580/app/backend/src/couchers/servicers/auth.py#L315

app/backend/src/couchers/servicers/auth.py
aapeliv: Just above, on line 348. We hash the password but throw away the result.

lucaslcode:
@ShmuelTreiger yes, and a few others if you do a search in that file. There shouldn't be any instances of that error code in `auth.py`

lucaslcode (dismissed):
Looks good! There's just four tests still failing from the changed status code.

ShmuelTreiger:
> Looks good! There's just four tests still failing from the changed status code.

[It looks like the same 2 tests failed twice](https://gitlab.com/couchers/couchers/-/pipelines/250190364/test_report) or am I reading this wrong?

test_invalid_token is in test.auth, but it seems to be testing interceptors directly? (Hence the failure, the interceptor is raising UNAUTHENTICATED, which auth.py no longer raises)

Edit: I see. One's in models, one's in migrations. What's the difference?

lucaslcode:
@aapeliv you can sort this out better than me. Not sure why the interceptor would be causing NOT_FOUND now if that code wasn't changed. 

Maybe something to do with the `real_api_session` test function?

ShmuelTreiger:
Sorry, I wasn't clear. Interceptor is returning UNAUTHENTICATED. I changed the test to check for NOT_FOUND. Just want to double check that everything is as it should be before I revert changes, given it's in test.auth, which seems to me like it should be testing auth.py

kalvdans:
> One's in models, one's in migrations. What's the difference?

"models" runs the test with a database set up with the correct tables, whereas "migrations" starts with a completely empty database and applies all database migration files. It is documented at https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/tests/test_fixtures.py#L56-L61

lucaslcode:
@ShmuelTreiger I wouldn't worry too much about the name

aapeliv:
@ShmuelTreiger: you changed some more errors to `NOT_FOUND`. I had a look, and I think it's a better error code. But that means you need to fix two more tests to check the right error codes. There's `test_ChangeEmail_wrong_token` inside `src/tests/test_account.py`, and `test_invalid_token` in the auth tests.

ShmuelTreiger:
@lucaslcode I'm not worried about the name, just didn't know if they were separate tests I needed to fix or the same one.

@aapeliv I understand. I just wanted to make sure test_invalid_token is working as intended. But I'll push my fix I made last night.

app/backend/src/tests/test_auth.py
```diff
@@ -140,7 +140,7 @@ def test_invalid_token(db):
         res = api.GetUser(api_pb2.GetUserReq(user=user2.username))
 
     assert e.value.code() == grpc.StatusCode.UNAUTHENTICATED
-    assert e.value.details() == "Unauthorized"
```
aapeliv: This is the last test that's failing. You should change this back to `"Unauthorized"`, instead of `errors.INVALID_TOKEN`.

This is not related to login, it's just checking that if you hit the API with the wrong token, it'll return `UNAUTHENTICATED`. If I now am on top of things, this is basically the only place that should be returning `UNAUTHENTICATED`.

ShmuelTreiger: Yup, forgot to change that back, oops

aapeliv (approved):
Feel free to merge when checks go green

PR #539, opened 2021-01-29
Split pages into places and events

aapeliv:
Also renames "point of interest" to "place"


aapeliv:
I don't think this is ideal, open to suggestions!

lucaslcode (approved):
I think this is fine for now. Ultimately we will have separate components for the `GuidePage` and `PlacePage` and just abstract as much as possible into share components, but I don't think there's much point doing that yet.

PR #540, opened 2021-01-29
Implement slugify in postgres

aapeliv:
This moves slug generation into the database. This is nicer for a variety of reasons.

On an abstract level, the slug is some generated data from the title and is part of the data, and shouldn't need to be recomputes all the time (in the future we could use a generated postgres column, but I'm not sure how it hooks into sqlalchemy).

On a practical level, this makes it much easier to return full info about a page/community in GIS/GeoJSON functions without having to go through python, so the full data for maps can be sent straight from the database which is very efficient.


kalvdans:
Cool! I heard you could implement postgres on top of postgres :)

I was really hoping we could push slugification to the frontends though, to minimize the amount of data needed to be sent over the API.

That said, I don't really care which language slugify is implemented in as long as we don't store the slugified text in the database. Please get rid of the code duplication, use a single implementation of the function in tests and live.

aapeliv:
> Please get rid of the code duplication, use a single implementation of the function in tests and live.

There are two implementations: one for migrations (used in prod, as well as half the tests) and the other used when creating all without migrations (used in the other half of tests only).

I'm not sure how to make do it very cleanly. The migrations should be the "history" of the DB, and the current pure sql should be the "current version". So if the function changes, we'd create a new migration but just update the current version, I imagine.

kalvdans:
> I'm not sure how to make do it very cleanly. The migrations should be the "history" of the DB, and the current pure sql should be the "current version". So if the function changes, we'd create a new migration but just update the current version, I imagine.

custom functions used only during runtime doesn't belong to the database scheme imho and shouldn't be part of the migrations at all.

aapeliv:
> custom functions used only during runtime doesn't belong to the database scheme imho and shouldn't be part of the migrations at all.

I agree in principle, but it's a stored procedure and the database stores it. In one way or another, it must be saved in the database as a persistent object.

kalvdans:
> > custom functions used only during runtime doesn't belong to the database scheme imho and shouldn't be part of the migrations at all.
> 
> I agree in principle, but it's a stored procedure and the database stores it. In one way or another, it must be saved in the database as a persistent object.

Ah, too bad, then I think I will :thumbsdown: this PR and keep doing the slugify in Python and Javascript. It is not important that the slug is done in the exact same way everywhere.

aapeliv:
> It is not important that the slug is done in the exact same way everywhere.

It is imperative that the slugs are the same. Otherwise entities don't have stable URLs, which is bad for discoverability and a nightmare for SEO.

Look, I think we'll have to figure out a way to do stored procedures one day anyway. I think this is pretty clean.

kalvdans (dismissed):
After the meeting you convinced me that slugify could be seen more as new column than a generic function, so that convinces me it should be in the migrations. Great job and please merge!

PR #549, opened 2021-01-31
Fix flaky CI

### LINKED ISSUE #401: The backend coverage upload is flaky in CI
I've no clue why, but it's annoying and flaky.

See e.g. https://gitlab.com/couchers/couchers/-/jobs/897744854

aapeliv:
Closes #401


darrenvong (approved):
Guess it must be working since CI is passing 😄

PR #550, opened 2021-01-31
Split unseen host request message count into sent and received

### LINKED ISSUE #548: Host requests notifications always appear on the "Hosting" tab
I think this is due to the count of host requests sent and received being a combined count, which was fine in the alpha as they live together. In the beta, they are split into a "hosting" and "surfing" tab, which means when you get an unread message on the "surfing" tab, the notification count still shows up under "hosting".

The API only returns this as a combined count at the moment and so there's no easy way to tell apart which are "hosting" requests and which are "surfing" requests at the moment, so that will probably need updating. Then frontend needs updating too to use these new separated counts.

aapeliv:
Closes #548


darrenvong (approved):
LGTM

PR #529, opened 2021-01-25
Backend/feature/profile fields api

ShmuelTreiger:

Issue #410, opened 2020-12-10
Profile/user page design


lucaslcode:
Users need the following additional simple fields implemented for the beta ui. More complex fields in seperate issues.

- pronouns
- Hometown
- meet up status (like hosting status)
- education

**Markdown fields** - the second level fields are pre-filled as markdown headings, there by default but can be deleted
- About me
  - current mission
  - what I do in my free time
- My travels
  - why I use Couchers,
  - my favourite travel story
- Things I like
  - Art
  - Books
  - Movies
  - Music
- Additional information

**Info for surfers**
- Camping requests ok?
- Pets ok?

**Info for hosts**
- Has pets (Yes/No and additional free text)
- Private/shared space
- I smoke at home (y/n)
- I drink at home (y/n)
- Has housemates (y/n/details if yes)
- Has children (y/n/details if yes)
- parking (y/n/free off-site/free on-site/paid off-site/paid on-site)
- other host info (markdown)
  - pre-filled heading: What I can share with guests

These fields need to be modified:

- hosting status get rid of 'difficult' (maybe it should just be removed from frontend initially?)
- Remove "color" 
- Remove "multiple groups OK?"
- area should be a markdown field with pre-filled headings:
  - Public transport
  - Parking
  - Accessibility

aapeliv:
> * current mission*
> * Why i use couchers
> * What I do in free time
> * Favourite travel story
> * Favourite art, books, movies & music (one field or 4?)*
> * Additional info about me*
> * Favourite quote

Would it be possible to have these in one markdown field? I feel like people would like to customize this as well?

lucaslcode:
It would certainly be easier

lucaslcode:
Updated this issue to reflect slightly more generalized markdown fields with default, pre-filled headers

ShmuelTreiger:
I began a fix in https://github.com/Couchers-org/couchers/tree/backend/feature/profile-fields-api

1) I couldn't figure out how to do the pre-filled headings in markdown

2) I couldn't find where the models are tested, so I didn't add any tests

3) Haven't migrated

Other than that, I've done the basic changes

lucaslcode:
Great! Tests are in `https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/tests/test_api.py` though there are probably some other broken ones too. Open a pull request for review and comments.

ShmuelTreiger:
So a couple questions/issues I'm having:
[...]

PR #555, opened 2021-01-31
[MRG] Test all updated fields and clarify why two users are needed

kalvdans:
Update after comments on https://github.com/Couchers-org/couchers/pull/529#issuecomment-768201670

### DIFF
--- a/app/backend/src/tests/test_api.py
+++ b/app/backend/src/tests/test_api.py
@@ -166,7 +166,6 @@ def test_update_profile(db):
         assert user.occupation == "Testing"
         assert user.about_me == "I rule"
         assert user.about_place == "My place"
-        assert user.color == "#111111"
         assert user.hosting_status == api_pb2.HOSTING_STATUS_CAN_HOST
         assert "Binary" in user.languages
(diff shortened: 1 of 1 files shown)


app/backend/src/tests/test_api.py
```diff
@@ -162,9 +162,18 @@ def test_update_profile(db):
         assert user.lat == 0.01
         assert user.lng == -2
         assert user.radius == 321
+        assert user.gender == "Bot"
+        assert user.occupation == "Testing"
+        assert user.about_me == "I rule"
+        assert user.about_place == "My place"
+        assert user.color == "#111111"
```
ShmuelTreiger: We're taking out color in #410

ShmuelTreiger (changes requested):
Other than color, looks good to me

ShmuelTreiger (approved):

app/backend/src/tests/test_api.py
kalvdans: Oh, thanks for noticing, and thanks again for fixing it in this branch :) Merging!

PR #556, opened 2021-01-31
Sort imports

lucaslcode:
wew

Merge #530 into develop first


app/frontend/package.json
```diff
@@ -40,6 +40,17 @@
       "prettier/@typescript-eslint",
       "prettier/react"
     ],
+    "plugins": [
+      "eslint-plugin-simple-import-sort"
```
darrenvong, line 44:
```suggestion
      "simple-import-sort"
```
Apparently you can omit the `eslint-plugin` bit according to their README

darrenvong (approved):
Only a small comment about removing the `eslint-plugin` prefix, but up to you to change or not

app/frontend/package.json
lucaslcode: I'm going to skip it for now so you don't have to re-approve, feel free to change it upstream with another PR.

PR #566, opened 2021-02-01
Run yarn format

aapeliv:


lucaslcode:
This was done in #556 , does it need to be done here?

aapeliv:
> This was done in #556 , does it need to be done here?

No it doesn't. Thanks! I didn't notice that.

PR #568, opened 2021-02-02
General Hifi styling preparations

lucaslcode:


app/frontend/src/theme.ts
```diff
@@ -1,37 +1,69 @@
-import { createMuiTheme, ThemeOptions } from "@material-ui/core";
+import { createMuiTheme, makeStyles, ThemeOptions } from "@material-ui/core";
 import type {} from "@material-ui/lab/themeAugmentation";
 
 const spacing = (factor: number) => `${0.5 * factor}rem`;
-const borderRadius = 10;
+const borderRadius = 12;
+
+const useGlobalStyles = makeStyles((theme) => ({
+  h1Large: {
```
darrenvong: Is there any reason why we couldn't include this in the main theme too?

lucaslcode: Custom variants are in v5 but not v4

darrenvong: This [stackoverflow](https://stackoverflow.com/questions/58168798/add-custom-theme-variable-object-to-createmuitheme) answer seem to suggest it works since v4.4.3 🤔  but I think you have to declare the additional types [like in here](https://material-ui.com/customization/theming/#custom-variables) for TypeScript to stop complaining.

darrenvong: Maybe v5 has better generic types so this wouldn't be necessary once it's out of beta?

lucaslcode: You can but you can only access it from the theme object, not as a variant of typography. https://github.com/mui-org/material-ui/issues/22257

lucaslcode:
@darrenvong maybe this is all that's necessary for now? Anything else I should add in?

app/frontend/src/theme.ts
darrenvong: I guess custom variants won't work with this anyway, so having it as a custom property on the main theme is probably fine for now for keeping everything together?

app/frontend/src/components/Button/Button.tsx
```diff
@@ -46,15 +49,24 @@ export default function Button<D extends ElementType = "button", P = {}>({
       setWaiting(false);
     }
   }
+  //height is fontSize * lineHeight, but they use rem
+  //so just use a &nbsp; to get the right height for loading
```
darrenvong: What happens without these non breaking whitespace? Does the circular progress jump around?

darrenvong: Oh I see it from deleting the whitespace. Couldn't we enforce this with min height on the button itself? 🤔

lucaslcode: The problem is `theme.typography.button.fontSize` is in rem, so you can't multiply it cleanly...

lucaslcode: Or you mean just set it manually?

lucaslcode: I set `minHeight` in the latest commit. But I kinda prefer the old way, this way can break if the button font size is made bigger.

darrenvong: I think the height is actually `<number of lines> * line height`. If we explicitly set line height as well in the theme for buttons and get this `minHeight` referencing that, it'd fix the problem of having to update this in future?

lucaslcode: It can be either a multiplier or specified height. That would work as long as it's one line only (which it should be).

darrenvong: Even with the non breaking white space, it'd go back down to one line when the button is loading, so I think it's fine. What do you mean by the multiplier option? I thought `minHeight` only works with length units like `px`/`rem` etc?

lucaslcode: If you don't put any units, it's a multiplier of fontSize (I actually didn't know you could specify the height exactly)

lucaslcode: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

darrenvong: Oh I see, yeah usually the multiplier is best as it scales properly with the user options, but I think it should be ok if we use `rem` to set the line height and reference that in `minHeight`. It should scale together with the font size since they are referencing the same root value unlike the `em` example shown in the article.

darrenvong: Apparently Material UI got that set at 1.75, so I guess `1.75 * <font_size> rem` (works out to be `1.53125rem`) works?

lucaslcode:
Had to use `calc` to add the padding, then I realized I can just use two for the line height * font size.

darrenvong (approved):

PR #569, opened 2021-02-03
Page transfer and fix check constraints

aapeliv:


app/backend/src/couchers/servicers/pages.py
```diff
@@ -4,16 +4,17 @@
...
 def _check_update_permission(page: Page, user_id):
-    if page.owner_user_id == user_id:
-        return True
-    # elif page.owner_cluster_id # TODO
-    return False
+    if page.owner_user:
+        return page.owner_user_id == user_id
+    elif page.owner_cluster:
+        return page.owner_cluster.admins.filter(User.id == user_id).one_or_none() is not None
+    raise Exception("Page must have an owner.")
```
kalvdans: raise a `RuntimeError` if it is a programming mistake (preferably with an assert statement).

Can we add a database invariant that ensures a page has a owner?

aapeliv: There should be a database invariant. Maybe I can just remove the raise?

kalvdans: Yes and the elif line. Then we will get a `NoneType has no attribute admins` runtime error which is as good as an assert.

aapeliv:
I noticed the check constraints weren't done properly, so this PR also fixes that with a few more tests.

kalvdans (dismissed):
Looks good to me!

PR #571, opened 2021-02-03
Set fsync=off in gitlab CI

aapeliv:
Should speed up tests in CI a bit

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -137,6 +137,7 @@ test:backend:
   services:
     - name: postgis/postgis:13-3.0
       alias: postgres
+      command: postgres -c fsync=off
   variables:
     POSTGRES_PASSWORD: c765064a49d18a95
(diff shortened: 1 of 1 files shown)


aapeliv:
This seems to save about 2 minutes from backend tests!

Before: https://gitlab.com/couchers/couchers/-/jobs/1003340831
After: https://gitlab.com/couchers/couchers/-/jobs/1006393636

.gitlab-ci.yml
```diff
@@ -137,6 +137,7 @@ test:backend:
   services:
     - name: postgis/postgis:13-3.0
       alias: postgres
+      command: ["postgres", "-c", "fsync=off"]
```
kalvdans, line 140: `["postgres", "--fsync=off"]` also works (and "postgres", "-F" but maybe too short.

In my setup I also have `--synchronous_commit=off --full_page_writes=off` but they have probably less impact.

kalvdans (approved):
And it should save gitlab precious hard disk write cycles! (especially if they are using SSD)

PR #573, opened 2021-02-05
Run autoflake and remove unused imports and variables

aapeliv:
I had to carefully run the tests and add in any fixtures that need to be imported for pytest magic


kalvdans (approved):
Great february cleaning

PR #574, opened 2021-02-05
Pagination

lucaslcode:


lucaslcode:
@aapeliv what is `GetUpdates` for??

aapeliv:
Given a message id, it returns all messages after that id, across all chats. (I think.) So you could keep track of the latest message you've seen and always just hit `GetUpdates` to stay up to date on all messages. Though that requires dividing them up into the right group chats, etc.

lucaslcode:
Gotcha. I think we won't use it, react-query just invalidates the old stuff.

lucaslcode:
Backend portion of this is ready for review @Couchers-org/backend. See following

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -12,8 +12,10 @@
 from couchers.utils import Timestamp_from_datetime
 from pb import api_pb2, conversations_pb2, conversations_pb2_grpc
 
-# TODO: custom pagination length
+# TODO: Still needs custom pagination: GetUpdates
 PAGINATION_LENGTH = 20
+DEFAULT_LARGE_PAGINATION_LENGTH = 20
+DEFAULT_SMALL_PAGINATION_LENGTH = 10
```
lucaslcode: I thought it makes sense to have a larger default for messages than chats

lucaslcode (commented):
Basically, the pagination logic was pretty busted. Requests used `last_id` but responses gave `next_id` so you couldn't pass it directly. And most of the query filters were using `<` not `<=`, so it should have been `last_id` anyway.

Also listing host requests mixed up message id and request id (I think that was me :D)

Maybe we need to write better tests for the pagination logic, there are basically none at the moment.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -113,6 +115,8 @@ def _add_message_to_subscription(session, subscription, **kwargs):
 class Conversations(conversations_pb2_grpc.ConversationsServicer):
     def ListGroupChats(self, request, context):
         with session_scope() as session:
+            page_size = request.number if request.number != 0 else DEFAULT_SMALL_PAGINATION_LENGTH
```
aapeliv: Might be good to have a max value here, or similar

app/backend/src/couchers/servicers/conversations.py
aapeliv: Yeah, that's sensible, though maybe we could just do it on the frontend and set some reasonable maximum here.

Also, maybe worth either renaming to `DEFAULT_{MESSAGES/CONVERSATIONS}_PAGE_SIZE` or adding a comment as to when large/small is used?

aapeliv (commented):
The backend side looks decent, if it works :).

There's a whole lot of refactoring that could be done, but I think it's not worth concentrating on right now. We can wait until we got closer to real-time notifications, and we'll need to add some notification triggers in places. One of the main things is that we could rely the ORM more and use sqlalchemy relationships instead of writing all the joins and filters repeatedly.

I'm not entirely sure, but it might also be good to eventually move to the page size/token pagination style as well. That depends on how the frontend ends up using the messages though (whether it is very sophisticated and uses the last message ids, etc, or whether it just reloads everything and uses the message IDs as purely opaque pagination tokens).

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -120,12 +151,47 @@ export default function GroupChatView() {
...
+    scrollRef.current?.scrollBy(0, -60);
+    /*if (isFetchingNextPage || !scrollRef.current) return;
+    const newScrollHeight = scrollRef.current.scrollHeight;
+    if (newScrollHeight === prevScrollHeight.current) return;
+    scrollRef.current.scrollTo(
+      0,
+      newScrollHeight - (prevScrollHeight.current ?? newScrollHeight)
+    );
+    prevScrollHeight.current = newScrollHeight;*/
+  }, [isFetchingNextPage]);
```
lucaslcode: @darrenvong what do you think of this? It's difficult doing infinite load "upwards"! The current way is less janky than the commented-out one, but still not amazing. And I don't really like using id like that.

darrenvong: What happens with this when the next page loads? Does it scroll all the way to the oldest message on that page? Can't really think of a cleaner way to do this and have it scroll in smoothly, as I doubt we can do better than the browser with `scrollIntoView`. Would the `smooth` option help make this less janky too?

app/frontend/src/features/messages/Messages.tsx
```diff
@@ -43,28 +43,32 @@ export function HostRequestsSentNotification() {
...
   const header = (
     <>
       <PageTitle>Messages</PageTitle>
       <TabBar
         value={messageType}
         setValue={(newType) =>
-          history.push(`${messagesRoute}/${newType !== "all" ? newType : ""}`)
+          history.push(
+            `${messagesRoute}/${
+              newType !== "groupchats" ? newType : "groupchats"
```
lucaslcode:
```suggestion
              newType
```

app/frontend/src/App.tsx
```diff
@@ -15,20 +15,17 @@ import AuthProvider from "./features/auth/AuthProvider";
...
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles({
   padding: {
-    paddingBottom: theme.spacing(7),
+    paddingBottom: theme.shape.navPaddingMobile,
     paddingLeft: theme.spacing(2),
     paddingRight: theme.spacing(2),
-    [`${theme.breakpoints.up("xs")} and (orientation: landscape)`]: {
-      paddingBottom: theme.spacing(6),
-    },
```
lucaslcode, line 25: Am I correct to get rid of this? I don't see why landscape needs special treatment. Either the breakpoint is larger than md, and it uses the desktop padding, or it isn't, so it uses the mobile padding.

app/frontend/src/features/messages/Messages.tsx
darrenvong: What's happening with the other tabs? Are we commenting them out until we actually have content for them or are they going to be removed eventually?

app/frontend/src/App.tsx
darrenvong: Yeah I'm not sure what this was trying to achieve... let's remove it for now to simplify the CSS and we can always fix it if people report visual bugs

app/frontend/src/features/messages/Messages.tsx
lucaslcode: Yeah until they have content - at least "All" I'm not sure will be in the beta due to the current separation of host requests and messages.

app/frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -32,15 +34,24 @@ export interface AvatarProps extends BoxProps {
...
+  isProfileLink = true,
   ...otherProps
 }: AvatarProps) {
   const classes = useStyles();
+
+  const muiAvatar = (user: User.AsObject) => (
+    <MuiAvatar className={classes.avatar} alt={user.name} src={user.avatarUrl}>
+      {user.name.split(/\s+/).map((name) => name[0])}
+    </MuiAvatar>
+  );
```
darrenvong: Couldn't this be inlined in line 69? The JSX output seems simple enough to me that I think it'll be fine to do that

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -78,13 +100,21 @@ export default function GroupChatView() {
...
   const {
-    data: messages,
+    data: messagesRes,
     isLoading: isMessagesLoading,
     error: messagesError,
-  } = useQuery<Message.AsObject[], GrpcError>(
+    fetchNextPage,
+    isFetchingNextPage,
+    hasNextPage,
+  } = useInfiniteQuery<GetGroupChatMessagesRes.AsObject, GrpcError>(
```
darrenvong, line 110: Damn, I can't imagine how much trickier this will be if we were still using Redux. So nice this has barely changed the core of the code 🤩

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -0,0 +1,94 @@
...
+import CircularProgress from "@material-ui/core/CircularProgress";
+import classNames from "classnames";
+import React, { ReactNode, useCallback, useLayoutEffect, useRef } from "react";
+
+import useOnVisibleEffect from "../useOnVisibleEffect";
+import { messageElementId } from "./MessageView";
+
+const useStyles = makeStyles((theme) => ({
+  loader: {
+    //px for easier scroll compensation
```
darrenvong: Is this comment now outdated?

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const { ref: loadMoreRef } = useOnVisibleEffect(null, handleLoadMoreVisible);
+
+  useLayoutEffect(() => {
+    if (isFetchingNextPage) return;
+    const messageEl = document.getElementById(
+      messageElementId(prevTopMessageId.current ?? 0)
+    );
+    messageEl?.scrollIntoView();
+  }, [isFetchingNextPage, loadMoreRef]);
```
darrenvong:
```suggestion
  }, [isFetchingNextPage]);
```
Since `loadMoreRef` is meant to be stable? Also no lint warnings when removed so I think React knows it.

app/frontend/src/features/messages/useOnVisibleEffect.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from "react";
+import { useInView } from "react-intersection-observer";
+
+export default function useOnVisibleEffect<Param>(
+  param: Param,
+  onVisible?: (param: Param) => void
+) {
```
darrenvong:
```suggestion
export default function useOnVisibleEffect(
  onVisible?: () => void
) {
```
Now I'm seeing this being used the generic way, I think this might make more sense instead, since the components that use this can pass the info they pass through via `param` right now directly with a wrapper function like `() => onVisible(message.messageId)` etc

app/frontend/src/theme.ts
```diff
@@ -35,6 +44,28 @@ const themeOptions: ThemeOptions = {
...
+        width: "0.5rem",
+        background: "rgba(0,0,0,0.1)",
+      },
+      "&::-webkit-scrollbar-thumb": {
+        borderRadius: "20px",
+        background: "rgba(0,0,0,0.2)",
+      },
+      "&::-webkit-scrollbar-thumb:hover": {
+        background: "rgba(0,0,0,0.3)",
+      },
```
darrenvong, line 102: Does that mean it'll still look ugly on Firefox since this is WebKit specific CSS? 😢

darrenvong (changes requested):
Only some minor comments, but some solid work here mate, some tricky stuff for sure!

One potential concern that we may have to address - once the app gets some real usage, it's likely a chat can have up to hundreds if not thousands of messages, which could cause performance problems since it will be quite easy to break the recommended limit here at that point - https://web.dev/dom-size/. I'm looking at the no more than 1500 DOM nodes rather than 60 one since Fb Messenger seems to break this anyway lol. But maybe we'll cross that bridge once we get there? 🤷

app/frontend/src/theme.ts
lucaslcode: yes, but firefox supports `scrollbarWidth` at least

lucaslcode:
> a chat can have up to hundreds if not thousands of messages, which could cause performance problems

Yes, before I figured out the janky scrolling (I was using `useEffect` instead of `useLayoutEffect`), I tried using `react-virtuoso`, but performance was awful. `react-window` is the standard for virtualized lists, but it requires the item size to be known, not sure how we'd get around that.

darrenvong:
> `react-window` is the standard for virtualized lists, but it requires the item size to be known

Yeah I think there's an experimental version that never got completed by the author, although I think it only seems to support scrolling down. Considering both Messenger and WhatsApp web don't seem to render their stuff in virtual list and they've got a pretty big user base... maybe we can get away with this for now until we get more complaints about the chat being sluggish.

darrenvong (dismissed):

darrenvong (approved):

PR #576, opened 2021-02-06
Fix broken logout

darrenvong:
Setting the authenticated state to false first, which kicks the users out, before waiting for the next re-render to clear the user, seems to be the easiest fix to the problem.


aapeliv (approved):
LGTM

PR #577, opened 2021-02-06
Split up places and guides in GIS servicer.

aapeliv:
Also sends slug in geojson, and separates the layers in the frontend.

Adds a location to a guide in the dummy data


kalvdans (dismissed):
I'm not the best to review postgis/sqlalchemy wizardry or frontend stuff but nothing sticks out so I'm ok with merging :)

lucaslcode (dismissed):

Issue #543, opened 2021-01-30
Discussions


aapeliv:
This needs:

- [x] A (slight) redesign of the discussion cards in the communities page (Maybe just adding avatars)
- [x] A `Comment` component to display a comment with a reply button (and a text input that gets revealed?)
  - [x] Show replies
  - [x] Reply button reveals form that can be used to reply to that top level comment
- [x] ~~A `Reply` component, a reply to a comment that doesn't have a reply button (there is only 1 level of replies)~~
- [x] A `CommentTree` component that shows
  - [x] a list of comments
  - [x] all of the comments' replies
- [x] A page that lists all the discussions in a community
  - [x] ~~Refactor this later to share more markup with discussion section if it makes sense~~
  - [x] Tests
  - [x] Storybook
- [x] A page that shows the comments/replies for a single discussion
  - [x] Redo to match design
  - [x] Storybook
  - [x] Render single discussion with community markup - see [this PR discussion more details](https://github.com/Couchers-org/couchers/pull/1197#issuecomment-835881976)
- [x] A page for creating a new discussion in a community
  - [x] Functional form
  - [x] Styling
  - [x] Storybook
  - [x] Tests

These components should be in storybook

This will use #536 but can be done with normal text or the existing markdown editing component for now

lucaslcode:
Frontend barebones are done in #578, however it needs to be refactored to use `react-query` and `react-hook-forms`

PR #578, opened 2021-02-06
Discussions and comment section (threads)

aapeliv:


aapeliv:
Contrary to what I said in the meeting, I've instead gone the opposite and made the thread id compulsory. This makes the API a lot simpler to use, and allows using a truly "drop-in" comment section. I decided to give this a shot because the API for threads is such that it requires a lot of reworking to get the different entities that might have a comment section in there (because the thread_id is shared between the actual thread, comments and replies). In the end, an extra int foreign key is not very expensive.

This is up for debate, what do others think?

kalvdans:
what is the difference between a discussion and a thread? can we merge them into one entity in the API?

aapeliv:
> what is the difference between a discussion and a thread?

Discussion is a separate page for discussing things in a community or group. They will eventually be votable, I imagine, and have a bunch of other fancy features (for example, at the moment discussions have a slug and owners, threads don't). A thread is a "comment section", that can be embedded on any page.

> can we merge them into one entity in the API?

I think it's good to keep them separate, they will deviate over time, I believe.

darrenvong:
Is a thread a recursive data structure with how it's currently modelled? i.e. a reply to a thread is a thread itself? Sorry if these might have been answered in previous discussions, I should keep up more with backend discussions too...

aapeliv:
> Is a thread a recursive data structure with how it's currently modelled? i.e. a reply to a thread is a thread itself?

Yes, the API is recursive, in that you get and post onto `thread_id`s. But in fact there are only three levels: the root, comments, and replies. Replies therefore can't have replies on them.

darrenvong:
Right, so I guess we are abandoning the Reddit style replies where things can nest to crazy levels for simplicity? Probably a good idea as I personally find heavily nested threads quite hard to follow

aapeliv:
> Right, so I guess we are abandoning the Reddit style replies where things can nest to crazy levels for simplicity? Probably a good idea as I personally find heavily nested threads quite hard to follow

Yeah, that was the conclusion, I guess. I think the conclusion was that too deep nesting is just confusing!

app/frontend/src/components/Comments/CommentBox.tsx
```diff
@@ -0,0 +1,118 @@
...
+        );
+      } catch (e) {
+        console.error(e);
+        setError(e.message);
+      }
+      setLoading(false);
+    })();
+  }, [threadId]);
+
+  const onComment = async (threadId: number, content: string) => {
```
lucaslcode:
```suggestion
  const handleComment = async (threadId: number, content: string) => {
```

app/frontend/src/components/Comments/CommentBox.tsx
```diff
@@ -0,0 +1,118 @@
...
+  },
+}));
+
+interface CommentBoxProps {
+  threadId: number;
+}
+
+// Reply with more Reply objects as children
+interface MultiLevelReply extends Reply.AsObject {
+  children: Array<Reply.AsObject>;
```
lucaslcode: please change this to `replies` as children usually refers to react nodes.

app/frontend/src/components/Comments/CommentBox.tsx
```diff
@@ -0,0 +1,118 @@
...
+              <>
+                <Card className={classes.card}>
+                  Reply: by user id {reply.authorUserId}, posted at{" "}
+                  {reply.createdTime!.seconds}.
+                  <Markdown source={reply.content} />
+                </Card>
+              </>
+            ))}
+            <NewComment
+              onComment={(content) => onComment(comment.threadId, content)}
```
lucaslcode:
```suggestion
              onComment={(content) => handleComment(comment.threadId, content)}
```
Convention is the event is `onEvent` and the function is `handleEvent` (though are other places in the code which don't currently reflect this)

app/frontend/src/components/Comments/NewComment.tsx
```diff
@@ -0,0 +1,59 @@
...
+interface NewCommentProps {
+  onComment: (comment: string) => Promise<void>;
+}
+
+export default function NewComment({ onComment }: NewCommentProps) {
+  const classes = useStyles();
+  const [preview, setPreview] = useState(false);
+  const [comment, setComment] = useState("");
+
+  const onSave = async () => {
```
lucaslcode:
```suggestion
  const handleSubmit= async () => {
```

app/frontend/src/components/Comments/NewComment.tsx
```diff
@@ -0,0 +1,59 @@
...
+            </Button>
+          </Box>
+        </Grid>
+        {preview && (
+          <Grid item xs={12} md={6}>
+            <Markdown source={comment} />
+          </Grid>
+        )}
+      </Grid>
+      <Button onClick={onSave}>Comment</Button>
```
lucaslcode:
```suggestion
      <Button onClick={handleSubmit} type="submit">Comment</Button>
```
Not 100% sure about `type="submit"` but do it if it lets you (good for a11y)

lucaslcode (changes requested):
Just some minor style things (I'm assuming you clicked around to test)

aapeliv:
@lucaslcode: fixed those comments. And yep, was just testing manually...

app/frontend/src/components/Comments/NewComment.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <>
+      <p>Write a comment:</p>
+      <Grid container spacing={2}>
+        <Grid item xs={12} md={preview ? 6 : 12}>
+          <TextField
+            label="Text field"
+            rowsMax={5}
+            multiline
+            fullWidth
+            onChange={(e) => setComment(e.target.value)}
```
darrenvong, line 37: We should wrap this in a form and do it with `react-hook-form` like in other places, since it optimises for state change like this that otherwise would make this re-render on every keystroke and cause performance issues.

app/frontend/src/components/Comments/CommentBox.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const [error, setError] = useState("");
+
+  const [comments, setComments] = useState<Array<MultiLevelReply>>([]);
+
+  useEffect(() => {
+    (async () => {
+      setLoading(true);
+      try {
+        const thread = await service.threads.getThread(threadId);
+        setComments(
```
darrenvong, line 42: Probably ok for now... but definitely should do this in React Query to avoid having to deal with all the refetching ourselves. Is this meant to be working with pagination as well already?

darrenvong (changes requested):

app/frontend/src/components/Comments/CommentBox.tsx
lucaslcode: > Is this meant to be working with pagination as well already?

Seems not, that's easier with react-query anyway

app/frontend/src/components/Comments/NewComment.tsx
lucaslcode: @aapeliv if you make a note on #543 that this needs to be converted to react query and use react-hook-form, you can leave it for us to do

lucaslcode:
I'm happy to approve this pending backend review. I will leave a note in #543 that we need to use `react-query` and `react-hook-forms`.

aapeliv:
Yeah, the frontend is crappy (and also ugly). I'm hoping someone who actually knows frontend could fix this up in another PR, but wanted to do an "MVP" so the API calls are clear, etc, and hopefully it speeds that up!

I'm happy to also fix it up myself if yous want (though I can't make it pretty)!

app/backend/src/couchers/servicers/communities.py
```diff
@@ -203,5 +206,19 @@ def ListEvents(self, request, context):
         return communities_pb2.ListEventsRes()
 
     def ListDiscussions(self, request, context):
-        raise NotImplementedError()
-        return communities_pb2.ListDiscussionsRes()
+        with session_scope() as session:
+            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
+            next_page_id = int(request.page_token) if request.page_token else 0
```
kalvdans, line 211: Maybe a nicer error message if page_token can't be converted to integer. Can be done later though.

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -0,0 +1,74 @@
...
+
+def discussion_to_pb(discussion: Discussion, user_id):
+    return discussions_pb2.Discussion(
+        discussion_id=discussion.id,
+        slug=discussion.slug,
+        created=Timestamp_from_datetime(discussion.created),
+        creator_user_id=discussion.creator_user_id,
+        owner_group_id=discussion.owner_cluster.id
+        if discussion.owner_cluster.official_cluster_for_node_id is None
+        else None,
```
kalvdans: Ugly formatting, can you precalculate on a separate line or put parentheses around?

app/backend/src/tests/test_communities.py
```diff
@@ -139,6 +151,19 @@ def _create_place(token, title, content, address, x):
         )
 
 
+def _create_discussion(token, community_id, group_id, title, content):
```
kalvdans: Drop the leading _, this is a test file anyway not exporting any functions.

app/backend/src/tests/test_discussions.py
```diff
@@ -0,0 +1,169 @@
...
+        res = api.GetDiscussion(
+            discussions_pb2.GetDiscussionReq(
+                discussion_id=discussion_id,
+            )
+        )
+
+        assert res.title == "dummy title"
+        assert res.content == "dummy content"
+        assert res.slug == "dummy-title"
+        assert to_aware_datetime(res.created) > now() - timedelta(seconds=10) and to_aware_datetime(res.created) < now()
```
kalvdans: Use Python's chained comparisons and have a more generous time limit, we don't want random failures because of loaded servers. You could set `test_start_time = now()` before creating the discussion and then change this to
```
assert test_start_time <= to_aware_datetime(res.created) <= now()
```

app/pb/communities.proto
```diff
@@ -68,6 +69,7 @@ message Community {
   uint32 member_count = 10;
   uint32 admin_count = 11;
   uint32 nearby_user_count = 12;
+  int64 thread_id = 13;
 }
```
kalvdans, line 73: Would be nice to pass the number of comments here too, but could be added later.

app/pb/discussions.proto
```diff
@@ -0,0 +1,44 @@
...
+  string slug = 2;
+  google.protobuf.Timestamp created = 3;
+  int64 creator_user_id = 4;
+  oneof owner {
+    int64 owner_group_id = 5;
+    int64 owner_community_id = 6;
+  }
+  string title = 7;
+  string content = 8;
+  int64 thread_id = 9;
```
kalvdans, line 30: Would be nice to pass number of comments here too, can be done later.

app/pb/groups.proto
```diff
@@ -75,6 +76,7 @@ message Group {
   bool admin = 9;
   uint32 member_count = 10;
   uint32 admin_count = 11;
+  int64 thread_id = 12;
```
kalvdans, line 79: Likewise, pass number of comments but not necessary now.

kalvdans (dismissed):
Backend part approved, some minor comments. You hereby get my auto-approval, no need to ask for another review if you fix those.

app/backend/src/couchers/servicers/communities.py
aapeliv: I think it's fine like this, it'll be clear from the logs and the exceptions are sanitized anyway at the end. Adding the check everywhere would be annoying and bloat the code, IMO. But we can refactor the page token stuff eventually to have shared helper functions or something, in which case it of course is worth it.

It's (almost certainly) a problem in something we're doing (or someone is poking around on the API) if a non-int ends up here. For the user, it'll just show up as "unknown internal error" or similar.

app/backend/src/couchers/servicers/discussions.py
aapeliv: Agree that it's ugly, I'll see if I can make black format it nicer with parentheses.

aapeliv: Fixed with precomputation

app/backend/src/tests/test_discussions.py
aapeliv: Thanks, great idea. I've cleaned it up

aapeliv:
I've fixed up a bunch of stuff. Great idea with the timing.

> You hereby get my auto-approval

Thanks :D

Regarding getting number of comments: I think the API requires some refactoring, I want to push that out until a bit later when we see a bit more what the access patterns are, etc. At the moment the frontend has to make too many API calls for each page, and we need to figure out a long-term solutions for that. I've some ideas, but I don't want to jump in and guess, I want to have a good think about it.

PR #581, opened 2021-02-06
Bug reporting

### LINKED ISSUE #292: Bug reporting
For users to report bugs from the app. It magically opens a github issue in the main repo like this. Protobuf definition:

```
syntax = "proto3";

package org.couchers.bugs;

import "google/protobuf/empty.proto";

service Bugs {
  rpc Version(google.protobuf.Empty) returns (VersionInfo) {
    // Returns backend version info
  }

  rpc ReportBug(ReportBugReq) returns (ReportBugRes) {
    // Report a bug
  }
}

message VersionInfo {
  string version = 1;
}

message ReportBugReq {
  string subject = 1;
  string description = 2;
  string steps = 3;
  string results = 4;

  string frontend_version = 5;
  string user_agent = 6;
  string page = 7;
  uint32 user_id = 8;
}

message ReportBugRes {
  string report_identifier = 1;
}
```

darrenvong:
This is pretty much ready apart from styling the button to be next to the search bar on the right, but maybe that's best done after the #568 is merged first as I suspect I'll get some merge conflicts here...

Closes #292 once this is done.


app/frontend/src/features/BugReport.tsx
```diff
@@ -0,0 +1,147 @@
...
+                report was sent to the devs! The bug ID is {reportIdentifier}
+              </Alert>
+            )}
+            <DialogContentText>
+              Please note that this information, as well as diagnostic
+              information including which page you are on, what browser you are
+              using, and your user ID will be saved to a public list of bugs.
+            </DialogContentText>
+            <TextField
+              id="bug-report-subject"
```
darrenvong, line 115: A good thing I learnt while I was writing tests for this - we need to add an `id` for these text fields for material UI to link up text field labels properly so they remain accessible for screen readers. +1 on React Testing Library for encouraging these low hanging fruits to make sure I do it right!

aapeliv:
I tried it out, it's great! Except you get no feedback when you click "Submit" (doesn't turn gray or start loading), so I accidentally made 4 issues, spamming the button, thinking it wasn't submitting :P

darrenvong:
~~I thought the button is meant to go into a loading spinner when it's being clicked 🤔  maybe because the function is not actually async. Gonna add in the loading state from the mutation and see if it makes a difference~~

Fixed by adding in said loading state from mutation! It will be less ugly once rebased with Lucas' change in #568

app/frontend/src/features/BugReport.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+async function fillInAndSubmitBugReport(
+  subjectFieldLabel: string,
+  descriptionFieldLabel: string,
+  stepsFieldLabel: string = "",
+  resultsFieldLabel: string = ""
+) {
+  const subjectField = await screen.findByLabelText(subjectFieldLabel);
+  const descriptionField = await screen.findByLabelText(descriptionFieldLabel);
+  const stepsField = screen.queryByLabelText(stepsFieldLabel);
+  const resultsField = screen.queryByLabelText(resultsFieldLabel);
```
lucaslcode, line 32: Why do these have different functions?

app/frontend/src/features/BugReport.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+      expect(reportBugButton).not.toHaveTextContent("Report a bug");
+    });
+  });
+
+  describe('when the "report a bug" button is clicked', () => {
+    const subjectFieldLabel = "Brief description of the bug";
+    const descriptionFieldLabel = "What's the problem?";
+    const stepsFieldLabel = "What did you do to trigger the bug?";
+    const resultsFieldLabel =
+      "What happened? What did you expect should have happened?";
```
lucaslcode, line 106: Maybe not now, but do you think we should use test ids or something that's more stable than labels? They might not even statically exist later if we use translations.

app/frontend/src/features/BugReport.tsx
lucaslcode: Should we open a low priority bug to add ids to all textFields then?  / Make it a required prop?

app/frontend/src/features/BugReport.tsx
```diff
@@ -0,0 +1,173 @@
...
+    register,
+    handleSubmit,
+    reset: resetForm,
+  } = useForm<BugReportFormData>();
+  const userId = useAuthContext().authState.userId;
+  const {
+    data: reportIdentifier,
+    error,
+    isLoading,
+    mutate: reportBug,
```
lucaslcode, line 63: there is also a `mutateAsync` function if you want a promise (I think you mentioned that?)

lucaslcode (approved):
This is great, and good testing example to follow :D

app/frontend/src/features/BugReport.test.tsx
darrenvong: `findBy*` throws an error if the text field can't be found, whereas `queryBy*` returns null. It's only needed because I'm trying to make this reusable and for some test cases, I'm not filling in the `stepsField` and `resultsField` on purpose just to see if submit still works since they are optional.

app/frontend/src/features/BugReport.test.tsx
darrenvong: This is an interesting one... the author of React Testing Library [advise to still assert by text rather than using test IDs](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text) even in a multi-lingual app so we can ensure at least one language is being translated correctly.

But yeah, definitely something worth re-visiting if this becomes a pain as I don't feel as strongly as him on this. The Twitter thread he linked to on that paragraph also suggests some good solution on selecting by the translation key text, run the translation function on it (which we can probably set to English in the tests) and assert the result of that instead.

app/frontend/src/features/BugReport.tsx
darrenvong: Yeah that sounds like a good idea 👍

app/frontend/src/features/BugReport.tsx
darrenvong: But then we'd have to handle errors in a try-catch. I prefer the non-async version (but still async under the hood) but with all the logic done in the `on*` callbacks to keep everything together in the `useMutation` so all the logics are nicely colocated, so I think I'll leave this as it is.

app/frontend/src/features/BugReport.test.tsx
lucaslcode: Yeah I just don't like the idea of having to open multiple files to make a label change. But with translation keys it makes sense.

PR #588, opened 2021-02-07
Show that the enum values are unimportant

kalvdans:
Show that the numerical enum values are unimportant to sqlalchemy by assigning them
with enum.auto().


aapeliv (approved):
LGTM contingent on tests.

PR #589, opened 2021-02-07
Accept non-existent session token in auth.Deauthenticate.

aapeliv:
Also sends an empty cookie expiring now() to remove it from the browser as an extra.

### DIFF
--- a/app/backend/src/couchers/servicers/auth.py
+++ b/app/backend/src/couchers/servicers/auth.py
@@ -21,7 +21,7 @@ from couchers.interceptors import AuthValidatorInterceptor
 from couchers.models import LoginToken, PasswordResetToken, SignupToken, User, UserSession
 from couchers.servicers.api import hostingstatus2sql
 from couchers.tasks import send_login_email, send_password_reset_email, send_signup_email
-from couchers.utils import create_session_cookie, parse_session_cookie
+from couchers.utils import create_session_cookie, now, parse_session_cookie
 from pb import auth_pb2, auth_pb2_grpc
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


aapeliv:
I've also cleaned the tests a little bit so they give nicer error messages... just because...

app/backend/src/tests/test_auth.py
```diff
@@ -204,4 +204,35 @@ def test_password_reset_invalid_token(db, fast_passwords):
...
+
+    reply_token = get_session_cookie_token(metadata_interceptor)
+
+    # delete all login tokens
+    with session_scope() as session:
+        session.query(LoginToken).delete()
+
+    # log out with non-existent token should still return a valid result
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Deauthenticate(empty_pb2.Empty(), metadata=(("cookie", f"couchers-sesh={reply_token}"),))
```
kalvdans: unused variable; test `reply` or remove the assigment

app/backend/src/couchers/servicers/auth.py
```diff
@@ -354,11 +354,18 @@ def Deauthenticate(self, request, context):
...
-            # probably caused by token not existing
-            context.abort(grpc.StatusCode.UNKNOWN, errors.LOGOUT_FAILED)
+        # if we had a token, try to remove the session
+        if token:
+            self._delete_session(token)
+
+        # set the cookie to an empty string and expire immediately, should remove it from the browser
+        context.send_initial_metadata(
+            [
+                ("set-cookie", create_session_cookie("", now())),
```
kalvdans, line 364: Would be easier if we used Max-Age insead of Expires, but that's a separate change.

kalvdans (approved):
LGTM!
I don't think there was any issue to close.

PR #598, opened 2021-02-08
Annotate markdown fields in API and database

kalvdans:
Patch that sprinkles "CommonMark without images" everywhere, feel free to thumbsdown if you think it pollutes the files rather than helping.

Not sure if references are to allowed to be Markdown formatted or just a string of words.


lucaslcode:
Maybe this should wait until after #529 to avoid merge conflicts

app/backend/src/couchers/models.py
```diff
@@ -534,7 +534,7 @@ class Message(Base):
     # time sent, timezone should always be UTC
     time = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
 
-    # the message text if not control
+    # the CommonMark-compliant message text if not control
```
aapeliv: I believe the group chat messages are plaintext. Is that right @lucaslcode?

app/backend/src/couchers/models.py
```diff
@@ -394,7 +394,7 @@ class Reference(Base):
 
     reference_type = Column(Enum(ReferenceType), nullable=False)
 
-    text = Column(String, nullable=True)
+    text = Column(String, nullable=True)  # CommonMark without images
```
aapeliv: The reference also is currently plaintext, I think

app/pb/api.proto
```diff
@@ -299,7 +299,7 @@ message Reference {
   int64 from_user_id = 1;
   int64 to_user_id = 2;
   ReferenceType reference_type = 3;
-  string text = 4;
+  string text = 4;  // CommonMark without images
```
aapeliv: As above

app/pb/api.proto
```diff
@@ -322,7 +322,7 @@ message AvailableWriteReferenceTypesRes {
 message WriteReferenceReq {
   int64 to_user_id = 1;
   ReferenceType reference_type = 2;
-  string text = 3;
+  string text = 3;  // CommonMark without images
```
aapeliv: As above

app/pb/conversations.proto
```diff
@@ -109,7 +109,7 @@ message Message {
   uint64 author_user_id = 2;
   google.protobuf.Timestamp time = 3;
   oneof content {
-    MessageContentText text = 4;
+    MessageContentText text = 4;  // CommonMark without images
```
aapeliv: As above

aapeliv (changes requested):

aapeliv:
> Maybe this should wait until after #529 to avoid merge conflicts

#529 has now been merged.

kalvdans:
> #529 has now been merged.

Thanks for the reminder! Merged latest develop and fixed your comments.

app/pb/conversations.proto
aapeliv: I think this is still plain text

aapeliv (changes requested):

aapeliv (approved):
LGTM!

PR #607, opened 2021-02-08
Separate docker-compose into dev and test, and update docs

### LINKED ISSUE #518: Simplify docker-compose setup
Currently there's too many containers in the main `docker-compose.yml` file. Some have reported that it's very slow and sluggish, and it confuses new devs. We should move the testing stuff and media container to another file. It also takes forever to run due to building yarn stuff inside docker.

Does anyone actually run the frontend inside docker?

aapeliv:
Closes #518


kalvdans (approved):
I don't use docker but the change seems ok.

PR #612, opened 2021-02-09
fix: saving simplified user address

### LINKED ISSUE #559: Address clean up from Nominatim
In the legacy vue frontend, we stripped some parts of the address from Nominatim:

https://github.com/Couchers-org/couchers/blob/f95e212833a84d9c470e73530f78587676befc02/app/vue/src/components/EditableLocation.vue#L316-L334

This makes the display addresses a bunch nicer, so instead of:

```
Forbes Street, Macdonaldtown, Newtown, Sydney, Council of the City of Sydney, New South Wales, 2006, Australia
500, West 120th Street, Morningside Heights, Manhattan Community Board 9, Manhattan, New York County, New York, 10027, United States
```

We have:

```
Macdonaldtown, Sydney, Australia
Morningside Heights, Manhattan, New York, United States
```

Tay0108:
Closes #559.

### DIFF
--- a/app/frontend/src/components/MapSearch.tsx
+++ b/app/frontend/src/components/MapSearch.tsx
@@ -3,6 +3,7 @@ import { AutocompleteChangeReason } from "@material-ui/lab/Autocomplete";
 import { LngLat } from "maplibre-gl";
 import React, { useState } from "react";
 
+import { NominatimPlace, simplifyPlaceDisplayName } from "../utils/nominatim";
 import Autocomplete from "./Autocomplete";
 import { SearchIcon } from "./Icons";
 
@@ -38,6 +39,7 @@ const useSearchStyles = makeStyles((theme) => ({
 
 interface SearchOption {
   name: string;
+  simplifiedName: string;
   location: LngLat;
 }
 
@@ -68,8 +70,9 @@ export default function MapSearch({
       return;
     }
     setSearchOptionsLoading(true);
-    const url =
-      NOMINATIM_URL! + "search?format=jsonv2&q=" + encodeURIComponent(value);
+    const url = `${NOMINATIM_URL!}search?format=jsonv2&q=${encodeURIComponent(
+      value
+    )}&addressdetails=1`;
     const options = {
       method: "GET",
       headers: {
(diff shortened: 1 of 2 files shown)


app/frontend/src/components/MapSearch.tsx
```diff
@@ -78,13 +81,17 @@ export default function MapSearch({
...
+
       setSearchOptions(
-        data.map((obj) => ({
-          name: obj["display_name"],
-          location: new LngLat(Number(obj["lon"]), Number(obj["lat"])),
-        }))
+        nominatimResults.map((result) => {
+          return {
+            name: result["display_name"],
+            simplifiedName: simplifyPlaceDisplayName(result),
```
lucaslcode, line 91: This is much better! I would be up for just ditching the full name and using only `name: simplifyPlaceDisplayName(result)` so the list of options has the better-looking name too. Up to you!

app/frontend/src/utils/nominatim.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const simplifyPlaceDisplayName = (place: NominatimPlace) => {
+  const addressFields = [
+    "neighbourhood",
+    "suburb",
+    "city",
+    "country",
+    "town",
+    "village",
```
lucaslcode:
```suggestion
    "town",
    "village",
    "neighbourhood",
    "suburb",
    "city",
    "state",
    "country",
```
How about this? Town and village should be first, and in many places like the US and Australia including the state is important. I just hope it is omitted automatically by nominatim where appropriate😓

lucaslcode (changes requested):
Nice job

app/frontend/src/utils/nominatim.ts
lucaslcode: Are villages smaller than towns? Maybe that should be swapped.

Tay0108: From what I know villages are smaller ;) Will swap it ;)

app/frontend/src/components/MapSearch.tsx
Tay0108: Wanted to do that but the problem with this approach is that nominatim is still doing search on fields we are not including. The result could be e.g. displaying "Sela" after typing "fa" because there is railway containing "fa". It could be missleading for users. There is a`zoom` query param which we can use to narrow search (https://nominatim.openstreetmap.org/ui/reverse.html?format=html&lat=51.484622236435&lon=8.337727189064&zoom=14) but still there would be cases that there is something we don't want in options list.
image

app/frontend/src/utils/nominatim.ts
Tay0108: Should the city be in front of neighbourhood & suburb?

app/frontend/src/components/MapSearch.tsx
Tay0108: Or maybe we can do filtering on our end?

aapeliv: In the vue frontend, the autocomplete/search list was raw nominatim output, but then in the automatically generated "display address" box we'd put the culled one

app/frontend/src/utils/nominatim.ts
lucaslcode: Cities are bigger imo

app/frontend/src/components/MapSearch.tsx
lucaslcode: I see, perhaps leave this then

lucaslcode (approved):

PR #614, opened 2021-02-10
feat: allow users to paste csv into profileTagInput

### LINKED ISSUE #262: "countries I've visited" won't accept comma-separated list
Subject: "countries I've visited" won't accept comma-separated list
Description:
Updating "countries I've visited" by pasting a comma-separated list of countries does not result in individual country tags for each comma-separated value, but instead creates one single long tag.

Steps:
I copied and pasted the following list: Austria, Belgium, Canada, Costa Rica, Czech Republic, France, Germany, Gibraltar, Greece, Hong Kong, Hungary, Ireland, Italy, Jamaica, Japan, Liechtenstein, Luxembourg, Mexico, Morocco, Netherlands, New Zealand, Portugal, Puerto Rico, Spain, Switzerland, Turkey, United Kingdom, United States, Virgin Islands, British, Virgin Islands, U.S.

Results:
When I pasted that list, it created one single, long tag. It should have created one country tag for each comma-separated value. Perhaps a country table should be used for people to select from (so people can't make up countries). We will need a table of all possible countries for the platform anyway, so perhaps we could query that table for valid/recognized values.

Backend version: cf99b0d
Frontend version: cf99b0d
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:81.0) Gecko/20100101 Firefox/81.0
Page: https://app.couchers.org/profile
User ID: 57

Tay0108:
Closes #262.

If you have any ideas how to do it better please let me know because I'm not 100% satisfied with it 😅

### DIFF
--- a/app/frontend/src/features/profile/ProfileTagInput.tsx
+++ b/app/frontend/src/features/profile/ProfileTagInput.tsx
@@ -205,20 +205,20 @@ export default function ProfileTagInput({
               // For some reason I came across situations when there were undefined values in this array.
               newValue = newValue.filter((element) => element !== undefined);
 
-            if (allowCsv) {
-              const lastIndex = newValue.length - 1;
-              const latestEntry = newValue[lastIndex];
-              const previousEntries = newValue.slice(0, lastIndex);
-              uniqueValues = new Set([
-                ...previousEntries,
-                ...latestEntry.split(",").map((value) => value.trim()),
-              ]);
+              if (allowCsv) {
+                const lastIndex = newValue.length - 1;
+                const latestEntry = newValue[lastIndex];
+                const previousEntries = newValue.slice(0, lastIndex);
+                uniqueValues = new Set([
+                  ...previousEntries,
+                  ...latestEntry.split(",").map((value) => value.trim()),
+                ]);
+              } else {
+                uniqueValues = new Set(newValue);
+              }
             } else {
-              uniqueValues = new Set(newValue);
+              uniqueValues = new Set([]);
             }
-          } else {
-            uniqueValues = new Set([]);
-          }
             setPendingValue(
               Array.from(uniqueValues).filter((value) => !/^\s*$/.test(value))
             );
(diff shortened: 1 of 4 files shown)


app/frontend/src/features/profile/ProfileTagInput.tsx
```diff
@@ -120,6 +121,7 @@ export default function ProfileTagInput({
   options,
   label,
   id,
+  allowCsv,
```
lucaslcode:
```suggestion
  allowCsv = true,
```
If it's true by default, you don't have to add it to each field (I can't see any reason we wouldn't want it most of the time).

app/frontend/src/features/profile/ProfileTagInput.tsx
```diff
@@ -198,7 +200,19 @@ export default function ProfileTagInput({
...
+              const latestEntry = newValue[lastIndex];
+              const previousEntries = newValue.slice(0, lastIndex);
+              uniqueValues = new Set([
+                ...previousEntries,
+                ...latestEntry.split(",").map((value) => value.trim()),
+              ]);
+            } else {
+              uniqueValues = new Set(newValue);
+            }
+            setPendingValue(Array.from(uniqueValues));
```
lucaslcode:
```suggestion
            setPendingValue(Array.from(
              //remove whitespace-only values
              uniqueValues.filter((value) => !/^\s*$/.test(value))
            ));
```
This is extra but good to have in there

lucaslcode (dismissed):
Optional changes

Tay0108:
Btw. there is a reason that I didn't merge it yet - I've found a bug (existing on current develop version) that if you press enter on an empty input it removes entries that were already there. Working on fix.

lucaslcode (changes requested):
Looks good but run `yarn format`, there's some funny indentation. (Personally I set up vs code to run prettier every time I save so I don't have to do that.)

Tay0108:
> Looks good but run `yarn format`, there's some funny indentation. (Personally I set up vs code to run prettier every time I save so I don't have to do that.)

I was sure I did that! Also, shouldn't we have some .prettierrc file? 

Usually I use key shortcuts to do that but for some reason they are not working for me on this project in both VS Code & Webstorm.

lucaslcode (approved):
> shouldn't we have some .prettierrc file

We could but we're just using the defaults atm

PR #617, opened 2021-02-11
Backend/feature/usr hiding

### LINKED ISSUE #610: User hiding
Users should be hidden (basically from all queries) if:

* the user is deleted, or
* the user is banned, or
* the user is jailed

This should be done through by filtering with a `hybrid_property` in queries with SQLAlchemy. Similar to the way in which tokens are filtered with `is_valid`:

https://github.com/Couchers-org/couchers/blob/910e9cb0cb8d11d01fca810c816e9e687922dfee/app/backend/src/couchers/models.py#L263-L265

Filtered e.g:

https://github.com/Couchers-org/couchers/blob/66b0ea2ee6211f19e74d52fb0d7ea7788b9782d1/app/backend/src/couchers/servicers/auth.py#L177-L182

This is a global property of a user, separate from user blocking (one user hiding another user, #414)

ShmuelTreiger:
Added a rudimentary user deletion
Added is_hidden hybrid field

Unsure how to proceed. Do I need to find every query for users and filter for hidden users? Or is there one place I can insert a universal filter? 

@aapeliv


app/backend/src/couchers/models.py
```diff
@@ -131,6 +132,10 @@ class User(Base):
     def is_jailed(self):
         return self.accepted_tos < 1 or self.is_missing_location
 
+    @hybrid_property
+    def is_hidden(self):
+        return self.is_banned and self.is_jailed and self.is_deleted
```
aapeliv: These should be ORs, but you need to use the `|` (bitwise) OR so that sqlalchemy can do its magic.

aapeliv (changes requested):
The user shouldn't be able to delete themselves via the API, they need to contact support and request it for the time being.

> Do I need to find every query for users and filter for hidden users?

Yep, that's the main thing. Have to go through everywhere where users are queried and filter out any that are hidden.

ShmuelTreiger:
> The user shouldn't be able to delete themselves via the API, they need to contact support and request it for the time being.

It seems to me this may be something we will want to eventually allow (i.e. click through a couple of pages, type a confirmation message, and click delete), in which case I think we should leave it.

Also, will support just go into the database manually to delete them? Or do I need to add some sort of interface for us to use? It seems to me that, given we will eventually want to make deletion more granule (i.e. delete profile, but keep pages and reviews, etc.) that we should start some sort of delete function.

> Yep, that's the main thing. Have to go through everywhere where users are queried and filter out any that are hidden.

Got it. I'll get working on it.

app/backend/src/couchers/models.py
ShmuelTreiger: oops. duh

ShmuelTreiger: I just re-read what you wrote and realized that I completely misunderstood. Fixing now.

ShmuelTreiger:
Replaced with backend/feature/user-hiding

PR #619, opened 2021-02-11
Add RPC to get account info

aapeliv:
This is needed so the frontend knows whether to ask the user for a password or not when setting a new password.

### DIFF
--- a/app/pb/account.proto
+++ b/app/pb/account.proto
@@ -9,6 +9,9 @@ service Account {
   // This is a secure service: a user needs to be authenticated and not jailed to call functions here, refer to auth.proto and jail.proto
 
   // account management APIs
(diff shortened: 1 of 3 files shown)


lucaslcode (approved):

app/pb/account.proto
```diff
@@ -22,6 +25,18 @@ service Account {
   }
 }
 
+message GetAccountInfoRes {
+  enum LoginMethod {
+    MAGIC_LINK = 0;
+    PASSWORD = 1;
+  }
+
+  LoginMethod login_method = 1;
```
kalvdans, line 34: Can't an account have both login methods enabled?

aapeliv: I don't think it's a good idea security-wise. A user would presumably either want link or password login, an attacker would want the weakest :P

PR #622, opened 2021-02-12
feat: adding proper title & favicon

Tay0108:
It annoys me that we are using React app title with default favicon so I thought I'll change that ;) If that helps you guys distingush it from other tabs then I completely understand it and this can be closed 😅


lucaslcode:
@aapeliv will be happy, it was annoying him too

lucaslcode (approved):

Issue #368, opened 2020-12-03
Account management frontend


lucaslcode:
Implement these rpcs

```
rpc ResetPassword(ResetPasswordReq) returns (google.protobuf.Empty) {
    // Sends a forgot password email to the given user if the user exists, returns no output (so you can't go around guessing email addresses)
  }

  rpc CompletePasswordReset(CompletePasswordResetReq) returns (google.protobuf.Empty) {
    // Triggered when the user goes to the link sent in the forgot password email
  }

  rpc CompleteChangeEmail(CompleteChangeEmailReq) returns (google.protobuf.Empty) {
    // When a user opens a changed email confirmation, the frontend should call this function with the token
  }
```

To-do list:
- [x] ResetPassword
- [x] CompletePasswordReset
- [x] CompleteChangeEmail
- [x] ChangePassword
  - [x] Improve UX by checking if user has password with GetAccountInfo
- [x] ChangeEmail

Edit: and below

aapeliv:
Also these?

https://github.com/Couchers-org/couchers/blob/0342766b1e21a2c142f93146f90b266c4d0131bf/app/pb/account.proto#L13-L21

aapeliv:
Superset of #211

darrenvong:
I don't think there are designs for these either? I can do this next. Guess I'll have to make it up and adapt based off something like the edit profile form (but a lot simpler)...

darrenvong:
How do you get a change email token? Is that the same token you get from `ResetPassword`?

aapeliv:
> How do you get a change email token? Is that the same token you get from `ResetPassword`?

No, it's from `account.proto`:

https://github.com/Couchers-org/couchers/blob/0342766b1e21a2c142f93146f90b266c4d0131bf/app/pb/account.proto#L18-L21

Issue #501, opened 2021-01-11
URL plan


aapeliv:
We need to agree on a consistent URL convention.

aapeliv:
Here's my proposal (as examples).

## Sydney example

Title: Sydney
Explanation: Page for the Sydney Community
URL: https://couchers.org/community/14362/sydney-nsw-australia
Crumbs: Australia > NSW > Sydney

Title: Sydney Opera House
Explanation: A "place"/"point of interest" page
URL: https://couchers.org/place/39293/sydney-opera-house-australia
Crumbs: Australia > NSW > Sydney > Sydney Opera House

Title: Best Food in Sydney
Explanation: A guide to food in Sydney
URL: https://couchers.org/guide/72833/best-food-in-sydney
Crumbs: Australia > NSW > Sydney > Best Food in Sydney

Title: Sydney Art Tour
Explanation: An event for an art tour around Sydney
URL: https://couchers.org/event/31312/sydney-art-tour
Crumbs: Australia > NSW > Sydney > Sydney Art Tour

Title: Best Views in Sydney for NYE fireworks?
Explanation: A reddit-style dicussion page about best place to view fireworks
URL: https://couchers.org/discussion/85548/best-views-in-sydney-for-nye-fireworks
Crumbs: Australia > NSW > Sydney > Best Views in Sydney for NYE fireworks?

Title: East Coast Hitchhikers
Explanation: A hitchhiking-focussed subgroup of the Australia community
URL: https://couchers.org/group/18421/east-coast-hitchhikers-australia
Crumbs: Australia > East Coast Hitchhikers

kalvdans:
What are Crumbs?

aapeliv:
Like little links to "parents" under the title, or similar. They're maybe more often called breadcrumbs I think.

You can see them on the landing page as well, above the title: https://couchers.org/issues/reviews/. They provide a navigation hierarchy.

Here it's more a list of "parents" which should have a link to the next child and provide one way to get to this content.

(I think it's really important that each piece of content we have lives in the hierarchy and has a well-defined place: that allows people to build a picture of where stuff is and easily find content they need.)

aapeliv:
> Title: Sydney Art Tour
> Explanation: An event for an art tour around Sydney
> URL: https://couchers.org/event/31312/sydney-art-tour
> Crumbs: Australia > NSW > Sydney > Sydney Art Tour

[...]

PR #623, opened 2021-02-12
Reset password functionality

darrenvong:
One of probably a few parts of #368. Breaking this up at this point before the changes get too large.

### DIFF
--- a/app/backend/src/couchers/tasks.py
+++ b/app/backend/src/couchers/tasks.py
@@ -30,6 +30,7 @@ def send_login_email(user, token, expiry_text):
 def send_password_reset_email(user, token, expiry_text):
     logger.info(f"Sending password reset email to {user=}:")
     password_reset_link = urls.password_reset_link(password_reset_token=token.token)
+    logger.info(f"Link is: {password_reset_link}")
     return email.send_email_template(
         user.email, "password_reset", template_args={"user": user, "password_reset_link": password_reset_link}
     )
--- a/app/frontend/src/AppRoutes.tsx
+++ b/app/frontend/src/AppRoutes.tsx
@@ -6,8 +6,8 @@ import { useAuthContext } from "./features/auth/AuthProvider";
 import Jail from "./features/auth/jail/Jail";
 import Login from "./features/auth/login/Login";
 import Logout from "./features/auth/Logout";
-import CompleteResetPasswordPage from "./features/auth/resetPassword/CompleteResetPasswordPage";
-import ResetPasswordPage from "./features/auth/resetPassword/ResetPasswordPage";
+import CompleteResetPasswordPage from "./features/auth/password/CompleteResetPasswordPage";
+import ResetPasswordPage from "./features/auth/password/ResetPasswordPage";
 import Signup from "./features/auth/signup/Signup";
 import CommunityPage from "./features/communities/CommunityPage";
 import DiscussionPage from "./features/communities/DiscussionPage";
@@ -31,6 +31,7 @@ import { PageType } from "./pb/pages_pb";
 export const loginRoute = "/login";
 export const loginPasswordRoute = `${loginRoute}/password`;
 export const resetPasswordRoute = "/passwordreset";
+export const changePasswordRoute = "/change-password";
 
 export const signupRoute = "/signup";
 export const profileRoute = "/profile";
--- a/app/frontend/src/AppRoutes.tsx
+++ b/app/frontend/src/AppRoutes.tsx
@@ -6,6 +6,7 @@ import { useAuthContext } from "./features/auth/AuthProvider";
 import Jail from "./features/auth/jail/Jail";
 import Login from "./features/auth/login/Login";
 import Logout from "./features/auth/Logout";
+import CompleteResetPasswordPage from "./features/auth/resetPassword/CompleteResetPasswordPage";
 import ResetPasswordPage from "./features/auth/resetPassword/ResetPasswordPage";
 import Signup from "./features/auth/signup/Signup";
 import CommunityPage from "./features/communities/CommunityPage";
@@ -66,8 +67,11 @@ export default function AppRoutes() {
       <Route path={`${signupRoute}/:urlToken?`}>
         <Signup />
       </Route>
-      <Route path={resetPasswordRoute}>
+      <Route exact path={resetPasswordRoute}>
         <ResetPasswordPage />
+      </Route>
+      <Route exact path={`${resetPasswordRoute}/:resetToken`}>
+        <CompleteResetPasswordPage />
       </Route>
       <Route path={tosRoute}>
         <TOS />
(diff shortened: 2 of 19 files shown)


app/backend/src/couchers/tasks.py
```diff
@@ -30,6 +30,7 @@ def send_login_email(user, token, expiry_text):
 def send_password_reset_email(user, token, expiry_text):
     logger.info(f"Sending password reset email to {user=}:")
     password_reset_link = urls.password_reset_link(password_reset_token=token.token)
+    logger.info(f"Link is: {password_reset_link}")
```
darrenvong, line 33: Small modification on the backend so I can actually see the reset password link in the console in dev mode.

app/frontend/src/features/profile/ProfilePage.tsx
```diff
@@ -34,6 +41,13 @@ export default function ProfilePage() {
         >
           Edit my place
         </ListItem>
+        <ListItem
```
darrenvong, line 44: Not sure where the best place to put this, so this makes the most sense to me for now. I suspect we'd want to move this to the profile page eventually under some options button maybe?

app/frontend/src/setupTests.ts
```diff
@@ -13,6 +13,7 @@ global.localStorage = createLocalStorageMock();
 
 afterEach(() => {
   global.localStorage.clear();
+  jest.restoreAllMocks();
```
darrenvong, line 16: Pretty much every test suite has an/some error case(s) that lead(s) to `console.error` needing stubbed out and being reset so that other tests still can log errors and allow us to see problems, so I thought worth moving it here so it's done automatically for all test suites by default.

app/frontend/src/AppRoutes.tsx
```diff
@@ -28,6 +33,8 @@ import { PageType } from "./pb/pages_pb";
 
 export const loginRoute = "/login";
 export const loginPasswordRoute = `${loginRoute}/password`;
+export const resetPasswordRoute = "/passwordreset";
+export const changePasswordRoute = "/change-password";
```
lucaslcode, line 37: We should decide if separate words have dashes or not. You could just leave it for now for #501 to decide.

app/frontend/src/features/auth/login/PasswordForm.tsx
```diff
@@ -1,13 +1,32 @@
...
 
-import { loginRoute } from "../../../AppRoutes";
+import { loginRoute, resetPasswordRoute } from "../../../AppRoutes";
 import Button from "../../../components/Button";
 import TextField from "../../../components/TextField";
 import { useAuthContext } from "../AuthProvider";
 
+const useStyles = makeStyles((theme) => ({
+  button: {
+    "& + &": {
```
lucaslcode, line 13: I thought I knew all the css selectors!

app/frontend/src/features/auth/login/PasswordForm.tsx
```diff
@@ -1,13 +1,32 @@
...
+      marginInlineStart: theme.spacing(1),
+    },
+  },
+  forgotPasswordLink: {
+    ...theme.typography.body1,
+    color: "inherit",
+    display: "block",
+    marginBlock: theme.spacing(2, 2),
```
lucaslcode: I have been using these, but I just looked at Can I Use and `marginBlock/marginInline` are only 64%. `marginBlockEnd/Start` and similar is 91%, so probably fine. However! I am using Edge, and it says that `marginBlock` isn't supported without a flag, but it seems to work for me and that flag isn't enabled. What do you think?

If compatibility was out of the question I would definitely say we should always use `*inline` and not `*left/*right` due to RTL languages. I'm less opinionated about `*block`, though I do like you can set vertical or horizontal without specifying both.

app/frontend/src/features/auth/password/CompleteResetPasswordPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  );
+
+  useEffect(() => {
+    if (resetToken) {
+      completePasswordReset(resetToken);
+    }
+  }, [completePasswordReset, resetToken]);
+
+  return isLoading ? (
+    <Typography variant="body1">Password reset in progress...</Typography>
```
lucaslcode, line 31: `<BodyText>`

app/frontend/src/features/auth/password/CompleteResetPasswordPage.tsx
```diff
@@ -0,0 +1,44 @@
...
+  ) : isSuccess ? (
+    <>
+      <Alert severity="success">
+        Your password has been reset successfully!
+      </Alert>
+      <Typography variant="body1" component={Link} to={loginRoute}>
```
lucaslcode, line 37: As above

app/frontend/src/features/auth/password/ChangePasswordPage.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <>
+      <PageTitle>Change password</PageTitle>
+      {error && <Alert severity="error">{error.message}</Alert>}
+      {isSuccess && (
+        <Alert severity="success">
+          Your password change has been processed. Check your email for
+          confirmation.
+        </Alert>
+      )}
+      <Typography className={classes.infoText} variant="body1">
```
lucaslcode, line 74: `<BodyText>` or maybe we should get rid of that component....

lucaslcode (dismissed):

app/frontend/src/features/auth/password/ChangePasswordPage.tsx
darrenvong: I noticed we start using `Typography` more for other heading levels. I guess since all the styles seem to be standardised/controlled through the theme's typography variants, `BodyText` feels like a redundant wrapper to me at this point?  🤔

app/frontend/src/features/auth/login/PasswordForm.tsx
darrenvong: I noticed you used them and look them up on MDN and they sound great for what you said! But I didn't see the "can I use" stats 🤦   looks like `marginInline` support is non-existent for Safari and I suspect we'll have a fair number of users using that if they are on a Mac. Uhhhh it's literally the last annoying browser left once you can drop IE support.

app/frontend/src/AppRoutes.tsx
darrenvong: Yeah the dashes seem consistent with Aapeli's proposal there tbf on places that have multiple words, so maybe we should change `passwordreset` to `password-reset` too. I only left that just to match what's currently in the backend.

app/frontend/src/features/auth/login/PasswordForm.tsx
darrenvong: Never mind, it's only the shorthand property they don't support, updating to explicitly use `marginBlockStart` and `marginBlockEnd` then...

lucaslcode (approved):

Issue #495, opened 2021-01-08
Smart date entry form field


CouchersBot:
Subject: Profile birthday
Description:
Wasn't clear how I had to entre my birthday
DD/MM/Year
MM/DD/Year
Year/MM/DD
....

Steps:
Entered my birthday the European way: DD/MM/Year

Results:
IN the end I found the correct way but might be helpful if it states how the website want it entered.

Backend version: develop-4a30e13d
Frontend version: develop-4a30e13d
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0.2 Safari/605.1.15
Page: https://app.couchers.org/profile
User ID: 170

lucaslcode:
Related to #600

PR #624, opened 2021-02-12
Fixing minor visual bugs

### LINKED ISSUE #606: Visual bugs
Poked around on iOS 14.4, mobile Safari, iPhone 11 Pro.

* The main content overflows under the bottom menu bar on every page (the "Save" button here is partly hidden):

signal-2021-02-08-164731_003

* Initiating host request has these empty blobs for start/end dates.

signal-2021-02-08-164731_001

* Missing space under this message:

signal-2021-02-08-164731_002

Tay0108:
Closes #606.

There is one problem - date picking is not a visual bug - Chrome has its own component for doing date picking, Safari does not. Should I replace those TextFields with some MaterialUI DatePicker?

### DIFF
--- a/app/frontend/src/App.tsx
+++ b/app/frontend/src/App.tsx
@@ -17,7 +17,7 @@ import { theme } from "./theme";
 
 const useStyles = makeStyles((theme) => ({
   padding: {
-    paddingBottom: theme.spacing(7),
+    paddingBottom: theme.spacing(10),
     paddingLeft: theme.spacing(2),
     paddingRight: theme.spacing(2),
     [`${theme.breakpoints.up("xs")} and (orientation: landscape)`]: {
--- a/app/frontend/src/features/messages/surfing/NewHostRequest.tsx
+++ b/app/frontend/src/features/messages/surfing/NewHostRequest.tsx
@@ -18,6 +18,11 @@ import { validateFutureDate } from "../../../utils/validation";
 import { useUser } from "../../userQueries/useUsers";
 
 const useStyles = makeStyles((theme) => ({
+  buttonContainer: {
+    display: "flex",
+    paddingTop: theme.spacing(1),
+    justifyContent: "center",
+  },
   form: {
     "& > *": {
       marginTop: theme.spacing(2),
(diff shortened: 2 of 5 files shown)


lucaslcode:
Thanks, will check this out tonight!

#495 is related to the dates and would fix that I'm guessing, so if you wanted to work on that it could be another PR.

app/frontend/src/App.tsx
```diff
@@ -17,7 +17,7 @@ import { theme } from "./theme";
 
 const useStyles = makeStyles((theme) => ({
   padding: {
-    paddingBottom: theme.spacing(7),
+    paddingBottom: theme.spacing(10),
```
lucaslcode, line 20: How did you choose 10? This will conflict with my PR awaiting review and I did 9 (can't remember why) but I could change it to 10.

lucaslcode (approved):

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -112,14 +117,16 @@ export default function NewHostRequest() {
...
           />
-          <Button
-            type="submit"
-            color="primary"
-            onClick={onSubmit}
-            loading={mutation.isLoading}
-          >
-            Send
-          </Button>
+          <div className={classes.buttonContainer}>
```
lucaslcode: use `Box`

lucaslcode (changes requested):

app/frontend/src/App.tsx
Tay0108: I chose 10 because in figma there is 16px gap between navigation and button. Navigation has 64px (doesn't matter what device I check on) so if I put 10 `padding-bottom` evaluates to 80px (64+16). But what I did not see is that in Figma navigation is slightly bigger (73px). So it's up to you ;)

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
Tay0108: Should I replace every `buttonContainer` `div` with `Box`? I did not introduce it, seen this div wrapper in some component so I thought I will copy for consistency.

lucaslcode: Just change this one for now and we can decide what to do overall and fix the others later

lucaslcode (approved):

PR #626, opened 2021-02-13
Add doc explaining the "Life of an RPC call through the backend"

aapeliv:
I hope this helps explain how the backend works, as it's quite opaque to those new to gRPC


lucaslcode (approved):
nice one

PR #627, opened 2021-02-13
Join groups/communities

### LINKED ISSUE #621: Joining groups and communities
Copy paste from wiki:

Aapeli, 25/1, after discussion with Itsi:

* Users automatically become members of communities whose polygon they are inside (either batch job + when changing one's location). A user may join any other community by clicking "become mebmer", and they'll get notifications, etc, as if they lived there. A user does not automatically stop being a member of a community. It's not possible to stop being a member of a community you're contained in, but it is possible to leave a community that you're not inside. (It will later be possible to turn off notifications for communities).
* Membership in groups is completely voluntary, by clicking "join group"/"leave group" buttons. There are no private/invite only groups, etc.

aapeliv:
Closes #621


aapeliv:
Screwed up the base branch here... Fixed.

I was originally going to do this on top of the bg job PR as I was going to add a bg job, but that can happen later.

aapeliv:
@lucaslcode: I'm sorry, the frontend here is a new low...

lucaslcode (commented):
Looks okay to me (I didn't look extremely carefully at backend, but seems okay if the tests work), but do you need to add logic to the user changing their location which adds them to relevant communities?

Some things to think about:
- If I change location, am I still in the previous community?
- If a new community is added where I am, am I added automatically?

aapeliv:
For now user membership in communities will be done in a background job (the task exists, triggering it is awaiting the bg PR). An idempotent job will just make it easier to guarantee we don't miss any edge cases (user move/created, community added/removed, community bounds changed, etc), since this inclusion is impossible to guarantee in the database (I believe).

> do you need to add logic to the user changing their location which adds them to relevant communities?

Probably a good idea on top of the job.

> If I change location, am I still in the previous community?

Yes, but you can leave the group manually.

> If a new community is added where I am, am I added automatically?

Yes. Will be added by the background job.

kalvdans (approved):
Backend part looks fine, with tests and everything!

PR #629, opened 2021-02-13
Organize urls

### LINKED ISSUE #501: URL plan

lucaslcode:
Closes #501 

This is the best approach I could think of. I also considered making all routes functions but seemed a bit excessive, and inelegant to get the required `:routeParam`s in there.

### DIFF
--- a/app/frontend/src/features/messages/Messages.tsx
+++ b/app/frontend/src/features/messages/Messages.tsx
@@ -1,7 +1,16 @@
 import * as React from "react";
 import { Route, Switch, useHistory, useParams } from "react-router-dom";
 
-import { messagesRoute } from "../../AppRoutes";
+import {
+  archivedMessagesRoute,
+  groupChatsRoute,
+  hostingRequestsRoute,
+  hostRequestRoute,
+  meetRoute,
+  messagesRoute,
+  newHostRequestRoute,
+  surfingRequestsRoute,
+} from "../../AppRoutes";
 import NotificationBadge from "../../components/NotificationBadge";
 import PageTitle from "../../components/PageTitle";
 import TabBar from "../../components/TabBar";
@@ -44,7 +53,7 @@ export function HostRequestsSentNotification() {
 
 const labels = {
   all: "All",
-  groupchats: <MessagesNotification />,
+  chats: <MessagesNotification />,
   hosting: <HostRequestsReceivedNotification />,
   surfing: <HostRequestsSentNotification />,
   meet: "Meet",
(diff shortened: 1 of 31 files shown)


app/frontend/src/features/messages/Messages.tsx
```diff
@@ -44,7 +53,7 @@ export function HostRequestsSentNotification() {
 
 const labels = {
   all: "All",
-  groupchats: <MessagesNotification />,
+  chats: <MessagesNotification />,
```
lucaslcode, line 56: changed to chats for now since we aren't differentiating between group chats and dms in the ui

app/frontend/src/features/messages/Messages.tsx
```diff
@@ -44,7 +53,7 @@ export function HostRequestsSentNotification() {
...
   hosting: <HostRequestsReceivedNotification />,
```
lucaslcode, line 57: I also considered something like `[hostingRequestsRoute.split("/")[hostingRequestsRoute.length - 1]]: 
 <HostRequestsReceivedNotification />,` but seems way too ugly. But it would mean it doesn't need manual syncing.

aapeliv:
How about going one step further and moving all the route functions to a separate `.ts` file?

app/frontend/src/features/messages/Messages.tsx
lucaslcode: `[hostingRequestsRoute.split("/").pop()]` ?

aapeliv (dismissed):
The backend is obviously good, and to me the frontend looks good as well, but I will defer a full frontend review to a more frontend-y person :)

kalvdans:
We should document the decisions made in https://github.com/Couchers-org/couchers/issues/501 to a markdown document. I can do that in this PR if no one beats me to it.

app/frontend/src/features/messages/Messages.tsx
darrenvong: Hmm not sure about that, seems a little over abstracted for me and more obscure to read. If the label does update, I suspect we'd want to rename the component name on the right hand side too, so it wouldn't be too much work to update both in one go? I doubt this will change enough to be a pain?

darrenvong (dismissed):
Frontend side looks fine to me, I mean I'd say backend is too unless @aapeliv want to give a thumb up too 😄

lucaslcode:
@darrenvong just rebased to include your new pages and also changed the backend to put - in the urls as discussed

app/frontend/src/features/messages/Messages.tsx
lucaslcode: agreed

darrenvong (dismissed):

darrenvong (approved):

PR #634, opened 2021-02-15
Enforce main page uniqueness with partial index instead of extra field

aapeliv:
What do you think about this @kalvdans?

It simplifies code a bit.


aapeliv:
The same could be done for `official_cluster_for_node_id`, what do you think?

kalvdans:
That's some heavy postgresql magic there! An index made from two columns.

Auto-lgtm from me on changes that reduces the amount of redundant state.

kalvdans (approved):

PR #635, opened 2021-02-15
Remove thread_id from cluster

aapeliv:
Clusters erroneously had two threads: one for the cluster object itself, one for the main page. This removes the one from the cluster object itself, and uses the one for the main page instead.


kalvdans (dismissed):
Great cleanup, thanks!

PR #636, opened 2021-02-15
Organise storybook

lucaslcode:
- [x] Sort existing stories
- [x] Add all current components
- [x] Add page-specific components

Taking suggestions for a more descriptive name than "Small" and "Large" components


lucaslcode:
I decided to add repeatable page-specific components only, not "structural" components which only exists to break pages up into smaller chunks.

darrenvong:
> Taking suggestions for a more descriptive name than "Small" and "Large" components

Maybe "Simple" and "Composite" components? Word idea directly stolen from [Storybook intro guide](https://www.learnstorybook.com/intro-to-storybook/react/en/simple-component/) itself 😄

app/frontend/src/components/Icons/Icons.stories.mdx
```diff
@@ -0,0 +1,31 @@
+import { Meta, Story, Canvas } from "@storybook/addon-docs/blocks";
+import * as Icons from "./index";
+
+<Meta title="Components/Small/Icons" />
+
+# Couchers Icons
+
+<Canvas>
```
darrenvong, line 8: The `ExpandMoreIcon` is missing?

app/frontend/src/features/messages/messagelist/MessageView.stories.tsx
```diff
@@ -1,19 +1,19 @@
...
-import AuthProvider from "../auth/AuthProvider";
-import Messages from "./index";
-import ControlMessageView from "./messagelist/ControlMessageView";
-import MessageView, { MessageProps } from "./messagelist/MessageView";
+import messages from "../../../test/fixtures/messages.json";
+import { addDefaultUser } from "../../../test/utils";
+import AuthProvider from "../../auth/AuthProvider";
+import Messages from "../index";
+import ControlMessageView from "./ControlMessageView";
+import MessageView, { MessageProps } from "./MessageView";
```
darrenvong, line 10: The control message on this is broken :(

app/frontend/src/features/messages/groupchats/GroupChatsTab.stories.tsx
```diff
@@ -12,7 +12,7 @@ import GroupChatView from "./GroupChatView";
 const queryClient = new QueryClient();
 
 export default {
-  title: "GroupChatsTab",
+  title: "Messages/GroupChatsTab",
```
darrenvong, line 17: The chat page in this is also broken... I wonder if it's the mocks playing up...

app/frontend/src/components/ScoreBar.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+import ScoreBar from "./ScoreBar";
+
+export default {
+  title: "Components/Large/ScoreBar",
+  component: ScoreBar,
+} as Meta;
+
+const Template: Story<any> = (args) => (
+  <>
+    <ScoreBar {...args}>Community standing</ScoreBar>
```
darrenvong, line 12: We might want to change text colours depending on if we are keeping this purple... the black text on purple makes the text impossible to read

app/frontend/src/features/messages/surfing/HostRequestListItem.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+import HostRequestListItem from "./HostRequestListItem";
+
+export default {
+  title: "Messages/HostRequestListItem",
+  component: HostRequestListItem,
+} as Meta;
+
+const Template: Story<any> = (args) => (
+  <>
+    <HostRequestListItem {...args} />
```
darrenvong, line 13: The loading state of this looks a bit odd, it's showing partially loaded data before the avatar/user is being loaded etc. Wonder if that's just the mocks being weird again...

darrenvong:
Also the font on some of these components don't display in the default Ubuntu font in the Storybook entries. I suspect it's because they are not using the `Typography` component so the styles don't get picked up.

Wonder if it's best to stay on the safer side and make everything use `Typography`, or at least the styling from it (not sure how without using the component though)? Or it probably doesn't matter since the font family is set at the `body`?

app/frontend/src/features/messages/groupchats/GroupChatsTab.stories.tsx
lucaslcode: It was because it had changed from being passed the chat, to querying it using the id in the url

app/frontend/src/features/messages/surfing/HostRequestListItem.stories.tsx
lucaslcode: It's because the service mock in storybook adds a 1 second loading time to everything. It could be noticeable in real usage, so I cleaned the skeletons up a bit.

lucaslcode:
Yes everything should use `Typography`, I don't think fonts are 100% guaranteed to be on `body` (or maybe not in complex cases or something)

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -85,6 +85,7 @@ export default function GroupChatView() {
   };
 
   const groupChatId = +(useParams<{ groupChatId?: string }>().groupChatId || 0);
+  console.log(groupChatId);
```
darrenvong: Forgot to remove from debugging?

lucaslcode: 😠

darrenvong (approved):

PR #637, opened 2021-02-15
Backend/bugfix/birthdate in past

### LINKED ISSUE #274: Should verify birthdate is in the past
Causes backend errors if not

OmerBaddour:
Modified **auth.py** to now also check that the birthdate entered upon sign up is in the past.

Retained the same form of try/except that was previously only used to check the form of the birthdate. This makes sense to me because the `is_valid_date()` function simply yields **True/False**. Invoking this function, checking for **True**, and then converting the string to a **datetime** instance is less straightforward than simply trying to convert the string to a **datetime** instance (especially since the **datetime** instance is needed by the database in the valid case).

Note: my code does not account for timezones. I prioritized simplicity and readability over this level of accuracy.

Note: also fixed a typo in a comment for unrelated code, since the comment is used to generate documentation.

### DIFF
--- a/app/backend/src/couchers/servicers/auth.py
+++ b/app/backend/src/couchers/servicers/auth.py
@@ -201,11 +201,11 @@ class Auth(auth_pb2_grpc.AuthServicer):
             if not signup_token:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
-			# check birthdate validity (YYYY-MM-DD format and in the past)
+            # check birthdate validity (YYYY-MM-DD format and in the past)
             try:
                 birthdate = datetime.fromisoformat(request.birthdate)
-				if birthdate >= datetime.now():
-            	    context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
+                if birthdate >= datetime.now():
+                    context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
             except ValueError:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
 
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/servicers/auth.py
```diff
@@ -201,9 +201,11 @@ def CompleteSignup(self, request, context):
             if not signup_token:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
-            # should be in YYYY-MM-DD format
+            # check birthdate validity (YYYY-MM-DD format and in the past)
             try:
                 birthdate = datetime.fromisoformat(request.birthdate)
+                if birthdate >= datetime.now():
+                    context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
```
aapeliv: This should be moved outside the try-catch block. The block is there to catch any errors in date formatting. What you've added shouldn't raise a ValueError.

Also, please use the `now()` function from `couchers.utils`. It's timezone aware and should be used everywhere. Timezone bugs are a pain.

aapeliv (changes requested):

app/backend/src/couchers/servicers/auth.py
```diff
@@ -201,11 +201,13 @@ def CompleteSignup(self, request, context):
             if not signup_token:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
-            # should be in YYYY-MM-DD format
+            # check birthdate validity (YYYY-MM-DD format and in the past)
             try:
                 birthdate = datetime.fromisoformat(request.birthdate)
             except ValueError:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
+            if utc.localize(birthdate) >= now():
```
aapeliv: Need to import pytz, then do `pytz.UTC.localize(...)`

aapeliv:
LGTM, contingent on tests passing

aapeliv (dismissed):
LGTM, contingent on tests passing

OmerBaddour:
Added birthdate test. Passes all tests.

aapeliv (approved):

aapeliv:
Closes #274

PR #645, opened 2021-02-15
Enforce single official cluster with partial index

aapeliv:
Instead of extra column and check constraints


aapeliv:
@kalvdans: I separated this into a new PR because it's a bit hairy

kalvdans (approved):
So indices are not just an optimization, they can also be used to state constraints?

Looks good with the test!

aapeliv:
> So indices are not just an optimization, they can also be used to state constraints?

Yes, if you declare the index unique, it will enforce uniqueness. It's the way uniqueness constraints are implemented in general in postgres. This is pretty powerful with partial indexes (as we're using here)!

Issue #391, opened 2020-12-07
Speed up CI


aapeliv:
They used to take like 1.5 min, now it's at ~6 min.

Almenon:
You can try running the tests in parallel:

https://www.tutorialspoint.com/pytest/pytest_run_tests_in_parallel.htm#:~:text=By%20default%2C%20pytest%20runs%20tests,to%20run%20tests%20in%20parallel.

https://docs.pytest.org/en/3.0.1/xdist.html

aapeliv:
They hit the same DB, though, which is problematic. Would require partitioning into separate schemas or similar. I'm not sure if PostGIS can be installed outside the default schema.

caleb15:
I think postgis is installed per database, and a db can have multiple schemas inside it.

https://postgis.net/install/

> PostGIS is an optional extension that must be enabled in each database you want to use it in

Almenon:
I think postgis is installed per database, and a db can have multiple
schemas inside it.

https://postgis.net/install/

PostGIS is an optional extension that must be enabled in each database you
want to use it in

WritePreview

aapeliv:
Related: https://github.com/Couchers-org/couchers/pull/571 (but does not close this issue)

aapeliv:
There's a few more things with CI: ~~we could do the gitlab DAG thing, so that backend and frontend could be built/tested at their own pace. Currently frontend is slow to build, fast to test, and backend is fast to build, slow to test.~~ (done in #646)

Another thing is that we could move the frontend build out of docker, and this should speed it up a whole lot.

I think the biggest slowdown is just that we're regenerating all models from scratch for each test, and as the number of tables and complexity of models grows, the tests will slow down in proportion.

ShmuelTreiger:
In my experience, running pytest in parallel led to hung tests and increased error rate (though that was 2017, so I suppose it could have been fixed since then).

kalvdans:
Related: https://github.com/Couchers-org/couchers/pull/885 opens up for running a single migration test instead of running all tests twice.

PR #646, opened 2021-02-15
Make CI run in parallel and out-of-order

aapeliv:
This makes it a bunch faster


kalvdans (dismissed):
Looks good if it works :)

aapeliv:
Relates to #391

aapeliv:
> Looks good if it works :)

It doesn't 😢. But I'll fix it :P

Issue #547, opened 2021-01-30
Search redesign


aapeliv:
~~Fuzzy on name, simple~~

Update: backend is in place, we just need to implement the design.

To start with, we don't need an omni-search, just select what you want to search for first.

Then, we should display everything (eg. all users) on the map like the soft beta map does (from a big geojson), and when a search is performed, pass the result ids to the layer as a filter.

Also, make the map use the light theme

lucaslcode:
Also search for groups, discussions, places, etc?

aapeliv:
Backend part now merged in #650 

> Also search for groups, discussions, places, etc?

Yes, name is ambiguous, I meant "communities features" search :).

Tay0108:
I don't know how to filter features (markers) on clustered layer - do you have any ideas? Or maybe we should turn if off for search?

aapeliv:
It seems difficult: https://stackoverflow.com/questions/36179464/mapbox-gl-js-filtering-marker-clusters.

Let's leave it out for now, I'm not sure what the prettiest way is, I guess just turn clustering off?

Tay0108:
Yeah, I think this would be the best for now, because you would see a lot of clusters and no users after you zoom in.

Issue #546, opened 2021-01-30
Advanced user search/filtering


aapeliv:
i.e. host search

lucaslcode:
Closing in favour of #547

PR #507, opened 2021-01-17
Backend/feat/add language fluency

Almenon:
Resolves the backend portion of https://github.com/Couchers-org/couchers/issues/409

PR #650, opened 2021-02-16
Search API

aapeliv:
Draft 1


lucaslcode:
Looks good to me.

app/pb/search.proto
```diff
@@ -0,0 +1,112 @@
...
+
+  oneof result {
+    org.couchers.api.core.User user = 2;
+    org.couchers.api.communities.Community community = 3;
+    org.couchers.api.groups.Group group = 4;
+    org.couchers.api.pages.Page place = 5;
+    org.couchers.api.pages.Page guide = 6;
+    org.couchers.api.discussions.Discussion discussion = 7;
+    // org.couchers.api.events.Event event = 8;
+  }
```
darrenvong: The generated type for this could be a bit clunky to work with, as we'd have to check for each result item what kind of result it is to display accordingly on the frontend. I wonder if there's a good way to unify this into some common interface so we can loop through the results easily and display it through some "high level" views? And if the user want more details, maybe it can expand to one of these more detailed entities?

lucaslcode: I think it's fine, just have `<SearchResultCard>` that is a big switch statement returning `<UserSearchResultCard>`, `<GroupSearchResultCard>`, etc. Or would that have other problems?

darrenvong: That's fine, but I was thinking if there's a way to keep the display logic simpler on the frontend then that's always better. May not be possible if each card is going to be quite different though looking at the designs so far... 🤔

lucaslcode: I see. Yes I think they're intrinsically too different to try and unify.

aapeliv: Yeah, I could return them as repeated users, communities, etc. But you'd have to order them in the frontend then

kalvdans: Would it be good, for each search result, to return an URL and a text snippet containing the searched term? To ease the presentation of the search results.

kalvdans:
Is there any particular reason to use *Value and not primitive types in the search query? For example maxAge could easily be passed as 200 if we don't want to filter on it.

app/pb/search.proto
aapeliv: > return an URL

URL to where? If you mean to the resultant object (e.g. user), I think that should go inside the `User` object itself instead. I've some thoughts on that.

> a text snippet containing the searched term

You mean like a snippet of the match? I think that'd be a good idea. I was just thinking that it might be annoying to implement :P

kalvdans: An URL to the resource according to our URL plan :) Yes a snippet with the search term highlighted like a google search result (client can highlight)

aapeliv: > An URL to the resource according to our URL plan :)

Yeah... That should probably be included in the objects themselves. And I've been starting to think we should refactor the whole API to be more resource-oriented.

The only thing is that including a URL is very web-centric. What does it mean for mobile apps or third party apps? Of course we could have our own URI scheme of some form, `couchers://community/1/world` (or just return the path only)?

> Yes a snippet with the search term highlighted like a google search result (client can highlight)

Yeah, I'll add that. Would it be better if the highlighting was through markdown? Or just choose some symbol, `*` or `|`, and highlight sections with that? We can probably use the postgres full-text search to get pretty decent search for now: it doesn't match verbatim, it matches based on word roots or whatever.

kalvdans: > That should probably be included in the objects themselves.

But then we lose the whole point that it is hard to deal with oneof in the frontends...

> The only thing is that including a URL is very web-centric.

It's not. Just because an RFC1738 uniform resource locator happens to start with https, doesn't forbid any app to parse it and pick out the ID numbers.

aapeliv:
> Is there any particular reason to use *Value and not primitive types in the search query? For example maxAge could easily be passed as 200 if we don't want to filter on it.

I mean, you could do that for `maxAge`. But it's not very clean, and the frontend has to specify all primitive filters then whenever doing a request, and then what do you set maxAge to? What about people who go wonky and set a birthday in the really far past? Then backend and frontend have to coordinate on what the max valid age is.

The other thing that we should probably move to at some point is the whole `FieldMask` thing (https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.FieldMask) that protobuf has. It basically is an extra field you add that contains the mask of what you want to update/set in a query. Much nicer, but I'm not sure what it looks like from the python and javascript sides.

app/pb/search.proto
```diff
@@ -0,0 +1,112 @@
...
+  string next_page_token = 2;
+}
+
+message UserSearchReq {
+  google.protobuf.StringValue query = 3;
+
+  google.protobuf.StringValue gender = 4;
+
+  // coarsened, times less than one day mean no filter
+  google.protobuf.Timestamp last_active = 5;
```
kalvdans: Omitting the value means no filter.

aapeliv: Thanks, will fix

kalvdans:
> But it's not very clean, and the frontend has to specify all primitive filters then whenever doing a reques

I agree, it would also be harder to add more filters in a backwards-compatible way.

> The other thing that we should probably move to at some point is the whole `FieldMask` thing.

Cool! I guess that you get nothing for free in the Python generated stubs and has to parse it like a normal repeated string.

app/pb/search.proto
lucaslcode: > return an URL

We won't use this in the frontend. We will have to handle each type of result differently no matter what, and we have route functions now like `routeToUser(userId)`.

> and a text snippet containing the searched term

I wouldn't spend too much time on that, while it would be good for guides/discussions/etc. it isn't in the current design (and we have lots of other stuff to do 😱 )

aapeliv: Oh my... I love postgres:

https://www.postgresql.org/docs/current/textsearch-controls.html#TEXTSEARCH-HEADLINE

aapeliv:
I've updated according to comments, please add new comments if you have any. Otherwise I'll try to get to this over the weekend.

kalvdans (dismissed):
API looks good

app/pb/search.proto
```diff
@@ -0,0 +1,115 @@
...
+  bool include_communities = 5;
+  bool include_groups = 6;
+  bool include_places = 7;
+  bool include_guides = 8;
+  bool include_discussions = 9;
+  // bool include_events = 10;
+
+  // at most one of the following
+  Area search_in_area = 11;
+  int64 search_in_community_id = 12;
```
lucaslcode: Add oneof here so this is enforced?

lucaslcode: Actually maybe not since it's at least one of

darrenvong:
From the thread earlier - if we go with a resource oriented approach (at some point? sounds like might be too much work to refactor now?), and the URL conforms to what we agreed on already, is there any reason why we couldn't use this easily (or even directly) on the frontend? We're sort of doing that already with things like login URLs in our emails.

The main reason I see left would be the visual differences. Might it be worth pushing back on the designs if it's gonna make our lives easier at least for the initial version? I'm just wary of letting design dictating too much and making our lives harder unnecessarily just because they design some funky stuff. Although it's probably ok in this case... mostly thinking out on virtual paper with this.

app/pb/search.proto
aapeliv: Yeah, could add `oneof`, but I feel like it just makes it more complicated... Maybe I should though, to be clear? ...Yeah, I should probably add it.

darrenvong: Wouldn't `oneof` restrict our ability to search by both? Or is it a bit redundant to allow that since we'd only get one community matching this ID when we use this option?

app/pb/search.proto
```diff
@@ -0,0 +1,115 @@
...
+}
+
+message UserSearchReq {
+  google.protobuf.StringValue query = 3;
+
+  google.protobuf.StringValue gender = 4;
+
+  // coarsened
+  google.protobuf.Timestamp last_active = 5;
+  google.protobuf.StringValue language = 6;
```
darrenvong: Do we want to allow user search by multiple languages? e.g. find me all users that speak English and Spanish. I guess this would only support one language at the moment right?

app/pb/search.proto
```diff
@@ -0,0 +1,115 @@
...
+  google.protobuf.BoolValue has_references = 7;
+  bool friends_only = 8;
+
+  google.protobuf.UInt32Value age_min = 9;
+  google.protobuf.UInt32Value age_max = 10;
+
+  repeated org.couchers.api.core.HostingStatus hosting_statuse_filter = 11;
```
darrenvong:
```suggestion
  repeated org.couchers.api.core.HostingStatus hosting_status_filter = 11;
```
typo 😛

aapeliv: Ah, yeah, thanks!

app/pb/search.proto
aapeliv: I'd rather imagine people would like to find someone who speaks English OR Spanish?

For now, before #507, we can't do anything but a string match, so multiple keywords doesn't make sense.

aapeliv:
> From the thread earlier - if we go with a resource oriented approach (at some point? sounds like might be too much work to refactor now?), and the URL conforms to what we agreed on already, is there any reason why we couldn't use this easily (or even directly) on the frontend? We're sort of doing that already with things like login URLs in our emails.
> 
> The main reason I see left would be the visual differences. Might it be worth pushing back on the designs if it's gonna make our lives easier at least for the initial version? I'm just wary of letting design dictating too much and making our lives harder unnecessarily just because they design some funky stuff. Although it's probably ok in this case... mostly thinking out on virtual paper with this.

Are you proposing we'd just send back instead something simple like this:

```proto3
message Result {
  double relevance = 1;
  string result_title = 2;
  string result_url = 3;
  string snippet = 3;
}
```

If this is what you mean, and if it makes life easier for you to get up an initial version, I can add the title and URL (temporarily, removed later) into the result so you can ignore the actual objects?

app/pb/search.proto
darrenvong: Ahh yeah OR would make more sense. I was thinking more like a non-dev for a sec there, using "and" to mean "and/or" 😅 

Does that mean you plan to update this in a future version of this once the linked issue is merged?

aapeliv: > Does that mean you plan to update this in a future version of this once the linked issue is merged?

Should've been more clear: I don't know 🤷‍♂️...

Maybe this should just be removed for the time being?

darrenvong:
> Are you proposing we'd just send back instead something simple like this:
> 
> ```
> message Result {
>   double relevance = 1;
>   string result_title = 2;
>   string result_url = 3;
>   string snippet = 3;
> }
> ```

Yeah that matches well with the simple version I'm thinking of. It'd definitely make getting an initial version of the UI easier to start with (less variants to style, almost close to no logic apart from reading data from the response). I guess the follow-up question I have is would we be happy to release that in the early beta version and follow up with the more specific UIs later? If not, maybe this isn't saving us that much time in the long run.

darrenvong:
After thinking more about this... scrap what I said and just leave the `oneof` thing over the different object types as it is. I don't think it's worth taking the shortcut if we need to go back to change it later, so might as well do it right the first time. With this split into different component mentioned, it's probably not as bad as my gut instinct thoughts initially.

app/pb/search.proto
darrenvong: Hmm I guess even having a basic version (i.e. to work with one language) of this working can be nice initially?

aapeliv: I'm going to fix up the languages stuff, so I'll add this on top, with a filter to have "or" on multiple languages, and a minimum fluency.

app/pb/search.proto
aapeliv: Yeah, the idea is that these are both "geo restrictions".

I've clarified this in the pb

aapeliv:
This is a f**king mess...

aapeliv:
I've now implemented the main chunk of a full-text search for most content. However, it's a textual search, not an exact search, so it's not good for titles.

I think w need to add a separate RPC for fuzzy search of titles only using a trigram index, which will be good for names, etc.

Should that be integrated together? Then the interleaving with ranking is a bit hard?

aapeliv:
This is now ready for review. I've patched up the frontend so it should work and frontend can be extended to take advantage of new features.

The code is fairly rough, and there's pretty much no tests (except to make sure basically that the functions run and work *somehow*). I feel like we'll end up iterating this on a fair bit, and it's in general a bit hard (+ super tedious) to test this. I've also not yet done search for discussions (these are bit difficult because you have to condense the whole tree of discussions into one searchable object). I think some stuff will come up as frontend is implemented.

Caveats:
1. Currently we have no indexes in place to make this fast. These should be added as we go and this matures. (The search won't be too slow with a couple thousand rows anyway.)
2. Poorly tested.
3. Code is a bit rough/hard to read due to SQLAlchemy mapping <> postgres special functions
4. I've removed some bits from the earlier protos that ended up being annoying/not worth it (in particular whether the match was in title or content)
5. I've commented out stuff from protos that I think should be implemented but are not yet done

app/pb/search.proto
```diff
@@ -0,0 +1,135 @@
...
+
+  // restrictions on area
+  // oneof search_in {
+  //   // search in given area
+  //   Area search_in_area = 11;
+  //   // search inside the area defined by the community
+  //   int64 search_in_community_id = 12;
+  // }
+
+  // whether to only perform a trigram based search on titles and not do full text search
```
kalvdans, line 60: Does the trigram search influence the results? Leave it out of the API docs if it is an implementation detail.

aapeliv:
Closes backend part of #546 and #547.

app/pb/search.proto
aapeliv: Yes it does, it's a different indexing/search strategy:

* Trigram looks at triplets of characters, and matches using a reverse index to find similar words. It's basically fuzzy search and good with typos etc.
* Full text search divides the text into lexemes and matches against dictionaries, etc, to find exact word matches against "standardized words" (so different forms of the same word match)

I think trigram will be much better for finding names of places, of people, and just matching against short strings. Full text is better for finding specific topics in large text

app/pb/search.proto
```diff
@@ -0,0 +1,135 @@
...
+}
+
+message Area {
+  double lat = 1;
+  double lng = 2;
+  double radius = 3; // m
+}
+
+message SearchReq {
+  string query = 1;
```
kalvdans, line 42: Please document the format of the search string:

* Are AND and OR parsed and interpreted?
* Will minus in front of a word remove those results from the hits?
* Will quotation marks force a literal search for that string?

app/pb/search.proto
```diff
@@ -0,0 +1,135 @@
...
+  google.protobuf.BoolValue parking = 26;
+  google.protobuf.BoolValue camping_ok = 27;
+
+  uint32 page_size = 1;
+  string page_token = 2;
+}
+
+message UserSearchRes {
+  // this will only contain users
+  repeated Result results = 1;
```
kalvdans, line 143: // Results are sorted with highest rank first

app/pb/search.proto
```diff
@@ -0,0 +1,135 @@
...
+    org.couchers.api.groups.Group group = 5;
+    org.couchers.api.pages.Page place = 6;
+    org.couchers.api.pages.Page guide = 7;
+    // org.couchers.api.events.Event event = 8;
+    // org.couchers.api.discussions.Discussion discussion = 9;
+  }
+}
+
+message SearchRes {
+  repeated Result results = 1;
```
kalvdans, line 87: // Results are sorted with highest rank first

app/backend/src/couchers/servicers/search.py
```diff
@@ -0,0 +1,423 @@
...
+            func.setweight(func.to_tsvector(regconfig, _join_with_space([func.coalesce(bit, "") for bit in D])), "D")
+        )
+    return tsv
+
+
+def _build_doc(A, B=[], C=[], D=[]):
+    """
+    Builds the raw document (without to_tsvector and weighting), used for extracting snippet
+    """
+    doc = _join_with_space([func.coalesce(bit, "") for bit in A])
```
kalvdans, line 64: Reduce to a oneliner:

`return " ".join(func.coalesce(bit, "") for bit in A + B + C + D)`

app/backend/src/couchers/servicers/search.py
```diff
@@ -0,0 +1,423 @@
...
+from couchers.servicers.communities import community_to_pb
+from couchers.servicers.groups import group_to_pb
+from couchers.servicers.pages import page_to_pb
+from couchers.utils import create_coordinate, to_aware_datetime
+from pb import search_pb2, search_pb2_grpc
+
+# searches are a bit expensive, we'd rather send back a bunch of results at once than lots of small pages
+MAX_PAGINATION_LENGTH = 50
+
+regconfig = "english"
```
kalvdans: capitalize constants

kalvdans (changes requested):
That is some heavy postgres magic going on there!

Tests are a bit thin, but it is a very conventient skeleton to add to later.

app/backend/src/couchers/servicers/search.py
aapeliv: There's some underlying SQLAlchemy magic going on on top of this, therefore the ugliness with `_join_with_space`, etc.

I had issues with `str.join`: it's expecting what it's joining to be strings, and they're not. I was disappointed to see also that `sum([A,B]) != A + B`...

aapeliv: I've added a comment in code

app/pb/search.proto
aapeliv: Added a comment

aapeliv:
> That is some heavy postgres magic going on there!

Yes, the stuff here is pretty heavy on postgres/sqlalchemy stuff. I've added a new document to describe how it works and point to relevant postgres docs + document some implementation details.

aapeliv:
@kalvdans: rebased, cleaned, ready for merge after review!

app/backend/src/couchers/servicers/search.py
kalvdans: `sum()` have a second argument to start off the summation. Needed when concatenating lists like `sum(list_of_lists, [])`

kalvdans (approved):
Yay for extensive documentation!

PR #651, opened 2021-02-16
Change email + updates to change password

darrenvong:
Updated change password form to use the new `getAccountInfo` API to hide the "old password" field if the user doesn't have one. Also realised I didn't wrap it in a `PrivateRoute` for that in the last PR, so fixed that in this too.

Also did the same for the change email form.

### DIFF
--- a/app/frontend/src/features/auth/email/ChangeEmailPage.tsx
+++ b/app/frontend/src/features/auth/email/ChangeEmailPage.tsx
@@ -24,6 +24,15 @@ export default function ChangeEmailPage() {
   const isMdOrWider = useMediaQuery(theme.breakpoints.up("md"));
 
   const {
+    handleSubmit,
+    register,
+    reset: resetForm,
+  } = useForm<ChangeEmailFormData>();
+  const onSubmit = handleSubmit(({ currentPassword, newEmail }) => {
+    changeEmail({ currentPassword, newEmail });
+  });
+
+  const {
     data: accountInfo,
     error: accountInfoError,
     isLoading: isAccountInfoLoading,
@@ -36,13 +45,13 @@ export default function ChangeEmailPage() {
     mutate: changeEmail,
   } = useMutation<Empty, GrpcError, ChangeEmailFormData>(
     ({ currentPassword, newEmail }) =>
-      service.account.changeEmail(newEmail, currentPassword)
+      service.account.changeEmail(newEmail, currentPassword),
+    {
+      onSuccess: () => {
+        resetForm();
+      },
+    }
   );
-
-  const { handleSubmit, register } = useForm<ChangeEmailFormData>();
-  const onSubmit = handleSubmit(({ currentPassword, newEmail }) => {
-    changeEmail({ currentPassword, newEmail });
-  });
 
   return (
     <>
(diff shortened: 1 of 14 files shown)


app/frontend/src/features/auth/email/ChangeEmailPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+  } = useAccountInfo();
+
+  const {
+    error: changeEmailError,
+    isLoading: isChangeEmailLoading,
+    isSuccess: isChangeEmailSuccess,
+    mutate: changeEmail,
+  } = useMutation<Empty, GrpcError, ChangeEmailFormData>(
+    ({ currentPassword, newEmail }) =>
+      service.account.changeEmail(newEmail, currentPassword)
```
lucaslcode: I think clearing the form with `reset()` on success would be nice. This applies to all the account-related forms.

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+                type="password"
+                fullWidth={!isMdOrWider}
+              />
+            )}
+            <TextField
+              id="newEmail"
+              inputRef={register({ required: true })}
+              label="New email"
+              name="newEmail"
+              fullWidth={!isMdOrWider}
```
lucaslcode, line 90: Still not sure this is ideal...

lucaslcode (changes requested):
Nice work and thorough testing. One minor thing

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
darrenvong: That wouldn't work though if we want to display the success message without keeping track of that in a separate variable, since clearing on success would also clear that it's succeeded. Not sure I'm too keen on creating a separate variable just for that.

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
darrenvong: What's your concern with it?

darrenvong: Tangentially related and not sure of a good way to solve this either... but I really dislike the text field on Material UI forcing us to hardcode a width value if we don't want full width and it doesn't support letting the label taking up whatever width it needs properly. It's gonna be brittle when we eventually localise the app, when we don't know how long each translated text is gonna be...

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
lucaslcode: I think it will work, you call `reset` from `useForm`, that doesn't affect the mutation right?

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
lucaslcode: Just setting a visual-related field on every single field. Maybe a solution will be more obvious once the hifi is implemented. 

Re: the other comment, isn't that a problem no matter what? We could always not use the Mui field.

lucaslcode: Are there any text fields in the hifi?

darrenvong: Had a quick look and they look more like traditional text fields - box with rounded borders, but with the label outside of the text field which will solve this problem actually. Better UX/accessibility in my opinion too, I've not been a fan of that shrinking label as soon as it's focused/has value in it.

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
darrenvong: Oh yeah that would work, sorry got confused since the comment is below the mutation!

lucaslcode (approved):

darrenvong:
Relates to #368. Tagging so it shows up in that issue too

PR #655, opened 2021-02-18
does not create excessive WALs which can cause db crash

OmerBaddour:


kalvdans:
I didn't know postgres kept history even though no replication server has been configured. What is the error from postgres you are seeing?

aapeliv:
I diagnosed a bit with Omer last night. Docker creates a tmpfs and he had only allocated 2 GB of memory to docker. Because we keep on rebuilding and tearing down all of postgis, it creates a lot of changes in postgres that then causes lots of WAL.

> I didn't know postgres kept history even though no replication server has been configured.

It does, it's the way in which it guarantees integrity in the case of unexpected power loss. The whole replication server thing is really a hack on top of the Write Ahead Log (well not a hack anymore, it has grown and matured). Setting it to `minimal` means it's enough for recovery from bad shutdowns but not enough for replication (physical or logical).

aapeliv (approved):

PR #658, opened 2021-02-18
Community page components

lucaslcode:
- [x] refactor to use react-query
- [x] Components:
  - [x] Header image (will use map for now)
  - [x] Circular icon button
  - [x] Section title
  - [x] Horizontal scroller
  - [x] Load more card
  - [x] Place card
  - [x] Event card (placeholder)
  - [x] Discussion list items ~and list~
- [ ] Assemble


darrenvong:
Haven't looked at this properly yet but this is only part 1 and already over 1k loc changed?? 😱

lucaslcode:
Yes but it might be only one part actually, and also half the lines are Storybook boilerplate!

app/frontend/package.json
```diff
@@ -17,6 +16,7 @@
     "react-dom": "^17.0.0",
     "react-hook-form": "^6.15.1",
     "react-intersection-observer": "^8.31.0",
+    "react-lines-ellipsis": "^0.14.1",
```
lucaslcode, line 19: Don't really like this dependency, though small, it has no types. But I didn't want to waste time solving this myself. (This being having ellipsis overflow with an arbitrary number of lines.)

app/frontend/src/components/Card.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {
+  Card,
+  CardActionArea,
+  CardActions,
+  CardContent,
+  CardMedia,
+} from "./Card";
+
+export default {
+  title: "Components/Simple/Card",
```
lucaslcode: Is this simple or compound

app/frontend/src/components/CircularProgress.tsx
```diff
@@ -6,9 +6,9 @@ import {
 import classNames from "classnames";
 import React, { ForwardedRef } from "react";
 
-const useStyles = makeStyles({
+const useStyles = makeStyles((theme) => ({
   root: {},
-});
+}));
```
lucaslcode, line 11: Should revert this

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -0,0 +1,56 @@
...
+    flexDirection: "row",
+    alignItems: "center",
+    overflowX: "auto",
+    WebkitOverflowScrolling: "touch",
+    padding: theme.spacing(2),
+    "& > *": {
+      flex: "0 0 auto",
+    },
+  },
+  loader: {},
```
lucaslcode:
```suggestion
```

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -0,0 +1,56 @@
...
+/// TODO: Padding is not showing at end of list (maybe storybook only?)
+
+interface HorizontalScrollerProps {
+  fetchNext?: () => void;
+  isFetching?: boolean;
+  className?: string;
+  children?: ReactNode;
+}
+
+export default function HorizontalScroller({
```
lucaslcode, line 31: This scroller is good for mobile only. The custom css scrollbar solution is crap (they don't disappear) and default mobile scrollbar looks good.
However I think horizontal scrollers are bad design on desktop anyway, so I think it's fine?

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+    },
+  },
+}));
+
+interface CircularIconButtonProps extends IconButtonProps {
+  children?: ReactNode;
+  onClick: () => void;
+}
+
+export default function CircularIconButton({
```
lucaslcode, line 20: This component looks a bit out of place. I might be sneaky and change the design later...

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+              <Skeleton className={classes.userLoading} />
+            )}{" "}
+            • {posted}
+          </Typography>
+          <Typography variant="h2" component="h3" className={classes.title}>
+            {discussion.title}
+          </Typography>
+          <Typography variant="body1">{textTruncated}</Typography>
+          <div className={classes.replies}>
+            {(thread?.pages.length ?? 0) > 0 ? (
```
lucaslcode, line 97: listing the first 3 replies because the card seemed really bare otherwise. No way to get number of replies currently, and no up/downvoting or tags.

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+          <Typography
+            variant="caption"
+            component="p"
+            className={classes.subtitle}
+            noWrap
+          >
+            Created by {event.creatorName}
+          </Typography>
+          <ul className={classes.detailsList}>
+            {/*Leaving this here but out for now. Seem
```
lucaslcode, line 72: note

app/frontend/src/stories/__mocks__/service.ts
```diff
@@ -137,3 +141,86 @@ export const hostRequest = {
   lastSeenMessageId: 0,
   latestMessage: message1,
 };
+
+export const community = {
```
lucaslcode: This file doesn't seem like a good place for this mock data. TBH I'm a bit iffy about the whole file. Is there a more accepted way to mock data in storybook without redefining it in each story?

app/frontend/src/utils/stripMarkdown.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable no-useless-escape */
```
lucaslcode: Was getting this error but too scared to remove them. I think it's not correct?

lucaslcode:
Just components for now without assembling into the page, since it's big already.

darrenvong:
Apparently there's a type error that failed the build. Also are there any communities data on the test environment that can actually test this without pulling this branch down locally?

app/frontend/src/components/Card.stories.tsx
darrenvong: I'd say compound based on the number of sub-components that are used with this, since `Card` without any of them is a bit useless on its own?

app/frontend/src/components/Card.tsx
```diff
@@ -0,0 +1,28 @@
...
+  CardActionAreaProps,
+  CardActions as MuiCardActions,
+  CardActionsProps,
+  CardContent as MuiCardContent,
+  CardContentProps,
+  CardMedia as MuiCardMedia,
+  CardMediaProps,
+  CardProps,
+} from "@material-ui/core";
+
```
darrenvong: Are there any real benefits of creating these components if they are doing nothing other than wrapping the same component from Material UI anyway?

app/frontend/src/components/HorizontalScroller.tsx
darrenvong: Do you mean it's crap because the horizontal scrollbar always remain on screen? Yeah I think it's an anti-pattern to hide that on desktop since the user won't really expect to scroll like they do on a touch screen unless there's some visual feedback that there are something for them to actually scroll along.

Also agreed on horizontal scrollers (they're basically more or less like carousels?) being a bad UI on desktop generally. It's easy for users to miss contents. I much prefer if we render the initial items and have a "show all/show more" button that'd either show more items below it, or take the user to a dedicated pages that show them all the items. Maybe something for the UI/UX meeting to discuss.

app/frontend/src/features/communities/CommunityPage.tsx
```diff
@@ -21,48 +18,87 @@ import {
...
+  useListNearbyUsers,
+  useListPlaces,
+  useListSubCommunities,
+} from "./useCommunity";
 
 export default function CommunityPage() {
-  const [loading, setLoading] = useState(false);
-  const [error, setError] = useState("");
-  const [community, setCommunity] = useState<Community.AsObject | null>(null);
+  const history = useHistory();
```
darrenvong, line 35: I assume in your "assemble" task, you're gonna break up this giant PITA file right? This component is doing too many things...

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+                      variant="body2"
+                      className={classes.replies}
+                      key={reply.threadId}
+                    >
+                      {replyUsers?.get(reply.authorUserId) ? (
+                        firstName(replyUsers.get(reply.authorUserId).name)
+                      ) : (
+                        <Skeleton className={classes.userLoading} />
+                      )}
+                      : {stripMarkdown(reply.content)}
```
darrenvong, line 112: Doesn't the markdown editor already sanitise text before it gets sent?

Also does this mean replies don't get rendered with formatting at all?

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    marginBottom: 0,
+  },
+  subtitle: { marginBottom: theme.spacing(0.5) },
+  icon: {
+    display: "block",
+    marginInlineEnd: theme.spacing(0.5),
+    fontSize: "1rem",
+  },
+  detailsList: {
+    "ul&": {
```
darrenvong, line 30: Is this a specificity hack on Material UI's default stylings?

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+      : [];
+  const { data: replyUsers } = useUsers(replyUserIds);
+
+  const date = timestamp2Date(discussion.created);
+  const posted = timeAgo(date, false);
+  const strippedText = useMemo(
+    () => stripMarkdown(discussion.content.replace("\n", " ")),
+    [discussion.content]
+  );
+  const textTruncated =
```
darrenvong, line 69: Couldn't this be done in CSS instead? Pretty sure that's what `textOverflow` is for - https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
```diff
@@ -0,0 +1,44 @@
...
+      position: "relative",
+      left: "50%",
+      right: "50%",
+      marginLeft: "-50vw",
+      marginRight: "-50vw",
+    },
+    [theme.breakpoints.up("md")]: {
+      width: "100%",
+      height: "16rem",
+    },
```
darrenvong:
```suggestion
    [theme.breakpoints.up("md")]: {
      position: "relative",
      left: "50%",
      right: "50%",
      marginLeft: "-50vw",
      marginRight: "-50vw",
      width: "100%",
      height: "16rem",
    },
```
These styles could be merged?

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
```diff
@@ -0,0 +1,44 @@
...
+const useStyles = makeStyles((theme) => ({
+  root: {
+    width: "100vw",
+    height: "8rem",
+    [theme.breakpoints.up("md")]: {
+      position: "relative",
+      left: "50%",
+      right: "50%",
+      marginLeft: "-50vw",
+      marginRight: "-50vw",
```
darrenvong, line 17: Is this trying to do some kinda centering on larger screens? Can we be sure 50% is always 50vw though?

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+export default function DiscussionCard({
+  discussion,
+  className,
+}: {
+  discussion: Discussion.AsObject;
+  className?: string;
+}) {
+  const classes = useStyles();
+  const { data: creator } = useUser(discussion.creatorUserId);
+  //although we are only using 1 page of query, still use infinite as that
```
darrenvong, line 46: Does that mean this query is being/going to be reused elsewhere that need to be multiple pages? Otherwise what's the advantage of this over a normal `useQuery`? Not sure I follow the "this is how its stored" comment.

app/frontend/src/features/communities/useCommunity.ts
```diff
@@ -0,0 +1,122 @@
...
+export const useListNearbyUsers = (communityId?: number) =>
+  useInfiniteQuery<ListNearbyUsersRes.AsObject, GrpcError>(
+    communityNearbyUsersKey(communityId!),
+    ({ pageParam }) =>
+      service.communities.listNearbyUsers(communityId!, pageParam),
+    {
+      enabled: !!communityId,
+      getNextPageParam: (lastPage) =>
+        lastPage.nextPageToken ? lastPage.nextPageToken : undefined,
+    }
```
darrenvong, line 121: Same question as on `DiscussionCard` - considering we're only displaying single page of content on the "landing page" of a community at the moment.

app/frontend/package.json
darrenvong: Saw that you typed it yourself - not popular enough for it to have types available on DefinitelyTyped?

app/frontend/src/stories/__mocks__/service.ts
darrenvong: Not that I know of because we're using grpc... usually a library that mocks the network request is enough but because we're using grpc and all the responses are in buffer, it makes this a bit of a pain the arse really when we're outside of jest.

This is doing the auto-mocking of what jest would have done so I think it's close to the best we can do really. The iffy bit is there's no easy way to reset/clean up the mocks so it's possible that when we move between Storybook entries that use the same mock, if one story file alters the mock, it could break the other story, so that's the only iffy bit. I suspect we'll need to some auto-cleanup inside the proxy after it's called or something 🤷 

Although I think it might be a good idea to not dump things in here and only create the minimal object structure required for individual stories to override the mocks as and when they need to. This mimics quite well how the tests are doing this as it just makes instrumenting the mocks to respond with a success/fail promise a lot easier than what this essentially has become a global variable dump mess otherwise.

app/frontend/src/utils/stripMarkdown.ts
darrenvong: Which bit is the linting erroring on?

darrenvong: Also how come this source code is slightly different from the repo you linked? I assume you didn't just install it since it comes with another markdown editor that we don't need?

darrenvong (changes requested):
Code looks mostly fine, nice bit of refactoring going here! Need to look around the app a bit more once the build/type error is fixed.

app/frontend/package.json
lucaslcode: Sadly not

app/frontend/src/stories/__mocks__/service.ts
lucaslcode: agreed and I will do this the same way you did.

app/frontend/src/utils/stripMarkdown.ts
lucaslcode: Whoops! It's actually https://github.com/stiang/remove-markdown

lucaslcode: -1 to the line number
```
/home/lucas/couchers/app/frontend/src/utils/stripMarkdown.ts
  33:24  warning  Unnecessary escape character: \*  no-useless-escape
  33:28  warning  Unnecessary escape character: \+  no-useless-escape
  36:49  warning  Unnecessary escape character: \*  no-useless-escape
  36:53  warning  Unnecessary escape character: \+  no-useless-escape
  53:20  warning  Unnecessary escape character: \-  no-useless-escape
  55:27  warning  Unnecessary escape character: \:  no-useless-escape
  58:17  warning  Unnecessary escape character: \!  no-useless-escape
  58:29  warning  Unnecessary escape character: \[  no-useless-escape
  58:31  warning  Unnecessary escape character: \(  no-useless-escape
  58:40  warning  Unnecessary escape character: \)  no-useless-escape
  60:27  warning  Unnecessary escape character: \[  no-useless-escape
  60:29  warning  Unnecessary escape character: \(  no-useless-escape
  60:38  warning  Unnecessary escape character: \)  no-useless-escape
  71:19  warning  Unnecessary escape character: \*  no-useless-escape
  72:19  warning  Unnecessary escape character: \*  no-useless-escape
```

lucaslcode: > In most regex flavors, the only special characters or metacharacters inside a character class are the closing bracket ], the backslash \, the caret ^, and the hyphen -. The usual metacharacters are normal characters inside a character class, and do not need to be escaped by a backslash. To search for a star or plus, use [+*]. Your regex will work fine if you escape the regular metacharacters inside a character class, but doing so significantly reduces readability.

I will fix them.

app/frontend/src/components/Card.tsx
lucaslcode: I thought we may want to style the cards, and we can do it in the file?

app/frontend/src/features/communities/CommunityPage.tsx
lucaslcode: Yep

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
lucaslcode: I specifically don't want any markdown since this is a truncated preview, not the proper reply. Actually I forgot to add `noWrap` here.

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
lucaslcode: For some reason the default browser styling for ul wasn't being overriden unless I did this. I guess specificity.

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
lucaslcode: Sadly not, it only works for text content on one line. That is why I needed the `react-line-ellipsis` component (but I don't use it here)

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
lucaslcode: Whoops, the bottom one is supposed to be desktop

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
lucaslcode: It's to escape the container on mobile (changed to `breakpoints.down["sm"]`, and display full width. I prefer this solution than using the newly-implemented full screen routes, since I want the container except for this component.

> Can we be sure 50% is always 50vw though?
No, but it should be if it's used in a centered container. [Taken from here.
](https://css-tricks.com/full-width-containers-limited-width-parents/#no-calc-needed)

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
lucaslcode: Yes, I thought something will break if `["thread", threadId]` is used for both an infinite query and a normal query (no `pages` in a normal query).

app/frontend/src/features/communities/useCommunity.ts
lucaslcode: Yes I figured these will all be needed soon.

app/frontend/src/stories/__mocks__/service.ts
```diff
@@ -1,44 +1,10 @@
...
-  name: "Funny Kid",
-  userId: 3,
-  username: "funnykid",
-  avatarUrl: funnykid,
-} as User.AsObject;
-
-const userMap = new Map(
-  [user1, user2, user3].map((user) => [user.userId, user])
-);
+const userMap = new Map(users.map((user) => [user.userId, user]));
```
darrenvong, line 7: Hope the conflict I'll get here won't be too hard to fix... this map thing is driving me nuts since it doesn't work with username to user (I need that for the user's page stuff 😠 )

app/frontend/src/stories/__mocks__/service.ts
darrenvong: Another idea for new storybook entries, maybe `useEffect` cleanup could be used to clean up the proxy between story files?

I think I didn't do it because the friends connection stories had some hacky stuff with refs for the mocks to return the updated state after the mutation so it didn't work, but only just realised it shouldn't matter for stories that haven't been written with that hack. Probably need to re-visit as a low priority thing later to figure out a better way to write it...

app/frontend/src/components/Card.tsx
darrenvong: Are we going off the assumption from now we won't rely on the base Material UI styles from their simple components and will need to override them?

app/frontend/src/stories/__mocks__/service.ts
lucaslcode: Can you just change the mock so if it the key doesn't parse as a number, it does a username lookup?

darrenvong: I ended up changing it back to a regular object so the mock just do a look up on this. It's only 3 users so not that bad to write it out in full for both IDs and usernames.

darrenvong:
Are there any test data in the test db to check this without pulling down the branch?

lucaslcode:
Only the world community:
https://9203377f--frontend.preview.coucher.org/community/1/world

There's mostly storybook changes only here anyway

app/frontend/src/components/CircularProgress.tsx
darrenvong: Why do we need to keep this in then? 🤔

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
darrenvong: Amazing... the funky tricks available in CSS is mind bending sometimes 😮

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+    title: string;
+    creatorName: string;
+    location: string;
+    startTime: { seconds: number; nanos: number };
+  };
+  className?: string;
+}) {
+  const classes = useStyles();
+  const date = timestamp2Date(event.startTime);
+  return (
```
darrenvong, line 56: Not sure if this is only in Storybook... but I feel this should show the whole name since the test name you got isn't even that long?
Screenshot 2021-02-26 at 00 41 27

lucaslcode: Maybe I'll change it to "by". I don't want it to wrap.

app/frontend/src/components/CircularProgress.tsx
lucaslcode: We will definitely have a custom loading indicator with the logo

app/frontend/src/components/Card.tsx
lucaslcode: I'll remove it.

app/frontend/src/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+import classNames from "classnames";
+import React, { useMemo } from "react";
+import LinesEllipsis from "react-lines-ellipsis";
+
+import { Page } from "../../../pb/pages_pb";
+import stripMarkdown from "../../../utils/stripMarkdown";
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    maxWidth: 160,
```
darrenvong, line 18: Since the text gets truncated, wouldn't we want to set the `width` instead so they all have the same width?

Although not sure about having magic numbers like this here. Feel like it's probably better to let the parent component that will presumably render these cards in a list to decide its width depending on different screen sizes etc.

darrenvong (dismissed):
Left a comment but I think it may be better to address it in your assemble step

darrenvong (approved):

PR #660, opened 2021-02-19
[MRG] Bug: Conversion error

ShmuelTreiger:


aapeliv (approved):
Thanks for spotting!

kalvdans:
Aiee, what a blunder! A hole in our unit tests.

ShmuelTreiger:
My blunder, tbf. And discovered while creating unit tests for it, so it worked in a way

PR #661, opened 2021-02-19
[MRG] Fixing unknown to unspecified

ShmuelTreiger:
So I see that I messed this up when I did my initial changes.

That being said, I don't 100% understand the difference? When would it be unspecified vs unknown?


aapeliv:
> That being said, I don't 100% understand the difference? When would it be unspecified vs unknown?

The profile update function allows you to do one of three things:

1. Unspecified: "I don't want to change my sleeping arrangement field, I just wanted to change my display name"
2. Unknown: "I want to update my sleeping arrangement to be empty/unset/null/unknown (i.e. I want to remove the value)"
3. Anything else: "I want to set my S.A. to 'shared room'"

The first one, (1) allows us to recycle the same profile update function, so you can update other parts of your profile without touching your sleeping arrangement. This is because in protobuf every field^ must have a value: if you don't set it, you get the default. For enumerations the default is 0. So if the frontend doesn't set a sleeping arrangement value, it'll be 0, whatever that happens to be in the enum.

(^: it applies to primitive fields such as int/enum/string/etc, but doesn't always apply, e.g. you can detect whether a submessage is present or not)

app/backend/src/couchers/servicers/api.py
```diff
@@ -365,8 +365,8 @@ def UpdateProfile(self, request, context):
                 else:
                     user.other_host_info = request.other_host_info.value
 
-            if request.sleeping_arrangement != api_pb2.SLEEPING_ARRANGEMENT_UNKNOWN:
-                user.sleeping_arrangement = sleepingarrangement2sql[request.sleeping_arrangement]
+            if request.sleeping_arrangement != api_pb2.SLEEPING_ARRANGEMENT_UNSPECIFIED:
+                user.sleeping_arrangement = smokinglocation2sql[request.sleeping_arrangement]
```
aapeliv: Now you undid this smoking locaiton change

aapeliv (changes requested):

ShmuelTreiger:
Got it, that makes sense, thanks. And thanks for catching my reversion.

aapeliv (approved):
Looks good if tests pass!

PR #664, opened 2021-02-21
Storybook entries for account management pages

### LINKED ISSUE #368: Account management frontend

darrenvong:
Completes and closes #368... should have added this in my last PR rather than waiting for the big tidy up first lol, but I guess that helps give some structure for where to put this

### DIFF
--- a/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
+++ b/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
@@ -135,7 +135,7 @@ describe("ChangePasswordPage", () => {
       // gonna be true
       expect(await screen.findByLabelText("New password")).toBeVisible();
       expect(
-        screen.getByText("Please enter a password")
+        screen.getByText("Please enter a password.")
       ).toBeVisible();
       expect(screen.queryByLabelText("Old password")).not.toBeInTheDocument();
     });
--- a/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
+++ b/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
@@ -135,7 +135,7 @@ describe("ChangePasswordPage", () => {
       // gonna be true
       expect(await screen.findByLabelText("New password")).toBeVisible();
       expect(
-        screen.getByText("Leave the fields blank to unset your password.")
+        screen.getByText("Please enter a password")
       ).toBeVisible();
       expect(screen.queryByLabelText("Old password")).not.toBeInTheDocument();
     });
--- a/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
+++ b/app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
@@ -35,6 +35,11 @@ describe("ChangePasswordPage", () => {
         screen.getByRole("heading", { name: "Change password" })
       ).toBeVisible();
       expect(await screen.findByLabelText("Old password")).toBeVisible();
+      expect(
+        screen.getByText(
+          /Please enter a password, or leave the "New password" and "Confirm password" fields blank/i
+        )
+      ).toBeVisible();
       expect(screen.getByLabelText("New password")).toBeVisible();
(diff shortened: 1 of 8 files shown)


lucaslcode (changes requested):
Great, much neater than how I was doing it. I will copy it.

Just something I didn't pick up on before, the message shouldn't say "Leave the fields blank to unset your password" if the user doesn't have a password already.

app/frontend/src/features/auth/password/ChangePasswordPage.tsx
```diff
@@ -82,8 +82,10 @@ export default function ChangePasswordPage() {
             </Alert>
           )}
           <Typography className={classes.infoText} variant="body1">
-            Please enter a new password, or leave the "New password" and
-            "Confirm password" fields blank to unset your password.
+            Please enter a new password,{" "}
+            {accountInfo?.hasPassword
+              ? 'or leave the "New password" and "Confirm password" fields blank to unset your password.'
+              : "or leave the fields blank to unset your password."}
```
lucaslcode: Still the same problem, no? The user has no password and they see `Please enter a new password, or leave the fields blank to unset your password.` But they already don't have a password.

darrenvong: Oh yeah what was I thinking 🤦

app/frontend/src/features/auth/password/ChangePasswordPage.tsx
```diff
@@ -82,8 +82,9 @@ export default function ChangePasswordPage() {
             </Alert>
           )}
           <Typography className={classes.infoText} variant="body1">
-            Please enter a new password, or leave the "New password" and
-            "Confirm password" fields blank to unset your password.
+            {accountInfo?.hasPassword
+              ? 'Please enter a password, or leave the "New password" and "Confirm password" fields blank to unset your password.'
+              : "Leave the fields blank to unset your password."}
```
lucaslcode:
```suggestion
              : "Please enter a password."}
```

app/frontend/src/features/auth/password/ChangePasswordPage.test.tsx
```diff
@@ -129,6 +134,9 @@ describe("ChangePasswordPage", () => {
       // Wait for new password field/form to show up first, otherwise old password not visible is always
       // gonna be true
       expect(await screen.findByLabelText("New password")).toBeVisible();
+      expect(
+        screen.getByText("Leave the fields blank to unset your password.")
```
lucaslcode:
```suggestion
        screen.getByText("Please enter a password")
```

lucaslcode (changes requested):
😊

app/frontend/src/features/auth/password/ChangePasswordPage.tsx
darrenvong: I did think after I committed (and was too late to turn back on my laptop 💤 ) if it makes any sense to clear your password if you don't have one already 😂

lucaslcode (changes requested):
Test failed. But it looks like it should have worked???

darrenvong:
It's because it's a string value and so require precise match, but the value missed a full stop 😬

lucaslcode (approved):
If it's green

PR #668, opened 2021-02-21
Schema

OmerBaddour:
Added .pngs of the currently used and eventual schema, as well as a file containing the images and instructions to generate them yourself


kalvdans:
The png:s seems to be empty files.

aapeliv:
> The png:s seems to be empty files.

Odd, in the tree they aren't: https://github.com/Couchers-org/couchers/blob/schema/docs/full_er_diagram.png. Maybe a bug in GitHub?

kalvdans:
> Odd, in the tree they aren't: https://github.com/Couchers-org/couchers/blob/schema/docs/full_er_diagram.png. Maybe a bug in GitHub?

Seems so. Probably because they are huge, both in filesize and in resolution. Can we ask DBeaver for vector graphics? (too lazy to install it myself...)

kalvdans (approved):
Looks good to me regardless of png or svg graphics.

Issue #601, opened 2021-02-08
Add location to signup


lucaslcode:
Currently it is only a text field "city", which is ignored when the user sets their location from jail

aapeliv:
What's needed from backend for this?

I'm not sure what the current design is like, but would it be okay to have two API calls: basic details, and then the other stuff? I'm thinking we'd want to put the higher effort stuff on a separate screen so if people think it's too much, we still get their details and can spam them to finish signing up 😈

lucaslcode:
Don't know if we need two calls. Just need to add lat/lng and TOS agreement to the CompleteSignup call

aapeliv:
OK, I can do that

PR #675, opened 2021-02-22
Add location and accepted tos to CompleteSignupReq

aapeliv:
Does backend part of #601


app/pb/auth.proto
```diff
@@ -115,10 +115,16 @@ message CompleteSignupReq {
...
   string name = 3;
-  string city = 4;
   string birthdate = 5; // in YYYY-MM-DD format, in the timezone of the birth place
   string gender = 6;
   org.couchers.api.core.HostingStatus hosting_status = 7;
+
+  string city = 8;
+  double lat = 9;
+  double lng = 10;
+  double radius = 11;
```
kalvdans: Document units :)

aapeliv: Done

app/backend/src/couchers/servicers/auth.py
```diff
@@ -232,10 +232,13 @@ def CompleteSignup(self, request, context):
                 email=signup_token.email,
                 username=request.username,
                 name=request.name,
-                city=request.city,
                 gender=request.gender,
                 birthdate=birthdate,
                 hosting_status=hostingstatus2sql[request.hosting_status],
+                city=request.city,
+                geom=create_coordinate(request.lat, request.lng),
+                geom_radius=request.radius,
```
lucaslcode, line 243: No checks to see if a valid location was sent?

lucaslcode (commented):
Should I do a frontend branch on top of this?

app/backend/src/couchers/servicers/auth.py
aapeliv: These are not checked elsewhere either. I think postgis will complain if the given coordinate is invalid (or it might wrap it around, I'm not entirely sure).

For the radius, we might add database check constraints to enforce it being in some sensible range?

aapeliv:
> Should I do a frontend branch on top of this?

That'd be great, thanks!

app/backend/src/couchers/servicers/auth.py
lucaslcode: I mean more, if for some reason it isn't included, the user will be at null island. Maybe that's fine and the frontend has responsibility for making sure it's set?

aapeliv: I think that was the conclusion last time, I think someone was very vocal about not mixing serialization layer stuff with business logic... I'd be more than happy to add checks to make sure coordinates aren't on null island now though.

lucaslcode: Oh yes, that was me. I do still kinda think people should be *able* to have 0,0 as a location. But now I also think it's splitting hairs.

aapeliv: Added some checks

lucaslcode (dismissed):

kalvdans (dismissed):
Backend part looks ok for me!

lucaslcode (approved):

PR #678, opened 2021-02-22
Simple message emails

### LINKED ISSUE #175: Email alerts for messages
Putting this in alpha milestone but that is probably optimistic!

aapeliv:
Closes #175

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -7,11 +7,11 @@ from datetime import timedelta
 
 from sqlalchemy.sql import func, or_
 
-from couchers import config
+from couchers import config, email, urls
 from couchers.db import session_scope
 from couchers.email.dev import print_dev_email
 from couchers.email.smtp import send_smtp_email
-from couchers.models import GroupChat, GroupChatSubscription, LoginToken, Message, SignupToken, User
+from couchers.models import GroupChat, GroupChatSubscription, LoginToken, Message, MessageType, SignupToken, User
 from couchers.utils import now
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -40,3 +44,74 @@ def process_purge_signup_tokens(payload):
...
+
+            unseen_messages = (
+                session.query(GroupChat, Message, subquery.c.count_unseen)
+                .join(subquery, subquery.c.message_id == Message.id)
+                .join(GroupChat, GroupChat.conversation_id == subquery.c.group_chat_id)
+                .order_by(subquery.c.message_id.desc())
+                .all()
+            )
+
+            user.last_notified_message_id = max([message.id for _, message, _ in unseen_messages])
```
kalvdans: Drop the [ ], to instead pass a *generator expression* to `max`. Two characters less to write and doesn't create an intermediary array.

app/backend/templates/unseen_messages.md
```diff
@@ -0,0 +1,36 @@
...
+
+{% endfor %}
+
+{% if unseen_messages|length > 5 %}
+... and {{ total_unseen_message_count - 5 }} more messages in {{ unseen_messages|length - 5}} other chats!
+{% endif %}
+
+Check it out here:
+
+{% if html %}
```
kalvdans, line 26: So much for writing one markdown to generate both text and html, haha!

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -40,3 +44,74 @@ def process_purge_signup_tokens(payload):
...
+            session.query(User)
+            .join(GroupChatSubscription, GroupChatSubscription.user_id == User.id)
+            .join(Message, Message.conversation_id == GroupChatSubscription.group_chat_id)
+            .filter(Message.time >= GroupChatSubscription.joined)
+            .filter(or_(Message.time <= GroupChatSubscription.left, GroupChatSubscription.left == None))
+            .filter(Message.id > User.last_notified_message_id)
+            .filter(Message.id > GroupChatSubscription.last_seen_message_id)
+            .filter(Message.time < now() - timedelta(minutes=30))
+            .filter(Message.message_type == MessageType.text)  # TODO: only text messages for now
+            .group_by(User)
```
kalvdans: Is the group_by needed if you are only asking for a list of users?

kalvdans (changes requested):
Looks fine, but I think we need unit tests to sleep well, not worrying that we have spammed everyone overnight :)

app/backend/src/couchers/jobs/handlers.py
aapeliv: Probably not, seems to be leftover from trying other things.

app/backend/templates/unseen_messages.md
aapeliv: At least they're next to each other :)

Or we could change `button` to just emit `<{{ link }}>` in non-HTML mode

aapeliv:
> Looks fine, but I think we need unit tests to sleep well, not worrying that we have spammed everyone overnight :)

I'll add a basic test, but this is annoying to test, and I'm hoping we'll rip it out soon with a better notification system...

aapeliv:
Fixed up comments and added two basic tests.

app/backend/src/couchers/migrations/versions/065886328b03_add_last_notified_message_id_to_.py
```diff
@@ -0,0 +1,25 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.add_column("users", sa.Column("last_notified_message_id", sa.BigInteger(), nullable=False))
+    op.execute("ALTER TYPE backgroundjobtype ADD VALUE 'send_message_notifications'")
+
+
+def downgrade():
+    raise Exception("Can't downgrade.")
```
kalvdans, line 25: Should have this in all migrations, half of the downgrades are probably broken because they are not tested :)

kalvdans (approved):
Looks good now with the tests!

app/backend/templates/unseen_messages.md
kalvdans: The way to do links in markdown is simply
```
[Friend Requests]({{ group_chats_link }})
```

app/backend/src/couchers/migrations/versions/065886328b03_add_last_notified_message_id_to_.py
aapeliv: That's true...

PR #681, opened 2021-02-23
Add required id prop to TextField component

### LINKED ISSUE #590: Make `id` a required prop for the TextField component
According to the [accessibility section in the docs for `TextField`](https://material-ui.com/components/text-fields/#accessibility), adding an `id` is needed for Material UI to label our text fields correctly, which is important for keeping our app accessible for screen readers.

It's quite a low hanging fruit considering the library does the hard work of doing the rest right so we should fix this some time. A good first issue for anyone looking to get started!

javiwq:
Closes #590 

TODO:
- [x]  Add required `id` prop to TextField component
- [x] Add unique id to instances of TextField in other components

### DIFF
--- a/app/frontend/src/components/TextField.tsx
+++ b/app/frontend/src/components/TextField.tsx
@@ -24,6 +24,7 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 export default function TextField({
+  id,
   className,
   ...otherProps
 }: Omit<OutlinedTextFieldProps, "variant">) {
@@ -31,6 +32,7 @@ export default function TextField({
   return (
     <MuiTextField
       {...otherProps}
+      id={id}
       variant="outlined"
       className={classNames(classes.root, className, {
         [classes.multiline]: otherProps.multiline,
--- a/app/frontend/src/components/TextField.tsx
+++ b/app/frontend/src/components/TextField.tsx
@@ -23,11 +23,15 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+interface TextfieldProps extends OutlinedTextFieldProps {
+  id: string;
+}
+
 export default function TextField({
   id,
   className,
   ...otherProps
-}: Omit<OutlinedTextFieldProps, "variant">) {
+}: Omit<TextfieldProps, "variant">) {
   const classes = useStyles();
   return (
     <MuiTextField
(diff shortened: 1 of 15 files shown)


app/frontend/src/features/auth/login/UsernameForm.tsx
```diff
@@ -74,6 +74,7 @@ export default function UsernameForm() {
     <>
       <form onSubmit={onSubmit}>
         <TextField
+          id="username/email"
```
lucaslcode:
```suggestion
          id="username-email"
```
I think the original is indeed valid html5 but having a slash isn't conventional I think.

lucaslcode (commented):
Thanks so much!
Looks good ~~but I will check the build tonight since our CI checks don't run on forks.~~ ✔️ 

Also I don't recognize your username, if you want to join the slack, send me an email (it should be in the git logs) (:

lucaslcode (changes requested):
Just that one thing!
Also, let me know if you'd like to to deal with the merge conflicts.

javiwq:
Thank you for your review! I committed that one change. :) 
Could you or someone with write access resolve the merge conflicts?

lucaslcode:
Will do!

app/frontend/src/components/TextField.tsx
```diff
@@ -23,14 +23,20 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+interface TextfieldProps extends OutlinedTextFieldProps {
```
Tay0108, line 26: This can just use TextFieldProps from material-ui since we have parametrized variant now

app/frontend/src/components/TextField.tsx
```diff
@@ -23,14 +23,20 @@ const useStyles = makeStyles((theme) => ({
...
+  id: string;
+}
+
 export default function TextField({
+  id,
   className,
   ...otherProps
-}: Omit<OutlinedTextFieldProps, "variant">) {
+}: Omit<TextfieldProps, "variant">) {
```
Tay0108, line 34: This can just use TextFieldProps from material-ui since we have parametrized variant now

Tay0108:
Made PR with resolved merge conflicts: https://github.com/Couchers-org/couchers/pull/693

lucaslcode:
This had to be in a new PR in order to merge, but your commits are there! Thanks!

PR #687, opened 2021-02-23
Actually test the arguments to the mocked function

kalvdans:
Two tests had to be disabled since they fails when the actual test code is run.

Do you want to fix the code for the failing tests in this PR or a separate PR?


aapeliv:
> Do you want to fix the code for the failing tests in this PR or a separate PR?

I can fix it later in another pr

aapeliv (approved):

PR #692, opened 2021-02-24
[MRG] Removing redundant code (A>B>C, don't need to check A>C)

ShmuelTreiger:


app/backend/src/tests/test_requests.py
```diff
@@ -51,15 +51,6 @@ def test_create_request(db):
...
 
-        with pytest.raises(grpc.RpcError) as e:
-            api.CreateHostRequest(
-                requests_pb2.CreateHostRequestReq(
-                    to_user_id=user2.id, from_date=today_plus_2, to_date=today_minus_2, text="Test request"
-                )
-            )
-        assert e.value.code() == grpc.StatusCode.INVALID_ARGUMENT
-        assert e.value.details() == errors.DATE_TO_BEFORE_TODAY
-
```
kalvdans: You can leave the test as it is, just don't be so picky about the error text. (remove the e.value.details() check)

That way, we can be sure that the logic is still the same despite the removed code.

kalvdans (dismissed):
Ok for me to merge with our without fixing my little remark

app/backend/src/tests/test_requests.py
ShmuelTreiger: Makes sense

app/backend/src/tests/test_requests.py
```diff
@@ -54,21 +54,21 @@ def test_create_request(db):
...
                 requests_pb2.CreateHostRequestReq(
-                    to_user_id=user2.id, from_date=today_minus_3, to_date=today_plus_2, text="Test request"
+                    to_user_id=user2.id, from_date=today_plus_2, to_date=today_minus_2, text="Test request"
                 )
             )
         assert e.value.code() == grpc.StatusCode.INVALID_ARGUMENT
-        assert e.value.details() == errors.DATE_FROM_BEFORE_TODAY
-
+        assert e.value.details() == errors.DATE_FROM_AFTER_TO
+
```
aapeliv: Please remove this whitespace

app/backend/src/couchers/servicers/requests.py
```diff
@@ -68,15 +68,16 @@ def CreateHostRequest(self, request, context):
...
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.DATE_FROM_BEFORE_TODAY)
 
-            if request.to_date < least_current_date():
-                context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.DATE_TO_BEFORE_TODAY)
-
+            # B >= C
             if request.from_date >= request.to_date:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.DATE_FROM_AFTER_TO)
 
+            # No need to check A > C
```
aapeliv: Could use `least_current_date()` instead of A, `request.from_date` instead of B, etc, but doesn't matter much.

aapeliv (commented):
Just small nitpick with whitespace. Should also be stripped if you run isort and black

aapeliv (approved):
Looks good to me! Thanks :)

Sorry for slow approval, was busy the last few days!

ShmuelTreiger:
No worries, I'm not in any rush

PR #697, opened 2021-02-25
Community page part 2

lucaslcode:
image
image
image
image


lucaslcode:
This is probably enough for now. Some things I'm not happy with, particularly the "nav buttons" on desktop, the cards on desktop, the "see more" button, and the font should be bigger on desktop (but we should implement that separately/not sure the best way).

lucaslcode:
Also probably some awful css in there, apologies in advance

lucaslcode:
I broke backend tests by messing with discussion ordering, but frontend preview still works: https://frontend-feature-community-page-assembly--frontend.preview.coucher.org/
However it is better running this locally - there is mock more community data

app/frontend/src/AppRoutes.tsx
```diff
@@ -127,15 +147,72 @@ export default function AppRoutes() {
...
+      </AppRoute>
+      <AppRoute
+        isPrivate
+        path={`${communityEventsRoute}/:communityId/:communitySlug?`}
+      >
+        Events
+      </AppRoute>
+      <AppRoute
+        isPrivate
+        path={`${communityRoute}/:communityId/:communitySlug?`}
```
darrenvong: It might be worth having the full template string in the `routes` file so it reads `path={communityPlacesRoute}` etc?

app/frontend/src/components/UserSummary.tsx
```diff
@@ -53,17 +54,16 @@ export default function UserSummary({ user }: UserSummaryProps) {
...
-              <Box className={classes.hostingAbilityContainer}>
-                {user ? <HostStatus user={user} /> : <Skeleton width={100} />}
-              </Box>
-            </Box>
+            <ScoreBar value={user?.verification || 0}>
+              Community Standing
+            </ScoreBar>
+            <div className={classes.hostingAbilityContainer}>
+              {user ? <IconText icon={CouchIcon} text={hostingStatusLabels[user.hostingStatus]} /> : <Skeleton width={100} />}
+            </div>
```
darrenvong: I moved the highlighted bit to a `HostingStatus` component in my last PR to make this more re-usable, since there were variations of it across the designs.

After chatting with @andreaowu on this, we used our creative license and agreed that fixing the avatar, user name/location summary and community standing bar, then make additional texts/components like this hosting status an optional thing that can be plugged back into this component the best way to keep it reusable without being awkward.

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
```diff
@@ -1,34 +1,64 @@
...
+    </div>
   );
+
+  if (linkTo)
+    return (
+      <Link to={linkTo} className={classes.link}>
+        {body}
+      </Link>
+    );
+  else return body;
```
darrenvong: Could use a ternary statement here instead since it's not too complex?

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
```diff
@@ -1,34 +1,64 @@
...
+        {...otherProps}
+        id={id}
+        className={classes.button}
+        children={children}
+        disabled={disabled}
+      />
+      <Typography variant="caption" align="center">
+        <label htmlFor={id}>{label}</label>
+      </Typography>
+    </div>
```
darrenvong:
```suggestion
      <IconButton
        {...otherProps}
        aria-labelledby={id}
        className={classes.button}
        children={children}
        disabled={disabled}
      />
      <Typography id={id} variant="caption" align="center">
        {label}
      </Typography>
    </div>
```
I don't think this is the right HTML to use, since `label`s are usually used with form inputs. Might be better to go with an `aria-labelledby` with the matching ID attached to the `Typography` component instead as that should work correctly according to the dev tool's accessibility tree tab.

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -0,0 +1,445 @@
...
+        flexBasis: `calc(33.33% - ${theme.spacing(1)})`,
+      },
+    },
+  },
+  discussionCard: {
+    marginBottom: theme.spacing(1),
+  },
+}));
+
+export default function CommunityPage() {
```
darrenvong, line 141: This is still a quite a big component... Could the places and discussions not be broken up further into its own with the relevant hooks moving in with them?

darrenvong (changes requested):
Only a few things I spotted from a quick look, may have to revisit as this is another beasty PR and so it's difficult to spot everything in one sitting 😅

app/frontend/src/AppRoutes.tsx
lucaslcode: Then the associated function can't use that variable

app/frontend/src/components/UserSummary.tsx
lucaslcode: I think this doesn't exist anymore after I merged in develop

app/frontend/src/AppRoutes.tsx
darrenvong: Couldn't we still leave that variable within the `routes` file and have this re-exposed as the variables themselves? So something like (better naming welcomed, although maybe it's not that confusing if you pronounce "route" the American way):

```typescript
const communityRootRoute = "/community";
export const communityRoute = `${communityRootRoute}/:communityId/:communitySlug?`;
```

Unless there are use cases where we'd ever go to `/community`, `/place` on its own that I overlooked?

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
```diff
@@ -1,34 +1,64 @@
...
+  body: {
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "center",
+  },
+  link: {
+    textDecoration: "none",
+  },
+  button: {
+    marginBottom: theme.spacing(0.5),
```
darrenvong, line 20: I've seen this in a few places now where we've resorted to using fractional values. Wonder if the 8px spacing scale is not granular enough/too rigid for some layouts and if we should move down to 4px intervals instead.

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -0,0 +1,233 @@
...
+        <div className={classes.topInfo}>
+          <Breadcrumbs aria-label="breadcrumb" className={classes.breadcrumbs}>
+            {community.parentsList
+              .filter((parent) => !!parent.community)
+              .map((parent) => (
+                <Link
+                  to={routeToCommunity(
+                    parent.community!.communityId,
+                    parent.community!.slug
+                  )}
```
darrenvong:
```suggestion
              .filter((parent) : parent is ParentWithCommunity => !!parent.community)
              .map((parent) => (
                <Link
                  to={routeToCommunity(
                    parent.community.communityId,
                    parent.community.slug
                  )}
```
where `ParentWithCommunity` is this interface extending `Parent.AsObject`: `{ community: CommunityParent.AsObject; }` seems to "correct" the type enough so you don't need non-null assertion

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -54,12 +57,19 @@ export default function EventCard({
   const classes = useStyles();
   const date = timestamp2Date(event.startTime);
   return (
-    <Card className={classNames(classes.root, className)}>
-      <CardActionArea>
+    <Card className={className}>
+      <Link
+        to={routeToEvent(event.eventId ?? 0, event.slug ?? "")}
+        className={classes.link}
+        component={CardActionArea}
```
darrenvong: Getting this warning in the console for this component:
```
Warning: Invalid value for prop `navigate` on <a> tag. Either remove it from the element, or pass a string or number value to keep it in the DOM. For details, see https://reactjs.org/link/attribute-behavior
```

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -54,12 +57,19 @@ export default function EventCard({
...
+      >
         <CardContent>
```
darrenvong:
```suggestion
      >
      <CardActionArea>
        <CardContent>
        ... some other stuff
      </CardActionArea>
```
Doing that seems to work, will have to fix the link styling to not get the horrible default blue text.

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -73,8 +77,12 @@ export default function DiscussionCard({
 
   return (
     <Card className={classNames(classes.root, className)}>
-      <CardActionArea>
-        <CardContent>
+      <Link
+        to={routeToDiscussion(discussion.discussionId, discussion.slug)}
+        className={classes.link}
+        component={CardActionArea}
```
darrenvong: Same comment as in `EventCard`

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <DialogContent>
+          {newDiscussionMutation.error && (
+            <Alert severity="error">
+              {newDiscussionMutation.error.message}
+            </Alert>
+          )}
+          <NewComment
+            onComment={async (content) =>
+              newDiscussionMutation.mutate({
+                title: "test",
```
darrenvong, line 120: Shouldn't be add another field to let the user specify a title as well? Every discussions is fixed with "test" as the discussion title otherwise...

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
```diff
@@ -0,0 +1,153 @@
...
+          discussions?.pages
+            .flatMap((res) => res.discussionsList)
+            .map((discussion) => (
+              <DiscussionCard
+                discussion={discussion}
+                className={classes.discussionCard}
+                key={`discussioncard-${discussion.threadId}`}
+              />
+            ))
+        )}
```
darrenvong: If you wrap all of this with an extra bracket, then you don't need the optional chaining on line 126. Bit of an expression hack so that `discussions` is known to be defined by TS.

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
```diff
@@ -9,35 +10,47 @@ const useStyles = makeStyles((theme) => ({
...
+
+  //display a map if there's no image
   //if no location, just display a zoomed out map of the world
-  const zoom = community.mainPage?.location ? 13 : 2;
+  const zoom = community.mainPage?.location ? 13 : 1;
   const lngLat = new LngLat(
     community.mainPage?.location?.lng ?? 0,
     community.mainPage?.location?.lat ?? 0
   );
+  console.log(community.mainPage);
```
darrenvong:
```suggestion
```

app/frontend/src/features/communities/CommunityPage/HeaderImage.tsx
```diff
@@ -9,35 +10,47 @@ const useStyles = makeStyles((theme) => ({
...
   return (
-    <Box className={classes.root}>
+    <Box className={classNames(classes.root, className)}>
       <Map grow interactive={false} initialCenter={lngLat} initialZoom={zoom} />
     </Box>
```
darrenvong:
```suggestion
    <div className={classNames(classes.root, className)}>
      <Map grow interactive={false} initialCenter={lngLat} initialZoom={zoom} />
    </div>
```

app/frontend/src/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -40,26 +50,36 @@ export default function PlaceCard({
...
-          image="https://loremflickr.com/320/240"
-          className={classes.image}
-        />
+    <Card className={className}>
+      <Link
+        to={routeToPlace(place.pageId, place.slug)}
+        className={classes.link}
+        component={CardActionArea}
+      >
+        <CardMedia src={place.photoUrl} className={classes.image} />
```
darrenvong: Getting this runtime warning in the console from MUI:

```
Warning: Failed prop type: Material-UI: Either `children`, `image`, `src` or `component` prop must be specified.
```

Might be worth falling back to some stock image if `photoUrl` is undefined?

app/frontend/src/features/communities/CommunityPage/PlacesSection.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <>
+      <SectionTitle icon={<InfoIcon />}>Places</SectionTitle>
+      {placesError && <Alert severity="error">{placesError.message}</Alert>}
+      {isPlacesLoading && <CircularProgress />}
+      <HorizontalScroller className={classes.cardContainer}>
+        {
+          //Is there a better way to check for empty state?
+          places &&
```
darrenvong: Probably not, considering the `pages[0].<typeList>` changes depending on the data type. Could abstract this into a helper but might be tricky to get it type safe without delving into some hairy generics 😅

app/frontend/src/features/communities/useCommunity.ts
```diff
@@ -120,3 +126,24 @@ export const useListNearbyUsers = (communityId?: number) =>
...
+    {
+      title: string;
+      content: string;
+      ownerCommunityId: number;
+    }
+  >(
+    ({ title, content, ownerCommunityId }) =>
+      service.discussions.createDiscussion(title, content, ownerCommunityId),
+    {
+      onSuccess(_data, { ownerCommunityId }) {
```
darrenvong:
```suggestion
      onSuccess(_, { ownerCommunityId }) {
```
Nit: since we aren't using `data` at all?

darrenvong (changes requested):

darrenvong:
> Some things I'm not happy with, particularly the "nav buttons" on desktop

Design only made a bare minimum start on the desktop hi-fi, but the position of where those are looking quite different from mobile again such that it could be a layout pain if they stick with it...

It's probably ok where they are for now as long we fix the ugly default blue text link styling on them?

darrenvong:
Oh and another general comment, should probably extract all UI strings (directly in the UI, in aria-labels and labels etc) into some constant file like what @andreaowu did?

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
lucaslcode: This is just placeholder to see if the discussions actually work

app/frontend/src/AppRoutes.tsx
lucaslcode: Done just for these ones, not previous

app/frontend/src/routes.ts
```diff
@@ -44,24 +44,55 @@ export const routeToSearch = (query: string) => `${searchRoute}/${query}`;
...
+
+const communityBaseRoute = "/community";
+export const communityRoute = `${communityBaseRoute}/:communityId/:communitySlug?`;
+export const routeToCommunity = (id: number, slug: string) =>
+  `${communityRoute}/${id}/${slug}`;
+
+const communityPlacesBaseRoute = `${communityRoute}/places`;
+export const communityPlacesRoute = `${communityPlacesBaseRoute}/:communityId/:communitySlug?`;
+export const routeToCommunityPlaces = (id: number, slug: string) =>
+  `${communityPlacesRoute}/${id}/${slug}`;
```
```suggestion
  `${communityPlacesBaseRoute}/${id}/${slug}`;
```
Otherwise redirect is broken. Same for all the dynamic community related route creators here

darrenvong (changes requested):

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
darrenvong: But if the API is already there, shouldn't we just let the title be set? Or is it still subject to change?

darrenvong (commented):
There's a few merge conflicts... hopefully not too nasty

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
lucaslcode: I didn't want to put any more in this PR - this dialog is just the minimum for me to see if the discussions actually appeared since there was no mock data

darrenvong (approved):
Since you got part 3 going... I assume any hardcoded stuff you can address there and in subsequent PR parts to come

PR #698, opened 2021-02-25
fixed messages sending notification email

### LINKED ISSUE #691: clicking on a request message causes notification email to be sent
Subject: clicking on a request message causes notification email to be sent
Description:
whenever I visit the hosting request message tab and click on the message to view it, the system sends me an email saying I have a request. (Request to jesse from Itsi)

Steps:
Clicked on the request message to view it

Results:
I shouldn't get an email notification every time I click on it...

Backend version: develop-2b68134b
Frontend version: develop-2b68134b
User Agent: Mozilla/5.0 (Android 11; Mobile; rv:86.0) Gecko/86.0 Firefox/86.0
Page: https://develop--frontend.preview.coucher.org/messages/hosting
User ID: 7

dnalhg:
Viewing host request messages no longer sends a notification email. 

Moved the send email function call from GetHostRequest to CreateHostRequest. The send email function requires the host request to have a to_user attribute which is only populated after the host request is added to the database so the host request object is fetched again after it is added to the database. Not sure if this is the cleanest method.

### DIFF
--- a/app/backend/src/couchers/servicers/requests.py
+++ b/app/backend/src/couchers/servicers/requests.py
@@ -115,13 +115,8 @@ class Requests(requests_pb2_grpc.RequestsServicer):
             host_request.status = HostRequestStatus.pending
             host_request.from_last_seen_message_id = message.id
             session.add(host_request)
-
-            host_request = (
-                session.query(HostRequest)
-                .filter(HostRequest.conversation_id == conversation.id)
-                .filter(or_(HostRequest.from_user_id == context.user_id, HostRequest.to_user_id == host.id))
-                .one_or_none()
-            )
+            session.flush()
+            
             send_host_request_email(host_request)
 
             return requests_pb2.CreateHostRequestRes(host_request_id=host_request.conversation_id)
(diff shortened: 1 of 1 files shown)


app/backend/src/couchers/servicers/requests.py
```diff
@@ -116,6 +116,14 @@ def CreateHostRequest(self, request, context):
             host_request.from_last_seen_message_id = message.id
             session.add(host_request)
 
+            host_request = (
+                session.query(HostRequest)
+                .filter(HostRequest.conversation_id == conversation.id)
+                .filter(or_(HostRequest.from_user_id == context.user_id, HostRequest.to_user_id == host.id))
+                .one_or_none()
+            )
```
aapeliv: You don't need to grab the host request from the database again, SQLAlchemy (SA) keeps track of it and you can use the one from above.

However, at this point, SA hasn't sent it to the database yet, so it's just an object in memory and so has no auto-generated ID column yet. You need to do a `session.flush()` so SA will flush it to the database returning back an ID (tracked by SA automatically). It's a good idea to flush before sending the email anyway, that way if for example a DB constraint is violated, the function will fail before the email gets sent out. (On failure the database transaction will `ROLLBACK` so we get that ACID goodness and no gibberish data is stored in the databse).

(The `with session_scope ...` auto-commits on exit, that's why there's few `session.flush()`/`session.commit()`s around in the code.)

[Sorry for the verbose comment :P.]

aapeliv (changes requested):

aapeliv (approved):
Looks good! Thanks :D

aapeliv:
Closes #691

PR #702, opened 2021-02-25
added backend api to get all user communities #696

### LINKED ISSUE #696: List a user's communities/groups/etc
There's currently no way to list the communities, groups, and pages that a user is a part of or owns.

This should definitely exist *somewhere*.

dnalhg:
Added API to get all of a users' communities (Nodes)


app/pb/communities.proto
```diff
@@ -103,6 +107,17 @@ message ListCommunitiesRes {
   string next_page_token = 2;
 }
 
+message ListUserCommunitiesReq {
+  uint32 page_size = 2;
+  string page_token = 3;
```
dnalhg, line 115: not sure if I need to keep the page size info in here since I'm not sure how this data will be displayed/fetched for the front end

app/backend/src/couchers/servicers/communities.py
```diff
@@ -245,3 +245,26 @@ def LeaveCommunity(self, request, context):
...
+        with session_scope() as session:
+            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
+            next_node_id = int(request.page_token) if request.page_token else 0
+            nodes = (
+                session.query(Node)
+                .filter(Node.id >= next_node_id)
+                .order_by(Node.id)
+                .limit(page_size + 1)
+                .all()
+            )
```
dnalhg: should I fetch more than just the communities. Should I also return the groups, pages etc or should these be separate APIs?

aapeliv: It would be good to build one API call for each: communities, groups, and pages (which needs a filter of places/guide or both)

app/backend/src/couchers/servicers/communities.py
```diff
@@ -245,3 +245,26 @@ def LeaveCommunity(self, request, context):
...
+
+            # Filter for communities containing user of interest
+            relevant_communities = []
+            for node in nodes:
+                if node.official_cluster.members.filter(User.id == context.user_id).one_or_none() is not None:
+                    relevant_communities.append(node)
```
aapeliv: The filtering here should be done in the database. You can do a join across the association table:

```py3
            nodes = (
                session.query(Node)
                .join(ClusterSubscription, ClusterSubscription.cluster_id == Cluster.id)
                .filter(ClusterSubscription.user_id == context.user_id)
                .filter(Cluster.is_official_cluster)
                .filter(Node.id >= next_node_id)
                .order_by(Node.id)
                .limit(page_size + 1)
                .all()
            )
```

This should get you all the nodes that the user has a membership to.

(Also, if you get `page_size+1` nodes and then filter, you'll most likely not get enough communities, and often you'll get empty ones.)

app/backend/src/tests/test_communities.py
```diff
@@ -377,6 +377,24 @@ def test_ListCommunities(testing_communities):
...
+            c1r1c1_id = get_community_id(session, "Country 1, Region 1, City 1")
+            c1r1c2_id = get_community_id(session, "Country 1, Region 1, City 2")
+            c1r2_id = get_community_id(session, "Country 1, Region 2")
+            c1r2c1_id = get_community_id(session, "Country 1, Region 2, City 1")
+
+        with communities_session(token1) as api:
+            res = api.ListUserCommunities(
+                communities_pb2.ListCommunitiesReq()
+            )
+            assert [c.community_id for c in res.communities] == [w_id, c1_id, c1r1_id, c1r1c1_id, c1r1c2_id, c1r2_id, c1r2c1_id]
```
aapeliv, line 418: I don't think user1 should be in `c1r1c2`? The user is not explicitly a member and their location does not overlap the community area.

app/backend/src/couchers/servicers/communities.py
aapeliv: The `members` relationship also defines a backref (`cluster_memberships`):

https://github.com/Couchers-org/couchers/blob/08ca715f3c2912f5508f62cf3e53f1bf5890d562/app/backend/src/couchers/models.py#L791-L799

Another way would be to use this somehow. Though then you'd need to join on `Node`s, and I'm not sure what SA gymnastics are required for this!

app/pb/communities.proto
aapeliv: Yes, the frontend will do pagination, either through explicit pages or infinite scroll automagically.

aapeliv (changes requested):
See comments inline.

One other thing is that we probably actually want to get communities for arbitrary users.

You could implement this by adding a `uint32 user_id = ?;` to the `Req`, and then filtering based on that user ID. You can use `0` to denote "requestor".

app/backend/src/tests/test_communities.py
dnalhg: this test was using user2 so I think it should be fine

app/backend/src/couchers/servicers/communities.py
```diff
@@ -245,3 +245,83 @@ def LeaveCommunity(self, request, context):
...
+
+    def ListUserGuides(self, request, context):
+        with session_scope() as session:
+            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
+            next_page_id = int(request.page_token) if request.page_token else 0
+            user_id = context.user_id if request.user_id == 0 else request.user_id
+            guides = (
+                session.query(Page)
+                .filter(Page.owner_user_id == user_id)
+                .filter(Page.type == PageType.guide)
```
dnalhg: Places and Guides untested. Original ListPlaces and ListGuides were untested with the comment:
TODO: requires transferring of content

Not too sure what that means

app/backend/src/tests/test_communities.py
aapeliv: Ah, sorry, my bad.

Please change the `user1_id` and `token1` to be 2s instead!

app/pb/communities.proto
```diff
@@ -64,6 +64,22 @@ service Communities {
...
+    // Gets a list of groups a member belongs to
+  }
+
+  rpc ListUserPlaces(ListUserPlacesReq) returns (ListUserPlacesRes) {
+    // Gets a list of groups a member belongs to
+  }
+
+  rpc ListUserGuides(ListUserGuidesReq) returns (ListUserGuidesRes) {
+    // Gets a list of groups a member belongs to
+  }
```
aapeliv: Please move groups to `groups.proto`, and places/guides to `pages.proto`

app/backend/src/couchers/servicers/communities.py
aapeliv: Places and guides are currently created by users, then transferred to communities/groups. So communities couldn't have any places/groups before transfer was implemented.

Transfer has been implemented now, but tests haven't been written 🤔

app/backend/src/couchers/servicers/communities.py
```diff
@@ -245,3 +245,83 @@ def LeaveCommunity(self, request, context):
             session.query(ClusterSubscription).filter(ClusterSubscription.user_id == context.user_id).delete()
 
             return empty_pb2.Empty()
+    
+    def ListUserCommunities(self, request, context):
+        with session_scope() as session:
+            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
+            next_node_id = int(request.page_token) if request.page_token else 0
+            user_id = context.user_id if request.user_id == 0 else request.user_id
```
aapeliv: You could simplify this to `request.user_id or context.user_id`; if `request.user_id` is 0, it'll be falsey in a boolean context, so it'll fall through to using `context.user_id`

aapeliv (changes requested):
See comments. This is starting to look very good!

I know the communities/groups/pages stuff is a bit confusing and it's not well documented here in code, sorry :S

app/pb/communities.proto
dnalhg: should I also move the methods to services/groups.py and services/pages.py?

aapeliv: Yeah, if you move the RPC def to those protos, you have to move servicers as well

app/pb/pages.proto
```diff
@@ -27,6 +27,14 @@ service Pages {
   rpc TransferPage(TransferPageReq) returns (Page) {
     // Transfer a page you own to either a group or a community
   }
+
+  rpc ListUserPlaces(ListUserPlacesReq) returns (ListUserPlacesRes) {
+    // Gets a list of groups a member belongs to
```
aapeliv: Wrong comment

aapeliv (changes requested):
Sorry, I didn't notice you pushed more commits!

This is looking very good, thanks! Please add one test at least where you explicitly set the user id to something that's not the current user to double check it's working! Also a small nitpick with one comment.

There's been a lot of emails from this repo lately :). Please re-request a review next time so the status changes and I'll notice it easier.

aapeliv (commented):
Sorry, one more small thing: could you add a comment in the protos that the `user_id` being zero gives the user's own places/guides/communities.

Other than that, this looks very good and is good to merge!

aapeliv (approved):
Awesome! Smash that green button ;)

PR #703, opened 2021-02-25
Frontend/feature/image upload

MichaelGuldborg:
I've added functionality for updating the user profile picture in the UserForm

I've added a component that handles taking an image file input (ImageInput)
And added a component for showing a circular avatar image (AvatarInput)
I've added a service function to upload images to the backend

**Waiting** for #704 to be merged


app/frontend/src/components/AvatarInput.tsx
```diff
@@ -0,0 +1,29 @@
...
+import React from "react";
+
+import {ImageInput, ImageInputProps} from "./ImageInput";
+
+export interface AvatarInputProps extends ImageInputProps {
+    src?: string;
+    username?: string;
+}
+
+const AvatarInput: React.FC<AvatarInputProps> = ({id, name, username, src, ...rest}) => {
```
lucaslcode:
```suggestion
function AvatarInput({id, name, username, src, ...rest}: AvatarImageProps) {
```
style nit

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -0,0 +1,29 @@
...
+    return (
+        <ImageInput id={id} name={name} {...rest}>
+            <IconButton component="span">
+                <Avatar
+                    src={src}
+                    imgProps={{style: {objectFit: 'contain'}}}
+                    style={{height: 120, width: 120, objectFit: 'contain'}}
```
lucaslcode: We tend to use jss (`const useStyle = makeStyles((theme) => ({...}))`) rather than inline styles (unless there is a reason you wanted inline here?)

Also I think it would be good if the parent could set the size (perhaps by this component accepting a `className` prop) rather than it being fixed.

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+
+export interface ImageInputProps {
+    id?: string;
+    name?: string;
+    onChange: (file: File) => void;
+}
+
+export const ImageInput: React.FC<ImageInputProps> = ({id, name, children, onChange}) => {
```
lucaslcode:
```suggestion
export default function ImageInput({id, name, children, onChange}: ImageInputProps) {
```

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+    const inputId = id + '-file-input';
+    return (
+        <React.Fragment>
```
lucaslcode:
```suggestion
        <>
```
shorthand

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+                        return;
+                    }
+                    const file = event.target.files[0];
+                    onChange && onChange(file);
+                }}
+            />
+            <label htmlFor={inputId}>
+                {children}
+            </label>
+        </React.Fragment>
```
lucaslcode:
```suggestion
        </>
```

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+    onChange: (file: File) => void;
+}
+
+export const ImageInput: React.FC<ImageInputProps> = ({id, name, children, onChange}) => {
+    const inputId = id + '-file-input';
+    return (
+        <React.Fragment>
+            <input
+                accept="image/*"
+                style={{display: 'none'}}
```
lucaslcode: use `className`
(Hmm if this is the only style in the file, maybe you don't have to).

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,34 @@
...
+                    onChange && onChange(file);
+                }}
+            />
+            <label htmlFor={inputId}>
+                {children}
+            </label>
+        </React.Fragment>
+    )
+}
+export default ImageInput;
```
lucaslcode: Newline. There are some other formatting things too, run `yarn format`

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -87,12 +93,34 @@ export default function EditProfileForm() {
...
+                      id={'profile-picture'}
+                      name={'image'}
+                      src={value}
+                      onChange={async (file) => {
+                          // TODO move media endpoint to config
+                        const response = await service.api.uploadFile(file);
+                        const thumbnailUrl = 'https://dev-user-media.coucher.org/media/img/avatar/' + response.key + '.jpg';
+                        setValue('avatarUrl',thumbnailUrl)
+                      }}
+                  />
```
lucaslcode: I think you could move this all into the AvatarInput component. Not sure if others agree or not.

That way you don't have to specify onChange and there would be one place to change the upload logic.

lucaslcode: Additionally, if you use a [react-query mutation](https://react-query.tanstack.com/guides/mutations) you'll get free loading state

lucaslcode (dismissed):
Nice one

Some small style nitpicks and an optional idea to move the logic into the component. IT would also be great to have the new components in storybook!

lucaslcode:
> Aapeli agrees that it would be better that the url is part of the response from the backend rather than the way I've done it currently

What do you mean by this?

lucaslcode:
Also the upload is not saved to the profile currently, it needs to be added to the UpdateProfileReq request in `service/user.ts`

app/frontend/src/components/ImageInput.tsx
andreaowu: In order to make the code organized, we should just keep it consistent and just use `className`.

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -77,6 +80,9 @@ export default function EditProfileForm() {
     updateUserProfile({ profileData: data, setMutationError: setErrorMessage });
   });
 
+
+
+
```
andreaowu: nit: delete extra lines

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -87,12 +93,34 @@ export default function EditProfileForm() {
       {user ? (
         <>
           <form onSubmit={onSubmit}>
+
+
```
andreaowu: nit: delete extra lines

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -87,12 +93,34 @@ export default function EditProfileForm() {
...
+            <Controller
+                name={'avatarUrl'}
+                control={control}
+                render={({value}) => {
+                  return <AvatarInput
+                      id={'profile-picture'}
+                      name={'image'}
+                      src={value}
+                      onChange={async (file) => {
+                          // TODO move media endpoint to config
```
andreaowu: nit: indentation

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -87,12 +93,34 @@ export default function EditProfileForm() {
...
+                        const response = await service.api.uploadFile(file);
+                        const thumbnailUrl = 'https://dev-user-media.coucher.org/media/img/avatar/' + response.key + '.jpg';
+                        setValue('avatarUrl',thumbnailUrl)
+                      }}
+                  />
+                }}
+            />
+
+
```
andreaowu: nit: delete extra lines

app/frontend/src/features/profile/ProfileForm.tsx
andreaowu: +1 otherwise every time we have to upload, this code needs to be copied again (basically what Lucas said)

app/frontend/src/components/AvatarInput.tsx
MichaelGuldborg: I think that's a good idea, I'll change it

app/frontend/src/components/ImageInput.tsx
MichaelGuldborg: I'll change it and remember to use className in the future aswell

app/frontend/src/features/profile/ProfileForm.tsx
MichaelGuldborg: I totally agree, I'll move this logic into the ImageInput component.
I'm not sure how to avoid having the onChange since I need to use the setValue function to set the thumbnail url on the form, right?

MichaelGuldborg:
> Also the upload is not saved to the profile currently, it needs to be added to the UpdateProfileReq request in service/user.ts

I've added the avatarKey and avatarUrl to the request values but it still does not seem to work

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -87,12 +95,32 @@ export default function EditProfileForm() {
...
+              render={({ value }) => {
+                return (
+                  <AvatarInput
+                    className={classes.avatar}
+                    id={"profile-picture"}
+                    name={"image"}
+                    src={value}
+                    onChange={async (values) => {
+                      setValue("avatarUrl", values.thumbnail_url);
+                      setValue("avatarKey", values.key);
```
lucaslcode: Ok after thinking about it, I think the way to do this is:
- only have `avatarKey` here in the form
- handle the `avatarUrl` and displaying it within the `AvatarInput` with local state
- the [`Controller`](https://react-hook-form.com/api/#Controller) render function also has `onChange(newValue)` as a parameter, so pass that to `AvatarInput`
- then within `AvatarInput` in the "real" `onChange`, you can set the url in the local component state, then call from the props `onChange(avatarKey)`

Let me know if that makes no sense :P

app/frontend/src/service/user.ts
```diff
@@ -49,7 +49,8 @@ export type UpdateUserProfileData = Pick<
   | "countriesVisited"
   | "countriesLived"
   | "additionalInformation"
->;
+  | "avatarKey"
```
lucaslcode, line 51: You need to apply the new data to the gRPC request message in this file in `updateProfile` function

app/frontend/src/features/profile/ProfileForm.tsx
MichaelGuldborg: It makes sense,
I've fixed it now 👌

MichaelGuldborg:
@aapeliv 
The updateUser request gives an error when it includes the avatarKey
And some of the uploaded images gives "Internal Server Error" when I request them after upload

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -24,16 +26,23 @@ export function AvatarInput({
   src,
   username,
   className,
+  onChange,
   ...rest
 }: AvatarInputProps) {
   const classes = useStyles();
+  const [imageUrl, setImageUrl] = useState(src);
+  const handleChange = (values: ImageInputValues) => {
+    const randomInt = Math.floor(Math.random() * 100); // force reload onChange
```
lucaslcode: Did you have problems when not doing this? I would have thought every upload has a different key and therefore a different url

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -24,16 +26,23 @@ export function AvatarInput({
...
+    setImageUrl(values.thumbnail_url + "?rand=" + randomInt);
+    onChange && onChange(values.key);
```
lucaslcode:
```suggestion
    onChange?.(values.key);
```
You don't have to change this but I just learnt this recently and I think it's cool :D

app/frontend/src/features/profile/ProfileForm.tsx
```diff
@@ -67,7 +68,7 @@ export default function EditProfileForm() {
       lat: user?.lat ?? undefined,
       lng: user?.lng ?? undefined,
       radius: user?.radius ?? undefined,
-      // avatarUrl: user?.avatarUrl ?? undefined,
+      avatarKey: undefined,
```
lucaslcode: Does this need to be `avatarKey: user.avatarKey` - otherwise maybe the avatar will be unset if the user is changing things that aren't the avatar?

lucaslcode: But that might mess with the thumbnail...

lucaslcode (commented):
Looks good, let's wait for backend

app/frontend/src/features/profile/ProfileForm.tsx
lucaslcode: Maybe the solution would be an `initialImage` prop or something

aapeliv:
>  The updateUser request gives an error when it includes the avatarKey

Fixed the typo

>  And some of the uploaded images gives "Internal Server Error" when I request them after upload

Fixed that one too, in #709.

Sorry, the backend has been incredibly buggy...

app/frontend/src/components/AvatarInput.tsx
MichaelGuldborg: That's pretty cool actually 🤔

app/frontend/src/features/profile/ProfileForm.tsx
MichaelGuldborg: That's a good point.
It seems that the backend needs the avatarkey to update the profile image. I assumed that if it was undefined or empty then the backend would just ignore it.

@aapeliv

app/frontend/src/components/AvatarInput.tsx
MichaelGuldborg: I did have some issues when uploading the same image twice, which is why I added it.
But I'll check if it can be omitted

app/frontend/src/service/user.ts
```diff
@@ -129,6 +130,7 @@ export async function updateProfile(
 ): Promise<Empty> {
   const req = new UpdateProfileReq();
 
+  const avatarKey = new wrappers.StringValue().setValue(profile.avatarKey);
```
lucaslcode:
```suggestion
  const avatarKey = profile.avatarKey ? new wrappers.StringValue().setValue(profile.avatarKey) : undefined;
```

This might solve this - the purpose of the wrapper is to differentiate between unsetting (ie. setting to nothing - wrapper is present but value is null) and just not making any changes (wrapper itself not present). This should do since you can't clear the image at this point.

lucaslcode:
Currently upload works, but changing the profile without changing the image causes an unknown backend error. I suspect there is no check for avatarKey existing but having a blank value @aapeliv .

app/frontend/src/service/user.ts
MichaelGuldborg: I've changed it to your suggestion

lucaslcode:
@MichaelGuldborg  It's disappeared from the profile edit page - did it get lost in the merge?

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+        name={name}
+        type="file"
+        onChange={async (event) => {
+          if (!event.target.files?.length) {
+            console.log("No files selected");
+            return;
+          }
+          const file = event.target.files[0];
+          const response = await mutation.mutateAsync(file);
+          onChange && onChange(response);
```
darrenvong:
```suggestion
          onChange(response);
```

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+export interface ImageInputProps {
+  id?: string;
+  name?: string;
+  onChange: (values: ImageInputValues) => void;
+  children: React.ReactNode;
+}
+
+export function ImageInput({ id, name, children, onChange }: ImageInputProps) {
+  const classes = useStyles();
+  const inputId = id + "-file-input";
```
darrenvong:
```suggestion
  const inputId = `${id}-file-input`;
```

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <>
+      <input
+        className={classes.input}
+        accept="image/*"
+        id={inputId}
+        name={name}
+        type="file"
+        onChange={async (event) => {
+          if (!event.target.files?.length) {
+            console.log("No files selected");
```
darrenvong:
```suggestion
```

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -122,6 +127,23 @@ export default function EditProfileForm() {
       {user ? (
         <>
           <form onSubmit={onSubmit}>
+            <Controller
+              name={"avatarKey"}
```
darrenvong:
```suggestion
              name="avatarKey"
```

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -122,6 +127,23 @@ export default function EditProfileForm() {
...
+              control={control}
+              render={({ onChange }) => {
+                return (
+                  <AvatarInput
+                    className={classes.avatar}
+                    id={"profile-picture"}
+                    name={"avatarKey"}
```
darrenvong:
```suggestion
                    id="profile-picture"
                    name="avatarKey"
```

app/frontend/src/components/AvatarInput.tsx
darrenvong: This is unnecessary since `onChange` is guaranteed by the interface to always exist?

app/frontend/src/components/AvatarInput.tsx
darrenvong: So is it safe to actually omit this? Seems a bit hacky to me if this can be resolved by the backend instead.

app/frontend/src/service/api.ts
```diff
@@ -48,3 +51,23 @@ export async function ping() {
 
   return response.toObject();
 }
+
+export async function uploadFile(file: File): Promise<ImageInputValues> {
```
darrenvong, line 62: Not sure if this manual work is necessary 🤔  We can always get the URL for this in a `useEffect` and populate the form's action that way?

I feel like it'd be simpler to use HTML form to handle this as it already deals with form data properly out of the box.

app/frontend/src/service/api.ts
```diff
@@ -1,11 +1,14 @@
...
+import { ImageInputValues } from "../components/ImageInput";
 import {
   CancelFriendRequestReq,
   PingReq,
   RespondFriendRequestReq,
   SendFriendRequestReq,
 } from "pb/api_pb";
-import client from "service/client";
+
+import client from "./client";
```
darrenvong: Build is failing on the import order of these.

darrenvong (changes requested):
Also out of curiosity once this is merged, how do you use this locally?

darrenvong:
Also this is crashing with `Error: There is already a source with this ID`, which I suspect your Map fix will fix this?

app/frontend/src/service/api.ts
lucaslcode: I think it's better this way, since otherwise you would be making a media upload request every time you go to edit anything on your profile, plus an html form `submit()` doesn't return a promise, so it will end up more complicated.

lucaslcode:
image

To test locally you need to run media server docker container, but it's only in the production docker-compose file atm

app/frontend/src/components/AvatarInput.tsx
darrenvong:
```suggestion
    onChange(values.key);
```

app/frontend/src/service/api.ts
darrenvong: Does this need to be part of the main form though if this is a standalone thing? It shouldn't be sent every time if it's in its own form. What would you need the promise for in that case?

app/frontend/src/components/ImageInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+        className={classes.input}
+        accept="image/*"
+        id={inputId}
+        name={name}
+        type="file"
+        onChange={async (event) => {
+          if (!event.target.files?.length) return;
+          const file = event.target.files[0];
+          const response = await mutation.mutateAsync(file);
+          onChange(response);
```
darrenvong: Not sure auto submitting like this is good UX - what if the user selects the wrong image and want to correct it?

lucaslcode: They can choose a new image - it does auto-upload, but the upload will only be placed in the resource (in the case, profile) when the form as a whole is submitted.

lucaslcode: What I mean is, this auto-submits and auto-receives a key, but the key isn't put into the profile automatically, just into the form state.

lucaslcode: If you're content with that and you approve, feel free to merge this pr

darrenvong: Ahh I see, sounds like more of a problem for Aapeli's wallet on S3 bucket cost than actual technical problem here for me here 😂

darrenvong: Although storage (half) joke aside, it's still a bit shitty for people on mobile data for this to always upload?

lucaslcode:
image

lucaslcode:
How's this, will do some tests if it's ok

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+      reader.readAsDataURL(file);
+    });
+    setImageUrl(base64);
+    setFile(file);
+  };
+
+  const handleConfirm = async () => {
+    const response = await mutation.mutateAsync();
+    //const randomInt = Math.floor(Math.random() * 100); // force reload onChange
+    //setImageUrl(response.thumbnail_url + "?rand=" + randomInt);
```
darrenvong:
```suggestion
```

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -1,54 +1,166 @@
...
+      validate: () => !isConfirming || UPLOAD_PENDING_ERROR,
+    },
+  });
+
+  const handleChange = async (event: React.ChangeEvent<HTMLInputElement>) => {
+    if (!event.target.files?.length) return;
+    const file = event.target.files[0];
+    const base64 = await new Promise<string>((resolve, reject) => {
+      const reader = new FileReader();
+      reader.onload = () => resolve(reader!.result as string);
```
darrenvong:
```suggestion
      reader.onload = () => resolve(reader.result as string);
```

app/frontend/src/components/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const CANCEL_UPLOAD = "Cancel upload";
+export const CONFIRM_UPLOAD = "Confirm upload";
+export const getAvatarLabel = (name: string) => `${name} avatar`;
```
darrenvong:
```suggestion
export const getAvatarLabel = (name: string) => `${name}'s profile picture`;
```

app/frontend/src/components/AvatarInput.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { useForm } from "react-hook-form";
+
+import AvatarInput from "./AvatarInput";
+import Button from "./Button";
+
+export default {
+  component: AvatarInput,
+  title: "Components/Compound/AvatarInput",
+} as Meta;
+
```
darrenvong, line 13: This is broken when pressing confirm. I think `uploadFile` from service in this will need mocking.

darrenvong (changes requested):
Pretty minor things left in the comment now, but this looks fine to me. Go ahead and add the tests if you are planning to 👍

app/frontend/src/components/AvatarInput.tsx
lucaslcode: Typescript doesn't know it can't be undefined, since it starts as undefined and the event can be triggered from anything in theory

darrenvong: `reader` won't be undefined though as you instantiated on the line above? I checked locally and it didn't give me an error without it.

lucaslcode: Oh quite right

app/frontend/src/components/AvatarInput.stories.tsx
lucaslcode: I used `ChangeEmail.stories.tsx` as a mock example but it isn't mocking... any idea why?

darrenvong: Maybe it's because `AvatarInput` is importing `service` instead of `service/index`?

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+      MOCK_INITIAL_SRC
+    );
+
+    userEvent.upload(screen.getByLabelText(SELECT_AN_IMAGE), MOCK_FILE);
+
+    await waitFor(() => {
+      expect(screen.getByLabelText(CONFIRM_UPLOAD)).toBeVisible();
+    });
+
+    userEvent.click(screen.getByLabelText(CONFIRM_UPLOAD));
```
darrenvong: Is this not meant to click `CANCEL_UPLOAD` instead based on the test description?

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    await waitFor(() => {
+      expect(uploadFileMock).toHaveBeenCalled();
+    });
+
+    userEvent.click(screen.getByRole("button", { name: SUBMIT }));
+
+    await waitFor(() => {
+      expect(submitForm).toHaveBeenCalledWith({ avatarInput: MOCK_KEY });
+    });
+  });
```
darrenvong, line 85: Check the image src here has also been updated to the new src?

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    await waitFor(() => {
+      expect(uploadFileMock).toHaveBeenCalledTimes(2);
+    });
+
+    expect(screen.getByAltText(getAvatarLabel(NAME))).not.toHaveProperty(
+      "src",
+      MOCK_INITIAL_SRC
+    );
+
```
darrenvong:
```suggestion
    expect(screen.getByAltText(getAvatarLabel(NAME))).toHaveProperty(
      "src",
      "thumb0.jpg"
    );
```
Checking it has the src of the first upload implies it's not `MOCK_INITIAL_SRC`

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+      expect(submitForm).toHaveBeenCalledWith({ avatarInput: MOCK_KEY });
+    });
+  });
+
+  it("doesn't submit without confirming/cancelling", async () => {
+    userEvent.upload(screen.getByLabelText(SELECT_AN_IMAGE), MOCK_FILE);
+
+    await waitFor(() => {
+      expect(screen.getByLabelText(CONFIRM_UPLOAD)).toBeVisible();
+    });
```
darrenvong:
```suggestion
      expect(await screen.findByLabelText(CONFIRM_UPLOAD)).toBeVisible();
```
For single assertion like this, you can use `findBy*` instead. It's shorthand equivalent to what you got there.

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    userEvent.click(screen.getByRole("button", { name: SUBMIT }));
+
+    await waitFor(() => {
+      expect(screen.getByText(UPLOAD_PENDING_ERROR)).toBeVisible();
+    });
```
darrenvong: Same comment as above

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+    await waitFor(() => {
+      expect(screen.getByLabelText(CONFIRM_UPLOAD)).toBeVisible();
+      expect(screen.getByLabelText(CANCEL_UPLOAD)).toBeVisible();
+    });
+
+    userEvent.click(screen.getByLabelText(CONFIRM_UPLOAD));
+
+    await waitFor(() => {
+      expect(uploadFileMock).toHaveBeenCalled();
```
darrenvong:
```suggestion
      expect(uploadFileMock).toHaveBeenCalledTimes(1);
```
So we know it's not going rogue and being called unnecessarily somehow?

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+    );
+  });
+
+  it("uploads upon confirmation and submits key", async () => {
+    userEvent.upload(screen.getByLabelText(SELECT_AN_IMAGE), MOCK_FILE);
+
+    await waitFor(() => {
+      expect(screen.getByLabelText(CONFIRM_UPLOAD)).toBeVisible();
+      expect(screen.getByLabelText(CANCEL_UPLOAD)).toBeVisible();
+    });
```
darrenvong:
```suggestion
    expect(await screen.findByLabelText(CONFIRM_UPLOAD)).toBeVisible();
```
Since both of these buttons render together, it's ok to only wait for one of them, since that'd be long enough for the other to appear too.

darrenvong (changes requested):
Some solid testing effort there 💪

darrenvong:
Also tests for negative cases? i.e. what happens if an invalid file is given, and when the file reader throws an error

lucaslcode:
@darrenvong as far as I can tell, there's no way to check for valid images. However the server does and I added error handling.

app/frontend/src/components/AvatarInput.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+    expect(submitForm).not.toHaveBeenCalled();
+  });
+
+  it.skip("displays an error for an invalid file", async () => {
+    userEvent.upload(
+      screen.getByLabelText(SELECT_AN_IMAGE),
+      new File([new Blob(undefined)], "")
+    );
+    expect(await screen.findByText(INVALID_FILE)).toBeVisible();
+  });
```
lucaslcode: No matter what I tried, couldn't make an invalid file to test this....

darrenvong: I got it working by making `readAsDataURL` emit an `error` event on all `FileReader` instances for this test specifically, which I guess is what happens when there's an error as these seem to use the event callback pattern rather than throwing an error directly, so I think it's fine to do? More interested in hitting the error handling path rather than how it's achieved here...

```suggestion
  it("displays an error for an invalid file", async () => {
    jest.spyOn(FileReader.prototype, "readAsDataURL").mockImplementation(() => {
      FileReader.prototype.dispatchEvent(new Event("error"));
    });
    userEvent.upload(
      screen.getByLabelText(SELECT_AN_IMAGE),
      new File([new Blob(undefined)], "")
    );
    expect(await screen.findByText(INVALID_FILE)).toBeVisible();
  });
```

darrenvong:
If the server will throw an error, couldn't we simulate this since those are already mocked? What happens if `handleConfirm` throws from the `mutateAsync` call?

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -0,0 +1,173 @@
...
+      });
+      setImageUrl(base64);
+      setFile(file);
+    } catch {
+      setReaderError(INVALID_FILE);
+    }
+  };
+
+  const handleConfirm = async () => {
+    const response = await mutation.mutateAsync();
```
darrenvong: What happens if this throws? I think the whole UI will probably crash here...

lucaslcode: Interestingly not, but I would have expected it to! Wrapped it in try just in case.

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -105,11 +105,15 @@ export function AvatarInput({
...
-    confirmedUpload.current = response;
-    setFile(null);
+    try {
+      const response = await mutation.mutateAsync();
+      field.onChange(response.key);
+      setImageUrl(response.thumbnail_url);
+      confirmedUpload.current = response;
+      setFile(null);
+    } catch {
+      //no need to do anything here, error goes to mutation.error
```
darrenvong: Would it be easier to use `mutate` (the non async awaitable version) in that case? 🤔 
Then we wouldn't need this catch at all as it'd still be handled by React Query's mutation error?

darrenvong (approved):

PR #705, opened 2021-02-25
Remove moment library from package.json

### LINKED ISSUE #365: replace moment library
As it is advised not to use moment anymore we need to find out how to replace its uses in our code.

MichaelGuldborg:
I replaced all moment imports with vanilla js functions
#365


lucaslcode:
There's also `timeago.ts`

MichaelGuldborg:
Aahh I missed timeago.ts

app/frontend/src/utils/timeAgo.ts
```diff
@@ -1,9 +1,42 @@
...
+const yearMillis = 31557600000;
+const twoYearMillis = 31557600000 * 2;
+
+export function timeAgo(input: Date | string) {
+  if (input === undefined) return "";
+  const date = new Date(input);
+  const diffMillis = Date.now() - date.getTime();
 
-export function timeAgo(date: Date) {
   //the backend fuzzes times to 15 minute intervals
```
darrenvong: Would be good to have some unit tests for these since we are writing our own logic/ditching `moment`

andreaowu: +1

MichaelGuldborg: I've added a unit test to the timeago function

lucaslcode:
There is a build error from the merge - I recently added another `fuzzed` parameter to `timeAgo` which defaults to false, which decides whether to do the < 15 minutes thing or not. Sorry!

app/frontend/src/utils/timeAgo.ts
```diff
@@ -1,15 +1,15 @@
...
-const dayMillis = 86400000;
-const twoDayMillis = 172800000;
-const weekMillis = 604800000;
-const twoWeekMillis = 1209600000;
-const monthMillis = weekMillis * 4;
-const twoMonthMillis = weekMillis * 8;
-const yearMillis = 31557600000;
-const twoYearMillis = 31557600000 * 2;
+export const minuteMillis = 60000;
+export const quaterHourMillis = 900000;
```
darrenvong:
```suggestion
export const quarterHourMillis = 900000;
```
Nitpick: spelling mistake

app/frontend/src/utils/timeAgo.test.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  dayMillis,
+  hourMillis,
+  minuteMillis,
+  monthMillis,
+  quaterHourMillis,
```
darrenvong:
```suggestion
  quarterHourMillis,
```

app/frontend/src/utils/timeAgo.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  twoMonthMillis,
+  twoWeekMillis,
+  twoYearMillis,
+  weekMillis,
+  yearMillis,
+} from "./timeAgo";
+
+const timeAgoMap = {
+  [minuteMillis]: "< 15 minutes ago",
+  [quaterHourMillis]: "15 minutes ago",
```
darrenvong:
```suggestion
  [quarterHourMillis]: "15 minutes ago",
```

app/frontend/src/utils/timeAgo.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  [yearMillis]: "1 year ago",
+  [twoYearMillis]: "2 years ago",
+};
+
+test("timeAgo function", () => {
+  const now = Date.now();
+
+  Object.keys(timeAgoMap).forEach((key: string) => {
+    const millis = parseInt(key);
+    const value = timeAgoMap[millis];
```
darrenvong:
```suggestion
    const expectedValue = timeAgoMap[millis];
```

app/frontend/src/utils/timeAgo.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    const date = new Date(now - millis);
+    const s = timeAgo(date);
+    expect(s).toBe(value);
```
darrenvong:
```suggestion
    const timeString = timeAgo(date);
    expect(timeString).toBe(expectedValue);
```
Probably both nitpicks as the tests are small enough you can guess the meaning, but I like to be explicit.

darrenvong (commented):
Normally I wouldn't block an approval on these but since you have a build error, may as well fix that together ;)

MichaelGuldborg:
Valid point @darrenvong
I've updated the PR

darrenvong (approved):

PR #706, opened 2021-02-25
Test: regenerating migrations

aapeliv:
I'm having a look at what it takes to regenerate migrations, as well as what the advantages are (in particular if the tests will run faster).

I don't want this to be merged. I think the strategy should be as follows:

1. When we migrate alpha to beta, we take it down, and just upgrade the alpha container to beta. This will run all the automatic + manual migrations on the alpha data.
2. Do a `pg_dump` of the data only
3. Regenerate all migrations
4. Do a `pg_restore` onto the new setup

We should check the outcome of 1 and 3 are roughly equivalent as well, and this can be done as a dry-run before the proper migration. We can take down the app for a day or two to get this right.


aapeliv:
I've had a look at the schemas for both, there's a few small discrepancies, but other than that all seems well:

* `backgroundjobstate` enum still has state `reserved`
* `hostrequesteventtype` is leftover
* Some constraint names were typed in wrong when done manually
* Host requests weirdness
* Fields have been reordered
* Many sequences are created `AS INTEGER` (as opposed to the default big integer)
* User session expiry is not correct

aapeliv:
It actually took 15 sec longer to run tests than on `develop`!

PR #709, opened 2021-02-26
Media scaling bugfix

aapeliv:
Fixes calculation error when width < height causing invalid crops


kalvdans (approved):
I would generate the images in the tests. But super that you added tests!

PR #710, opened 2021-02-26
Add TOS & Location to CompleteSignupForm

Tay0108:
Frontend for https://github.com/Couchers-org/couchers/issues/601


aapeliv:
Jail should also support these, in addition to signup. E.g. if an alpha user has not set a location and logs in, they should be jailed and required to update it on first log in to beta. Similarly if we update the TOS, everyone needs to be jailed before they can continue to the app, etc.

app/frontend/src/features/auth/jail/Jail.tsx
```diff
@@ -28,6 +26,7 @@ export default function Jail() {
   const isAuthenticated = authState.authenticated;
 
   const [loading, setLoading] = useState(false);
+  // eslint-disable-next-line
```
lucaslcode: What happened here?

app/frontend/src/features/auth/jail/Jail.tsx
```diff
@@ -40,6 +39,7 @@ export default function Jail() {
     })();
   }, [authActions]);
 
+  // eslint-disable-next-line
```
lucaslcode: Also here?

app/frontend/src/features/auth/jail/Jail.tsx
```diff
@@ -57,15 +57,6 @@ export default function Jail() {
...
       </Backdrop>
-      {jailInfo?.hasNotAcceptedTos && (
-        <TOSSection updateJailed={updateJailed} className={classes.section} />
-      )}
-      {jailInfo?.hasNotAddedLocation && (
-        <LocationSection
-          updateJailed={updateJailed}
-          className={classes.section}
-        />
-      )}
```
lucaslcode: These can stay in for backwards-compatibility/later use

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -30,17 +36,23 @@ type SignupInputs = {
...
 };
 
 const useStyles = makeStyles((theme) => ({
   genderRadio: {
     display: "flex",
     flexDirection: "row",
   },
-  genderRadioButton: {},
+  locationMap: {
+    width: "100%",
```
lucaslcode, line 52: `EditUserLocationMap` has a `grow` prop, can that be used instead? (possibly not)

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -165,16 +191,33 @@ export default function CompleteSignup() {
...
-          <TextField
-            id="location"
-            variant="standard"
+          <Controller
             name="location"
-            fullWidth
+            control={control}
             inputRef={register({
               required: "Enter your location",
             })}
```
lucaslcode:
```suggestion
            inputRef={register}
```
That `required` bit isn't doing anything since there's no actual field and you're handling the check manually

lucaslcode (changes requested):
👍🏻

app/frontend/src/features/auth/jail/Jail.tsx
Tay0108: Disabled warnings because CI would have problems with them and we probably want to reuse this component in the future. But since Aapeli said, that I should left TOS and Map in jail I'll remove it.

darrenvong: What warnings do you get? Seems like a pretty bog standard `useState` to me so shouldn't complain?

lucaslcode: Probably "unused" warning

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
Tay0108: No, it does not work:
image

lucaslcode (approved):
I changed the base branch so this will merge into the backend PR.

PR #712, opened 2021-02-26
about me page

andreaowu:
This PR:
- creates About page
- makes ProfilePage the center for everything profile-related (card, tabbar, pages to show) and updates appropriate routes
- deletes UserPage and routes
- fixes ProfilePage for mobile web
- makes container wider

`/profile` goes to user's profile
`/profile/<username>` goes to `username`'s profile

Screen Shot 2021-02-26 at 5 18 27 PM


app/frontend/src/components/TabBar/TabBar.tsx
```diff
@@ -34,7 +39,7 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
     >
       {Object.entries(labels).map(([value, label]) => (
         <Tab
-          classes={{ root: classes.tabRoot }}
+          classes={{ root: short ? classes.tabRootShort : classes.tabRoot }}
```
darrenvong: I ran into this problem when I tried to re-use this, that the `minWidth` was making the tabs too wide and so you get horizontal scrolling even on desktop. I removed it altogether and let the tab content dictate what width it needs instead and it seems to work fine. It might be worth doing that instead to simplify the component interface a bit? Seems like MUI already applies `minWidth` according to different breakpoints.

app/frontend/src/AppRoutes.tsx
```diff
@@ -124,15 +123,12 @@ export default function AppRoutes() {
...
         <EditHostingPreferencePage />
       </AppRoute>
-      <AppRoute isPrivate path={profileRoute}>
+      <AppRoute isPrivate path={`${profileRoute}/:type?`}>
         <ProfilePage />
       </AppRoute>
       <AppRoute isPrivate path={`${messagesRoute}/:type?`}>
         <Messages />
       </AppRoute>
-      <AppRoute isPrivate path={`${userRoute}/:username`}>
```
darrenvong, line 133: We'll still need this route, otherwise we won't be able to visit other user's profile anymore with `/user/<username>`.

app/frontend/src/features/profile/profilePage/About.tsx
```diff
@@ -0,0 +1,63 @@
...
+  OCCUPATION,
+  OVERVIEW,
+  WHO,
+} from "../../constants";
+
+const useStyles = makeStyles((theme) => ({
+  card: {
+    padding: theme.spacing(2),
+    width: "65%",
+  },
```
darrenvong: These styles don't seem to be used? 🤔

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
```diff
@@ -20,15 +41,54 @@ const useStyles = makeStyles({
...
     },
   },
-});
+}));
+
+type DetailType = keyof typeof SECTION_LABELS;
 
 export default function ProfilePage() {
   const classes = useStyles();
   const { data: user } = useCurrentUser();
```
darrenvong: If this page is meant to be shared for both the current user and viewing other users, this won't work. We'll probably need to delegate this to a consuming parent component and get them to pass in a user here instead.

app/frontend/src/utils/date.ts
```diff
@@ -2,10 +2,29 @@
...
 export function formatDate(date: string, short: boolean = false): string {
   return moment(date, "YYYY-MM-DD").format(short ? "D MMM" : "D MMM YYYY");
 }
 
 export function timestamp2Date(timestamp: Timestamp.AsObject): Date {
   return new Date(Math.floor(timestamp.seconds * 1e3 + timestamp.nanos / 1e6));
 }
+
+export function monthYear(date: Date): string {
+  return `${MONTHS[(date.getMonth() + 1) % 12]} ${date.getFullYear()}`;
```
darrenvong: I've been needing this in my branch too and I found that we can do it through the platform with the [`Intl.DateTimeFormat` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), so I don't think we need this. The advantage of using this is it's also configurable to format date for different language locales based on the user's language settings (with `navigator.language`). Both of these seem to have decent browser support so I think it's safe to go with it!

darrenvong (changes requested):

app/frontend/src/components/TabBar/TabBar.tsx
darrenvong: As to why it was there in the first place, let's say it's one of those "why the hell did I do this?" moment when I realised I was also the criminal of this 🙈

app/frontend/src/AppRoutes.tsx
andreaowu: (Why) do we need to be able to view another user's profile with `/user/<username>` if we can do that with `/profile/username`?

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
```diff
@@ -1,21 +1,20 @@
...
   editHostingPreferenceRoute,
   editProfileRoute,
-  favoritesRoute,
-  homeRoute,
-  photosRoute,
-  profileRoute,
-  referencesRoute,
 } from "../../../routes";
+import { service } from "../../../service";
+import { getCurrentUser } from "../../../service/user";
```
darrenvong: It's better to access this through the `service` import above even though it's more characters to type, since we auto mock the `service` module in the current test setup (rather than having to mock individual files in `service`) to make things easier for writing tests.

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
```diff
@@ -43,47 +42,60 @@ const useStyles = makeStyles((theme) => ({
...
+        setLoading(true);
+        try {
+          setUser(await (username ? service.user.getUser(username) : getCurrentUser()));
+        } catch (e) {
+          console.error(e);
+          setError(e.message);
+        }
+        setLoading(false);
+    })();
+  }, [username]);
```
darrenvong: Hmm it'd be more ideal to use the existing `useCurrentUser` or `useUserByUsername` hook instead since it uses React Query so we get these error and loading state for free from the lib. Also it re-uses users that have been previously fetched from the cache while it refetches in the background so things will be faster to load too.

Although not sure we can do that with how it's currently structured since we can't conditionally use a hook in React...

darrenvong (changes requested):

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
darrenvong: The best way I can think of is to store the username in the auth state, so we can default to the current user's username if one isn't given by the URL. Thoughts on this @lucaslcode ?

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
```diff
@@ -20,15 +46,65 @@ const useStyles = makeStyles({
...
+  const { username } = useParams<{
+    username?: string;
+  }>();
+
+  const {
+    data: user,
+    isLoading: loading,
+    error,
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+  } = username ? useUserByUsername(username) : useCurrentUser();
```
lucaslcode:
```suggestion
  const currentUser = useCurrentUser();
  const {
    data: user,
    isLoading: loading,
    error,
  } = useUserByUsername(username  ?? currentUser.username);
```

andreaowu:
Erm...did I accidentally just delete everything?

lucaslcode:
I think you can reopen this somehow

lucaslcode:
~~Wait, it's merged!~~ Wait it's not

andreaowu:
It's ok, I figured it out, will commit in a bit

darrenvong (commented):
Screenshot 2021-02-27 at 00 38 20

Not sure if this is related to the `Container` component change, but the card looks a bit off centre and stuck/overflown to the right on anything at the `md` breakpoint or below

andreaowu:
Fixed, thanks for catching that!
Screen Shot 2021-02-26 at 6 12 40 PM

darrenvong (approved):
Looks good!

PR #714, opened 2021-02-27
home page

andreaowu:
Screen Shot 2021-02-27 at 8 12 56 PM


darrenvong (changes requested):
I'm getting this error in the console with the preview (and a blank screen, so pretty critical) when clicking to the "home" tab...

```
Error: Minified React error #31; visit https://reactjs.org/docs/error-decoder.html?invariant=31&args[]=object%20with%20keys%20%7Bvalue%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at wi (react-dom.production.min.js:140)
    at react-dom.production.min.js:150
    at Da (react-dom.production.min.js:176)
    at qs (react-dom.production.min.js:271)
    at Al (react-dom.production.min.js:250)
    at Tl (react-dom.production.min.js:250)
    at El (react-dom.production.min.js:250)
    at ml (react-dom.production.min.js:243)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:18)
```

app/frontend/src/features/profile/profilePage/ProfilePage.tsx
```diff
@@ -87,7 +88,7 @@ export default function ProfilePage() {
                 <About {...{ user }} />
               </TabPanel>
               <TabPanel value="home">
-                <div>home</div>
+                <Home {...{ user }}></Home>
```
darrenvong:
```suggestion
                <Home user={user}}></Home>
```
Nitpick: Use the more explicit syntax to make it clear it's only one prop please :) The spread syntax is typically used for a "wildcard"/larger collection of props where you want to forward them to child components.

app/frontend/src/features/profile/profilePage/Home.tsx
```diff
@@ -0,0 +1,125 @@
...
+            text={`${user.lastMinute?.value || ""}`}
+          />
+          <LabelAndText label={MAX_GUESTS} text={`${user.maxGuests}`} />
+          <LabelAndText
+            label={ACCEPT_SMOKING}
+            text={`${user.smokingAllowed}`}
+          />
+          <LabelAndText
+            label={ACCEPT_DRINKING}
+            text={`${user.drinkingAllowed}`}
```
darrenvong:
```suggestion
            text={`${user.drinkingAllowed?.value}`}
```
This is probably one of the culprit of that console error, since it's a `BoolValue.AsObject` rather than a primitive `boolean`, so you got to "unwrap" it. I'm surprised this error doesn't show up in dev 🤔 

Also since this isn't a string field, not sure if `false` is a sensible default if the user hasn't set a value on this, since we can't really tell if they allow drinking or not if they didn't say. Or should we not show the field altogether? @lucaslcode thoughts?

app/frontend/src/features/profile/profilePage/Home.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </div>
+        <div className={classes.info}>
+          <Typography variant="h1">{ABOUT_HOME}</Typography>
+          <LabelAndText
+            label={SPACE}
+            text={`${user.sleepingArrangement || ""}`}
+          />
+          <LabelAndText
+            label={HAS_HOUSEMATES}
+            text={`${user.hasHousemates} ${user.housemateDetails}`}
```
darrenvong: Same problem as above here, these are both `*.AsObject` rather primitive values

app/frontend/src/features/profile/profilePage/Home.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <LabelAndText label={HOST_DRINKING} text={`${user.drinksAtHome}`} />
+          <LabelAndText
+            label={HOST_PETS}
+            text={`${user.hasPets?.value || ""} ${user.petDetails}`}
+          />
+          <LabelAndText
+            label={HOST_KIDS}
+            text={`${user.hasKids?.value || ""} ${user.kidDetails}`}
+          />
+          <LabelAndText label={PARKING} text={`${user.parking}`} />
```
darrenvong: Same problem here with these user fields that aren't being "unwrapped"

app/frontend/src/features/profile/profilePage/Home.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <Typography variant="body1">{user.parkingDetails}</Typography>
+      <Divider />
+      <Typography variant="h1">{SLEEPING_ARRANGEMENT}</Typography>
+      <Typography variant="body1">{user.sleepingDetails}</Typography>
+      <Divider />
+      <Typography variant="h1">{HOUSE_RULES}</Typography>
+      <Typography variant="body1">{user.houseRules}</Typography>
+      <Divider />
+      <Typography variant="h1">{ADDITIONAL}</Typography>
+      <Typography variant="body1">{user.otherHostInfo}</Typography>
```
darrenvong: And all of these apart from `parkingDetails` 😅 . It's an enum so you might want to provide a mapping of those to more user friendly strings, otherwise I think it'll print out a meaningless numeric code.

app/frontend/src/features/profile/profilePage/Home.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <LabelAndText
+            label={WHEELCHAIR}
+            text={`${user.wheelchairAccessible?.value || ""}`}
+          />
+        </div>
+        <div className={classes.info}>
+          <Typography variant="h1">{ABOUT_HOME}</Typography>
+          <LabelAndText
+            label={SPACE}
+            text={`${user.sleepingArrangement || ""}`}
```
darrenvong: Same comment as `parkingDetails` below about this being an enum.

lucaslcode:
I wonder if it might be good not to show any fields that are undefined or an empty string (or for enums, ENUM_TYPE_UNSPECIFIED). Maybe another widget or extend `LabelText` so it just returns `null` in those cases?

darrenvong:
My concern with that would be that top section could look a bit bare if we don't show the undefined fields, and probably would give off a less uniform feel considering those fields aren't free form fields. But then I'm also not sure about leaving them empty either... although if I remember correctly, that's what Couchsurfing did and it didn't look that bad?

andreaowu:
What about "Ask me" or something like that?

darrenvong (approved):
"Ask me" sounds good to me! Love the accompanying boolean helper with it as well 🙌

PR #715, opened 2021-02-28
[MRG] Backend/feature/user hiding

ShmuelTreiger:
A few things:

1) ~~By far the biggest issue with this is that I have essentially duplicate properties in models. I couldn't figure out how to circumvent this issue. Full details at the end of this comment.~~ Fixed by using just the right combination of |, (), and ~
2) I searched the code base for places where all users were queried and added a filter on hidden users where it seemed appropriate. I ended up rolling a bunch of these back. In any event, it's entirely possible I misjudged some instances in both directions. Edit: Found a ton more, with I'm sure more to go
3) ~~I started a basic repository for standardized sql queries. I feel like that's easier to use and maintain. Very easy to delete and rollback if just filtering each individual query is better.~~ Nixed
4) I wasn't sure where the test I created fit best, so I created its own test file. Can easily be moved wherever.
5) ~~By design, creating a user who is banned raises an error. However, a banned user is necessary to check that user hiding works. I am unsure how to circumvent this issue. Consequently, the commented out line of code in my test.~~ Implemented workaround

~~Details:~~

~~As far as I can understand it, when querying the database, sqlalchemy saves boolean expressions instead of evaluating them (e.g. "self.accepted_tos < 1", instead of T or F). This makes sense, since it wants to apply this expression to evaluate every query result. However, we need to both be able to access a user's properties, while also being able to query on them (i.e. sometimes we need to know if this user.is_jailed==True, other times we're querying the database for all jailed users). I cannot figure out a way to convert between boolean expression and bool. I often got errors to the effect of "sqlalchemy.sql.elements.AsBoolean received, bool expected". The only solution which I've seen suggested and didn't try is to query the database for a specific user every time we want to check a given user's value, but this seemed excessive to me.~~


app/backend/src/couchers/migrations/versions/b4645870f298_adding_is_deleted.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision = "02eda17d1e9b"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_index("idx_events_geom", table_name="events")
+    op.drop_index("idx_nodes_geom", table_name="nodes")
+    op.drop_index("idx_page_versions_geom", table_name="page_versions")
```
aapeliv: Please remove these unrelated drops (and creates). We'll have to figure out how to stop sqlalchemy from touching the unrelated indexes

app/backend/src/couchers/models.py
```diff
@@ -186,10 +187,22 @@ class User(Base):
     def is_jailed(self):
         return self.accepted_tos < 1 or self.is_missing_location
 
+    @hybrid_property
+    def is_jailed_for_sql(self):
+        return or_(self.accepted_tos < 1, self.is_missing_location_for_sql)
+
```
aapeliv: You should create a hybrid property like you've done, but I think you need to use the `|` operator instead of `or_`, that allows SA to evaluate both in python and in the db. I think that's what you were writing about in your comment.

Something like this:

```py3
    @hybrid_property
    def is_jailed_for_sql(self):
        return (self.accepted_tos < 1) | self.is_missing_location_for_sql
```

Then you can drop the `_for_sql` endings

app/backend/src/couchers/standardized_queries.py
```diff
@@ -0,0 +1,7 @@
+from sqlalchemy import not_
+
+from couchers.models import User
+
+
+def query_users_who_arent_hidden(session):
+    return session.query(User).filter(not_(User.is_hidden_for_sql))
```
aapeliv: This should be left at the query site, as not all queries use exactly this same pattern (e.g. when joining from non-User table to User).

Also, you can simplify the not by writing `(~User.is_hidden)` instead

aapeliv (dismissed):

aapeliv:
Also, it might be nicer to write the properties as something mostly truthy, such as `User.visible` instead of `User.hidden`. This makes it nicer to query since you don't need to add the negations everywhere, so you write `session.query(User).filter(User.visible)` instead of `session.query(User).filter(~User.hidden)`.

app/backend/src/couchers/migrations/versions/b4645870f298_adding_is_deleted.py
ShmuelTreiger: I wasn't sure about this. Meant to ask, but forgot

app/backend/src/couchers/models.py
ShmuelTreiger: | throws all sorts of errors. I've attached two examples, the first from using | in is_missing_location, the second from using | in is_jailed. I arrived at the or_() solution from stack overflow.

[using or when querying on missing location.txt](https://github.com/Couchers-org/couchers/files/6058258/using.or.when.querying.on.missing.location.txt)
[using or for normal jail testing.txt](https://github.com/Couchers-org/couchers/files/6058259/using.or.for.normal.jail.testing.txt)

ShmuelTreiger:
>By design, creating a user who is banned raises an error. However, a banned user is necessary to check that user hiding works. I am unsure how to circumvent this issue. Consequently, the commented out line of code in my test.

As far as I'm aware, this is the only remaining problem

app/backend/src/tests/test_api.py
```diff
@@ -74,20 +74,20 @@ def test_coords(db):
...
-        assert res.lat == 0.0
-        assert res.lng == 0.0
-        assert res.radius == 0.0
+    with api_session(token1) as api:
+        res = api.GetUser(api_pb2.GetUserReq(user=user2.username))
+        assert res.city == user2.city
+        lat, lng = user2.coordinates or (0, 0)
+        assert res.lat == lat
+        assert res.lng == lng
+        assert res.radius == user2.geom_radius
```
aapeliv: Why did you modify this test?

app/backend/src/tests/test_visible_users.py
```diff
@@ -0,0 +1,17 @@
...
+
+def test_hidden_users(db):
+    user1, token1 = generate_user()
+    user2, token2 = generate_user()
+    user3, token3 = generate_user(is_deleted=True)
+    user4, token4 = generate_user(accepted_tos=0)
+
+    with session_scope() as session:
+        user2 = get_user_by_field(session, user2.username)
+        user2.is_banned = True
```
aapeliv: You can just set this on line 8

app/backend/src/couchers/db.py
```diff
@@ -135,19 +135,24 @@ def get_user_by_field(session, field):
...
     elif is_valid_email(field):
         logger.debug(f"Field matched to type email")
-        return session.query(User).filter(User.email == field).one_or_none()
+        query = query.filter(User.email == field)
     else:
         logger.debug(f"Field {field=}, didn't match any known types")
         return None
 
+    query = query.filter(User.is_visible).one_or_none()
+    return query
```
aapeliv: No need to assign again to query here, just return the `.one_or_none()` result

aapeliv (commented):
Whenever you add the visibility check, please move it to be the first thing in the query, like so: `session.query(User).filter(User.visible)`

app/backend/src/tests/test_api.py
ShmuelTreiger: User 2 can't get user 1 anymore (user 1 is invisible). I explained it in the commit notes.

app/backend/src/tests/test_visible_users.py
ShmuelTreiger: 'generate_user(is_banned=True)' throws errors, since you can't have a user who is both banned and active. See point 5 in my original comment in this thread.

ShmuelTreiger:
Looking for some feedback on that last test I uploaded, specifically:
1) ~~Help on how to use pytest fixture properly. I couldn't figure it out~~ Nixed this idea, partially because it wasn't working as I wanted it to, partially because it wasn't proving particularly helpful
2) ~~Unsure if a number of test are necessary (marked by a question mark)~~ Lucas answered
3) ~~Similarly, unsure about some desired behavior, also marked with a ?~~ Still unsure about some behavior, but leaving for now
4) I'm sure there are use cases I haven't considered

app/backend/src/tests/test_api.py
aapeliv: Please add a bit in the test to check the user is invisible

app/backend/src/tests/test_visible_users.py
aapeliv: OK

app/backend/src/tests/test_api.py
ShmuelTreiger: I have a test for it in app/backend/src/tests/test_visible_users.py

ShmuelTreiger:
1) ~~I did some of these queries in two ways, either as one large query or broken up. The large query is what I understood I should be doing from the backend meeting, but I really think that doing so many unnecessary, large joins will kill us down the road.~~ Did some research and doing one query seems to be recommended.

2) ~~I'm not sure all these filters/tests are necessary. For example, if you can't see an invisible user, you shouldn't be able to send them a host request. The only edge case would be if you open the webpage to send the request and the user is hidden (i.e. banned) between then and when you click send. I don't know what would happen in this edge case if we don't account for it, so maybe it's necessary? I'm just unsure.~~ Lucas answered

3) Would love opinions if there's anywhere else I should be filtering for visible users

lucaslcode:
1. What aspect of the joins do you think is dangerous? AFAIK in terms of speed, joins > extra queries >> any kind of python filtering. 

You could even create a function that takes a list of user_id fields, loops through them doing a join and filter and returning the query, so then you can do `query.user_visibility_filter(Table.from_user, Table.to_user).other_sql_stuff()`

2. It's necessary - we can't trust incoming requests, since it's possible to create a request manually without a frontend (and people can write their own frontends)

3. Haven't looked through thoroughly, but I think the answer is nearly everywhere

ShmuelTreiger:
>What aspect of the joins do you think is dangerous? AFAIK in terms of speed, joins > extra queries >> any kind of python filtering.

That hasn't been my (admittedly limited) experience with SQL. I've only ever worked on one big database before, and we had problems with big complicated joins, and it was faster to run separate queries.

>You could even create a function that takes a list of user_id fields, loops through them doing a join and filter and returning the query, so then you can do query.user_visibility_filter(Table.from_user, Table.to_user).other_sql_stuff()

Not 100% sure I understand, but do you mean something similar to what I did in my commented out query in api.ListFriends?

>It's necessary - we can't trust incoming requests, since it's possible to create a request manually without a frontend (and people can write their own frontends)

That makes a lot of sense, thanks!

>Haven't looked through thoroughly, but I think the answer is nearly everywhere

I keep finding more and more places. Which is why I'm sure I'm missing plenty.

ShmuelTreiger:
This is mostly done, a few things:

1. ~~What's desired functionality for direct and group messages if a user is hidden? Do the messages go away for the other parties or are they still visible? I largely haven't written filters or tests for conversations.py because I'm unsure how this should work.~~ Will bring to product team

2. ~~One thing that occurs to me, should I be checking that the user making the API request is visible. For example, if someone runs a search query, should the first thing be to check that context.user_id is a visible user before attempting the search?~~

3. ~~Should there be some sort of intercept for deleted users?~~ Done

4. ~~Need some help with a frontend change~~

app/backend/src/tests/test_api.py
ShmuelTreiger: Update: now moved to test_api

ShmuelTreiger:
Assuming all checks pass, I propose merging this, and starting a new issue for dealing with conversations.

app/backend/src/couchers/models.py
```diff
@@ -184,11 +185,15 @@ class User(Base):
...
 
-    @property
+    @hybrid_property
     def is_missing_location(self):
-        return not self.geom or not self.geom_radius
+        return (self.geom == None) | (self.geom_radius == None)
+
+    @hybrid_property
+    def is_visible(self):
+        return ~(self.is_banned | self.is_deleted | self.is_jailed)
```
aapeliv: Jailed users should be visible

app/backend/src/couchers/servicers/api.py
```diff
@@ -464,13 +472,17 @@ def ListFriendRequests(self, request, context):
         with session_scope() as session:
             sent_requests = (
                 session.query(FriendRelationship)
+                .join(User, User.id == FriendRelationship.to_user_id)
+                .filter(User.is_visible)
```
aapeliv, line 539: I wonder if it's possible to do something like `.filter(FriendRelatoinship.to_user.is_visible)`? Might be some SA magic that we could use? Though maybe joins are nice and explicit?

app/backend/src/couchers/servicers/api.py
```diff
@@ -574,7 +590,10 @@ def WriteReference(self, request, context):
             rating=request.rating,
         )
         with session_scope() as session:
-            if not session.query(User).filter(User.id == request.to_user_id).one_or_none():
+            if (
+                not session.query(User).filter(User.is_visible).filter(User.id == reference.from_user_id).one_or_none()
```
aapeliv: No need to check `from_user`, they're guaranteed to exist and be visible (needs to be handled in auth)

app/backend/src/couchers/servicers/auth.py
```diff
@@ -94,6 +94,9 @@ def _create_session(self, context, session, user, long_lived):
         if user.is_banned:
             context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.ACCOUNT_SUSPENDED)
 
+        if user.is_deleted:
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.ACCOUNT_DELETED)
```
aapeliv: This should behave as if the user didn't exist (so this needs to be handled before `_create_session`, I think). Just do `is_visible` in the search for the user in login?

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -398,9 +406,9 @@ def CreateGroupChat(self, request, context):
             )
             session.add(your_subscription)
 
-            for recipient in request.recipient_user_ids:
+            for recipient in recipient_user_ids:
```
aapeliv: I reckon just move the hidden check into this for loop. Due to db atomicity, it'll roll everything back.

So just add a check here before the friendship check to make sure the user exists and is visible, no need to do the big check above. (Also it should outright fail if there's invalid users, not filter them out).

(I see now there's no explicit check to make sure the user exists, but it'd turn up either in the friend check or when the database foreign key doesn't work)

The messagign stuff needs a refactorign anyway, no need to be super elegant

app/backend/src/couchers/servicers/groups.py
```diff
@@ -199,10 +211,14 @@ def JoinGroup(self, request, context):
...
-            current_membership = cluster.members.filter(User.id == context.user_id).one_or_none()
-            if current_membership:
+            user_in_group = cluster.members.filter(User.id == context.user_id).one_or_none()
+            if user_in_group:
                 context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.ALREADY_IN_GROUP)
 
+            user_visible = session.query(User).filter(User.is_visible).filter(User.id == context.user_id).one_or_none()
+            if not user_visible:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
+
```
aapeliv: No need to check if `context.user_id` exists/is visible/etc. That needs to be handled in auth

app/backend/src/couchers/servicers/groups.py
```diff
@@ -223,10 +239,14 @@ def LeaveGroup(self, request, context):
...
-            if not current_membership:
-                context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.NOT_IN_GROUP)
+            user_in_group_and_visible = (
+                cluster.members.filter(User.is_visible).filter(User.id == context.user_id).one_or_none()
+            )
+            if not user_in_group_and_visible:
+                user_in_group = cluster.members.filter(User.id == context.user_id).one_or_none()
+                if not user_in_group:
+                    context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.NOT_IN_GROUP)
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
```
aapeliv: Ditto

app/backend/src/couchers/servicers/requests.py
```diff
@@ -60,8 +60,12 @@ def CreateHostRequest(self, request, context):
         with session_scope() as session:
             if request.to_user_id == context.user_id:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.CANT_REQUEST_SELF)
-            # just to check the host exists
-            host = session.query(User).filter(User.id == request.to_user_id).one_or_none()
+
+            # just to check guest and host exist and are visible
+            guest = session.query(User).filter(User.is_visible).filter(User.id == context.user_id).one_or_none()
+            if not guest:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
```
aapeliv: Again, no need to check context user id

aapeliv (changes requested):
This is looking very good. Please rebase on `develop` as it's pretty far behind. The rebase luckily isn't very bad (I did it earlier). Thanks for comprehensive testing!

I think the auth logic needs to be corrected. When a user is banned, they get an error saying they were banned. When they're deleted, the logic needs to be the same as if the user never existed. `get_session_for_token` needs to filter away deleted/banned users.

The other thing is that given that logic in the auth interceptor, we always assume `context.user_id` is correct and that user exists in the database and is allowed to do the call, so no need to check visibility of that user in handlers, etc.

Sorry for taking forever to review this PR.

app/backend/src/couchers/models.py
ShmuelTreiger: This was the original spec you gave me. That being said, I agree. I've also come to think that "is_visible" might not be nuanced enough. See my spreadsheet here:

https://docs.google.com/spreadsheets/d/1Y1e-QvbXVE_nrvJyvdm-BVOqzEV-_519qTv9vNGk4DY/edit?usp=sharing

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: As it stands now, that would catch banned users as well.

I'll just add a check 'not user.is_deleted'

ShmuelTreiger:
>The other thing is that given that logic in the auth interceptor, we always assume context.user_id is correct and that user exists in the database and is allowed to do the call, so no need to check visibility of that user in handlers, etc.

This contradicts what Lucas said above:

>It's necessary - we can't trust incoming requests, since it's possible to create a request manually without a frontend (and people can write their own frontends)

And no worries on taking your time, everyone's busy

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: I feel like the check should still come at the beginning. I can loop through all recipients to check they're visible. What if it's only sending a message to one person and that person is invisible, this solution wouldn't catch that.

ShmuelTreiger: Using is_visible on an individual user was proving impossible, so I just added a length check

app/backend/src/couchers/models.py
aapeliv: Yeah, I had a look at your spreadsheet earlier and it looked good. Let's go with how it's there?

app/backend/src/couchers/servicers/conversations.py
aapeliv: > What if it's only sending a message to one person and that person is invisible, this solution wouldn't catch that.

Why?

aapeliv:
> This contradicts what Lucas said above:

We have the auth interceptor that specifically makes sure the user is authenticated and exists. So any requests that hit the handler within a secure service are guaranteed to have an existent user.

I think Lucas's reply was in relation to you saying this:

> if you can't see an invisible user, you shouldn't be able to send them a host request

But this isn't caught by the authentication interceptor. A user could create a custom request and get an invisible user's ID by just guessing, or having it from before, or anything, and then craft a request (although with an authenticated user).

So we can't trust that the requests are in good faith and are constructed from what we'd assume ("there's no way to do X in the frontend, so when we get requests, they cannot be asking for X"), but we can trust that the user is auth'd and exists.

ShmuelTreiger:
Got it. Completely misunderstood.

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: Because it comes after and individual message has been sent.

https://github.com/Couchers-org/couchers/blob/4ea4ef3daa6fc658629d59d1d1dfbfe3f018ca08/app/backend/src/couchers/servicers/conversations.py#L363-L409

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: Tried it, got this error: https://stackoverflow.com/questions/16589208/attributeerror-while-querying-neither-instrumentedattribute-object-nor-compa

Suggested "workaround" is using joins

ShmuelTreiger:
OK, I think this is pretty much done. I think the next area of improvement would be in hiding/archiving messages, I didn't do any of that. Do we even have an archive function yet?

app/backend/src/couchers/servicers/api.py
aapeliv: Thanks for trying. Joins are probably more explicit anyway!

app/backend/src/couchers/servicers/conversations.py
aapeliv: That just makes sure you don't already have another DM with someone when you try to make a new one. It's not sending the individual message

app/backend/src/couchers/models.py
```diff
@@ -189,11 +190,15 @@ class User(Base):
...
 
-    @property
+    @hybrid_property
     def is_missing_location(self):
-        return not self.geom or not self.geom_radius
+        return (self.geom == None) | (self.geom_radius == None)
+
+    @hybrid_property
+    def is_visible(self):
+        return ~(self.is_banned | self.is_deleted)
```
aapeliv, line 201: Doesn't matter so much, but is this clearer as `~is_banned & ~is_deleted`?

app/backend/src/couchers/servicers/api.py
```diff
@@ -412,9 +412,27 @@ def UpdateProfile(self, request, context):
...
+        users2 = aliased(User)
+
         with session_scope() as session:
             rels = (
                 session.query(FriendRelationship)
+                .join(
+                    users1,
+                    and_(
+                        users1.id == FriendRelationship.from_user_id,
+                        FriendRelationship.from_user_id is not context.user_id,
```
aapeliv: Why is this needed? If `context.user_id` is not visible, they shouldn't make it this far, so no need to worry for that.

Also, I don't think `is not` works here, should be `!=` so SA can overload it.

[Unrelated note: I believe it's often a bad idea to have expressions in the join condition for SQL performance reasons, it trashes the query planner's ability to use indexes and often causes full table scans which are very slow with more rows.]

app/backend/src/couchers/servicers/api.py
```diff
@@ -492,6 +511,8 @@ def RespondFriendRequest(self, request, context):
         with session_scope() as session:
             friend_request = (
                 session.query(FriendRelationship)
+                .join(User, FriendRelationship.from_user_id == User.id)
```
aapeliv: I like to keep the expression as `Foreign.column == This.column` for consistency, `Foreign` is `User` here and `This` is `FriendRelationship`

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -342,20 +342,32 @@ def SearchMessages(self, request, context):
...
+                for user in (
+                    session.query(User.id).filter(User.is_visible).filter(User.id.in_(request.recipient_user_ids)).all()
+                )
+            ]
+
+        # make sure all requested users are visible
+        if len(recipient_user_ids) != len(request.recipient_user_ids):
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.USER_NOT_FOUND)
+
+        if len(recipient_user_ids) < 1:
```
aapeliv: Know it was wrong before as well, but this should just be `if not recipient_user_ids`

app/backend/src/couchers/servicers/groups.py
```diff
@@ -199,8 +211,8 @@ def JoinGroup(self, request, context):
             if not cluster:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.GROUP_NOT_FOUND)
 
-            current_membership = cluster.members.filter(User.id == context.user_id).one_or_none()
-            if current_membership:
+            user_in_group = cluster.members.filter(User.id == context.user_id).one_or_none()
+            if user_in_group:
```
aapeliv, line 215: Please don't change this if there's no good reason.

app/backend/src/couchers/servicers/references.py
```diff
@@ -55,11 +57,23 @@ def ListReferences(self, request, context):
             if not request.from_user_id and not request.to_user_id:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.NEED_TO_SPECIFY_AT_LEAST_ONE_USER)
 
+            users1 = aliased(User)
+            users2 = aliased(User)
```
aapeliv: Would it be better to call these `from_users` and `to_users`? More descriptive?

app/backend/src/couchers/servicers/references.py
```diff
@@ -55,11 +57,23 @@ def ListReferences(self, request, context):
...
+                    .filter(Reference.from_user_id == request.from_user_id)
+                )
             if request.to_user_id:
-                query = query.filter(Reference.to_user_id == request.to_user_id)
+                # join the from_users, because only interested if the writer is visible
+                query = (
+                    query.join(users2, users2.id == Reference.from_user_id)
+                    .filter(~users2.is_banned)  # instead of is_visible; if user is deleted, reference still visible
+                    .filter(Reference.to_user_id == request.to_user_id)
+                )
```
aapeliv, line 75: Not sure why I can't comment here, but I think the subquery also needs a visibility check? (`sub = session...`)

app/backend/src/couchers/tasks.py
```diff
@@ -1,5 +1,6 @@
 import logging
 
+from sqlalchemy import not_
```
aapeliv: Unused import

app/backend/src/tests/test_api.py
```diff
@@ -297,13 +331,23 @@ def test_friend_request_flow(db):
     with api_session(token1) as api:
         api.SendFriendRequest(api_pb2.SendFriendRequestReq(user_id=user2.id))
 
+    with session_scope() as session:
+        friend_request = (
+            session.query(FriendRelationship)
+            .filter(FriendRelationship.from_user_id == user1.id and FriendRelationship.to_user_id == user2.id)
+            .one_or_none()
+        )
+        session.expunge(friend_request)
```
aapeliv: Please don't use expunge unless really necessary (even in tests): it causes a lot of errors that are hard to understand especially if one hasn't dealt with them before!

Instead just save the FR id as `friend_request_id = friend_request.id` and use that outside this scope

app/backend/src/tests/test_groups.py
```diff
@@ -262,8 +213,91 @@ def test_ListOtherUserGroups(testing_communities):
             assert [g.group_id for g in res.groups] == [hitchhikers_id, foodies_id, skaters_id]
 
 
+# These tests are separate from above as they mutate the database
```
aapeliv: `testing_communities` takes quite a while to set up. Would it be better to add some hidden users in that setup script and then not mutate the database with `make_user_invisible` within the test at all?

app/backend/src/tests/test_visible_users.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+from couchers.db import get_user_by_field
+from couchers.models import User
+from tests.test_fixtures import db, generate_user, make_user_invisible, session_scope
+
+
+def test_visible_user_filter(db):
+    user1, token1 = generate_user()
+    user2, token2 = generate_user()
```
aapeliv, line 9: Just this should work:

```py3
user2, token2 = generate_user(is_banned=True)
```

app/backend/src/tests/test_api.py
```diff
@@ -507,6 +642,53 @@ def test_CancelFriendRequest(db):
...
+    with api_session(token1) as api:
+        api.SendFriendRequest(api_pb2.SendFriendRequestReq(user_id=user2.id))
+
+    with session_scope() as session:
+        friend_request = (
+            session.query(FriendRelationship)
+            .filter(FriendRelationship.from_user_id == user1.id and FriendRelationship.to_user_id == user2.id)
+            .one_or_none()
+        )
+        session.expunge(friend_request)
```
aapeliv: Please don't expunge

aapeliv (changes requested):
See comments. Some main things:

1. Please don't user `session.expunge` unless really necessary: it's very error prone and causes issues. Similarly, please don't user `get_user_by_field` except for like the two API calls that accept usernames/IDs/emails.
2. I don't understand why you are using `and_`s in a bunch of the join conditions? I've added a comment on one, but it applies to all cases.
3. I think it'd be better to call the aliased tables `from_users`, etc instead of `users1`, makes it clearer

app/backend/src/couchers/models.py
ShmuelTreiger: I think not, but I can see arguments both ways

aapeliv: OK, keep as is!

app/backend/src/couchers/servicers/groups.py
ShmuelTreiger: I changed it as part of a bigger change that got rolled back. I think it's clearer, but doesn't really matter

app/backend/src/tests/test_visible_users.py
ShmuelTreiger: It doesn't. Auth interceptor gets it. Try it yourself

app/backend/src/couchers/servicers/groups.py
aapeliv: There's a few places where this kind of stuff is done (`if [not] (...).one_or_none`), maybe we should write a quick wrapper helper or just start standardizing. This is also wasteful currently: the whole user gets shipped to the backend. Something like `if (...).count() [!/=]= 0` or something...

Anyway, leave this if you don't feel like cleaning up, it's fine either way (user_in_group or current_membership)

ShmuelTreiger: I'm not understanding your first comment

Happy to change back while I clean up the rest, like I said, doesn't really matter

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: You mean

`if not len(recipient_user_ids) == len(request.recipient_user_ids):`

?

app/backend/src/tests/test_visible_users.py
aapeliv: Ah... sorry about that.

How about `session.query(User).filter(User.id == user2.id).one().is_banned = True`? Then no need to refresh/expunge the user since it's not written to

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: I thought it would be faster, but If it's slower then ya, I'll take it out.

app/backend/src/tests/test_visible_users.py
ShmuelTreiger: It works without the expunge and refresh, just forgot to take them out.

app/backend/src/couchers/servicers/groups.py
aapeliv: No need to change it back, it's good as is.

I mean that there are several places where all we do is check whether some object exists (in here, whether there is a `User` that matches that filter in `cluster.members`). Doing `one_or_none()` does a `SELECT * FROM ... WHERE ...` that fetches the whole `User` object from the database — just so we can check whether that object is `None` or not. That's pretty wasteful, instead we'd want to do something like `SELECT count(*) FROM ... WHERE ...`. We do this in quite a few different places, and should probably refactor the full `SELECT *` out and replace with `SELECT count(*)`. (That's outside the scope of this PR though and should be done separately.)

app/backend/src/tests/test_visible_users.py
ShmuelTreiger: Would you prefer a query instead of get_user_by_field?

app/backend/src/couchers/servicers/conversations.py
aapeliv: No, this line is checking if the `recipient_user_ids` list is empty or not. The pythonic way is to check whether the list is "falsey", i.e. whether `not recipient_user_ids` evaluates to `True`, which happens if and only if `recipient_user_ids` is an empty list.

app/backend/src/couchers/servicers/groups.py
ShmuelTreiger: Is count quicker? If so, then ya, we do it a lot. Make an issue for it. Probably good first issue.

app/backend/src/tests/test_visible_users.py
aapeliv: Yes please

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: Got it. I'll change it. But ya, unrelated to this PR.

app/backend/src/couchers/servicers/groups.py
aapeliv: Yeah, `count(*)` just returns ~1 bytes + overhead to the backend from the database (and can probably be done using only an index, without having to visit the table). `*` returns the whole user object to the backend and requires definitely a table access. (Also, `User` models has some `column_property`s that also need to be computed whenever we grab the whole thing.)

This requires a tad bit of investigation though, to make sure SQLAlchemy doesn't do something dumb (like `SELECT count(*) FROM (SELECT *) AS t ...` which would negate a bunch of the point. Will open an issue.

app/backend/src/tests/test_groups.py
ShmuelTreiger: If it's faster, then yes, that would make more sense

app/backend/src/couchers/servicers/api.py
aapeliv: I don't know if it's faster, but I think it's much clearer from a code maintainability POV to take it out

app/backend/src/tests/test_groups.py
aapeliv: Yeah, that's the reason for the somewhat unusual class setup in this test file. Thanks!

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: Ya, you're right, I misread the indentations. Still, I think a quick length check at the beginning will save a bunch of backend work if one of the recipients is invisible. Plus recipient.is_visible is proving super hard to work with.

app/backend/src/couchers/servicers/references.py
ShmuelTreiger: I don't think so. Because in the first query we filter out everything sent to/from banned users. Even if we pick it up in the second query, the outerjoin will get rid of it. We can add it, but it's an extra join. I think not adding is faster (and maybe clearer?)

ShmuelTreiger:
>I think it'd be better to call the aliased tables from_users, etc instead of users1, makes it clearer

I'm not sure it does. It's a lot of from_user == from_user which I find kind of confusing. But you can take a look for yourself.

app/backend/src/couchers/servicers/gis.py
```diff
@@ -37,7 +37,7 @@ def _query_to_geojson_response(session, query):
 class GIS(gis_pb2_grpc.GISServicer):
     def GetUsers(self, request, context):
         with session_scope() as session:
-            query = session.query(User.username, User.id, User.geom).filter(User.geom != None)
+            query = session.query(User.username, User.id, User.geom).filter(User.is_visible).filter(User.geom != None)
```
ShmuelTreiger, line 40: Is GIS tested anywhere? I just realized I never tested this change.

aapeliv: No, these endpoints are not tested!

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -342,20 +342,32 @@ def SearchMessages(self, request, context):
             )
 
     def CreateGroupChat(self, request, context):
-        if len(request.recipient_user_ids) < 1:
+        with session_scope() as session:
```
aapeliv, line 345: Do everything in the same transaction, just wrap it all in a single session

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -342,20 +342,32 @@ def SearchMessages(self, request, context):
...
+            recipient_user_ids = [
+                user.id
+                for user in (
+                    session.query(User.id).filter(User.is_visible).filter(User.id.in_(request.recipient_user_ids)).all()
+                )
+            ]
+
+        # make sure all requested users are visible
+        if len(recipient_user_ids) != len(request.recipient_user_ids):
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.USER_NOT_FOUND)
```
aapeliv: Combine this with the `set`:

`if len(set(recipient_user_ids)) != len(request.recipient_user_ids):`

app/backend/src/couchers/servicers/references.py
aapeliv: Ah, you're right.

You have to move these joins out of the if queries then, into line 58. You don't only want to filter out banned stuff if you're filtering for a user, they're separate things.

app/backend/src/couchers/servicers/references.py
```diff
@@ -252,7 +272,9 @@ def ListPendingReferencesToWrite(self, request, context):
         with session_scope() as session:
             q1 = (
                 session.query(literal(True), HostRequest)
-                .outerjoin(Reference, Reference.host_request_id == HostRequest.conversation_id)
+                .outerjoin(Reference, HostRequest.conversation_id == Reference.host_request_id)
```
aapeliv: Why did you swap the order? I prefer to have it `.join(Table, Table.col == Table2.col2)`, with the `Table` on "that" side of the join condition?

aapeliv (changes requested):
Could you standardize and move the `.filter(User.is_visible)` as early on in the query as possible, please?

Also, I added a comment about the join condition ordering, it's very small, but if you could keep it in that order, would be sweet

app/backend/src/couchers/servicers/conversations.py
ShmuelTreiger: I wanted a different error code. User Not Found vs Invalid recipient

app/backend/src/couchers/servicers/references.py
ShmuelTreiger: I don't understand. The joins are dependent upon if the request wants to see references about a user or written by the user? Where would you move them?

(Also, I assume you mean lines 59 and 60?)

ShmuelTreiger:
Needs migration, but I'll wait until it's done

PR #716, opened 2021-02-28
editing profile

andreaowu:
The following fields update correctly:
- name
- location
- meetup status
- hosting status
- occupation
- languages
- about
- about my home
- countries lived / visited

Screen Shot 2021-02-28 at 7 16 31 AM
Screen Shot 2021-02-28 at 6 22 45 PM


app/frontend/src/features/profile/constants.ts
```diff
@@ -10,17 +18,17 @@ export const smokingLocationLabels = {
...
   [HostingStatus.HOSTING_STATUS_UNSPECIFIED]: "Blank hosting ability",
   [HostingStatus.HOSTING_STATUS_UNKNOWN]: "Unknown hosting ability",
 };
 
 export const meetupStatusLabels = {
-  [MeetupStatus.MEETUP_STATUS_WANTS_TO_MEETUP]: "Wants to meet up",
-  [MeetupStatus.MEETUP_STATUS_OPEN_TO_MEETUP]: "Maybe wants to meet up",
-  [MeetupStatus.MEETUP_STATUS_DOES_NOT_WANT_TO_MEETUP]: "Can't meet up",
+  [MeetupStatus.MEETUP_STATUS_WANTS_TO_MEETUP]: MEETUP,
+  [MeetupStatus.MEETUP_STATUS_OPEN_TO_MEETUP]: MAYBE_ACCEPTING,
```
darrenvong:
```suggestion
  [MeetupStatus.MEETUP_STATUS_OPEN_TO_MEETUP]: MAYBE_MEETUP,
```

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -76,6 +105,9 @@ export default function EditProfileForm() {
   const onSubmit = handleSubmit((data) => {
     resetUpdate();
     updateUserProfile({ profileData: data, setMutationError: setErrorMessage });
+    if (!errorMessage) {
```
darrenvong: `updateUserProfile` is an async operation, but the mutation exposed is not so this will always succeed as `errorMessage` wouldn't be populated at this time. The mutation exposed is not async here as all of the logic (`onSuccess`, `onError` etc) is inside the hook that exposes this. Might be worth moving this into `onSuccess` in `useUpdateUserProfile` so the logic remain in one place.

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -53,9 +64,18 @@ export default function Overview({ user }: OverviewProps) {
...
-      <Typography variant="h1">{user.name}</Typography>
-      <Typography variant="body1">{user.city}</Typography>
+      <Typography variant="h1" className={classes.intro}>
+        {user.name}
+      </Typography>
+      <Typography variant="body1" className={classes.intro}>
+        {user.city}
+      </Typography>
       <Divider />
+      <CardActions className={classes.cardActions}>
```
darrenvong: This is shown on all profiles at the moment. Need some additional to check that the user ID matches the current user's before showing it.

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -143,31 +276,20 @@ export default function EditProfileForm() {
...
-                  onChange={onChange}
-                  value={value}
-                />
-              )}
+              defaultValue={user.aboutPlace}
+              inputRef={register}
+              className={classes.field}
             />
-
             <Controller
```
darrenvong, line 292: I noticed that this (and other dropdowns) seem to have a different font size compared to the rest of the text fields. Do you mind fixing it while you're on the page please?

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -199,17 +320,16 @@ export default function EditProfileForm() {
...
+            <div className={classes.buttonContainer}>
               <Button
                 type="submit"
                 variant="contained"
                 color="primary"
                 onClick={onSubmit}
               >
-                Save
+                {SAVE}
               </Button>
```
darrenvong: This is stuck right against the bottom of the screen on the latest preview unlike in your screenshot. 🤔  
We probably want a bit of spacing below this?

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -113,12 +145,114 @@ export default function EditProfileForm() {
...
+                      control={<Radio />}
+                      label={FEMALE}
+                    />
+                    <FormControlLabel
+                      value={MALE_FORM_VALUE}
+                      control={<Radio />}
+                      label={MALE}
+                    />
+                    <FormControlLabel
+                      value={OTHER_FORM_VALUE}
```
darrenvong, line 238: This causes any freeform value to be set to "other". Is this intentional? I thought we want to let people to set their gender to whatever they want post registration.

darrenvong (changes requested):

app/frontend/src/features/profile/edit/ProfileForm.tsx
lucaslcode: In any case, any references to gender should refer to "non-binary" rather than "other"

andreaowu: The radio button doesn't actually have a label on it, it's blank. I just needed a value for the form, but it's not the label that shows.

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -60,6 +61,13 @@ interface OverviewProps {
 
 export default function Overview({ user }: OverviewProps) {
   const classes = useStyles();
+  const [currentUser, setCurrentUser] = useState({} as User.AsObject);
+
+  useEffect(() => {
+    (async () => {
+      setCurrentUser(await getCurrentUser());
+    })();
+  }, []);
```
```suggestion
  const currentUserId = useAuthContext().authState.userId;
```

My bad, I should have mentioned in the last review that this is available on the auth context already and as this is an authenticated page, it's safe to use/should always have a value.

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -71,11 +79,13 @@ export default function Overview({ user }: OverviewProps) {
         {user.city}
       </Typography>
       <Divider />
-      <CardActions className={classes.cardActions}>
-        <Button component={Link} to={editProfileRoute}>
-          {EDIT_PROFILE}
-        </Button>
-      </CardActions>
+      {user.userId === currentUser.userId && (
```
```suggestion
      {user.userId === currentUserId && (
```

darrenvong (changes requested):

darrenvong (approved):

PR #717, opened 2021-02-28
Implement grpc timeout and nice errors

### LINKED ISSUE #659: Frontend rpc calls don't timeout
A [deadline needs to be set](https://github.com/grpc/grpc-web#setting-deadline) in the metadata. This could be done in an interceptor.

lucaslcode:
Closes #659

### DIFF
--- a/app/frontend/src/components/Alert.tsx
+++ b/app/frontend/src/components/Alert.tsx
@@ -16,6 +16,7 @@ const useStyles = makeStyles((theme) => ({
 
 interface AlertProps extends MuiAlertProps {
   severity: MuiAlertProps["severity"];
+  children: string;
 }
 
 export default function Alert({
@@ -25,21 +26,19 @@ export default function Alert({
 }: AlertProps) {
   const classes = useStyles();
 
+  const oldErrorKey = Object.keys(grpcErrorStrings).find((oldError) =>
+    children.includes(oldError)
+  );
+
   return (
     <MuiAlert {...otherProps} className={classNames(classes.root, className)}>
-      {React.Children.map(children, (child) => {
-        if (typeof child !== "string")
-          throw new Error("Alert should only have string children");
-
+      {
         // Search for the error in the ugly grpc error object keys
         // Replace it with the nice error if found
-        const oldErrorKey = Object.keys(grpcErrorStrings).find((oldError) =>
-          child.includes(oldError)
-        );
-        if (oldErrorKey)
-          return grpcErrorStrings[oldErrorKey as keyof typeof grpcErrorStrings];
-        else return child;
-      })}
+        oldErrorKey
+          ? grpcErrorStrings[oldErrorKey as keyof typeof grpcErrorStrings]
+          : children
+      }
     </MuiAlert>
   );
 }
--- a/app/frontend/src/components/Alert.tsx
+++ b/app/frontend/src/components/Alert.tsx
@@ -6,6 +6,8 @@ import {
 import classNames from "classnames";
 import React from "react";
 
+import { grpcErrorStrings } from "../constants";
+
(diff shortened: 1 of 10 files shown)


app/frontend/src/components/Alert.tsx
```diff
@@ -16,9 +18,28 @@ interface AlertProps extends MuiAlertProps {
...
+
+        // Search for the error in the ugly grpc error object keys
+        // Replace it with the nice error if found
+        const oldErrorKey = Object.keys(grpcErrorStrings).find((oldError) =>
+          child.includes(oldError)
+        );
+        if (oldErrorKey)
+          return grpcErrorStrings[oldErrorKey as keyof typeof grpcErrorStrings];
+        else return child;
+      })}
```
lucaslcode: This is gross but don't think there's a better way.

app/frontend/src/service/client.ts
```diff
@@ -37,12 +38,23 @@ export class AuthInterceptor {
   }
 }
 
-const interceptor = new AuthInterceptor();
+class TimeoutInterceptor {
+  async intercept(request: any, invoker: (request: any) => any) {
+    const deadline = Date.now() + grpcTimeout;
+    const metadata = request.getMetadata();
+    metadata.deadline = deadline;
+    const response = await invoker(request);
```
lucaslcode, line 47: In theory this works but I was only able to get a "Deadline exceeded" error once, even setting `grpcTimeout` to one and limiting my bandwidth! 🤔

app/frontend/src/components/Alert.tsx
darrenvong: Do we ever use `Alert` with more than one children? If not, and it's a bit crap for a generic component, but could type the `children` as string so we can operate on `children` directly to do this. Will simplify the code a fair bit (no need for `React.Children` and the error throwing bit etc)

app/frontend/src/service/client.ts
darrenvong: What happens if this is set to 0?

darrenvong: What happens if this is set to 0?

app/frontend/src/constants.ts
```diff
@@ -7,4 +7,15 @@ export const userLocationDefaultRadius = 500;
...
-export const reactQueryRetries = 3;
+export const reactQueryRetries = 2;
+
+export const grpcTimeout = 10000; //milliseconds
+
+export const grpcErrorStrings = {
+  "Deadline exceeded":
+    "Server took too long to respond. Please check your Internet connection or try again later.",
+  "Http response at 400 or 500 level":
+    "Couldn't find the server. Please check your Internet connection or try again later.",
```
aapeliv: "find" is a confusing term for end users, I think. "Couldn't connect to server"?

aapeliv (commented):
This looks fine to me

app/frontend/src/components/Alert.tsx
lucaslcode: Yes quite a few places it has `<Alert>Some text and a {var}</Alert>` which I wanted to support, but now I think I will do as you say since it will probably be constants/I18n keys later anyway.

app/frontend/src/service/client.ts
lucaslcode: Still nothing... I can see there is correctly a  `grpc-timeout: 10000m` header on the request.

lucaslcode:
Had to rebase, sorry.

I had a look into the grpc-web code, and setting the deadline doesn't do anything on the client side except include it in the header. So I'm a bit confused about what the point is.

Edit: It does seem to timeout correctly so whatever

darrenvong (approved):

Issue #414, opened 2020-12-10
User blocking


lucaslcode:
Api to block a user which then causes NOT_FOUND error when that user tries to access your profile, send a request, events created, etc.

Api to unblock too

PR #724, opened 2021-03-01
Backend/feature/user blocking

ShmuelTreiger:
I made an attempt at user blocking #414 . Running into the attached error, hoping someone can help me figure out what's wrong.
[TypeError: fake_handler() missing 1 required positional argument: 'request'](https://github.com/Couchers-org/couchers/files/6062531/Error.txt)


aapeliv:
> Running into the attached error, hoping someone can help me figure out what's wrong.

You need to give the API a request, even if it's empty, e.g. `api.GetBlockedUsers(empty_pb2.Empty())`

app/backend/src/couchers/models.py
```diff
@@ -1216,3 +1216,18 @@ def ready_for_retry(self):
...
+    """
+    Table of blocked users
+    """
+
+    __tablename__ = "user_blocking"
+
+    id = Column(BigInteger, primary_key=True)
+
+    blocking_user = Column(ForeignKey("users.id"), nullable=False)
+    blocked_user = Column(ForeignKey("users.id"), nullable=False)
```
aapeliv: These should be postfixed with `_id`, and you should create relationships based on them

This table also needs a uniqueness constraint each user can block another user at most once.

app/backend/src/couchers/models.py
```diff
@@ -1216,3 +1216,18 @@ def ready_for_retry(self):
...
     def __repr__(self):
         return f"BackgroundJob(id={self.id}, job_type={self.job_type}, state={self.state}, next_attempt_after={self.next_attempt_after}, try_count={self.try_count}, failure_info={self.failure_info})"
+
+
+class UserBlocking(Base):
+    """
+    Table of blocked users
+    """
+
+    __tablename__ = "user_blocking"
```
aapeliv: `user_blocks`

app/backend/src/couchers/models.py
```diff
@@ -1216,3 +1216,18 @@ def ready_for_retry(self):
 
     def __repr__(self):
         return f"BackgroundJob(id={self.id}, job_type={self.job_type}, state={self.state}, next_attempt_after={self.next_attempt_after}, try_count={self.try_count}, failure_info={self.failure_info})"
+
+
+class UserBlocking(Base):
```
aapeliv: `UserBlock`

app/backend/src/couchers/models.py
```diff
@@ -1216,3 +1216,18 @@ def ready_for_retry(self):
...
+    Table of blocked users
+    """
+
+    __tablename__ = "user_blocking"
+
+    id = Column(BigInteger, primary_key=True)
+
+    blocking_user = Column(ForeignKey("users.id"), nullable=False)
+    blocked_user = Column(ForeignKey("users.id"), nullable=False)
+    blocked = Column(Boolean, nullable=False, default=True)
```
aapeliv: Presence of a row in this table indicates that the blocking user has blocked the blocked user. We don't need the bool here (what does it mean to be blocked but have this false?)

app/backend/src/couchers/servicers/api.py
```diff
@@ -643,6 +644,77 @@ def InitiateMediaUpload(self, request, context):
...
+            user_blocking = (
+                session.query(UserBlocking)
+                .filter(UserBlocking.blocking_user == context.user_id)
+                .filter(UserBlocking.blocked_user == request.user_id)
+                .one_or_none()
+            )
+            if user_blocking:
+                user_blocking.blocked = True
+                user_blocking.time_blocked = func.now()
+                session.commit()
```
aapeliv: Just return an error that the user has already been blocked

app/pb/api.proto
```diff
@@ -55,6 +55,12 @@ service API {
   rpc InitiateMediaUpload(google.protobuf.Empty) returns (InitiateMediaUploadRes) {
     // Initiate a media upload
   }
+
+  rpc BlockUser(UserBlockingReq) returns (google.protobuf.Empty);
+
+  rpc GetBlockedUsers(google.protobuf.Empty) returns (GetBlockedUsersRes);
+
+  rpc UnblockUser(UserBlockingReq) returns(google.protobuf.Empty);
```
aapeliv: Please adhere to the style of `rpc SomeFunction` takes request `SomeFunctionReq`, so `BlockUser` should take `BlockUserReq`. Please also comment these functions!

It might also be good to separate these out into a separate `blocking.proto`, or similar?

app/pb/api.proto
```diff
@@ -385,3 +391,19 @@ message InitiateMediaUploadRes {
...
+message BlockedUserRelationship{
+  int64 id = 1;
+  int64 blocking_user_id = 2;
+  int64 blocked_user_id = 3;
+  bool blocked = 4;
+  google.protobuf.Timestamp time_blocked = 5;
+}
+
+message GetBlockedUsersRes {
+  repeated BlockedUserRelationship blocked_user_relationships = 1;
```
aapeliv: This can just return a `repeated int64 blocked_user_ids = 1`. The requestor knows their user id, and the time the block was created doesn't need to be returned

aapeliv (changes requested):
The main task is to modify **all** queries where one user requests info about another one to hide all users they've blocked. That's a huge exercise and I'm not sure exactly how we should implement that!

app/backend/src/couchers/models.py
ShmuelTreiger: >create relationships based on them

How do you mean?

>This table also needs a uniqueness constraint each user can block another user at most once.

In api.user_blocking(), it checks if the user has already been blocked and doesn't create a new block. That being said, happy to figure out how to add a uniqueness constraint

app/backend/src/couchers/models.py
ShmuelTreiger: When unblocking, instead of dropping the row, I change blocked to false (was taught that it's always safer to not delete data)

app/backend/src/couchers/models.py
aapeliv: > How do you mean?

Add SQLAlchemy `relationship` objects so you can use them easier in SQLA code, something like this:

https://github.com/Couchers-org/couchers/blob/620a41fd8d2ca07b5456d2f8deaacc0a356ab44d/app/backend/src/couchers/models.py#L450-L451

> In api.user_blocking(), it checks if the user has already been blocked and doesn't create a new block. That being said, happy to figure out how to add a uniqueness constraint

Yes, but it'd be nice to have an additional database check in place so we can guarantee the data is correct. Similar to this:

https://github.com/Couchers-org/couchers/blob/620a41fd8d2ca07b5456d2f8deaacc0a356ab44d/app/backend/src/couchers/models.py#L835

app/pb/api.proto
ShmuelTreiger: I considered this, but thought better to make it more versatile for future applications. Happy to minimize it.

ShmuelTreiger:
>The main task is to modify all queries where one user requests info about another one to hide all users they've blocked. That's a huge exercise and I'm not sure exactly how we should implement that!

Yes, this something I've been considering and don't have a solution for yet

app/backend/src/couchers/models.py
aapeliv: Our table structure is so complicated that to do "soft deletes" in general is a big pain. We've made a conscious decision to maintain only what information is required at the present state of the database.

We will probably need some kind of history as well, but for now we can do with the WAL from postgres if we ever badly need to recover history

app/backend/src/couchers/models.py
ShmuelTreiger: I'm having trouble understanding how the relationship works/should work. Specifically, both blocking_user_id and blocked_user_id reference the same model.field and I couldn't figure out how that works.

aapeliv: The relationships allow you to do things like

```py3
block = session.query(UserBlock).filter(...).one()

if block.blocking_user.username == "itsi":
  raise Exception("No one can block Itsi")
```

Where sqlalchemy will do the automagic to get the blocking user's username from the db.



> Specifically, both blocking_user_id and blocked_user_id reference the same model.field and I couldn't figure out how that works.

You need to specify a `foreign_keys` value as in the snippet above, so SA knows which row to use

ShmuelTreiger:
I implemented one blocking filter. I can't figure out how to do it without two queries. Would love feedback/insight on this.

ShmuelTreiger:
I got it working with one query. Let me know which you prefer.

ShmuelTreiger:
@aapeliv Let me know what you think of this solution, using relationships to grab relevant blocks and then filtering not in that list.

app/backend/src/couchers/models.py
```diff
@@ -269,6 +272,14 @@ def mutual_friends(self, target_id):
             .all()
         )
 
+    def blocked_blocking_users(self):
+        relevant_blocks = []
+        for userblock in self.is_blocking_user:
+            relevant_blocks.append(userblock.blocked_user_id)
+        for userblock in self.is_blocked_user:
+            relevant_blocks.append(userblock.blocking_user_id)
+        return relevant_blocks
```
aapeliv: What's the purpose of this function? Please document it thoroughly (through a doc string) and maybe see if you can come up with a better name! (It's just a bit confusing)

app/backend/src/couchers/models.py
```diff
@@ -1226,3 +1237,21 @@ def ready_for_retry(self):
...
+    """
+    Table of blocked users
+    """
+
+    __tablename__ = "user_blocks"
+    __table_args__ = (UniqueConstraint("blocking_user_id", "blocked_user_id"),)
+
+    id = Column(BigInteger, primary_key=True)
+
+    blocking_user_id = Column(ForeignKey("users.id"), nullable=False)
```
aapeliv, line 1270: `blocker`? Is that a better word? `blocker` and `blocked` user?

app/backend/src/couchers/servicers/api.py
```diff
@@ -423,12 +423,17 @@ def ListFriends(self, request, context):
         users2 = aliased(User)
 
         with session_scope() as session:
+            user = get_user_by_field(session, str(context.user_id))
```
aapeliv: Please don't use this function except in the few places where we take user input (login/user's page).

There's a standard `user = session.query(...).one()` query

aapeliv (changes requested):
This is quite incomplete, I think.

I think the blocking needs to happen through a join across the blocks table. Maybe an outer join which then filters to make sure the user in question has not blocked the relevant user.

I'm somewhat worried about the performance impact, but I suppose it's something we have to deal with.

Let's discuss blocking during the Wednesday backend meeting?

app/backend/src/couchers/models.py
ShmuelTreiger: I'll add a comment, but basically, for the most part, this should return everyone involved in a block with you so that we can filter for them (i.e. session.query(Users).filter(request.user is not in context.user.blocked_blocking_users)

app/backend/src/couchers/models.py
ShmuelTreiger: Happy to change it, but marginal at best. I'm not sure what the correct answer is

ShmuelTreiger: We can ask around at the meeting

ShmuelTreiger:
Would a join be faster than using relationships? Relationships allow something of an elegant solution as opposed to what I had before (take a look at previous commit). Sure, we can discuss tomorrow.

aapeliv:
> Would a join be faster than using relationships? Relationships allow something of an elegant solution as opposed to what I had before (take a look at previous commit). Sure, we can discuss tomorrow.

Relationships are just joins, they're SA magic for joins. Might be marginally slower to parse and construct the query or something, but that's nothing to worry about. The big worry is to build something that scales reasonably with more users, and for that the SA query building doesn't matter.

app/backend/src/couchers/models.py
ShmuelTreiger: We forgot to ask

app/backend/src/couchers/servicers/references.py
```diff
@@ -58,19 +58,24 @@ def ListReferences(self, request, context):
 
             to_users = aliased(User)
             from_users = aliased(User)
+            # relevant_blocks = all_blocked_or_blocking_users(context.user_id)
```
ShmuelTreiger, line 61: Wasn't sure if you should be able to see reviews from blocked/blocking users so you can know what they're saying about you (and others)

app/backend/src/couchers/db.py
```diff
@@ -283,3 +284,17 @@ def can_moderate_node(session, user_id, node_id):
...
+def all_blocked_or_blocking_users(user_id):
+    with session_scope() as session:
+        relevant_user_blocks = (
+            session.query(UserBlocks)
+            .filter(or_(UserBlocks.blocking_user_id == user_id, UserBlocks.blocked_user_id == user_id))
+            .all()
+        )
+
+        return [
+            user_block.blocking_user_id if user_block.blocking_user_id != user_id else user_block.blocked_user_id
```
aapeliv, line 298: `user_block.blocking_user_id or user_block.blocked_user_id`

app/backend/src/couchers/db.py
```diff
@@ -283,3 +284,17 @@ def can_moderate_node(session, user_id, node_id):
     return _can_moderate_any_cluster(
         session, user_id, [cluster.id for _, _, _, cluster in get_node_parents_recursively(session, node_id)]
     )
+
+
+def all_blocked_or_blocking_users(user_id):
```
aapeliv, line 289: Needs a better name, e.g. `blocked_users_for(user_id)`

aapeliv (commented):
This is still way too verbose and manual to have at every query site. As discussed in the meeting, it needs to be some simple SQL alchemy magic. Something like `.filter(User.is_visible_for(request.user_id))`, where that can then do some fancy stuff

app/backend/src/couchers/db.py
ShmuelTreiger: I don't understand

app/backend/src/couchers/db.py
ShmuelTreiger: It's also blocking users (i.e. users blocking them). It can be relevant_user_blocks, but that's less informative.

ShmuelTreiger: Or do you just want me to add for? all_blocked_or_blocking_users_for(user_id)

ShmuelTreiger:
>.filter(User.is_visible_for(request.user_id))

This would mean running a separate query on each user in a search (i.e. every user on couchers). I don't see how to make it feasible. Maybe someone else can if they want to take over.

You could move `all_blocked_or_blocking_users(context.user_id)` into each query:

`.filter(~User.id.in_(all_blocked_or_blocking_users(context.user_id)))`

But that would mean running a separate query for each filter, instead of once to get all the blocked and blocking users (2-4 times as often)

ShmuelTreiger:
Also needs migrations, but also going to hold off for now

aapeliv:
> > .filter(User.is_visible_for(request.user_id))
> 
> This would mean running a separate query on each user in a search (i.e. every user on couchers).

What do you mean?

ShmuelTreiger:
If you add a function on the user model, that means each user is going to individually query Userblocks (or you might be able to get away with a join)*. I've modeled three options for redesigning the queries in #1163. Lmk what you think. 

*You might be able to use a cache, but that seems suboptimal

PR #725, opened 2021-03-01
the rest of about me edit profile page, fix gender and pronoun fill-in

andreaowu:
Screen Shot 2021-03-01 at 11 25 48 AM
Screen Shot 2021-03-01 at 11 35 18 AM
Screen Shot 2021-03-01 at 11 35 26 AM

### DIFF
--- a/app/frontend/src/features/constants.ts
+++ b/app/frontend/src/features/constants.ts
@@ -45,13 +45,13 @@ export const SECTION_LABELS = {
 export const COUNTRIES_VISITED = "Countries I've Visited";
 export const COUNTRIES_LIVED = "Countries I've Lived In";
 export const FEMALE = "Female";
-export const FEMALE_FORM_VALUE = "female";
+export const FEMALE_PRONOUNS = "she / her";
 export const GENDER = "Gender";
 export const HOSTING_STATUS = "Hosting status";
 export const LANGUAGES_SPOKEN = "Languages I speak";
 export const MALE = "Male";
-export const MALE_FORM_VALUE = "male";
-export const OTHER_FORM_VALUE = "other";
+export const MALE_PRONOUNS = "he / him";
+export const PRONOUNS = "Pronouns";
 export const SAVE = "save";
 
 // Home
--- a/app/frontend/src/features/profile/edit/ProfileForm.tsx
+++ b/app/frontend/src/features/profile/edit/ProfileForm.tsx
@@ -60,12 +60,6 @@ const useStyles = makeStyles((theme) => ({
     "& > .MuiInputBase-root": {
       width: "100%",
     },
-    "& .MuiOutlinedInput-notchedOutline": {
-      borderColor: theme.palette.grey[500],
-    },
-    "& .MuiOutlinedInput-root:hover .MuiOutlinedInput-notchedOutline": {
-      borderColor: theme.palette.grey[900],
-    },
     [theme.breakpoints.up("md")]: {
       "& > .MuiInputBase-root": {
         width: 400,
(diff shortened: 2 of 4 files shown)


app/frontend/src/features/constants.ts
```diff
@@ -42,16 +42,21 @@ export const SECTION_LABELS = {
 };
 
 // Edit Profile
+export const AGE = "Age";
+export const AGE_MAX = 100;
+export const AGE_MIN = 21;
```
lucaslcode: Unused?

app/frontend/src/features/constants.ts
```diff
@@ -42,16 +42,21 @@ export const SECTION_LABELS = {
...
 export const COUNTRIES_VISITED = "Countries I've Visited";
 export const COUNTRIES_LIVED = "Countries I've Lived In";
 export const FEMALE = "Female";
 export const FEMALE_FORM_VALUE = "female";
+export const FEMALE_PRONOUNS = "she / her";
```
lucaslcode: These are man/woman on the signup page, I thought that's more gender-related.
In any case, I think the value should be capitalized, so it appears that way on the profile.

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -56,6 +62,12 @@ const useStyles = makeStyles((theme) => ({
     "& > .MuiInputBase-root": {
       width: "100%",
     },
+    "& .MuiOutlinedInput-notchedOutline": {
+      borderColor: theme.palette.grey[500],
+    },
+    "& .MuiOutlinedInput-root:hover .MuiOutlinedInput-notchedOutline": {
+      borderColor: theme.palette.grey[900],
+    },
```
lucaslcode: Is this to match the hifi more closely? I'm not sure it's 100% necessary, but if we want this, I think it should be in mui theme overrides or in a component - unless these colours shouldn't apply on other pages

lucaslcode (commented):
I'm sorry I didn't notice in the previous PRs, but the overview section is displaying blanks. I think maybe there it's okay to omit blank ones, since age, gender and "coucher since" will always be shown? Also, brackets for pronouns shouldn't appear if there are no pronouns. Feel free to leave those for another pr though if you think it's too unrelated.

Other stuff just minor, if you don't want to change anything, request again and I'll approve

Edit: And fluent languages are being displayed with "," in between instead of ", "

app/frontend/src/features/constants.ts
andreaowu: I feel like woman / man is weird...I never see those as gender categories on any sites. Why did we pick woman / man over female / male?

app/frontend/src/features/profile/edit/ProfileForm.tsx
andreaowu: This is currently in [TextField](https://github.com/Couchers-org/couchers/blob/24e81c4757e6079907c27713a2c78167b667b3fd/app/frontend/src/components/TextField.tsx#L17-L22), but reversed. `TextField` is used a bunch throughout the UI so I didn't want to change it and mess things up

andreaowu:
re: omitting blank ones - people's profiles aren't going to look uniform, which I don't really like. I think if people don't fill it in, we can just replace it with "Ask me" or just leave it blank. 

I'll fix the brackets and languages.

app/frontend/src/features/profile/edit/ProfileForm.tsx
lucaslcode: I think if this is the correct way, then change them all! :D

app/frontend/src/features/constants.ts
lucaslcode: Because I am a fool

lucaslcode (dismissed):
No need to squash commits, it can be good to see the process and makes reviewing a bit easier

app/frontend/src/features/constants.ts
andreaowu: LOL no you're not!

lucaslcode (approved):

PR #726, opened 2021-03-02
Profile references filter

darrenvong:
I dipped my hands into the backend and made some small changes, mainly so the reference object returns an ID so I can use it for the React key when rendering the list of references in a list. None of the existing properties were unique so this was the best option.

~~The only thing missing that I'll need to wrap up is probably add some frontend tests.~~ Now there are tests for them!

All references:
Screenshot 2021-03-05 at 03 25 57

No references available for a specific type:
Screenshot 2021-03-02 at 02 54 20

References for a specific type (from hosts only in this case):
Screenshot 2021-03-05 at 03 27 20

Mobile view:
Screenshot 2021-03-05 at 03 37 39


app/frontend/src/features/profile/view/useReferences.ts
```diff
@@ -0,0 +1,54 @@
...
+        service.user.getReferencesReceived({
+          count: 50,
+          userId: user.userId,
+          offset: 0,
+        }),
+        service.user.getReferencesGiven({
+          count: 50,
+          userId: user.userId,
+          offset: 0,
+        }),
```
darrenvong: I thought for now, rather than having "all" needing to combine the two reference type lists (given/received), then have "given" and "received" as two separate queries (so the filter becomes a thing to control which query in the cache to flick between), it's easier to combine them and do the filtering on the frontend.

Not feeling 100% about this though, the main limitation being doing pagination with this might be a pain... but maybe not with `useInfiniteQuery` eventually as the combined results themselves can be its own page?

app/backend/src/couchers/servicers/api.py
```diff
@@ -646,12 +646,13 @@ def InitiateMediaUpload(self, request, context):
 
 def paginate_references_result(request, query):
     total_matches = query.count()
-    references = query.order_by(Reference.time).offset(request.start_at).limit(request.number).all()
+    references = query.order_by(Reference.time.desc()).offset(request.start_at).limit(request.number).all()
```
darrenvong, line 651: This since we probably want the latest reference written to come back first?

andreaowu:
Can we align the photo with name + bar? I think it'd look better if vertically, the space between the name + bar was vertically in the middle of the photo and the name and bar are equidistant to this invisible middle vertical line (does that make sense? I can try to mock what I mean if it doesn't)

darrenvong:
Do you mean something like this? (Obviously minus the blue overlay, that was just a dev tool thing to make the box easier to see)
Screenshot 2021-03-03 at 03 27 11

Might be some CSS fiddling needed to get it to work as I'm not sure why the basic flexbox tricks aren't working 😅

app/frontend/.storybook/main.js
```diff
@@ -6,10 +6,7 @@ module.exports = {
     "@storybook/preset-create-react-app",
   ],
   webpackFinal: (config) => {
-    config.resolve.alias["../../service$"] = require.resolve(
-      "../src/stories/__mocks__/service.ts"
-    );
-    config.resolve.alias["../../../service$"] = require.resolve(
+    config.resolve.alias["service/index"] = require.resolve(
```
darrenvong, line 9: Fixes the broken Storybook mocks from absolute imports 😬

aapeliv (dismissed):
Backend looks ok

app/frontend/src/features/profile/view/References.tsx
```diff
@@ -1,149 +1,286 @@
...
+                key={reference.referenceId}
+                isReceived
+                user={userToShow}
+                reference={reference}
+              />
+            ) : null;
+          })}
+        </List>
+      )}
+      {referencesGiven && referencesGiven.totalMatches > 0 && (
```
lucaslcode: Needs an empty state

app/frontend/src/features/profile/constants.ts
```diff
@@ -62,3 +62,5 @@ export const parkingDetailsLabels = {
 export default function booleanConversion(value: boolean | undefined) {
   return value === undefined ? UNSURE : value ? "Yes" : "No";
 }
+
+export const referencesQueryStaleTime = 10 * 60 * 1000;
```
lucaslcode, line 66: Why is this different to global stale time?

app/frontend/src/components/UserSummary.tsx
```diff
@@ -19,7 +19,14 @@ export const useStyles = makeStyles((theme) => ({
     width: "100%",
   },
   title: {
-    marginBottom: theme.spacing(1),
+    marginTop: 0,
+  },
+  titleAndBarContainer: {
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "space-around",
```
lucaslcode, line 26: Re andrea's comment, I think to achieve what she is talking about, leave this as `space-around` but in the `root` class remove `alignItems`, which means it will go back to the default of `stretch` and the spacing will be balanced.

lucaslcode (dismissed):
minor

app/frontend/src/features/profile/constants.ts
darrenvong: If they are like those Couchsurfing, references aren't meant to be editable after they've been left right? That means they are reasonably stable data imo and so I thought it wouldn't make sense to refetch them that frequently. Probably should set the cache time value as the same too.

There is no stale time set explicitly so by default, it goes stale as soon as it hits the cache

app/frontend/src/components/UserSummary.tsx
darrenvong: That didn't make any difference to what I got now. I'm unsure if she meant what I got in the screenshot with the dev tool overlay or what I got now 🤔  @andreaowu ?

lucaslcode (approved):

PR #727, opened 2021-03-02
change relative paths to absolute paths

andreaowu:


lucaslcode:
I agree we should use absolute paths, but this makes it difficult to distinguish between modules and our code. Is there anything we can do about that? Does something like `import MessagesPage from "/features/messages/MessagesPage"` (with a / at the start) work?

andreaowu:
Hmm I don't think so?

lucaslcode:
How about this: https://stackoverflow.com/questions/42749973/es6-import-using-at-sign-in-path-in-a-vue-js-project-using-webpack

And/or using the [custom groups](https://stackoverflow.com/questions/42749973/es6-import-using-at-sign-in-path-in-a-vue-js-project-using-webpack) function of `simple-import-sort`

That way we can have internal files starting with @ like `import x from "@/features/x"` And we could have them in a separate group.

Anyway I'm happy to give that a try in a separate pr

darrenvong:
I think this would be quite tricky without either hacking/overriding the webpack config from Create React App (CRA), or eject from CRA so we have full control of the webpack config, as I believe it only supports having `src` as the `basePath` set by default.

lucaslcode:
We could do the group then, at least.
Anyway, build is broken here because the mocks aren't working on the tests, I couldn't figure out why...

darrenvong:
Uhhhh I couldn't find a good solution without ones that mention needing to modify jest's `moduleDirectories` config ([like this one](https://stackoverflow.com/questions/50171412/jest-typescript-absolute-paths-baseurl-gives-error-cannot-find-module)), which CRA doesn't support configuring so we _might_ need to hack/override some config inside it if we want to keep this. Unless you can get the `moduleNameMapper` working, because apparently that config can be [changed/supported by CRA](https://create-react-app.dev/docs/running-tests/#configuration). I didn't have much success when I tried though...

lucaslcode:
Try this adding to `yarn test` command: https://blog.shovonhasan.com/using-absolute-imports-with-jest-in-create-react-app/

andreaowu:
I added `--modulePaths=src` to it, I'm not sure why the tests are failing either :(

andreaowu:
I'm going to break this up into smaller PR's, don't know where the test is failing.

andreaowu:
*facepalm it's because a bunch of these try to `import {service} from 'service'` instead of `service/index`

lucaslcode (approved):
Super weird that fixed it, may be a bug in the jest mocks resolution

PR #730, opened 2021-03-03
absolute path for features/auth directory

andreaowu:


andreaowu:
Found the problem and fixed the big PR, so let's just go with that

PR #734, opened 2021-03-04
Backend/feature/num replies top level thread

### LINKED ISSUE #665: No way to get number of replies for top-level threads
For example, the `Discussion` message has a `thread_id` and replies to the discussion have `num_replies` themselves, but there's no way to the the number of replies on the base thread.

OmerBaddour:
Works and created a test for verification.

### DIFF
--- a/app/backend/src/couchers/servicers/threads.py
+++ b/app/backend/src/couchers/servicers/threads.py
@@ -13,7 +13,7 @@ from pb import threads_pb2, threads_pb2_grpc
 logger = logging.getLogger(__name__)
 
 # Since the API exposes a single ID space regardless of nesting level,
-# we construct the API id by appending the the nesting level to the
+# we construct the API id by appending the nesting level to the
 # database ID.
 
 
--- a/app/backend/src/couchers/servicers/threads.py
+++ b/app/backend/src/couchers/servicers/threads.py
@@ -29,6 +29,7 @@ def unpack_thread_id(thread_id: int) -> (int, int):
 class Threads(threads_pb2_grpc.ThreadsServicer):
     def GetThread(self, request, context):
         database_id, depth = unpack_thread_id(request.thread_id)
+        request.page_size = request.page_size if request.page_size < 100000 else None  # upper bound
         page_size = request.page_size or 1000
         page_start = unpack_thread_id(int(request.page_token))[0] if request.page_token else 2 ** 50
 
--- a/app/backend/src/couchers/servicers/threads.py
+++ b/app/backend/src/couchers/servicers/threads.py
@@ -58,15 +58,18 @@ class Threads(threads_pb2_grpc.ThreadsServicer):
                     )
                     for r, n in res[:page_size]
                 ]
+
+                # query without .limit() to get the total number of responses
                 all_res = (
                     session.query(Comment, func.count(Reply.id))
                     .outerjoin(Reply, Reply.comment_id == Comment.id)
                     .filter(Comment.thread_id == database_id)
                     .group_by(Comment.id)
-                    .order_by(Comment.created.desc())
                     .all()
                 )
-                num_responses = len(all_res)
+                num_responses = 0
+                for r in all_res:
+                    num_responses += 1 + r[1]
 
             elif depth == 1:
                 if not session.query(Comment).filter(Comment.id == database_id).one_or_none():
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/threads.py
```diff
@@ -58,6 +59,18 @@ def GetThread(self, request, context):
...
+                all_res = (
+                    session.query(Comment, func.count(Reply.id))
+                    .outerjoin(Reply, Reply.comment_id == Comment.id)
+                    .filter(Comment.thread_id == database_id)
+                    .group_by(Comment.id)
+                    .all()
+                )
+                num_responses = 0
+                for r in all_res:
+                    num_responses += 1 + r[1]
```
aapeliv: Please compute this within postgres, that'll speed it up a lot, and also you won't need to ship the actual data from the db. Something like

```py3
num_responses = (session.query(func.sum(1 + func.count(Reply.id))
  .outerjoin(Reply, Reply.comment_id == Comment.id)
  .filter(Comment.thread_id == database_id)
  .scalar())
```

This is completely untested and I'm not sure about the sum there, as well as the `.scalar()` (I'm pretty sure that's used somewhere in our codebase, `grep` for it, or just use `.one()[0]`)

app/backend/src/couchers/servicers/threads.py
```diff
@@ -80,6 +93,8 @@ def GetThread(self, request, context):
                     )
                     for r in res[:page_size]
                 ]
+                all_res = session.query(Reply).filter(Reply.comment_id == database_id).all()
+                num_responses = len(all_res)
```
aapeliv: You can just compute `func.count(Reply)` (I'm not sure if `func.count(Reply.id)` is faster or not?) to compute this in the db and not have to send all reply objects into the python backend.

app/pb/threads.proto
```diff
@@ -54,6 +54,9 @@ message GetThreadRes {
   // A token that can be sent as `page_token` to retrieve the next page.
   // If this field is omitted, there are no subsequent pages.
   string next_page_token = 2;
+
+  // number of responses (comments and replies) associated with the thread
+  uint32 num_responses = 3;
```
aapeliv: Please move this above `next_page_token`: for consistency, it would be best for that to be last. You can either renumber of leave this as `3`

aapeliv (changes requested):

app/backend/src/couchers/servicers/threads.py
```diff
@@ -29,6 +29,7 @@ def unpack_thread_id(thread_id: int) -> (int, int):
 class Threads(threads_pb2_grpc.ThreadsServicer):
     def GetThread(self, request, context):
         database_id, depth = unpack_thread_id(request.thread_id)
+        request.page_size = request.page_size if request.page_size < 100000 else None  # upper bound
```
kalvdans: This line has no effect, remove. There's no point in changing the request structure, as it is only inputs to the function. Use local variables instead.

app/backend/src/tests/test_threads.py
```diff
@@ -142,3 +142,38 @@ def test_threads_pagination(db):
...
+            api.PostReply(threads_pb2.PostReplyReq(thread_id=cat_id, content=animal)).thread_id
+            for animal in ["cheetah", "lynx", "panther"]
+        ]
+
+        # test num_responses
+        ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=PARENT_THREAD_ID))
+        assert ret.num_responses == 9
+
+        ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=dog_id))
+        assert ret.num_responses == 3
```
kalvdans, line 179: It would be useful to pass page_size=1 here to check that num_replies could be queried without actually downloading all the replies.

app/pb/threads.proto
```diff
@@ -51,9 +51,12 @@ message GetThreadRes {
   // sorted in inverse time order, latest reply first.
   repeated Reply replies = 1;
 
+  // number of responses (comments and replies) associated with the thread
+  uint32 num_responses = 2;
+
```
kalvdans, line 56: It would be useful to return num_responses in the `Discussions.GetDiscussion()` and `Pages.GetPage()` RPC calls, where the client first get ahold of a thread_id. This could be added in a later PR of course.

aapeliv: I think it's good to keep the threads stuff isolated/modular in its own part. The frontend currently wouldn't use that, they just have a component for "put this given thread (by ID) here".

kalvdans (approved):
LGTM!

aapeliv (approved):

PR #735, opened 2021-03-04
Desktop font sizes

lucaslcode:
Gave up on MUI `responsiveFontSizes`, it only works >1rem


darrenvong:
> it only works >1rem

Is that because they multiply the value, so anything below 1 when you multiply gets smaller? 😬

darrenvong (commented):
It looks like design has also added a `lineHeight` of 1.5 (150%, but the unitless multiplier is preferred) on all of them too, so we should also include that?

lucaslcode:
Fonts of less than 1rem are just skipped...

darrenvong (approved):

PR #736, opened 2021-03-04
Increasing test coverage in test_api

ShmuelTreiger:
Are we aware that, in order to use api.GetUser(user_id) you need to cast user_id to a string? Is this desired behavior?


aapeliv (approved):
Yes, it's known

PR #737, opened 2021-03-05
Community page part the third

lucaslcode:


app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -177,79 +160,57 @@ export default function CommunityPage() {
...
-            disabled
-          >
-            <CouchIcon />
-          </CircularIconButton>
-        </div>
-      </div>
+      <div className={classes.body}>
+        <Hidden smDown>
+          <NavDesktop community={community} />
+        </Hidden>
```
lucaslcode: On a scale of 1-10, how terrible is this idea?

lucaslcode: @darrenvong @andreaowu

darrenvong: I have mixed feels about that helper... but I think it's a balancing act. If this is the only thing that differs then I think it's ok.

Although if we start having them sprinkled all over our layouts, then at that point we may as well have two versions which I believe we'd want to avoid right now and try to make sure the layout work across all sizes as much as possible.

app/frontend/src/routes.ts
```diff
@@ -67,27 +67,18 @@ export const routeToEvent = (id: number, slug: string) =>
...
-  `${communityEventsBaseRoute}/${id}/${slug}`;
+const communityTabs = [
+  "overview",
+  "find-host",
+  "events",
+  "local-points",
+  "discussions",
+  "hangouts",
+] as const;
+export type CommunityTab = typeof communityTabs[number];
```
lucaslcode: `as const`, my new favourite TS magic. That lets `CommunityTab` be a union of those strings instead of just `string`!

lucaslcode:
I think this can be merged and split into a few issues to continue.

darrenvong:
What are the main changes in this part?

app/frontend/src/routes.ts
darrenvong: Ahh yeah, `as const` is probably the one exception where it's mostly fine/good to use a type assertion to help narrow the type.

Although it doesn't guarantee the array will remain constant, as it is still a type assertion so you could still do `communityTabs.push(...)` and things would break. But hopefully that'd be weird enough (as there wouldn't be a good reason to do that and then use `as const`) that if anyone tries to do that, we'd pick that up in PR reviews...

app/frontend/src/routes.ts
```diff
@@ -67,27 +67,18 @@ export const routeToEvent = (id: number, slug: string) =>
...
+export const communityRoute = `${communityBaseRoute}/:communityId/:communitySlug/:page?`;
+export const routeToCommunity = (
+  id: number,
+  slug: string,
+  page?: CommunityTab
+) => `${communityBaseRoute}/${id}/${slug}` + (page ? `/${page}` : "");
```
```suggestion
) => `${communityBaseRoute}/${id}/${slug}${page ? `/${page}` : ""}`;
```

app/frontend/src/routes.ts
darrenvong: Hmm although if you are only using this to generate a union type... couldn't you just write that literally instead? I.e. 
```typescript
export type CommunityTab =
  | "overview"
  | "find-host"
  | "events"
  | "local-points"
  | "discussions"
  | "hangouts";
```

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
```diff
@@ -1,64 +0,0 @@
-import {
```
darrenvong, line 1: How come we don't need this component anymore? I'm guessing this isn't being used outside of this WIP community page?

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -161,6 +138,10 @@ export default function CommunityPage() {
     }
   }, [community, communitySlug, history]);
 
+  const { page = "overview" } = useParams<{ page: CommunityTab }>();
```
darrenvong:
```suggestion
  const { page = "overview" } = useParams<{ page: string }>();
```
Since this comes from the URL so it can be anything, hence the pattern below with `tab` to narrow it to the right typed string?

app/frontend/src/features/messages/Messages.tsx
```diff
@@ -63,7 +63,7 @@ type MessageType = keyof typeof labels;
 
 export default function Messages() {
   const history = useHistory();
-  const { type = "chats" } = useParams<{ type: keyof typeof labels }>();
+  const { type = "chats" } = useParams<{ type: MessageType }>();
```
darrenvong:
```suggestion
  const { type = "chats" } = useParams<{ type: string }>();
```
Same comment as in community page.

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
```diff
@@ -2,18 +2,48 @@ import {
...
+import eventImagePlaceholder from "./eventImagePlaceholder.svg";
 
 const useStyles = makeStyles((theme) => ({
+  link: { textDecoration: "none", color: "inherit" },
+  image: {
+    backgroundColor: theme.palette.grey[200],
+    height: 80,
+    objectFit: "contain",
+    [theme.breakpoints.up("sm")]: {
+      height: 100,
```
darrenvong, line 28: Hmm might be better to use % for these height so it scales with the card's height?

app/frontend/src/theme.ts
```diff
@@ -26,6 +29,14 @@ declare module "@material-ui/core/styles/shape" {
 const themeOptions: ThemeOptions = {
   breakpoints,
   overrides: {
+    MuiCssBaseline: {
+      "@global": {
+        a: {
```
darrenvong, line 34: Do we have any default link styles from design yet? It'd be good to include here too if they exist?

app/frontend/src/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -15,21 +17,31 @@ import stripMarkdown from "utils/stripMarkdown";
 import placeImagePlaceholder from "./placeImagePlaceholder.svg";
 
 const useStyles = makeStyles((theme) => ({
+  link: { textDecoration: "none", color: "inherit" },
   image: {
     backgroundColor: theme.palette.grey[200],
     height: 80,
     objectFit: "contain",
+    [theme.breakpoints.up("sm")]: {
```
darrenvong, line 24: Same comment on the hardcoded `height` as in event card

darrenvong (changes requested):
Yeah this is definitely a good breaking point for further PRs 👍

app/frontend/src/theme.ts
lucaslcode: I think it's better to put them in Mui <Link> or something - this is just a css reset

app/frontend/src/routes.ts
lucaslcode: lol, I was using the array for something but I must have refactored it out...

app/frontend/src/theme.ts
darrenvong: Good point, yeah better to put in `MuiLink` instead

app/frontend/src/features/communities/CommunityPage/CircularIconButton.tsx
lucaslcode: Yes, the design changed so it's no longer used

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
lucaslcode: That would be equivalent to "auto" because the parent doesn't have specified height

app/frontend/src/features/constants.ts
```diff
@@ -101,6 +103,23 @@ export const WHEELCHAIR = "Wheelchair accessible";
...
+  given: "Given to others",
+};
+export const referenceBadgeLabel = {
+  [ReferenceType.REFERENCE_TYPE_FRIEND]: "Friend",
+  [ReferenceType.REFERENCE_TYPE_HOSTED]: "Guest",
+  [ReferenceType.REFERENCE_TYPE_SURFED]: "Hosted",
+};
+export const NO_REFERENCES = "No references of this kind yet!";
+export const getReferencesGivenHeading = (name: string) =>
+  `References ${name} wrote`;
```
darrenvong: The merge probably just merged the old stuff with this branch, but this is not needed as I've moved these into `features/profile/constants.ts`

app/frontend/src/features/communities/CommunityPage/EventCard.tsx
darrenvong: Uhh the pain of height always having to be explicitly set so you can't read its current value and base its % on that... I guess this will become a decorative image eventually so it's probably not as big of a deal.

darrenvong (approved):

PR #739, opened 2021-03-05
fix: sanitizing name/email before submitting it to backend

### LINKED ISSUE #258: Invalid username
Subject: invalid username
Description:
Whatever username i try to sign up with it just says invalid username

Steps:
Sign up as a new user

Results:
Says invalid username for any username i try

Backend version: bf2ff87
Frontend version: bf2ff87
User Agent: Mozilla/5.0 (Android 6.0.1; Mobile; rv:68.0) Gecko/68.0 Firefox/68.0
Page: https://app.couchers.org/signup/Ft_mArU2qVsbbXPAMtVMb4o8qWncLBNky3QsxwQkeO4=
User ID: 0

Tay0108:
Closes https://github.com/Couchers-org/couchers/issues/258

Let me know if there are other places we should do this. Also this function is used for sanitizing both email and login, couldn't come up with a better name, feel free to suggest sth better ;) 

Do we display username somewhere? Does it make sense to display it the way user entered it during sign up (capitalized or not)?

### DIFF
--- a/app/frontend/src/features/auth/email/ChangeEmailPage.tsx
+++ b/app/frontend/src/features/auth/email/ChangeEmailPage.tsx
@@ -3,6 +3,7 @@ import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { Error as GrpcError } from "grpc-web";
 import { useForm } from "react-hook-form";
 import { useMutation } from "react-query";
+import { sanitizeName } from "utils/validation";
 
 import Alert from "../../../components/Alert";
 import Button from "../../../components/Button";
@@ -29,7 +30,8 @@ export default function ChangeEmailPage() {
     reset: resetForm,
   } = useForm<ChangeEmailFormData>();
   const onSubmit = handleSubmit(({ currentPassword, newEmail }) => {
-    changeEmail({ currentPassword, newEmail });
+    const sanitizedEmail = sanitizeName(newEmail);
+    changeEmail({ currentPassword, newEmail: sanitizedEmail });
   });
 
   const {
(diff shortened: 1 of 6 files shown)


app/frontend/src/features/auth/email/ChangeEmailPage.tsx
```diff
@@ -29,7 +30,8 @@ export default function ChangeEmailPage() {
     reset: resetForm,
   } = useForm<ChangeEmailFormData>();
   const onSubmit = handleSubmit(({ currentPassword, newEmail }) => {
```
darrenvong, line 32: Would be nice to update and add to the existing tests on this page to check it's the sanitised input that's being called.

darrenvong (changes requested):
I think apart from the URL, personally not aware of the username being displayed anywhere else. Is that right @lucaslcode ?

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
Tay0108: I created 2 new tests for password and without password forms. If you'd rather have me edit existing tests let me know ;)

lucaslcode (approved):

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
darrenvong: Haha more _useful_ tests are always welcomed, but don't want to blow the scope of this too much for ones that don't have tests yet. This tests the helper as well which I much prefer than unit testing it in this case as it's only a thin wrapper around built-in functionality so we know that should work on its own already.

darrenvong (approved):

PR #740, opened 2021-03-06
chore: add sort-object-keys lint rule

darrenvong:
And fix all the warnings zzzzzzzzz. This was probably the most brutal and boring chore ever...

### DIFF
--- a/app/frontend/package.json
+++ b/app/frontend/package.json
@@ -43,16 +43,8 @@
     ],
     "rules": {
       "simple-import-sort/imports": "warn",
-      "simple-import-sort/exports": "error",
+      "simple-import-sort/exports": "warn",
       "sort-imports": "off",
-      "sort-keys": [
-        "warn",
-        "asc",
-        {
-          "caseSensitive": false,
-          "natural": true
-        }
-      ],
       "import/order": "off",
       "import/first": "error",
       "import/newline-after-import": "error",
--- a/app/frontend/package.json
+++ b/app/frontend/package.json
@@ -42,9 +42,17 @@
       "simple-import-sort"
     ],
     "rules": {
-      "simple-import-sort/imports": "error",
+      "simple-import-sort/imports": "warn",
       "simple-import-sort/exports": "error",
       "sort-imports": "off",
+      "sort-keys": [
+        "warn",
+        "asc",
+        {
+          "caseSensitive": false,
+          "natural": true
+        }
+      ],
       "import/order": "off",
       "import/first": "error",
       "import/newline-after-import": "error",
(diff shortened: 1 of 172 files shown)


app/frontend/src/components/Map.tsx
```diff
@@ -10,21 +10,21 @@ const URL = process.env.REACT_APP_API_BASE_URL;
...
     position: "absolute",
     top: 0,
-    left: 0,
     width: "100%",
-    height: "100%",
+  },
+  root: {
+    height: 200,
+    position: "relative",
+    width: 400,
```
lucaslcode, line 29: Moving this here has made 400px width take priority over 100% and `grow` no longer works. Also I'm not sure I like these keys being alphabetical, usually I order them as they appear in the page. What do you think?

darrenvong: Ahh crap should have thought of this, I wonder how many things this breaks because the ordering of these matters for CSS. I don't have a problem with ordering by component position. Also I think you mentioned something similar in Slack already, but having `height` and `width` so far apart (imagine there are more properties in between than only `position` in this case) also feels a bit odd to me as they'd be considered related properties to me. Maybe this isn't a good idea and we should just try our best to order it where makes sense...

lucaslcode: On the other hand, we probably should rely on specificity more than ordering. But yeah the lint rule might be a bit un-subtle?

darrenvong: Depends on what you mean by specificity, but I think if we need to use overly specific selectors, it kinda defeats using CSS-in-JS solution a bit, which is supposed to create unique classes so that we shouldn't need highly specific selectors often.

Yeah, unfortunately it's a bit of an all or nothing as it can't be configured to skip checking top level keys in an object. The CSS is basically where most of the warnings come from.

darrenvong: Although something like:
```typescript
{
  grow: {
    "div&": { ... }
  }
}
```
probably isn't that bad if we want to keep this? 😅  The styling is still scoped to this component after all...

darrenvong:
Ok got rid of the sort-keys rule but kept the re-ordering work. Fixed the broken `Map` grow thing and from a glance across Storybook, that seems to be the only thing broken so 🤞  nothing extra is screwed up because of this

app/frontend/package.json
```diff
@@ -42,8 +42,8 @@
       "simple-import-sort"
     ],
     "rules": {
-      "simple-import-sort/imports": "error",
-      "simple-import-sort/exports": "error",
+      "simple-import-sort/imports": "warn",
+      "simple-import-sort/exports": "warn",
```
darrenvong, line 46: Changed these to `warn` as it's annoying they fail builds in dev when you know you'll go back and fix this later/before you commit. They still fail on CI so it's just as good as error on build time

lucaslcode (approved):

PR #742, opened 2021-03-07
remove edit name / gender from profile editing and add it to settings page

andreaowu:
Screen Shot 2021-03-09 at 6 09 10 PM
Screen Shot 2021-03-09 at 7 18 31 PM


app/frontend/src/features/auth/email/ChangeEmailPage.tsx
```diff
@@ -55,7 +63,9 @@ export default function ChangeEmailPage() {
 
   return (
     <>
-      <PageTitle>Change email</PageTitle>
+      <PageTitle>{CHANGE_NAME_GENDER}</PageTitle>
```
lucaslcode: ~`CHANGE_EMAIL`? I think the test is failing here - maybe will pass if you use "Change email"~ See below comment about new page.

app/frontend/src/features/auth/constants.tsx
```diff
@@ -1,5 +1,12 @@
 export const ABOUT_US = "About us";
+export const CHANGE_NAME_GENDER = "Change Name / Gender";
+export const CHANGE_PASSWORD = "Change Password";
+export const CHECK_EMAIL =
+  "Your email change has been received. Check your new email to complete the change.";
 export const COLLABORATIVE = "Built by the community and always free";
+export const CONTACT =
+  "Admins will be happy to change your name and/or gender label. Please contact couchersorg@gmail.com!";
```
lucaslcode:
```suggestion
  "Admins will be happy to change your name and/or gender label. Please contact contact@couchers.org!";
```

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -213,101 +196,6 @@ export default function EditProfileForm() {
...
-                );
-              }}
-            />
-            <ProfileTextInput
-              label={HOMETOWN}
-              name="hometown"
-              defaultValue={user.hometown}
-              inputRef={register}
-              className={classes.field}
-            />
```
lucaslcode: Keep this, it's not in the signup

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -213,101 +196,6 @@ export default function EditProfileForm() {
...
-                      />
-                    </RadioGroup>
-                  </>
-                );
-              }}
-            />
-            <Controller
-              control={control}
-              defaultValue={user.pronouns}
-              name="pronouns"
```
lucaslcode: I think we need to keep this, it's not in the signup. Maybe it should be free-form only?

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
```diff
@@ -55,7 +63,9 @@ export default function ChangeEmailPage() {
...
+      <Typography variant="body1">{CONTACT}</Typography>
+      <PageTitle>{CHANGE_PASSWORD}</PageTitle>
```
lucaslcode: Password change is a different page. Maybe there needs to be a new page "Account settings" And put all three forms (name/gender contact info, email change, password page) on?

app/frontend/src/features/profile/edit/ProfileForm.tsx
```diff
@@ -121,15 +113,6 @@ export default function EditProfileForm() {
...
       {user ? (
         <>
-          <form onSubmit={onSubmit}>
-            <ProfileTextInput
-              label="Name"
-              name="name"
-              defaultValue={user.name}
-              inputRef={register}
-              className={classes.field}
-            />
```
lucaslcode: Maybe we should keep this, or allow it to be changed on the other page. There is/will be a separate field for  "legal name" used for verification purposes.

lucaslcode (dismissed):
Removed fields will need to be removed from `updateProfile()` in `service/user.ts` or they will be set to a blank string when editing another part of the profile.

The `updateProfile` api call takes a bunch of `StringValue`s - this is because protobuf strings are `""` by default. Wrapping the string in a StringValue allows the backend to know if that field was untouched vs. set to a blank string on purpose.

However, we aren't using this capability in the frontend atm, we just set a predefined list of fields. This is also happening in the [pending profile pic PR](https://github.com/Couchers-org/couchers/pull/703#discussion_r586028666).

Also we will need to prevent the backend accepting gender changes - I can do that when this is ready to merge.

app/frontend/src/features/profile/edit/ProfileForm.tsx
andreaowu: Should we call this "display name" or something instead?

app/frontend/src/features/profile/edit/ProfileForm.tsx
andreaowu: I think she / her and he / him are standard enough for those to be options, and the last one can be free form?

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
andreaowu: That's what I tried to make this page for

andreaowu:
The only field that ended up getting removed is gender, so I guess we just need to prevent BE from accepting gender changes?

app/frontend/src/features/auth/email/ChangeEmailPage.tsx
lucaslcode: This file is `ChangeEmailPage.tsx` but it has a section about changing name and gender, and a section titled "Change Password" followed by a form for changing your email address :D

andreaowu: Oops :D

app/frontend/src/routes.ts
```diff
@@ -3,8 +3,7 @@ export const baseRoute = "/";
 export const loginRoute = "/login";
 export const loginPasswordRoute = `${loginRoute}/password`;
 export const resetPasswordRoute = "/password-reset";
-export const changePasswordRoute = "/change-password";
-export const changeEmailRoute = "/change-email";
+export const settingsRoute = "/change-email";
```
lucaslcode: "account-settings" but I will do this now when I make the backend change

lucaslcode (approved):
Approved pending successful build

PR #744, opened 2021-03-07
Events

aapeliv:
This is now ready for review, hoping to get it merged in.

After that, the following (big chunks) will be in later PRs:

* Proper timezone handling
* Notifications
* Corresponding `ListEvents` in communities, groups
* Search functionality
* Automatically repeating events (currently need to schedule each instance manually) — not so urgent

It was too big to do in one PR.


app/backend/src/couchers/errors.py
```diff
@@ -27,6 +27,17 @@
 DATE_TO_AFTER_ONE_YEAR = "You cannot request to stay with someone for longer than one year."
 DIRECT_MESSAGE_ONLY_FRIENDS = "You can only send a message to friends."
 DISCUSSION_NOT_FOUND = "Discussion not found."
+EVENT_CANT_REMOVE_OWNER_AS_ORGANIZER = "Transfer the event elsewhere before removing yourself as an organizer."
+EVENT_CANT_UPDATE_ALL_TIMES = "Sorry, you cannot update all events if you're modifying start or end times."
+EVENT_CANT_UPDATE_OLD_EVENT = "You can't update an event that's more than 1 day old."
+EVENT_EDIT_PERMISSION_DENIED = "You're not allowed to edit that event."
+EVENT_ENDS_BEFORE_STARTS = "The event must end after it starts"
+EVENT_IN_PAST = "The event must be in the future"
```
kalvdans: Some of the added error strings does not end with a dot.

app/backend/src/couchers/models.py
```diff
@@ -1029,30 +1037,116 @@ class ClusterEventAssociation(Base):
...
+        "Cluster",
+        backref=backref("owned_events", lazy="dynamic"),
+        uselist=False,
+        foreign_keys="Event.owner_cluster_id",
+    )
+
+    __table_args__ = (
+        # Only one of owner_user and owner_cluster should be set
+        CheckConstraint(
+            "(owner_user_id IS NULL AND owner_cluster_id IS NOT NULL) OR (owner_user_id IS NOT NULL AND owner_cluster_id IS NULL)",
```
kalvdans: Can be written shorter `((owner_user_id ISNULL) != (owner_cluster_id ISNULL))` (untested)

app/backend/src/couchers/models.py
```diff
@@ -1029,30 +1037,116 @@ class ClusterEventAssociation(Base):
...
+    )
+
+
+class EventOccurence(Base):
+    __tablename__ = "event_occurences"
+
+    id = Column(BigInteger, communities_seq, primary_key=True, server_default=communities_seq.next_value())
+    event_id = Column(ForeignKey("events.id"), nullable=False, index=True)
+
+    creator_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
```
kalvdans, line 1112: How is creator_user_id different from Event.creator_user_id ? Please write a comment.

app/backend/src/couchers/models.py
```diff
@@ -1029,30 +1037,116 @@ class ClusterEventAssociation(Base):
...
+    photo_key = Column(ForeignKey("uploads.key"), nullable=True)
+
+    # a null geom is an online-only event
+    geom = Column(Geometry(geometry_type="POINT", srid=4326), nullable=True)
+    # physical address, iff geom is not null
+    address = Column(String, nullable=True)
+    # videoconferencing link, etc, must be specified if no geom, otherwise optional
+    link = Column(String, nullable=True)
+
+    timezone = "Etc/UTC"
```
kalvdans, line 1123: capitalize constants to not confuse them with a Column name.

app/pb/events.proto
```diff
@@ -0,0 +1,304 @@
...
+
+  rpc RemoveEventOrganizer(RemoveEventOrganizerReq) returns (google.protobuf.Empty) {
+    // Remove yourself as an event organizer
+  }
+
+  rpc ListMyEvents(ListMyEventsReq) returns (ListMyEventsRes) {
+    // Lists all subscribed, attending
+  }
+
+  // TODO: DownloadCalendarInvite
```
kalvdans, line 68: Client-side

app/pb/events.proto
```diff
@@ -0,0 +1,304 @@
...
+enum AttendanceState {
+  ATTENDANCE_STATE_NOT_GOING = 0;
+  ATTENDANCE_STATE_MAYBE = 1;
+  ATTENDANCE_STATE_GOING = 2;
+}
+
+message Event {
+  int64 event_id = 1; // = occurence_id
+  // it's the next event occurence
+  bool is_next = 2;
```
kalvdans: Please define "next". Is it only in my current polygon and does already started occurrences count?

app/pb/events.proto
```diff
@@ -0,0 +1,304 @@
...
+  // already happened
+  bool is_past = 3;
+  // in the future
+  bool is_future = 4;
```
kalvdans: Redundant, client can calculate these from start_date and end_date.

app/backend/src/tests/test_events.py
```diff
@@ -0,0 +1,1726 @@
...
+        assert res.is_future
+        assert res.title == "Dummy Title"
+        assert res.slug == "dummy-title"
+        assert res.content == "Dummy content."
+        assert not res.photo_url
+        assert res.WhichOneof("mode") == "offline_information"
+        assert res.offline_information.lat == 0.1
+        assert res.offline_information.lng == 0.2
+        assert res.offline_information.address == "Near Null Island"
+        assert time_before < to_aware_datetime(res.created) < now()
```
kalvdans: `<=`

app/backend/src/tests/test_events.py
```diff
@@ -0,0 +1,1726 @@
...
+        assert time_before < to_aware_datetime(res.last_edited) < now()
```
kalvdans: `<=` to not fail on a fast computer with 100 Hz clock accuracy.

app/backend/src/tests/test_events.py
```diff
@@ -0,0 +1,1726 @@
...
+                    link="https://app.couchers.org/meet/",
+                ),
+                parent_community_id=c_id,
+                start_time=Timestamp_from_datetime(start_time),
+                end_time=Timestamp_from_datetime(end_time),
+                timezone="UTC",
+            )
+        )
+
+        assert res.is_next
```
kalvdans, line 179: Can two simultaneous events both be "next"?

app/pb/events.proto
```diff
@@ -0,0 +1,304 @@
...
+  google.protobuf.Timestamp created = 13;
+  google.protobuf.Timestamp last_edited = 14;
+  int64 creator_user_id = 15;
+
+  // the UTC timestamps of start and end times
+  google.protobuf.Timestamp start_time = 16;
+  google.protobuf.Timestamp end_time = 17;
+
+  // the tzdata timezone identified
+  string timezone = 18;
```
kalvdans: identified? I think this is the timezone decided when creating the event.

kalvdans (changes requested):
Looks solid in general, some minor comments only.

app/pb/events.proto
aapeliv: It's a typo, should be "identifier", thanks for spotting

app/backend/src/couchers/models.py
aapeliv: Idea is to turn this into a column in next pr without too much hassle

app/pb/events.proto
aapeliv: Thanks, great catch

app/backend/src/couchers/models.py
aapeliv: I think as it's written now, it's clearer.

(We have the same type of constraints in several places, I think.)

Too bad pg doesn't have an XOR operator

app/backend/src/tests/test_events.py
aapeliv: No. I've thought about it, and I think occurences of the same event should not overlap.

aapeliv:
@kalvdans: to clarify, there are events (`Event`) and occurences (`EventOccurence`) in the backend representation. An event is something like "Weekly NYC Couchers meetup", and an occurence is an instance of that, e.g. the time this Friday that the NYC meetup goes to meet up. The purpose for having these two parts is to that users can subscribe to one "stream" of events (through `EventSubscription`) and be notified (in future PR) of new occurences of an event. So if I'm in the NYC Couchers meetup, I want to get an email every week telling me where we're meeting up and at what time, who's going (`EventOccurenceAttendance`, separately from who is subscribed), etc.

Please let me know if this is not clear in comments (I guess it's not: sorry). Where should I document this? I can add it in something like `docs/events.md` or in models? protos (but this is backend implementation details)? I'm not sure.

The frontends/API layer just has one "event" object, and the way to get info about other related events is through `ListEventOccurences`. That's why there's `is_next`.

Regarding `is_next`: I think the right way is to add an exclusion constraint to make sure there is only one event in a given stream at a time, no overlaps.

Thoughts?

app/pb/events.proto
aapeliv: My idea was to generate the file (an icalendar, `.ics`) on the backend and just send that to the frontends. It's a standard format for describing events, and I think much better to do that in the backend instead of having to replicate it to each frontend implementation. There will probably also be some details that are just easier to do in the backend (e.g. tz support/conversions).

app/backend/src/couchers/models.py
kalvdans: `!=` is xor when applied to boolean values.

app/pb/events.proto
kalvdans: Or we can provide a javascript function to generate an ics file from GetEvent response, that all frontends can use.

aapeliv: Not all frontends will be written in javascript.

I am also hoping to have better integration with calendar software eventually. It's possible to add a URL that some software will update the event from, so we could potentially push new scheduled occurences directly to the user's calendar. This is not something we can do client side.

Also, client side icalendar handling libraries seem to be pretty crappy compared to https://pypi.org/project/icalendar/

kalvdans: I disagree but both ways have their merits, so resolving conversation :)

app/backend/src/couchers/models.py
kalvdans: there will be hassle anyway but resolving conversation.

kalvdans:
> Thoughts?

I'm all for having a lot of comments in the .proto files, like your story with "Weekly NYC Couchers meetup". So the documentation for is_next is very much good now but can be clarified that it is always true for non-recurring events.

Implementation details, like having two tables (one for event series and one for event occurrances) should go into `models.py` in my opinion.

aapeliv:
I've changed this to use a tstzrange (timestamp with timezone range) to represent the time during which the event takes place, and then used that to make sure there is only one occurence at a time. This fixes up the is_next bit.

I also cleaned up all existing constraints.

aapeliv:
Ready for re-review @kalvdans

app/pb/events.proto
```diff
@@ -0,0 +1,301 @@
...
+
+service Events {
+  // This is a secure service: a user needs to be authenticated and not jailed to call functions here, refer to auth.proto and jail.proto
+
+  rpc CreateEvent(CreateEventReq) returns (Event) {
+    // Creates an event, only one instance, add more with ScheduleEvent
+  }
+
+  rpc ScheduleEvent(ScheduleEventReq) returns (Event) {
+    // Schedules a new occurence of an existing event
```
kalvdans: occurrence is spelled with two r (here and more places)

aapeliv: Wow... Yeah, I completely didn't notice. Let's hope find & replace suffices!

app/backend/src/couchers/migrations/versions/c035cdcabd40_add_new_events_tables.py
```diff
@@ -75,7 +75,7 @@ def upgrade():
         op.f("ix_cluster_event_associations_event_id"), "cluster_event_associations", ["event_id"], unique=False
     )
     op.create_table(
-        "event_occurences",
+        "event_occurrences",
```
kalvdans, line 78: Don't touch old migrations! (your own words :) )

kalvdans (changes requested):
Aww, the spelling checks made it hard to see the actual changes. Anyway the is_next documentation is clearer now, thanks. I think it looks ok besides some new issues introduced by the spelling corrections.

We might live with wrongly spelled tables and fix that separately later.

app/backend/src/couchers/migrations/versions/c035cdcabd40_add_new_events_tables.py
aapeliv: This is a new migration in this PR so it's fine. But yes, shouldn't touch migrations already on `develop` (or a bit more leniently, those that have been applied in prod)

aapeliv:
> I think it looks ok besides some new issues introduced by the spelling corrections.

I think it's all fine. This is the first PR to do events (there were some existing tables but no data/code that used them).

So there shouldn't be any remnants of wrong spelling. (There is one occurrence of "occurence" in the migrations to do with `slugify`, and I didn't change that, even though it just seems to be a comment).

app/backend/src/couchers/migrations/versions/c035cdcabd40_add_new_events_tables.py
kalvdans: My mistake looking at the last commit only 😊

kalvdans (approved):

Issue #753, opened 2021-03-08
404 shouldn't change the url


lucaslcode:
404 page currently redirects, but better behaviour would be keeping the incorrect URL, but showing the 404 page.

vrouilhac:
Hello, I look at the code and I think I can solve this. In your routes files, you only need to remove the path props on the last route which is the 404 route

lucaslcode:
Quite right! If you're keen, feel free to make a PR (:

vrouilhac:
> Quite right! If you're keen, feel free to make a PR (:

Thank you :pray:

darrenvong:
Closed by #765

PR #765, opened 2021-03-09
404 Redirect while keeping current URL

vrouilhac:
@lucaslcode 

From issue : #753 

I removed the 'exact path' props from 404 Route component and as it is the last it act as a all other routes grabber and still show the 404 page while keeping the URL user entered. 

I then removed the redirect component which became useless with this configuration and after checking for references I also removed 'notFoundRoute export on routes.ts.

### DIFF
--- a/app/frontend/src/AppRoutes.tsx
+++ b/app/frontend/src/AppRoutes.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Redirect, Switch } from "react-router-dom";
+import { Switch } from "react-router-dom";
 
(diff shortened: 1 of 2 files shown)


lucaslcode (approved):
Thanks! If you'd like to work on anything else, have a look at the other issues (:

vrouilhac:
> Thanks! If you'd like to work on anything else, have a look at the other issues (:

Thank you, actually, I'd love to contribute to this project. I don't know much about backend with Python so I'll focus on React. Good work from your team on this project!

PR #766, opened 2021-03-09
code organization, navbar overhaul, add friend button

### LINKED ISSUE #754: Navigation tweaks
On mobile, search is hidden and bug report button is inaccessible off the screen

### LINKED ISSUE #759: Profile action buttons
Friend Request - should only be visible when appropriate (eg. send friend request button only if they aren't friends already).

### LINKED ISSUE #760: Account settings page
- A page to hold the links/forms for changing password, email, gender.
- Remove those links from the bottom of profile page
- Link to this page from somewhere that makes sense

andreaowu:
#754 #760 #759 

Screen Shot 2021-03-09 at 9 02 41 AM
Screen Shot 2021-03-09 at 9 03 22 AM
Screen Shot 2021-03-09 at 9 03 30 AM

(I keep pushing onto the wrong branch so that's why this PR came with three different fixes, oops)

**Friends**
My profile
Screen Shot 2021-03-11 at 8 24 39 PM

"Add friend" button clicked, and then:
Screen Shot 2021-03-11 at 8 24 47 PM

Did not try to add friend:
Screen Shot 2021-03-11 at 8 24 54 PM


app/frontend/src/components/Navigation/NavButton.tsx
```diff
@@ -22,17 +18,38 @@ const useStyles = makeStyles((theme) => ({
...
+  label: {
+    alignSelf: "center",
+    marginTop: 0,
+  },
 }));
 
+interface NavButtonProps {
+  route: string;
+  label: string;
+  labelVariant:
```
lucaslcode: Can this variant type be imported from mui?

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -1,68 +1,60 @@
...
+  {
     name: "Messages",
     route: messagesRoute,
   },
   {
-    icon: <LocationIcon fontSize="inherit" />,
-    name: "Map",
-    route: mapRoute,
+    name: "Communities",
+    route: communityRoute,
```
lucaslcode: Should we leave this as the map until communities are in?

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -76,32 +68,77 @@ const useStyles = makeStyles((theme) => ({
...
 
 export default function Navigation() {
   const classes = useStyles();
-
   const authenticated = useAuthContext().authState.authenticated;
+  const [open, setOpen] = React.useState(false);
+
+  const logoutButton = <NavButton route={logoutRoute} label={LOG_OUT} />;
+
+  const drawer = (
```
lucaslcode:
```suggestion
  const drawerItems = (
```
Since it's not the whole drawer

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -119,38 +156,67 @@ export default function Navigation() {
...
+              aria-label="open drawer"
+              onClick={handleDrawerOpen}
+              edge="start"
+            >
+              <MenuIcon />
+            </IconButton>
+            <Drawer
+              variant="temporary"
+              anchor="left"
+              open={open}
```
lucaslcode, line 174: this needs a `handleClose` prop so clicking the overlay closes the drawer

app/frontend/src/features/profile/functions/useUpdateHostingPreferences.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { renderHook } from "@testing-library/react-hooks";
-import useUpdateHostingPreferences from "features/profile/useUpdateHostingPreferences";
+import useUpdateHostingPreferences from "features/profile/functions/useUpdateHostingPreferences";
```
lucaslcode: thoughts on "hooks" folder name instead of "functions"? I think "hooks" is more common, but it should be obvious they are hooks because of the name "use*"

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -119,38 +156,67 @@ export default function Navigation() {
...
-          classes={{
-            root: classes.flex,
-          }}
-        >
-          {menu.map((item) => (
-            <NavButton
-              route={item.route}
-              label={item.name}
-              key={`${item.name}-nav-button`}
+          <Hidden mdDown>
```
lucaslcode:
```suggestion
          <Hidden smDown>
```

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -119,38 +156,67 @@ export default function Navigation() {
...
+            <Grid
```
lucaslcode: This `<Grid container>` has no `<Grid item>` children, so maybe it can just be a div?

lucaslcode (changes requested):
Was there a verdict in slack in the end? We can discuss it at the frontend meeting too

app/frontend/src/components/Navigation/NavButton.tsx
andreaowu: I couldn't find what type it is :(

app/frontend/src/components/Navigation/Navigation.tsx
andreaowu: We're going to have to change it later, so might as well just change it now?

lucaslcode: Yes, but if we do the soft beta thing, it might still not be in yet

app/frontend/src/components/Navigation/NavButton.tsx
lucaslcode: What is the type of the `variant` prop of `Typography`?

andreaowu: Dunno, I just put what I saw on the [docs](https://material-ui.com/api/typography/#props)

lucaslcode (dismissed):
Sorry for not picking this up before, but there needs to be top padding or content is covered by the header on mobile (use/change `theme.shape.navPaddingMobile`). This only happens on some pages for some reason (dashboard and profile)

app/frontend/src/components/Navigation/NavButton.tsx
darrenvong: `Exclude<TypographyProps['variant'], undefined>` should do it. The utility type is needed because `variant` is optional on `TypographyProps` and we probably don't want `undefined` to be a possible value here.

app/frontend/src/AppRoute.tsx
```diff
@@ -17,6 +17,7 @@ export const useStyles = makeStyles((theme) => ({
       paddingTop: theme.shape.navPaddingDesktop,
     },
     paddingBottom: theme.shape.navPaddingMobile,
```
lucaslcode:
```suggestion
```

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -72,36 +63,82 @@ const useStyles = makeStyles((theme) => ({
...
+  const [open, setOpen] = React.useState(false);
+
+  const logoutButton = <NavButton route={logoutRoute} label={LOG_OUT} />;
+
+  const drawerItems = (
+    <div>
+      <List>
+        {menu.map(({ name, route }) => (
+          <ListItem button key={name}>
+            <NavButton route={route} label={name} labelVariant="h2" />
```
lucaslcode, line 127: I think the drawer needs to close on click. Maybe a `closeDrawer` prop for `NavButton`?

app/frontend/src/theme.ts
```diff
@@ -57,7 +57,7 @@ const themeOptions: ThemeOptions = {
   shape: {
     borderRadius,
     navPaddingDesktop: spacing(10),
-    navPaddingMobile: spacing(8),
+    navPaddingMobile: spacing(10),
```
lucaslcode: This should exactly be the navbar height (3.5), some pages may not want padding between the bar and content (for example, if there is a banner image)

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -75,12 +78,20 @@ export default function Overview({ user }: OverviewProps) {
...
-      {user.userId === currentUserId && (
+      {user.userId === currentUserId ? (
         <CardActions className={classes.cardActions}>
           <Button component={Link} to={editProfileRoute}>
             {EDIT_PROFILE}
           </Button>
         </CardActions>
+      ) : (
+        <CardActions className={classes.cardActions}>
+          {mutationError && <Alert severity="error">{mutationError}</Alert>}
```
lucaslcode: This should be outside `CardActions`, or it looks funny when it is present

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -75,12 +78,20 @@ export default function Overview({ user }: OverviewProps) {
...
+          <AddFriendButton
+            userId={user.userId}
+            setMutationError={setMutationError}
+          />
+        </CardActions>
```
lucaslcode: There should be a check here for `user.friendshipStatus`, showing this on `NOT_FRIENDS`, maybe a grayed out pending button for `PENDING` (actually maybe nothing for now, that will be quicker) and nothing for `FRIENDS`

lucaslcode (dismissed):
Last one sorry!

app/frontend/src/components/Navigation/NavButton.tsx
andreaowu: How did you find out what the type is?

darrenvong: VS Code shows you if you write out something like `type Variant = TypographyProps['variant']` and hover over where it says `TypographyProps['variant']`. Or you can click on `TypographyProps` and it should jump you into that file with the type definition

andreaowu: Sounds good, thanks!

app/frontend/src/theme.ts
andreaowu: Why do some pages not want padding between bar and content? (Which pages wouldn't want that?)

lucaslcode: Communities page wouldn't because of the banner image

andreaowu: It's bad design if we need to go to every page to add spacing due to one page not wanting that spacing?

lucaslcode: Fair, how about separating Nav bar height and Nav bar padding variables so in those instances, we can do a negative margin accurately (or is there a better way?)

lucaslcode: Fair, how about separating Nav bar height and Nav bar padding variables so in those instances, we can do a negative margin accurately (or is there a better way?)

lucaslcode (approved):

PR #769, opened 2021-03-10
Added functionality to clear profile picture

### LINKED ISSUE #731: No way to unset/clear profile picture
https://github.com/Couchers-org/couchers/blob/404cf8820d8afeb352838487f862d94fe7a9b711/app/backend/src/couchers/servicers/api.py#L212-L213

dnalhg:
Fixes #731 

Makes the user avatar nullable when updating a user profile which will allow the picture to be cleared.


aapeliv (approved):
Looks good!

PR #775, opened 2021-03-11
fix: hide send message/unfriend buttons on friend list + tests

### LINKED ISSUE #667: Unable to send message to friend
Subject: Unable to send message to friend
Description:
I am Unable to send message to friend from the "My Connections" page

Steps:
- access https://develop--frontend.preview.coucher.org/connections/friends/
- under "Your friends", click the mail button

Results:
Expected: a popup window opens, allowing me to write a message to my friend
Actual: nothing happens

Backend version: develop-869f44fb
Frontend version: develop-1e5907fb
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.182 Safari/537.36
Page: https://develop--frontend.preview.coucher.org/connections/friends/
User ID: 15

darrenvong:
And some general clean ups from hardcoded messages to constants as agreed.

Closes #667


app/frontend/src/features/auth/Logout.tsx
```diff
@@ -17,6 +18,8 @@ export default function Logout() {
       queryClient.resetQueries("hostRequestMessages");
       queryClient.resetQueries("hostRequests");
       queryClient.resetQueries("ping");
+      queryClient.resetQueries(friendRequestKey("Sent"));
+      queryClient.resetQueries(friendRequestKey("Received"));
```
lucaslcode: I think we should just clear absolutely everything. That might fix the offline cache bug too.

I mean, even the user cache shouldn't be left on a public computer (for example) after logging out, if users are only visible to couchers members

app/frontend/src/queryKeys.ts
```diff
@@ -40,3 +40,7 @@ export const referencesKey = (
   userId: number,
   type: "received" | "given" | "all"
 ) => ["references", { type, userId }];
+
+export type FriendRequestType = "Sent" | "Received";
+export const friendRequestKey = (type: FriendRequestType) =>
+  `friendRequests${type}`;
```
lucaslcode:
```suggestion
  ["friendRequests", type];
```
Second param could even be an object, since it's "filters".

app/frontend/src/reactQueryClient.tsx
```diff
@@ -22,13 +20,6 @@ export const queryClient = new QueryClient({
   },
 });
 
-const persistor = createLocalStoragePersistor();
-
-persistQueryClient({
-  maxAge: 14 * 24 * 60 * 60 * 1000,
-  persistor,
-  queryClient,
-});
```
lucaslcode: Did you mean to commit this?

app/frontend/src/test/serviceMockDefaults.ts
```diff
@@ -1,11 +1,12 @@
+import { User } from "pb/api_pb";
 import { HostRequestStatus } from "pb/conversations_pb";
 import messages from "test/fixtures/messages.json";
 import users from "test/fixtures/users.json";
 
 const userMap = new Map(users.map((user) => [user.userId, user]));
 
 export async function getUser(userId: string) {
-  return userMap.get(+userId);
+  return userMap.get(+userId) as User.AsObject;
```
lucaslcode: Why this change?

lucaslcode (changes requested):
Can you remove the all tab while you're in there?

app/frontend/src/reactQueryClient.tsx
lucaslcode: Oh I see your commit message. I will check out the plugin source

lucaslcode:
[Looking here](https://github.com/tannerlinsley/react-query/blob/master/src/createLocalStoragePersistor-experimental/index.ts), it seems the cache updating is throttle to only happen every seconds, so I guess it's a race condition?

As a workaround, maybe set the local storage key to a constant, and then directly remove that from local storage on logout?

https://github.com/tannerlinsley/react-query/issues/1951

app/frontend/src/reactQueryClient.tsx
lucaslcode: https://github.com/tannerlinsley/react-query/issues/1951

app/frontend/src/features/auth/Logout.tsx
darrenvong: Yeah that might make this more maintainable going forward tbh, otherwise it's easy to forget to remove a query here. I guess there's probably still value in having the cache while the user is logged in so that between page refreshes it'll still be faster than no cache

app/frontend/src/test/serviceMockDefaults.ts
darrenvong: It's because I've used the real return type of `getUser` and the fixtures don't actually have the full user objects (a lot of missing keys deemed required in the type). Maybe I'll just add in everything as in the default user and stop being lazy 😅  I imagine eventually once we have better test coverage, we'll need those properties on the user anyway. Since it's already in a separate JSON file, I don't see it being a problem (mainly bloating up the test file with properties you don't care about in a specific test suite if it were inlined)

darrenvong: Oh crap actually, might still need this or a non-null assertion since `userMap.get` returns `User.AsObject | undefined` but it should never be undefined in our case as we should always be using known userIds in our tests.

darrenvong:
The workaround of interacting with directly in logout didn't work :( I suspect this might also be a bottleneck on how quickly local storage can accept write to disk? 🤷

app/frontend/src/reactQueryClient.tsx
```diff
@@ -22,13 +20,6 @@ export const queryClient = new QueryClient({
   },
 });
 
-const persistor = createLocalStoragePersistor();
```
lucaslcode, line 25: Adding `throttleTime: 100` fixed the issue for me.

lucaslcode (changes requested):
Really don't want to get rid of offline cache sorry :P But this worked for me

darrenvong:
> Really don't want to get rid of offline cache sorry :P But this worked for me

Lol didn't know why I didn't try that first. Glad it was only a race condition and not some kinda disk write bottleneck! Fixed!

lucaslcode (approved):

PR #776, opened 2021-03-11
Adding documentation

ShmuelTreiger:


docs/database.md
```diff
@@ -20,13 +22,21 @@ popd
 # nuke the current database
 rm -rf ./data/postgres/pgdata
 
+# *********
+# Windows note:
+# If you're using WSL2 and want to run the tests locally (i.e. not in linux or docker),
+# run the next two commands in the windows command line, not in Linux, then switch back
+# *********
```
lucaslcode: I don't think this is necessarily true, they should be run in the `couchers/app` folder, wherever that is. Mayber there's a `popd` missing here.

ShmuelTreiger: I get a connection issue if I run those commands in Linux, but not when I run them in windows. But you're right, it's not absolute, I'll edit to be conditional.

lucaslcode (approved):

PR #779, opened 2021-03-12
fix: replace date entry fields with component from @material-ui/pickers

### LINKED ISSUE #495: Smart date entry form field

### LINKED ISSUE #751: Replace date entry fields with component from @material-ui/pickers
Relevant documentation:
- [@material-ui/pickers installation](https://material-ui-pickers.dev/getting-started/installation) (use the day.js version)
- [react-hook-forms](https://react-hook-form.com/get-started#Quickstart)

Rough steps:
- use `yarn add` to add necessary packages
- Replace the date `<TextField>`s with [`<KeyboardDatePicker>`](https://material-ui-pickers.dev/api/KeyboardDatePicker) in the following files:
  - [NewHostRequest.tsx](https://github.com/Couchers-org/couchers/blob/997e12a688b91b8c212279d0cef8bf3d8e7ceccb/app/frontend/src/features/messages/surfing/NewHostRequest.tsx#L83) (from and to dates)
  - [CompleteSignupForm.tsx](https://github.com/Couchers-org/couchers/blob/997e12a688b91b8c212279d0cef8bf3d8e7ceccb/app/frontend/src/features/auth/signup/CompleteSignupForm.tsx#L172)

Note: the `id`, `name`, and `inputRef` props should remain the same on the new component

angelika7:
fixes #751


andreaowu:
What do you think about making a common component for this so we don't have to copy and paste a bunch of properties for it everywhere?

Also, for labels, please put the string in `constants.ts`

app/frontend/package.json
```diff
@@ -3,10 +3,13 @@
   "version": "0.1.0",
```
lucaslcode, line 3: Please delete `package-lock.json`, it is used by `npm`, but we use `yarn` for everything. (Note, that's not this file, I just can't comment on that one)

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -40,6 +42,7 @@ type SignupInputs = {
   acceptTOS: boolean;
   hostingStatus: HostingStatus;
   location: ApproximateLocation;
+  meetingDT: string;
```
lucaslcode: There's no meeting field, delete this line and just use the birthdate field.

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,24 +172,41 @@ export default function CompleteSignup() {
...
-              required: "Enter your birthdate",
-              validate: (stringDate) =>
-                validatePastDate(stringDate) ||
-                "Must be a valid date in the past.",
-            })}
-            helperText={errors?.birthdate?.message}
-          />
+          <MuiPickersUtilsProvider utils={DateFnsUtils}>
+            <Controller
+              {...{ control }}
```
lucaslcode: This works, but we prefer to use `control={control}` because it's clearer you are only setting one prop.

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,24 +172,41 @@ export default function CompleteSignup() {
...
+              defaultValue={getValues().meetingDT}
```
lucaslcode: Looks like you're trying to set "dd.mm.rrrr" as a default, but it needs to be a valid date or date string. I would do this (see my other comment for an explanation): 
```suggestion
              defaultValue={new Date().toISOString().split("T")[0]}
```

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,24 +172,41 @@ export default function CompleteSignup() {
...
+                required: "Enter your birthdate",
+                validate: (stringDate) =>
+                  validatePastDate(stringDate) ||
+                  "Must be a valid date in the past.",
+              })}
+              
+              render={({ onChange, value }) => (
+                <KeyboardDatePicker
+                  className={authClasses.formField}
+                  error={errors.hasOwnProperty("meetingDT")}
```
lucaslcode: `errors.birthdate` is an object if there's an error, or undefined if there's no error. Objects are truthy, and undefined is falsey, but typescript requires exactly true or false for this prop. So you can do a little trick to convert it to true/false:
`error={!!errors.birthdate}`
If the error is an object, the first ! turns it into `false`, and the second ! turns it into `true`. [More info here](https://stackoverflow.com/questions/784929/what-is-the-not-not-operator-in-javascript)

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,24 +172,41 @@ export default function CompleteSignup() {
...
+                  helperText={errors?.birthdate?.message}
+                  id="birthdate"
+                  clearable
+                  format="DD.MM.YYYY"
+                  label={getValues().meetingDT}
```
lucaslcode:
```suggestion
                  label="Birthdate"
```

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
-            }}
-            inputRef={register({
-              required: "Enter a to date",
-              validate: dateValidate,
-            })}
-            helperText={formErrors?.toDate?.message}
-          />
+          <MuiPickersUtilsProvider utils={DateFnsUtils}>
+            <Controller
+              {...{ control }}
```
lucaslcode: Same as other comment

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+              defaultValue={"dd.mm.rrrr"}
```
lucaslcode: Same as other comment

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+              name="fromDate"
+              inputRef={register({
+                required: "Enter a from date",
+                validate: dateValidate,
+              })}
+              
+              render={({ onChange, value }) => (
+                <KeyboardDatePicker
+                  helperText={formErrors?.fromDate?.message}
```
lucaslcode: I guess you also need an `errors` prop like the other one

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+                    shrink: true,
+                  }}
+                />
+              )}
+            />
+          </MuiPickersUtilsProvider>
+          <MuiPickersUtilsProvider utils={DateFnsUtils}>
+            <Controller
+              {...{ control }}
+              defaultValue={"dd.mm.rrrr"}
```
lucaslcode: Same as other comments

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+              name="toDate"
+              inputRef={register({
+                required: "Enter a to date",
+                validate: dateValidate,
+              })}
+              
+              render={({ onChange, value }) => (
+                <KeyboardDatePicker
+                  helperText={formErrors?.toDate?.message}
```
lucaslcode: error prop

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+                  id="to-date"
+                  clearable
+                  format="DD.MM.YYYY"
+                  label="Date to"
+                  onChange={onChange}
+                  value={value}
+                  animateYearScrolling={true}
+                  fullWidth
+                  disableToolbar
+                  emptyLabel=''
```
lucaslcode: Maybe this is a good place to put the "DD.MM.YYYY" label? What happens if you omit this line?

darrenvong (commented):
We use `yarn` to install our packages so won't need a `package-lock.json`. Can you remove the file from the PR please?

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,24 +172,41 @@ export default function CompleteSignup() {
...
+                  onChange={onChange}
```
lucaslcode: My fault for not realising, but this isn't working because the birthdate field expects a string in the format "YYYY-MM-DD", but onChange gives it a Date() object.

Date object has a function `toIsoString()` which gives a string like "2021-03-12T23:00:00.000Z". You can get everything before the "T" by splitting the string on T, then taking the first element of the resulting array. [See this fiddle.](https://jsfiddle.net/qanfd57z/) Since this comes up a lot, I would create a very small function `date2DateString(date: Date)` that does this, and put it in `src/utils/date.ts`

But it gets even more complicated! Because the user might be just typing, so the date might not be valid, and if you try `toISOString()` on an invalid date, there will be an error. So you have to check for that.

The end result is this:
```suggestion
                  onChange={(date) => {
                    //user might be typing, so check the date is valid before doing conversions
                    date?.isValid()
                      ? onChange(date?.toISOString().split("T")[0])
                      : onChange(date);
                  }}
```

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+              })}
+              
+              render={({ onChange, value }) => (
+                <KeyboardDatePicker
+                  helperText={formErrors?.fromDate?.message}
+                  id="from-date"
+                  clearable
+                  format="DD.MM.YYYY"
+                  label="Date from"
+                  onChange={onChange}
```
lucaslcode: Same as other comment

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,34 +82,70 @@ export default function NewHostRequest() {
...
+              })}
+              
+              render={({ onChange, value }) => (
+                <KeyboardDatePicker
+                  helperText={formErrors?.toDate?.message}
+                  id="to-date"
+                  clearable
+                  format="DD.MM.YYYY"
+                  label="Date to"
+                  onChange={onChange}
```
lucaslcode: Same as other comment

lucaslcode (changes requested):
Looks like this isn't as straightforward as I thought! Sorry about that!

Andrea is right, it would be good to create a component for this, with everything inside and including `MuiPickersUtilsProvider`. To do that, create the file in the components folder. Maybe `DateField.tsx` or something like that. You can refer to `HeaderButton.tsx` as an example.

For props, I think it should accept `control: Control; name: string; inputRef: (ref: any) => void; className: string; id: string; error: string; label: string`. I think that's it?

You could also add the component to storybook, use `HeaderButton.stories.tsx` as an example.

app/frontend/src/components/DateField.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+interface DateFieldProps {
+  control: Control;
+  name: string;
+  inputRef: (ref: any) => void;
+  className: string;
+  id: string;
+  error: boolean;
+  label: string;
+  helperText: any
```
andreaowu: nit: can we alphabetize these? here and below

app/frontend/src/features/messages/constants.ts
```diff
@@ -9,3 +9,5 @@ export const hostRequestStatusLabels = {
 };
 
 export const MARK_LAST_SEEN_TIMEOUT = 500;
+export const DATE_FROM ="Date from"
+export const DATE_TO ="Date to";
```
andreaowu, line 13: nit: space after `=`

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -80,33 +86,31 @@ export default function NewHostRequest() {
...
-              shrink: true,
-            }}
+          <DateField 
+            control={control}
+            name="fromDate" 
             inputRef={register({
               required: "Enter a from date",
               validate: dateValidate,
             })}
+            className=''
```
andreaowu: is this needed? what about making `className` optional in the component?

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -169,23 +172,21 @@ export default function CompleteSignup() {
...
+          <DateField 
+            control={control}
+            name="birthdate" 
             inputRef={register({
-              required: "Enter your birthdate",
-              validate: (stringDate) =>
-                validatePastDate(stringDate) ||
-                "Must be a valid date in the past.",
-            })}
+                required: "Enter your birthdate",
```
andreaowu: should we make this string and line 182 constants?

app/frontend/src/components/DateField.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import DateField from "./DateField";
+
+export default {
+  component: DateField,
+  title: "Components/Simple/DateField",
+} as Meta;
+
+const Template: Story<any> = (args) => (
+    <>
```
andreaowu: is this necessary?

app/frontend/src/components/DateField.tsx
```diff
@@ -0,0 +1,63 @@
...
+            helperText={helperText}
+            id={id}
+            InputLabelProps={{
+              shrink: true,
+            }}
+            label={label}
+            onChange={(date) => {
+              //user might be typing, so check the date is valid before doing conversions
+              date?.isValid()
+                ? onChange(date?.format().split("T")[0])
```
lucaslcode: What happened when you tried `.toISOString()`?

app/frontend/package.json
```diff
@@ -3,14 +3,19 @@
...
     "@material-ui/lab": "^4.0.0-alpha.56",
+    "@material-ui/pickers": "3.2.10",
     "classnames": "^2.2.6",
+    "dayjs": "^1.10.4",
     "google-protobuf": "^3.15.3",
     "grpc-web": "^1.2.1",
+    "i": "^0.3.6",
     "maplibre-gl": "^1.13.0-rc.4",
     "markdown-it": "^12.0.2",
+    "npm": "^7.6.3",
```
lucaslcode: Remove npm and i, `yarn remove i npm`.

lucaslcode (changes requested):
Just some very tiny stuff, and also please run `yarn format`

app/frontend/src/components/DateField.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import * as React from "react";
+
+import DateField from "./DateField";
+
+export default {
+  component: DateField,
+  title: "Components/Simple/DateField",
+} as Meta;
+
+const Template: Story<any> = args => <DateField {...args} />;
```
lucaslcode: Datefield requires `control`, so you can putting a dummy form in here like `const { control } = useForm(...)`

lucaslcode: check if it works by running `yarn storybook`

app/frontend/src/components/DateField.tsx
angelika7: After clicking e.g 18.03.2021 I got different dates. First `console.log()`, when I used `.toISOString()`. Second with `.format()`. The second one was correct. It was connected with timezone. 
`2021-03-17T23:00:00.000Z`
`2021-03-18T00:00:00+01:00`

app/frontend/src/components/DateField.stories.tsx
```diff
@@ -8,10 +10,23 @@ export default {
...
+    <DateField
+      control={control}
+      error={false}
+      helperText=""
+      id="date-field"
+      inputRef={register({
+        required: "Enter your birthdate",
+        validate: (stringDate) =>
+          validatePastDate(stringDate) || "Must be a valid date in the past.",
+      })}
```
lucaslcode:
```suggestion
      inputRef={register}
```
and remove the unused import

lucaslcode (commented):
Last thing!

app/frontend/src/components/DateField.tsx
```diff
@@ -0,0 +1,65 @@
...
+        name={name}
+        render={({ onChange, value }) => (
+          <KeyboardDatePicker
+            animateYearScrolling={true}
+            autoOk
+            className={className}
+            clearable
+            disableToolbar
+            error={error}
+            format="DD.MM.YYYY"
```
andreaowu: Should / Could we change the format based on where the user is?

Also, we should have a `minDate`

app/frontend/src/components/DateField.tsx
```diff
@@ -0,0 +1,65 @@
...
+  control: Control;
+  error: boolean;
+  helperText: any;
+  id: string;
+  inputRef: (ref: any) => void;
+  label: string;
+  name: string;
+}
+
+export default function DateField({
```
andreaowu: nit: Can we call this `Datepicker` instead?

app/frontend/src/features/messages/constants.ts
```diff
@@ -9,3 +9,5 @@ export const hostRequestStatusLabels = {
 };
 
 export const MARK_LAST_SEEN_TIMEOUT = 500;
+export const DATE_FROM = "Date from";
+export const DATE_TO = "Date to";
```
andreaowu: nit: alphabetize these

app/frontend/src/components/DateField.tsx
```diff
@@ -0,0 +1,65 @@
...
+  className,
```
andreaowu: should `className` be optional?

app/frontend/src/components/Datepicker.tsx
```diff
@@ -0,0 +1,66 @@
...
+            error={error}
+            format="DD.MM.YYYY"
+            fullWidth
+            helperText={helperText}
+            id={id}
+            InputLabelProps={{
+              shrink: true,
+            }}
+            label={label}
+            minDate={new Date(1899, 12, 1)}
```
andreaowu: Can we make this an optional prop that defaults to today?

andreaowu:
Can you include a screenshot of what it looks like? 

Also, for comments that have been resolved, feel free to click `Resolve conversation`

app/frontend/src/components/DateField.tsx
andreaowu: Bump on: changing the format based on where the user is?

andreaowu:
Can we add some tests for Datepicker?

lucaslcode:
> Can we add some tests for Datepicker?

Let's open an issue for this - I don't want to drag out @angelika7 's first PR too long.

Add `minDate` as an optional prop for the component, and then let's merge!

lucaslcode (approved):

andreaowu:
No tests for a component in the name of "don't drag the PR on for too long" is a bad reason to not have tests. 

Opened issue #821 and assigned to @angelika7

PR #781, opened 2021-03-13
Include profile pic/remove username in connections page

### LINKED ISSUE #761: Connections page tweaks
Superset of/fixes #667 

- Make message button work
- Remove delete button
- Remove all tab (for now)
- Remove usernames
- Add profile pics

darrenvong:
Literally swapping what's on the friends list to re-use the existing `UserSummary` component for this. Achieves the "remove username" and "show profile photo" in one. Since there aren't any designs for this, I thought this is the easiest way to do this and also for maintaining visual consistency. Thoughts?

Closes #761

**Web**
Screenshot 2021-03-16 at 00 34 04

**Mobile Web**
Screenshot 2021-03-16 at 00 34 29

### DIFF
--- a/app/frontend/src/components/TabBar/TabBar.tsx
+++ b/app/frontend/src/components/TabBar/TabBar.tsx
@@ -1,12 +1,15 @@
-import { Tab, Tabs } from "@material-ui/core";
+import { Tab } from "@material-ui/core";
+import { TabList } from "@material-ui/lab";
 
 export interface TabBarProps<T extends Record<string, React.ReactNode>> {
-  value: keyof T;
-  setValue: (value: keyof T) => void;
+  ariaLabel: string;
   labels: T;
+  setValue: (value: keyof T) => void;
+  value: keyof T;
 }
 
 export default function TabBar<T extends Record<string, React.ReactNode>>({
+  ariaLabel,
   value,
   setValue,
   labels,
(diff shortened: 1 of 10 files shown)


andreaowu (dismissed):

andreaowu:
Looks good, can you include a screenshot?

app/frontend/src/components/TabBar/TabBar.tsx
```diff
@@ -15,7 +18,8 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
     setValue(newValue);
   };
   return (
-    <Tabs
+    <TabList
```
lucaslcode, line 21: What is the difference?

lucaslcode (commented):
The top text of `UserSummary` is H2 which isn't really right on this page...

app/frontend/src/components/TabBar/TabBar.tsx
darrenvong: This injects/handles the right aria props for accessibility according to the docs - https://material-ui.com/components/tabs/#experimental-api - so I thought why not use it. Otherwise I think it's the same as I believe `TabList` uses `Tabs`

darrenvong:
> Looks good, can you include a screenshot?

@andreaowu Updated original message with screenshots. I need to be better at doing this...

> The top text of `UserSummary` is H2 which isn't really right on this page...

Hmm good point. At least it's not skipping levels but doesn't exactly feel right either. I think I'll alter it to allow the heading level to be overriden from props, since it's impossible for the component on its own to know where it'll be used, which determines what heading level it should be at...

lucaslcode (approved):

Issue #762, opened 2021-03-08
Search results tweaks


lucaslcode:
- Fix text alignment with Profile Pic
- Add some more info
- Add empty state for "About me"

andreaowu:
Use [UserSummary](https://github.com/Couchers-org/couchers/blob/997e12a688b91b8c212279d0cef8bf3d8e7ceccb/app/frontend/src/components/UserSummary.tsx) to do this

PR #782, opened 2021-03-13
Feature? Add more information to search result card and refactored some code

justjooshing:
Added more details to the search cards and added UserDataLabels to reuse since we're using the label/text components like this in multiple areas.
Some questions about hosting/meetup label status, and whether they should be in `UserSummary` instead of children, so it sits more nicely at the top of the card, but I'm not sure where else we're using UserSummary and whether we want those there as well.
Also changed the position of `children` in `UserSummary` so the Avatar icon doesn't take up the whole vertical space (and therefore the children have the whole width available).


andreaowu:
Can you attach a screenshot of what this looks like?

Hosting / meetup label status should not be in `UserSummary`. `UserSummary` was designed in a way such that the last line can be switched out as necessary.

justjooshing:
Yep ok, that makes sense. I'm not sure how to better align the Avatar image in that case though.

Desktop and mobile tiles below
image
image

but it's lacking some API stuff from the design, and definitely needs a bit more love re colours, and the label statuses once we have the logos available for them.

Also at the moment the whole tile is clickable, whereas with the design there's a button for the user profile
image

Also the `name`, `age`, `location` in the `UserSummary` card is a bit busy sometimes, so I wonder whether we can display it differently or truncate it somehow?
image

justjooshing:
Also for this issue, sorry [#762](https://github.com/Couchers-org/couchers/issues/762)

app/frontend/.env.development
```diff
@@ -1,4 +1,4 @@
-REACT_APP_API_BASE_URL=http://localhost:8888
+REACT_APP_API_BASE_URL=https://dev-api.coucher.org
```
darrenvong: Since these are dev environment variables, they should be pointing to the local API rather than the dev preview API. Also some of us prefer to use a local copy of the API and I rather not have to change this back each time.

It's fine if you don't want to set up the Docker stuff, you can still change this locally on your machine but not commit this. With all that said, can you revert this change please?

app/frontend/src/components/UserSummary.tsx
```diff
@@ -42,31 +42,33 @@ interface UserSummaryProps {
...
+            <Typography variant="h2" className={classes.title}>
+              {!user ? <Skeleton /> : `${user.name}, ${user.age}, ${user.city}`}
+            </Typography>
+          }
+          secondary={
+            <>
+              <ScoreBar value={(user?.communityStanding || 0) * 100}>
+                {COMMUNITY_STANDING}
+              </ScoreBar>
+            </>
```
darrenvong:
```suggestion
              <ScoreBar value={(user?.communityStanding || 0) * 100}>
                {COMMUNITY_STANDING}
              </ScoreBar>
```
Unnecessary left over fragment if we're moving this out of here.

app/frontend/src/features/profile/view/About.tsx
```diff
@@ -25,27 +19,13 @@ export default function About({ user }: AboutProps) {
...
-            ? dateTimeFormatter.format(timestamp2Date(user.joined))
-            : ""
-        }
-      />
+      <UserAgeGenderPronouns user={user} />
+      <UserLanguages user={user} />
+      <UserHomeTown user={user} />
+      <UserOccupation user={user} />
+      <UserEducation user={user} />
+      <UserSinceJoined user={user} />
```
darrenvong: I'd leave it how it was before and made the changes inline here tbh. These components aren't doing anything extra and all this does is introduce an extra layer of indirection (now whoever reads this have to open another file just to trace what's going on) without much benefit imo.

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -110,15 +106,8 @@ export default function Overview({ user }: OverviewProps) {
...
-        <LabelAndText
-          label={LAST_ACTIVE}
-          text={
-            user.lastActive
-              ? `${timeAgo(timestamp2Date(user.lastActive))}`
-              : "Unknown"
-          }
-        />
+        <UserReferences user={user} />
+        <UserLastActive user={user} />
```
darrenvong: Same comment as in the `About` page.

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
...
         <CardActionArea>
           <CardContent>
-            <div className={classes.resultHeader}>
-              <Avatar user={user} />
-              <Container>
-                <Typography variant="h2">{user.name}</Typography>
-                <Typography variant="subtitle1">
-                  {hostingStatusLabels[user.hostingStatus]}
+            <UserSummary user={user}>
+              {/* I think this container would sit more nicely inside UserSummary, to align
```
darrenvong: There's a variation of this used in the host requests view that rather than displaying the hosting/meetup status, it displays information related to the host request instead, so `UserSummary` is meant to be the maximum common elements between all variations + `children` to allow for the minor variation, so no, please don't put this `Container` in `UserSummary`.

app/frontend/src/features/user/UserHeader.tsx
```diff
@@ -51,7 +51,7 @@ export default function UserHeader({ children, user }: UserHeaderProps) {
       {user.lastActive && (
         <Box mb={2}>
           <Typography component="p" variant="caption" gutterBottom>
-            {timeAgo(timestamp2Date(user.lastActive))}
+            {user.lastActive ? timeAgo(timestamp2Date(user.lastActive)) : undefined}
```
darrenvong:
```suggestion
            {user.lastActive && timeAgo(timestamp2Date(user.lastActive))}
```

app/frontend/src/features/user/UserDataLabels.tsx
```diff
@@ -0,0 +1,68 @@
+import LabelAndText from "components/LabelAndText";
```
darrenvong: Based on my comment on `About` and `Overview` pages... I'd just delete this file altogether because [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

darrenvong (changes requested):

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
...
+              >
+                <Typography
+                  className={classes.statusLabel}
+                  display="inline"
+                  variant="subtitle1"
+                  color="primary"
+                >
+                  {/* This "ask me about hosting/about meeting up" text can be changed when we add the icons in
+                      but at the moment just having Ask Me without a label is unclear" */}
+                  {hostingStatusLabels[user.hostingStatus] === "Ask me"
```
darrenvong:
```suggestion
                  {hostingStatusLabels[user.hostingStatus] === UNSURE
```
We're trying to not have loose strings around our code and trying to centralise them all in constants file so they can be found more easily/only need to change in one place if this message changes again etc.

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
...
+                    ? `${hostingStatusLabels[user.hostingStatus]} about hosting`
+                    : hostingStatusLabels[user.hostingStatus]}
+                </Typography>
+                <Typography
+                  className={classes.statusLabel}
+                  display="inline"
+                  variant="subtitle1"
+                  color="secondary"
+                >
+                  {meetupStatusLabels[user.meetupStatus] === "Ask me"
```
darrenvong:
```suggestion
                  {meetupStatusLabels[user.meetupStatus] === UNSURE
```

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
...
+                <Typography
+                  className={classes.statusLabel}
+                  display="inline"
+                  variant="subtitle1"
+                  color="primary"
+                >
+                  {/* This "ask me about hosting/about meeting up" text can be changed when we add the icons in
+                      but at the moment just having Ask Me without a label is unclear" */}
+                  {hostingStatusLabels[user.hostingStatus] === "Ask me"
+                    ? `${hostingStatusLabels[user.hostingStatus]} about hosting`
```
darrenvong: Can we pull this into a string helper function into `profile/constants` please (for the reasons explained as above)?

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
...
-            <ScoreBar value={user.verification * 100}>
-              {VERIFICATION_SCORE}
-            </ScoreBar>
-
-            {user.aboutMe.length < 300
-              ? user.aboutMe
-              : user.aboutMe.substring(0, 300) + "..."}
+              <Typography variant="h6" className={classes.about}>
+                {missingAbout
+                  ? `${firstName} hasn't said anything about themselves yet`
```
darrenvong: Same comment about pulling into `profile/constants` as above.

app/frontend/src/components/UserSummary.tsx
```diff
@@ -42,31 +42,33 @@ interface UserSummaryProps {
...
+          }
+        />
+      </div>
+      {children}
```
darrenvong: > Also changed the position of children in UserSummary so the Avatar icon doesn't take up the whole vertical space (and therefore the children have the whole width available)

Hmm not sure about this change if we are expecting anything in `children` to be dynamic. Might be better to [top align the avatar like how it's shown here](https://www.figma.com/file/z1aQUklCQewLsEFUo2HsLz/Couchers-Platform-Design?node-id=5156%3A45)? Ignore the fact the "moderator ..." text is above the bar in this design, this is where our dev "creative license" comes in to override their decision as supporting that would introduce inconsistencies/maintenance nightmares...

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -36,31 +51,61 @@ const useStyles = makeStyles((theme) => ({
 
 export default function SearchResult({ user }: { user: User.AsObject }) {
   const classes = useStyles();
+  const firstName = user.name.split(" ")[0];
+  const missingAbout = user.aboutMe.length === 0;
+  console.log(user);
```
darrenvong:
```suggestion
```

app/frontend/.env.development
justjooshing: Yep good catch

app/frontend/src/components/UserSummary.tsx
justjooshing: It's really because at the moment we have two components in our flexboxed .root class. 
So `ListItemAvatar` is taking up the whole left hand side, and if we keep `children` in `ListItemText` 's secondary component then it's restricted to the right hand side only. But I can move it back in no problemo if you think it's better?

image

app/frontend/src/features/profile/view/About.tsx
justjooshing: Yep that makes sense. Would there be any benefit in having some larger components (for say LastActive/References, or Language/Gender) as these are duplicated in Search Results and About or Overview. Or just leave them all as `LabelAndText`?

app/frontend/src/features/search/SearchResult.tsx
justjooshing: Cool thanks for the clarification, this matches your comment earlier about dev 'creative licence'.

app/frontend/src/features/search/SearchResult.tsx
justjooshing: Yep good suggestion

app/frontend/src/features/user/UserHeader.tsx
justjooshing: Not sure why I changed this, since we're already handling user.lastActive above. Reverting back

app/frontend/src/features/search/SearchResult.tsx
justjooshing: Good idea

app/frontend/src/features/profile/view/About.tsx
darrenvong: If they are styled the same as what we see in the user profile (which looks like that is the case from the screenshots), then sure, that sounds like a decent use case to me to make them separate components so they can be shared 👍

app/frontend/src/components/UserSummary.tsx
darrenvong: Yeah I think it's fine for the `ListItemAvatar` to take up the left hand side space, as we probably want the supplementary text/components to be aligned with whatever that's on the right. If we want a component to take up the full width/go below the avatar, I'd probably put this as an adjacent sibling component to the `UserSummary` component instead.

app/frontend/src/features/profile/view/About.tsx
```diff
@@ -19,12 +24,28 @@ export default function About({ user }: AboutProps) {
...
+
+      <LabelAndText
+        label={AGE_GENDER}
+        text={`${user.age} / ${user.gender} ${
+          user.pronouns ? `(${user.pronouns})` : ""
+        }`}
+      />
+      <LabelAndText
+        label={LANGUAGES_FLUENT}
+        text={user.languagesList.toString().replace(",", ", ") || "Not given"}
```
darrenvong: Can you extract the "Not given" text to the constant file as well please?

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -118,8 +122,15 @@ export default function Overview({ user }: OverviewProps) {
...
       <div className={classes.info}>
-        <UserReferences user={user} />
-        <UserLastActive user={user} />
+        <LabelAndText label={REFERENCES} text={`${user.numReferences || 0}`} />
+        <LabelAndText
+          label={LAST_ACTIVE}
+          text={
+            user.lastActive
+              ? timeAgo(timestamp2Date(user.lastActive))
+              : "Unknown"
```
darrenvong: Another good one for the constant file!

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -73,38 +66,45 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
...
+                label={AGE_GENDER}
+                text={`${user.age} / ${user.gender} ${
+                  user.pronouns ? `(${user.pronouns})` : ""
+                }`}
+              />
+              <LabelAndText
+                label={LANGUAGES_FLUENT}
+                text={
+                  user.languagesList.toString().replace(",", ", ") ||
+                  "Not given"
```
darrenvong: Make this a constant/re-use from constant file please

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -73,38 +66,45 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
...
+              <LabelAndText
+                label={REFERENCES}
+                text={`${user.numReferences || 0}`}
+              />
+              <LabelAndText
+                label={LAST_ACTIVE}
+                text={
+                  user.lastActive
+                    ? timeAgo(timestamp2Date(user.lastActive))
+                    : "Unknown"
```
darrenvong: Same comment as above

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -41,26 +55,57 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
...
-              {COMMUNITY_STANDING}
-            </ScoreBar>
-            <ScoreBar value={user.verification * 100}>
-              {VERIFICATION_SCORE}
-            </ScoreBar>
-
-            {user.aboutMe.length < 300
-              ? user.aboutMe
-              : user.aboutMe.substring(0, 300) + "..."}
+              <Typography variant="h6" className={classes.about}>
```
darrenvong:
```suggestion
              <Typography component="h3" variant="h6" className={classes.about}>
```
This shouldn't be a `h6` since it skips a couple of heading levels which is bad practice for accessibility. Better to use a `h3` but style it like a `h6` for visual users.

darrenvong (dismissed):
Screenshot 2021-03-14 at 17 35 35

I also noticed the `Age/Gender` and the value (same applies for all of those label and text things) don't seem to align horizontally properly. Can we look into fixing that as well?

app/frontend/src/components/UserSummary.tsx
```diff
@@ -42,31 +42,33 @@ interface UserSummaryProps {
...
-          <>
-            <ScoreBar value={(user?.communityStanding || 0) * 100}>
-              {COMMUNITY_STANDING}
-            </ScoreBar>
-            {children}
-          </>
-        }
-      />
-    </div>
+    <>
```
darrenvong: Don't need this fragment anymore since `children` is going back into the `secondary` component

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -41,26 +55,57 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
       <Card className={classes.card}>
         <CardActionArea>
           <CardContent>
-            <div className={classes.resultHeader}>
-              <Avatar user={user} />
-              <Container>
-                <Typography variant="h2">{user.name}</Typography>
-                <Typography variant="subtitle1">
+            <UserSummary user={user}>
+              <Container
```
lucaslcode: I wrote this `Container` a few months ago when I was young and foolish, can you get rid of it and replace it with a normal `div`

lucaslcode:
I'm going to disagree with everyone, I think the `aboutMe` stuff and below shouldn't be children of `UserSummary`, since a) it isn't part of the user summary - it's details, and b) it should indeed take the full width, that's how it is in the design and it looks a little weird otherwise

darrenvong:
> I'm going to disagree with everyone, I think the `aboutMe` stuff and below shouldn't be children of `UserSummary`, since a) it isn't part of the user summary - it's details, and b) it should indeed take the full width, that's how it is in the design and it looks a little weird otherwise

That's a good catch actually, the `children` of `UserSummary` aren't really meant to be a dumping ground for everything other than some supplementary text which still fits with the user summary (e.g. the hosting status text and a user's host/meetup status) tbh when I first modified it.

>  Me: If we want a component to take up the full width/go below the avatar, I'd probably put this as an adjacent sibling component to the UserSummary component instead.

I think I'm already half agreeing with you on that one haha

app/frontend/src/features/profile/view/About.tsx
justjooshing: Extracted/reused all

app/frontend/src/features/search/SearchResult.tsx
justjooshing: Good point, I didn't realise I could just specify component level while keeping variant styling 👍

app/frontend/src/components/UserSummary.tsx
justjooshing: good catch, thanks

app/frontend/src/features/search/SearchResult.tsx
justjooshing: 👍👍

app/frontend/src/features/profile/view/About.tsx
andreaowu: Maybe I was unclear in communicating this, but I meant putting all of those into one component

app/frontend/src/features/profile/view/About.tsx
```diff
@@ -46,6 +35,7 @@ export default function About({ user }: AboutProps) {
             : ""
         }
       />
+
```
andreaowu: nit: remove this?

justjooshing: Do you mean the falsy outcome?

andreaowu: No, the extra line :P

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -41,26 +55,36 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
       <Card className={classes.card}>
         <CardActionArea>
           <CardContent>
-            <div className={classes.resultHeader}>
-              <Avatar user={user} />
-              <Container>
-                <Typography variant="h2">{user.name}</Typography>
-                <Typography variant="subtitle1">
+            <UserSummary user={user} />
+            <div className={classes.statusLabelWrapper}>
```
lucaslcode: Sorry for not being clear, but do put this div in `UserSummary`. That way you won't need the left margin, and the vertical margin will be more consistent.

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -41,26 +55,36 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
...
-              {COMMUNITY_STANDING}
-            </ScoreBar>
-            <ScoreBar value={user.verification * 100}>
-              {VERIFICATION_SCORE}
-            </ScoreBar>
-
-            {user.aboutMe.length < 300
-              ? user.aboutMe
-              : user.aboutMe.substring(0, 300) + "..."}
+            <Typography component="h3" variant="h6" className={classes.about}>
```
lucaslcode:
```suggestion
            <Typography variant="body1" className={classes.about}>
```

lucaslcode: Because I don't think this actually needs to be bigger, and also it isn't a heading

darrenvong: I was thinking that soon after I left that comment, but didn't have time to write back until now but guess it's ok since you got there :D

justjooshing: great, done

app/frontend/src/features/search/SearchResult.tsx
justjooshing: Ok, back in!

app/frontend/src/features/profile/view/About.tsx
justjooshing: done

app/frontend/src/features/user/UserDataLabels.tsx
justjooshing: I readded the About section as a full component, per @andreaowu's suggestion, but am open to change if we think it's too much to show in the search results. Also not sure if my component name is good enough

app/frontend/src/features/profile/view/About.tsx
justjooshing: done, comment below :)

app/frontend/src/features/search/SearchResult.tsx
justjooshing: How does it fit with what Darren said though?

> That's a good catch actually, the children of UserSummary aren't really meant to be a dumping ground for everything other than some supplementary text which still fits with the user summary (e.g. the hosting status text and a user's host/meetup status) tbh when I first modified it.

lucaslcode: Semantically it makes sense (to me) that the hosting status is part of the "user summary", and also it makes the alignment better.

app/frontend/src/features/profile/constants.ts
```diff
@@ -64,3 +65,16 @@ export default function booleanConversion(value: boolean | undefined) {
 }
 
 export const referencesQueryStaleTime = 10 * 60 * 1000;
+
+export const aboutText = (user: User.AsObject) => {
+  const firstName = user.name.split(" ")[0];
```
lucaslcode: There is a firstName function somewhere in `src/utils/`

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -41,26 +54,37 @@ export default function SearchResult({ user }: { user: User.AsObject }) {
...
+                    {meetupStatusLabels[user.meetupStatus]}
+                  </Typography>
+                </div>
+              </div>
+            </UserSummary>
+            <Typography variant="body1" className={classes.about}>
+              {aboutText(user)}
+            </Typography>
+            <RemainingAboutLabels user={user} />
+            <LabelsReferencesLastActive user={user} />
```
lucaslcode: \*feels bad\* but now this seems like *too many* fields for a search result...

app/frontend/src/features/user/UserTextAndLabel.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <LabelAndText
+      label={AGE_GENDER}
+      text={`${user.age} / ${user.gender} ${
+        user.pronouns ? `(${user.pronouns})` : ""
+      }`}
+    />
+    <LabelAndText
+      label={LANGUAGES_FLUENT}
+      text={
+        user.languagesList.toString().replace(",", ", ") ||
```
lucaslcode:
```suggestion
        user.languagesList?.join(", ") ||
```
Not sure if it needs ? or not

app/frontend/src/features/search/SearchResult.tsx
lucaslcode: Maybe an appropriately named prop that either show's all fields or just a couple? And put the component in storybook so it's somewhat documented

lucaslcode (changes requested):
Sorry to drag it out :(

app/frontend/src/features/search/SearchResult.tsx
lucaslcode: Or maybe split the first component into two, one with just Age / Gender and languages, and the rest, with the latter only on the profile page

justjooshing: > *feels bad*

ha! No it's fine

> but now this seems like too many fields for a search result...

@andreaowu and I wanted to see whether it would work or not

image
vs
image

I've separated them out now

app/frontend/src/features/user/UserTextAndLabel.tsx
justjooshing: I dont think so because languagesList is an array. So if we have user we'll have user.languagesList === [ ], which isn't falsy.
I could do user?.languagesList but I don't think thats needed since we have the ||

lucaslcode: Sorry I did indeed mean `user?.languagesList`.

lucaslcode (dismissed):

lucaslcode (approved):

PR #785, opened 2021-03-13
feature flag as env var

andreaowu:


app/frontend/.env.development
```diff
@@ -1,3 +1,4 @@
+IS_BETA_ENABLED=false
```
darrenvong: I think because of how Create React App needs it [as referenced here](https://create-react-app.dev/docs/adding-custom-environment-variables/), this needs to be prefixed with `REACT_APP_` for it to work.

app/frontend/.env.production
```diff
@@ -1,3 +1,4 @@
+IS_BETA_ENABLED=false
```
darrenvong: I think all environment variables are strings, so even if you set it as `false`, it will still be truthy. I think the solution is to not set it/have this line in this file for production.

darrenvong (dismissed):

darrenvong (dismissed):

app/frontend/.env.development
```diff
@@ -1,5 +1,5 @@
 REACT_APP_API_BASE_URL=http://localhost:8888
-REACT_APP_TILE_URL="https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
-REACT_APP_TILE_ATTRIBUTION="Map data &copy; <a href="https://osm.org/copyright">OpenStreetMap</a> contributors"
-REACT_APP_NOMINATIM_URL="https://nominatim.openstreetmap.org/"
-REACT_APP_MAPBOX_KEY = "pk.eyJ1IjoiY291Y2hlcnMiLCJhIjoiY2tpYzY4eHd2MGVpcTJ0bGdhdGhvbHhlbiJ9.u5zvDeFE9H8itTK_dNp7Pg"
\ No newline at end of file
+REACT_APP_IS_BETA_ENABLED=true
```
darrenvong: Hmm based on the change you made in the other branch, should we have this the other way around instead? i.e. remove this line here and have `REACT_APP_IS_BETA_ENABLED` set in prod

darrenvong: Yeah let's remove this here so we can still see it in dev, based on the branching logic of `(flag) ? null : show JSX` you got on the other PR.

darrenvong (dismissed):

darrenvong (dismissed):

darrenvong (approved):

PR #786, opened 2021-03-13
remove community standing and verification bars for beta

### LINKED ISSUE #752: Hide all community standing/verification bars (for now)
These are not implemented yet.

Maybe we want to leave one grayed-out in a certain place (profile?) to show it is coming, but not in other places.

andreaowu:
#752 

Built on top of #785


app/frontend/src/components/Bar/BarWithHelp.tsx
```diff
@@ -26,7 +26,7 @@ export default function BarWithHelp({
   description,
 }: BarWithHelpProps) {
   const classes = useStyles();
-  return (
+  return (process.env.REACT_APP_API_BASE_URL ? <></> :
```
darrenvong:
```typescript
  return (process.env.REACT_APP_IS_BETA_ENABLED ? // scorebar stuff
    : null
```
Shouldn't it be something like that which references the feature flag instead? Also it's React convention to return `null` if we don't want to render anything at all.

app/frontend/src/components/Bar/ScoreBar.tsx
```diff
@@ -38,7 +38,7 @@ interface ScoreBarProps extends ContainerProps {
 
 export default function SearchResult({ value, children }: ScoreBarProps) {
   const classes = useStyles();
-  return (
+  return (process.env.REACT_APP_API_BASE_URL ? <></> :
```
darrenvong: Same comment as above.

darrenvong (dismissed):

app/frontend/src/components/Bar/BarWithHelp.tsx
andreaowu: Oops

app/frontend/src/components/Bar/BarWithHelp.tsx
```diff
@@ -26,7 +26,7 @@ export default function BarWithHelp({
   description,
 }: BarWithHelpProps) {
   const classes = useStyles();
-  return (process.env.REACT_APP_API_BASE_URL ? <></> :
+  return process.env.IS_BETA_ENABLED ? null : (
```
darrenvong: Should it not be the other way around instead? This is currently saying hide the bars if beta is enabled 🤔

app/frontend/src/components/Bar/ScoreBar.tsx
```diff
@@ -38,7 +38,7 @@ interface ScoreBarProps extends ContainerProps {
 
 export default function SearchResult({ value, children }: ScoreBarProps) {
   const classes = useStyles();
-  return (process.env.REACT_APP_API_BASE_URL ? <></> :
+  return process.env.IS_BETA_ENABLED ? null : (
```
darrenvong: Same as above

app/frontend/src/components/Bar/BarWithHelp.tsx
darrenvong: ~~Or is beta meant the "soft launch" in this case?~~
Yeah the bars are still being shown in the preview...

andreaowu: Yep, I tested it earlier and they disappeared but I was using the wrong env var

andreaowu:
Dunno what's going on in branching but if #785 goes in first, it'll be all good

app/frontend/src/components/Bar/ScoreBar.tsx
```diff
@@ -38,8 +38,10 @@ interface ScoreBarProps extends ContainerProps {
 
 export default function SearchResult({ value, children }: ScoreBarProps) {
   const classes = useStyles();
-  return (
+  return process.env.REACT_APP_IS_BETA_ENABLED ? null : (
     <Container disableGutters className={classes.root}>
+      {`hello: ${process.env.REACT_APP_MAPBOX_KEY}`}
+      {/* {`bye: ${process.env.REACT_APP_IS_BETA_ENABLED}`} */}
```
darrenvong:
```suggestion
      {`hello: ${process.env.REACT_APP_MAPBOX_KEY}`}
      {/* {`bye: ${process.env.REACT_APP_IS_BETA_ENABLED}`} */}
```
I'm guessing this is not meant to be left in? 😄

app/frontend/src/components/Bar/ScoreBar.tsx
```diff
@@ -38,8 +38,10 @@ interface ScoreBarProps extends ContainerProps {
 
 export default function SearchResult({ value, children }: ScoreBarProps) {
   const classes = useStyles();
-  return (
+  return process.env.REACT_APP_IS_BETA_ENABLED ? null : (
```
darrenvong: In prod this would be `false`, so would end up showing the bars which is the opposite of what we want? 🤔

darrenvong (approved):

PR #787, opened 2021-03-14
Fix maps

### LINKED ISSUE #758: Fix maps
MapPage is broken. I think this is a minor styles-related issue.

There is a larger issue with `EditUserLocationMap` - it reloads when typing and flyTo when selection a location doesn't work well.

lucaslcode:
Fixes #758 and some related stuff

I noticed the search box doesn't say "No results" when there are no results, but I couldn't get it to work.


app/frontend/src/components/Map.tsx
```diff
@@ -69,8 +67,11 @@ export default function Map({
     return { url };
   };
 
+  const mapRef = useRef<mapboxgl.Map>();
   useEffect(() => {
     if (!containerRef.current) return;
+    //don't create a new map if it exists already
+    if (mapRef.current) return;
```
lucaslcode, line 74: This is an "easy" solution, but I'm not sure it's really the most correct solution.

darrenvong: Is this because it gets re-run on re-renders? Does this `useEffect` need to run more than once? I guess most of the dependencies are probably messing this up as they are mostly objects/functions...

darrenvong (commented):
The URLs on the purple markers are broken. When I clicked on one, I get something like this in the URL:

`https://5cf873bb--frontend.preview.coucher.org/guide/:pageId/:pageSlug?/21/edinburgh`

Oh I think it's because the places and guide route helpers are broken.

app/frontend/src/components/Map.tsx
lucaslcode: Honestly, I'm not really sure what was happening. I tried making sure all the dependencies were using `useCallback` but it didn't help. I used a dependency tracking hook and it said that map had a first render every time I typed, but in that case, why were more and more maps being added to the same div? Maybe because they originate from outside react?

lucaslcode: But this is working so 🤷🏻

darrenvong (approved):

PR #788, opened 2021-03-14
Remove gender from service.user.updateProfile

lucaslcode:
Gender was still being set on the service call, so the backend wasn't allowing any profile change.


darrenvong (approved):
Lol did you push to `develop` directly? I see those revert commits on this 😂

PR #790, opened 2021-03-14
fix messages

### LINKED ISSUE #757: Small messaging tweaks
- [x] "Create a new group chat" -> "Create a new chat"
  - [x] Make sure dm also work from that dialog
- [ ] Change NavButtons to normal IconButtons
- [x] Chats with unread messages should be bold/have a badge
- [x] Don't let SendField grow (because it covers the message list)
- [x] Desktop tweaks?
- [x] Check all margins are appropriate
- [ ] New Request page - fix layout (all fields same size)

andreaowu:
#757 

Screen Shot 2021-03-13 at 11 16 19 PM
Screen Shot 2021-03-13 at 11 16 47 PM
Screen Shot 2021-03-13 at 10 40 34 PM
Screen Shot 2021-03-13 at 10 40 40 PM
Screen Shot 2021-03-14 at 8 21 54 PM
Screen Shot 2021-03-14 at 8 22 07 PM


app/frontend/src/features/messages/useSendFieldStyles.ts
```diff
@@ -3,13 +3,14 @@ import { makeStyles } from "@material-ui/core";
...
   button: {
     [theme.breakpoints.up("md")]: {
-      margin: 0,
       marginInlineStart: theme.spacing(2),
+      marginBottom: theme.spacing(1),
       marginTop: theme.spacing(1),
     },
     display: "block",
     flexShrink: 0,
     marginInline: "auto",
```
lucaslcode: I know I wrote this (lol) but can you change this line to `marginInlineStart` and `marginInlineEnd` since this shorthand is not well supported yet.

lucaslcode (dismissed):
I noticed you can set a title on a DM, this shouldn't be allowed I think? Should we just make it produce a backend error, or prevent the possibility in the frontend, or both? Or open a bug and save it for later.

lucaslcode (approved):
I thought maybe you have to check for `title` before sending the data, in case you choose 2 friends, type something, then remove one, but looks like the title form state gets cleared on unmount.

PR #796, opened 2021-03-14
Backend/feature/coordinates to timezone

### LINKED ISSUE #795: Function to convert (lat, lng) into timezone
Need a function to convert (longitude, latitude) into a timezone, so that when users input their address (which is used to derive (longitude, latitude)), we can infer their timezone. This timezone would then be used to view all time data through, such as the time of events.

OmerBaddour:
Added function to utils, and test_utils.py

### DIFF
--- a/app/backend/requirements.txt
+++ b/app/backend/requirements.txt
@@ -17,3 +17,4 @@ pyyaml
 requests
 Shapely
 sqlalchemy
+timezonefinder[numba]
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -7,6 +7,7 @@ from geoalchemy2.shape import from_shape, to_shape
 from google.protobuf.timestamp_pb2 import Timestamp
 from shapely.geometry import Point, Polygon, shape
 from sqlalchemy.sql import func
+from timezonefinder import TimezoneFinder
 
 from couchers.config import config
 
@@ -94,6 +95,16 @@ def get_coordinates(geom):
     shp = to_shape(geom)
     # note the funiness with 4326 normally being (x, y) = (lng, lat)
     return (shp.y, shp.x)
+
+
+def to_timezone_lng_lat(lng, lat):
+    """
+    Returns the timezone name corresponding the (lng, lat) input
+    The timezone names are defined in this timezone database:
+    https://www.iana.org/time-zones
+    """
+    tf = TimezoneFinder()
+    return tf.timezone_at(lng, lat)
 
 
 def http_date(dt=None):
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -97,9 +97,9 @@ def get_coordinates(geom):
     return (shp.y, shp.x)
 
 
-def to_timezone(lat, lng):
+def timezone_at_coordinate(lat, lng):
     """
-    Returns the timezone name corresponding the lat, lng input, or None if the input is
+    Returns the timezone name corresponding the lat, lng input, or None if the input does not correspond to a timezone or is invalid.
     The timezone names are defined in this timezone database:
     https://www.iana.org/time-zones
     """
@@ -107,7 +107,7 @@ def to_timezone(lat, lng):
     try:
         return tf.timezone_at(lat=lat, lng=lng)
     except ValueError as ve:
-        print("Error: to_timezone():", ve)
+        print("Error: timezone_at_coordinate():", ve)
 
 
 def http_date(dt=None):
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -97,14 +97,17 @@ def get_coordinates(geom):
     return (shp.y, shp.x)
 
 
-def to_timezone_lng_lat(lng, lat):
+def to_timezone(lat, lng):
     """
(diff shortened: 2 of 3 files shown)


app/backend/requirements.txt
```diff
@@ -17,3 +17,4 @@ pyyaml
 requests
 Shapely
 sqlalchemy
+timezonefinder[numba]
```
aapeliv, line 20: This file should be sorted

app/backend/src/couchers/utils.py
```diff
@@ -96,6 +97,16 @@ def get_coordinates(geom):
     return (shp.y, shp.x)
 
 
+def to_timezone_lng_lat(lng, lat):
+    """
+    Returns the timezone name corresponding the (lng, lat) input
+    The timezone names are defined in this timezone database:
+    https://www.iana.org/time-zones
+    """
+    tf = TimezoneFinder()
```
aapeliv: Does this reload all of TF every time this function is called? Do we need to create a cached `TimezoneFinder` object and call that each time we want to do the actual computation?

Does this function throw an exception on failure? When can it fail?

app/backend/src/tests/test_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+@pytest.fixture(autouse=True)
+def _(testconfig):
+    pass
+
+
+def test_to_timezone_lng_lat():
+    tf = TimezoneFinder()
+    lat, lng = 52.5061, 13.358
+    assert tf.timezone_at(lng=lng, lat=lat) == "Europe/Berlin"
```
aapeliv: You need to test the actual function in `utils.py`, not the TimezoneFinder module (I hope they have their own tests!)

aapeliv:
@OmerBaddour: how're you doing on this?

aapeliv:
Also, please make it `(lat, lng)`, we live in EPSG4326

OmerBaddour:
Had a busy week will work on this today

app/backend/src/couchers/utils.py
OmerBaddour: I can initialize with `TimezoneFinder(in_memory=True)` to [cache the relevant data](https://timezonefinder.readthedocs.io/en/latest/1_usage.html#initialisation)

app/backend/src/couchers/utils.py
```diff
@@ -96,6 +97,19 @@ def get_coordinates(geom):
     return (shp.y, shp.x)
 
 
+def to_timezone(lat, lng):
+    """
+    Returns the timezone name corresponding the lat, lng input, or None if the input is
```
aapeliv: if the input is?

app/backend/src/couchers/utils.py
```diff
@@ -96,6 +97,19 @@ def get_coordinates(geom):
...
+    The timezone names are defined in this timezone database:
+    https://www.iana.org/time-zones
+    """
+    tf = TimezoneFinder(in_memory=True)
+    try:
+        return tf.timezone_at(lat=lat, lng=lng)
+    except ValueError as ve:
+        print("Error: to_timezone():", ve)
```
aapeliv: This seems incomplete.

Does TF not have its own exception we could catch?

app/backend/src/tests/test_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_to_timezone():
+    # test valid lat, lng
+    lat, lng = 52.5061, 13.358
+    assert to_timezone(lat=lat, lng=lng) == "Europe/Berlin"
+
+    # test invalid lat, lng
+    lat, lng = -91, -181
+    assert to_timezone(lat=lat, lng=lng) == None
```
aapeliv: Use `is` when comparing to `None`

app/backend/src/couchers/utils.py
```diff
@@ -96,6 +97,19 @@ def get_coordinates(geom):
     return (shp.y, shp.x)
 
 
+def to_timezone(lat, lng):
```
aapeliv: Let's call this function `timezone_at_coordinate`

aapeliv (changes requested):

app/backend/src/couchers/utils.py
OmerBaddour: I followed the sequence of methods and functions that are called, and only a `ValueError` can be raised.

app/backend/src/couchers/utils.py
```diff
@@ -96,6 +97,19 @@ def get_coordinates(geom):
...
+    """
+    Returns the timezone name corresponding the lat, lng input, or None if the input does not correspond to a timezone or is invalid.
+    The timezone names are defined in this timezone database:
+    https://www.iana.org/time-zones
+    """
+    tf = TimezoneFinder(in_memory=True)
+    try:
+        return tf.timezone_at(lat=lat, lng=lng)
+    except ValueError as ve:
+        print("Error: timezone_at_coordinate():", ve)
```
aapeliv, line 110: Please don't `print`, you can `logger.debug` if you wish.

Also please explicitly `return None`, it's just cleaner that way!

aapeliv (commented):
Looks good, small nit.

Is there any place within the WGS84 bounds that also returns `None`, or is that only outside the bounds?

Issue #797, opened 2021-03-14
Host request view doesn't give request info


lucaslcode:
Also there is an overflow menu that only says "placeholder" which can be removed.

Also there is no padding between the "accept/reject/etc" buttons and message box.

andreaowu:
Don't need to do this, I'm going to make the actual request box.

andreaowu:
Oh wait jk, this is for the message after making a request.

justjooshing:
@andreaowu Am I confused? I thought I did this card?

> this is for the message after making a request.

Do i need to do something else for this? ^

lucaslcode:
Andrea's comment was from before your PR

justjooshing:
Ha ok, I was so confused :joy:

PR #808, opened 2021-03-15
Frontend/chore/add host request info

justjooshing:
For [#797](https://github.com/Couchers-org/couchers/issues/797)

Do we think the pending icon is obvious enough?
image

Also, on mobile view the whole message section is quite small
image


justjooshing:
Uhhh - uhoh, why did my search branch push to here too?

lucaslcode:
I changed the base branch.

lucaslcode:
I think it's too tall, especially for mobile. How is it on the design?

justjooshing:
> I think it's too tall, especially for mobile. How is it on the design?

I agree. In the design there's no back arrow or "Request from surfer", but that's because there's the messages panel on the left hand side, but I can't see a mobile design.

andreaowu:
This is what it looks like in the mocks, we should probably get rid of the hosting status in this PR?
Screen Shot 2021-03-15 at 4 17 08 PM

justjooshing:
Not really sure where we can go from here. This is removing the hosting confirmation icon, and also the number of nights (just keeping the days), but it still looks too big:
image

And additionally without the hosting status
image

What do you think?

app/frontend/src/features/messages/surfing/HostRequestView.tsx
```diff
@@ -182,13 +171,25 @@ export default function HostRequestView() {
...
       </Box>
-
       <UserSummary user={otherUser}>
         <HostingStatus hostingStatus={otherUser?.hostingStatus} />
+        {hostRequest && (
+          <>
+            <div className={classes.requestStatus}>
+              <Typography component="div" variant="h2">
+                {`Requesting to stay for
+                ${daysDifference(hostRequest.toDate, hostRequest.fromDate)}`}
```
andreaowu: Make this a function in constants.ts

app/frontend/src/features/messages/surfing/HostRequestView.tsx
```diff
@@ -182,13 +171,25 @@ export default function HostRequestView() {
...
+              </Typography>
+              <HostRequestStatusIcon hostRequest={hostRequest} />
+            </div>
+            <Typography component="p">
+              {formatDate(hostRequest.fromDate, true)} -{" "}
+              {formatDate(hostRequest?.toDate, true)}
```
andreaowu: Can this be wrapped in `${}`?

justjooshing: Yep done

justjooshing:
I can also remove the "Requesting for X nights" for that small amount of extra room?

image
image

lucaslcode:
Sure! You could also put (2 nights) on the same line as the dates. (Possibly bold?) Whatever you think is better.

justjooshing:
Tidied up slightly more, although I'm not sure about using h3 and then losing the font-weight so it's unemboldened, but looking at body etc they didn't seem to be what I needed.

image

`body1` stays large on smaller screens, and `body2` stays small on larger screens, whereas `h3` is adaptive in size, so I couldn't use one as `h3` and the other as `bodyX`. I also tried the LabelAndText but had the same problem with the `text` sizing staying the same while the `label` sizing shrinks, and also that the label sat halfway across the page hahaha!
But really, did we know about this issue with TextAndLabel?

image
image

app/frontend/src/features/messages/surfing/HostRequestView.tsx
```diff
@@ -162,33 +149,29 @@ export default function HostRequestView() {
...
-        </Menu>
       </Box>
-
       <UserSummary user={otherUser}>
-        <HostingStatus hostingStatus={otherUser?.hostingStatus} />
+        {hostRequest && (
+          <div className={classes.requestedDatesWrapper}>
+            <Typography
+              component="p"
+              variant="h3"
```
lucaslcode, line 166: I think this is good, you want the header styling but it isn't a header. So why not.

lucaslcode (changes requested):
This looks good, the only thing that concerns me is there's no indication of the current request status except for the control messages ("X Accepted the request") (which means pending requests don't have any indication).
Now that the community score bar is out, do you want to try putting that in?

lucaslcode:
Also body1 should be 1rem on desktop and body2 should be 0.875rem. Maybe I forgot it when adding the responsive text sizes...

lucaslcode:
Looking at the mockups, the actual request page doesn't have the indication, only the item in the list of requests. Do you maybe want to update the listitem to match the mockup a bit better (it has the status icon now, but no text to make it clear)
(PS. I think we should keep the message preview although it is not in the mockups)

app/frontend/src/features/messages/surfing/HostRequestView.tsx
justjooshing: No moreso the (2 nights) bit. That has the same component/variant combination, but with fontWeight: initial, just because body1 and body2 don't change fontSize dynamically (whereas h3 does).

justjooshing:
The surfer/host text 
image

image

The calendar actually isnt in this commit, but I wanted to show you how they don't align - I'm looking at it now, but the CalendarIcon has spacing around it, whereas the StatusIcons dont - but I'm not sure where to fix this.

justjooshing:
How it looks without the calendar
image

justjooshing:
> This looks good, the only thing that concerns me is there's no indication of the current request status except for the control messages ("X Accepted the request") (which means pending requests don't have any indication).
> Now that the community score bar is out, do you want to try putting that in?

Does something like this work?
image

app/frontend/src/features/messages/constants.ts
```diff
@@ -24,12 +22,15 @@ export const NO_GROUP_CHAT = "No group chats yet";
...
+    return `This request is still pending`;
+  } else
+    return isHost
+      ? `You have ${Object.values(hostRequestStatusLabels)[
+          status
+        ].toLowerCase()} this request`
+      : `Your request was ${Object.values(hostRequestStatusLabels)[
+          status
+        ].toLowerCase()}`;
+};
```
lucaslcode:
```suggestion
export const hostingStatusText = (
  isHost: boolean,
  status: HostRequestStatus
) => {
  if (status === HostRequestStatus.HOST_REQUEST_STATUS_PENDING) {
    return `This request is still pending`;
  } else
    return isHost
      ? `You have ${hostRequestStatusLabels[status].toLowerCase()} this request`
      : `Your request was ${hostRequestStatusLabels[status].toLowerCase()}`;
};
```

lucaslcode (changes requested):
Extremely tiny thing, otherwise lgtm!

app/frontend/src/features/messages/constants.ts
justjooshing: Yep, definitely better, I was trying to do something like that but couldn't figure it out without TS yelling at me 😂

lucaslcode (approved):

PR #812, opened 2021-03-16
References

### LINKED ISSUE #413: References
Should specify if the user was hosting, surfing, at an event, or other. Perhaps with associated host request or event ID.

aapeliv:
This is a complete rewrite of the references system to incorporate non-friend references after host requests.


app/backend/src/couchers/errors.py
```diff
@@ -61,6 +62,7 @@
 MISSING_PAGE_PARENT = "Missing page parent."
 MISSING_PAGE_TITLE = "Missing page title."
 MISSING_PASSWORD = "The user has a password but the request did not provide one."
+NEED_TO_SPECIFY_AT_LEAT_ONE_USER_ID = "You need to specify at least one user id."
```
kalvdans: spelling

app/backend/src/couchers/models.py
```diff
@@ -655,23 +625,107 @@ class HostRequest(Base):
     from_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
     to_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
 
-    # dates as "YYYY-MM-DD", in the timezone of the host
-    from_date = Column(String, nullable=False)
-    to_date = Column(String, nullable=False)
+    from_date = Column(Date, nullable=False)
```
kalvdans, line 635: document timezone

app/backend/src/couchers/tasks.py
```diff
@@ -102,6 +102,40 @@ def send_friend_request_email(friend_relationship):
     )
 
 
+def send_host_reference_email(reference, both_written):
+    """
+    both_written iif both the surfer and hoster wrote a reference
```
kalvdans: iff*

app/backend/src/couchers/utils.py
```diff
@@ -19,6 +20,20 @@ def Timestamp_from_datetime(dt: datetime):
     return pb_ts
 
 
+def parse_date(d: date):
```
kalvdans: d: str

app/pb/references.proto
```diff
@@ -0,0 +1,104 @@
...
+    // Returns a paginated list of references satisfying some filtering criterion
+  }
+
+  rpc WriteFriendReference(WriteFriendReferenceReq) returns (Reference) {
+    // Write a reference for your friend. Raises INVALID_ARGUMENT if you
+    // try to write for yourself, or NOT_FOUND if the user doesn't exist.
+    // Raises FAILED_PRECONDITION if you try to write a reference for someone you already wrote a friend reference for.
+  }
+
+  rpc WriteHostRequestReference(WriteHostRequestReferenceReq) returns (Reference) {
```
kalvdans, line 21: s/host request completed/surfed/

kalvdans (commented):
some comments from mobile :)

app/backend/src/couchers/utils.py
aapeliv: Good catch, thanks!

app/pb/references.proto
aapeliv: I clarified a bit. This same function is used both when surfing or hosting

aapeliv:
Closes #413

aapeliv:
Regarding timezones: the idea is that each table that contains local times also contains the tzdata identifier for which timezone those dates/times are in. Mostly that timezone identifier should be copied from the timezone identifier from the creating user, whose timezone identifier is set automatically from their location — see #795 (or overwritten manually, maybe?). This is more important for things like events that handle times and not just dates: but I thought it'd be nice have the time you can write references be until midnight local time (since it'll be visible in the UI) instead of like UTC. And also it's good that people can only start writing references after they've hosted, so midnight of the to_date. (Though now thinking about it, there might be an off-by-one error here).

app/pb/references.proto
```diff
@@ -0,0 +1,104 @@
...
+  repeated Reference references = 1;
+
+  string next_page_token = 2;
+}
+
+message WriteFriendReferenceReq {
+  int64 to_user_id = 1;
+  string text = 2;  // plain text
+  bool was_safe = 3;
+  int32 rating = 4;
```
lucaslcode: Based on the mockups, I think this should have `bool was_comfortable` (maybe appropriate is a better word than comfortable, though?) and rating should be a float from 0-1. I am going to ask for clarification in the slack.

lucaslcode: @aapeliv

darrenvong:
@aapeliv Just a heads up I merged in the frontend portion for working with this new refactored references API

darrenvong:
Got lazy and fixed the frontend conflicts directly in the web editor, hopefully I didn't f*** anything up...

lucaslcode (dismissed):

app/pb/references.proto
aapeliv: Done.

app/pb/references.proto
```diff
@@ -72,15 +72,17 @@ message ListReferencesRes {
 message WriteFriendReferenceReq {
   int64 to_user_id = 1;
   string text = 2;  // plain text
-  bool was_safe = 3;
-  int32 rating = 4;
+  bool was_appropriate = 3;
```
lucaslcode, line 75: Sorry I meant keep `was_safe` and add `was_appropriate`...

app/backend/src/couchers/models.py
```diff
@@ -655,23 +625,113 @@ class HostRequest(Base):
...
+    reference_type = Column(Enum(ReferenceType), nullable=False)
+
+    host_request_id = Column(ForeignKey("host_requests.id"), nullable=True)
+
+    text = Column(String, nullable=True)  # plain text
+
+    rating = Column(Float, nullable=False)
+    was_appropriate = Column(Boolean, nullable=False)
+
+    visible_from = Column(DateTime(timezone=True), nullable=False)
```
kalvdans: Can be calculated from `time`, no need to store in db.

kalvdans (commented):
Fixing timezone later will be a pain after introducing things like `today()`... Should be addressed in this PR in my opinion.

app/backend/src/couchers/models.py
aapeliv: The reason I've added it is that it's really quite annoying to compute. Logic is approximately this:

1. If it's a friend reference, it's immediately visible
2. Otherwise, it's visible if the other reference from the person you surfed/hosted has written the reference OR if the reference is over 2 weeks old

Storing this means it's very easy to get all visible references, avoiding a few hairy joins in some places. But I can also remove this if you feel it's important.

Also, currently it's ready for timezone support, which would need to included in that join!

aapeliv:
> Fixing timezone later will be a pain after introducing things like `today()`...

`today()` is not used much in the actual code, it's mostly in tests, so I think it's fine.

> Should be addressed in this PR in my opinion.

I'd prefer to fix all of that up in another PR where we introduce tzs for users, which then go into host requests, and from there into here. The main stuff left in this is the `least_current_date` which needs to be completely removed. I guess we need a `today_in_tz(tz=...)` function instead.

Most of this is otherwise tz aware already

app/backend/src/couchers/models.py
kalvdans: It would be more annoying to remember to recalculate the whole `visible_from` column on every friend/unfriend/reference written/timezone change. Cache invalidation is the second hardest problem in computer science :)

kalvdans:
Sounds like you have a plan how to tackle the timezone later, so I take your word for it :)

app/backend/src/couchers/models.py
aapeliv: I'll change it.

But this is set once, except for a hosting/surifng reference if the other person writes a reference within 2 weeks. Doesn't get updated if you friend/unfriend/tz change,etc.

kalvdans: Well if visible_from should be based on an old snapshot of the database then we can't calculate it on demand, then we need a column for it. Just have to make up our mind :D

aapeliv: Done, turned into a subquery: I agree, it's cleaner this way! Thanks :)

aapeliv:
@kalvdans: I removed `visible_from`, and agree now, it's cleaner. I also cleaned up a bunch of the timezone stuff, but will leave it to another PR to implement non-UTC stuff.

Ready for re-review now!

kalvdans (approved):
Backend looks fine now!

PR #819, opened 2021-03-18
[MRG] Adding GetFriendRelationship

### LINKED ISSUE #738: Get a single friend request/relationship
There's currently no way to get a single friend request/relationship. I'm thinking this would be useful, for example:

-Someone sends you a friend request
-You go to their page to see who it is
-Ideally, you can reply to the request right there on their page

-Someone sends you a friend request
-You get a notification
-You reply to friend request straight from the notification (i.e. how it is on FB)

It's entirely possible this can all be done with respondfriendrequest and I'm just out of the loop?

ShmuelTreiger:


lucaslcode:
I still don't quite understand the use case for this? If it's needed in the backend somewhere, you can use the database directly, and afaik we don't need this for frontend.

ShmuelTreiger:
I've thought about it more, and you're probably right

Issue #792, opened 2021-03-14
Host Request


andreaowu:
[ ] Add "Request" button to `Overview.tsx`
[ ] Panel with request details

PR #822, opened 2021-03-18
make host request

andreaowu:
#792 

https://user-images.githubusercontent.com/1359259/112509320-fa89bf00-8d55-11eb-9e4a-801b6c5b2579.mov


app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +45,172 @@ const useStyles = makeStyles((theme) => ({
...
+  const guests = Array.from({ length: 8 }, (_, i) => {
+    const num = i + 1;
+    return (
+      <MenuItem key={num} value={num} ref={register}>
+        {num}
+      </MenuItem>
+    );
+  });
+
+  const watchToDate = (watch("fromDate") as unknown) as Date;
```
lucaslcode: Our `DatePicker` component's value is a string "YYYY-MM-DD", so maybe this cast isn't working 100%? Try `new Date(watch("fromDate"))`

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -57,39 +57,39 @@ export default function ProfilePage() {
...
+          <Card className={classes.detailsCard}>
+            <TabContext value={currentTab}>
+              <TabBar
+                value={currentTab}
+                setValue={setCurrentTab}
+                labels={SECTION_LABELS}
+                ariaLabel="tabs for user's details"
+              />
+              {isRequesting && (
+                <NewHostRequest user={user} setIsRequesting={setIsRequesting} />
```
lucaslcode: Using [`Collapse`](https://material-ui.com/components/transitions/#collapse) here would be nice

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -62,9 +61,10 @@ const useStyles = makeStyles((theme) => ({
 
 interface OverviewProps {
   user: User.AsObject;
+  setIsRequesting: Dispatch<SetStateAction<boolean>>;
```
lucaslcode: `(value: boolean) => void` gives more flexibility imo but if you think being specific is better, leave as is

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
lucaslcode: Okay seems like this isn't the problem

lucaslcode: This should be `watchFromDate` but that also isn't the problem

lucaslcode: I think, just below this, you can add a `useEffect` something like this (bad pseudocode sorry):
```
useEffect(() => {
  if (watchFromDate > getValues("toDate")) setValue("toDate", watchFromDate);
}, [watchFromDate])
```

app/frontend/src/components/Datepicker.tsx
```diff
@@ -32,15 +32,14 @@ export default function Datepicker({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Controller
         control={control}
-        defaultValue={new Date().toISOString().split("T")[0]}
+        defaultValue={new Date()}
```
lucaslcode: Was this causing problems? This component gives a string "YYYY-MM-DD" rather than the date, so I think the original is correct.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -53,10 +52,8 @@ export default function Datepicker({
             label={label}
             minDate={minDate}
             onChange={(date) => {
-              //user might be typing, so check the date is valid before doing conversions
-              date?.isValid()
-                ? onChange(date?.format().split("T")[0])
-                : onChange(date);
+              // user might be typing, so check the date is valid before doing conversions
+              date?.isValid() ? onChange(date?.format()) : onChange(date);
```
lucaslcode: Same, this isn't working because it should be a YYYY-MM-DD string. I could be convinced though that the conversion should be made in `service.ts` rather than here...

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,178 @@ const useStyles = makeStyles((theme) => ({
...
+    const num = i + 1;
+    return (
+      <MenuItem key={num} value={num} ref={register}>
+        {num}
+      </MenuItem>
+    );
+  });
+
+  const watchFromDate = watch("fromDate", new Date().toString());
+  useEffect(() => {
```
lucaslcode, line 128: Changing the type back to string means this will need modifying, but you can do `new Date("YYYY-MM-DD")` then do the tomorrow stuff

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,178 @@ const useStyles = makeStyles((theme) => ({
...
 
   return (
-    <Box>
+    <>
       <Typography variant="h1">
-        {hostLoading ? <Skeleton width="100" /> : title ?? null}
+        {hostLoading ? (
+          <Skeleton width="100" />
+        ) : (
+          sendRequest(user.name) ?? null
```
lucaslcode: `?? null` not necessary

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,178 @@ const useStyles = makeStyles((theme) => ({
...
+        )}
       </Typography>
-      {mutation.error && (
-        <Alert severity={"error"}>{mutation.error?.message}</Alert>
-      )}
+      {error && <Alert severity={"error"}>{error.message}</Alert>}
+      {isSuccess && <Alert severity={"success"}>{SEND_REQUEST_SUCCESS}</Alert>}
```
lucaslcode: We usually omit the curly braces for plain strings

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -65,15 +73,21 @@ export default function ProfilePage() {
...
             <Card className={classes.detailsCard}>
               <TabContext value={currentTab}>
                 <TabBar
                   value={currentTab}
                   setValue={setCurrentTab}
                   labels={SECTION_LABELS}
-                  ariaLabel="tabs for user's details"
+                  ariaLabel={SECTION_LABELS_A11Y_TEXT}
                 />
+                <Collapse in={isRequesting}>
```
lucaslcode, line 103: This looks so great, and it's so easy! We should keep these transitions in mind.

lucaslcode (dismissed):
Also please remove `package-lock.json`

app/frontend/src/features/profile/view/ProfilePage.tsx
andreaowu: Definitely :D

andreaowu:
Chatted offline about this - timezones changing strings back and forth along with creating new Date()s is causing issues.

Result is to keep passing `Date`s around on the UI and convert it back to `ISOString()` when passing to backend.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -32,15 +32,14 @@ export default function Datepicker({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Controller
         control={control}
-        defaultValue={new Date().toISOString().split("T")[0]}
+        defaultValue={minDate}
```
lucaslcode, line 35: This means when choosing a birthday, it will be preset at year 1900 😁 maybe it doesn't matter.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -53,10 +52,8 @@ export default function Datepicker({
             label={label}
             minDate={minDate}
             onChange={(date) => {
-              //user might be typing, so check the date is valid before doing conversions
-              date?.isValid()
-                ? onChange(date?.format().split("T")[0])
-                : onChange(date);
+              // user might be typing, so check the date is valid before doing conversions
+              date?.isValid() ? onChange(date.toDate()) : onChange(date);
```
lucaslcode: This was a hacky thing for the string value. I wonder if `onChange={onChange}` will work now it's all `Date`s?

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,187 @@ const useStyles = makeStyles((theme) => ({
...
+    marginTop: theme.spacing(2),
+  },
 }));
 
-export default function NewHostRequest() {
-  const classes = useStyles();
-
-  const userId = +useParams<{ userId: string }>().userId;
+interface NewHostRequestProps {
+  isSuccessRequest: (value: boolean) => void;
```
lucaslcode: Can this be `setIsSuccessRequest` or maybe even `setIsRequestSuccess`?

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,187 @@ const useStyles = makeStyles((theme) => ({
...
-              type="submit"
-              color="primary"
-              onClick={onSubmit}
-              loading={mutation.isLoading}
-            >
-              Send
-            </Button>
-          </Box>
-        </form>
+        <div>
```
lucaslcode: Remove this div if it's not necessary

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -57,23 +68,47 @@ export default function ProfilePage() {
...
+  const [isRequesting, setIsRequesting] = useState(false);
+  const [isSuccessRequest, setIsSuccessRequest] = useState(false);
+
+  const handleClose = () => {
+    setIsSuccessRequest(false);
+  };
+
   return (
     <>
+      <Snackbar
```
lucaslcode: Use our SuccessSnackbar (feel free to change the anchor to top)

app/frontend/src/utils/date.ts
```diff
@@ -18,24 +18,42 @@ const monthNames = [
...
-};
+/** Compares whether date1 is equal to or in the future of date2 */
+function isSameOrFutureDate(date1: Date, date2: Date): boolean {
+  return (
+    isSameDate(date1, date2) ||
+    date1.setHours(0, 0, 0, 0) > date2.setHours(0, 0, 0, 0)
+  );
+}
+
+export { dateTimeFormatter, formatDate, isSameOrFutureDate, numNights, timestamp2Date };
```
lucaslcode: Run `yarn format`

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -57,23 +68,47 @@ export default function ProfilePage() {
     username ?? (currentUser.data?.username || "")
   );
 
+  const [isRequesting, setIsRequesting] = useState(false);
+  const [isSuccessRequest, setIsSuccessRequest] = useState(false);
+
+  const handleClose = () => {
+    setIsSuccessRequest(false);
+  };
+
```
lucaslcode, line 82: Can we have a useEffect here with a dependency on is requesting, that uses `element.scrollIntoView()` if it's true? (Because tapping request on mobile appears to do nothing atm because it appears off screen)

lucaslcode: To scroll to the request box (I think a ref on Collapse should work)

app/frontend/src/features/messages/surfing/NewHostRequest.tsx
```diff
@@ -30,106 +46,187 @@ const useStyles = makeStyles((theme) => ({
...
+              name="text"
+              rows={6}
+              inputRef={register}
+              multiline
+              fullWidth
+              placeholder={REQUEST_DESCRIPTION}
+            />
+            <CardActions className={classes.send}>
+              <Button onClick={() => setIsRequesting(false)}>{CANCEL}</Button>
+              <Button type="submit" onClick={onSubmit}>
```
lucaslcode: Can this have loading state?

app/frontend/src/service/requests.ts
```diff
@@ -78,13 +78,17 @@ export async function getHostRequestMessages(
...
-export async function createHostRequest(
-  data: Required<CreateHostRequestReq.AsObject>
-) {
+export type CreateHostRequestWrapper = Omit<
+  Required<CreateHostRequestReq.AsObject>,
+  "toDate" | "fromDate"
+> & { toDate: Date; fromDate: Date };
+
+export async function createHostRequest(data: CreateHostRequestWrapper) {
+  console.log(data);
```
lucaslcode: Remove

lucaslcode (dismissed):
👍👍 Just some small stuff. 

`package-lock.json` is still present

lucaslcode:
Also - please make necessary changes to the completeSignup function in the service, since it is still expecting the date string so is broken atm

app/frontend/src/components/Datepicker.tsx
andreaowu: Wasn't that the case regardless? We can change the `minDate` for the Datepicker for birthday if we don't want that to be the case?

lucaslcode: Before it defaulted to today. I think maybe that will work in most instances.

andreaowu: Oh wait I see, I think that's fine

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -111,7 +119,7 @@ export default function CompleteSignupForm() {
 
     authActions.signup({
       acceptTOS: acceptedTOS,
-      birthdate: data.birthdate,
+      birthdate: data.birthdate.toISOString(),
```
lucaslcode: This needs `.split("T")[0]`

app/frontend/src/components/Datepicker.tsx
```diff
@@ -52,12 +51,7 @@ export default function Datepicker({
             }}
             label={label}
             minDate={minDate}
-            onChange={(date) => {
-              //user might be typing, so check the date is valid before doing conversions
-              date?.isValid()
-                ? onChange(date?.format().split("T")[0])
-                : onChange(date);
-            }}
+            onChange={(date) => onChange(date)}
```
lucaslcode: I wouldn't have commented if there were no other changes, but since there are this can just `onChange={onChange}`

lucaslcode (changes requested):
Last one

app/frontend/src/components/Datepicker.tsx
andreaowu: You should comment on this even if there's no other changes :P

Thanks for your thorough review!!

lucaslcode (approved):

Issue #815, opened 2021-03-17
Unread notification badge in messages is clipped on mobile


lucaslcode:
![image](https://user-images.githubusercontent.com/32044095/111549496-2fa17a80-8774-11eb-85ed-ea4c333b6cb4.png)

justjooshing:
There's not a lot of room for it to sit with in similar style as is set currently on small devices.
How do we think it looks with the default styling? I think having them overlaid (overlayed? overlain?) over the text doesn't look as clean.
![image](https://user-images.githubusercontent.com/60824744/111648912-c7967700-87fb-11eb-9550-14e8d82325f2.png)

Because even right: theme.spacing(-0.5) it still gets slightly cut off 
![image](https://user-images.githubusercontent.com/60824744/111649316-2a880e00-87fc-11eb-8bac-99ca0244e714.png)

Alternatively, we could increase the width of each of the tabs?

lucaslcode:
How does it look with overflow allowed?

PR #823, opened 2021-03-18
Set default notification badge styling for small devices

justjooshing:
Issue [#815](https://github.com/Couchers-org/couchers/issues/815).

Currently with default styling on small devices, but can change if we want to increase the tabsize instead. Images here:
https://github.com/Couchers-org/couchers/issues/815#issuecomment-802022033


justjooshing:
Looks like: 
image

lucaslcode (approved):

PR #824, opened 2021-03-18
fix: split two env variables into separate lines

Tay0108:
Had some problems with map data fetching due to this.


andreaowu (approved):
Pending on build passing

PR #825, opened 2021-03-19
Fix bouncy scrolling on mobile web

### LINKED ISSUE #817: on mobile firefox, top & bottom screen bounds "bounce" on swipe
Subject: on mobile firefox, top & bottom screen bounds "bounce" on swipe
Description:
On firefox mobile, when scrolling up or down the top bar bounces up and down and so does the bottom of the screen when you reach the bottom. This prevents you from pressing the save button when editing your profile. It also makes swiping on the map very difficult/cumbersome. See attached video.

Steps:
Used firefox on my mobile device. Seems to happen whenever swiping up or down. See attached video.

Results:
The bouncing of the top and bottom bounds of the screen should not happen. The bounds should be fixed so that they do not bounce around.

Backend version: develop-3462c04b
Frontend version: develop-3317e6da
Page: https://develop--frontend.preview.coucher.org/profile/edit#loc=13/43.0653/-89.42332
User ID: 7

darrenvong:
Considering I can just about reproduce this on mobile Chrome (Firefox works fine for me on Android even without this fix...), fixing the map overflow on the edit profile page seems to have done the trick.

Might be good to have an iPhone user to check and see how it behaves on Safari too? 🤷 

Fixes #817

### DIFF
--- a/app/frontend/src/components/EditUserLocationMap.tsx
+++ b/app/frontend/src/components/EditUserLocationMap.tsx
@@ -24,7 +24,7 @@ const useStyles = makeStyles({
   root: {
     height: 200,
     position: "relative",
-    width: 400,
+    maxWidth: 400,
   },
 });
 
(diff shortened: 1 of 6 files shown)


app/frontend/src/components/EditUserLocationMap.tsx
```diff
@@ -24,7 +24,7 @@ const useStyles = makeStyles({
   root: {
     height: 200,
     position: "relative",
-    width: 400,
+    maxWidth: 400,
```
lucaslcode, line 27: really going all out on this PR

lucaslcode:
Let's see if it is fixed for jesse

lucaslcode (approved):

PR #828, opened 2021-03-20
Add manual mock to allow "service" to be imported without "index"

darrenvong:
A small workaround so we can import `service` rather than having to do `service/index`. I think it's worth it given this limitation is imposed by our tests rather than actually the import not working!


lucaslcode (approved):
Assuming this did in fact pass CI

Issue #756, opened 2021-03-08
Put something in the dashboard until communities are done


lucaslcode:
Also get rid of the "Create X" buttons

lucaslcode:
@ItsiW  thoughts?

andreaowu:
If this is just for soft launch, what about 
```
Welcome to Couchers! Thanks for your patience as we build out: 
- list of features that we're currently working on for beta launch
```

ItsiW:
Yeah something simple like that is good for now. We might want a little explanation about transition from alpha for our current users. I'll work on some copy for beta.

For listing features I was thinking of creating a priority order page on the site which we can link to, that way we can signal things like hangouts and moderation tools are in the scope

andreaowu:
When those are ready, please drop it here and I'll get it changed :)

ItsiW:
Sorry for the delay. I've added some stuff from the alpha home page

Welcome to Couchers.org!

We're in the process of building out the platform, so we appreciate your patience as we create new features. 

If you've been here before, you may have noticed that some things have changed. That's because we're migrating from our early alpha version to our new beta version. If you were on the alpha version, don't worry, most of your profile data has been kept!

## Bugs

Bugs are to be expected, we are still in the testing phase and it's great that you're here to help with that. If you notice any bugs, please report them with the red button in the top right corner.

Please discuss anything else about the app on the [Community Forum](https://community.couchers.org).

## Features

We're working on a few more things before we launch, such as events, forums and community pages, as well as improved designs, and we'll let you know as those get released.

## Help out

If you're interested in helping out on this project, we'd love for you to [sign up](https://couchers.org/signup)! We are looking for people with all kinds of experience to help us build this amazing product.

PR #830, opened 2021-03-20
dashboard info

andreaowu:
#756

Screen Shot 2021-03-19 at 9 51 36 PM

### DIFF
--- a/app/frontend/src/features/Home.tsx
+++ b/app/frontend/src/features/Home.tsx
@@ -1,22 +1,78 @@
+import { Link, Typography } from "@material-ui/core";
 import Button from "components/Button";
+import Divider from "components/Divider";
 import TextBody from "components/TextBody";
+import {
+  BUGS,
+  BUGS_DESCRIPTION_1,
+  BUGS_DESCRIPTION_2,
+  BUGS_DESCRIPTION_3,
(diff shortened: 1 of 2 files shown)


app/frontend/src/features/Home.tsx
```diff
@@ -1,22 +1,78 @@
...
-      <Button component={Link} to={newPlaceRoute}>
-        Create a new place
+      <Button component={RouterLink} to={newPlaceRoute}>
+        {CREATE_PLACE}
       </Button>
-      <Button component={Link} to={newGuideRoute}>
-        Create a new guide
+      <Button component={RouterLink} to={newGuideRoute}>
+        {CREATE_GUIDE}
       </Button>
```
darrenvong: I wonder if there's much value keeping this around and if it's better/easier to just delete it, as I doubt we'll want to keep any of this when we build out the dashboard properly. What do you think?

andreaowu: Yeah I was thinking that too, but I wasn't sure whether we would actually need it. Guess we can just make it when / if we do!

darrenvong (dismissed):

darrenvong (approved):

PR #834, opened 2021-03-21
Frontend/bugfix/place creation

### LINKED ISSUE #771: An unknown backend error occurred when trying to create a new place
Subject: An unknown backend error occurred when trying to create a new place
Description:
Tried to create a new place "Madison, Wisconsin" with place content of "test content". I got the "An unknown backend error occurred. Please consider reporting a but!" error message.

Steps:
Created the page as described above, went to a different tab for a while, came back and clicked "create page" and got the error.

Results:
Got the error. I shouldn't have got the error -- it should have created the page! Or a more detailed error message (maybe another place exists with the exact same name?)

Backend version: develop-8c645f0e
Frontend version: develop-63759ab3
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:86.0) Gecko/20100101 Firefox/86.0
Page: https://develop--frontend.preview.coucher.org/place/new#loc=13/43.07476/-89.38376
User ID: 7

spreeni:
Closes #771

Fixed a small query error in the backend, a range issue in map coordinates and page url redirection.

### DIFF
--- a/app/backend/src/couchers/db.py
+++ b/app/backend/src/couchers/db.py
@@ -287,13 +287,14 @@ def can_moderate_at(session, user_id, shape):
     """
     Returns True if the user_id can moderate a given geo-shape (i.e., if the shape is contained in any Node that the user is an admin of)
     """
-    cluster_ids = list(
+    cluster_ids = [
+        cluster_id for (cluster_id,) in
         session.query(Cluster.id)
         .join(Node, Node.id == Cluster.parent_node_id)
         .filter(Cluster.is_official_cluster)
         .filter(func.ST_Contains(Node.geom, shape))
         .all()
-    )
+    ]
     return _can_moderate_any_cluster(session, user_id, cluster_ids)
 
 
--- a/app/frontend/src/components/EditLocationMap.tsx
+++ b/app/frontend/src/components/EditLocationMap.tsx
@@ -75,7 +75,7 @@ export default function EditLocationMap({
   };
 
   const onMove = (e: MapMouseEvent | MapTouchEvent) => {
-    centerCoords.current = e.lngLat;
+    centerCoords.current = e.lngLat.wrap();
 
     (map.current!.getSource("location") as GeoJSONSource).setData(
       pointGeoJson(centerCoords.current!)
--- a/app/frontend/src/features/communities/redirect.ts
+++ b/app/frontend/src/features/communities/redirect.ts
@@ -1,7 +1,7 @@
 import { Page, PageType } from "pb/pages_pb";
 import { routeToGuide, routeToPlace } from "routes";
 
-export const pageURL = (page: Page.AsObject) => 
+export const pageURL = (page: Page.AsObject) =>
   page.type === PageType.PAGE_TYPE_GUIDE
-  ? routeToGuide(page.pageId, page.slug)
-  : routeToPlace(page.pageId, page.slug);
+    ? routeToGuide(page.pageId, page.slug)
+    : routeToPlace(page.pageId, page.slug);
--- a/app/frontend/src/features/communities/redirect.ts
+++ b/app/frontend/src/features/communities/redirect.ts
@@ -1,10 +1,7 @@
 import { Page, PageType } from "pb/pages_pb";
 import { routeToGuide, routeToPlace } from "routes";
 
-export const pageURL = (page: Page.AsObject) => {
-  if (page.type === PageType.PAGE_TYPE_GUIDE) {
-    return routeToGuide(page.pageId, page.slug);
-  } else {
-    return routeToPlace(page.pageId, page.slug);
-  }
-};
+export const pageURL = (page: Page.AsObject) => 
+  page.type === PageType.PAGE_TYPE_GUIDE
+  ? routeToGuide(page.pageId, page.slug)
+  : routeToPlace(page.pageId, page.slug);
--- a/app/frontend/src/features/communities/redirect.ts
+++ b/app/frontend/src/features/communities/redirect.ts
@@ -1,8 +1,10 @@
 import { Page, PageType } from "pb/pages_pb";
-import { guideRoute, placeRoute } from "routes";
+import { routeToGuide, routeToPlace } from "routes";
 
 export const pageURL = (page: Page.AsObject) => {
-  return `${page.type === PageType.PAGE_TYPE_GUIDE ? guideRoute : placeRoute}/${
-    page.pageId
-  }/${page.slug}`;
+  if (page.type === PageType.PAGE_TYPE_GUIDE) {
+    return routeToGuide(page.pageId, page.slug);
+  } else {
+    return routeToPlace(page.pageId, page.slug);
+  }
 };
--- a/app/frontend/src/components/EditUserLocationMap.tsx
+++ b/app/frontend/src/components/EditUserLocationMap.tsx
@@ -223,7 +223,10 @@ export default function EditUserLocationMap({
       if (!user) {
         navigator.geolocation.getCurrentPosition((position) => {
           flyToSearch(
-            new LngLat(position.coords.longitude, position.coords.latitude).wrap()
+            new LngLat(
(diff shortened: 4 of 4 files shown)


app/backend/src/couchers/db.py
```diff
@@ -287,13 +287,14 @@ def can_moderate_at(session, user_id, shape):
...
-    cluster_ids = list(
+    cluster_ids = [
+        cluster_id for (cluster_id,) in
         session.query(Cluster.id)
         .join(Node, Node.id == Cluster.parent_node_id)
         .filter(Cluster.is_official_cluster)
         .filter(func.ST_Contains(Node.geom, shape))
         .all()
-    )
+    ]
```
aapeliv, line 297: Did this end up being needed? It seems to work on the eastern hemisphere?

app/frontend/src/components/EditLocationMap.tsx
```diff
@@ -75,7 +75,7 @@ export default function EditLocationMap({
   };
 
   const onMove = (e: MapMouseEvent | MapTouchEvent) => {
-    centerCoords.current = e.lngLat;
+    centerCoords.current = e.lngLat.wrap();
```
aapeliv, line 78: Is there another map component that requires similar treatment?

aapeliv (commented):
Thanks! Excellent debugging :)

For those reading this without context, Yannic wrote to me on slack: "I found out that the map returns longitude in a range (0,360) so you first need to wrap it to (-180,180) so the backend can work with it."

So it appears if you search for something east of the default location (in Sydney... this needs a bit of work elsewhere), it will fly to the east across the antimeridian and have a longitude above 180, which needs to be wrapped.

Yannic: I added a few small comments, other than that this is good to go!

(Maybe there should be some nice check in the backend too about coordinates, but I think it's fine to leave it as a comment that coords have to be in the EPSG4326 bounds. I'm quite certain postgis will catch that though (in this case it didn't because we used the coordinate for something beforehand).)

app/backend/src/couchers/db.py
spreeni: On my machine I needed it. For me it resulted in an error when checking for moderating rights in the function `_can_moderate_any_cluster` for the query-filter `.filter(ClusterSubscription.cluster_id.in_(cluster_ids))`. The previous list of cluster IDs was a list of tuples like `[(3,), (5,), (11,)]` which the `in_` function would collapse on as the elements are no known basic data type.

app/frontend/src/components/EditLocationMap.tsx
spreeni: That is a very good question that I wanted to pose in the chat as well. For the pages issue this was the only one affected but I wanted to have a look at the other components. As far as I see the user locations work though, so it is probably caught. We have to keep this in mind though when querying if something is contained in our country geodata.

spreeni:
Ah sorry, could have elaborated more. Thanks for adding the info, Aapeli. So yeah, for negative longitudes (Western hemisphere) it will result in a longitude value as follows: `long_new = 360 + long`. So for a longitude of -20 it would result in a longitude value of 340.

app/frontend/src/components/EditLocationMap.tsx
aapeliv: There is another one in the `EditUserLocationMap.tsx`, please add a wrap to that one too.

It seems postgis *does not* do the wrapping, so I'm working on a fix for that.

spreeni: I did wrap the calls in `EditUserLocationMap.tsx`. 

But I also realized that `displaceLngLat` for points close to the poles or to the border between the 180 and -180 longitude could result in an error if we step out of the ranges. Currently our data is saved as geometrical data, right? If we save it as geographical data, I could calculate the true latitude/longitude, but with geometrical data I think it would be better to just cap the results of the function.

aapeliv: Would it be possible to just `.wrap()` the coords when they are returned? That way there wouldn't be these issues (I imagine there's also an issue with the little handle if the point is on one side of the antimeridian, and the little drag handle is on the other?).

I can do a PR to update the backend/db, a double application of this function seems to be what we need: https://postgis.net/docs/ST_WrapX.html

It'd just be nice to have these constraints in the database, but it's annoying if we have to add them for every single column that has a point.

lucaslcode:
Frontend looks good to me, I'll leave @aapeliv to approve

aapeliv (dismissed):
Yep, backend looks good, thanks! Happy to merge

app/frontend/src/components/EditLocationMap.tsx
darrenvong: I have to say all of those maths helpers in `EditUserLocationMap` feels like black magic to me that I don't understand (except `displaceLngLat` which has a gis stack exchange link)... Can we add some reference as to how those formulae come about somehow?

spreeni: @aapeliv: 
The wrap only wraps longitude, not latitude overflow. And I am not really sure if we want the wrapping to happen there as we deal with purely geometrical shapes. Here we calculate a circle in which the user lives and calculate bounding points for it. If, let's say, we have something close to the North pole and one point would be wrapped from (90.2, 0) to (89.8, 180) I think it could lead to some funky shape interactions if we want to see if something is contained in a node as the geometric shape will not support that wrap. 

I saw this geography type that we could use, but it would mean changing the database and queries significantly, maybe not worth it:
https://postgis.net/workshops/postgis-intro/geography.html

Here it says that the drawbacks are that you have to cast types and execution speed:
https://blog.crunchydata.com/blog/postgis-and-the-geography-type

Generally, I think this is very much an edge case. But it would be nice if the map does not crash if the user wants to set their location to the south or north pole to pretend to be santa clause. :)

aapeliv: I would not worry about the poles: that's not a big issue for us as people don't live within a few km (max radius) from the pole :).

In terms of `geom` vs `geog`: indeed geom is faster (I believe this comes from doing computations in the cartesian plane instead of on the spheroid/whatever defined by the Spatial Reference System: cartesian is just fine for an approximation when not close to the poles and when doing computations on short distances). It seems geog would wrap automatically, but I'm afraid of the performance penalty (we will end up doing quite a few point in polygon queries with community features). If needed, we can always cast to geog on the fly.

Fiji seems to be the only properly populated place that crosses the antimeridian, and even then, it's only one island with population of a few thousand people (looks funny on OSM: https://www.openstreetmap.org/relation/571747#map=8/-17.780/180.000).

app/frontend/src/features/communities/redirect.ts
```diff
@@ -1,8 +1,10 @@
 import { Page, PageType } from "pb/pages_pb";
-import { guideRoute, placeRoute } from "routes";
+import { routeToGuide, routeToPlace } from "routes";
 
 export const pageURL = (page: Page.AsObject) => {
```
lOdwrot: You can use shorter syntax with inline function:
export const pageURL = (page: Page.AsObject) => page.type === PageType.PAGE_TYPE_GUIDE
 ? routeToGuide(page.pageId, page.slug)
 :  routeToPlace(page.pageId, page.slug)

app/frontend/src/components/EditLocationMap.tsx
spreeni: Ok, that makes sense to me. Although poor Fijians :( 

Then I would just include a latitude wrap for the function and would cap the latitude at 90° and -90° so it may show up a bit weird but does not crash for these cases.

spreeni:
Added clipping of invalid displacement coordinates to avoid errors and so that the user location circle does not go across the whole world map.

aapeliv (dismissed):
Looks good to me, did you run `yarn format` on the frontend?

aapeliv (approved):
LGTM

Issue #798, opened 2021-03-14
Accepting friend request works, but the request stays in the loading state...


lucaslcode:
...in the "Friend Requests" section, until you reload. Maybe the query cache isn't invalidated on success?

vrouilhac:
Hi, I'd like to see if I can understand what is going on here :D

vrouilhac:
~~I'm not sure but in the file FriendRequestReveive.tsx you set the loading state when isLoading is true but also if isSuccess is true~~

~~in useFriendRequests isLoading is set to 0 ( which is PENDING ) but I'm not sure it updates after that assignment~~

telalpal:
Hi @lucaslcode ,
it's my first try to contribute, so probably I am doing smth wrong.
I tried to pick up this issue or add labels/reviewers at PR but seems I don't have such rights. 
Please take a look if you will have time https://github.com/Couchers-org/couchers/pull/837.
Thanks!

darrenvong:
Closed by #837

PR #837, opened 2021-03-22
Fix query name in invalidation logic, issue #798

telalpal:
After sending a queries to accept/decline income Friends Requests it was tried to invalidate "friendRequestsReceived" and "friendRequestsSent". The problem is that there no such queries and mentioned ones are placed under "friendRequests". So all changes are about this.

### DIFF
--- a/app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
+++ b/app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
@@ -28,7 +28,7 @@ export default function useCancelFriendRequest() {
         setMutationError("");
       },
       onSuccess: (_, { userId }) => {
-        queryClient.invalidateQueries("friendRequestsSent");
+        queryClient.invalidateQueries(["friendRequests", {"type":"sent"}]);
         queryClient.invalidateQueries(["user", userId]);
       },
     }
--- a/app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
+++ b/app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
@@ -2,6 +2,7 @@ import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { Error } from "grpc-web";
 import { useMutation, useQueryClient } from "react-query";
 import { service } from "service";
+import { friendRequestKey } from "queryKeys";
 
 import { SetMutationError } from ".";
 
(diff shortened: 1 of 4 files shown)


app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
```diff
@@ -28,7 +28,7 @@ export default function useCancelFriendRequest() {
         setMutationError("");
       },
       onSuccess: (_, { userId }) => {
-        queryClient.invalidateQueries("friendRequestsSent");
+        queryClient.invalidateQueries(["friendRequests", {"type":"sent"}]);
```
darrenvong: Can you update this to use the `friendRequestKey` helper function in `src/queryKeys.ts` instead please? We are gradually moving towards centralising all of the query keys we use across the app so this type of mistakes will happen less often.

app/frontend/src/features/connections/friends/useRespondToFriendRequest.ts
```diff
@@ -27,11 +27,11 @@ export default function useRespondToFriendRequest() {
       },
       onMutate: async ({ setMutationError }) => {
         setMutationError("");
-        await queryClient.cancelQueries("friendRequestsReceived");
+        await queryClient.cancelQueries(["friendRequests", {"type":"received"}]);
```
darrenvong: Same comment as in the `useCancelFriendRequest` hook

app/frontend/src/features/connections/friends/useRespondToFriendRequest.ts
```diff
@@ -27,11 +27,11 @@ export default function useRespondToFriendRequest() {
...
       },
       onSuccess: () => {
         queryClient.invalidateQueries("friendIds");
-        queryClient.invalidateQueries("friendRequestsReceived");
+        queryClient.invalidateQueries(["friendRequests", {"type":"received"}]);
```
darrenvong: And here too please!

darrenvong (changes requested):
Wow I updated this area recently and missed these oops, nice catch and thanks for your contribution!

app/frontend/src/features/connections/friends/useCancelFriendRequest.ts
telalpal: Sure, just pushed.
Missed this file :-)

darrenvong (approved):
Feel free to look at other issues you're interested in working on and write a comment on it if any doesn't make sense or need more info!

PR #843, opened 2021-03-22
Fix failed frontend build

darrenvong:
A recent merged change I did came from someone who's not part of the couchers org so CI didn't run against their PR to pick this up. Should have probably pulled the branch down locally to check. Oops.


andreaowu (dismissed):
Pending on passing build

Tay0108 (approved):
Build passed

PR #844, opened 2021-03-22
Fix EditUserLocationMap on mobile

### LINKED ISSUE #820: EditUserLocationMap can't edit on mobile
Moving the map works, but can't move the circle

lucaslcode:
Was missing touchmove events


app/frontend/src/components/EditUserLocationMap.tsx
```diff
@@ -77,9 +77,15 @@ export default function EditUserLocationMap({
 
     map.current!.getCanvas().style.cursor = "grab";
 
-    const moveEvent = (e: MapMouseEvent | MapTouchEvent) => onCircleMove(e);
-    map.current!.on("mousemove", moveEvent);
-    map.current!.once("mouseup", (e) => onCircleUp(e, moveEvent));
+    if (e.type === "touchstart") {
```
darrenvong, line 80: Couldn't we attach the event handlers to all of these events without this if-else branching? Or will that cause the mouse event to fire twice?

lucaslcode: The mouse move event gets removed in the mouse up event, so I'm worried about adding a bunch of touch move events that never get removed

darrenvong: There's a `map.current!.off("touchmove", ...)` straight after the same for `mousemove`. Would that not remove it on every call of `onCircleUp`?

darrenvong: On the topic of removing handlers... shouldn't we also be removing those assigned in `initializeMap`? Otherwise we could get some bad memory leaks each time a map is created...

lucaslcode: I just tried this. I occasionally got weird behaviour without the if branching, I guess it was double-firing.

I tried adding a Destructor return function, but it's no good and also not needed because the `postMapInitialize` `useEffect` returns immediately if the map already exists

darrenvong: That sounds like that early `useEffect` exit thing is stopping us from cleaning up the map if it unmounts.

In their (only) React tutorial it looks like there's a `map.remove()` function that'd clean up everything the map holds onto when called. Can we call that in this component with a `useEffect` like:

```typescript
useEffect(() => {
    return () => {
      map.current!.remove();
    };
}, []);
```
`map` is a ref so the lint rule shouldn't complain about missing it out.

lucaslcode: Done

darrenvong:
Merge conflicts 😅 
Happy to approve once that's fixed!

app/frontend/src/components/EditUserLocationMap.tsx
```diff
@@ -77,9 +77,15 @@ export default function EditUserLocationMap({
...
+      const touchMoveEvent = (e: MapTouchEvent) => onCircleMove(e);
```
darrenvong: I think it might be clearer to name these `handleTouchMove` and `handleMouseMove` instead of `touchMoveEvent` and `moveEvent`? It's a bit confusing to read an event being attached on an event 🤔 

And maybe do the same for the `onCircleUp` and `onHandleUp` and name the second parameter `handleMove` or something similar.

darrenvong (changes requested):
Ok this is the last one! And also there's some merge conflicts that need fixing too 😅

app/frontend/src/components/EditUserLocationMap.tsx
lucaslcode: I can do `handleTouchMove`/`handleMouseMove`  but I wanted to avoid the others because the you get `handleHandleUp`....

darrenvong: Oh sorry I wasn't being clearer, I meant the variable name in the parameter of `onCircleUp`/`onHandleUp`. You can still rename the second parameter so it reads something like:
```typescript
const onCircleUp = (
    e: MapMouseEvent | MapTouchEvent,
    handleMove: (x: any) => void
  )
```
?

darrenvong (approved):

PR #847, opened 2021-03-23
Frontend/feature/new references api call

### LINKED ISSUE #832: Update references to use the new API in references.proto
This is a breaking API change so #812 can't be merged until the frontend has been updated to use the new references API too.

I'll also try to make this work with pagination too as the new API now supports this nicely with the different filters expected as in the [Figma design here](https://www.figma.com/file/z1aQUklCQewLsEFUo2HsLz/Desktop-Mockups?node-id=4661%3A19717)

darrenvong:
This is a lot bigger than I expected... it's basically a half re-write of the references stuff to support pagination as well on the frontend. The components got a bit out of control with the original setup, so had to break things up a lot more than the initial implementation.

**Desktop**
Screenshot 2021-03-23 at 02 17 23

Screenshot 2021-03-23 at 02 23 03

**Mobile**
Screenshot 2021-03-23 at 02 25 10

Screenshot 2021-03-23 at 02 25 37

Closes #832 once merged.


app/frontend/src/utils/hasAtLeastOnePage.ts
```diff
@@ -0,0 +1,20 @@
+import { InfiniteData } from "react-query";
+
+type EndsWithList = `${string}List`;
+type ObjectWithListValue<TData> = {
+  [Key in keyof TData]: Key extends EndsWithList ? unknown[] : unknown;
+};
+
```
darrenvong, line 7: This is quite possibly the most bit of TypeScript black magic I managed to fit in around 15 lines of code (minus the comments)! The cool bit about this is since our list data always end with `List` from the generated types, this takes advantage of that constraint on the type for `listKey`. The result is we get autocomplete values in VS Code telling you exactly which key(s) you can use 🤯

lucaslcode:
I haven't looked yet, but why not use the ALL filter for displaying all references rather than just the sent and received separately?

darrenvong:
> I haven't looked yet, but why not use the ALL filter for displaying all references rather than just the sent and received separately?

`ListReferences` needs at least one of `from_user_id` or `to_user_id`, so as soon as I set one of them, it'd only give me back all references from one direction (given or received), so I need two queries here to get both ways.

lucaslcode:
In that case I almost think we don't need "all"

darrenvong:
> In that case I almost think we don't need "all"

Still need "all" for references received to distinguish references received by friends, guests and hosts for the various filters? I haven't add the type for references given as we always fetch back everything at the moment.

app/frontend/src/features/profile/view/AllReferencesList.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const givenUserIds =
+    referencesGivenRes?.pages
+      .map((page) => page.referencesList.map((reference) => reference.toUserId))
+      .flat() ?? [];
+  const receivedUserIds =
+    referencesReceivedRes?.pages
+      .map((page) =>
+        page.referencesList.map((reference) => reference.fromUserId)
+      )
+      .flat() ?? [];
```
lucaslcode: Are the names swapped on these?

app/frontend/src/features/profile/view/ReferenceList.tsx
```diff
@@ -0,0 +1,50 @@
...
+              isReceived ? reference.fromUserId : reference.toUserId
+            );
+            return userToShow ? (
+              <ReferenceListItem
+                key={reference.referenceId}
+                isReceived
+                user={userToShow}
+                reference={reference}
+              />
+            ) : null;
```
lucaslcode, line 44: I think it's better for `<ReferenceListItem>` to handle undefined users with a skeleton, otherwise content will be jumping around. For example, if I had visited the profile of some of the users earlier, so they are cached, but not others.

lucaslcode: Actually that's not true because you block it with a loading indicator...

app/frontend/src/features/profile/view/ReferencesView.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <>
+      {referencesError && (
+        <Alert severity="error">{referencesError.message}</Alert>
+      )}
+      {isReferenceUsersLoading || isReferencesLoading ? (
+        <CircularProgress />
+      ) : referencesRes &&
+        referencesRes.pages.length &&
+        referencesRes.pages[0].referencesList.length ? (
```
lucaslcode: `hasAtLeastOnePage`?

app/frontend/src/service/references.ts
```diff
@@ -0,0 +1,46 @@
...
+type GetReferencesGivenInput = GetReferencesBaseInput;
+
+export async function getReferencesGivenByUser({
+  userId,
+  pageToken = "0",
+}: GetReferencesGivenInput) {
+  const req = new ListReferencesReq();
+  req.setFromUserId(userId);
+  req.setReferenceTypeFilterList([]);
+  req.setPageSize(25);
```
lucaslcode: That's... quite a lot. How about 5?

app/frontend/src/service/references.ts
```diff
@@ -0,0 +1,46 @@
...
+  userId,
+  pageToken = "0",
+  referenceType,
+}: GetReferencesReceivedInput) {
+  const req = new ListReferencesReq();
+  req.setToUserId(userId);
+  req.setReferenceTypeFilterList(
+    referenceType !== "all" ? [referenceType] : []
+  );
+  req.setPageSize(25);
```
lucaslcode: Same. And maybe it should be a constant somewhere, REFERENCE_PAGE_SIZE.

lucaslcode (changes requested):
Big one.

I think we should change "All references" to "All received references". I don't see any point combining sent and received - the all refers to reference type (host, surfer, event, etc)

app/frontend/src/features/profile/view/AllReferencesList.tsx
darrenvong: Yeah for given references, the current profile is the `fromUserId` and for received references, the current profile is the `toUserId`. I feel like maybe if these are two separate API/grpc calls, we could maybe stop having to do this mental overhead of figuring out which way around and just return the userId...

app/frontend/src/features/profile/view/ReferenceList.tsx
darrenvong: Yeah at the moment it's all users or loading at the list level from the parent. Do you think showing users as soon as possible rather than waiting for all might be better user experience? Might give the perception that it's not taking as long to load 🤔

app/frontend/src/service/references.ts
darrenvong: Lol literally took what was the upper limit on the backend but can change it 👍

darrenvong:
> I think we should change "All references" to "All received references". I don't see any point combining sent and received

How else would you display them next to each other though? We'd still need somewhere to house both queries and this feels the best place to keep it consistent with `ReferencesGivenList` and `ReferencesReceivedList`, as that's more like a switch controller for choosing which list to display depending on the dropdown option selected.

darrenvong:
Hmm actually, is it worth just cutting out the references given/sent part in "all references" altogether, considering there's an entire view for just seeing references given as its own option? Dunno about you but when I'm looking at someone's profile, I probably care more about references written about them by others than references they left other people...

That'd actually simplify the code a fair bit as I would be able to remove "all references" and use the same `ReferencesGivenList` for the "all references" option... (gahh when do I learn to not let design dictate engineering 😂 )

lucaslcode:
> it worth just cutting out the references given/sent part in "all references" altogether

Yes that's what I mean. There's not need to display them together

app/frontend/src/features/profile/view/ReferenceList.tsx
darrenvong: Actually this wouldn't work if I don't use a blocking loading indicator in the parent that uses this, since it is using `useUsers` and there's no way to tell which individual query is still loading so we might end up with the case where the cached user will display correctly, but the errored users (let's say because of dodgy internet) will be stuck on an infinite loading state. Think I'll just leave it for now and we can re-visit if this becomes a UX issue when more users test this out.

lucaslcode (approved):

Issue #821, opened 2021-03-18
write tests for datepicker


andreaowu:
follow-up for #779

thdk:
relates to #1127

lucaslcode:
This was actually done in #848. Btw you can merge all your approved PRs @thdk !

PR #848, opened 2021-03-23
tests: add test to Datepicker, change date format

angelika7:
tests #821


lucaslcode:
Nice job! I will have a look later but @darrenvong should look as well

app/frontend/src/components/DateFormats.tsx
```diff
@@ -0,0 +1,212 @@
+export const dateFormats = {
+  "af-ZA": "YYYY/MM/DD",
```
darrenvong: Where did you get this information from? I'm wondering if this is included in some date time library we're using (`dayjs`?) and so we don't need to maintain a manual mapping like this ourselves.

darrenvong: Would this [localised format suggestion in the docs](https://day.js.org/docs/en/display/format#localized-formats) work?

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    testWrapper = render(<Form />, { wrapper });
+  });
+
+  it("should submit with proper date for clicking", async () => {
+    await waitFor(() => {
+      fireEvent.click(
+        testWrapper.container.querySelector(".MuiIconButton-label") as Element
+      );
+      fireEvent.click(screen.getByText("23"));
```
darrenvong:
```suggestion
      userEvent.click(screen.getByText("23"));
```
It's better to use `userEvent` for UI tests (like what you have below) since it mimics the same events a normal browser would fire.

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    testWrapper = render(<Form />, { wrapper });
+  });
+
+  it("should submit with proper date for clicking", async () => {
+    await waitFor(() => {
+      fireEvent.click(
+        testWrapper.container.querySelector(".MuiIconButton-label") as Element
+      );
```
darrenvong:
```suggestion
      userEvent.click(screen.getByRole("button", { name: CHANGE_DATE }));
```
Since this is a normal button, we shouldn't need to use `querySelector` to find it, which means this is a sign our current button isn't very accessible. It'll be better to add an aria-label so then we can locate the button with React Testing Library's more accessibility friendly selectors.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -40,10 +50,9 @@ export default function Datepicker({
...
             autoOk
             className={className}
-            clearable
             disableToolbar
             error={error}
-            format="DD.MM.YYYY"
+            format={getLocaleFormat()}
             fullWidth
             helperText={helperText}
             id={id}
```
darrenvong, line 57:
```suggestion
            id={id}
            KeyboardButtonProps={{
              'aria-label': CHANGE_DATE,
            }}
```
Adapted from the [date picker doc's](https://material-ui.com/components/pickers/) example code. You can export the `CHANGE_DATE` variable and put it in this file I think, since it's specific to this component only.

This will then hopefully make my other suggested change in the test work and make the button more accessible in the process!

app/frontend/src/components/Datepicker.tsx
```diff
@@ -5,6 +5,16 @@ import {
...
+
+const getLocaleFormat = () => {
+  for (const [key, value] of Object.entries(dateFormats)) {
+    if (key === navigator.language) {
+      return value.toString();
+    }
+  }
+};
+
 interface DatepickerProps {
```
darrenvong, line 17: `helperText` should be typed as `React.ReactNode` here, the same type as the `helperText` prop in the underlying `KeyboardDatePicker` (you can see it when hovering it over if using VS Code). We should avoid using `any` in our code if possible.

darrenvong (changes requested):
Thanks for following up with the tests!

darrenvong:
Can you rebase from develop as well when you next get a chance to look through this please? The build is failing on some old code that's not your fault and it's been fixed since!

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+  it("should submit with proper date for clicking", async () => {
+    await waitFor(() => {
+      userEvent.click(testWrapper.container.querySelector('.MuiIconButton-label') as Element);
+      userEvent.click(screen.getByText("23"));
+
+      userEvent.click(screen.getByRole("button", { name: SUBMIT }));
+    });
+    expect(submitForm).toHaveBeenCalledWith({ datefield: "2021-03-23" });
```
lucaslcode: We had to change the date picker to return a `Date` object instead of the `YYYY-MM-DD` string, so this is failing.

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+        </form>
+      );
+    };
+
+    testWrapper = render(<Form />, { wrapper });
+  });
+
+  it("should submit with proper date for clicking", async () => {
+    await waitFor(() => {
+      userEvent.click(testWrapper.container.querySelector('.MuiIconButton-label') as Element);
```
lucaslcode:
```suggestion
      userEvent.click(screen.getByLabelText(CHANGE_DATE));
```

app/frontend/src/components/DateFormats.tsx
lucaslcode: Any luck with this @angelika7 ?

lucaslcode (changes requested):

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -21,41 +15,72 @@ afterAll(() => {
...
-            label="Date field"
-            name="datefield"
-          />
-          <input type="submit" name={SUBMIT} />
-        </form>
-      );
-    };
-
-    testWrapper = render(<Form />, { wrapper });
+    cleanup();
```
darrenvong: How come this is needed? React Testing Library supposedly [calls this automatically after each test](https://testing-library.com/docs/react-testing-library/api/#cleanup) 🤔

app/frontend/src/components/Datepicker.tsx
```diff
@@ -3,13 +3,32 @@ import {
...
+      ? dateFormats[navigator.language as keyof typeof dateFormats]
+      : "DD/MM/YYYY";
+  //convoluted way to convert and single letter date formats to double letter
+  //eg. "M/D/YYYY" to "MM/DD/YYYY"
+  //this is because keyboard typing is a total disaster for single letter formats
+  locale = locale
+    .replace("DD", "D")
+    .replace("MM", "M")
+    .replace("D", "DD")
+    .replace("M", "MM");
```
darrenvong:
```typescript
    .replace("D", "DD")
    .replace("M", "MM")
    .replace("DD", "D")
    .replace("MM", "M");
```
Would we not want to swap it around like this? So it goes 'DD' -> 'DDDD' -> 'DD' (same for 'M') for date formats that are already correct?

lucaslcode:
This was such a mess, I spend 3 hours debugging, trying to get single letter formats (like M/D/YYYY) to work with typing. In the end I just decided to convert them to the respective double letter.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -42,16 +61,21 @@ export default function Datepicker({
...
+            }}
             InputLabelProps={{
               shrink: true,
             }}
             label={label}
             minDate={minDate}
-            onChange={(date) => onChange(date?.toDate())}
+            onChange={(date) => {
+              if (date?.isValid()) onChange(date?.toDate());
+            }}
```
lucaslcode, line 68: The previous one is better, because if you leave the field like "05/04/202_" and submit, you get an error, whereas with this change, it just submits the most recent valid date.

However, for some reason the tests don't pass without this change and I couldn't figure out why

app/frontend/src/components/constants.ts
```diff
@@ -5,3 +5,218 @@ export const INVALID_FILE = "Invalid file.";
 export const SELECT_AN_IMAGE = "Select an image";
 export const UPLOAD_PENDING_ERROR =
   "Confirm or cancel the picture upload before saving.";
+
+//these come from https://github.com/unicode-org/cldr-json/tree/master/cldr-json/cldr-dates-modern/main
+//but really https://stackoverflow.com/a/9893752
+export const dateFormats = {
```
lucaslcode: I tried using "L" with the LocalizedFormats plugin, but it doesn't work with typing (it thinks the mask is _ so you can only type a single digit)

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+        name="datefield"
+      />
+      <input type="submit" name={SUBMIT} />
+    </form>
+  );
+};
+
+describe("DatePicker", () => {
+  beforeEach(() => {
+    submitForm.mockClear();
```
darrenvong: Don't think this is needed either as `jest.restoreAllMocks()` is called after each test in the global test configuration.

app/frontend/src/components/Datepicker.tsx
darrenvong: Does that mean if I do `05/04/2021`, then click around and change it to like `01/06/20__`, it'll still submit `05/04/2021`?

app/frontend/src/components/Datepicker.tsx
lucaslcode: Why? It's the same in the end, no?

app/frontend/src/components/Datepicker.tsx
lucaslcode: If by click around you mean clear it and type "010620" then yes

app/frontend/src/components/Datepicker.tsx
darrenvong: Oh yeah never mind, when I first read this I thought the ordering might mess up things. This still feels a bit weird and confusing... since we're reading directly from it and manipulating straight after, why not do it in the mapping directly ?

app/frontend/src/components/Datepicker.tsx
darrenvong: I got curious, revert to the old code locally and when you do that interaction, it crashed the app on this bit of code: `data.birthdate.toISOString().split("T")[0]`, so that might be why the test is failing? I think it's done its job here in catching the problem! :D

darrenvong (approved):

lucaslcode (approved):

Issue #827, opened 2021-03-20
Group chat view doesn't scroll to bottom when you send a new message


darrenvong:
**Steps to reproduce**
1. Go to a group chat with enough messages such that scrolling is required
2. Send a new message

**Expected behaviour**
The message scroll view automatically scrolls down to the last message (either one you just sent or the latest message from another user)

**Actual behaviour**
The message scroll view is still stuck to the last message before you sent the message, so that you need to scroll down before you can see the new message(s).

lucaslcode:
Also, on mobile when the keyboard opens, it reduces the vertical size of the window and the bottom message is no longer visible. I wonder if we can react to window resizing and, if the messages were scrolled to the bottom, re-scroll to the bottom.

justjooshing:
@lucaslcode how can I see this on actual mobile? Or a different way instead of chrome dev tools (because I don't get the keyboard opening to see what I need to change on chrome)

lucaslcode:
You could perhaps mimic it by reducing the vertical height of the window. 

I'm not sure if the same events get emitted for the keyboard though. It might require some investigation by connecting the Dev tools to chrome on a real device on the develop preview, and doing `monitorEvents(document.body)` in the console, and seeing what happens when the keyboard opens.

lucaslcode:
Some googling tells me this might not work. Maybe you can do something with `onfocus` instead

lucaslcode:
I have moved the milestone but keep working on it

lucaslcode:
Closed in #849

PR #849, opened 2021-03-23
scroll to bottom on new message, and keep at bottom with window resize

justjooshing:
First part of issue [#827](https://github.com/Couchers-org/couchers/issues/827)


justjooshing:
@lucaslcode this definitely scrolls to the bottom on window resize, but im not yet sure about the keyboard biz

justjooshing:
ahh wait, im not cleaning up my event listener again

lucaslcode:
Haven't looked thoroughly but you'll only want to scroll if it was scrolled to the bottom before (if I am scrolled halfway through my messages and resize, I probably don't want it to go to the bottom)

justjooshing:
> Haven't looked thoroughly but you'll only want to scroll if it was scrolled to the bottom before (if I am scrolled halfway through my messages and resize, I probably don't want it to go to the bottom)

ah, yes good point lol ok ignore all this then hahaha

justjooshing:
@lucaslcode 

I'm thinking maybe something like onclick for the input window, but since we're not using Redux I'd annoyingly have to pass it up and down (and save the state in the parent of both), unless theres another way?

So similar code, but instead of useLayoutEffect, it's just go onclick for the TextField or form

image

lucaslcode:
I think you can do it within InfiniteMessageList with `clientHeight`

justjooshing:
Struggled with this one, and tried to make it as spaghetti free as possible, but please go ham with any ways to make it cleaner.

I'm also never sure how to handle useEffect/useLayoutEffect hooks that have no dependencies, but have a setState setter, because I don't want it just to run on load, but it's just yells warnings of infinite loops (which wont happen here).

It sends the view to the bottom of the message thread if the messageId changes (newest message is always at 0 index).

It also assumes we're at the bottom of the message thread (currentlyAtBottom) if the total height is less than the distance to the top and the view window, and sends us back to the bottom on window resize. Or doesn't if we're not at the bottom.

In case (like me) you weren't sure what these stood for:
scrollHeight:  height of the total messages (including offscreen) 
scrollTop: distance to top of scrollHeight
clientHeight: height of view window

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -66,11 +75,51 @@ export default function InfiniteMessageLoader({
...
+        // Send to bottom on window resize
+        scrollRef.current.scroll(0, scrollRef.current.scrollHeight);
+      }
+      setCurrentClientHeight(scrollRef.current.clientHeight);
+    };
+
+    if (!scrollRef.current) return;
+    window.addEventListener("resize", updateMessagePosition);
+    return () => window.removeEventListener("resize", updateMessagePosition);
+  }, [currentClientHeight]);
```
lucaslcode: You don't need the state for client height, since you can just access it from `scrollRef.current.clientHeight`. Then you don't need `currentClientHeight` as a dependency, and you can just add the event listener (it can be a normal effect rather than layout I think).

lucaslcode:
```suggestion
   const updateMessagePosition = useCallback(() => {
    if (!scrollRef.current) return;
    const currentlyAtBottom =
      scrollRef.current.scrollHeight <=
      scrollRef.current.scrollTop + scrollRef.current.clientHeight;

    if (currentlyAtBottom) {
      // Send to bottom on window resize
      scrollRef.current.scroll(0, scrollRef.current.scrollHeight);
    }
  }, []);
    
  useEffect(() => {
    window.addEventListener("resize", updateMessagePosition);
    return () => window.removeEventListener("resize", updateMessagePosition);
  }, []);
```
(I haven't actually checked this code)

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -66,11 +75,51 @@ export default function InfiniteMessageLoader({
...
+
+  //** Scroll to the bottom on new message  **//
+  // Saving messages count to know whether a new message has been created
+  const { messageId } = children.props.messages[0];
```
lucaslcode: This isn't really idiosyncratic, and it's a bit flakey. Not sure a good alternative. Maybe listening for changes in `scrollHeight`? You could also use [`React.Children.count()`](https://reactjs.org/docs/react-api.html#reactchildren) and do something if the number changes, but that's also flakey (we don't know if messages are the immediate children. Maybe someone else has some other ideas.

lucaslcode (changes requested):

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
lucaslcode: This will also fix the dependency issues which is causing the build to fail (I think)

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
lucaslcode: Maybe just adding another prop - `latestMessageId`

darrenvong: `children` aren't meant to be used this way at all and using it like that completely defeats its point, since as soon as this has more than one element, this will break. The fact you're trying to do it by using `any` is a big code smell.

@lucaslcode I see you've done something similar with `earliestMessageId` but in the opposite direction (scrolling up)? So yeah, maybe passing `latestMessageId` down as a prop is the way to go 🤷 , at least it'll be a lot more reliable than this as we'll be able to compute that from the React Query data from the parent component that'll be using this.

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: The reason I had to save `clientHeight` is so we could compare the change in height. If we use scrollRef.current.clientHeight, then we wouldn't be looking at the height that it's changing from, so never passes `currentlyAtBottom`. 

FB messenger chat always sends you back to the bottom when you click the keyboard (no matter how far up you've scrolled), and asking my friends noone's felt like it was annoying, so we also could just do that (instead of me trying calculating whether they're at the bottom or not).

Makes it much neater overall

`
  const updateMessagePosition = useCallback(() => {
    if (!scrollRef.current) return;
    scrollRef.current.scroll(0, scrollRef.current.scrollHeight);
  }, []);

  useEffect(() => {
    window.addEventListener("resize", updateMessagePosition);
    return () => window.removeEventListener("resize", updateMessagePosition);
  });`

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: > Maybe listening for changes in `scrollHeight`?

Using scrollHeight isn't an option because of the infinite loader, if you scroll up, the scrollHeight changes, which send you down to the bottom, same as React.children.count(), because there seems to be 20 children loaded in a "full" infinite loader element. 

Where can I get/make `latestMessageIid` to pass it down to?

justjooshing:
Ok this is looking much neater. Up to your opinion, but atm if you open the keyboard it'll send you back down to the latest message no matter where you were in the thread (like messenger does)

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -31,19 +32,18 @@ const useStyles = makeStyles((theme) => ({
...
   fetchNextPage: () => void;
   isFetchingNextPage: boolean;
   hasNextPage: boolean;
   isError: boolean;
   className?: string;
-  // I'm not sure on the appropriate type, but I needed children.props
-  // and kept getting "Error: Property 'props' does not exist on type 'boolean | ReactChild | ReactFragment | ReactPortal'"".
-  // and "Error: Property 'props' does not exist on type 'string'" even though props DOES exist
-  children: any;
+  children: ReactChild;
```
lucaslcode: `ReactNode` is the correct type for children

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -82,44 +82,24 @@ export default function InfiniteMessageLoader({
...
-  }, [currentClientHeight]);
+  }, []);
 
   //** Scroll to the bottom on new message  **//
-  // Saving messages count to know whether a new message has been created
-  const { messageId } = children.props.messages[0];
-  const [newestMessageID, setNewestMessageID] = useState(messageId);
-
+  const [savedMessageId, setSavedMessageId] = useState(latestMessageId);
   useEffect(() => {
```
lucaslcode: Needs to be `useLayoutEffect` if you're scrolling or manipulating the DOM.

lucaslcode: The other one doesn't need it because  the scrolling happens in a callback, not in the effect.

lucaslcode (changes requested):
Simplest way is often the best! One improvement I just thought of - we probably only want to scroll to the bottom if it was our message (not if someone else just sent us a message, that would be an unexpected scroll jump). Maybe instead of the message id, you can pass the whole message, check if the current user is the author, and scroll only then?

Could you also merge develop into this branch so the preview works again and I can try on my phone :)

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: > not if someone else just sent us a message

Yes! Bloody hell, this infuriates me when I'm trying to read something in a group chat and new messages keep pushing me to the bottom. Good catch

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -32,18 +32,20 @@ const useStyles = makeStyles((theme) => ({
 
 interface InfiniteMessageLoaderProps {
   earliestMessageId?: number;
-  latestMessageId?: number;
+  latestMessage?: { messageId: number; authorUserId: number };
```
lucaslcode: `Message.AsObject`

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -32,18 +32,20 @@ const useStyles = makeStyles((theme) => ({
...
   className?: string;
-  children: ReactChild;
+  children: ReactNode;
 }
 
 export default function InfiniteMessageLoader({
   earliestMessageId,
-  latestMessageId,
+  latestMessage,
+  currentUserId,
```
lucaslcode: Instead of drilling props, use `const currentUserId = useAuthStore().authState.userId`

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -91,15 +93,25 @@ export default function InfiniteMessageLoader({
     return () => window.removeEventListener("resize", updateMessagePosition);
   }, []);
 
-  //** Scroll to the bottom on new message  **//
-  const [savedMessageId, setSavedMessageId] = useState(latestMessageId);
-  useEffect(() => {
+  //** Scroll to the bottom after sending own new message  **//
+  const [savedMessageId, setSavedMessageId] = useState(
+    latestMessage?.messageId
+  );
```
lucaslcode: Sorry for not noticing this earlier. It would be more correct for this to use `useRef` and therefore not be in the layout effect dependency array. When you set savedMessageId, you don't actually want to trigger a re-render or run the layout effect again, you're just saving it for later comparison, and it's `latestMessage` that triggers the layout effect.

As is, the layout effect gets triggered again after you `setSavedMessageId` (but the second time obviously doesn't do anything.)

lucaslcode (changes requested):
It works great! Very minor nits

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: done 👍

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: done 👍

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
justjooshing: I think i've done this lol

app/frontend/src/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -66,11 +77,33 @@ export default function InfiniteMessageLoader({
...
+      if (!scrollRef.current) return;
+      scrollRef.current.scroll(0, scrollRef.current.scrollHeight);
+    };
+    window.addEventListener("resize", updateMessagePosition);
+    return () => window.removeEventListener("resize", updateMessagePosition);
+  }, []);
+
+  //** Scroll to the bottom after sending own new message  **//
+  const savedMessageId = useRef(latestMessage?.messageId);
+  useEffect(() => {
```
lucaslcode:
```suggestion
  useLayoutEffect(() => {
```

lucaslcode (changes requested):

lucaslcode (approved):

PR #852, opened 2021-03-24
Update colour palette and add logo

### LINKED ISSUE #816: Put the logo in and update the palette

darrenvong:
Screenshot 2021-03-24 at 18 41 54

Screenshot 2021-03-24 at 18 42 10

Screenshot 2021-03-24 at 18 42 42

Screenshot 2021-03-24 at 18 42 51

Closes #816 once merged.


app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -56,17 +58,12 @@ const useStyles = makeStyles((theme) => ({
     top: 0,
   },
   flex: {
-    [theme.breakpoints.up("md")]: {
```
darrenvong, line 59: Some redundant styles here now that we are using the `Hidden` helper

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -91,7 +92,13 @@ const useStyles = makeStyles((theme) => ({
     paddingRight: 0,
   },
   nav: {
+    alignItems: "center",
```
darrenvong, line 92: Safari stretches the nav bar height by default without this. Not sure why other browsers don't do this as default value for `align-items` is `stretch` so technically Safari was right... 🤷

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -91,7 +92,13 @@ const useStyles = makeStyles((theme) => ({
...
     display: "flex",
+    flex: 0,
```
darrenvong, line 94: Firefox makes the left nav bar wider than it needs to be without this. Again not sure why other browsers don't do this. Luckily with this combo, it works on all of them now. The fun of dealing with inconsistencies in the frontend...

app/frontend/src/resources/logo.svg
```diff
@@ -0,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 150 150">
+  <g fill="#E47701">
```
darrenvong: If I didn't check this on Safari (other browsers happily hide the other redundant bits with `<switch>` for some reason in the original SVG...), didn't even realise how much crap there were in the original logo, including duplicates of it that we don't need!

darrenvong (commented):
Lol this issue definitely took longer than expected due to all these weird things I stumbled along the way 🙃

lucaslcode (dismissed):
Should we remove all instances of button `color="secondary"`?

aapeliv:
The favicon is too small :)

Screen Shot 2021-03-24 at 5 19 33 pm

darrenvong:
Updated with all `color="secondary"` I can find + larger favicon

lucaslcode (approved):

PR #853, opened 2021-03-24
Prevent zooming on mobile

lucaslcode:
Hopefully this prevents iOS safari zooming into text fields


lucaslcode:
@aapeliv how's this

darrenvong:
https://www.30secondsofcode.org/blog/s/zooming-accessibility-tips

Especially:
> while others such as iOS 10 ignore the suggestion entirely, so you should be aware that you might not be delivering a consistent user experience for all of your users.

I don't have an iPhone to test but not sure if this would work? And it seems anti-accessibility to disable zooming for other browsers/Android that may not be affected by this...

This might be a better solution instead? https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone

lucaslcode:
It did work but you're right, it's bad accessibility. Trying this.

darrenvong:
@aapeliv should probably check this? I don't have a physical iPhone 😅

aapeliv (approved):
Yep, works now!

Issue #829, opened 2021-03-20
The last message in a group chat is right above the send message input in non-Chromium browsers


darrenvong:
**Browsers affected**
Firefox, Safari

**Reproduction Steps**
1. Go to a group chat where there are enough messages for it to require scrolling
2. Look at the spacing between the last message and the message input

**Expected behaviour**
There should be some spacing between the bottom of the last message box and the top of the message input

**Actual behaviour**
No spacing exist between the bottom of the last message and the top of the message input. See screenshot:

<img width="1281" alt="Screenshot showing the visual bug between last message and input box" src="https://user-images.githubusercontent.com/13669362/111857170-9ca94180-8927-11eb-882a-767699dc3d20.png">

I put it as low priority as it's not exactly unusable, since the messages aren't actually getting in the way of the input...

vrouilhac:
I'll look at the code may be a CSS property browser compatibility?

vrouilhac:
I made some research and it seems that the problem appears on firefox only ( have only checked chrome, firefox, opera ).
The problem is that the list ( name it parent ) contains an element ( name it mid ). The 'mid' element then contains each message. 
The parent overflow is set to auto which displays a scroll bar. The mid element is flex which displays each message is column. 

The problem is that the 'mid' element goes outside of the parent element due to margin collapsing. I replicate this error in a simpler way if you want to check this out: https://codepen.io/crysyrood/pen/MWJaLQQ ( open it in firefox to see the problem )

I'm searching how to change the behaviour so that the padding of the parent element is taken into account even if the 'mid' element is flex.

vrouilhac:
I solved this using padding in 'mid' element. The scrolling behaviour is on parent, and scrolling behaviour is applied to content edges of box model so the padding is outside of current scrollable element. I added the padding to 'mid' because 'mid' is part of the content edges of parent ( which is scrollable ). More info on this : 

https://bugzilla.mozilla.org/show_bug.cgi?id=748518#c13
https://www.w3.org/TR/CSS2/visufx.html#overflow-clipping
https://www.w3.org/TR/CSS2/box.html#box-padding-area

PR #859, opened 2021-03-25
padding added to child of scrollable

vrouilhac:
I explained the solution here #829. I hope this is the expected behaviour.


vrouilhac:
87D881DC-8665-4D13-9BE4-3CFBCD7D930B


Here is a simple understanding of what’s happening 😊

lucaslcode (changes requested):
LGTM just remove the commented out line entirely and the `className` since it's no longer used

lucaslcode (approved):

Issue #768, opened 2021-03-09
Requests Beta UX


lucaslcode:
Post-beta, we plan to unify requests and messages, Until then, we need to have good UX to overcome some design limitations:
- You can only send messages to friends
- You can't send messages to a closed request (cancelled, rejected, etc) (this is intentional)
- You can't send messages to a confirmed request (this should probably be changed? Or maybe users should be pointed towards messages?)

 ~~I think at least we need to auto-add friends when requests are confirmed? Not really sure, but in it's current state it's very confusing.~~

Add message button to `Overview.tsx`, which may require a new route like /messages/user/:userId that copies /messages/chats/:messageId but uses GetDirectMessage and asks to create a DM with that user if one doesn't exist already

Correct behavior:
- Not friends, show "Request" button on profile
  - if click "Request" - do normal request flow
- Are friends, show "Message" and "Request" button on profile
  - if click "Request" - do normal request flow
  - if click "Message" and have no pending "Request" - create new message thread between me and friend
  - if click "Message" and have pending "Request" - bring to message thread with request

Completed ~~Requests:~~
- ~~Change the backend so confirmed requests with the end date in the future are not "closed"~~
- ~~For confirmed requests in the past, cancelled and rejected requests, disable the send field and button with a helpful message explaining why~~

lucaslcode:
> if click "Message" and have no pending "Request" - create new message thread between me and friend
> if click "Message" and have pending "Request" - bring to message thread with request

This is complicated to implement, I think let's just skip it since this will be refactored later.

PR #860, opened 2021-03-25
Disable msg sending for closed/past requests

lucaslcode:
Part of #768. I think the rest can possibly be moved to beta milestone.

This disables the send button for past, cancelled and rejected requests, and puts an informative message in the text box. I also changed the backend to allow message sending to confirmed requests (unless they are in the past).

Note margins and other stuff are being worked on in other prs

image


app/frontend/src/features/messages/useSendFieldStyles.ts
```diff
@@ -28,6 +28,10 @@ const useSendFieldStyles = makeStyles((theme) => ({
       display: "flex",
     },
   },
+  requestClosedLabel: {
+    transform: "none",
+    padding: theme.spacing(2, 2),
+  },
```
lucaslcode, line 32: Workaround for `TextField` label not wrapping correctly for a multiline input

app/frontend/src/features/messages/requests/HostRequestSendField.tsx
```diff
@@ -97,6 +101,11 @@ export default function HostRequestSendField({
 
   const isButtonLoading = isLoading || isResponseLoading;
 
+  const isRequestClosed =
+    hostRequest.toDate < new Date().toISOString().split("T")[0] ||
```
darrenvong, line 105: Feels like we are doing this same check on the backend as well... Couldn't we return this too as part of the requests object?

lucaslcode: Good idea

lucaslcode: Actually on second thought, I want to just leave this as is, because we are getting rid of host request messages later so I don't want to waste time on it (and we will just have to remove that field from the protobuf message again later)

darrenvong (dismissed):

darrenvong (approved):

PR #476, opened 2020-12-25
Some geo functions

aapeliv:
This is partly to learn how all this geo stuff works, partly as I think this is useful functionality, and also to figure out a bit how the data pipelines would look like.

This actually only checks if the point is in the sea or not, it check against coastlines extracted from OSM, but does not check internal waters. So e.g. the Great Lakes all report as land.

Issue #470, opened 2020-12-24
Geo considerations


aapeliv:
The point of this issue is to collect some thoughts I have about geo stuff. It's mostly for me to keep track of what I've looked at and collect research. Feel free to comment and discuss.

By "geo" I mean stuff relating to real-world geography here, countries, areas, cities, etc. The main problem is that OSM data is very heavy and it's also not super well standardised and well curated. It's for instance hard to refer to a single object in their database because the object's ID is not guaranteed to be permanent, and things might change. We also can't directly rely on their APIs as they have limits and we'd like our platform to be spiffy with e.g. typeahead suggestions as people type.

# Countries

We need 4 kinds of country info on people's profiles: countries they've lived in (self reported), countries they've visited (self , reported), and countries they've visited through couchers (verified visited?), countries they've hosted people from.

We need some way of having user location history. So if someone moves e.g. to another country/community, then all those who surfed with them don't turn up as having been in a completely different place. For now we haven't done history of objects except for stuff that explicitly is versioned (e.g. page versions). The discussion on that is still open, but maybe we should just copy the relevant information from the user's profile, since it (information about where someone surfed) should never be updated, and it might be useful to have that for security/privacy reasons, etc.

Countries change, but they change rarely, and if someone visits a country now and that country stops existing, we should presumably not have the database shit itself. However, we should be aware of that.

There are 2 well-standardised, good ways of identifying countries (and a lot of other ones). There's the ISO 3166 alpha-2 and alpha-3 codes (it tells the number of characters in the identifier). E.g. Finland is `FI` and `FIN`, respectively. Both seem to be OK, but alpha-2 seems more common, and OSM seems to support this better.

aapeliv:
# Getting country data out of OSM

It's a bit tricky. The data is not perfect, and we're not going to start taking sides on what's an official country and what's not. Here's a log of what I've been doing:

* Get all relations with ISO 3166-1 codes. This is problematic though, it returns these non-official areas: CP, XK, TA, AC, IC, DG.

```ql
rel[type=boundary]["ISO3166-1:alpha2"]; out count;
```

* Get all country boundaries?

```ql
rel[boundary=administrative][admin_level=2]
```

* The solution might be to get relations based on a list of country codes that have `admin_level=2` or `admin_level=3`, then label points with the smallest region containing that point.

lucaslcode:
There may come a time when we DO have to take a side. I think OSM's "on the ground" policy is reasonable and very applicable to hospex.

PR #864, opened 2021-03-26
[MRG] Backend/feat/regions

ShmuelTreiger:
Some notes:

1) Under this schema there's no distinction, for example, between Scotland an England, they're both just UK. Not necessarily wrong, just something to consider. (Similarly, it doesn't have the various US states which I suppose are kind of analogous, so maybe it's for the best.)

2) Do we want to allow "other" countries somehow?

3) I didn't think it was worth massively re-doing countries as their own data type, or reworking how we treat them (i.e. as a "|" delineated string).

4) I'm sure this bit of code

```
                    else:
                        if user.countries_lived != "":
                            user.countries_lived += "|"
                        user.countries_lived += country
```

Can be done more prettily, but I'm bad at that kind of stuff. Input welcome.

5) Do we want to rename "country" as "region"?


aapeliv (changes requested):
Thanks for taking this up! Some comments:

1. Let's use the alpha-3 three letter ISO code for regions, when I looked into it, it was more unique (the 2 letter codes sometimes get reshuffled). No need to include attribution for that.
1. Could you switch to storing the countries visited/lived in separate tables? Similar to language ability (but without the actual ability level): that would allow filtering based on country through a fast join and is just a lot cleaner than a manual array encoding in-db.
1. Please call them regions :)
1. Could you please copy the `regions.json` from the other pr verbatim? The file you've created has some encoding issues with non-ASCII chars (and is missing alpha3, see above).

aapeliv:
>     1. Under this schema there's no distinction, for example, between Scotland an England, they're both just UK. Not necessarily wrong, just something to consider. (Similarly, it doesn't have the various US states which I suppose are kind of analogous, so maybe it's for the best.)

Yeah, that's ok. States and sub-regions are not a very well standardised thing, and then we'd have to start adding sub-regions for whenever people wanted to have their country's states or whatever, etc. I mean we could have some overlapping regions if needed.



>     2\. Do we want to allow "other" countries somehow?

No, not needed at the moment.

>     3\. I didn't think it was worth massively re-doing countries as their own data type, or reworking how we treat them (i.e. as a "|" delineated string).

See comment above. This is important for fast filtering and e.g. grouping people by countries lived, etc, etc.

> Do we want to rename "country" as "region"?

Yes please.

ShmuelTreiger:
> 1. Let's use the alpha-3 three letter ISO code for regions, when I looked into it, it was more unique (the 2 letter codes sometimes get reshuffled). No need to include attribution for that.

Sure. I was going based off #470 for guidance.
Included attribution is more for ease of updating more than anything, but happy to remove

> 4. Could you please copy the `regions.json` from the other pr verbatim? 

Not sure what you mean by this. Which pr? You mean from the website I grabbed it from? I did copy it verbatim?

>The file you've created has some encoding issues with non-ASCII chars (and is missing alpha3, see above).

I noticed the ASCII issue, but figured it would get dealt with in the front end

aapeliv:
> > 1. Could you please copy the `regions.json` from the other pr verbatim?
> 
> Not sure what you mean by this. Which pr?

This one: https://github.com/Couchers-org/couchers/pull/476

PR #865, opened 2021-03-26
fix: login/signup tweaks

### LINKED ISSUE #857: Login/signup tweaks
- [ ] Padding between fields
- [ ] Move the left text on `CompleteSignup` up
- [ ] Signup map reloads page when pressing enter

Last one needs an ugly workaround - split the `<form>` into two, each side of the map (so the map isn't included). It will still function correctly because react-hook-form doesn't use this element, except for submitting on <Enter> presses.

Tay0108:
Closes #857 

I got rid of one unnecessary route and tweaked things from the issue. I think splitting sign up into two forms is hacky and should be fixed properly in the future but for soft beta release it should do.


lucaslcode:
> I think splitting sign up into two forms is hacky and should be fixed properly in the future

Agreed but I remember trying to "properly" do it for an hour with preventDefault and preventBubble (or whatever it's called) and events were originating from all kinds of strange places and I gave up in the end 😭

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -127,71 +130,76 @@ export default function CompleteSignupForm() {
...
-          <InputLabel className={authClasses.formLabel} htmlFor="location">
-            {LOCATION_LABEL}
-          </InputLabel>
+        <>
+          <form
+            className={`${authClasses.form} ${classes.firstForm}`}
+            onSubmit={completeSignup}
+          >
+            <InputLabel className={authClasses.formLabel} htmlFor="username">
+              Username
```
darrenvong: There's a few of these hardcoded strings here and the validation messages etc. Can you pull those out to the constants file while you're at these files?

darrenvong (dismissed):
Also agreed about this two form workaround being hacky. We should probably create an issue for investigation when we have more time post beta...

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
Tay0108: Sure, I'm on it!

app/frontend/src/features/auth/constants.tsx
```diff
@@ -29,9 +33,13 @@ export const LOCATION_LABEL = "Your location";
 export const LOGIN = "Log in";
 export const LOGIN_HEADER = "Welcome back!";
 export const LOGIN_PAGE = "Log in page";
+export const MAN = "Man";
```
lucaslcode: I know I originally said this but someone persuaded me it should indeed be "Male"

app/frontend/src/features/auth/constants.tsx
```diff
@@ -42,7 +50,20 @@ export const SELECT_LOCATION = "Please select your location.";
...
+
 export const SUBMIT = "Submit";
+
+export const THANKS = "Thanks!";
+
 export const USERNAME = "Username";
+export const USERNAME_REQUIRED = "Enter your username";
+export const USERNAME_TAKEN = "This username is taken.";
+
+export const WOMAN = "Woman";
```
lucaslcode: Female

lucaslcode (changes requested):

lucaslcode (approved):

PR #869, opened 2021-03-27
Fix same-file input

### LINKED ISSUE #866: tried uploading pic for my profile but came up blank (didn't load)
Subject: tried uploading pic for my profile but came up blank (didn't load)
Description:
My photo did not appear after selecting it. The first pic I chose worked fine but i changed my mind and pressed the x, selected another one but it didn't load.

Steps:
I clicked to add a pic, selected one and it appeared without any problems. I changed my mind, pressed the x, then selected another photo, but the photo didn't appear. it was just white. 

Results:
It should have shown the second photo just like the first one did.

Frontend version: develop-579a9506
Page: https://develop--frontend.preview.coucher.org/profile/edit#loc=13/43.0653/-89.42332
User ID: 7

lucaslcode:
closes #866

### DIFF
--- a/app/frontend/src/components/AvatarInput.tsx
+++ b/app/frontend/src/components/AvatarInput.tsx
@@ -140,8 +140,8 @@ export function AvatarInput({
           accept="image/*"
           id={id}
           type="file"
-          onClick={handleClick}
           onChange={handleChange}
+          onClick={handleClick}
           ref={inputRef}
         />
         <label className={classes.label} htmlFor={id} ref={field.ref}>
--- a/app/frontend/src/components/AvatarInput.tsx
+++ b/app/frontend/src/components/AvatarInput.tsx
@@ -74,7 +74,10 @@ export function AvatarInput({
   const [file, setFile] = useState<File | null>(null);
   const [readerError, setReaderError] = useState("");
   const mutation = useMutation<ImageInputValues, Error>(
-    () => (file ? service.api.uploadFile(file) : Promise.reject(INVALID_FILE)),
+    () =>
+      file
+        ? service.api.uploadFile(file)
+        : Promise.reject(new Error(INVALID_FILE)),
     {
       onSuccess: (data: ImageInputValues) => {
         field.onChange(data.key);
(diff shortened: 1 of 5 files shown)


darrenvong (commented):
Would be good add a test for this to reduce likelihood of regression in future

lucaslcode:
The test passes anyway, obviously it's a real-browser thing. Should I still put it in?

darrenvong:
Would you not expect it to pass if you've fixed it now? Or do you mean the test still passes even without the fix?

lucaslcode:
Yes, I tried running the test without the fix and it passed.

darrenvong:
Oh that's 💩 , such an odd bug considering we clear the file state on cancel ~~so it should work~~ 🤔  Never mind, I just read that file inputs can never be a controlled component [from the React docs](https://reactjs.org/docs/uncontrolled-components.html#the-file-input-tag)...

app/frontend/src/components/AvatarInput.tsx
```diff
@@ -131,7 +140,9 @@ export function AvatarInput({
           accept="image/*"
           id={id}
           type="file"
+          onClick={handleClick}
           onChange={handleChange}
+          ref={inputRef}
```
darrenvong, line 145: ~~Actually looking at [react hook form docs here on `useController`](https://react-hook-form.com/api#useController), shouldn't we be using the `ref` from the `field` object instead so value updates are managed by the lib rather than us having to manually clear the input value on cancel?~~

Uhh scrap what I said there (and the message I deleted earlier if you saw that lol as I realised inputs can't be controlled components) since we aren't actually submitting the file state as well, so in that case, I think I like the preview state being kept here for good separation of concerns.

darrenvong: In fact... do we even need a Controller? This is a native HTML input field so it's not a controlled component unlike Material UI fields. Wouldn't it be simpler to pass `register` in from the parent component with `useForm` and do `ref={register}` instead?

darrenvong:
https://github.com/testing-library/user-event/blob/f633a52ffc7bfdcdcf98a044815e35dd0c231ca1/src/upload.ts#L76-L80

This is why the test is still passing, `userEvent.upload` always fire the change event which I guess differs from the browser...

darrenvong (dismissed):
We should probably file a bug on the `userEvent` package and get them to fix that, so I'd say leave the test in with the skip. When it works, then it's a good regression test and also act as good documentation for the component's behaviour.

lucaslcode:
Still doesn't work: https://github.com/testing-library/user-event/issues/632

Let's leave it like this for now.

darrenvong (approved):
That's a shame the update fixed one problem but introduced another 😞  let's hope it will be fixed soon...

PR #874, opened 2021-03-28
Implement MessageUserButton

### LINKED ISSUE #768: Requests Beta UX

lucaslcode:
closes #768 

I actually think we should avoid state in `history`, but it's too convenient in this case....

Needs testing

### DIFF
--- a/app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
+++ b/app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
@@ -21,7 +21,13 @@ describe("CreateGroupChat with router state", () => {
   it("initially shows the create dialog with a user pre-filled", async () => {
     expect(screen.getByLabelText(FRIENDS)).toBeVisible();
     expect(screen.getByText(users[0].name)).toBeVisible();
-    expect(screen.getByText(NEW_CHAT, { selector: "h2" })).toBeVisible();
+    expect(
+      screen.queryByRole("heading", {
+        name: NEW_CHAT,
+        level: 2,
+        hidden: true,
+      })
+    ).toBeVisible();
   });
 });
 
@@ -33,6 +39,12 @@ describe("CreateGroupChat without router state", () => {
 
   it("doesn't initially show the create dialog", async () => {
     expect(screen.getByLabelText(FRIENDS)).not.toBeVisible();
-    expect(screen.getByText(NEW_CHAT, { selector: "h2" })).not.toBeVisible();
+    expect(
+      screen.queryByRole("heading", {
+        name: NEW_CHAT,
+        level: 2,
+        hidden: true,
+      })
+    ).toBeNull();
   });
 });
--- a/app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
+++ b/app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
@@ -1,0 +1,38 @@
+import { render, screen } from "@testing-library/react";
+import { FRIENDS, NEW_CHAT } from "features/messages/constants";
+import CreateGroupChat from "features/messages/groupchats/CreateGroupChat";
+import { User } from "pb/api_pb";
+import users from "test/fixtures/users.json";
+import { getHookWrapperWithClient } from "test/hookWrapper";
+
+describe("CreateGroupChat with router state", () => {
+  beforeEach(() => {
+    const { wrapper } = getHookWrapperWithClient({
+      initialRouterEntries: [
+        {
+          pathname: "create-group-chat-test",
+          state: { createMessageTo: users[0] as User.AsObject },
+        },
+      ],
+    });
+    render(<CreateGroupChat />, { wrapper });
+  });
+
+  it("initially shows the create dialog with a user pre-filled", async () => {
+    expect(screen.getByLabelText(FRIENDS)).toBeVisible();
+    expect(screen.getByText(users[0].name)).toBeVisible();
+    expect(screen.getByText(NEW_CHAT, { selector: "h2" })).toBeVisible();
+  });
+});
+
+describe("CreateGroupChat without router state", () => {
+  beforeEach(() => {
+    const { wrapper } = getHookWrapperWithClient();
+    render(<CreateGroupChat />, { wrapper });
+  });
+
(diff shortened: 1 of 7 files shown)


app/frontend/src/service/conversations.ts
```diff
@@ -125,3 +127,18 @@ export function markLastSeenGroupChat(
...
 }
+
+export async function getDirectMessage(userId: number) {
+  const req = new GetDirectMessageReq();
+  req.setUserId(userId);
+  try {
+    const res = await client.conversations.getDirectMessage(req);
+    return res.getGroupChatId();
+  } catch (e) {
+    if (e?.code === StatusCode.NOT_FOUND) {
```
darrenvong, line 138: Probably more of a backend API nitpick, but still feels a bit shady to me to throw an error on not found when it's not something the user could have done to cause this error... it might be cleaner if we can return some extra boolean like `hasExistingDm` or something that we can check on in the frontend and put the response under `oneof`? I think that's the proto equivalent of a union type right even though the generated code doesn't really do it?

darrenvong: Not saying you have to do it in this here though, maybe something to think about in the upcoming request/message API refactor?

darrenvong (commented):
This looks good to me generally - I like that this is re-using the create chat dialog UI so you can only do it in one place, which I think it's better for consistency than what my WIP stuff was doing, where we discussed we'd have to display a confirmation dialog before creating the chat.

Would be good to add some tests for this! It'll be easier to do it incrementally as we re-touch certain parts of the app again rather than trying to do it standalone which I think will be too boring lol

app/frontend/src/features/profile/actions/MessageUserButton.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+>;
+jest.mock("react-router-dom", () => {
+  const pushMock = jest.fn();
+  return {
+    ...jest.requireActual("react-router-dom"),
+    useHistory: () => ({
+      push: pushMock,
+    }),
+  };
+});
```
lucaslcode: I would prefer to pass the (mocked) history to `hookWrapper`, but I encountered an issue where it is impossible to pass both history and state, because you need to use `<Router history={...}>`, but `createMemoryHistory()` doesn't support state in the entries. It is fixed in `history` v5, but `react-router-dom` uses v4.

app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+        },
+      ],
+    });
+    render(<CreateGroupChat />, { wrapper });
+  });
+
+  it("initially shows the create dialog with a user pre-filled", async () => {
+    expect(screen.getByLabelText(FRIENDS)).toBeVisible();
+    expect(screen.getByText(users[0].name)).toBeVisible();
+    expect(screen.getByText(NEW_CHAT, { selector: "h2" })).toBeVisible();
```
darrenvong:
```suggestion
    expect(screen.getByRole('heading', { level: 2, name: NEW_CHAT })).toBeVisible();
```
Better to use `getByRole` here so it checks that it's accessible through the accessibility tree by things like screen readers.

app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+describe("CreateGroupChat without router state", () => {
+  beforeEach(() => {
+    const { wrapper } = getHookWrapperWithClient();
+    render(<CreateGroupChat />, { wrapper });
+  });
+
+  it("doesn't initially show the create dialog", async () => {
+    expect(screen.getByLabelText(FRIENDS)).not.toBeVisible();
+    expect(screen.getByText(NEW_CHAT, { selector: "h2" })).not.toBeVisible();
```
darrenvong: Same comment as above

app/frontend/src/features/profile/actions/MessageUserButton.test.tsx
darrenvong: Do you mean if `history` had worked, you'd probably render the other component side-by-side and have it under `Router`? What's wrong with using `initialEntries` in `MemoryRouter`? Shouldn't it direct you to the right page with the state if you put them under the right `Route`s?

darrenvong (dismissed):
Some minor comments on the tests, but good work here!

app/frontend/src/features/messages/groupchats/CreateGroupChat.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  it("initially shows the create dialog with a user pre-filled", async () => {
+    expect(screen.getByLabelText(FRIENDS)).toBeVisible();
+    expect(screen.getByText(users[0].name)).toBeVisible();
+    expect(
+      screen.queryByRole("heading", {
+        name: NEW_CHAT,
+        level: 2,
+        hidden: true,
```
darrenvong: Is this needed if the heading is visible? 🤔

darrenvong (approved):

PR #882, opened 2021-03-30
Fix profile naming convention and missing edit fields

### LINKED ISSUE #863: Profile discrepancies when editing
Some labels are mismatched, and some fields are displayed in profile/home but missing in the editor.

Original bug report:
Subject: "Who I am" field shows as "About" when editing profile
Description:
The first major text box when editing my profile is named "About." This is misleading, because the title appearing on the profile is "Who I am"

Steps:
Compared the section titles in the "Edit Profile" page with the "Profile page"

Results:
The title of the "About" section in the "Edit profile" page should be changed to "Who I am"

Frontend version: develop-7addcde4
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36
Page: https://develop--frontend.preview.coucher.org/profile/edit#loc=13/43.05851/180
User ID: 25

justjooshing:
For [#863](https://github.com/Couchers-org/couchers/issues/863).

I also made it a bit nicer on desktop (imo, but feel free to disagree), and will do something similar to the Edit Profile page this afternoon
image
image

Edit profile page on desktop currently (same structure as on mobile)
image

Profile page/home tab
image


justjooshing:
One thing I did just realise though looking again at the designs, is the Private/Shared Space label. 
I initially removed it, but see that Sleeping Arrangement isn't a textbox in the design rather than options, but we're using the user.sleepingArrangement options for the Private/Shared space label.

Original:
image
image

Design: 
image

Should we rename them to user.space.value instead?

lucaslcode:
Let's avoid renaming fields in this PR because it makes all the backend tests fail. 

[The user has](https://github.com/Couchers-org/couchers/blob/de503afba00120583f9be71bfbdcde720a103dd7/app/pb/api.proto#L232) the sleeping arrangements enum and another free text sleeping details field. That matches the design, right? We can keep the current "Private / Shared space" label and add a larger "Sleeping arrangement details" text box below.

justjooshing:
[edit: damn the commit above should say re-add, not read 🙄 ]

Cool done

I also noticed that none of these were showing, but I'm working on it in a different branch since I wasn't sure how urgent it was/how much time we want to spend meeting the design compared to this high priority fix
image

justjooshing:
> FAIL src/features/profile/view/ProfilePage.test.tsx (13.411 s)
>   ● Profile page › when viewing another user's profile › and the 'report user' option is clicked › reports the user successfully
>     Unable to find role="alert"

Hmm, not sure what I changed to mess that up though

lucaslcode (commented):
Tests worked for me locally, it will probably work next commit.

Don't forget to extract the labels in to constants while you're here 🙃

justjooshing:
Edit profile page now has everything centered, and the multi input things like languages/countries lived etc are grouped more nicely
image
image

Looks almost as it was on mobile
image
image

I'm not convinced on the name/avatar placement of either though - so im happy for design suggestions!

justjooshing:
The only thing currently still missing is the list of countries visited/lived in

aapeliv:
> The only thing currently still missing is the list of countries visited/lived in

Are you blocked by the backend for this stuff? Could we disable/do something quick and dirty for it now. This is coming soon in the backend with languages/regions being worked on (should be ready for frontend work in a few days), but it'd be great to get soft beta out before that.

justjooshing:
Didn't get a chance to finish it off today, but I'll be able to push it
first thing tomorrow morning!

justjooshing:
My travels section without the map:
image

lucaslcode (approved):

PR #885, opened 2021-03-31
[MRG] Diff schema created by models.py and migrations

kalvdans:
I managed to get some rudimentary sorting of pg_dump output, but it still displays some differences.

```diff
@@ -33,25 +33,6 @@

 
 COMMENT ON EXTENSION unaccent IS 'text search dictionary that removes accents';
 
-
---
--- Name: alembic_version alembic_version_pkc; Type: CONSTRAINT; Schema: public; Owner: chn
---
-
-ALTER TABLE ONLY public.alembic_version
-    ADD CONSTRAINT alembic_version_pkc PRIMARY KEY (version_num);
-
-
---
--- Name: alembic_version; Type: TABLE; Schema: public; Owner: chn
---
-
-CREATE TABLE public.alembic_version (
-    version_num character varying(32) NOT NULL
-);
-
-
-ALTER TABLE public.alembic_version OWNER TO chn;
 
 --
 -- Name: background_jobs id; Type: DEFAULT; Schema: public; Owner: chn
@@ -1267,6 +1248,13 @@

 
 
 --
+-- Name: idx_users_geom; Type: INDEX; Schema: public; Owner: chn
+--
+
+CREATE INDEX idx_users_geom ON public.users USING gist (geom);
+
+
+--
 -- Name: initiated_uploads fk_initiated_uploads_initiator_user_id_users; Type: FK CONSTRAINT; Schema: public; Owner: chn
 --
 
@@ -2298,7 +2286,7 @@

 --
 
 CREATE TABLE public.sessions (
-    api_calls integer DEFAULT 0 NOT NULL,
+    api_calls integer NOT NULL,
     created timestamp with time zone DEFAULT now() NOT NULL,
     deleted timestamp with time zone,
     expiry timestamp with time zone DEFAULT (now() + '90 days'::interval) NOT NULL,
@@ -2356,17 +2344,17 @@

 CREATE FUNCTION public.slugify(text text) RETURNS text
     LANGUAGE sql IMMUTABLE STRICT
     AS $_$
-    SELECT regexp_replace(
+SELECT regexp_replace(
+  regexp_replace(
+    substring(
       regexp_replace(
-        substring(
-          regexp_replace(
-            lower(unaccent("text")),
-            '[^a-z0-9\-_]+', '-', 'gi'
-          ) from 0 for 64
-        ), '\-+$', ''
-      ), '^\-', ''
-    );
-    $_$;
+        lower(unaccent("text")),
+        '[^a-z0-9\\-_]+', '-', 'gi'
+      ) from 0 for 64
+    ), '\-+$', ''
+  ), '^\-', ''
+);
+$_$;
 
 
 ALTER FUNCTION public.slugify(text text) OWNER TO chn;
```


kalvdans:
Who would guess postgres stores actual text for functions and not the parsed AST.... :/

aapeliv:
> Who would guess postgres stores actual text for functions and not the parsed AST.... :/

How else would you support arbitrary languages? You can write postgres functions in python with pl/python O.o

aapeliv:
I'll look more tomorrow. It looks like CI doesn't have pg_dump

aapeliv:
@kalvdans: you're the debian guy. It seems buster doesn't have postgres 13 utilities in their packages;

https://packages.debian.org/search?keywords=postgres+client&searchon=names&suite=buster&section=all

We're using the 13 server, and postgres docs state "pg_dump cannot dump from PostgreSQL servers newer than its own major version; it will refuse to even try" (https://www.postgresql.org/docs/current/app-pgdump.html)

kalvdans:
> @kalvdans: you're the debian guy. It seems buster doesn't have postgres 13 utilities in their packages;
> 
> https://packages.debian.org/search?keywords=postgres+client&searchon=names&suite=buster&section=all
> 
> We're using the 13 server, and postgres docs state "pg_dump cannot dump from PostgreSQL servers newer than its own major version; it will refuse to even try" (https://www.postgresql.org/docs/current/app-pgdump.html)
```
chn@rosa:~$ dpkg --search /usr/bin/pg_dump
postgresql-client-common: /usr/bin/pg_dump
```

aapeliv:
That's postgres 11 though?

kalvdans:
> That's postgres 11 though?

I'm using Debian Bullseye which have postgres 13. I don't know how forwards compatible pg_dump is, it doesn't say on the manual page.

kalvdans:
> I don't know how forwards compatible pg_dump is, it doesn't say on the manual page.

It's not.
```
pg_dump: server version: 13.0 (Debian 13.0-4); pg_dump version: 11.11 (Debian 11.11-0+deb10u1)
pg_dump: aborting because of server version mismatch
```

aapeliv:
It doesn't look like there's a stock python docker container with bullseye :(

kalvdans:
Do you think it could work by running `docker run postgis/postgis:13-3.0 pg_dump` instead of `pg_dump`?

aapeliv:
Yeah, we could definitely run it just as either a shell script, or a CI script (basically a shell script) and invoke pg_dump/docker

kalvdans:
> Yeah, we could definitely run it just as either a shell script, or a CI script (basically a shell script) and invoke pg_dump/docker

What do you mean? I thought about having the pg_dump command as a config variable, and define it as `docker run postgis/postgis:13-3.0 pg_dump` in the CI and just `pg_dump` when testing locally.

aapeliv:
I mean we can run it in CI as commands like in the tests (https://github.com/Couchers-org/couchers/blob/81781a7bb9314bec841188474fda06fe574082be/.gitlab-ci.yml#L173-L208)

But I think it's better to do as you're suggesting, just make the pg_dump cmd a variable

kalvdans:
It didn't work... The command `docker` was not found inside the container running the unit tests. Of course.

aapeliv:
Tried to just install `pg_dump` v13 from the postgres package repos in the test script, let's see if that works:

https://github.com/Couchers-org/couchers/commit/c85281de3fca47ff2135dd1bb86ebdb6d1379e6b, https://github.com/Couchers-org/couchers/commit/fdd53218ff772a0bf31c9df792663a39d12acf50, https://github.com/Couchers-org/couchers/commit/83b7a36aaeeabf8f932052743bc308cde1096f3c, https://github.com/Couchers-org/couchers/commit/aab03ec70d1bc3ae456e344eac948dfcb039b5a9

kalvdans:
@aapeli now only a single failure left, which I don't know how to solve:
```diff
--- migrations
+++ model
@@ -1252,6 +1252,13 @@
 --
 
 CREATE INDEX idx_page_versions_geom ON public.page_versions USING gist (geom);
+
+
+--
+-- Name: idx_users_geom; Type: INDEX; Schema: public; Owner: postgres
+--
+
+CREATE INDEX idx_users_geom ON public.users USING gist (geom);
 
 
 --
 ```

**Comment by kalvdans** [comment]
TODO from the backend meeting: create new migrations instead of changing existing ones.

**Comment by kalvdans** [comment]
All tests passes, ready for a new review @aapeliv! Will create another PR for the slugify escape bug

File: app/backend/src/tests/test_db.py [hunk 1]
```diff
@@ -77,3 +85,57 @@ def test_get_parent_node_at_location(testing_communities):
...
         assert get_parent_node_at_location(session, create_1d_point(51)).id == w_id
+
+
+def pg_dump():
+    return subprocess.run(
+        ["pg_dump", "-s", config["DATABASE_CONNECTION_STRING"]], stdout=subprocess.PIPE, encoding="ascii", check=True
+    ).stdout
+
+
+def sort_pg_dump_output(output):
```
aapeliv, line 96: Could you add a docstring for this function?

app/backend/src/tests/test_db.py
```diff
@@ -77,3 +85,57 @@ def test_get_parent_node_at_location(testing_communities):
...
+
+def test_sort_pg_dump_output():
+    assert sort_pg_dump_output(" (\nb,\nc,\na\n);\n") == " (\na,\nb,\nc\n);\n"
+
+
+def strip_leading_whitespace(lines):
+    return [s.lstrip() for s in lines]
+
+
+def test_migrations():
```
aapeliv, line 126: Could you add a comment/docstring as to what this test is doing?

aapeliv (changes requested):
Thanks, this is looking great!

Could you just add comments to `sort_pg_dump_output` and `test_migrations`. It's hard to decipher what's going on due to the regexpes and lambda functions :)... I guess you're basically sorting each section (delimited by `--`?) of the produced dump?

aapeliv (approved):
LGTM!

PR #904, opened 2021-04-01
[MRG] Backend/feat/regions

ShmuelTreiger:
Picks up from #864


app/backend/src/couchers/models.py
```diff
@@ -174,6 +172,14 @@ class User(Base):
...
     new_email_token_expiry = Column(DateTime(timezone=True), nullable=True)
 
+    @property
+    def regions_visited(self):
+        return "|".join(region.region_code for region in self._regions_visited)
+
+    @property
+    def regions_lived(self):
+        return "|".join(region.region_code for region in self._regions_lived)
+
```
aapeliv: There's no need to use this funky pipe-delimited thing now that we're storing things properly.

You can remove both of these properties and use the `_regions_visited` directly (remove underscore as in later comment)

app/backend/src/couchers/models.py
```diff
@@ -287,6 +293,28 @@ class LanguageAbility(Base):
...
 
+class RegionsVisited(Base):
+    __tablename__ = "regions_visited"
+    __table_args__ = (UniqueConstraint("user_id", "region_code"),)
+
+    id = Column(BigInteger, primary_key=True)
+    user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
+    region_code = Column(String(3), nullable=False)
+
+    user = relationship("User", backref="_regions_visited")
```
aapeliv: No underscore in backref

app/backend/src/couchers/models.py
```diff
@@ -287,6 +293,28 @@ class LanguageAbility(Base):
...
+
+class RegionsLived(Base):
+    __tablename__ = "regions_lived"
+    __table_args__ = (UniqueConstraint("user_id", "region_code"),)
+
+    id = Column(BigInteger, primary_key=True)
+    user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
+    region_code = Column(String(3), nullable=False)
+
+    user = relationship("User", backref="_regions_lived")
```
aapeliv: Ditto

app/backend/src/couchers/servicers/api.py
```diff
@@ -295,11 +298,35 @@ def UpdateProfile(self, request, context):
...
-            if request.countries_visited.exists:
-                user.countries_visited = "|".join(request.countries_visited.value)
+            if request.regions_visited.exists:
+                for region in user._regions_visited:
+                    session.delete(region)
+
+                for region in request.regions_visited.value:
+                    if not region_is_allowed(region):
+                        context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_REGION)
+                    else:
```
aapeliv: No need to have an `else` here, it keeps the indentation down. If the above `abort` gets hit, it won't continue here, so you can get rid of the `else` and de-indent the next statement

app/backend/src/couchers/servicers/api.py
```diff
@@ -295,11 +298,35 @@ def UpdateProfile(self, request, context):
...
-            if request.countries_lived.exists:
-                user.countries_lived = "|".join(request.countries_lived.value)
+            if request.regions_lived.exists:
+                for region in user._regions_lived:
+                    session.delete(region)
+
+                for region in request.regions_lived.value:
+                    if not region_is_allowed(region):
+                        context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_REGION)
+                    else:
```
aapeliv: Ditto

app/backend/src/couchers/servicers/api.py
```diff
@@ -745,8 +772,8 @@ def user_model_to_pb(db_user, session, context):
             api_pb2.LanguageAbility(code=ability.language_code, fluency=fluency2api[ability.fluency])
             for ability in db_user.language_abilities
         ],
-        countries_visited=db_user.countries_visited.split("|") if db_user.countries_visited else [],
-        countries_lived=db_user.countries_lived.split("|") if db_user.countries_lived else [],
+        regions_visited=db_user.regions_visited.split("|") if db_user.regions_visited else [],
+        regions_lived=db_user.regions_lived.split("|") if db_user.regions_lived else [],
```
aapeliv: You can get rid of the pipe-delimited stuff, just use

`=[region.region_code for region in user.regions_visited]`

app/backend/src/tests/test_api.py
```diff
@@ -31,42 +31,48 @@ def test_ping(db):
...
-    # TODO: this list serialisation will be fixed hopefully soon
-    assert res.user.countries_visited == user.countries_visited.split("|")
-    assert res.user.countries_lived == user.countries_lived.split("|")
-    assert res.user.additional_information == user.additional_information
-
-    assert res.user.friends == api_pb2.User.FriendshipStatus.NA
-    assert len(res.user.mutual_friends) == 0
+    # Need the user to have a live database connection to pull region and language data
+    # Alternatively, could remove these from api.ping
+    # TODO on that note, test for language
```
aapeliv: You can do this easier/cleaner by either:

1. Just checking what the user's regions are when it's created, I guess it's just empty?
2. Have one session scope to grab the user's regions lived/visited and then store them in a list and check against that. You'll need to do the same as you do here, but call it something other than `user`, since that will mess with the sqlalchemy stuff, as I think you noticed.

app/backend/src/tests/test_api.py
```diff
@@ -190,10 +212,14 @@ def test_update_profile(db):
...
-        assert "UK" in user.countries_lived
-        assert "Aus" in user.countries_lived
+        regions_visited = {"CXR", "NAM"}
+        regions_lived = {"USA", "ITA"}
+        for region in user.regions_visited:
+            regions_visited.remove(region)
+        for region in user.regions_lived:
+            regions_lived.remove(region)
+        assert not regions_visited
+        assert not regions_lived
```
aapeliv: You can do something like

```
assert set(user.regions_visited) == {"CXR", "NAM"}
```

app/backend/src/couchers/errors.py
```diff
@@ -44,6 +44,7 @@
 INVALID_MESSAGE = "Invalid message."
 INVALID_NAME = "Name not supported."
 INVALID_RECIPIENTS = "Invalid recipients list."
+INVALID_REGION = "Invalid region code."
```
aapeliv: Change to "Invalid region.", the "code" will confuse users if they ever see this

aapeliv (dismissed):
Thanks, this is looking very close to ready.

Please don't change code not related to regions in the backend (the communities are not the same as regions lived/visited, please remove the `TODO`s from unrelated tests), on the landing site (that has nothing to do with this) or in the frontend (that needs to be updated anyway because their code has arbitrary country names sprinkled around, let them clean up the whole thing, it's easier to find where stuff needs to change if it's all breaking :P)

app/backend/src/couchers/models.py
ShmuelTreiger: I figured leaving it like this would minimize refactoring across the app. I'll change it.

aapeliv: I think you're basically already modifying everywhere where it's used? Either way, it should be refactored out :)

app/backend/src/tests/test_api.py
ShmuelTreiger: 1. I added in regions visited and lived to generate_user(). Should I not? I figured those are useful properties to include
2. That makes more sense, though I didn't have any issues with variable renaming if thats what you meant.

app/backend/src/couchers/models.py
ShmuelTreiger: No, def places where I was able to leave the code as it was (except to change country to region)

app/pb/api.proto
```diff
@@ -231,8 +231,8 @@ message User {
   string my_travels = 104;
   string things_i_like = 105;
   string about_place = 16;
-  repeated string countries_visited = 18;
-  repeated string countries_lived = 19;
+  repeated string regions_visited = 40;
+  repeated string regions_lived = 41;
```
aapeliv, line 235: Ahh, this is good to change the field ids!

site/markdown/solutions/profit-and-incentives.md
```diff
@@ -28,4 +28,4 @@ We can run ads on the website and phone apps. It would provide a predictable, re
...
+A frequently raised opinion is the idea of running the platform as a co-op. Users could voluntarily buy ownership of the platform as a one-off or subscription payment, allowing them to vote on decisions or leaders. You can see the enthusiasm for this idea on the [comments of Couchsurfing&#8482;'s official response to the outrage over the forced contributions](https://blog.couchsurfing.com/we-hear-you/#comments). The platform would be entirely community-owned, and so the incentives would run in direct line with the community. Potential downsides are that it may be hard to predict the dynamics of widespread ownership (for instance an undesired political element may evolve), and it may disadvantage users or communities with less disposable income.
```
aapeliv: Please don't edit this at all

aapeliv:
Combined into #926

PR #905, opened 2021-04-01
Add ability to respond to friend request on profile

### LINKED ISSUE #591: Can't accept friend request on user's page
Subject: Can't accept friend request on user's page
Description:
Upon visiting a user's page who has sent you a friend request, it says "Pending friend request..." instead of having some kind of button to accept a request.

You have to go to Connections -> Friends to accept requests.

Steps:

Results:

Backend version: develop-7a5e0536
Frontend version: develop-7a5e0536
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:86.0) Gecko/20100101 Firefox/86.0
Page: https://develop--frontend.preview.coucher.org/user/aapeli
User ID: 4

javiwq:
Closes #591 

On another user's profile, iff there is a pending friend request, checks if request has been sent or recieved. If sent, shows a disabled 'pending' button (like original implementation). If received, shows 'pending' button that opens a menu with accept or reject.


app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+import PendingFriendRequestSent from "./PendingFriendRequestSent";
+import RespondToFriendRequestProfileButton from "./RespondToFriendRequestProfileButton";
+
+interface PendingRequestProps {
+  userId: number;
+  setMutationError: SetMutationError;
+}
+
+function CheckSentRequests(userId: number) {
+  const friendRequestsSent = useFriendRequests("sent").data;
```
lucaslcode: Use `{ data }` like you did for the other one

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const friendRequestToUser = friendRequestsSent?.find(
+    (e) => e.userId === userId
+  );
+  if (friendRequestToUser !== undefined) {
+    const request = friendRequestToUser;
+    return request;
+  } else {
+    return false;
+  }
```
lucaslcode:
```suggestion
  return friendRequestToUser ? friendRequestToUser : false;
```
You could argue that what you wrote is more clear, and it's better for code to be clear and easily understandable rather than just have fewer lines. However in js/ts we abuse truthiness/falsiness pretty often so I would say the suggestion is usual enough to make it more concise like this.

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const friendRequestsReceived = data;
+  const friendRequestFromUser = friendRequestsReceived?.find(
+    (e) => e.userId === userId
+  );
+  if (friendRequestFromUser !== undefined) {
+    const request = friendRequestFromUser;
+    return request;
+  } else {
+    return false;
+  }
```
lucaslcode: as above

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
```
lucaslcode: These functions should be hooks. `useRequestSentToUser` / `useRequestReceivedFromUser` (or a better name, if you can think of one). The reason is they contain hooks themselves, and hooks and only be in other hooks or components (you probably had a lint error which went away when you put a capital letter. Sentence case is for components only, and hooks are allowed in components)

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+          state={request.state}
+          setMutationError={setMutationError}
+        />
+      );
+    }
+  } else if (CheckSentRequests(userId) !== false) {
+    return <PendingFriendRequestSent />;
+  }
+  return null;
+}
```
lucaslcode: Since these functions are actually hooks, they can't be used in branching statements (like `if`).  Just put your two hooks at the top and then use branching after to return the components you want (usually with a ternary operator rather than an if statement).

lucaslcode: (The reason hooks can't be used in branching, is because React expects them to always appear in the same order behind the scenes, or it will get their values confused. [More info here](https://www.netlify.com/blog/2019/03/11/deep-dive-how-do-react-hooks-really-work/).)

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+interface RespondToFriendRequestProfileButtonProps {
+  friendRequestId: number;
+  state: FriendRequest.FriendRequestStatus;
+  setMutationError: SetMutationError;
+}
+
+export const RESPOND_TO_FRIEND_REQUEST_MENU_ID =
+  "respond-to-friend-request-actions-menu";
+
+function RespondToFriendRequestProfileButton({
```
lucaslcode: `RespondFriendRequestButton` is fine, since it's in the `profile` feature folder

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+        onClose={handleClose("menu")}
+        open={isOpen.menu}
+      >
+        {" "}
+        {isLoading || isSuccess ? (
+          <CircularProgress />
+        ) : (
+          <>
+            <MenuItem onClick={handleClick("accepted")}>
+              {
```
lucaslcode: These braces (and similar below) aren't necessary (are they??)

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+                <IconButton
+                  aria-label="Accept request"
```
lucaslcode: move to a constant

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  }}
+                >
+                  <CheckIcon />
+                </IconButton>
+              }
+            </MenuItem>
+            <MenuItem onClick={handleClick("rejected")}>
+              {
+                <IconButton
+                  aria-label="Decline request"
```
lucaslcode: move to a constant

app/frontend/src/features/profile/actions/PendingFriendRequestSent.tsx
```diff
@@ -0,0 +1,25 @@
...
+    backgroundColor: theme.palette.grey[100],
+  },
+}));
+
+export default function PendingFriendRequestSent() {
+  const classes = useStyles();
+  return (
+    <Button
+      startIcon={<PersonAddIcon />}
+      className={classNames({ [classes.disabledButton]: true })}
```
lucaslcode: `classNames(classes.disabledButton)` works. But if there's only one class, just use the class directly: 
```suggestion
      className={classes.disabledButton}
```

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {isLoading || isSuccess ? (
+          <CircularProgress />
+        ) : (
+          <>
+            <MenuItem onClick={handleClick("accepted")}>
+              {
+                <IconButton
+                  aria-label="Accept request"
+                  onClick={() => {
+                    reset();
```
lucaslcode: What does this reset do?

lucaslcode: I see it's copied from the other page. Try without - `reset` is used to clear the error and data, but since we're using `setMutationError` I don't think it's needed. If it works, remove it from the other page too :)

lucaslcode: I usually fake an error in chrome dev tools by going network->throttling->offline, then clicking

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {PENDING}
+      </Button>
+      <Menu
+        anchorEl={menuAnchor.current}
+        id={RESPOND_TO_FRIEND_REQUEST_MENU_ID}
+        onClose={handleClose("menu")}
+        open={isOpen.menu}
+      >
+        {" "}
+        {isLoading || isSuccess ? (
```
lucaslcode: On success I guess the parent component removes this one. Maybe add a comment here just to make it clear.

app/frontend/src/features/profile/actions/PendingFriendRequestSent.tsx
lucaslcode: However, I'm surprised this class is necessary - shouldn't it be appropriately styled anyway if `disabled={true}`?

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+          friendRequestId={request.friendRequestId}
+          state={request.state}
+          setMutationError={setMutationError}
+        />
+      );
+    }
+  } else if (CheckSentRequests(userId) !== false) {
+    return <PendingFriendRequestSent />;
+  }
+  return null;
```
lucaslcode: What happens while this component is checking if a request is sent or received? We don't want elements popping in

lucaslcode (changes requested):
This is a great start!

If you feel up to it, it would be good to create `Overview.test.tsx` that checks if the right buttons are displayed in the right cases.

lucaslcode:
Also - there's a syntax error with the build. I suspect something happened with the merge, maybe revert that commit and try again (or I can try)

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -95,9 +96,17 @@ export default function Overview({ user, setIsRequesting }: OverviewProps) {
...
-            {user.friends !== User.FriendshipStatus.FRIENDS ? (
-              <AddFriendButton
-                isPending={user.friends === User.FriendshipStatus.PENDING}
+            {user.friends !== User.FriendshipStatus.FRIENDS &&
+              user.friends !== User.FriendshipStatus.PENDING && (
+                <AddFriendButton
+                  userId={user.userId}
+                  setMutationError={setMutationError}
+                />
+              )}
```
```suggestion
            {user.friends === User.FriendshipStatus.NOT_FRIENDS && (
                <AddFriendButton
                  userId={user.userId}
                  setMutationError={setMutationError}
                />
             )}
```
You can check the user is `NOT_FRIENDS` instead to simplify the logic.

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+import RespondToFriendRequestProfileButton from "./RespondToFriendRequestProfileButton";
+
+interface PendingRequestProps {
+  userId: number;
+  setMutationError: SetMutationError;
+}
+
+function CheckSentRequests(userId: number) {
+  const friendRequestsSent = useFriendRequests("sent").data;
+  const friendRequestToUser = friendRequestsSent?.find(
```
darrenvong: You can determine if you've sent a request to the user through the `user.friends` field (which you get through the user object when you're on the profile page component) and so all of the logics here to figure out from list of friend request sent is unnecessary.

app/frontend/src/features/profile/actions/PendingRequest.tsx
darrenvong: We could even simplify this further by doing `!!friendRequestToUser`? That way, it's returning the same data type (as a boolean) which makes the intent clearer that we're checking if a friend request has been sent or not, rather than interested in the full object itself in this context.

I think this logic is only needed for checking request received based on my other comment above.

app/frontend/src/features/profile/actions/PendingRequest.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const request = friendRequestToUser;
+    return request;
+  } else {
+    return false;
+  }
+}
+
+function CheckReceivedRequests(userId: number) {
+  const { data } = useFriendRequests("received");
+  const friendRequestsReceived = data;
```
darrenvong:
```suggestion
  const { data: friendRequestReceived } = useFriendRequests("received");
```
You can directly rename in an object destructure rather than doing it separately

app/frontend/src/features/profile/actions/PendingFriendRequestSent.tsx
javiwq: You're right, I had copy pasted from previous code but indeed the class is not necessary! It styles anyways with `disabled={true}`

app/frontend/src/features/connections/friends/AddFriendButton.tsx
```diff
@@ -49,16 +39,13 @@ export default function AddFriendButton({
   return (
     <Button
       startIcon={<PersonAddIcon />}
-      className={classNames({ [classes.disabledButton]: isPending })}
-      disabled={isPending}
+      className={classNames()}
```
darrenvong:
```suggestion
```
Redundant code since no class names are needed

app/frontend/src/features/connections/friends/AddFriendButton.tsx
```diff
@@ -1,33 +1,23 @@
-import { makeStyles } from "@material-ui/core";
 import classNames from "classnames";
```
darrenvong: You may want to pay attention to lint warnings to see if this is still needed or not. Remove if unused, which I think will be the case

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
javiwq: Actually that part wasn't necessary to have in there now that I'm testing around. On success in fact the entire button disappears!

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
javiwq: No, you're right.

app/frontend/src/features/profile/actions/PendingRequest.tsx
javiwq: I changed it to the suggestion by @lucaslcode because the purpose is not just to check whether a request has been sent or received, but also return the friend request object so its `friendRequestId` and `state` properties can be accessed.

app/frontend/src/features/profile/actions/PendingRequest.tsx
javiwq: Thanks so much for all your feedback. Question about this - by "all the logics", do you refer to the `?` in `friendRequestsSent?.find(`? That is there because `useFriendRequests()` returns an object which is possibly undefined.

app/frontend/src/features/profile/actions/RespondFriendRequestButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+interface RespondToFriendRequestProfileButtonProps {
+  friendRequestId: number;
+  state: FriendRequest.FriendRequestStatus;
+  setMutationError: SetMutationError;
+}
+
+export const RESPOND_TO_FRIEND_REQUEST_MENU_ID =
+  "respond-to-friend-request-actions-menu";
+
+function RespondToFriendRequestProfileButton({
```
darrenvong:
```suggestion
function RespondToFriendRequestButton({
```

app/frontend/src/features/profile/actions/RespondFriendRequestButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { CheckIcon, CloseIcon, PersonAddIcon } from "components/Icons";
+import Menu, { MenuItem } from "components/Menu";
+import { PENDING } from "features/connections/constants";
+import { ACCEPT_REQUEST, DECLINE_REQUEST } from "features/profile/constants";
+import { SetMutationError } from "features/connections/friends";
+import useRespondToFriendRequest from "features/connections/friends/useRespondToFriendRequest";
+import { FriendRequest } from "pb/api_pb";
+import React, { useRef, useState } from "react";
+
+interface RespondToFriendRequestProfileButtonProps {
```
darrenvong:
```suggestion
interface RespondToFriendRequestButtonProps {
```

app/frontend/src/features/profile/actions/RespondFriendRequestButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export const RESPOND_TO_FRIEND_REQUEST_MENU_ID =
+  "respond-to-friend-request-actions-menu";
+
+function RespondToFriendRequestProfileButton({
+  friendRequestId,
+  state,
+  setMutationError,
+}: RespondToFriendRequestProfileButtonProps) {
```
darrenvong:
```suggestion
}: RespondToFriendRequestButtonProps) {
```

app/frontend/src/features/profile/actions/RespondFriendRequestButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+                setMutationError,
+              });
+            }}
+          >
+            <CloseIcon />
+          </IconButton>
+        </MenuItem>
+      </Menu>
+    </>
+  ) : null;
```
darrenvong:
```suggestion
  return (
    <>
      <Button
        startIcon={<PersonAddIcon />}
        innerRef={menuAnchor}
        onClick={handleClick("menu")}
        aria-controls={RESPOND_TO_FRIEND_REQUEST_MENU_ID}
      >
        {PENDING}
      </Button>
      <Menu
        anchorEl={menuAnchor.current}
        id={RESPOND_TO_FRIEND_REQUEST_MENU_ID}
        onClose={handleClose("menu")}
        open={isOpen.menu}
      >
        <MenuItem onClick={handleClick("accepted")}>
          <IconButton
            aria-label={ACCEPT_REQUEST}
            onClick={() => {
              reset();
              respondToFriendRequest({
                accept: true,
                friendRequestId,
                setMutationError,
              });
            }}
          >
            <CheckIcon />
          </IconButton>
        </MenuItem>
        <MenuItem onClick={handleClick("rejected")}>
          <IconButton
            aria-label={DECLINE_REQUEST}
            onClick={() => {
              reset();
              respondToFriendRequest({
                accept: false,
                friendRequestId,
                setMutationError,
              });
            }}
          >
            <CloseIcon />
          </IconButton>
        </MenuItem>
      </Menu>
    </>
  );
```
This check is unnecessary - since it has already been done by the parent component so we know there is a pending friend request here at this point.

app/frontend/src/features/profile/actions/RespondToFriendRequestProfileButton.tsx
javiwq: Indeed I took it from the `FriendRequestsReceived` implementation, but after faking some errors and trying it out `reset` does not seem to be needed :)

app/frontend/src/features/profile/actions/PendingRequest.tsx
javiwq: I attempted to add a circular progress element now while it is loading

javiwq:
Thank you both for your comments! I went through them and cleaned up the code. I just saw #955 though, so I will change the implenetation to reflect the API change (which I think @aapeliv is working on as I type this) next.

darrenvong (changes requested):
A few other loose things I saw earlier while I was looking too early 🙈

app/frontend/src/features/profile/actions/PendingRequest.tsx
darrenvong: Never mind, I got a bit confused by all of these logic as I thought everything is available on the `friends` field in the user object until I found that's not the case, hence I opened the issue you saw Aapeli looked at

darrenvong:
> I will change the implenetation to reflect the API change (which I think @aapeliv is working on as I type this) next.

Sorry about that! When I understood why you had to write two hooks to determine the pending status, I didn't like the fact we need to do so much data wrangling on the frontend to figure that information out - this should have been available from the API. The updated API should remove the need for you to keep the hooks altogether :)

javiwq:
No need to be sorry @darrenvong! Indeed this is better, and in retrospect I should've raised the issue myself when I realised I'd have to make a work around. Anyways, I applied your second round of comments, thanks again for the review! I will make the bigger change to the new API tomorrow :)

darrenvong:
Thanks for understanding! We are all learning here each time we write code and reviewing each other's work! You will know for the future to ask when you've done this a few times :)

app/frontend/src/features/profile/actions/PendingRequest.tsx
darrenvong: Ha also ignore what I said here then, this may soon be irrelevant with the updated API!

javiwq:
Closing this pull request, I think it will be better to make a new PR after the API change is merged :)

PR #906, opened 2021-04-01
Redirect to profile page after update hosting preference

### LINKED ISSUE #862: Saving changes on Edit Home and Edit Profile is inconsistent
Subject: Saving changes on Edit Home and Edit Profile is inconsistent
Description:
Saving changes on the "Edit profile" and "Edit home" leads to two different behaviors. I think both of them should redirect me to the profile page, or there should be a "Back to profile" button on both.

Steps:
Went to my profile page, clicked the "Edit profile" button (later: "Edit home") and applied some changes.

Results:
Saving my changes on "Edit Profile" redirects to the profile page. Saving my changes on "Edit home" keeps me on the "Edit Home" page.  They should both do one or the other to be consistent.

Frontend version: develop-7addcde4
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36
Page: https://develop--frontend.preview.coucher.org/profile
User ID: 25

anand-prem:
**Saving changes on Edit Home and Edit Profile is inconsistent**

Made both Edit Home and Edit Profile flow consistent by redirected to Profile page.
 
Issue: [#862](https://github.com/Couchers-org/couchers/issues/862l)

@aapeliv, I raised the MR but I can't select reviewers or assignees. Please do the needful.

### DIFF
--- a/app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
+++ b/app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
@@ -1,0 +1,42 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { Route, Switch } from "react-router-dom";
+import { profileRoute } from "routes";
+import { service } from "service";
+import wrapper from "test/hookWrapper";
+import { getUser } from "test/serviceMockDefaults";
+
+import { addDefaultUser, MockedService } from "../../../test/utils";
+import { SAVE } from "../../constants";
+import EditHostingPreference from "./EditHostingPreference";
+
+const mockProfilePage = "Mock Profile Page";
+const getUserMock = service.user.getUser as MockedService<
+  typeof service.user.getUser
+>;
+const MockProfileComponent = () => <div title={mockProfilePage} />;
+
+describe("EditHostingPreference", () => {
+  beforeEach(() => {
+    getUserMock.mockImplementation(getUser);
+    addDefaultUser();
+  });
+
+  it("should redirect to Profile route after successful update", async () => {
+    render(
+      <Switch>
+        <Route path={profileRoute}>
+          <MockProfileComponent />
+        </Route>
+        <Route>
+          <EditHostingPreference />
+        </Route>
+      </Switch>,
+      { wrapper }
+    );
+
+    userEvent.click(await screen.findByRole("button", { name: SAVE }));
+
+    expect(await screen.findByTitle(mockProfilePage)).toBeInTheDocument();
+  });
+});
--- a/app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.ts
+++ b/app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.ts
@@ -1,6 +1,8 @@
 import { useAuthContext } from "features/auth/AuthProvider";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { useMutation, useQueryClient } from "react-query";
+import { useHistory } from "react-router-dom";
+import { profileRoute } from "routes";
 import { HostingPreferenceData, service } from "service/index";
 import { SetMutationError } from "utils/types";
 
@@ -11,6 +13,7 @@ interface UpdateHostingPreferencesVariables {
 
 export default function useUpdateHostingPreferences() {
   const queryClient = useQueryClient();
+  const history = useHistory();
   const userId = useAuthContext().authState.userId;
   const { mutate: updateHostingPreferences, status, reset } = useMutation<
     Empty,
@@ -28,6 +31,7 @@ export default function useUpdateHostingPreferences() {
       },
       onSuccess: () => {
         queryClient.invalidateQueries(["user", userId]);
+        history.push(profileRoute);
       },
     }
   );


darrenvong (commented):
Change looks good - thanks for your first contribution!

Can you add a test for this please? `app/frontend/src/features/profile/actions/MessageUserButton.test.tsx` is a good example to look at to adapt to this if you are unsure on how to start. No pressure on having something as complete as that but a single test for this scenario you changed would be a good starting point for us to build/improve on in future 🙂

anand-prem:
@darrenvong sure thing. I will add.

The reason why I didn't added tests is because the test suite was failing always in my system. Any idea why that is failing? Is this any typescript or jest version mismatch? I'm on develop branch.

Screenshot 2021-04-03 at 8 57 04 AM


Also I couldn't find the file for tests for the component `/app/frontend/src/features/profile/edit/EditHostingPreference.tsx`

Can you help me to point out that as well?

darrenvong:
> The reason why I didn't added tests is because the test suite was failing always in my system. Any idea why that is failing?

Hmm you're not the first person who have seen this... something odd is happening with the jest setup with Create React App (the CLI tool we used to set up this project to start with) such that mocks don't seem to be working properly when they are files at the first level of the `src` folder. I introduced a workaround a few weeks ago but that seem to only please CI and not our machines ☹️ 

Try these things and hopefully one of these will work for you:
- Try going in `/app/frontend/src/__mocks__/service.ts` and trigger a save (Cmd/Ctrl + S depending on if you're on MacOS/Windows) on that file without changing anything. That should re-run the tests again and strangely that has fixed it to me, so try that first.
- Instead of running the whole test suite, try running the `EditHostingPreference` test suite only using the filter that you can set through the interactive test mode (press `p` then type `EditHostingPreference`). In the test, make sure to write `import { service } from "service/index";`. Importing from `service/index` seems to work fine for all of us before my workaround...

> Also I couldn't find the file for tests for the component /app/frontend/src/features/profile/edit/EditHostingPreference.tsx

That's because there aren't any yet for that component 😅  so you'll have to create a new file next to it with a similar structure to `MessageUserButton`.

Hope all of this helps!

anand-prem:
> Try going in /app/frontend/src/__mocks__/service.ts and trigger a save (Cmd/Ctrl + S depending on if you're on MacOS/Windows) on that file without changing anything.

Haha! Actually I also used that way only for running the tests ;)

Ok then I will try fix for that also and will create new test file for the component.

Thanks @darrenvong

anand-prem:
@darrenvong , in the `app/frontend/src/features/profile/edit/EditHostingPreference.tsx` component, it expects a user data. Any tests that I can refer to properly mock a user data?

anand-prem:
@darrenvong , I updated the PR with test cases also. Please review as I used to work on jest and enzyme before, I'm still a beginner with react-testing-libarary. So please share your thoughts for improvements. Thanks :)

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+} from "@testing-library/react";
+import { Route, Switch } from "react-router-dom";
+import { addDefaultUser, MockedService } from "../../../test/utils";
+import wrapper from "test/hookWrapper";
+import EditHostingPreference from "./EditHostingPreference";
+import { getUser } from "test/serviceMockDefaults";
+import { service } from "service";
+import { profileRoute } from "routes";
+
+const MockProfilePage = "Mock Profile Page";
```
darrenvong:
```suggestion
const mockProfilePage = "Mock Profile Page";
```
We tend to start variable names in upper case only if it's for actual React components

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+const MockProfileComponent = () => <div>{MockProfilePage}</div>;
+
+describe("EditHostingPreference", () => {
+  beforeEach(() => {
+    getUserMock.mockImplementation(getUser);
+    addDefaultUser();
+  });
+
+  it("should redirect to Profile route after successful update", async () => {
+    const { getByText } = render(
```
darrenvong:
```suggestion
    render(
```
Generally we don't need to do this anymore since we can use `getByText` from `screen`

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </Route>
+        <Route>
+          <EditHostingPreference />
+        </Route>
+      </Switch>,
+      { wrapper }
+    );
+    const saveButton = await waitFor(() => getByText(/save/i));
+    const rightClick = { button: 2 };
+    fireEvent.click(saveButton, rightClick);
```
darrenvong:
```suggestion
    userEvent.click(await screen.findByRole('button', { name: SAVE }));
```
A couple of things here:
- In React Testing Library (RTL), `*ByRole` should be preferred over `*ByText` since it also gives us a basic idea on how accessible our components are - [more info here](https://testing-library.com/docs/queries/about/#priority)
- `findBy*` is a shorthand for the `waitFor(() => ...)` and `getBy*` combo you wrote there
-  right click is unnecessary? The interaction here is triggered by a left click and RTL does that by default
- Prefer using `userEvent` from `@testing-library/user-event` [for reasons explained here](https://testing-library.com/docs/queries/about/#priority)
- Since we are not using the button for other purpose than clicking, I think it's clear enough to combine the two steps together

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const profilePage = await waitFor(() => screen.getByText(MockProfilePage));
+    expect(profilePage).toBeDefined();
```
darrenvong:
```suggestion
    expect(await screen.findByText(mockProfilePage)).toBeInTheDocument();
```
Prefer using assertions from [`@testing-library/jest-dom`](https://github.com/testing-library/jest-dom) over the primitive jest ones since [they give better errors](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-the-wrong-assertion).

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  fireEvent,
+  act,
+} from "@testing-library/react";
+import { Route, Switch } from "react-router-dom";
+import { addDefaultUser, MockedService } from "../../../test/utils";
+import wrapper from "test/hookWrapper";
+import EditHostingPreference from "./EditHostingPreference";
+import { getUser } from "test/serviceMockDefaults";
+import { service } from "service";
+import { profileRoute } from "routes";
```
darrenvong: Can you run `yarn lint --fix` so the imports are sorted?

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
+import {
+  render,
+  screen,
+  waitFor,
+  queryByAttribute,
```
darrenvong:
```suggestion
```
Unused imports

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+  fireEvent,
+  act,
```
darrenvong:
```suggestion
```
Unused imports

darrenvong (changes requested):
Even though there are quite a lot of comments, you made a great start especially not having used React Testing Library much before.

I've linked [this blog post by Kent C Dodds](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) in two of the review comments but really, I think you'll learn a lot about how to best use the library if you read the whole blog post, since the author himself wrote RTL!

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
darrenvong: Also `SAVE` is available as a constant in `features/constants`

anand-prem: Thanks for the comments with appropriate links as well. You are awesome. 👍🏻

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
anand-prem: I ran `yarn format`. I thought that will do the lint issues and all. Anyway I did this.

anand-prem:
@darrenvong made the changes as per the review comments.

darrenvong (approved):
Looks good! Feel free to look through the issues list and let us know if you fancy on working another one

Issue #774, opened 2021-03-10
Change login/signup picture


lucaslcode:
It is time to source these @ItsiW

andreaowu:
Have we decided on what we're doing for the photo? Happy to do the update it in code if we did!

ItsiW:
Miuriel is gathering all the community photos in one place. We'll start getting photos via the forum too, and then we can decide based on what's available. The stock photo is fine for now

andreaowu:
Which stock photo are you referring to?

ItsiW:
Ah looks like the photo was removed

andreaowu:
Login page has the right photo now, fixed by #794

lucaslcode:
@ItsiW are you sure? Check on mobile - I thought we didn't want an undiverse photo (desktop is ok)

ItsiW:
Good catch, I'll get Miuriel to pick one. Also is it easy to change it so the photo colours don't invert in dark mode? Or is that a my phone problem?

lucaslcode:
Not too sure. Maybe it will be fixed when we have an actual defined dark mode?

lucaslcode:
Moved to beta, see #911 for soft beta solution

lucaslcode:
@ItsiW you can close this or leave open, up to you

PR #911, opened 2021-04-01
Update mobile auth bg to be same as on desktop for soft beta

aapeliv:
These might need to be updated later with even better pictures, but this lets us release the soft beta.

Does a bit of #774 :P


aapeliv:
It's got `background-size: contain` now, which makes it too small, should be `cover`, but the design is made for a light picture, so it's hard to see the text on the dark background...

lucaslcode:
I am going to push to this branch

lucaslcode (approved):
@aapeliv double check and merge

aapeliv:
Works, merging

PR #920, opened 2021-04-02
Fix off by one day bug in reference start writing time

aapeliv:
The host request ends at midnight in the local timezone on the `to_date`. That's `to_date to timestamp + 1 day` (because `to_date to timestamp` is the midnight that starts that day).

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -638,14 +638,12 @@ class HostRequest(Base):
     # timezone aware start and end times of the request, can be compared to now()
     start_time = column_property(date_in_timezone(from_date, timezone))
     end_time = column_property(date_in_timezone(to_date, timezone) + text("interval '1 days'"))
+    end_time_to_write_reference = column_property(end_time + text("interval '14 days'"))
 
     status = Column(Enum(HostRequestStatus), nullable=False)
 
     to_last_seen_message_id = Column(BigInteger, nullable=False, default=0)
     from_last_seen_message_id = Column(BigInteger, nullable=False, default=0)
-
-    start_time_to_write_reference = column_property(date_in_timezone(to_date, timezone))
-    end_time_to_write_reference = column_property(date_in_timezone(to_date, timezone) + text("interval '14 days'"))
 
     from_user = relationship("User", backref="host_requests_sent", foreign_keys="HostRequest.from_user_id")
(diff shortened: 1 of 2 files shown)


kalvdans:
If you have to wait until midnight to write a reference after surfing it sounds good!

kalvdans (dismissed):

app/backend/src/couchers/models.py
```diff
@@ -638,15 +638,14 @@ class HostRequest(Base):
     # timezone aware start and end times of the request, can be compared to now()
     start_time = column_property(date_in_timezone(from_date, timezone))
```
ShmuelTreiger, line 639: Do we need/use start_time?

ShmuelTreiger (changes requested):
Looks like some tests broke

app/backend/src/couchers/models.py
aapeliv: It's not used right now, no

kalvdans (approved):
LGTM!

app/backend/src/couchers/models.py
ShmuelTreiger: Is it worth removing? My impression is column_properties are somewhat resource intensive?

app/backend/src/tests/test_references.py
```diff
@@ -134,26 +134,26 @@ def test_ListPagination(db):
...
-        ref6, _ = create_host_reference(session, user6.id, user1.id, timedelta(days=15, seconds=40), surfing=True)
+        ref6, _ = create_host_reference(session, user6.id, user1.id, timedelta(days=16, seconds=40), surfing=True)
 
-        ref7 = create_friend_reference(session, user7.id, user1.id, timedelta(days=15, seconds=30))
+        ref7 = create_friend_reference(session, user7.id, user1.id, timedelta(days=16, seconds=30))
 
-        ref8, _ = create_host_reference(session, user8.id, user1.id, timedelta(days=15, seconds=20), surfing=False)
-        ref9, _ = create_host_reference(session, user9.id, user1.id, timedelta(days=15, seconds=10), surfing=False)
+        ref8, _ = create_host_reference(session, user8.id, user1.id, timedelta(days=16, seconds=20), surfing=False)
+        ref9, _ = create_host_reference(session, user9.id, user1.id, timedelta(days=16, seconds=10), surfing=False)
```
ShmuelTreiger, line 156: Should we abstract out all the 16's for the next time we change this?

PR #921, opened 2021-04-02
[MRG} Lower unseen email time to 5 minutes (8 max)

### LINKED ISSUE #892: Lower unseen message email time
Currently unseen messages trigger a notification email if unseen for 30 minutes. This should be lowered to 5 minutes due to user complaints, and the background job also needs to run more often.

ShmuelTreiger:


aapeliv (approved):
It's 5-8 minutes now, but that's good

PR #922, opened 2021-04-02
[MRG] Alternative way to deal with user.regions

ShmuelTreiger:
@aapeliv What about something like this instead?


aapeliv:
Everything's a tradeoff, but this is why I'd argue that this change is worse than what you've now.

I know for simple lists it's tempting to write these helper functions. However, it has three problems:

1. You've created an abstraction of the regions table stuff into a list, but what you have now is actually not equivalent: I can't assign to `user.regions_visited = ["FIN", "SWE"]` because it's a read-only property. That means that we still have to dig into the underlying data structure and know two things (the list stuff, but then also how to assign to it).
1. It adds indirection. There's a lot of these list-type things through one-to-many joined lists, but a lot of them have more properties than a single list, e.g. language abilities. If you only had string lists, doing this (and adding some assignment hack for 1.) would probably make sense. But devs already have to know how the relationship stuff works, so it's not worth adding another thing to convolute it.
1. And finally, it's just more code in total to factor out a fairly easy list comprehension in two call sites

PR #924, opened 2021-04-02
Add link to connections page

lucaslcode:
The was no link since we took it out of the nav bar. This is more of a temporary fix (there's nothing as permanent as a temporary fix...)


aapeliv (approved):
LGTM, if you're happy with my approval :)

PR #930, opened 2021-04-03
fix: update chat in close to real time

### LINKED ISSUE #850: Group chats are not updated in "real time"
**Expected behaviour**
Messages in group chats should be updated in (near) real time.

**Actual behaviour**
Messages in group chats are currently not refreshed in (near) real time unless:
- The user quits the chat and go back in, or
- The user sends a message

**Other notes**
Should be fixable for now by changing the `refetchInterval` option in the `[groupChatMessages, groupChatId]` query to refetch every 2-3 second (hence "real time" in quotes). Probably should look into how grpc streams might help with this in the longer term as it'll likely be more efficient than spamming the server with this regular polling.

darrenvong:
Closes #850

Still a bit concerned that this might overload the backend when there are enough users even though this is set to refetch every 3 seconds (so it's not quite real time)... hopefully streaming or being able to use websocket so the connection is established once might help with this

### DIFF
--- a/app/frontend/src/features/messages/groupchats/GroupChatView.test.tsx
+++ b/app/frontend/src/features/messages/groupchats/GroupChatView.test.tsx
@@ -1,8 +1,7 @@
-import { render, screen, within } from "@testing-library/react";
+import { act, render, screen, within } from "@testing-library/react";
 import { MARK_LAST_SEEN_TIMEOUT } from "features/messages/constants";
 import GroupChatView from "features/messages/groupchats/GroupChatView";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
-import React from "react";
 import {
   mockAllIsIntersecting,
   mockIsIntersecting,
@@ -14,6 +13,8 @@ import messageData from "test/fixtures/messages.json";
 import { getHookWrapperWithClient } from "test/hookWrapper";
 import { getGroupChatMessages, getUser } from "test/serviceMockDefaults";
 import { addDefaultUser, MockedService, wait } from "test/utils";
+
+import { GROUP_CHAT_REFETCH_INTERVAL } from "./constants";
 
 const getGroupChatMock = service.conversations.getGroupChat as MockedService<
   typeof service.conversations.getGroupChat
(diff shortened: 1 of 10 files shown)


app/frontend/src/features/messages/groupchats/GroupChatView.test.tsx
```diff
@@ -125,6 +126,79 @@ describe("GroupChatView", () => {
...
+    expect(getGroupChatMock).toHaveBeenCalledTimes(1);
+    // 1 is groupChatId here
+    expect(getGroupChatMock).toHaveBeenCalledWith(1);
+    expect(getGroupChatMessagesMock).toHaveBeenCalledTimes(1);
+    expect(getGroupChatMessagesMock).toHaveBeenCalledWith(1, undefined);
+  });
+
+  describe("when there are new messages due to come in", () => {
+    beforeEach(() => {
+      jest.useFakeTimers();
```
darrenvong, line 138: Lol figuring this out was harder than making the change itself. But glad we have done it so we have an example and also so we don't have to wait for 3 seconds for this test to finish!

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -138,8 +145,8 @@ export default function GroupChatView() {
     (text) => service.conversations.sendMessage(groupChatId, text),
     {
       onSuccess: () => {
-        queryClient.invalidateQueries(["groupChatMessages", groupChatId]);
-        queryClient.invalidateQueries(["groupChats"]);
+        queryClient.invalidateQueries(groupChatMessagesKey(groupChatId));
+        queryClient.invalidateQueries([groupChatBaseKey]);
```
lucaslcode: I don't think this should be named like this. Maybe `groupChatsList`?

lucaslcode (changes requested):

app/frontend/src/features/messages/groupchats/GroupChatView.tsx
darrenvong: Yeah that's a better name, missed another place where this is used so I'll update there too

darrenvong:
Not sure why the build failed in the other steps, but it's not related to the frontend... probably related to the Docker Hub issue Aapeli mentioned earlier

lucaslcode (approved):

PR #932, opened 2021-04-03
Frontend/feature/new favicon

ItsiW:
I've updated the favicon, logo192 and logo512. The logo now goes edge-to-edge of the image, let me know if that's an issue.


aapeliv:
The negative space should probably be even bigger in the favicon, it's still a bit murky

darrenvong (commented):
`logo192` and `logo512` seem to have the wrong shade of the orange that's fixed in the `favicon`. Let's fix while we're at them?

darrenvong (approved):
Looks good to me!

PR #955, opened 2021-04-03
Friend request API change

### LINKED ISSUE #954: FriendshipStatus in user objects only returns pending status from GetUser API
Currently when there is a pending request, we can't tell easily if a pending request is sent by current user or received from the user they are looking at from the frontend without extra data wrangling by getting data from `ListFriendRequests` and figure out from that. This should be encoded in `FriendshipStatus` instead.

aapeliv:
Closes #954


aapeliv:
Would this API change solve #954 @darrenvong?

darrenvong:
Yeah this should do it, otherwise I was thinking of encoding that in `FriendshipStatus` by splitting `PENDING` to `RECEIVED` or `SENT` to avoid an extra layer of nesting

aapeliv:
> Yeah this should do it, otherwise I was thinking of encoding that in `FriendshipStatus` by splitting `PENDING` to `RECEIVED` or `SENT` to avoid an extra layer of nesting

You'd still need the friend request id though?

darrenvong:
Yeah good point, ignore what I said above then, probably easiest to return the whole object

app/pb/api.proto
```diff
@@ -57,6 +57,9 @@ message FriendRequest {
   FriendRequestStatus state = 2;
 
   int64 user_id = 3;
+
+  // we sent the request if true, received it if false
+  bool sent = 4;
```
kalvdans, line 62: This is redundant in `ListFriendRequestsRes`. Can't you add it only to `User`?

kalvdans (dismissed):
Just a small comment, otherwise ok!

javiwq:
I don't think the changes help simplify #905. We would still have to do extra data wrangling on the frontend, because `ListFriendRequests` returns two separate lists for received and sent requests. Unless I'm mistaken, I do think  encoding that in `FriendshipStatus` by splitting `PENDING` to `RECEIVED` or `SENT` is what would simplify this; indeed you still need the `friendRequestId`, but then you can just query the corresponding sent list or received list.

lucaslcode:
@ineswq in this pr each request has `sent: bool` and the whole pending request itself is now present in `user`. So you can know if it's sent or received with only the user object.

app/backend/src/couchers/servicers/api.py
```diff
@@ -429,23 +429,44 @@ def ListFriends(self, request, context):
...
-        with session_scope() as session:
-            from_user = session.query(User).filter(User.id == context.user_id).one_or_none()
-
-            if not from_user:
-                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
+        if context.user_id == request.user_id:
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.CANT_FRIEND_SELF)
 
+        with session_scope() as session:
+            user = session.query(User).filter(User.id == context.user_id).one()
```
ShmuelTreiger, line 432: Why did you rename this from "from_user"?

Also, do we want to control the error if the (from_)user isn't found?

app/backend/src/couchers/servicers/api.py
```diff
@@ -429,23 +429,44 @@ def ListFriends(self, request, context):
...
+                        and_(
+                            FriendRelationship.from_user_id == request.user_id,
+                            FriendRelationship.to_user_id == context.user_id,
+                        ),
+                    )
+                )
+                .filter(
+                    or_(
+                        FriendRelationship.status == FriendStatus.accepted,
+                        FriendRelationship.status == FriendStatus.pending,
```
ShmuelTreiger, line 455: What happens if it's rejected? Do we want a way for the recipient to stop incoming friend requests without blocking the user?

For example:
1) user_1 sends friend request to user_2. User_2 rejects. User_1 sends again. Repeat ad infinitum
2) Alternatively, user_1 sends request to user_2. User 2 ignores/leaves it pending. User 1 cancels and re-requests. Repeat ad infinitum

ShmuelTreiger (changes requested):
Wouldn't it be simpler to just return the friend request object? (Which I suggested and coded a month ago for this exact use case.) This seems like a ton of excess code to me.

app/backend/src/couchers/servicers/api.py
aapeliv: > Why did you rename this from "from_user"?

It's the user performing the API call, we call it user everywhere else too, just a good convention



> Also, do we want to control the error if the (from_)user isn't found?

it's guaranteed to exist

app/backend/src/couchers/servicers/api.py
aapeliv: > What happens if it's rejected? Do we want a way for the recipient to stop incoming friend requests without blocking the user?

not for now, you can re-send requests if they're rejected



> For example:
> 
>     1. user_1 sends friend request to user_2. User_2 rejects. User_1 sends again. Repeat ad infinitum
> 
>     2. Alternatively, user_1 sends request to user_2. User 2 ignores/leaves it pending. User 1 cancels and re-requests. Repeat ad infinitum

user 2 reports user 1

aapeliv:
> Wouldn't it be simpler to just return the friend request object?

it's good to have it in the user object instead of extra api query.

> Which I suggested and coded a month ago for this exact use case.

Could've been done that way too. That PR wasn't created for this use case though, at least it wasn't mentioned then, it seemed redundant and I didn't realise it'd be needed a tthe time. Either way, this would've been a perfectly valid use for what you were implementing there

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: from_user seems clearer to me, but I suppose it's mostly irrelevant

Even if someone random tries to hit our backend?

app/backend/src/couchers/servicers/api.py
```diff
@@ -429,23 +429,44 @@ def ListFriends(self, request, context):
...
+                    )
+                )
+                .one_or_none()
+                is not None
+            ):
                 context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.FRIENDS_ALREADY_OR_PENDING)
 
-            # Race condition!
+            # TODO: Race condition where we can create two friend reqs, needs db constraint! See comment in table
```
ShmuelTreiger, line 463: Aside from the race condition, what happens in the case where the request exists but was cancelled/rejected?

aapeliv: It shouldn't change the backend behaviour

ShmuelTreiger: Aren't we making a duplicate friend relationship?

aapeliv: Yep

aapeliv:
@ShmuelTreiger: why did you make that change? I don't see a problem with having cancelled/rejected friend requests in the database. In fact it's useful if we need info about them later.

Either way, if you want to not store cancelled/rejected fr reqs, they should just not be added to the database. Deleting them when you find them is not the way to go. Your change also has a hidden bug: if I send a request then cancel it and send you another one, this code will fail (when it shouldn't) because that DB query returns more than one result.

Also in general if you'd like to recommend change in another PR, please create a PR on top of that one PR, don't push directly to the branch. (git is good for reconciling changes between people, but it's good to keep one branch per person or you get back into issues with version control)

ShmuelTreiger:
> @ShmuelTreiger: why did you make that change?

Trying to help. Sorry.

>Either way, if you want to not store cancelled/rejected fr reqs, they should just not be added to the database.

Figured we might want to see if there was an attempt

>Your change also has a hidden bug: if I send a request then cancel it and send you another one, this code will fail (when it shouldn't) because that DB query returns more than one result.

The second one would delete the original

            if existing_relationship:
                if (
                    existing_relationship.status == FriendStatus.accepted
                    or existing_relationship.status == FriendStatus.pending
                ):
                    context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.FRIENDS_ALREADY_OR_PENDING)
                session.delete(existing_relationship) #FriendStatus.rejected or FriendStatus.Cancelled

 
> Also in general if you'd like to recommend change in another PR, please create a PR on top of that one PR, don't push directly to the branch. (git is good for reconciling changes between people, but it's good to keep one branch per person or you get back into issues with version control)

That makes sense. Sorry

aapeliv:
> Trying to help. Sorry.

Sorry, I didn't mean to come off as aggressive or question your help, just about the reasoning for the change. Appreciate your help as always! No need to apologize.

> > Your change also has a hidden bug: if I send a request then cancel it and send you another one, this code will fail (when it shouldn't) because that DB query returns more than one result.
> 
> That would delete the original

Ahh, I see. I got it mixed up on a quick reading. My bad. Sorry :/

There are two ways to do it:

1. keep a copy of all historic friend reqs in the database (like currently). But that requries a partial unique index to make sure there is at most one accepted or pending request (others can have multiple).
2. keep at most one copy of friend req in the database. This is what you're proposing and is simpler in not requiring a partial index (just a normal index). No history.

Actually, I think you're right. It is cleaner to have the latter. And it's simpler.

ShmuelTreiger:
> Ahh, I see. I got it mixed up on a quick reading. My bad. Sorry :/

No worries. I should have included that comment. If you keep the change, you can add it.

>Actually, I think you're right. It is cleaner to have the latter. And it's simpler.

I was basing it off what you told me with userblocking. Just to keep blocks and delete them otherwise.

Don't feel like you have to keep it. I won't be offended if you roll it back, or just delete them when they're rejected/cancelled.

I also wasn't sure if text_fixtures was the right place for get_FriendRequest, I was just tired of typing out the query anew every time.

lucaslcode:
Is this ready for frontend work?

aapeliv:
> Is this ready for frontend work?

Yes

aapeliv:
I undid that change. I'm open to changing the backend, but I think it requires a bit more work and can be done in a different PR. I don't know if it's worth doing now, this works too.

I also renamed `get_FriendRelationship` to `get_friend_relationship` to agree with other conventions (only tests themselves break the snake case rule by sometimes being called `test_RpcName`, etc)

app/backend/src/tests/test_api.py
```diff
@@ -585,7 +585,7 @@ def test_reject_friend_request(db):
         assert res.sent[0].user_id == user2.id
 
         # Check only one existing fr
-        fr = get_FriendRelationship(user1, user2)
+        fr = get_friend_relationship(user1, user2)
         assert fr is not None  # If multiple, would have already excepted
```
ShmuelTreiger: This was added as a check on the change I made and could break otherwise

app/backend/src/tests/test_api.py
```diff
@@ -533,7 +533,7 @@ def test_CancelFriendRequest(db):
         assert res.sent[0].user_id == user2.id
 
         # Check only one existing fr
-        fr = get_FriendRelationship(user1, user2)
+        fr = get_friend_relationship(user1, user2)
```
ShmuelTreiger: Ditto

ShmuelTreiger (changes requested):

aapeliv:
Thanks, fixed

lucaslcode (dismissed):
Anyone feel free to merge this once checks are green

lucaslcode (approved):

PR #956, opened 2021-04-03
Add username to bug reports

### LINKED ISSUE #953: Update bug reports to list username as well
Would be good to have `username (user_id)` (with id in brackets) in case username changing is implemented, etc. Currently we just have the user id and you have to memorize who's who and ask around, but that's no fun.

aapeliv:
Closes #953

### DIFF
--- a/app/backend/src/couchers/servicers/bugs.py
+++ b/app/backend/src/couchers/servicers/bugs.py
@@ -2,6 +2,8 @@ import grpc
 import requests
 
 from couchers.config import config
+from couchers.db import session_scope
+from couchers.models import User
 from pb import bugs_pb2, bugs_pb2_grpc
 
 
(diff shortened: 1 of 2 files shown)


lucaslcode (approved):

app/backend/src/couchers/servicers/bugs.py
```diff
@@ -19,6 +21,9 @@ def ReportBug(self, request, context):
         repo = config["BUG_TOOL_GITHUB_REPO"]
         auth = (config["BUG_TOOL_GITHUB_USERNAME"], config["BUG_TOOL_GITHUB_TOKEN"])
 
+        with session_scope() as session:
+            username = session.query(User.username).filter(User.id == request.user_id).one_or_none() or "<unknown>"
```
kalvdans, line 25: Error on this line fixed by https://github.com/Couchers-org/couchers/pull/963

PR #960, opened 2021-04-03
Change user route to /user, not /profile

### LINKED ISSUE #933: User page should be /user/{username} not /profile/{username}
As per title, /profile is still good for the current user's own profile

aapeliv:
Closes #933

### DIFF
--- a/app/frontend/src/routes.ts
+++ b/app/frontend/src/routes.ts
@@ -6,15 +6,15 @@ export const settingsRoute = "/account-settings";
 export const confirmChangeEmailRoute = "/confirm-email";
 
 export const signupRoute = "/signup";
-export const profileRoute = "/profile";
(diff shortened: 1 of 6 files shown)


app/frontend/src/routes.ts
```diff
@@ -32,8 +32,7 @@ export const logoutRoute = "/logout";
 export const connectionsRoute = "/connections";
 export const friendsRoute = `${connectionsRoute}/friends`;
 
-export const userRoute = "/user";
-export const routeToUser = (username: string) => `${profileRoute}/${username}`;
+export const routeToUser = (username: string) => `${userRoute}/${username}`;
```
darrenvong: Just a small nitpick - might make more sense to move this line up with all of the other user routes so they are grouped together in this file.

darrenvong (approved):

PR #961, opened 2021-04-03
refactor: use colour from theme for stylesheet

darrenvong:
Since there aren't that many around and I didn't want to include it randomly in my other feature PRs, quick fix to try to stick to the colour palette on the theme where possible rather than hardcoding colour values.


app/frontend/src/features/auth/AuthPage.tsx
```diff
@@ -18,19 +18,18 @@ import {
...
 import { loginRoute, signupRoute } from "routes";
 
 import { COUCHERS } from "../../constants";
 
 const useStyles = makeStyles((theme) => ({
   aboutUs: {
-    color: "#2a2a2a",
-    marginTop: "auto",
+    color: theme.palette.common.white,
+    marginTop: theme.spacing(2),
```
darrenvong, line 29: Ok plus a cheeky fix - the "about us" link on mobile screens were too close to the two buttons before this fix.

lucaslcode (approved):

PR #971, opened 2021-04-04
Remove SSO for discourse (the forum)

### LINKED ISSUE #907: Remove SSO
SSO for discourse (our forum) was implemented, but I think we decided to not use it in the end and is not functoinal. This should be removed (it can be recovered from history if we for whatever reason change our minds).

ceelo777:
Fixes #907 

- Remove SSO from servicers
- Remove SSO proto file
- Remove SSO calls from client-side
- Remove SSO error constant


lucaslcode (dismissed):
Use the main repo next time for CI :)

aapeliv:
It seems you can also remove `base64encode` and `base64decode` in `crypto.py`

aapeliv:
@ceelo777: there's unused imports for those bas64 functions now, other than that, this is ready for merge :)

aapeliv (approved):
LGTM, thanks!

PR #973, opened 2021-04-04
Search box

aapeliv:
Design:

Screen Shot 2021-04-04 at 12 13 26 pm

Before:

Screen Shot 2021-04-04 at 12 26 32 pm

After:

Screen Shot 2021-04-04 at 3 09 05 pm

### DIFF
--- a/app/frontend/src/features/search/SearchBox.tsx
+++ b/app/frontend/src/features/search/SearchBox.tsx
@@ -17,11 +17,9 @@ const useStyles = makeStyles((theme) => ({
     marginLeft: "5%",
   },
   box: {
-    "& > .MuiOutlinedInput-root": {
-      width: "70%",
-      [theme.breakpoints.down("md")]: {
-        width: "90%",
-      },
+    width: "70%",
+    [theme.breakpoints.down("md")]: {
+      width: "90%",
     },
   },
 }));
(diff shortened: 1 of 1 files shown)


app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -35,13 +38,22 @@ export default function SearchBox() {
   return (
     <>
       <form onSubmit={onSubmit} className={classes.root}>
-        <TextField
-          className={classes.box}
-          id="search-query"
-          name="query"
-          label="Search"
-          inputRef={register}
-        ></TextField>
```
lucaslcode, line 44: just use this with variant="standard"

app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -1,5 +1,10 @@
 import { makeStyles } from "@material-ui/core";
-import TextField from "components/TextField";
+import FormControl from "@material-ui/core/FormControl";
+import IconButton from "@material-ui/core/IconButton";
+import Input from "@material-ui/core/Input";
+import InputAdornment from "@material-ui/core/InputAdornment";
+import InputLabel from "@material-ui/core/InputLabel";
+import SearchIcon from "@material-ui/icons/Search";
```
lucaslcode: prefer importing from `@material-ui/core`

lucaslcode (changes requested):

app/frontend/src/features/search/SearchBox.tsx
aapeliv: It doesn't allow adding the little magnifying glass, I don't think? That's why I had to do the form control stuff

app/frontend/src/features/search/SearchBox.tsx
aapeliv: Did I do it right?

lucaslcode (approved):

PR #974, opened 2021-04-04
Visual markdown editor

lucaslcode:
Using @toast-ui/editor.

There were compatibility problems between this editor and `markdown-it`, the editor uses `<br>` for lines with no text, but `markdown-it` doesn't accept any html tags. So I switched to the toast ui viewer (and that way we aren't duplicating library functionality).

However, that means images will be displayed now. We can either filter them out of the markdown with regex (change it to a regular link?), or replace [their html sanitizing function](https://github.com/nhn/tui.editor/blob/master/apps/editor/src/js/htmlSanitizer.js) with our own (don't fancy doing that too much).

The markdown component takes a `topHeaderLevel` prop, so you can change the biggest header `# like this` to be H2 instead of H1, to fit in document flow better.

I also took the opportunity to hide these fields on the profile if they don't have any content.

Screens:
image
image
image
image

(So on the profile, the h1 is actually h2, and the h2 is h3)


darrenvong:
Not sure if it's a config issue or bug with the editor... but you can't unstrike a previously saved striked out text without deleting it altogether

app/frontend/src/features/profile/ProfileMarkdownInput.tsx
```diff
@@ -1,57 +1,37 @@
...
-    </>
+    <div className={className}>
+      <Typography variant="h2" id={id}>
+        {label}
+      </Typography>
+      <MarkdownInput
+        control={control}
+        defaultValue={defaultValue}
+        id={id}
+        labelId={`${id}-label`}
```
darrenvong, line 32:
```suggestion
        labelId={id}
```
The other value was flagging up as invalid in the accessibility view

app/frontend/src/components/Markdown.test.tsx
```diff
@@ -0,0 +1,14 @@
+import { increaseMarkdownHeaderLevel } from "./Markdown";
+
+describe("markdown header level increase", () => {
+  it.each`
+    source                                           | topLevel | result
+    ${"#hi\n##hello\nYo\n###omg"}                    | ${3}     | ${"###hi\n####hello\nYo\n#####omg"}
+    ${"#hi\n##hello\nYo\n###omg"}                    | ${1}     | ${"#hi\n##hello\nYo\n###omg"}
+    ${"#hi\n##hello\nYo\n###omg"}                    | ${6}     | ${"######hi\n######hello\nYo\n######omg"}
+    ${"#h1\n##h2\n###h3\n####h4\n#####h5\n######h6"} | ${3}     | ${"###h1\n####h2\n#####h3\n######h4\n######h5\n######h6"}
```
darrenvong, line 10: Might be worth testing all `topLevel` from 1-6 if you are doing this? I can see 2 and 4 missing in this

app/frontend/src/components/Markdown.tsx
```diff
@@ -1,26 +1,68 @@
...
+import { useEffect, useRef } from "react";
 
-interface MarkdownProps extends BoxProps {
+interface MarkdownProps {
   source: string;
+  topHeaderLevel?: number;
 }
 
 const useStyles = makeStyles((theme) => ({
   root: {
```
darrenvong, line 13: This currently displays things with special formatting like bold, italic etc in a different font, so I think you need to nest these styles in `.tui-editor-contents` like what you did for the editor.

darrenvong (changes requested):
Might also want to update the `Markdown` storybook entry to be "composite" too?

Apparently they also have [a React wrapper](https://github.com/nhn/tui.editor/tree/master/apps/react-editor). Why aren't we using that?

app/frontend/src/features/profile/ProfileMarkdownInput.tsx
lucaslcode: It was something else, I wasn't setting the h2 id to this value.

lucaslcode:
> you can't unstrike a previously saved striked out text

Works for me

> Apparently they also have a React wrapper. Why aren't we using that?

You can't set `id` on the containing div.

app/frontend/src/components/Markdown.tsx
```diff
@@ -1,26 +1,77 @@
...
-  );
+
+  const rootEl = useRef<HTMLDivElement>(null);
+  const viewer = useRef<ToastUIViewer>();
+  useEffect(() => {
+    let sanitizedSource = increaseMarkdownHeaderLevel(source, topHeaderLevel);
+    //remove all html except <br>
+    sanitizedSource = sanitizedSource.replace(/<(?!br)([^>]+)>/gi, "");
+    //change images ![]() to links []()
+    sanitizedSource = sanitizedSource.replace(/!(?!\[.*\]\(.*\))/gi, "");
```
lucaslcode: Stripping html here except <br>. I didn't want to do it in the backend, because then you have to keep track of which fields are markdown. And you can't do it in the editor, because the client isn't trustworthy. So viewer it is.

darrenvong:
> Works for me

https://user-images.githubusercontent.com/13669362/113638114-020e6980-966e-11eb-8e8c-11261045337c.mov

You don't get this? I'm actually clicking on the button in the recording and nothing happens 🤔 
That seems to only happen to greyed out text. If the text is striked out but still in black, unstriking works for me, a bit odd...

If you are a power user, going to the markdown source and manually remove the `~~` works, but seems inconvenient for those who sticks with WSYWIG. Maybe we should hide that button if we can't fix it? Power users can still do strike through with markdown if they really want to...

darrenvong (changes requested):
Can we move the existing `Markdown.stories.tsx` to composite too? Seems inconsistent to have one on simple and the other on composite when they are about as "simple" as each other

lucaslcode:
What if you click on the crossed out bit without selecting and press the button?

It's probably browser related

darrenvong:
> What if you click on the crossed out bit without selecting and press the button?
> 
> It's probably browser related

Still does nothing :/ this is happening on Brave (basically Chrome), Safari and Firefox for me on MacOS. Not sure if OS would make a big difference or not...)

lucaslcode:
Removed strikethrough button

darrenvong (dismissed):

darrenvong (approved):

Issue #980, opened 2021-04-05
Critical changes to signup flow


ItsiW:
Some changes before we can start pushing it on marketing channels

- [x] Change to the date selector component so it's easier to change years
- [ ] Add ("press Enter to search") to map search descriptive text
- [ ] Make magnifying glass replicate hitting enter on map search input
- [ ] Remove adjustable radius, and set it at 1km (although leave this as adjustable on the edit profile page)
- [ ] New text field underneath map for "Display address" that is auto-populated with the location selection

justjooshing:
The color contrast is also quite hard to read on mobile (edit: created new issue here [#983](https://github.com/Couchers-org/couchers/issues/983)
![image](https://user-images.githubusercontent.com/60824744/113581260-d31cd700-961e-11eb-8352-ee326e461ca9.png)

justjooshing:
@ItsiW What's difficult re the ease of changing years atm?
Just the default date being 1900? I found it easy to type in a date on mobile, but first tried using the calendar and scrolling by month from 1900 is ridiculous haha

ItsiW:
Looks like we'll have to change that link colour, could you put up an issue for that?

Yeah it's just that using the calendar widget is too difficult and I think most people would use that rather than typing. Even if it was 5 years away from your birthdate that's a lot of scrolling

lucaslcode:
@justjooshing is working on year input. @aapeliv is working on map related issues.

ItsiW:
Looking at the new date selector component, could we make it so that it opens up to the year selection? This removes an extra click and I feel the year selector option is slightly less than obvious

justjooshing:
Alright, just pushed that change, for the birthday signup flow only (the others will still open on date). Let me know what you think once it's merged, and we can adjust if we want @ItsiW 
https://github.com/Couchers-org/couchers/pull/991

PR #984, opened 2021-04-05
Allow easier choice of year and replace default with placeholder value

justjooshing:
For the first part of [#980](https://github.com/Couchers-org/couchers/issues/980)

It opens to 1900, which I'd prefer it to open around 2000, but I can't find how to change that (I think it just opens at the earliest possible date)
image

### DIFF
--- a/app/frontend/src/components/Datepicker.tsx
+++ b/app/frontend/src/components/Datepicker.tsx
@@ -41,7 +41,7 @@ export default function Datepicker({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Controller
         control={control}
-        defaultValue={Date()}
+        defaultValue={new Date()}
         inputRef={inputRef}
         name={name}
         render={({ onChange, value }) => (
--- a/app/frontend/src/components/Datepicker.tsx
+++ b/app/frontend/src/components/Datepicker.tsx
@@ -41,7 +41,7 @@ export default function Datepicker({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Controller
         control={control}
-        defaultValue={null}
+        defaultValue={Date()}
         inputRef={inputRef}
         name={name}
         render={({ onChange, value }) => (
@@ -49,7 +49,6 @@ export default function Datepicker({
             animateYearScrolling={true}
             autoOk
             className={className}
-            disableToolbar
             error={error}
             format={getLocaleFormat()}
             fullWidth
@@ -66,8 +65,6 @@ export default function Datepicker({
             onChange={(date) => {
               if (date?.isValid()) onChange(date?.toDate());
             }}
-            openTo="year"
-            placeholder="dd/mm/yyyy"
             views={["year", "month", "date"]}
             value={value}
             variant="inline"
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/Datepicker.tsx
```diff
@@ -66,6 +66,9 @@ export default function Datepicker({
             onChange={(date) => {
               if (date?.isValid()) onChange(date?.toDate());
             }}
+            openTo="year"
+            placeholder="dd/mm/yyyy"
```
lucaslcode: This isn't true for all locales.

I think a better solution instead of a placeholder just have `Datepicker` accept an optional `defaultValue` prop, and if undefined use today as a default value.

app/frontend/src/components/Datepicker.tsx
```diff
@@ -66,6 +66,9 @@ export default function Datepicker({
             onChange={(date) => {
               if (date?.isValid()) onChange(date?.toDate());
             }}
+            openTo="year"
```
lucaslcode: It doesn't always make sense to open to the year. I think either make this a prop, or get rid of `disableToolbar` so you can select the year that way. Or maybe both.

lucaslcode (changes requested):
Tests failing because they expect today's date to be the default

app/frontend/src/components/Datepicker.tsx
justjooshing: I can't think of what a suitable `defaultValue` would be. I looked at something like 01/01/2000, but it feels clunkier compared to just using today's date. Whats the impact of omitting the default value all together?

app/frontend/src/components/Datepicker.tsx
justjooshing: Yeah, having a look this morning, I think that it's better just to have the toolbar, but without an `openTo` prop.
If we don't have it then if you misclick the year you can't easily fix it - and we don't have to weirdly scroll from 1900.

app/frontend/src/components/Datepicker.tsx
lucaslcode: Not sure. Today is probably fine!

app/frontend/src/components/Datepicker.tsx
```diff
@@ -41,15 +41,14 @@ export default function Datepicker({
     <MuiPickersUtilsProvider utils={DateFnsUtils}>
       <Controller
         control={control}
-        defaultValue={null}
+        defaultValue={Date()}
```
lucaslcode: Tests are failing, I think because this returns a number not a date. Try `new Date()`

justjooshing: I really need to get better at testing - I'm not sure if it makes a difference but previously default was `minDate`

justjooshing: seems to have passed

lucaslcode: You can run `yarn test` or `yarn test filename` to run the tests locally

lucaslcode (approved):

PR #986, opened 2021-04-06
Discussion part 1: redesign discussion cards

darrenvong:
First PR of many for #543. Mostly code reshuffles and updated the discussion card to include user avatar in this. Tests and all that to come later...

**Web**
Screenshot 2021-04-06 at 03 28 11
Screenshot 2021-04-06 at 03 28 27

**Mobile**
Screenshot 2021-04-06 at 03 29 18


app/frontend/src/features/communities/discussion/DiscussionsListPage.tsx
```diff
@@ -0,0 +1,135 @@
+import { Dialog, DialogContent, makeStyles } from "@material-ui/core";
```
darrenvong, line 1: This is intentionally copy/pasted from `DiscussionSection` for now. I'll likely refactor this in an upcoming part to share more markups between the two if that makes sense later on. But for now, duplicating seems like the cheapest way to distinguish the two so the bottom button can behave differently (one links to the main discussion list, whereas the other actually loads more discussions if more are available).

app/frontend/src/AppRoutes.tsx
```diff
@@ -162,28 +162,32 @@ export default function AppRoutes() {
       {
         // COMMUNITIES
       }
-      <AppRoute isPrivate path={communityRoute}>
-        <CommunityPage />
-      </AppRoute>
+      {process.env.REACT_APP_IS_COMMUNITIES_ENABLED && (
```
darrenvong, line 165: So I/we can keep on pushing to `develop` without needing completely working code rather than needing a feature branch

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -95,37 +82,9 @@ export default function DiscussionCard({
             <Typography variant="h2" component="h3">
               {discussion.title}
             </Typography>
-            <Typography variant="body1">{textTruncated}</Typography>
-            <div className={classes.replies}>
-              {(thread?.pages.length ?? 0) > 0 ? (
```
darrenvong, line 100: The current cards don't show any replies so I just got rid of them to match the designs.

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
```diff
@@ -113,46 +99,45 @@ export default function DiscussionsSection({
...
+                discussion={discussion}
+                key={`discussioncard-${discussion.threadId}`}
+              />
+            ))
+        ) : (
+          <TextBody>{DISCUSSIONS_EMPTY_STATE}</TextBody>
+        )}
         {discussionsHasNextPage && (
           <div className={classes.loadMoreButton}>
-            <Link
```
darrenvong, line 146: Apparently nesting `<button>`s inside `<a>`s (what `Link` from React router generates) is [invalid HTML](https://stackoverflow.com/questions/6393827/can-i-nest-a-button-element-inside-an-a-using-html5#:~:text=No%2C%20it%20isn't%20valid,be%20no%20interactive%20content%20descendant.), so that's why I went with button to keep the styling but use the history API to do the linking.

app/frontend/src/AppRoutes.tsx
lucaslcode: Do `=== "true"` or else this will be true if the env var is false (because it's a string)

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -45,41 +47,26 @@ export default function DiscussionCard({
...
+    const strippedText = stripMarkdown(discussion.content.replace("\n", " "));
+    return strippedText.length > 300
       ? strippedText.substring(0, 298) + "..."
       : strippedText;
+  }, [discussion.content]);
 
   return (
     <Card className={classNames(classes.root, className)}>
       <Link to={routeToDiscussion(discussion.discussionId, discussion.slug)}>
-        <CardActionArea>
```
lucaslcode, line 79: Why is this gone? I think it sets hover styling for the link.

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
lucaslcode: How about `<Link component={Button}>`

lucaslcode (changes requested):
How does it look if the Avatar is a bit smaller and part of the surtitle? (And then the title and content don't have the wide left margin)

app/frontend/src/AppRoutes.tsx
darrenvong: Hmm wouldn't it be easier to not set the environment variable at all rather than setting it to false explicitly (or more like `"false"` when parsed by JS) in that case? So we only set it if we want it to be truthy

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
darrenvong: Oops I think you're right. I thought it was redundant since it didn't affect the button function 🙈

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
darrenvong: That caused a new warning saying `navigate` from React Router gets passed to `<a>`, so the button workaround might be the best...

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
darrenvong: `CardActionArea` renders a "button", and there doesn't seem to be a way to override it, so this also leads to the invalid interactive element nesting under `<a>` issue. I implemented the hover in CSS instead

darrenvong:
> How does it look if the Avatar is a bit smaller and part of the surtitle?

Not sure what you mean by this. Do you mean have a separate component for this and make the circle about the same height as the surtitle text?

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
darrenvong: Also it is button that just happens to have the same functionality as a link so it's not exactly semantically wrong to use a button...

darrenvong:
Screenshot 2021-04-08 at 23 06 04
@lucaslcode Made the avatar a bit smaller and lined it up with the surtitle. Is this what you mean?

lucaslcode:
I meant more like this but with better margins and stuff: 
image

app/frontend/src/AppRoutes.tsx
lucaslcode: In theory, but it's easy to forget and have it set to false (ie. wasn't it set to false just before you changed it?)

app/frontend/src/components/Button/Button.tsx
```diff
@@ -12,7 +12,7 @@ import CircularProgress from "../CircularProgress";
 
 const useStyles = makeStyles((theme) => ({
   contained: {
-    borderRadius: `${theme.shape.borderRadius * 2}px`,
+    borderRadius: theme.spacing(0.5),
```
lucaslcode: Why this change? Should border radius be set in relation to the theme border radius?

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -1,37 +1,57 @@
...
 import { timeAgo } from "utils/timeAgo";
 
 const useStyles = makeStyles((theme) => ({
-  cardContent: { height: "100%" },
+  avatar: {
+    gridArea: "avatar",
+  },
+  cardContent: {
+    display: "flex",
+    "&&": {
```
lucaslcode, line 22: What does this do?

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
```diff
@@ -1,37 +1,57 @@
...
+  },
+  discussionSummaryContainer: {
+    columnGap: theme.spacing(2),
+    display: "grid",
+    gridTemplateAreas: `
+      "avatar surtitle"
+      ". summary"
+    `,
+    gridTemplateColumns: `${theme.spacing(4)} auto`,
+    gridTemplateRows: `${theme.spacing(4)} auto`,
```
lucaslcode: I think we shouldn't use spacing for not-spacing values. If we later change the spacing, this shouldn't change.
So maybe just use `2rem`?

lucaslcode: (I have probably been guilty of this before)

app/frontend/src/components/Button/Button.tsx
darrenvong: The border radius was too high compared to the design, and I didn't want to hack around the theme's border radius with some weird maths. Maybe we should lower the value so it's more flexible, although that might involve touching some unrelated file in this PR if I do that...

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
darrenvong: It's a specificity hack (applying the class twice) needed to override MUI's `.MuiCardContent-root:last-child` rule. The ugly side of working with a framework when you want uniform padding and the framework decides to randomly add in more specific selectors with more padding to screw you over

app/frontend/src/features/communities/CommunityPage/DiscussionCard.tsx
darrenvong: Hmm good point, I'll change it to write in literal `rem` instead if these styles are still relevant after I play around with your suggested layout

darrenvong:
> I meant more like this but with better margins and stuff:
> image

Tried this layout with more spacing, I think I still prefer what I had from the design, but reduced the spacing between the the avatar and the title instead.

darrenvong:
Screenshot 2021-04-10 at 23 07 39

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
lucaslcode: Still not sure about this. Doesn't it break some functionality, like middle mouse click opening in new tab?

How about this - replacing the outer div with `<Button component="div">` and keeping the link? That should be correct html and semantics and still be a proper link

lucaslcode (changes requested):

app/frontend/src/features/communities/CommunityPage/DiscussionsSection.tsx
darrenvong: > Doesn't it break some functionality, like middle mouse click opening in new tab?

Whether this is "breaking"/missing some functionality depends on if you know the underlying thing is an actual link or not. If you are a new user, you probably don't as it looks like a button?

This also probably crosses the classic debate of "should we have links that look like buttons?". Personally not a fan of mixing the two as I think it's confusing UX, since you don't always know what you're gonna get when you see a button.

lucaslcode (approved):

PR #987, opened 2021-04-06
Search Redesign

### LINKED ISSUE #547: Search redesign

Tay0108:
Closes https://github.com/Couchers-org/couchers/issues/547

Basic search redesign to use map. Feel free to roast it :D I might not have much time right now to further work on it so I'd rather do some small fixes if needed and merge or maybe someone else can continue this from now on? Don't want to drag this issue for too long.


lucaslcode:
Converted to draft just to be clearer in the PR list

lucaslcode:
Chris has less time at the moment, so asked me to take this over

lucaslcode:
I think this is a sufficient first step. Not sure if we want to merge it or wait until communities are added back in.

lucaslcode:
@aapeliv I came across the problem that it's basically not possible to filter on clustered points. It's just unclustered atm (even on the initial view without search).

app/frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "react-query": "^3.13.6",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
+    "smoothscroll-polyfill": "^0.4.4",
```
darrenvong, line 26: Is this needed for Safari?

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -0,0 +1,167 @@
...
+  },
+}));
+
+interface SearchResultsListProps {
+  handleResultClick(user: User.AsObject): void;
+  map: MutableRefObject<MaplibreMap | undefined>;
+  selectedResult?: number;
+}
+
+export default function SearchResultsList({
```
darrenvong, line 54: I think this is broken for tablet breakpoint that's not quite at desktop width, as I see this and can't scroll down until it becomes small enough to switch to the horizontal scroller:
Screenshot 2021-04-26 at 00 10 38

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -16,77 +18,103 @@ import {
...
+            </div>
+          </UserSummary>
+          <Hidden mdUp>
+            <LinesEllipsis
+              text={aboutText(user)}
+              maxLine={4}
+              component="p"
+              className={classes.about}
+            />
+          </Hidden>
```
darrenvong, line 103: This seems to display in a different font to what we got in the theme when on breakpoints below md. Probably need to use `Typography` for the `component` prop?

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -1,51 +1,181 @@
...
+      center: [user.lng, user.lat],
+      zoom: selectedUserZoom,
+    });
+  };
+
+  const handleMapUserClick = (ev: any) => {
+    const username = ev.features[0].properties.username;
+    const id = ev.features[0].properties.id;
+    const coords = ev.features[0].geometry.coordinates;
+    handleResultClick({ username, userId: id, lng: coords[0], lat: coords[1] });
```
darrenvong:
```suggestion
    const [lng, lat] = ev.features[0].geometry.coordinates;
    handleResultClick({ username, userId: id, lng, lat });
```

app/frontend/src/features/search/users.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const addClusteredUsersToMap = (
+  map: MaplibreMap,
+  userClickedCallback?: (ev: any) => void
+) => {
+  map.addSource("clustered-users", sources["clustered-users"] as any);
+  addPinImages(map);
+  map.addLayer(layers["clusterLayer"] as any);
+  map.addLayer(layers["clusterCountLayer"] as any);
+  map.addLayer(layers["unclusteredPointLayer"] as any);
```
darrenvong: Are the typings that bad in mapbox GL? Not sure I like how relax the map related code are (we're more or less half back to vanilla JS with a lot of `any`s all over the place) considering it'd be pretty difficult to write tests for these so would be nice if TS can help more to catch mistakes/bugs...

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -1,51 +1,181 @@
...
+
+  const flyToUser = (user: Pick<User.AsObject, "lng" | "lat">) => {
+    map.current?.stop();
+    map.current?.flyTo({
+      center: [user.lng, user.lat],
+      zoom: selectedUserZoom,
+    });
+  };
+
+  const handleMapUserClick = (ev: any) => {
```
darrenvong, line 101: Not sure where best to put this but seems related to this function: is it intentional that clicking a user on the map only highlights them in the search result rather than taking you directly to the user's profile (like how the old map page works)?

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -16,77 +18,103 @@ import {
...
-import { routeToUser } from "routes";
+import LinesEllipsis from "react-lines-ellipsis";
 import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
-  card: {
-    borderRadius: theme.shape.borderRadius * 3,
+  root: {
+    textDecoration: "none",
+    // width: "100%",
```
darrenvong:
```suggestion
```
Doesn't look immediately obvious this will be needed even when we enable other community stuff back on the map?

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -16,77 +18,103 @@ import {
...
-                    className={classes.statusLabel}
-                    display="inline"
-                    variant="subtitle1"
-                  >
-                    {meetupStatusLabels[user.meetupStatus]}
-                  </Typography>
-                </div>
+    <Card
+      id={id}
+      className={classNames(classes.card, classes.root, className)}
```
darrenvong:
```suggestion
      className={classNames(classes.root, className)}
```
Unneeded `classes.card` since no CSS are in the style object there?

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -1,51 +1,181 @@
...
+      marginTop: theme.shape.navPaddingDesktop,
+    },
+  },
+  mapContainer: {
+    flexGrow: 1,
+    position: "relative",
+  },
+  mapResults: {
+    height: "14rem",
+    zIndex: 3,
```
darrenvong:
```suggestion
```
Not sure this is needed since the results still seem to show fine without?

darrenvong (changes requested):
This is a big one, but nice start for sure!

app/frontend/package.json
lucaslcode: Yes

aapeliv:
I played around on this a bit:

1. When you search, and it shows a few people, the computed bounded box has no padding, so e.g. if there's two users, they'll both be half-hidden on teh edges of teh viewport
2. The markers currently don't look good when they get faded: it gives no idea of how many users there actually are, and it's just confusing, it's not "one marker per person". (I also personally think the green markers don't look good in general)

Currently this doesn't have any location search. The way I was envisioning that to work is that the user would search for a location, which would hit Nominatim in the frontend, and then just send coordinates to the backend, which is already supported in the backend.

app/frontend/src/features/search/SearchResult.tsx
lucaslcode: It only accepts html tags. But I'll put the typography css in `classes.about`

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: Yes, if you click them again, it goes to their profile. Taking suggestions for a better idea! (But there needs to be a way to highlight them on the list/map first)

darrenvong: I think that bit where you have to click twice in the search result is probably fine (for me at least). What I meant was on the map itself - I was expecting clicking the marker to also go to the user's profile when it's only for one person but it doesn't do that anymore, which the old map did.

Maybe we should pop up something when the marker is clicked and the user summary (but probably a shorter one than one shown in search result?) is shown in the pop up? Then that could be clickable and takes the user to a profile. Shamelessly lifting the idea from other apps with map. Not sure how tricky that might be to do though, so maybe one for the next PR?

lucaslcode:
Fixed padding. I'm not sure what you mean with the markers, they _are_ one per person. The selected user has a full opacity icon.

You can't scale or change icon colour dynamically unless you're using a special icon format - I haven't experimented with it yet.

Location search to come in next branch.

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: Yeah I think save that for later.

app/frontend/src/features/search/SearchResultsList.tsx
darrenvong: This is still a problem. Maybe we should show this as the horizontal scroller at the `md` breakpoint too? Seems like the easiest fix and I don't think it'd look too bad

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -0,0 +1,167 @@
...
+                  user={result.user}
+                  onClick={handleResultClick}
+                  highlight={result.user.userId === selectedResult}
+                />
+              ) : null
+            )}
+        </HorizontalScroller>
+      ) : (
+        query && (
+          <TextBody className={classes.noResults}>No users found.</TextBody>
```
darrenvong: Put message in constant file?

app/frontend/src/features/search/users.ts
```diff
@@ -0,0 +1,145 @@
...
+  map.on("click", layers["clusterLayer"].id, zoomCluster);
+};
+
+export const addUsersToMap = (
+  map: MaplibreMap,
+  userClickedCallback?: MapClickedCallback
+) => {
+  map.addSource("all-objects", sources["all-objects"]);
+  addPinImages(map);
+  map.addLayer(layers["users"] as any);
```
darrenvong:
```suggestion
  map.addLayer(layers["users"]);
```
Should be fine since you typed everything now?

darrenvong (changes requested):
Last one with minor comments. Not sure if @aapeliv wants to have another look too? Otherwise I can approve after

darrenvong:
It says there are now some merge conflicts... once those are fixed this looks good to me as a starting point 👍

darrenvong:
Actually, might also want to update Storybook for the horizontal scroller and remove the text that says it only scrolls when below `xs`

darrenvong (changes requested):
Actually, might also want to update Storybook for the horizontal scroller and remove the text that says it only scrolls when below `xs`

lucaslcode:
This is done then probably, but do we want to merge as is now @aapeliv or wait until filters/communities are done?

aapeliv:
It'd be good to get something out by Friday when we update release notes, so that we have something new to entertain people with :P. So that rules out communities.

I'd say merge now, but up to you!

darrenvong (approved):

PR #988, opened 2021-04-06
Try to fix build by regenerating yarn.lock

aapeliv:


darrenvong:
Oops, might have merged the package upgrade stuff in a little too aggressively and screwed up the lock file 🙈

aapeliv:
> Oops, might have merged the package upgrade stuff in a little too aggressively and screwed up the lock file 🙈

Well let's hope this fixes it :)

darrenvong:
Uhh I think the problem is the latest version of Storybook has `babel-loader` >= `8.2.2` as a dependency and Create React App doesn't like it and wants `8.1.0` only in the project...

There's no known major bugs between the latest and the slightly older version of Storybook before the change, so might be easiest to revert this for the time being

aapeliv:
Yeah that sounds good. This PR wasn't useful, so closing!

PR #989, opened 2021-04-07
Fix failing build due to storybook upgrade

darrenvong:
Latest version of Storybook uses a higher version of `babel-loader` than Create React App, and CRA doesn't like it as it wants the exact version it needs in its dependency. Reverting to the older lock file and Storybook version is probably the easiest fix for now...


aapeliv (approved):
Looks good. The build succeeded, so feel free to merge.

Issue #990, opened 2021-04-07
Message send button on mobile should be on right of text box


CouchersBot:
Subject: Message send button on mobile should be on right of text box
Description:
My send button is blocked by my url bar (Firefox app has url bar on the bottom) 

Steps:
I used couchers and Firefox at the same time 

Results:
I was blocked from hitting send! Button should be on the right so it's easily accessible and same as other messaging mobile apps 

Backend version: develop-7df73a94
Frontend version: develop-7df73a94
User Agent: Mozilla/5.0 (Android 11; Mobile; rv:87.0) Gecko/87.0 Firefox/87.0
Page: https://app.couchers.org/messages/chats/98
User (spoofable): itsi (2)

justjooshing:
![image](https://user-images.githubusercontent.com/60824744/113839236-6bce6680-9787-11eb-8dd4-f0f65178c20c.png)

Like this @ItsiW?

PR #994, opened 2021-04-07
set send button to side of message box on all screen sizes

justjooshing:
[#990](https://github.com/Couchers-org/couchers/issues/990)

Removed breakpoint so the send button sits next to the input field on all screen sizes

### DIFF
--- a/app/frontend/src/features/messages/useSendFieldStyles.ts
+++ b/app/frontend/src/features/messages/useSendFieldStyles.ts
@@ -2,15 +2,11 @@ import { makeStyles } from "@material-ui/core";
 
 const useSendFieldStyles = makeStyles((theme) => ({
   button: {
-    [theme.breakpoints.up("md")]: {
-      marginInlineStart: theme.spacing(2),
-    },
     display: "block",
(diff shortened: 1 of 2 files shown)


app/frontend/src/features/messages/useSendFieldStyles.ts
```diff
@@ -8,9 +8,9 @@ const useSendFieldStyles = makeStyles((theme) => ({
     display: "block",
     flexShrink: 0,
     marginInlineEnd: "auto",
-    marginInlineStart: "auto",
     marginBottom: theme.spacing(1),
     marginTop: theme.spacing(1),
+    marginLeft: theme.spacing(1),
```
lucaslcode: use `marginInlineStart`
and get rid of `marginInlineEnd` (I think)

lucaslcode: Also maybe just get rid of the above breakpoint to simplify a bit

lucaslcode (changes requested):

lucaslcode (approved):

PR #995, opened 2021-04-07
Add dev/preview banner

aapeliv:
Also enabled grpc web debugging in previews

Screen Shot 2021-04-07 at 10 29 49 am
Screen Shot 2021-04-07 at 10 29 40 am


aapeliv:
Span hiding could be simplified with `useMediaQuery`

aapeliv:
Tried doing it with a [Chip](https://material-ui.com/components/chips/) but can't make the color right

Screen Shot 2021-04-07 at 1 33 39 pm

app/frontend/src/components/EnvironmentBanner.tsx
```diff
@@ -0,0 +1,43 @@
...
+  },
+}));
+
+export function EnvironmentBanner() {
+  const theme = useTheme();
+  const classes = useStyles();
+  const isBelowSm = useMediaQuery(theme.breakpoints.down("sm"));
+
+  if (process.env.REACT_APP_COUCHERS_ENV === "prod") {
+    return <></>;
```
darrenvong:
```suggestion
    return null;
```
React convention is to return null when you don't want to anything

darrenvong (changes requested):
This floating thing looks kinda ugly :p It might look a bit better to have it above the nav bar instead (similar to how the alpha site did it)?

aapeliv:
> This floating thing looks kinda ugly :p It might look a bit better to have it above the nav bar instead (similar to how the alpha site did it)?

It's just that it then offsets other stuff and e.g. the logged out landing index page starts scrolling, etc?

darrenvong:
> It's just that it then offsets other stuff and e.g. the logged out landing index page starts scrolling, etc?

True and it's probably gonna give false CSS positives in dev. How about having it on the bottom left next to where the React Query dev tools usually are?

aapeliv:
Just needs to be prominent enough that people don't get confused (in particular if they end up there by accident)... We already had someone ask me to move their profile from the preview db to prod.

lucaslcode:
Can we not merge this until #1003 is solved, there's a few work arounds I want to try

aapeliv:
> Can we not merge this until #1003 is solved, there's a few work arounds I want to try

Sure

lucaslcode:
Made it removeable, moved to bottom right, and made it pulse since I was worried about it not being seen down there

image

aapeliv:
Does this now obstruct buttons that might appear on the bottom on a phone?

I had `pointerEvents: "none"` and a bit of transparency before so you could click through it, but obviously you can't do that if you can click it away. Or do you just click it away if you want to press a button there?

darrenvong:
> Or do you just click it away if you want to press a button there?

I think that's ok since it will only be on dev builds?

darrenvong (approved):

PR #997, opened 2021-04-07
Simplify bug reporter

### LINKED ISSUE #992: Simplify bug reporting tool
We should only have 3 fields:

1. Brief description of the bug (title)
2. What did you do to trigger the bug?
3. What happened? What should have happened?

### LINKED ISSUE #1018: All input data is lost when clicking outside the bug reporting window
Subject: All input data is lost when clicking outside the bug reporting window
Description:
If I type up a nice, big, long bug report and accidentally click outside of the box, I lose all of the text I typed up. When I click on the bug reporting button again, the text I previously entered is no longer there.

Steps:
Typed up a huge amount of text in a bug report, then clicked outside of the bug reporting box and the bug reporting box closes (and I lost all the input text).

Results:
The bug reporting box should not close when clicking outside of it.

Backend version: develop-1f987511
Frontend version: develop-1f987511
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:87.0) Gecko/20100101 Firefox/87.0
Page: https://app.couchers.org/user
User (spoofable): jesse (57)

aapeliv:
Backend: removes steps to reproduce and adds bug URL to returned message.

Closes #992.
Closes #1018.

I replaced the success message on top of the form with a snackbar that appears after the report. That also clears the form.

Screen Shot 2021-04-07 at 12 52 22 pm


app/frontend/src/features/BugReport.test.tsx
```diff
@@ -151,15 +143,14 @@ describe("BugReport", () => {
 
       expect(await screen.findByRole("progressbar")).toBeVisible();
       const successMessage =
-        "Thank you for reporting that bug and making Couchers better, a report was sent to the devs! The bug ID is 1";
+        "Thank you for reporting that bug and making Couchers better, a report was sent to the developers! The bug ID is 1";
```
darrenvong: Use the `getBugReportSuccessMessage` you made in the constants file to simplify this

app/frontend/src/components/Alert.tsx
```diff
@@ -26,10 +26,9 @@ export default function Alert({
 }: AlertProps) {
   const classes = useStyles();
 
-  const oldErrorKey = Object.keys(
-    grpcErrorStrings
-  ).find((oldError): oldError is ObscureGrpcErrorMessages =>
-    children.includes(oldError)
+  const oldErrorKey = Object.keys(grpcErrorStrings).find(
+    (oldError): oldError is ObscureGrpcErrorMessages =>
+      children?.toString().includes(oldError) ?? false
```
darrenvong: We aren't really meant to interact with `children` directly this way if it's openly allowing any `ReactNode`. It'd be safer to use one of the [`React.Children` utility](https://reactjs.org/docs/react-api.html#reactchildren) on `children` first and apply this to the individual child element.

@lucaslcode I think we tried to keep it open to accept `ReactNode` but had to change this to only allow string, otherwise it doesn't map the obscure gprc error correctly to the nicer ones right?

darrenvong (changes requested):

app/frontend/src/components/Alert.tsx
aapeliv: OK, I've decided not to touch it, not worth it

app/frontend/src/features/BugReport.test.tsx
aapeliv: OK

aapeliv:
Yeah, that link thing was too complicated, reverted now, even though the backend returns a URL that could be used in the future. Someone smarter or more dedicated to having a link is welcome to add it :P

lucaslcode:
Just use the Mui alert directly instead of our alert, this once?

darrenvong:
Yeah should be fine to update that in `SuccessSnackbar`, since it's meant to be for successful cases so shouldn't have to worry about the need to remap grpc error messages.

aapeliv:
Better @darrenvong?

app/frontend/src/features/BugReport.tsx
```diff
@@ -9,27 +9,33 @@ import {
...
-  STEPS,
   SUBMIT,
   WARNING,
 } from "features/constants";
 import { Error as GrpcError } from "grpc-web";
+import { ReportBugRes } from "pb/bugs_pb";
 import React, { useState } from "react";
 import { useForm } from "react-hook-form";
 import { useMutation } from "react-query";
+import { Link } from "react-router-dom";
```
darrenvong: Sorry just seen this... I think we want to use `Link` from Material UI instead like Lucas said so we get the right link styling from the theme. Also `Link` from `react-router-dom` is only meant for links between pages in the app and not to external URLs

aapeliv: Yeah, thanks. My bad, I was told in the meeting to use it... but I was having trouble with it and this Link was used in other spots too, so I used it...

Fixed now

darrenvong: Yeah it's confusing they are named the same thing, I'm sure you won't be the last person to make this mistake :p

app/frontend/src/features/BugReport.tsx
```diff
@@ -86,6 +102,17 @@ export default function BugReport() {
 
   return (
     <>
+      {bug && (
+        <SuccessSnackbar>
+          <>
+            {BUG_REPORT_SUCCESS}
+            <Link href={bug.bugUrl} target="_blank">
```
darrenvong:
```suggestion
            <Link href={bug.bugUrl}>
```
I don't think this needs to be opened in a new tab/window? It's bad for accessibility to do this by default without letting the user know in some way.

darrenvong:
Build is failing with a broken test - it's to do with the changed API so the mock needs updating

darrenvong (changes requested):
Screenshot 2021-04-08 at 00 37 16
Actually the test is doing its job at catching this... the bug ID link isn't showing up when I just submitted a test bug 🤔

aapeliv:
>  Actually the test is doing its job at catching this... the bug ID link isn't showing up when I just submitted a test bug 🤔

Is this against a local backend updated with this PR?

darrenvong:
Ahh I did it against the preview, so the API is probably still outdated there. How do I enable the bug tool without the API key locally though?

kalvdans (dismissed):
Looks good backend-wise! Didn't review frontend.

app/frontend/src/features/BugReport.tsx
```diff
@@ -103,54 +127,43 @@ export default function BugReport() {
         aria-labelledby="bug-reporter"
         open={isOpen}
         onClose={handleClose}
+        // won't close when clicking outside
+        disableBackdropClick
+        // won't clsoe when pressing esc
+        disableEscapeKeyDown
```
darrenvong: Why are we disabling close by backdrop and with escape? That seems to go against the default expected behaviour of a modal...

aapeliv: People have been complaining about accidentally clicking outside the modal and then losing their hard work in writing a bug report. That I think is fair enough.

Happy to remove the escape disable though.

What do you think?

darrenvong: I see... maybe doing this as a modal isn't the best way? What about having it on a separate page altogether?

aapeliv: I think it's quite handy as a modal: people don't have to navigate back and forth when reporting bugs?

darrenvong: Yeah ok, let's re-enable escape but disable background click. Just looked around and that's not actually an expected behaviour in terms of accessibility, so it must be an additional "nice to have" pattern people/Material UI implement

darrenvong (approved):

PR #1005, opened 2021-04-07
Add bold to unread host/surfing request messages

### LINKED ISSUE #923: Unread request messages are not bold
Unread chats are bold, but not host requests

trevvorko:
Added styling to messages in both the hosting and surfing tab to be in
bold when the last message sent is unread.

Resolves #923

### DIFF
--- a/app/frontend/src/features/messages/requests/HostRequestListItem.tsx
+++ b/app/frontend/src/features/messages/requests/HostRequestListItem.tsx
@@ -7,6 +7,7 @@ import {
 } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 import { Skeleton } from "@material-ui/lab";
+import classNames from "classnames";
 import Avatar from "components/Avatar";
 import TextBody from "components/TextBody";
 import useAuthStore from "features/auth/useAuthStore";
@@ -31,6 +32,7 @@ const useStyles = makeStyles((theme) => ({
   hostStatusIcon: {
     marginInlineEnd: theme.spacing(1),
   },
+  unread: { fontWeight: "bold" },
 }));
 
 export interface HostRequestListItemProps {
@@ -48,7 +50,8 @@ export default function HostRequestListItem({
   const { data: otherUser, isLoading: isOtherUserLoading } = useUser(
     isHost ? hostRequest.fromUserId : hostRequest.toUserId
   );
-
+  const isUnread =
+    hostRequest.lastSeenMessageId !== hostRequest.latestMessage?.messageId;
   //define the latest message author's name and
   //control message target to use in short message preview
   const authorName =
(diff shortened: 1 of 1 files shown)


app/frontend/src/service/client.ts
```diff
@@ -93,7 +93,7 @@ if (process.env.NODE_ENV === "development") {
     apis.pages,
     apis.references,
     apis.requests,
-    apis.search,
```
trevvorko: I ran `yarn format` and it adjusted this tab here. Let me know if I should recommit without this change.

trevvorko:
@aapeliv I can't seem to add a reviewer or labels, I think this is because I am not considered a collaborator to the repo.

app/frontend/src/features/messages/requests/HostRequestListItem.tsx
```diff
@@ -105,7 +110,7 @@ export default function HostRequestListItem({
                 true
               )}`}
             </Typography>
-            <TextBody noWrap>
+            <TextBody noWrap className={isUnreadClass}>
```
darrenvong: We use the `classNames` package for dealing with this kind of logic, so you can do something like this instead:

```suggestion
            <TextBody noWrap className={classNames({ [classes.unread]: isUnread })} />
```

app/frontend/src/features/messages/requests/HostRequestListItem.tsx
```diff
@@ -48,6 +49,10 @@ export default function HostRequestListItem({
   const { data: otherUser, isLoading: isOtherUserLoading } = useUser(
     isHost ? hostRequest.fromUserId : hostRequest.toUserId
   );
+  const isUnreadClass =
+    hostRequest.lastSeenMessageId !== hostRequest.latestMessage?.messageId
+      ? classes.unread
+      : "";
```
darrenvong:
```suggestion
  const isUnread = hostRequest.lastSeenMessageId !== hostRequest.latestMessage?.messageId;
```
When used with `classNames` suggestion below

app/frontend/src/service/client.ts
darrenvong: Only a small change here so it's cool to keep this in this case :)

darrenvong (changes requested):
Thanks for your first contribution!

app/frontend/src/features/messages/requests/HostRequestListItem.tsx
trevvorko: This makes sense, thanks for letting me know. I made the change in the latest commit.

darrenvong:
Can you rebase with `develop` as well please? The repo is changing fast and stuff has broken as a result of branches not being up to date with `develop` so just want to make doubly sure this will be fine.

trevvorko:
Yeah I think I rebased correctly, let me know if there's something else I haven't done correctly, and thanks for being so helpful.

darrenvong (approved):
Looks good! Feel free to take a look at other issues you might be interested in working on

PR #1006, opened 2021-04-07
Donations

aapeliv:


app/backend/src/couchers/servicers/donations.py
```diff
@@ -0,0 +1,59 @@
...
+
+class Donations(donations_pb2_grpc.DonationsServicer):
+    def InitiateDonation(self, request, context):
+        if not enabled:
+            context.abort(grpc.StatusCode.UNAVAILABLE, errors.DONATIONS_DISABLED)
+
+        with session_scope() as session:
+            user = session.query(User).filter(User.id == context.user_id).one()
+
+            if request.amount < 2:
```
lennart-at-hubper, line 28: If the amount is in cents, this checks for requests < 2 cents, while an error message above seems to say that you want to check for < 2 dollars. I think this should be < 200, or you save the output of ` request.amount * 100` in a variable and use that.

lennartvanlaake: Scratch that, the code makes sense, my comment is wrong.

aapeliv:
@ItsiW: please write a nice email for when we receive a donation.

We can also optionally have a regime of 3 emails:
* when users do a one-off donation
* when users set up a recurring one
* when a recurring one is received

aapeliv:
This is now just missing some tests, I think. Then it's ready for the first iteration.

In the future we need some kind of way to list and cancel recurring donations, as well as download receipts/invoices. These are already stored in the db, but they just need to be exposed to the user.

Jesus82:
Hi @aapeliv, I have finished the layout for the donations page https://github.com/Couchers-org/couchers/pull/1553, but I see this commit is still not merged. Should I wait for this or could I move forward with the frontend in some way?

aapeliv:
> Hi @aapeliv, I have finished the layout for the donations page #1553, but I see this commit is still not merged. Should I wait for this or could I move forward with the frontend in some way?

Yes, please move forward. The API here is stable, this is just missing testing.

Do you have the backend set up locally, or are you testing against the dev instance? I can try to get this merged in the next day or two so you can test against the dev instance.

Jesus82:
> > Hi @aapeliv, I have finished the layout for the donations page #1553, but I see this commit is still not merged. Should I wait for this or could I move forward with the frontend in some way?
> 
> Yes, please move forward. The API here is stable, this is just missing testing.
> 
> Do you have the backend set up locally, or are you testing against the dev instance? I can try to get this merged in the next day or two so you can test against the dev instance.

Yep, if you can merge that could be awesome. I did only the layout, not the functionalities. As I see you have kind of done it, I will take your page and then copy some of my html there and that'll do the trick :)

aapeliv:
@Couchers-org/backend: this is now ready for review. I'm still working on more testing, but the logic for v1 is ready to merge. We should do this asap so @Jesus82 can easily develop against the dev instance.

app/backend/src/couchers/urls.py
```diff
@@ -52,3 +52,11 @@ def media_upload_url(path):
...
 def change_email_link(confirmation_token):
     return f"{config['BASE_URL']}/confirm-email/{confirmation_token}"
+
+
+def donation_cancelled_url():
+    return f"{config['BASE_URL']}/donate?cancelled=true"
+
+
+def donation_success_url():
+    return f"{config['BASE_URL']}/donate?success=true&session={{CHECKOUT_SESSION_ID}}"
```
kalvdans: What is the f"{{foo}}" supposed to mean?

app/proto/stripe.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
```
kalvdans, line 1: Move file to backend-specific place

app/proto/donations.proto
```diff
@@ -0,0 +1,24 @@
...
+package org.couchers.api.donations;
+
+import "annotations.proto";
+
+service Donations {
+  option (auth_level) = AUTH_LEVEL_SECURE;
+
+  rpc InitiateDonation(InitiateDonationReq) returns (InitiateDonationRes) {
+    // Initiate a donation
+  }
```
kalvdans, line 15: Document how a client is supposed to proceed with the donation (or refer to stripe docs)

app/backend/src/couchers/urls.py
aapeliv: I want the final string to be something like `https://couchers.org/donate?success=true&session={CHECKOUT_SESSION_ID}`: then stripe will fill in the checkout session id

app/proto/stripe.proto
aapeliv: This is not backend specific: it needs to be publicly available. It won't be called by users, but it'll be called by the stripe servers, and the proxy needs to do the translation. We could potentially create a new set of proto files that are public-facing but not part of the API used by users, but I think that's more hassle?

kalvdans: Aha, it is not a protobuf-based protocol at all, and documented at https://stripe.com/docs/api/events/types . How can we do this without a .proto file?

app/backend/src/couchers/servicers/donations.py
```diff
@@ -0,0 +1,167 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Donations(donations_pb2_grpc.DonationsServicer):
+    def __init__(self):
+        self.enabled = config["ENABLE_DONATIONS"]
+
+        if self.enabled:
+            stripe.api_key = config["STRIPE_API_KEY"]
```
kalvdans: The stripe Python modules uses global variables, seems it won't be rewritten anytime soon ( https://stripe.com/docs/api/events/types ). I couldn't find any documentation for the Python bindings either.

Better make json requests using the `request` python module. The json API has [proper docs](https://stripe.com/docs/api/authentication) too.

app/backend/src/couchers/servicers/donations.py
```diff
@@ -0,0 +1,167 @@
...
+                    # metadata allows us to store arbitrary metadata for ourselves
+                    metadata={"user_id": user.id},
+                )
+                user.stripe_customer_id = customer.id
+                # commit since we only ever want one stripe customer id per user, so if the rest of this api call fails, this will still be saved in the db
+                session.commit()
+
+            if request.recurring:
+                item = {
+                    "price": config.get("STRIPE_RECURRING_PRODUCT_ID"),
```
kalvdans: Use `config["STRIPE_RECURRING_PRODUCT_ID"]` to raise an error on the backend side instead of passing `null` to the stripe API.

More occurrences of .get() in this merge request could be converted as well.

app/backend/src/couchers/servicers/donations.py
```diff
@@ -0,0 +1,167 @@
...
+                payload=request.data,
+                sig_header=headers.get("stripe-signature"),
+                secret=config.get("STRIPE_WEBHOOK_SECRET"),
+            )
+            data = event["data"]
+            event_type = event["type"]
+            event_id = event["id"]
+            data_object = data["object"]
+        except Exception as e:
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.STRIPE_WEBHOOK_FAILED_SIG)
```
kalvdans: Specify which exceptions that we could expect (I guess `KeyError`)

app/backend/src/couchers/urls.py
kalvdans: Where is it documented? (Couldn't find this syntax at https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-success_url )

Also could we write the expression without f-strings to make it easier to see what the resulting string would be like?

kalvdans (changes requested):

aapeliv:
> Aha, it is not a protobuf-based protocol at all, and documented at https://stripe.com/docs/api/events/types . How can we do this without a .proto file?

We could in theory run a HTTP server side-by-side with the grpc one. But I really see no need for that extra complexity for this case. It's similar to the GIS bits.

app/backend/src/couchers/servicers/donations.py
aapeliv: Do we even really need to catch it here? We already have functionality for catching general exceptions and they'll end up as useful logs in our tracing?

app/backend/src/couchers/urls.py
aapeliv: I got rid of this

app/backend/src/couchers/servicers/donations.py
kalvdans: No, I guess stripe doesn't care about our HTTP return code value anyway. Just remove all error handing.

kalvdans:
> We could in theory run a HTTP server side-by-side with the grpc one. But I really see no need for that extra complexity for this case. It's similar to the GIS bits.

We could reuse the same HTTP server but I guess grpc has taken over it all. When I'm bored I can see if I can get the same HTTP server to serve both protobuf and json.

app/backend/src/couchers/servicers/donations.py
```diff
@@ -0,0 +1,161 @@
...
+                    donation.stripe_subscription_id = data_object["subscription"]
+            else:
+                raise Exception("Unknown payment type")
+        elif event_type == "payment_intent.succeeded":
+            customer_id = data_object["customer"]
+            with session_scope() as session:
+                user = session.execute(select(User).where(User.stripe_customer_id == customer_id)).scalar_one()
+                invoice_data = data_object["charges"]["data"][0]
+                # amount comes in cents
+                amount = float(invoice_data["amount"]) / 100
```
ShmuelTreiger, line 143: Should we round this to two decimal places?

app/backend/src/tests/test_donations.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # evt_1JB4HyE5kUmYuPWzATTZL7Mt: charge.succeeded
+    fire_stripe_event("evt_1JB4HyE5kUmYuPWzATTZL7Mt")
+
+    # evt_1JB4HyE5kUmYuPWzINvGWlRu: checkout.session.completed
+    fire_stripe_event("evt_1JB4HyE5kUmYuPWzINvGWlRu")
+
+    ## Now finally check everything was added to the DB
+    with session_scope() as session:
+        donation = session.execute(select(OneTimeDonation)).scalar_one()
```
ShmuelTreiger, line 98: Should we add a where? Better to be more robust/precise, especially protects against the test changing.

app/backend/src/tests/test_donations.py
```diff
@@ -0,0 +1,259 @@
...
+
+    # evt_1JB4K0E5kUmYuPWzSj00EC0m: invoice.paid
+    fire_stripe_event("evt_1JB4K0E5kUmYuPWzSj00EC0m")
+
+    # evt_1JB4K0E5kUmYuPWzGQN86BFq: invoice.payment_succeeded
+    fire_stripe_event("evt_1JB4K0E5kUmYuPWzGQN86BFq")
+
+    ## Now finally check everything was added to the DB
+    with session_scope() as session:
+        donation = session.execute(select(RecurringDonation)).scalar_one()
```
ShmuelTreiger, line 210: ditto

app/backend/src/tests/test_donations.py
```diff
@@ -0,0 +1,259 @@
...
+        assert donation.user_id == user_id
+        assert donation.amount == 25
+        assert (
+            donation.stripe_checkout_session_id == "cs_test_a1xQjLyaMWpDj2Quqc5JBFnVQXHcVuS5bu53Nr8Kz6xDTuXERmqNzz3dUi"
+        )
+        assert donation.stripe_subscription_id == "sub_Johj2mPZu6V3CG"
+
+        invoice = session.execute(select(Invoice)).scalar_one()
```
ShmuelTreiger, line 218: again

app/backend/src/tests/test_donations.py
```diff
@@ -0,0 +1,259 @@
...
+        donation = session.execute(select(OneTimeDonation)).scalar_one()
+        assert donation.user_id == user_id
+        assert donation.amount == 100
+        assert (
+            donation.stripe_checkout_session_id == "cs_test_a1ScieS4Qtnak4fI7qhUChYz0XBBzwRBKMjNESus7eiF0WvIcV07i3TFx5"
+        )
+        assert donation.stripe_payment_intent_id == "pi_1JB4HaE5kUmYuPWz9O9m5vd9"
+        assert donation.paid
+
+        invoice = session.execute(select(Invoice)).scalar_one()
```
ShmuelTreiger, line 107: also

app/backend/templates/donation_received.md
```diff
@@ -0,0 +1,25 @@
+---
+subject: "Thank you for your donation to Couchers.org!"
+---
+
+{% from "macros.html" import button, link, support_email %}
+
+Hi {{ user.name|couchers_escape }},
+
+Thank you for your donation of $ {{ amount }} to the Couchers.org Foundation. Your donation makes our work and the wellbeing of the community possible, and the project wouldn't exist without the support of you and other financial contributors!
```
ShmuelTreiger: IMO "without your support!"

No need to talk about other people

app/backend/templates/donation_received.md
```diff
@@ -0,0 +1,25 @@
...
+
+{% if html %}
+
+{{ button("Invoice", receipt_url)|couchers_safe }}
+
+Alternatively, click the following link: {{ link(receipt_url, html)|couchers_safe }}.
+
+{% else %}
+<{{ receipt_url|couchers_escape }}>
+{% endif %}
```
ShmuelTreiger: Is this URL permanent? And do they need to save the receipt or the URL? Whatever the answer, it should be in the email.

If they can view the invoice later from their account, that should be stated here as well.

Alternatively, it would be great if we could just attach the invoice for their records. Maybe future issue

app/backend/src/couchers/servicers/donations.py
aapeliv: We only do dollar amount donations. Stripe uses integer cents. I think it's fine to keep as is: if we put in dollars it should only return dollars...

app/backend/src/tests/test_donations.py
aapeliv: This is already robust to there being exactly one donation (it'll raise an exception and fail if there's none or more than one), I think that's enough: we then assert that it's the correct one.

app/backend/templates/donation_received.md
aapeliv: It's a URL from stripe, presumably permanent.

We will eventually add something to see all receipts and so on in the frontend, but I think this is fine for now. Email is also a good record.

app/backend/templates/donation_received.md
aapeliv: I'm going to update the email with Itsi's proper one

ShmuelTreiger:
Looking at this again, this architecture restricts us to accepting donations from active users. Do we want to restrict ourselves in this way? Will we set up alternative donation channels for non-users?

aapeliv:
> Looking at this again, this architecture restricts us to accepting donations from active users. Do we want to restrict ourselves in this way? Will we set up alternative donation channels for non-users?

Yes, that's the plan for now. Otherwise it's a pain: then we'd have to collect and verify other email addresses, etc.

ShmuelTreiger (approved):

PR #1010, opened 2021-04-07
Change dialog styling to use mui stuff

aapeliv:
Before:

Screen Shot 2021-04-07 at 6 51 35 pm
Screen Shot 2021-04-07 at 6 51 41 pm
Screen Shot 2021-04-07 at 6 53 21 pm

After:

Screen Shot 2021-04-07 at 6 52 12 pm
Screen Shot 2021-04-07 at 6 52 21 pm
Screen Shot 2021-04-07 at 6 53 33 pm

### DIFF
--- a/app/frontend/src/components/Dialog.tsx
+++ b/app/frontend/src/components/Dialog.tsx
@@ -29,24 +29,6 @@ const useStyles = makeStyles((theme) => ({
   contentText: {
     padding: theme.spacing(2),
   },
-  dialog: {
-    padding: 0,
-    "& .MuiDialog-scrollPaper": {
-      display: "flex",
-      flexDirection: "column",
-      justifyContent: "start",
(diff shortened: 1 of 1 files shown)


darrenvong:
Is the main difference here the dialog is centred properly again? I'll have a quick look visually once CI builds

aapeliv:
> Is the main difference here the dialog is centred properly again? I'll have a quick look visually once CI builds

Yes, and the sizing is right on medium screens: before it hit the breakpoint it would look really awful:

Screen Shot 2021-04-07 at 7 00 09 pm

darrenvong (approved):
Apparently the label overflow still happens on iPhone SE when I played around with this with the Safari dev tools. I think this is a wider problem to how 💩  the input is built in Material UI though. This is a big improvement so happy to have this merged as it is.

PR #1025, opened 2021-04-08
replace 192 and 512 logos

ItsiW:


aapeliv:
These still look blurry and not sharp to me?

ItsiW:
Are you sure? They're smallish images that shouldn't be blown up too much

aapeliv:
Yeah but they should be as sharp as they can be at 512/192 px when blown up. It'll look equally good when downsampled to be smaller.

aapeliv (approved):

PR #1027, opened 2021-04-08
Add contributor form into app

aapeliv:
There's a shared component, which is embedded on the dashboard. The backend tracks whether logged in users have filled in the form or not. Form submissions are sent to an AWS lambda function. I've also added a non-logged in route `/contribute` for those who have not signed up (for whatever reason?). Several form inputs are hidden and filled in automatically if the user is signed in.

* ~~Itsi would like the radio group to be turned into a set of checkboxes, but that's a lot of effort~~
* ~~Add a separate username field + lambda func tweaks~~
* ~~Would be good to get the user's email into the form so the form backend can email the person~~
* Needs some tweaks in the AWS lambda functions: email needs to be updated to reflect current situation, and emails to us need to be updated to include username.

It'd be great to eventually move the lambda function concoction into the backend, but that can come later.

### DIFF
--- a/app/frontend/src/features/Contribute.tsx
+++ b/app/frontend/src/features/Contribute.tsx
@@ -1,0 +1,103 @@
+import { makeStyles, Typography } from "@material-ui/core";
+import { LOGIN } from "features/auth/constants";
+import DesktopAuthBg from "features/auth/resources/desktop-auth-bg.jpg";
+import { Link } from "react-router-dom";
+import { loginRoute } from "routes";
+
+import { COUCHERS } from "../constants";
+import ContributorForm, {
+  CONTRIBUTE,
+  FILL_IN_THE_FORM,
+  JOIN_THE_TEAM,
+} from "./contributorForm";
+
+const useStyles = makeStyles((theme) => ({
+  bg: {
+    height: "100%",
+    width: "100%",
+    backgroundColor: theme.palette.common.white,
+    // backgroundAttachment: "scroll, local",
+    backgroundPosition: "top center",
+    backgroundRepeat: "no-repeat",
+    backgroundImage: `url(${DesktopAuthBg})`,
+    backgroundSize: "cover",
+    position: "fixed",
+    zIndex: -1000,
+  },
+  page: {
+    boxSizing: "border-box",
+    display: "flex",
+    height: "100vh",
+    alignItems: "flex-start",
+    flexDirection: "column",
+    padding: 0,
+    width: "100%",
+  },
+  content: {
+    flexDirection: "column",
+    height: "100%",
+    margin: "0 auto",
+  },
+  nav: {
+    display: "flex",
+  },
+  link: {
+    borderRadius: theme.shape.borderRadius / 3,
+    color: theme.palette.common.white,
+    fontSize: "1.25rem",
+    fontWeight: 500,
+    textAlign: "center",
+    padding: theme.spacing(1, 2),
+  },
+  loginLink: {
+    border: `1px solid ${theme.palette.primary.main}`,
+    marginRight: theme.spacing(3),
+  },
+  formWrapper: {
+    backgroundColor: theme.palette.background.default,
+    borderRadius: theme.shape.borderRadius / 3,
+    padding: theme.spacing(5, 3),
+  },
+  header: {
+    width: "100%",
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    padding: theme.spacing(2, 4),
+    marginBottom: theme.spacing(2),
+  },
+  logo: {
+    color: theme.palette.secondary.main,
+    fontFamily: "'Mansalva', cursive",
+    fontSize: "2rem",
+  },
+}));
+
+export default function Contribute() {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.page}>
+      <div className={classes.bg} />
+      <header className={classes.header}>
+        <div className={classes.logo}>{COUCHERS}</div>
+        <nav className={classes.nav}>
+          <Link
+            to={loginRoute}
+            className={`${classes.link} ${classes.loginLink}`}
+          >
+            {LOGIN}
+          </Link>
+        </nav>
+      </header>
+      <div className={classes.content}>
+        <div className={classes.formWrapper}>
+          <Typography variant="h1">{CONTRIBUTE}</Typography>
+          <Typography variant="subtitle1">{JOIN_THE_TEAM}</Typography>
+          <p>{FILL_IN_THE_FORM}</p>
+          <ContributorForm />
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/app/frontend/src/features/Contribute.tsx
+++ b/app/frontend/src/features/Contribute.tsx
@@ -5,11 +5,7 @@ import { Link } from "react-router-dom";
 import { loginRoute } from "routes";
 
 import { COUCHERS } from "../constants";
-import ContributorForm, {
-  CONTRIBUTE,
-  FILL_IN_THE_FORM,
-  JOIN_THE_TEAM,
-} from "./contributorForm";
+import ContributorForm, { CONTRIBUTE, JOIN_THE_TEAM } from "./contributorForm";
 
 const useStyles = makeStyles((theme) => ({
   bg: {
@@ -94,7 +90,6 @@ export default function Contribute() {
         <div className={classes.formWrapper}>
           <Typography variant="h1">{CONTRIBUTE}</Typography>
           <Typography variant="subtitle1">{JOIN_THE_TEAM}</Typography>
-          <p>{FILL_IN_THE_FORM}</p>
           <ContributorForm />
         </div>
       </div>
--- a/app/frontend/src/features/Contribute.tsx
+++ b/app/frontend/src/features/Contribute.tsx
@@ -12,7 +12,6 @@ const useStyles = makeStyles((theme) => ({
     height: "100%",
     width: "100%",
     backgroundColor: theme.palette.common.white,
-    // backgroundAttachment: "scroll, local",
     backgroundPosition: "top center",
     backgroundRepeat: "no-repeat",
     backgroundImage: `url(${DesktopAuthBg})`,
@@ -38,7 +37,7 @@ const useStyles = makeStyles((theme) => ({
     display: "flex",
   },
   link: {
-    borderRadius: theme.shape.borderRadius / 3,
+    borderRadius: theme.shape.borderRadius,
     color: theme.palette.common.white,
     fontSize: "1.25rem",
     fontWeight: 500,
@@ -51,7 +50,7 @@ const useStyles = makeStyles((theme) => ({
   },
   formWrapper: {
     backgroundColor: theme.palette.background.default,
-    borderRadius: theme.shape.borderRadius / 3,
+    borderRadius: theme.shape.borderRadius,
     padding: theme.spacing(5, 3),
   },
   header: {
--- a/app/frontend/src/features/contributorForm/ContributorForm.tsx
+++ b/app/frontend/src/features/contributorForm/ContributorForm.tsx
@@ -1,0 +1,318 @@
+import {
+  FormControl,
+  FormControlLabel,
+  FormHelperText,
+  FormLabel,
+  makeStyles,
+  Radio,
+  RadioGroup,
+} from "@material-ui/core";
+import classNames from "classnames";
+import Alert from "components/Alert";
+import Button from "components/Button";
+import CircularProgress from "components/CircularProgress";
+import TextField from "components/TextField";
+import { useAuthContext } from "features/auth/AuthProvider";
+import { useUser } from "features/userQueries/useUsers";
+import React, { useEffect, useState } from "react";
+import { Controller, useForm } from "react-hook-form";
+
+import {
+  AGE,
+  CONTRIBUTE_LABEL,
+  CONTRIBUTE_OPTIONS,
+  CONTRIBUTE_REQUIRED,
+  EMAIL,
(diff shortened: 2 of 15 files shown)


aapeliv:
@kalvdans: could you please review backend bit, it's tiny :)

aapeliv:
I've done an awful hack to get checkboxes and have them turned nicely into a comma-delimited string...

kalvdans (dismissed):
backend looks solid

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+  } = useForm<ContributorInputs>({
+    defaultValues: {},
+    mode: "onBlur",
+    shouldUnregister: false,
+  });
+
+  useEffect(() => {
+    if (authState.authenticated) {
+      (async () => {
+        const info = await service.account.getContributorFormInfo();
```
darrenvong: It's probably easier and safer to do this with React Query, since it's easy for this kinda async stuff to run into the setting state on unmounted component warning (happens whenever some async stuff is still loading but user navigates away from page before it finishes loading) that could cause memory leaks. React Query don't have this issue since it's global outside of React.

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+        setValue("username", info.username);
+        setValue("name", info.name);
+        setValue("email", info.email);
+        setValue("age", info.age);
+        setValue("gender", info.gender);
+        setValue("location", info.location);
```
darrenvong: Don't these fields also exist in the form? It's better to set them as `defaultValue` on the form fields instead, otherwise React hook form warns about this

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+      if (result.success) {
+        if (authState.authenticated) {
+          await service.account.markContributorFormFilled();
+        }
+        setSuccess(true);
+      } else {
+        setError("An unknown error occured");
+      }
+    } catch (e) {
+      setError(e.message);
```
darrenvong: Use mutation from React Query, again handles all of this error and loading state so slightly less boilerplate. You can more or less take the bit in the try-block and put it in the mutation but without needing the try-catch and setState/error stuff (react query catches error for us when things go wrong).

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                    name="expertise"
+                    label={EXPERTISE_LABEL}
+                    helperText={EXPERTISE_HELPER}
+                    fullWidth
+                    multiline
+                    rows={4}
+                    rowsMax={6}
+                  />
+                  <Button
+                    classes={{}}
```
darrenvong:
```suggestion
```

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+    const currentChoices = contribute.split(",").filter((v) => !!v);
+    if (currentChoices.includes(option)) {
+      setContribute(currentChoices.filter((opt) => opt !== option).join(","));
+    } else {
+      setContribute(currentChoices.concat(option).join(","));
+    }
+  };
+
+  return (
+    <>
```
darrenvong, line 169: This outer fragment `<>` is unnecessary since it's a single ternary

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+      {loading || authState.loading ? (
+        <CircularProgress />
+      ) : (
+        <>
+          {filled ? (
+            <>
+              <p>{ALREADY_FILLED_IN}</p>
```
darrenvong: Use `Typography` so it uses styles from the theme

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+              <Button onClick={() => setFilled(false)}>{FILL_IN_AGAIN}</Button>
+            </>
+          ) : (
+            <>
+              {success ? (
+                <>
+                  <p>{SUCCESS_MSG}</p>
```
darrenvong: Use `Typography`

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                  {!authState.authenticated && (
+                    <>
+                      <p>{PLEASE_SIGN_UP}</p>
```
darrenvong: Same as above

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                      <Button component={Link} to={signupRoute}>
+                        {SIGN_UP}
+                      </Button>
+                    </>
+                  )}
+                </>
+              ) : (
+                <form onSubmit={submit}>
+                  <p>{FILL_IN_THE_FORM}</p>
```
darrenvong: Use `Typography`

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                  {!authState.authenticated && (
+                    <>
+                      <Button component={Link} to={signupRoute}>
+                        {SIGN_UP}
+                      </Button>
+                      <p>{YOU_CAN_ALSO}</p>
```
darrenvong: Use `Typography`

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                        name="email"
+                        inputRef={register({
+                          required: EMAIL_REQUIRED,
+                        })}
+                        helperText={errors?.email?.message ?? " "}
+                        error={!!errors?.email?.message}
+                        className={classNames({
+                          [classes.hidden]: authState.authenticated,
+                        })}
+                      />
```
darrenvong: Better to conditionally not render this if user is authenticated rather than hiding it

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                  <Controller
+                    id="contribute"
+                    control={control}
+                    name="contribute"
+                    defaultValue=""
+                    rules={{ required: CONTRIBUTE_REQUIRED }}
+                    render={({ onChange, value }) => (
+                      <FormControl>
+                        <FormLabel>{CONTRIBUTE_LABEL}</FormLabel>
+                        <FormGroup aria-label="contribute">
```
darrenvong: Put the aria-label in constant file too

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                            />
+                          ))}
+                        </FormGroup>
+                        <FormHelperText error={!!errors?.contribute?.message}>
+                          {errors?.contribute?.message ?? " "}
+                        </FormHelperText>
+                      </FormControl>
+                    )}
+                  />
+                  <p>{QUESTIONS_OPTIONAL}</p>
```
darrenvong: Use `Typography`

app/frontend/src/features/Contribute.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+import { COUCHERS } from "../constants";
+import ContributorForm, { CONTRIBUTE, JOIN_THE_TEAM } from "./contributorForm";
+
+const useStyles = makeStyles((theme) => ({
+  bg: {
+    height: "100%",
+    width: "100%",
+    backgroundColor: theme.palette.common.white,
+    // backgroundAttachment: "scroll, local",
```
darrenvong:
```suggestion
    // backgroundAttachment: "scroll, local",
```
Remove if we don't need this?

app/frontend/src/features/Contribute.tsx
```diff
@@ -0,0 +1,98 @@
...
+  content: {
+    flexDirection: "column",
+    height: "100%",
+    margin: "0 auto",
+  },
+  nav: {
+    display: "flex",
+  },
+  link: {
+    borderRadius: theme.shape.borderRadius / 3,
```
darrenvong:
```suggestion
    borderRadius: theme.shape.borderRadius,
```
I've changed `borderRadius` value in the theme recently if you rebase with `develop` so don't need to divide by 3 anymore

app/frontend/src/features/Contribute.tsx
```diff
@@ -0,0 +1,98 @@
...
+    textAlign: "center",
+    padding: theme.spacing(1, 2),
+  },
+  loginLink: {
+    border: `1px solid ${theme.palette.primary.main}`,
+    marginRight: theme.spacing(3),
+  },
+  formWrapper: {
+    backgroundColor: theme.palette.background.default,
+    borderRadius: theme.shape.borderRadius / 3,
```
darrenvong:
```suggestion
    borderRadius: theme.shape.borderRadius,
```

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+    }
+    setLoading(false);
+  });
+
+  const toggleCheckbox = (
+    option: string,
+    contribute: string,
+    setContribute: (s: string) => void
+  ) => {
+    const currentChoices = contribute.split(",").filter((v) => !!v);
```
darrenvong, line 158: What's wrong with the original form value that comes from react hook form's `onChange`? Couldn't the individual checkbox values be collected in `onSubmit` instead just before it gets sent to the fetch call rather than on every change?

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+                      />
+                      <TextField
+                        inputRef={register}
+                        id="location"
+                        margin="normal"
+                        name="location"
+                        label={LOCATION_LABEL}
+                        helperText={LOCATION_HELPER}
+                        fullWidth
+                      />
```
darrenvong:
```suggestion
                      <TextField
                        inputRef={register}
                        id="location"
                        margin="normal"
                        name="location"
                        label={LOCATION_LABEL}
                        helperText={LOCATION_HELPER}
                        fullWidth
                      />
```
Set the `defaultValue` prop on this (got a react-hook-form warning on this) with the result of `getContributorFormInfo`, should be accessible if using react query to do this once loaded.

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+        <CircularProgress />
+      ) : (
+        <>
+          {filled ? (
+            <>
+              <p>{ALREADY_FILLED_IN}</p>
+              <Button onClick={() => setFilled(false)}>{FILL_IN_AGAIN}</Button>
+            </>
+          ) : (
+            <>
```
darrenvong, line 194: Don't need this fragment either - single ternary again

darrenvong (changes requested):
Hopefully I caught most things

app/frontend/src/features/contributorForm/ContributorForm.tsx
aapeliv: I have to do the filtering/concat on every change anyway, so I thought it'd be easier to just split/join here as well

app/frontend/src/features/contributorForm/ContributorForm.tsx
aapeliv: I tried for quite a while to get this to work, but then couldn't figure out how to get these into the form without `setValue`, though I've moved stuff around

app/frontend/src/features/contributorForm/ContributorForm.tsx
aapeliv: Thanks, it is quite a lot easier!

aapeliv:
@darrenvong: I rebased on `develop`, fixed up the easy comments, and then refactored to use react query.

I was still having trouble setting the results from the first query (contributor form info endpoint) into the form, so I've still got it as hidden when logged in and using `setValue` in `onSuccess` in the react query query

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,367 @@
...
+        headers: {
+          Accept: "application/json",
+          "Content-Type": "application/json;charset=UTF-8",
+        },
+        mode: "cors",
+        method: "POST",
+        body: JSON.stringify(data),
+      }
+    );
+    return (await response.json()) as FormResponse;
```
darrenvong:
```suggestion
    return response.json();
```
Unnecessary `await` if you are already returning a Promise - the caller (React Query) will await it correctly.

app/frontend/src/features/contributorForm/ContributorForm.tsx
darrenvong: Hmm yeah even doing it on submit is gonna be hacky and this is likely a one time thing people use, so happy to leave this as it is as this is still more readable than what I initially had in mind

darrenvong (changes requested):
Very minor thing left, but this is a really great refactor overall 🎉

darrenvong (approved):
Assume build succeeds as before

PR #1029, opened 2021-04-08
Fix bug where user can't erase full address

### LINKED ISSUE #1028: Bug in location selector

aapeliv:
If you'd try to erase the whole address, it wouldn't work because `updates.address` was falsey. This fixes it and seems to be a more reliable way in general, so also updated the others.

Closes #1028

### DIFF
--- a/app/frontend/src/components/EditLocationMap.tsx
+++ b/app/frontend/src/components/EditLocationMap.tsx
@@ -185,12 +185,7 @@ export default function EditLocationMap({
         updateLocation(null);
       } else {
         setError("");
-        updateLocation({
-          address: location.current.address,
-          lat: location.current.lat,
-          lng: location.current.lng,
-          radius: location.current.radius,
-        });
+        updateLocation({ ...location.current });
       }
     }
 
--- a/app/frontend/src/components/EditLocationMap.tsx
+++ b/app/frontend/src/components/EditLocationMap.tsx
@@ -158,15 +158,15 @@ export default function EditLocationMap({
     updates: Partial<ApproximateLocation>,
     update: boolean = true
   ) => {
-    if (updates.address) {
-      location.current.address = updates.address;
+    if ("address" in updates) {
+      location.current.address = updates.address!;
     }
-    if (updates.radius && !exact) {
-      location.current.radius = updates.radius;
+    if ("radius" in updates && !exact) {
+      location.current.radius = updates.radius!;
     }
-    if (updates.lat && updates.lng) {
-      location.current.lat = updates.lat;
-      location.current.lng = updates.lng;
+    if ("lat" in updates && "lng" in updates) {
+      location.current.lat = updates.lat!;
+      location.current.lng = updates.lng!;
       isBlank.current = false;
     }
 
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/EditLocationMap.tsx
```diff
@@ -158,15 +158,15 @@ export default function EditLocationMap({
     updates: Partial<ApproximateLocation>,
     update: boolean = true
   ) => {
-    if (updates.address) {
-      location.current.address = updates.address;
+    if ("address" in updates) {
```
darrenvong: Not sure I fully understand the rationale of this change... by the putting the `!` after these variables, you're basically lying to TS to say a value always exist when it might not. If we want to always set it (even if it's undefined?), why not remove the checks altogether?

aapeliv: `updates` is a subset of the keys in `ApproximateLocation`, so you can call `commit({radius: 5})` and it will only update the radius.

These ifs check which parts are updated and update only those, with a bit of extra logic sprinkled in (like can't update the radius from 0 if the location is exact, etc).

I guess there could be the possibility that someone called `commit({radius: undefined})` or similar, in which case that might cause a bug?

aapeliv: I suppose it might make more sense if this was something like `location.current = {...location.current, ...updates}`?

But there is some other logic here too: e.g. lag+lng should be updated together to avoid bugs, and if they're updated we need to set `isBlank` to false; radius shouldn't be updatable, etc.

darrenvong: > I guess there could be the possibility that someone called commit({radius: undefined}) or similar, in which case that might cause a bug?

That'd be my guess, although not too familiar with Mapbox deals with this

> I suppose it might make more sense if this was something like `location.current = {...location.current, ...updates}`?

Yeah that would work if you don't need additional checks, as things that are implicitly undefined (i.e. the property doesn't exist on the object at all) shouldn't overwrite `location.current` on the previous spread.

lucaslcode:
```suggestion
    if (updates.address !== undefined) {
```
This should work, and remove the !

aapeliv: I'm happy for this to be refined and cleaned up further, but could we defer it to another PR?

I'd like to get this one merged as it's causing a bad bug in the signup flow :)

app/frontend/src/components/EditLocationMap.tsx
```diff
@@ -158,15 +158,15 @@ export default function EditLocationMap({
...
-      location.current.radius = updates.radius;
+    if ("radius" in updates && !exact) {
+      location.current.radius = updates.radius!;
     }
-    if (updates.lat && updates.lng) {
-      location.current.lat = updates.lat;
-      location.current.lng = updates.lng;
+    if ("lat" in updates && "lng" in updates) {
+      location.current.lat = updates.lat!;
+      location.current.lng = updates.lng!;
```
lucaslcode: These you can also do !== undefined (assuming 0 is a valid value)

lucaslcode (changes requested):

lucaslcode (approved):

PR #1030, opened 2021-04-08
Return copy of approx location object instead of some reacty reference?

aapeliv:
Before:

https://user-images.githubusercontent.com/2975660/114080492-7701bf00-9879-11eb-9741-2d2acd36d356.mov

After:

https://user-images.githubusercontent.com/2975660/114080605-939df700-9879-11eb-9edc-e5d5f5d51608.mov

### DIFF
--- a/app/frontend/src/components/EditLocationMap.tsx
+++ b/app/frontend/src/components/EditLocationMap.tsx
@@ -185,12 +185,7 @@ export default function EditLocationMap({
         updateLocation(null);
       } else {
         setError("");
-        updateLocation({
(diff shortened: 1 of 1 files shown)


aapeliv:
If the parent uses the location for something, then it would be possible to end up with an invalid location (e.g. address = "") through this, if the parent stores merely a copy to the reference and doesn't handle the `null` return case correctly.

On the other hand, basically everywhere where we use this, the parent just calls `onChange("address", ...)` etc on all the components individually

PR #1036, opened 2021-04-09
Fix broken edit hosting preference test

darrenvong:
This is a bit weird, since if you run the test in interactive mode, it'd say it's failed temporarily and then eventually passes 🤔 
but looks like the main thing that broke it was using `window.scroll` on `EditHostingPreference` since jsdom doesn't support it.

### DIFF
--- a/app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
+++ b/app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
@@ -44,11 +44,11 @@ describe("EditHostingPreference", () => {
     updateHostingPreferenceMock.mockResolvedValue(new Empty());
   });
 
-  it("should redirect to Profile route after successful update", async () => {
+  it("should redirect to the user profile route after successful update", async () => {
     renderPage();
 
     userEvent.click(await screen.findByRole("button", { name: SAVE }));
 
     expect(await screen.findByTestId("user-profile")).toBeInTheDocument();
-  });
+  }, 20000);
 });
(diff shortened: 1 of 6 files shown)


lucaslcode:
It passes for me when I comment out the markdown editors

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -1,42 +1,54 @@
...
-    );
+  it("should redirect to Profile route after successful update", async () => {
+    renderPage();
 
     userEvent.click(await screen.findByRole("button", { name: SAVE }));
 
-    expect(await screen.findByTitle(mockProfilePage)).toBeInTheDocument();
-  });
+    expect(await screen.findByTestId("user-profile")).toBeInTheDocument();
+  }, 20000);
```
darrenvong, line 53: So it looks like it's not to do with the markdown editor. I wonder what makes this take that much longer to run than other tests that this needs an increased timeout value...

lucaslcode: Maybe it's just a heavier page, frontend tests take quite a long time in CI

lucaslcode (approved):

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
darrenvong: Yeah the markdown editor must be more heavy to render...

PR #1037, opened 2021-04-09
Add about and forum into nav bar

### LINKED ISSUE #939: Add website and forum to navbar
Before logout on the right hand side we want About (https://couchers.org) and Forum (https://community.couchers.org)

jiwon1219:
Added About (https://couchers.org) and Forum (https://community.couchers.org) to navigation bar.

Screen Shot 2021-04-08 at 10 23 47 PM

Resolves #939

### DIFF
--- a/app/frontend/src/components/Navigation/Navigation.tsx
+++ b/app/frontend/src/components/Navigation/Navigation.tsx
@@ -24,7 +24,7 @@ import {
   userRoute,
 } from "routes";
 
-import { COUCHERS, LOG_OUT } from "../../constants";
+import { COUCHERS, LOG_OUT, ABOUT, FORUM } from "../../constants";
 import NavButton from "./NavButton";
 
 const menu = [
@@ -212,6 +212,8 @@ export default function Navigation() {
         <SearchBox />
         <div className={classes.bug}>
           <Hidden smDown>
+            <NavButton route={"https://couchers.org"} label={ABOUT} />
+            <NavButton route={"https://community.couchers.org"} label={FORUM} />
             <NavButton route={logoutRoute} label={LOG_OUT} />
           </Hidden>
           <BugReport />
--- a/app/frontend/src/components/Navigation/Navigation.tsx
+++ b/app/frontend/src/components/Navigation/Navigation.tsx
@@ -17,14 +17,16 @@ import SearchBox from "features/search/SearchBox";
 import React from "react";
 import CouchersLogo from "resources/CouchersLogo";
 import {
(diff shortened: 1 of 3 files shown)


app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -212,6 +212,8 @@ export default function Navigation() {
         <SearchBox />
         <div className={classes.bug}>
           <Hidden smDown>
+            <NavButton route={"https://couchers.org"} label={ABOUT} />
+            <NavButton route={"https://community.couchers.org"} label={FORUM} />
```
javiwq: Instead of passing the route like this, for best practice you should define the route in `routes.ts` and import it at the top with the other routes, then pass it as a variable.

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -212,6 +212,8 @@ export default function Navigation() {
         <SearchBox />
         <div className={classes.bug}>
           <Hidden smDown>
```
javiwq, line 222: This adds the links in the Desktop version, now to add it to the Mobile version just add the new items to `drawerItems`, as per the logout route.

javiwq (changes requested):
Hi, thank you for your first contribution! This looks good, there are just a couple things that I noticed 🙂

app/frontend/src/components/Navigation/Navigation.tsx
jiwon1219: This makes sense, thank you for the advice! I made the change in the latest commit.

app/frontend/src/components/Navigation/Navigation.tsx
jiwon1219: Added the links for the mobile version!

Screen Shot 2021-04-09 at 10 59 53 AM

javiwq (approved):
Looks good! Thanks again 🙂

javiwq:
Oops, I had approved this but now I realize the links are broken, they lead to `/url `rather than the url itself :/ This seems to be because our `NavButton` component is a styled `NavLink` component from `react-router-dom` which links only to local routes. Sorry for missing this in my review, I didn't catch it until I tested the preview. Reopening issue #939 and submitted PR to revert the changes #1054. 

@jiwon1219 perhaps you can open up a new PR?

jiwon1219:
@ineswq 
Sorry for the inconvenience :( I've updated the fix and opened a new PR #1066.

PR #1038, opened 2021-04-09
Minimum User Age

### LINKED ISSUE #831: Minimum user age
When the user signs up, there aren't any checks on the date of birth they provide apart from it's in a valid date format that the backend can accept. This means it's possible to create a user that's just been born (since the date field defaults to today's date):

A 0-years-old profile!

I think it'd be good to add an additional check on this on the backend so we ensure the user signing up is at least 18?

ceelo777:
Fixes #831 
- Add minimum age requirement of 18 years for users
- Validate using birthdate
- Wrote test that passed for signups (checks whether birth date is valid)


app/backend/src/couchers/utils.py
```diff
@@ -44,6 +44,12 @@ def to_aware_datetime(ts: Timestamp):
 def now():
     return datetime.now(utc)
 
+def required_birthdate():
+    """
+    Most recent birthdate allowed to register (must be 18 years minimum)
+    """
+    dt = now().date()
```
aapeliv: use `today_ = today()` (underscore to not overwrite the `today` object)

`dt` normally stands for datetime, but you only have a date

app/backend/src/couchers/utils.py
```diff
@@ -44,6 +44,12 @@ def to_aware_datetime(ts: Timestamp):
 def now():
     return datetime.now(utc)
 
+def required_birthdate():
```
aapeliv: Maybe call this `minimum_allowed_birthdate` just to be super explicit

app/backend/src/tests/test_auth.py
```diff
@@ -269,9 +269,25 @@ def test_signup_invalid_birthdate(db):
                 radius=100,
                 accept_tos=True,
             )
-        )
+        )        
     assert e.value.code() == grpc.StatusCode.INVALID_ARGUMENT
     assert e.value.details() == errors.INVALID_BIRTHDATE
 
+    with auth_api_session() as (auth_api, metadata_interceptor):
```
aapeliv: you can use the same session as above: you'll need to separate the `pytest.raises` into it's own `with` (just break it at the comma and create a new with block to put the above in, or have a look at how `pytest.raises` is used elsewhere

app/backend/src/tests/test_auth.py
```diff
@@ -269,9 +269,25 @@ def test_signup_invalid_birthdate(db):
...
+                city="New York City",
+                birthdate="2000-12-31", # arbitrary birthdate older than 18 years
+                gender="Helicopter",
+                hosting_status=api_pb2.HOSTING_STATUS_CAN_HOST,
+                lat=1,
+                lng=1,
+                radius=100,
+                accept_tos=True
+            )
+        )
```
aapeliv, line 291: please also add a test where the birthdate is like 17 years, not in the future but not yet 18

app/backend/src/tests/test_auth.py
```diff
@@ -269,9 +269,25 @@ def test_signup_invalid_birthdate(db):
                 radius=100,
                 accept_tos=True,
             )
-        )
+        )
```
aapeliv: please autoformat the code by running `isort . && black .` in `app`

app/backend/src/couchers/servicers/auth.py
```diff
@@ -204,7 +204,7 @@ def CompleteSignup(self, request, context):
 
             birthdate = parse_date(request.birthdate)
 
-            if not birthdate or birthdate >= today():
+            if not birthdate or birthdate >= required_birthdate():
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_BIRTHDATE)
```
aapeliv, line 216: please find the `INVALID_BIRTHDATE` string in errors and change it to something like `You must be at least 18 years old to sign up.`

aapeliv (changes requested):
Looking good, some small nitpicks!

aapeliv (changes requested):
Ready to merge, but could you run the formatting commands one more time, there's a few formatting errors left!

ceelo777:
> Ready to merge, but could you run the formatting commands one more time, there's a few formatting errors left!

Oops, forgot to format, my bad!

aapeliv (approved):
LGTM!

PR #1039, opened 2021-04-09
Workaround for login/signup styles different in production

lucaslcode:
Unfortunately this requires manually adding `{index: 1}` to affected styles, so I guess we'll just have to keep an eye out...

### DIFF
--- a/app/frontend/src/components/Divider.tsx
+++ b/app/frontend/src/components/Divider.tsx
@@ -1,4 +1,4 @@
-import { Divider as MuiDivider,  } from "@material-ui/core";
+import { Divider as MuiDivider } from "@material-ui/core";
 import React from "react";
 import makeStyles from "utils/makeStyles";
 
--- a/app/frontend/src/components/Divider.tsx
+++ b/app/frontend/src/components/Divider.tsx
@@ -1,5 +1,6 @@
-import { Divider as MuiDivider, makeStyles } from "@material-ui/core";
+import { Divider as MuiDivider,  } from "@material-ui/core";
 import React from "react";
+import makeStyles from "utils/makeStyles";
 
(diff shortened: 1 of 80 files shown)


aapeliv:
Glad you found a workaround!


Just an idea looking at this: are you able to patch this everywhere by changing imports and wirting a shim around the original `makeStyles` like this:

```ts
makeStyles = (originalMakeStyles: (theme) => styleType) => (theme => ({originalMakeStyles(theme), {index: 1}))
```

This probably doesn't run, but maybe you get what I mean? Might be less effort and fix it everywhere?

lucaslcode:
Thank god for regex

darrenvong:
Does this affect the ordering in which the the styles are injected only? Also where is this documented? Did you have to dig into their source code for this?

app/frontend/src/components/Divider.tsx
```diff
@@ -1,5 +1,6 @@
-import { Divider as MuiDivider, makeStyles } from "@material-ui/core";
+import { Divider as MuiDivider,  } from "@material-ui/core";
```
darrenvong:
```suggestion
import { Divider as MuiDivider } from "@material-ui/core";
```

darrenvong (commented):
The formatting on a few files looks a bit off (I didn't bother commenting on all of them) - might want to run Prettier over everything?

lucaslcode:
https://cssinjs.org/jss-api?v=v10.6.0#create-style-sheet
Higher index means lower in `<head>`, and it defaults to 0

darrenvong (approved):
Seems like a perfectly fine solution to me than a workaround especially as we made it into a helper function! 🎉

PR #1040, opened 2021-04-09
Backend/feat/no password second email confirmation

### LINKED ISSUE #912: Old email needs to be confirmed when changing emails
If a user has no password, and they want to change their email address, currently they can just do that without any confirmation. In particular, if someone leaves their laptop unlocked with Couchers open, anybody can trivially hijack their account. We can't have that.

Either:

1. Require a user to have a password to change their email address
2. Require verification of both old and new email any time they want to change their email
3. Same as above but only if the user has no password

ShmuelTreiger:
One thing I'm not sure about is how the confirmation link works, specifically:

1) If I need to make a new url
2) I'm sure it needs some frontend change


ShmuelTreiger:
I had 'send_email_changed_confirmation_email' split into two methods, old and new. I combined them. I'm thinking I should probably split them again?

app/backend/src/couchers/db.py
```diff
@@ -177,9 +177,24 @@ def new_password_reset_token(session, user, hours=2):
     return password_reset_token, f"{hours} hours"
 
 
-def set_email_change_token(session, user, hours=2):
+def set_change_token_old_email(session, user, hours=2):
```
aapeliv: `set_email_change_token_for_{old/new}_email`

app/backend/src/couchers/servicers/account.py
```diff
@@ -118,11 +120,16 @@ def ChangeEmail(self, request, context):
         with session_scope() as session:
             user = session.query(User).filter(User.id == context.user_id).one()
 
-            # otherwise we're good
             user.new_email = request.new_email
-            token, expiry_text = set_email_change_token(session, user)
+            if user.hashed_password is None:
```
aapeliv: We should probably write a `@property` helper for `user.has_password`, it's a bit ugly to use this everywhere

aapeliv (dismissed):
I think rather than a two-step process, we should send out both emails simultaneously and perform the action when both are done.

app/backend/src/couchers/db.py
ShmuelTreiger: If we make it one process, then yes, this can be combined. Otherwise it makes more sense to keep separate.

app/backend/src/couchers/servicers/account.py
ShmuelTreiger: I can add that

ShmuelTreiger:
I disagree about a one step process. It gives the potential hacker more information. If I find a logged-in couchers account, there's currently no way to get the current email address (I believe) and IMO that's a good thing. Changing it to one step would expose the email address to the potential hacker.

aapeliv:
> no way to get the current email address (I believe) and IMO that's a good thing.

It has just been added.

If you find a logged in user there's already a tonne of exposed information. It's a problem of course but the only workaround is to not leave your account attended... Eventually we'll have some tool to log out of all other devices or something

ShmuelTreiger:
Disagree, but it's now simultaneous

app/backend/src/couchers/db.py
```diff
@@ -177,19 +177,26 @@ def new_password_reset_token(session, user, hours=2):
...
+    Returns both tokens and expiry text
     """
-    token = urlsafe_secure_token()
-    user.new_email_token = token
+    old_email_token = ""
+    if double_confirmation:
+        old_email_token = urlsafe_secure_token()
+        user.old_email_token = old_email_token
+        user.old_email_token_created = now()
+        user.old_email_token_expiry = now() + timedelta(hours=hours)
```
aapeliv, line 131: These need to be reset if there's no old email

app/backend/src/couchers/models.py
```diff
@@ -178,13 +178,20 @@ class User(Base):
     accepted_tos = Column(Integer, nullable=False, default=0)
 
     # for changing their email
+    old_email_token = Column(String, nullable=True)
+    old_email_token_created = Column(DateTime(timezone=True), nullable=True)
+    old_email_token_expiry = Column(DateTime(timezone=True), nullable=True)
     new_email = Column(String, nullable=True)
```
aapeliv: Move this line to above the others, please, the others are just for tokens

app/backend/src/couchers/servicers/account.py
```diff
@@ -117,12 +119,19 @@ def ChangeEmail(self, request, context):
...
 
-            send_email_changed_notification_email(user)
-            send_email_changed_confirmation_email(user, token, expiry_text)
+            if user.has_password:
+                old_email_token, new_email_token, expiry_text = set_email_change_tokens(
+                    session, user, double_confirmation=False
+                )
+                send_email_changed_notification_email(user)
+            else:
+                old_email_token, new_email_token, expiry_text = set_email_change_tokens(
```
aapeliv, line 140: Please add a comment here about how this works.

It's also probably cleaner to move the `send_email_changed_confirmation_email(...)` inside the branches instead of leaving it out, just makes it clearer how the logic works, even though it's an extra line

app/backend/src/couchers/servicers/auth.py
```diff
@@ -413,12 +418,31 @@ def CompletePasswordReset(self, request, context):
...
+    def ConfirmChangeEmailWithOldAddress(self, request, context):
+        with session_scope() as session:
+            user = (
+                session.query(User)
+                .filter(User.old_email_token == request.change_email_token)
+                .filter(User.old_email_token_created <= func.now())
+                .filter(User.old_email_token_expiry >= func.now())
+                .one_or_none()
+            )
+            if user:
```
aapeliv: abort with a shortcut (`if not user: ...abort`) instead of doing the main logic in the `if` branch

app/backend/src/couchers/servicers/auth.py
```diff
@@ -428,11 +452,14 @@ def CompleteChangeEmail(self, request, context):
                 .one_or_none()
             )
             if user:
-                user.email = user.new_email
-                user.new_email = None
                 user.new_email_token = None
                 user.new_email_token_created = None
                 user.new_email_token_expiry = None
+
+                if user.old_email_token == None:
```
aapeliv: `if not user.old_email_token` don't compare to `None` (except in SA queries)

app/backend/src/couchers/tasks.py
```diff
@@ -139,16 +139,24 @@ def send_email_changed_notification_email(user):
     email.enqueue_email_from_template(user.email, "email_changed_notification", template_args={"user": user})
 
 
-def send_email_changed_confirmation_email(user, token, expiry_text):
+def send_email_changed_confirmation_email(user, token, expiry_text, old):
```
aapeliv: Let's have one of these `send_*_email` functions per template, so divide this into two functions

app/pb/auth.proto
```diff
@@ -73,8 +73,12 @@ service Auth {
     // Triggered when the user goes to the link sent in the forgot password email
   }
 
-  rpc CompleteChangeEmail(CompleteChangeEmailReq) returns (google.protobuf.Empty) {
-    // When a user opens a changed email confirmation, the frontend should call this function with the token
+  rpc ConfirmChangeEmailWithOldAddress(ConfirmChangeEmailWithOldAddressReq) returns (google.protobuf.Empty){
```
aapeliv: `Confirm{New,Old}EmailAddress`?

aapeliv (changes requested):
Thanks for changing that. See some small comments in code. A few bigger comments:

1. I think you should add `{new,old}_email_confirmed` booleans in the models. This is just safer than trying to check if the tokens are `None` or not (because they can be `None` for many reasons). Safer in programmer error sense, which is vital for security-critical components. If not, then you must document very carefully in the user model that being null has important extra meaning.
2. Clear all of the email change info in the user when someone requests a change: this way there won't be any leftover stuff in there which might cause bad security bugs
3. Should we have just one `ConfirmEmailAddress` RPC call? From the POV of the frontend: they do the same thing, and it makes life easier in the frontend. This is currently using the same URL template for both anyway?

app/backend/src/couchers/servicers/account.py
ShmuelTreiger: There's a comment section under the method declaration (which needed some updating). Did you mean something different? Or should I move that comment down here?

app/pb/auth.proto
ShmuelTreiger: ConfirmOldEmailAddress sounds like you're confirming your old email address is correct for some reason. Ambiguous

ShmuelTreiger:
>Should we have just one ConfirmEmailAddress RPC call? From the POV of the frontend: they do the same thing, and it makes life easier in the frontend. This is currently using the same URL template for both anyway?

It depends if the front end is going to change or if we're just going to use one link. If two pages with two links, I'd differentiate. It's currently using the same URL. Will it continue to do so? I posed that question when I started this branch.

ShmuelTreiger:
Migration issue. I'll fix next week

ShmuelTreiger:
Backend tests pass. Once no changes, ready for frontend work.

app/pb/auth.proto
aapeliv: Maybe something like this?

```proto
message CompleteChangeEmailRes {
  // true if all emails confirmed and email changed, false otherwise
  bool success = 1;
  // one of these is true if success is false
  bool requires_confirmation_from_new_email = 2;
  bool requires_confirmation_from_old_email = 3;
}
```

app/backend/src/couchers/tasks.py
```diff
@@ -131,24 +131,43 @@ def send_password_changed_email(user):
...
     """
-    Send the user an email confirming their new email. Goes to the new address
+    Send an email to user's original email address requesting confirmation of email change
     """
     logger.info(
-        f"Sending email changed (confirmation) email to {user=} (old email: {user.email=}, new email: {user.new_email=})"
+        f"Sending email changed (confirmation) email to {user=}'s old email address, (old email: {user.email}, new email: {user.new_email=})"
     )
+
+    confirmation_link = urls.change_email_link(confirmation_token=token)
```
aapeliv, line 204: You use the same `change_email_link` for both new and old emails, and frontend won't know which of the two functions to call. I suggest you combine them into one and return as in other comment

ShmuelTreiger:
@aapeliv I don't think this is what you had in mind, but I couldn't figure out how else to do it. Happy to discuss if you have some time today. Sorry, I know it's been a while so it won't be fresh on your mind.

ShmuelTreiger:
@aapeliv have a look?

app/backend/src/couchers/servicers/account.py
```diff
@@ -107,34 +108,42 @@ def ChangeEmail(self, request, context):
...
+
+            # check password first
             _check_password(user, "password", request, context)
 
-        # not a valid email
-        if not is_valid_email(request.new_email):
-            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_EMAIL)
+            # not a valid email
+            if not is_valid_email(request.new_email):
+                context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_EMAIL)
```
aapeliv, line 125: Thanks for fixing up the multi-session mess here!

app/backend/src/couchers/servicers/account.py
```diff
@@ -31,7 +32,7 @@ def _check_password(user, field_name, request, context):
             # wrong password
             context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_USERNAME_OR_PASSWORD)
 
-    if request.HasField(field_name) and not user.hashed_password:
+    elif request.HasField(field_name):
```
aapeliv: Leave this as `elif not user.has_password and request.HasField(field_name)`; that's more explicit and much clearer what's going on

app/backend/src/couchers/servicers/auth.py
```diff
@@ -424,27 +424,48 @@ def CompletePasswordReset(self, request, context):
...
-                return empty_pb2.Empty()
+                user.confirmed_email_change_via_new_email = True
             else:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
+
+            if user.confirmed_email_change_via_old_email and user.confirmed_email_change_via_new_email:
+                user.email = user.new_email
+                user.new_email = None
+                user.confirmed_email_change_via_old_email = False
+                user.confirmed_email_change_via_new_email = False
```
aapeliv: These should be set to `None`?

app/backend/src/couchers/servicers/auth.py
```diff
@@ -424,27 +424,48 @@ def CompletePasswordReset(self, request, context):
...
+                res = auth_pb2.ConfirmChangeEmailRes(success=True)
+            elif user.confirmed_email_change_via_old_email:
+                res = auth_pb2.ConfirmChangeEmailRes(requires_confirmation_from_new_email=True)
+            else:
+                res = auth_pb2.ConfirmChangeEmailRes(requires_confirmation_from_old_email=True)
+
+            session.commit()
```
aapeliv: No need for commit here (when execution leaves the ctx manager; it'll automatically commit, and if it fails, it'll error out regardless)

app/backend/src/couchers/servicers/auth.py
```diff
@@ -424,27 +424,48 @@ def CompletePasswordReset(self, request, context):
...
+            if user.confirmed_email_change_via_old_email and user.confirmed_email_change_via_new_email:
+                user.email = user.new_email
+                user.new_email = None
+                user.confirmed_email_change_via_old_email = False
+                user.confirmed_email_change_via_new_email = False
+                res = auth_pb2.ConfirmChangeEmailRes(success=True)
+            elif user.confirmed_email_change_via_old_email:
+                res = auth_pb2.ConfirmChangeEmailRes(requires_confirmation_from_new_email=True)
+            else:
+                res = auth_pb2.ConfirmChangeEmailRes(requires_confirmation_from_old_email=True)
```
aapeliv: You can just return instead of `res = `...

app/backend/src/couchers/db.py
```diff
@@ -118,19 +118,32 @@ def new_password_reset_token(session, user, hours=2):
     return password_reset_token, f"{hours} hours"
 
 
-def set_email_change_token(session, user, hours=2):
+def set_email_change_tokens(session, user, double_confirmation, hours=2):
```
aapeliv: I think `confirm_old_email` is a tad more descriptive than `double_confirmation`?

app/backend/src/couchers/db.py
```diff
@@ -118,19 +118,32 @@ def new_password_reset_token(session, user, hours=2):
...
+    else:
+        user.old_email_token = None
+        user.old_email_token_created = None
+        user.old_email_token_expiry = None
+        user.confirmed_email_change_via_old_email = True
+
+    new_email_token = urlsafe_secure_token()
+    user.new_email_token = new_email_token
     user.new_email_token_created = now()
     user.new_email_token_expiry = now() + timedelta(hours=hours)
```
aapeliv, line 143: Need to also set `confirmed_email_change_via_old_email` to something

app/backend/src/couchers/models.py
```diff
@@ -179,9 +179,14 @@ class User(Base):
...
     # for changing their email
     new_email = Column(String, nullable=True)
+    old_email_token = Column(String, nullable=True)
+    old_email_token_created = Column(DateTime(timezone=True), nullable=True)
+    old_email_token_expiry = Column(DateTime(timezone=True), nullable=True)
+    confirmed_email_change_via_old_email = Column(Boolean, nullable=True, default=False)
     new_email_token = Column(String, nullable=True)
     new_email_token_created = Column(DateTime(timezone=True), nullable=True)
     new_email_token_expiry = Column(DateTime(timezone=True), nullable=True)
+    confirmed_email_change_via_new_email = Column(Boolean, nullable=True, default=False)
```
aapeliv: I think given your usage here, `confirmed_email_change_via_{new,old}_email` would be better named in the opposite, `need_to_confirm_{new,old}_email`.

Also; there's a lot of fields here that need to work in tandem: could you add a constraint to make sure that `old_email_*,confirmed_email_change_via_old_email` are all either or all NULL, and similar for `new_email_*,...`?

aapeliv (changes requested):
Sorry for taking a while to get back to this! I had a busy week on research.

This is looking really good; see the comments on a few improvements. But the architecture is spot on.

app/backend/src/couchers/servicers/account.py
ShmuelTreiger: What if I add a comment, instead of testing the same thing twice.

ShmuelTreiger: I feel like the comments here are sufficient. They kind of walk you through the possibilities.

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: Shouldn't really matter, but sure

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: Can you explain this to me in more detail sometime? I still can't figure out when to commit and when not

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: I used the res because I thought I needed to commit and just wanted to do it once

app/backend/src/couchers/servicers/account.py
aapeliv: OK, fair enough.

app/backend/src/couchers/db.py
ShmuelTreiger: It's set in the if else statement

ShmuelTreiger:
No worries at all. We all have things to do.

Is there even a reason to keep track of when tokens were created and checking that they were created in the past? Could we just keep track of the expiration? Would simplify things

app/backend/src/couchers/models.py
ShmuelTreiger: I can add constraints that old_email_* are all NULL or all not NULL (I think that's what you want?), but as it's currently set up, confirmed_email_change_via_old_email can be True while they are NULL (i.e. user has confirmed via old email but not yet through new email). I could change it so that even once it's confirmed the fields don't go to null until the whole process is complete. I suppose it doesn't really matter? Worst case scenario they just confirm the email change again using the same email/token and nothing really changes?

ShmuelTreiger:
I'm not a huge fan of the last commit (renaming the booleans). I think the end result is more confusing. Ultimately, doesn't really matter though.

ShmuelTreiger:
This seems dangerous to me: The default state is now the passing state for changing your email, protected by one context.abort.

I think we should:
1) Only check constraint the token(, its creation,) and its expiration
2) Revert the booleans so we can check not bool (so that bool = None fails)
3) Remove the token creation time altogether (just for simplicity)

ShmuelTreiger:
Migrations are failing and I can't figure out why. Other than that, seems in working order. Tests pass.

aapeliv:
> Migrations are failing and I can't figure out why. Other than that, seems in working order. Tests pass.

It seems to be missing the `phone_verified_conditions` check constraint (possibly related to there being two `__table_args__`s??) Checked the PR and it doesn't seem it includes that in the migration. `test_migrations` should've picked that up? I think it's fine if you just add that constraint in your migration.

@kalvdans: good for you to know this happened, maybe you have some ideas?

kalvdans:
> > Migrations are failing and I can't figure out why. Other than that, seems in working order. Tests pass.
> 
> It seems to be missing the `phone_verified_conditions` check constraint (possibly related to there being two `__table_args__`s??) Checked the PR and it doesn't seem it includes that in the migration. `test_migrations` should've picked that up? I think it's fine if you just add that constraint in your migration.
> 
> @kalvdans: good for you to know this happened, maybe you have some ideas?

Yes, it seems to be a merge accident in c65df71dd4ac2c9baa6e6230ccfb2f6a859efe3a, just before I create the migrations for the sms verification MR https://github.com/Couchers-org/couchers/pull/1250

ShmuelTreiger:
Yup. For some reason I was convinced test_migration was passing at some point. I should have figured this one out.

app/backend/src/couchers/db.py
```diff
@@ -118,19 +118,39 @@ def new_password_reset_token(session, user, hours=2):
...
-    return token, f"{hours} hours"
+    user.need_to_confirm_via_new_email = True
+
+    # Check for edge case, however remote
+    if old_email_token == new_email_token:
+        return set_email_change_tokens(user, confirm_with_both_emails)
```
aapeliv: Why would `old_email_token == new_email_token`?

app/backend/src/couchers/db.py
```diff
@@ -118,19 +118,39 @@ def new_password_reset_token(session, user, hours=2):
...
+    Make email change tokens which are valid for `hours` hours
 
     Note: does not call session.commit()
 
-    Returns token and expiry text
+    Returns two tokens and expiry text
     """
-    token = urlsafe_secure_token()
-    user.new_email_token = token
+    old_email_token = ""
```
aapeliv: You can move this into the `else`; in python, `if` statements don't have their own scope

app/backend/src/couchers/migrations/versions/5e89dd9ef181_adding_tokens_and_booleans_to_user_model.py
```diff
@@ -0,0 +1,55 @@
...
+         (need_to_confirm_via_old_email IS FALSE AND old_email_token IS NULL AND old_email_token_created IS NULL AND old_email_token_expiry IS NULL)",
+    )
+    op.create_check_constraint(
+        constraint_name="check_new_email_token_state",
+        table_name="users",
+        condition="(need_to_confirm_via_new_email IS NULL AND new_email_token IS NULL AND new_email_token_created IS NULL AND new_email_token_expiry IS NULL) OR \
+         (need_to_confirm_via_new_email IS TRUE AND new_email_token IS NOT NULL AND new_email_token_created IS NOT NULL AND new_email_token_expiry IS NOT NULL) OR \
+         (need_to_confirm_via_new_email IS FALSE AND new_email_token IS NULL AND new_email_token_created IS NULL AND new_email_token_expiry IS NULL)",
+    )
+
```
aapeliv, line 45: It might be good to set all the email verification columns to None here to make sure there's no inconsistent states.

aapeliv (changes requested):
I double checked the logic one final time: it seems correct.

See the couple of minor nitpicks.

Please also rebase onto/merge develop: you'll need to update the emails, I think. All variables should either be `couchers_escape` or `couchers_safe` (or whatever they're called).

Then I can fix up the frontend on this branch and get frontend to review that little bit?

app/backend/src/couchers/db.py
ShmuelTreiger: If the two URLs happen to be the same. Odds are close to zero, but if we ever scale up to millions of users, it's still close to zero but slightly larger.

app/backend/src/couchers/db.py
ShmuelTreiger: Didn't know that. Cool

app/backend/src/couchers/migrations/versions/5e89dd9ef181_adding_tokens_and_booleans_to_user_model.py
ShmuelTreiger: You mean default values?

ShmuelTreiger:
>Please also rebase onto/merge develop: you'll need to update the emails, I think. All variables should either be couchers_escape or couchers_safe (or whatever they're called).

Already merged that, but ya, I'll remerge once everything is done

app/backend/src/couchers/db.py
aapeliv: No need to test for that. It's practically impossible for that to happen (as a mathematician I'm not allowed to call it impossible, but it's basically impossible). Each token has 32 bytes (256 bits), of randomness. So the chance of collision is 1 in 2^512, which is way more than there are atoms in the universe!

app/backend/src/couchers/migrations/versions/5e89dd9ef181_adding_tokens_and_booleans_to_user_model.py
aapeliv: Yeah; basically to set all the old (before this PR) email change fields to `None`, so that if someone was in the process of changing their email it won't cause any illegal states. (They'll just have to start again, which is OK)

ShmuelTreiger: Once this merges, those tokens will essentially be invalid, since I changed the name, no? New field in the model

If they confirm via an old email/token, it should just throw an invalid token error. Slight user frustration, but hopefully they just retry the process

aapeliv: Not the `new_email_*` ones?

ShmuelTreiger:
@aapeliv added data migration based on what I found here:

https://stackoverflow.com/questions/24612395/how-do-i-execute-inserts-and-updates-in-an-alembic-upgrade-script/24623979

Not sure how to check if it's working properly

PR #1051, opened 2021-04-10
Refactor avatar input to image input

### LINKED ISSUE #928: Generic image upload component
Image upload component like the avatar upload except... square? For places and later, guides/pages/etc.

lucaslcode:
closes #928

### DIFF
--- a/app/frontend/src/components/ImageInput/ImageInput.tsx
+++ b/app/frontend/src/components/ImageInput/ImageInput.tsx
@@ -80,6 +80,8 @@ type SquareImgInputProps = {
   type: "rect";
   alt: string;
   grow?: boolean;
+  height?: number;
+  width?: number;
 };
 
 export function ImageInput(
@@ -183,8 +185,8 @@ export function ImageInput(
               src={imageUrl ?? imagePlaceholder}
               style={{ objectFit: !imageUrl ? "contain" : undefined }}
               alt={props.alt}
-              width={DEFAULT_WIDTH}
-              height={DEFAULT_HEIGHT}
+              width={props.width ?? DEFAULT_WIDTH}
+              height={props.height ?? DEFAULT_HEIGHT}
             />
           )}
           {mutation.isLoading && (
--- a/app/frontend/src/components/ImageInput/ImageInput.tsx
+++ b/app/frontend/src/components/ImageInput/ImageInput.tsx
@@ -20,6 +20,9 @@ import { useMutation } from "react-query";
 import { service } from "service";
 import { ImageInputValues } from "service/api";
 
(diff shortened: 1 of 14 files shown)


darrenvong:
Got a build error on `ImageInput` not having a default export

darrenvong:
I'll remember to look at this properly some time tomorrow hopefully...

app/frontend/src/components/ImageInput/ImageInput.tsx
```diff
@@ -49,23 +63,31 @@ const useStyles = makeStyles((theme) => ({
...
+type SquareImgInputProps = {
+  type: "rect";
+  alt: string;
+  grow?: boolean;
+  height?: number;
+  width?: number;
+};
+
+export function ImageInput(
+  props: (AvatarInputProps | SquareImgInputProps) & ImageInputProps
```
darrenvong: It would simplify things a bit here if we have all of these as `interfaces`, then have `AvatarInputProps` and `SquareImgInputProps` extend from `ImageInputProps`. That way, in the `ImageInput` props type, you don't need the `& ImageInputProps` bit, which I think makes it a little bit more intuitive to read

darrenvong (changes requested):
Pretty some minor comments but looks fine otherwise.

Is this being used anywhere in the app to test this for real? I think Storybook is being funny since it's not backed by a proper server to handle upload so it looks "broken"

app/frontend/src/components/ImageInput/ImageInput.tsx
lucaslcode: I did it that way originally, but I couldn't get it to infer the type once I branched on `type`

darrenvong: Still works fine for me according to VS Code when I changed everything to interface, considering the `type` field you got there is exactly what TS used to narrow the type in these situations.

type before the if ternary branch:
Screenshot 2021-04-21 at 01 56 35

`SquareImgInputProps` correctly narrowed in the else ternary branch:
Screenshot 2021-04-21 at 01 56 47

lucaslcode:
Used interfaces, and also added the component to the skeleton place and new place pages `/place/new` / `/place/:id`

darrenvong:
Build failed again on an interesting a11y lint error about not using the word "photo" or any synonym of that in the alt text of an image.

I wonder if it's best to give them an empty alt text instead, since for non-visual users, they aren't really adding extra information apart from "it looks nice", so probably class as a decorative image according to the [W3 guidance here](https://www.w3.org/WAI/tutorials/images/decorative/#image-used-for-ambiance-eye-candy)?

darrenvong (approved):

PR #1052, opened 2021-04-11
Reduce re-renders on edit location map

darrenvong:
Reduces the number of re-renders quite significantly, and scope it to only the components that **needs to be re-render a lot** to a smaller area (i.e. the radius slider). Previously, it was re-rendering on every character change when you update the display location and on every map move.

See before and after video for the difference (excuse the terrible typos :p) - the flashy squares is something from the React DevTools, shows quite nicely where things get re-rendered before/after this change.

**Before**

https://user-images.githubusercontent.com/13669362/114289153-8cd7d580-9a6d-11eb-82fc-cde8a794b7ea.mov

**After**

https://user-images.githubusercontent.com/13669362/114289157-95c8a700-9a6d-11eb-97ee-b359efe95013.mov

### DIFF
--- a/app/frontend/src/features/profile/edit/EditProfile.tsx
+++ b/app/frontend/src/features/profile/edit/EditProfile.tsx
@@ -206,6 +206,7 @@ export default function EditProfileForm() {
             />
           </form>
           <Controller
+            defaultValue=""
             name="location"
             control={control}
             render={() => (
--- a/app/frontend/src/components/EditLocationMap.tsx
+++ b/app/frontend/src/components/EditLocationMap.tsx
@@ -18,6 +18,7 @@ import {
   DISPLAY_LOCATION_NOT_EMPTY,
   getRadiusText,
   INVALID_COORDINATE,
+  LOCATION_ACCURACY,
   LOCATION_PUBLICLY_VISIBLE,
   MAP_IS_BLANK,
 } from "./constants";
@@ -73,8 +74,6 @@ export default function EditLocationMap({
 
   const map = useRef<mapboxgl.Map | null>(null);
 
-  const [, setDirty] = useState(0);
-
   // map is imperative so these don't need to cause re-render
   const location = useRef<ApproximateLocation>({
     address: initialLocation?.address ?? "",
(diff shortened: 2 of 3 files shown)


app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -206,6 +206,7 @@ export default function EditProfileForm() {
             />
           </form>
           <Controller
+            defaultValue=""
```
darrenvong, line 209: This is to stop the `react-hook-form` warning, even though we don't actually use the value of this field anyway...

lucaslcode (changes requested):
This no longer updates the location field after a search

darrenvong:
Oops didn't realise that was a functionality, good catch. Fixed now!

aapeliv:
I had the `update` boolean on the `commit` to reduce the number of times we bubble the location back to the parent in cases when the user is between an action. That's unrelated to whether a re-draw/setDirty should happen or not.

1. When you select a search result for a new location, this now sends two updates, one with (new address, old coords) and then a second one with (new address, new coords), e.g. with a loc in NY and searching for London, `updateLocation` is called twice, with this history:

```
2: {"address":"New York","radius":1000,"lat":40.716740506264216,"lng":-74.00396063906874}
3: {"address":"London, England, United Kingdom","radius":1000,"lat":40.716740506264216,"lng":-74.00396063906874}
4: {"address":"London, England, United Kingdom","radius":1000,"lat":51.50025931937692,"lng":-0.1343511113528848}
```

Notice 3 has the coordinates from 2.

2. When I drag now it lags and calls `updateLocation` every move event, I'm not sure if lag is because of the continuous `updateLocation` calls or otherwise (this might make the parent redraw the child?). Before these changes, dragging the circle was buttery smooth and would fire off one update at the end of the drag/move.

Before (it's not laggy/tearing at all when not recording):

https://user-images.githubusercontent.com/2975660/114307234-c0862000-9aac-11eb-9e0f-d3ed6a40f0c2.mov

After (laggy/tearing even when not recording):

https://user-images.githubusercontent.com/2975660/114307231-be23c600-9aac-11eb-9928-44aad51162cc.mov

darrenvong:
Thanks for checking it over! I think the lag in Storybook is exaggerated (as I don't see this in the main page itself) and is mostly down to how you're printing the history as well, which is re-rendering **and** committing (i.e. changing the DOM) on every event so it's hammering the DOM. The actual pages themselves that use the map don't have this crazy list so that'll likely explain why I wasn't seeing the lag when viewed on them.

Even though that's not a problem on the actual page, updating the ref state on every mousemove event is probably still an overkill even if it's not gonna cause re-renders, so I'll re-introduce the `update` flag you had before and that should fix the issue

darrenvong:
Should be addressed now 👍  also renamed that second param to `shouldUpdate` since seeing `updates` and `update` next to each other is a bit confusing...

lucaslcode (approved):

Issue #871, opened 2021-03-28
Return full image URLs in API calls


CouchersBot:
Subject: Uploaded profile picture resolution too-low
Description:
Profile picture is reduced to a significantly lower resolution, and the loss in quality is very evident at the scale displayed.

Steps:
Edit my profile picture.

Results:
My profile picture was updated with a very low resolution. It should have had a higher resolution.

Backend version: develop-8782c21f
Frontend version: develop-481b6d9f
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.90 Safari/537.36
Page: https://develop--frontend.preview.coucher.org/profile/natalia#loc=13/43.05859/-89.42751
User ID: 25

naspitha:
Here's the comparison between the original picture and the version displayed.
![image](https://user-images.githubusercontent.com/23509977/112742370-a5b39780-8f53-11eb-8b19-aa3db863b491.png)

lucaslcode:
Frontend only gets the thumbnail image, thumbnail size is set somewhere nebulous (there's a `thumbnail_size` param for the Flask media server), not sure what it is atm

aapeliv:
As discussed in the meeting, this should return a URL with the size replaced by `{size}` instead of `thumbnail`/`full`, so the frontend can pick.

Assigned to Ben but he's not on GitHub yet!

lucaslcode:
This should take care of images in emails too

aapeliv:
This is sorted out now I think

PR #1055, opened 2021-04-11
Return full image URL instead of thumbnail image

aapeliv:
This is a quick workaround to the blurry pictures in the UI.

The full fix is deferred to #871 and requires some frontend attention too.


kalvdans (approved):
lgtm!

PR #1063, opened 2021-04-11
Remove gender from update profile.

### LINKED ISSUE #918: Remove gender from UpdateProfile
To reduce abuse of gender-based features.

Users can still contact support if there is a legitimate need to change their gender.

ceelo777:
Fixes #918 

- Remove gender reference in api.proto
- Remove permission denied check (since no more gender in request) in api.py


aapeliv (changes requested):
Seems like you need to update the tests to remove the gender from them too:

https://gitlab.com/couchers/couchers/-/jobs/1170136962

aapeliv (approved):
LGTM

Issue #900, opened 2021-04-01
More testing


aapeliv:
We could do with a few more tests in the backend. There's some bits that are somewhat poorly tested. It'd be great to have 100% coverage on e.g. the auth code. (Password based login is not tested at all!)

ShmuelTreiger:
It would be good if you could specify anywhere else you see that's missing. You mentioned auth, but any others? Or should we just use this thread to collect as we find untested things?

lucaslcode:
https://develop--bcov.preview.couchershq.org

aapeliv:
I've compiled a list of stuff that needs added tests from the coverage report. In addition to these, there's a whole bunch of error checking branches that I don't think need testing that badly, but I mean, it wouldn't hurt if someone is bored :P

## High priority

```
src/couchers/crypto.py: everything
src/couchers/interceptors.py: auth interceptors, other stuff is lower priority
src/couchers/servicers/auth.py: everything
src/couchers/utils.py: everything
```

## Files that seem to contain fully untested RPCs

```
src/couchers/servicers/communities.py: ListPlaces, ListGuides, ListEvents
src/couchers/servicers/groups.py: ListPlaces, ListGuides, ListEvents
src/couchers/servicers/conversations.py: GetUpdates
src/couchers/servicers/gis.py: currently nothing is tested
src/couchers/servicers/search.py: UserSearch needs more testing
src/couchers/servicers/requests.py: MarkLastSeenHostRequest?
src/couchers/tasks.py: send_password_changed_email?
```

I think `gis.py` doesn't have tests because I knew it was a temporary solution, needs a rewrite as per #897.

aapeliv:
Friend requests are also completely untested

trevvorko:
If no one has already worked on this I'd like to contribute. I could probably write tests for the high priority section detailed above. Let me know if there's anything I should be aware of before starting.

aapeliv:
@ShmuelTreiger was going to work on this after some other issues, but I think it's fine for you to start working on this as I don't think Sam started yet. The tests are also quite decoupled so I'm sure there's also work for both of you if you wish to share!

ShmuelTreiger:
By all means. It's on my todo list, but feel free!

ShmuelTreiger:
Another area for improved testing would be ensuring that any api request which has a user_id:
[...]

PR #1067, opened 2021-04-12
Improved code coverage of auth.py

trevvorko:
Added more tests to cover more lines of auth.py. Mainly Authenticate
testing, but also some more missed error handling in CompleteSignUp,
Login, and Signup.

All tests pass, ran `isort` and `black`

Unable to assign reviewers, but I think @aapeliv or @ShmuelTreiger can take a look and let me know of any suggestions or comments.

Part of issue #900

### DIFF
--- a/app/backend/src/tests/test_auth.py
+++ b/app/backend/src/tests/test_auth.py
@@ -309,4 +309,272 @@ def test_signup_invalid_birthdate(db):
             assert e.value.details() == errors.INVALID_BIRTHDATE
 
 
+def test_invalid_signup_token_info(db):
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        with pytest.raises(grpc.RpcError) as e:
+            reply = auth_api.SignupTokenInfo(auth_pb2.SignupTokenInfoReq(signup_token="notarealtoken"))
+        assert e.value.code() == grpc.StatusCode.NOT_FOUND
+        assert e.value.details() == errors.INVALID_TOKEN
+
+
+def test_signup_invalid_email(db):
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email="a"))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.INVALID_EMAIL
+
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email="a@b"))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.INVALID_EMAIL
+
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email="a@b."))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.INVALID_EMAIL
+
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email="a@b.c"))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.INVALID_EMAIL
+
+
+def test_signup_existing_email(db):
+    # Signed up user
+    user, _ = generate_user()
+
+    # Attempt to signup again with the same email
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email=user.email))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.EMAIL_EXISTS
+
+
+def test_successful_authenticate(db):
+    user, _ = generate_user()
+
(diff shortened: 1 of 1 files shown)


app/backend/src/tests/test_auth.py
```diff
@@ -309,4 +309,272 @@ def test_signup_invalid_birthdate(db):
...
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        with pytest.raises(grpc.RpcError) as e:
+            reply = auth_api.Authenticate(auth_pb2.AuthReq(user=user.username, password="incorrectpassword"))
+        assert e.value.code() == grpc.StatusCode.NOT_FOUND
+        assert e.value.details() == errors.INVALID_USERNAME_OR_PASSWORD
+
+    # Invalid username
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        with pytest.raises(grpc.RpcError) as e:
+            reply = auth_api.Authenticate(auth_pb2.AuthReq(user="Notarealusername", password="password"))
```
aapeliv: Please make username lowercase, otherwise this will fail with the regex.

By the way, it's possible to stick any of three things in the user here: username, email, or user id. Look up `get_user_by_field` or similar

app/backend/src/tests/test_auth.py
```diff
@@ -309,4 +309,272 @@ def test_signup_invalid_birthdate(db):
...
+
+def test_unsuccessful_login(db):
+    # Invalid email
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Login(auth_pb2.LoginReq(user="fake@email.com"))
+    assert reply.next_step == auth_pb2.LoginRes.LoginStep.INVALID_USER
+
+    # Invalid username
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Login(auth_pb2.LoginReq(user="Notarealusername"))
```
aapeliv: Ditto

app/backend/src/tests/test_auth.py
```diff
@@ -309,4 +309,272 @@ def test_signup_invalid_birthdate(db):
...
+    # Signed up user
+    user, _ = generate_user()
+
+    # Attempt to signup again with the same email
+    with auth_api_session() as (auth_api, metadata_interceptor):
+        reply = auth_api.Signup(auth_pb2.SignupReq(email=user.email))
+    assert reply.next_step == auth_pb2.SignupRes.SignupStep.EMAIL_EXISTS
+
+
+def test_successful_authenticate(db):
```
aapeliv: Use the `fast_passwords` fixture in tests that do any password hashing: that will significantly speed the tests up.

aapeliv (changes requested):
This is looking very good, thanks!

I've added a few small comments, and I'm just running the CI on it.

One small thing, I think it'd probably be better to use non-existent but invalid emails in the tests. We've used `{whatever}@couchers.org.invalid` in a bunch of places. Just to protect in case someone runs the tests with actual SMTP credentials or something...

app/backend/src/tests/test_auth.py
trevvorko: I saw the user id being an option here, but when I try passing that to user I get this error within auth_pb2.AuthReq
`TypeError: 1 has type int, but expected one of: bytes, unicode`

So I was thinking there was some type restriction on what could be passed to AuthReq but I couldn't figure out how to fix it. Do you have any suggestions, or do you even think it's important I test username, email, and user id?

aapeliv: That field has to be passed in as a string, even if you're entering a number, due to the typing of gRPC.

But testing each of the three cases isn't *that* important, up to you if you feel like adding them.

trevvorko: Just as you were responding I found the code that expected it to be a string. You say it's due to gRPC typing, would it be beneficial to add functionality that would accept an int as well? Or would that be just a convenience thing for testing and not actually used in production?

aapeliv: We'd have to add another field, and then have something like 

```proto
message WhateverItsCalled {
  oneof user {
    int64 user_id = 1;
    string user_username = 2;
    string user_email = 3;
  }
}
```

it just moves the burden of regexping from the backend to the frontend. Easier to keep on the backend for now!

I think going forward we might want to get rid of the places where we do this though, user IDs should really never be exposed in the frontend

trevvorko: One final question, is it prefered I add another commit to fix these changes, and then end up merging both of them into develop (if approved), or should I amend the current commit so only one gets merged into develop?

aapeliv: Normally we add commits; it's more convenient for the reviewer to see the new changes, see the progression of CI/CD successes/failures, etc. But you can also amend if you prefer (as long as you only amend in a PR and not on `develop` as rewriting history is bad when multiple people have the branch checked out).

aapeliv (approved):
Looks great, thanks!

PR #1080, opened 2021-04-13
Fix up form flashing on load if it's already been filled

aapeliv:
If you've already filled in the form, then it'll flash on the screen before turning into the "You've already filled in the contributor form." message.

It's because after the query succeeds, loading stops, but `setFormFilled` is called only moments later in the `onSuccess` function... I didn't catch it earlier because it was fast before the preview being deployed in Sydney.

This is not a very neat way of doing it, open to other approaches.

### DIFF
--- a/app/frontend/src/features/contributorForm/ContributorForm.tsx
+++ b/app/frontend/src/features/contributorForm/ContributorForm.tsx
@@ -16,9 +16,10 @@ import CircularProgress from "components/CircularProgress";
 import TextField from "components/TextField";
 import { useAuthContext } from "features/auth/AuthProvider";
 import { GetContributorFormInfoRes } from "pb/account_pb";
+import { contributorFormInfoQueryKey } from "queryKeys";
 import React, { useState } from "react";
 import { Controller, useForm } from "react-hook-form";
-import { useMutation, useQuery } from "react-query";
+import { useMutation, useQuery, useQueryClient } from "react-query";
 import { Link } from "react-router-dom";
 import { signupRoute } from "routes";
 import { service } from "service";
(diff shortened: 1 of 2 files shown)


darrenvong:
Not sure I fully follow this... shouldn't the query loading state got you covered? So this is a workaround to ensure already filled is set first by blocking it with another loading state before unsetting that extra loading state to reveal it "safely"?

Yeah seems a bit hacky (on top of all of that crazy CSV stuff 🙈 )... unfortunately don't have time until probably tomorrow to properly look into this to suggest anything better

lucaslcode:
I think setting  `defaultValue` on the form elements should work without needing the hack. Since the form isn't rendered until it is loaded, `user` should be populated by the time those elements are created

aapeliv:
I think the `onSuccess` runs after `queryLoading` is set to false, so it causes a render just before `onSuccess` hides it again.

For the default values: we hide those fields if the user is logged in, they don't modify them, so defaultValue doesn't really work

lucaslcode:
Does this form show anywhere where the user is not logged in?

aapeliv:
Yes, there's an extra page, `/contribute` if we want to just directly link it to people externally

aapeliv:
It might be that I just don't understand react query loading states

aapeliv:
Okay, I fixed it up better: no need for an extra state variable, just read it from the data. I also noticed the contributor info was not invalidated after submitting the form, which now mattered as the data was used.

darrenvong:
> Okay, I fixed it up better: no need for an extra state variable, just read it from the data.

That was literally what I was gonna suggest next but only got time to come on here now 👍  nice refactor on the query key too

darrenvong (approved):

PR #1066, opened 2021-04-11
Add about/forum into navigation bar - fixed external link

jiwon1219:
@ineswq sorry I should've checked it more thoroughly. I apologize for any inconvenience it may have caused :( 
I've updated the external link which should resolve #939

I'm using storybook to test it and it shows the correct link (bottom left):
Screen Shot 2021-04-11 at 7 00 51 PM
Screen Shot 2021-04-11 at 7 01 05 PM

Comparison with other navigation bar link:
Screen Shot 2021-04-11 at 7 01 26 PM

PR #1081, opened 2021-04-13
Fix broken external links

javiwq:
Follows up on #1066

### DIFF
--- a/app/frontend/src/components/Navigation/ExternalNavButton.tsx
+++ b/app/frontend/src/components/Navigation/ExternalNavButton.tsx
@@ -1,0 +1,41 @@
+import { Typography, TypographyProps } from "@material-ui/core";
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  link: {
+    color: theme.palette.text.secondary,
+    display: "flex",
+    flex: "1",
+    fontSize: "2rem",
+    maxWidth: theme.spacing(21),
+    padding: theme.spacing(1, 1.5),
+    transition: theme.transitions.create(["color", "padding-top"], {
+      duration: theme.transitions.duration.short,
+    }),
+  },
+  label: {
+    alignSelf: "center",
+    marginTop: 0,
+  },
+}));
+
+interface ExternalNavButtonProps {
+  route: string;
+  label: string;
+  labelVariant?: Exclude<TypographyProps["variant"], undefined>;
+}
+
+export default function ExternalNavButton({
+  route,
+  label,
+  labelVariant,
+}: ExternalNavButtonProps) {
+  const classes = useStyles();
+  return (
+    <a href={route} className={classes.link}>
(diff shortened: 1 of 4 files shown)


app/frontend/src/components/Navigation/ExternalNavButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    padding: theme.spacing(1, 1.5),
+    transition: theme.transitions.create(["color", "padding-top"], {
+      duration: theme.transitions.duration.short,
+    }),
+  },
+  label: {
+    alignSelf: "center",
+    marginTop: 0,
+  },
+}));
```
darrenvong: I think it's better to pull this out from `NavButton` into its own file and have it called `useLinkStyles` or something similar and import it in both places, considering the styles are literally identical! Otherwise this is error prone for when we need to update styles for links

app/frontend/src/components/Navigation/ExternalNavButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  link: {
+    color: theme.palette.text.secondary,
+    display: "flex",
+    flex: "1",
+    fontSize: "2rem",
+    maxWidth: theme.spacing(21),
+    padding: theme.spacing(1, 1.5),
+    transition: theme.transitions.create(["color", "padding-top"], {
+      duration: theme.transitions.duration.short,
+    }),
```
darrenvong: What happens if we actually remove this? 🤔 
I got a feeling this is left over from when we had the old nav link style before so might be unnecessary now

darrenvong (changes requested):

app/frontend/src/components/Navigation/ExternalNavButton.tsx
```diff
@@ -0,0 +1,41 @@
+import { Typography, TypographyProps } from "@material-ui/core";
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  link: {
+    color: theme.palette.text.secondary,
+    display: "flex",
+    flex: "1",
+    fontSize: "2rem",
+    maxWidth: theme.spacing(21),
```
darrenvong:
```suggestion
    maxWidth: "10.5rem",
```
We've agreed recently that we probably shouldn't be using this `spacing` function for anything that aren't spacing related CSS (padding and margins).

darrenvong:
Sorry for the random comments that might be less related to your main change 😅 
but while you are at the area and the changes I suggested are fairly small, I thought we might as well clean it up along the way

app/frontend/src/components/Navigation/ExternalNavButton.tsx
javiwq: You're right! Nothing changes without it 🙂

javiwq:
Always down for a cleanup! I made the suggested changes and pulled the styling out to its own file.

darrenvong (approved):

PR #1086, opened 2021-04-15
Frontend/fix/map height too big

### LINKED ISSUE #1048: When landing on map page on mobile, map size is too big
Subject: When landing on map page on mobile, map size is too big
Description:
Clicked on map navigation link in drawer and upon landing of map page, bottom of map box extends beyond window of my mobile phone. See screenshot below.

Results:
Map box should be responsive vertically in addition to horizontally and should not extend beyond bottom of screen (which forces scrolling). Map page should not scroll at all (only map should be movable)

Backend version: develop-d732f04e
Frontend version: develop-d732f04e
Page: https://app.couchers.org/map#loc=1/0/0
User (spoofable): jesse (57)

justjooshing:
Should fix [#1048](https://github.com/Couchers-org/couchers/issues/1048), but I can't test on mobile outside of the dev tools

With extra padding, the map is cut off at the bottom, and you need to scroll to see it all (even though its 100vh - navPaddingMobile)
image

Without it sits nicely 
image

All of the pages had this weird extra top padding
image

Which they don't now
image

And the only one benefitting from it was the profile page
image

Which would have been cut off with this fix
image

But i've re-added the padding just to this one, so it looks the same as it did before


thdk:
Hi, I'm new here so forgive me if I did something wrong but I've tested this on android/chrome using the preview url

https://085e6c07--frontend.preview.couchershq.org/map#loc=1/52.2/9.4

and saw that the map was still cut off:



Probably because google chrome doesn't take the height of the address bar into account when using `height: 100vh`. Note that you can scroll down (by dragging on the small area on eather side of the map)  to hide the address bar, then the map nicely fits the entire screen.

You can set a height of 100% to each and every parent (including the body and html tag) to make sure the map will be visible. However then you won't be able to scroll down to hide the address bar anymore.

Another idea might be to scroll down when using one finger and pan the map when using two fingers. (although I usually don't like this behavior)

darrenvong:
Thanks for taking a look @thdk! Yeah unfortunately [`100vh` seems to be quite well known to not work well on mobile](https://chanind.github.io/javascript/2019/09/28/avoid-100vh-on-mobile-web.html) so I think what you suggested might be the best solution without resorting to a JavaScript based one which I don't really like either for something that feels like should be solvable by CSS these days...

justjooshing:
Ha thanks @thdk! So it seems all my fix did was remove the extra padding.
@darrenvong I wasn't sure whether it was a deliberate design choice to have that much padding, shall I revert it?
Also, do we really want the map to cover the whole screen anyway? Why not set the map height to 90vh instead? Then it'll fit regardless of whether the url bar is visible

lucaslcode:
> Why not set the map height to 90vh instead

I like this. Why use lots of code when little code does trick? (It's not robust, but I think a robust version solves a problem that doesn't exist)

thdk:
I think the map doesn't need fill the entire height as well. So we could indeed reduce the map height (for all browsers?) or we can leave it like it is because that's how google decided it should work :)

lucaslcode:
Actually we may need to have it accurately 100vh when we have the proper map design, for the fullscreen effect...
In this case I think it can be achieved with some `useLayoutEffect` and `innerHeight`. Not ideal, but it will be difficult to set `100%` in html and body for only one route (unless we can just do it for all routes? Does it break scrolling?)

darrenvong:
> it will be difficult to set 100% in html and body for only one route

Why only one route? I don't think setting that at least at the html and body is a bad idea since height doesn't "inherit" like width does, and it makes building certain layouts easier if we want to keep height proportional, even though it might means we have to "drill" height down, but hopefully not too many levels for intentional full screen layout?

app/frontend/src/features/map/MapPage.tsx
```diff
@@ -11,27 +12,39 @@ import { MAP_PAGE } from "./constants";
...
+  const [windowHeight, setWindowHeight] = useState(window.innerHeight);
+
+  //Update on orientation change
+  useLayoutEffect(() => {
+    const updateWindowHeight = () => {
+      setWindowHeight(window.innerHeight);
+    };
+    window.addEventListener("resize", updateWindowHeight);
+    return () => window.removeEventListener("resize", updateWindowHeight);
+  }, []);
```
darrenvong: If we are going down the JS code route... I think it'd be worth extracting this into a custom hook as I can see this being one of those utilities that would be re-usable for other components that need this

darrenvong: Also, we should add a debounce around the `updateWindowHeight` if sticking with this, otherwise the `resize` event could fire too often and cause too many unnecessary re-renders

lucaslcode:
After reading about the "100%" on body and html thing, I think it's probably preferable. Do you want to try and see which is nicer?

lucaslcode (changes requested):
As it is currently, profile page has too much padding on desktop.

The problem is, the margins are spread over a few places:
- h1 has 1.5rem margin top
- navBarPadding which you just removed
- probably some other places in individual pages?

I'd be tempted to remove the padding on h1, and reduce navBarPadding to 1.5, and get rid of the top padding on the profile page. This might break some things though where they expect h1 to have padding.

Hmm... if we assume that it is expected for h1 to have padding, maybe removing navBarPadding is correct. So just the profile page needs fixing. That way pages are responsible for their own top margin, and they don't have to worry about it if they use h1.

darrenvong:
> * h1 has 1.5rem margin top

I have to say this is kinda annoying me, as in some places (lol aka the discussion stuff I'm working on) I had to override it again since it creates undesirable spacing at the top so I agree with removing it from the h1.

[This article might have some strong opinions about margins](https://mxstbr.com/thoughts/margin/) as I'm a bit iffy about adding junk elements for spacing, but sums up my thoughts when it comes to reusable components having its own margin. The whole margin collapse thing is an annoyance that adds to your mental overhead to think about when you try to use a component with margin and lay things out with other things, when I'd rather plug it in a component/page specific container and let that deal with spacing concerns.

lucaslcode:
True. Maybe then let's remove all those `marginTop`s from the theme on the headers.

justjooshing:
Oke doke my dudes. I'll switch it all up 👌

justjooshing:
cool @lucaslcode @darrenvong I've done that
After removing the marginTop from the theme headers, I saw that actually the 1.5rem was coming from PageTitle. I've changed it from margin to padding which helps with the mental overhead, but can remove it all together if you want

justjooshing:
Hmm ok we're not there just yet. I have the map still overflowing on mobile, 
And then when set height: 100% the rest of the way down from body to the map, it thinks height is 0

image

justjooshing:
Cool, that's working now

app/frontend/src/theme.ts
```diff
@@ -87,7 +86,7 @@ const themeOptions: ThemeOptions = {
   shape: {
     borderRadius,
     navPaddingDesktop: spacing(10),
-    navPaddingMobile: spacing(navBarHeight + navBarPadding),
+    navPaddingMobile: spacing(navBarHeight),
```
lucaslcode, line 89: Can you also delete the variable from constants.ts or wherever it is?

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -39,9 +39,10 @@ const useStyles = makeStyles((theme) => ({
     textDecoration: "none",
   },
   root: {
+    paddingTop: theme.spacing(3),
     [theme.breakpoints.up("md")]: {
+      paddingTop: 0,
       display: "flex",
-      justifyContent: "space-around",
```
lucaslcode, line 44: I'm not at my desktop - is the profile page still centered when this is removed?

lucaslcode (dismissed):
Minor comments. I like that this makes `PageTitle` no longer useless, since it adds the top padding. Before, I was thinking we should get rid of it.

app/frontend/src/features/profile/view/ProfilePage.tsx
justjooshing: It is

app/frontend/src/theme.ts
justjooshing: It was in this file on line 18
unless you mean something else

lucaslcode: Sorry I missed it

lucaslcode (commented):
LGTM, I want @darrenvong to check this before merging

darrenvong (approved):
Looks good to me too! Quite happy with how simple this turns out once all the JS are removed :)

justjooshing:
> Looks good to me too! Quite happy with how simple this turns out once all the JS are removed :)

Yeah me too, so much cleaner!

PR #1089, opened 2021-04-16
Frontend/chore/typecheck unit tests

### LINKED ISSUE #851: Tests pass even when there are TypeScript errors
According to https://github.com/facebook/create-react-app/issues/5626, there's a limitation in how Create React App sets out jest, which means type checking isn't being done properly on test files. That means tests could still be passing even if they should break because of TS error.

The workaround seems to be adding `tsc` before running the tests so the test files goes through the TS type checker and only run the tests if succeeds. Will have to fix some broken test files as a result of this.

Also, linting also don't seem to be applied to test files which would normally fail the app build, so we may need to add a similar workaround to the above by running the `eslint` command against files ending in `test.ts[x]*`.

thdk:
This PR closes previous one #1062

First CI should fail because tsc now runs before running unit tests.
Next I'll push a commit to this PR that fixes the existing typing errors in the unit tests.

Fixes #851


thdk:
seems I forgot to run `yarn install` locally, that's why CI kept failing but worked locally... next run hopefully better luck :)

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -12,29 +12,53 @@ const updateHostingPreferenceMock = service.user
...
+          default:
+            acc[key] = {
+              value: newHostingPreferenceData[key],
+            };
+            return acc;
         }
-        return acc;
       },
-      {}
+      {} as Record<string, unknown>
```
darrenvong: What's wrong with using `Object.entries` as before? You could literally swap the switch logic you got there with the old logic and everything works fine without all the extra type assertions, which I'm not keen of doing unless absolutely needed as they are not type safe.

app/frontend/package.json
```diff
@@ -31,10 +31,11 @@
     "eject": "react-scripts eject",
     "format": "prettier --write src/",
     "lint": "eslint src/",
-    "test-ci": "CI=true react-scripts test --coverageReporters=\"text\" --coverageReporters=\"cobertura\" --coverageReporters=\"lcov\" --reporters=\"default\" --reporters=\"jest-junit\" --coverage",
+    "test-ci": "yarn tsc && CI=true react-scripts test --coverageReporters=\"text\" --coverageReporters=\"cobertura\" --coverageReporters=\"lcov\" --reporters=\"default\" --reporters=\"jest-junit\" --coverage",
```
darrenvong:
```suggestion
    "test-ci": "tsc && CI=true react-scripts test --coverageReporters=\"text\" --coverageReporters=\"cobertura\" --coverageReporters=\"lcov\" --reporters=\"default\" --reporters=\"jest-junit\" --coverage",
```
I think `tsc` alone is enough since anything running through npm/yarn scripts will find the executable correctly without the extra step of having to run it through another custom npm/yarn command

app/frontend/package.json
```diff
@@ -31,10 +31,11 @@
...
     "jest": "jest --modulePaths=src",
     "storybook": "start-storybook -p 6006 -s public",
-    "build-storybook": "NODE_PATH=. build-storybook -s public"
+    "build-storybook": "NODE_PATH=. build-storybook -s public",
+    "tsc": "tsc"
```
darrenvong:
```suggestion
```

darrenvong (changes requested):
Generally looks good, thanks for doing the chore work to make our codebase better! ✨

app/frontend/package.json
thdk: Yes, I shall change it back to just 'tsc' indeed

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
thdk: I can use object.entries as well, but even then I would like to add the type assertions. Without these type assertings, the 'key' variable can contain any string and the typescript compiler won't tell you if you make a typo or if a property gets removed or renamed.

app/frontend/package.json
thdk: I added this line so developers could do 'yarn tsc' locally using typescript version from node_modules instead of globally installed typescript version. Should I still remove it?

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -41,25 +41,22 @@ describe("useUpdateHostingPreference hook", () => {
...
+        case "smokingAllowed":
+        case "parkingDetails":
+        case "sleepingArrangement":
+          acc[key] = value;
+          return acc;
+        default:
+          acc[key] = { value: value };
+          return acc;
+      }
+    }, {} as Record<string, string | number | boolean | null | { value: string | number | boolean | null }>);
```
darrenvong:
```suggestion
    type Primitives = string | number | boolean | null;
    type HostPreferenceValue = Primitives | { value: Primitives };
    const newUserPref = Object.entries(newHostingPreferenceData).reduce(
      (acc: Record<string, HostPreferenceValue>, [key, value]) => {
        switch (key) {
          case "smokingAllowed":
          case "parkingDetails":
          case "sleepingArrangement":
            acc[key] = value;
            return acc;
          default:
            acc[key] = { value };
            return acc;
        }
      },
      {}
    );
```
I think annotating the accumulator object and type aliasing it makes the types a bit neater to read and eliminates most needs for type casting.

darrenvong (changes requested):

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
thdk: Yes, that looks better! Except, I believe the type for HostPreferenceValue should be: 
`number | { value: Primitives }`

Since I think this type can be used on other places as well, should I add the following to `utils/types.ts`?
```
export type Primitives = string | number | boolean | null;
export type ServicePrimitives = number | { value: Primitives };
```

Or what is a better name for the `ServicePrimitives` type?

thdk: But your suggestion still drops the type safety (and intellisense) for the properties of type `HostingPreferenceData`.

```
    const newUserPref = (Object.entries(newHostingPreferenceData) as [
      keyof HostingPreferenceData,
      Primitives
    ][]).reduce((acc, [key, value]) => {
      switch (key) {
        case "smokingAllowed":
        case "parkingDetails":
        case "sleepingArrangement":
          if (typeof value !== "number") {
            throw new Error(`Value of enum ${key} must be a number.`);
          }
          acc[key] = value;
          return acc;
        default:
          acc[key] = { value };
          return acc;
      }
    }, {} as Record<string, ServicePrimitives>);
```

Using the `keyof HostingPreferenceData` type gives the following intellisense in vscode:

image

and the following added type safety which I think is needed to prevent errors after properties either renamed or removed:

image

Without using the type `keyof HostingPreferenceData`:

image

thdk:
I merged develop into this branch. Is that ok?

app/frontend/package.json
darrenvong: I'm not sure when I'd run `tsc` separately when in VS Code tbh, since you get a live `tsc` run by the editor itself, so I think I'd rather not leave rarely/almost unused commands around

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -3,39 +3,64 @@ import useUpdateHostingPreferences from "features/profile/hooks/useUpdateHosting
...
-    const newUserPref = Object.entries(newHostingPreferenceData).reduce(
-      (acc: Record<string, unknown>, [key, value]) => {
-        if (key !== "smokingAllowed") {
-          acc[key] = { value };
-        } else {
+
+    const newUserPref = (Object.entries(newHostingPreferenceData) as [
+      keyof HostingPreferenceData,
+      Primitives
+    ][]).reduce((acc, [key, value]) => {
```
darrenvong:
```suggestion
    ][]).reduce((acc: Record<string, ServicePrimitives>, [key, value]) => {
```
Just personal preference to avoid type assertions if things work without it for reasons mentioned in the other comment

darrenvong:
On your point about type safety: this is unfortunately impossible since all of the `Object.*` types don't offer generics which would do it safely.

Even with your current code, there's nothing stopping anyone from doing this:
```typescript
const newUserPref = Object.entries({...newHostingPreferenceData, randomKey: "I shouldn't be here"}) as [
      keyof HostingPreferenceData,
      Primitives
    ][]).reduce(...)
```
When you then call `reduce` with that, `randomKey` is not part of `HostingPreferenceData`'s key set, but since you type asserted the result of `Object.entries`, TypeScript still does not protect you here since type assertions is an escape hatch and so not type safe.

But both my "randomKey" and your "I can write whatever" are extreme examples anyway, and really should be picked up in code review if anyone tries to do either of those things. Having said that, in this case, I'm happy to acknowledge the TS types are not perfect and your type assertions here would still offer better `key` autocomplete (as we clearly know that object can only have keys from the `HostingPreferenceData` type), so let's leave it in.

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -3,39 +3,64 @@ import useUpdateHostingPreferences from "features/profile/hooks/useUpdateHosting
...
-        return acc;
-      },
-      {}
-    );
+          return acc;
+        default:
+          acc[key] = { value };
+          return acc;
+      }
+    }, {} as Record<string, ServicePrimitives>);
```
darrenvong:
```suggestion
    }, {});
```

darrenvong (changes requested):
Sorry, definitely the last one! Good call on pulling out the types into the utils file

app/frontend/package.json
thdk: I thought I would run it before committing changes, but it seems I run `yarn test-ci` instead :) I agree on not leaving rarely used commands so I shall remove it. 

Note however: running `yarn test` still won't faill if affected tests contain typescript errors. But your CI build will fail.

thdk:
> Sorry, definitely the last one!

Don't worry, thanks for reviewing!

thdk:
question: should I rebase and cleanup this branch? Or other best practices for PR's I should know?

lucaslcode:
No need to squash but there are a bunch of external commits in this branch, so maybe rebase. Not sure how they got there. Usually merging in develop doesn't do that.

lucaslcode (dismissed):

darrenvong (dismissed):

thdk:
Fixed more tests after rebasing on develop

app/frontend/src/utils/types.ts
```diff
@@ -29,3 +29,6 @@ export type ProtoToJsTypes<T> = T extends StringValue.AsObject
 export type SetMutationError = React.Dispatch<
   React.SetStateAction<string | null>
 >;
+
+export type Primitives = string | number | boolean | null;
+export type ServicePrimitives = number | { value: Primitives };
```
lucaslcode: It's not clear to me what this type is for from the name. 

If it's some kind of protobuf message type, I would expect it to be `Primitives | { value: Primitives }`

But depending on where it is used, maybe better to just use that inline?

thdk: I don't know the internals enough yet to decide this. I was just looking that at this moment the values being assigned to the variable are either `number` or `{value: Primitives}` . But I can move it back inline.

lucaslcode: Yeah maybe do inline, that way we don't need to think of a descriptive name. And I don't think it will be needed that much (otherwise we can add it later)

thdk: I've moved `ServicePrimitives` to an inline type. Or did you mean to also move the `Primitives` type to _useUpdateHostingPreferences.test.tsx_?

lucaslcode: nope this is good

lucaslcode (approved):
lgtm

PR #1091, opened 2021-04-16
Friend Request on Profile

javiwq:
This should be close to complete, but has a bug I cannot figure out: Still displays option to respond to friend request on profile after a request has been declined. 

Accepting friend request from profile works, and persists: 

https://user-images.githubusercontent.com/35485667/115045301-ab5b2800-9ed6-11eb-87ba-ee87dcd71bb9.mov

However after declining a friend request the option to respond re-appears after refresh, even though the request has indeed been declined and no longer exists:

https://user-images.githubusercontent.com/35485667/115045826-2d4b5100-9ed7-11eb-8990-aa8cf85d6f44.mov

This bug exists when responding to friend request from the `connections`page also:

https://user-images.githubusercontent.com/35485667/115046241-887d4380-9ed7-11eb-9ff5-147d20a7ca1a.mov


lucaslcode:
`useRespondToFriendRequest` doesn't invalidate the user in the cache (maybe it wasn't needed before?), so the request status in the user object isn't being updated after it is changed. 

I would change it so that the whole friend request is passed instead of the request id, then in `useRespondToFriendRequest` in the `onSucess` you can add `queryClient.invalidate(userQueryKey(friendRequest.fromId))` and same to `toId`. Or something along those lines.

javiwq:
Thank you Lucas! That did it 👍

app/frontend/src/features/connections/friends/AddFriendButton.tsx
```diff
@@ -49,16 +38,12 @@ export default function AddFriendButton({
   return (
```
lucaslcode, line 57: I can't comment on the correct line, but this refers to the `onSuccess` function above.

At the moment, there's a bit of jumping around: normal state --click--> goes to loading state --success--> normal state (invalidation hasn't happened yet) --invalidation complete--> disappear

I think we should get rid of the third one by optimistically updating the query cache. That is, since it was successful, make the change immediately as well as invalidating, which will prevent going back to the normal state which waiting for the invalidation to complete.

To do that, in `onSuccess`, add something like this:  
`
      const cachedUser = queryClient.getQueryData<User.AsObject>([
        "user",
        userId,
      ]);
      queryClient.setQueryData(["user", userId], {
        ...cachedUser,
        friends: User.FriendshipStatus.PENDING,
      });`

app/frontend/src/features/profile/actions/PendingFriendReqButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+  };
+
+  const handleClose = (item: keyof typeof isOpen) => () => {
+    setIsOpen((prevState) => ({ ...prevState, [item]: false }));
+  };
+
+  return (
+    <>
+      {isLoading ? (
+        <CircularProgress />
```
lucaslcode: Instead of doing this, do `loading={isLoading}` on the button

app/frontend/src/features/profile/actions/PendingFriendReqButton.tsx
```diff
@@ -0,0 +1,110 @@
...
+                onClick={() => {
+                  respondToFriendRequest({
+                    accept: true,
+                    friendRequest,
+                    setMutationError,
+                  });
+                }}
+              >
+                {ACCEPT_FRIEND_ACTION}
+              </Button>
```
lucaslcode: Menu items don't need a button, see the menu which appears when you click the three dots

app/frontend/src/features/profile/actions/FriendActions.tsx
```diff
@@ -0,0 +1,47 @@
...
+  user: User.AsObject;
+  setMutationError: SetMutationError;
+}
+
+export default function FriendActions({
+  user,
+  setMutationError,
+}: FriendActionsProps) {
+  const { isLoading: isFriendRequestsLoading } = useFriendRequests("received");
+  const { isLoading: isFriendsLoading } = useFriendList();
```
lucaslcode: It's better for the component that uses the data to display loading indicators. They can have better styling more similar to the end result.

app/frontend/src/features/profile/actions/FriendActions.tsx
```diff
@@ -0,0 +1,47 @@
...
+      (component = (
+        <PendingFriendReqButton
+          friendRequest={user.pendingFriendRequest}
+          setMutationError={setMutationError}
+        />
+      ));
+
+  const isLoading = isFriendRequestsLoading || isFriendsLoading;
+
+  return isLoading ? <CircularProgress /> : <>{component}</>;
```
lucaslcode: Since you have the big ternary statement anyway, don't use a `component` variable, just return the ternary statement.

Alternatively, you may find it clearer to do
```
if (user.friends === User.FriendshipStatus.NOT_FRIENDS) {
  return ...
}
if (user....) {
  return ...
}
//etc
```

app/frontend/src/features/connections/friends/useRespondToFriendRequest.ts
```diff
@@ -30,9 +31,10 @@ export default function useRespondToFriendRequest() {
         setMutationError("");
         await queryClient.cancelQueries(friendRequestKey("received"));
       },
-      onSuccess: () => {
+      onSuccess: (_, { friendRequest }) => {
         queryClient.invalidateQueries("friendIds");
```
lucaslcode, line 64: Do an optimistic update of the `user` in the query cache here as before. That will reduce buttons jumping around.

app/frontend/src/features/connections/friends/FriendRequestsReceived.stories.tsx
```diff
@@ -46,6 +46,7 @@ WithPendingRequests.args = {
         friendRequestId: 1,
```
lucaslcode, line 46: This and similar needs to be changed to a `friendRequest`

app/frontend/src/features/connections/friends/FriendRequestsSent.stories.tsx
```diff
@@ -47,6 +47,7 @@ WithPendingRequests.args = {
         friendRequestId: 1,
```
lucaslcode, line 47: This and similar needs to be a `friendRequest`

app/frontend/src/features/connections/friends/useRespondToFriendRequest.test.tsx
```diff
@@ -27,6 +27,7 @@ describe("useRespondToFriendRequest hook", () => {
         friendRequestId: 1,
```
lucaslcode, line 27: `friendRequest`

app/frontend/src/features/profile/actions/PendingFriendReqButton.tsx
```diff
@@ -0,0 +1,110 @@
+import { CircularProgress } from "@material-ui/core";
```
lucaslcode: Use our component

lucaslcode (changes requested):
Sorry I was so slow to review!

app/frontend/src/features/connections/friends/FriendRequestsReceived.stories.tsx
javiwq: How do you mean this? Lines 46 - 49 are a mock `friendRequest`.

app/frontend/src/features/connections/friends/AddFriendButton.tsx
javiwq: Based on the [React Query docs](https://react-query.tanstack.com/guides/optimistic-updates) I added the optimistic updates to `onMutate` rather than `onSuccess`, but thank you for catching this because it looks a lot smoother now 🙂.

app/frontend/src/features/connections/friends/FriendRequestsSent.stories.tsx
javiwq: As above!

app/frontend/src/features/profile/actions/FriendActions.tsx
javiwq: This is here because if the component doesnt re-query the pending requests and the friends list, the component will not correctly update after a sent request has been responded to, and so I was using the `useFriendRequests` and `useFriendList` hooks to do that and attempting to display a loading circle if those queries are loading. Perhaps it is just better/cleaner to write a new Query directly in this component?

javiwq (commented):
No worries at all, thanks for all your comments. I made most suggested changes but just still have a couple questions.

app/frontend/src/features/connections/friends/FriendRequestsReceived.stories.tsx
lucaslcode: Sorry, careless reviewing

app/frontend/src/features/connections/friends/AddFriendButton.tsx
lucaslcode: Nice, I didn't say that because you need to roll it back on error which you have done now.

app/frontend/src/features/profile/actions/FriendActions.tsx
lucaslcode: I see what you mean, but it won't work for two reasons. Firstly you can't rely on queries to actually query new data - they might return something from the cache without fetching if it's still considered fresh.  (We have set users to stay fresh for 5 minutes I think). If it needs to be re-fetched, there needs to be invalidation somewhere. 

Secondly, it's too opaque - this component decides what to render from `user.friends`, making it responsible for having an up to date user object is mixing responsibilities, which could lead to difficult-to-trace bugs.

In any case, I don't think it's actually needed now since you added invalidating the user on the successful response.

javiwq: Makes sense! I've removed it.

lucaslcode (approved):
Looks good! Do you want to do some tests or save it for another pr?

javiwq:
I am going to work on References next so ideally another PR.

PR #1099, opened 2021-04-17
Invalidate user when visiting their profile

lucaslcode:


darrenvong (approved):

darrenvong:
Lol this was not needed before as the accept friend flow was not part of the profile

lucaslcode:
It's not technically needed now, but I always intended to do this anyway

PR #1100, opened 2021-04-17
[WIP/Help] Backend/feature/user hiding/make community group tests explicit

aapeliv:
Reopened #1094.

GitHub closed that one since I renamed the branch. I had to rename the branch, there were problems because it started with a string that was the same name as another branch.

### DIFF
--- a/app/backend/src/tests/test_communities.py
+++ b/app/backend/src/tests/test_communities.py
@@ -196,25 +196,28 @@ def testing_communities(request):
     user6, token6 = generate_user(username="user6", geom=create_1d_point(65), geom_radius=0.1)
     user7, token7 = generate_user(username="user7", geom=create_1d_point(80), geom_radius=0.1)
     user8, token8 = generate_user(username="user8", geom=create_1d_point(51), geom_radius=0.1)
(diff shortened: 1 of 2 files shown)


ShmuelTreiger:
To add more details:

I'm attempting to explicitly test hidden users in groups and communities. In the base branch (user hiding), I just added an invisible user to each group and community and assumed that if all tests are passing, everything must be OK. Less than ideal approach. However, the proximity stuff in communities has me beat. Adding users changes the results of all the tests. And then I couldn't figure out how to test GetNearbyUsers. So I figured the base branch will do for now and someone who understands this better can fix it up at some point if they feel like it.

aapeliv:
@ShmuelTreiger: sorry, I don't understand what's going on here.

If you add a hidden/deleted user, then that should be sufficient: they shouldn't show up in any of the queries, right?

aapeliv:
Should we close this one too and work instead on top of the new PR?

ShmuelTreiger:
Yes, I was going to just rebase it

aapeliv:
@ShmuelTreiger: I'll close this PR, feel free to re-open. It's better that way so it doesn't get confusing about who the author is, etc.

PR #1103, opened 2021-04-19
[MRG] Moving mutual_friends to api from user model + worked on some testing

ShmuelTreiger:
Moved mutual_friends to api and edited some other tests in test_api while I was there


app/backend/src/couchers/servicers/api.py
```diff
@@ -619,10 +669,9 @@ def user_model_to_pb(db_user, session, context):
         countries_lived=db_user.countries_lived.split("|") if db_user.countries_lived else [],
         additional_information=db_user.additional_information,
         friends=get_friends_status(session, context.user_id, db_user.id),
-        mutual_friends=[
-            api_pb2.MutualFriend(user_id=mutual_friend.id, username=mutual_friend.username, name=mutual_friend.name)
-            for mutual_friend in db_user.mutual_friends(context.user_id)
-        ],
+        mutual_friends=API.ListMutualFriends(
+            self=None, context=context, request=api_pb2.ListMutualFriendsReq(user_id=db_user.id)
```
ShmuelTreiger: I'm sure this can be done better, but I'm not sure how. I feel like using self=None is a codesmell

app/pb/api.proto
```diff
@@ -28,6 +28,8 @@ service API {
 
   rpc ListFriends(google.protobuf.Empty) returns (ListFriendsRes);
 
+  rpc ListMutualFriends(ListMutualFriendsReq) returns (ListMutualFriendsRes);
```
kalvdans: It should be an error to send in your own user_id here, in my opinion. Either way, it should be documented here what happens when you do so.

app/backend/src/couchers/servicers/api.py
kalvdans: Good question, I think it is ugly too. Please make a helper function on the top level, that don't take the "self" parameter.

kalvdans (dismissed):

app/pb/api.proto
ShmuelTreiger: The way it was handled in models was to return an empty list, so the current method does the same. I imagine it's so that when you look at your own profile it will just have an empty list instead of an error. We can pose the question to the dev team at large or else just leave it as is until it needs changing.

But yes, I'll add a note.

ShmuelTreiger:
Is something like this better?

kalvdans:
> Is something like this better?

Better but not perfect :) Please make the helper function take user_id as parameter instead of a request, and that it return a list of api_pb2.MutualFriend.

That way, we avoid creating objects just to unpack them the first thing we do.

app/pb/api.proto
kalvdans: I think it is perfectly fine for `org.couchers.api.core.GetUser()` rpc call to return empty list for your own profile, while in this separate rpc call throws an error. Anyway, the actual behaviour is not very important as long as it is documented :)

kalvdans (approved):
LGTM!

Issue #1021, opened 2021-04-08
alert only pops up at the top, 


CouchersBot:
Subject: alert only pops up at the top, 
Description:
when the save is at the bottom of the screen you can't see the alert and don't know if there is a problem or was it saved.

Steps:
hit save, and "nothing" happened. had to look for the problem manually

Results:
alert should either pop up near the save button or screen should scroll automatically to the alert/problem

also, might be good to anchor the save button, so no need to go all the way down.

Backend version: develop-1f987511
Frontend version: develop-1f987511
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.114 Safari/537.36
Page: https://app.couchers.org/user/edit#loc=8.51/52.2323/5.18
User (spoofable): miuriel (144)

PR #1111, opened 2021-04-19
Frontend/fix/scroll to top on save

justjooshing:
For uniformity with saving EditHostingPreferences as well
Both redirect back to profile on successful save, but previously EditProfile wasn't scrolling to the top (and directing the user to the error) on unsuccessful save

Will close issue [#1021](https://github.com/Couchers-org/couchers/issues/1021).


lucaslcode (changes requested):
This doesn't scroll up for form validation errors, since that function isn't called. I think `handleSubmit` also takes a second callback for when the form is not valid.

It's still not great though to just scroll up, since it takes a second to load. It would be good if the button could turn into a loading button when clicked, and scroll up on error once loading is complete.

This can be done by exposing `isLoading` and [`mutateAsync`](https://react-query.tanstack.com/guides/mutations#promises) from the `useUpdateUserProfile` mutation

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -151,11 +153,30 @@ export default function EditProfileForm() {
...
+            window.scroll({ top: 0 });
+          },
+        }
+      );
+    },
+    () => {
+      // All field validation errors should scroll to their respective field
+      // Except the avatar, so this scrolls to top on avatar validation error
+      if (errors.avatarKey) {
+        window.scroll({ top: 0 });
```
darrenvong: Maybe this one is good to check with design too... but I wonder if it'd be easier/better to use snackbars instead for the error state too so we don't need to litter these scroll ups all over our components

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -151,10 +153,30 @@ export default function EditProfileForm() {
...
+        },
+        {
+          // Scoll to top on submission error
+          onError: () => {
+            window.scroll({ top: 0 });
+          },
+        }
+      );
+    },
+    () => {
```
lucaslcode:
```suggestion
    (errors) => {
```
It isn't working for me - I guess it's because `errors` from `useForm` isn't set until after this callback is executed? But the errors are also passed into the function so try this.

lucaslcode (changes requested):

lucaslcode:
how is this going @justjooshing ?

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
```diff
@@ -15,22 +15,27 @@ export default function useUpdateUserProfile() {
...
+    isLoading,
+    isError,
+    status,
+  } = useMutation<Empty, Error, UpdateUserProfileVariables>(
+    ({ profileData }) => service.user.updateProfile(profileData),
+    {
+      onError: (error, { setMutationError }) => {
+        setMutationError(error.message);
+      },
+      onMutate: async ({ setMutationError }) => {
```
lucaslcode: Since there's a merge conflict anyway, remove async here too :P

lucaslcode (changes requested):
lgtm except for one tiny thing and merge conflict!

lucaslcode (approved):

Issue #983, opened 2021-04-05
[signup] Link color difficult to read


justjooshing:
On mobile the Couchers guidelinks link is difficult to read, and I'm sure the text/background contrast would fail accessibility tests (particularly the field labels).

![image](https://user-images.githubusercontent.com/60824744/113582942-ea5cc400-9620-11eb-99da-a7b547be98b5.png)

On desktop its fine (below) so maybe we should just make the mobile signup background white as well?
![image](https://user-images.githubusercontent.com/60824744/113583066-124c2780-9621-11eb-8e10-226b015743ee.png)

alexishonzik:
Hi! I am interested in tackling this issue. I am currently downloading the repo, and was wondering if I could have access to the Slack channel to ask a question. Thank you!

lucaslcode:
Hi @alexishonzik! Send me an email: Lucas (ä t) couchers . org

alexishonzik:
@lucaslcode Just did! thanks!

alexishonzik:
Hi! To clarify, would you prefer to have the form wrapper like the desktop version or to have the entire background be white? I have added two images so that you can compare for your preference. Let me know which is best!
Just changing the background color:
<img width="809" alt="Screen Shot 2021-04-19 at 2 14 15 PM" src="https://user-images.githubusercontent.com/44992497/115283993-1ba5bb80-a11a-11eb-8e61-2e1af42bd075.png">

Creating a form wrapper:
<img width="811" alt="Screen Shot 2021-04-19 at 2 09 23 PM" src="https://user-images.githubusercontent.com/44992497/115284000-1e081580-a11a-11eb-968d-d4ee52de21cf.png">

lucaslcode:
The fields should probably take the whole width on mobile, so the former.

alexishonzik:
I realized that I cloned 'https://github.com/Couchers-org/couchers.git', but I don't believe I have write access to publish the branch I created. Should I fork the repo and create a PR through that?

lucaslcode:
Yep!

lucaslcode:
closed in #1112

PR #1112, opened 2021-04-19
Frontend/design/signup background

alexishonzik:
This PR is to resolve Issue #983. I removed the div that placed the background image for the form on the mobile view, so that the background is now white. This resolves the problem with not being able to clearly read the form.

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -175,7 +175,7 @@ class API(api_pb2_grpc.APIServicer):
 
             return user_model_to_pb(user, session, context)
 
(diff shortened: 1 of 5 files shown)


lucaslcode (changes requested):
Thanks! Can you also remove the one in `login.tsx` and then remove the `backgroundBlurImage` style, since it will no longer be used?

alexishonzik:
Just changed `login.tsx` and removed `backgroundBlurImage`!

lucaslcode (approved):
Thanks for your contribution!

lucaslcode:
closes #983

Issue #1041, opened 2021-04-09
Text color of navigation text in top bar not the same as Couchers.org wordmark


CouchersBot:
Subject: Text color of navigation text in top bar not the same as Couchers.org wordmark
Description:
The orange of the navigation text in the top bar (Dashboard, events, messages, etc) is not the same orange as the Couchers.org wordmark.

Results:
The oranges should be the same.

Backend version: develop-d732f04e
Frontend version: develop-1f987511
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:87.0) Gecko/20100101 Firefox/87.0
Page: https://app.couchers.org/
User (spoofable): jesse (57)

jesseallhands:
![image](https://user-images.githubusercontent.com/79384268/114246747-016d2000-9959-11eb-9168-e4803639bda1.png)

Sorry about the multiple reporting. I'm not getting a snackbar that tells me I submitted the bug report successfully...

lucaslcode:
closed in #1113

PR #1113, opened 2021-04-19
nav bar color switched

laurenfalvo:
this PR is to resolve issue #1041 I changed the color of the navigation text to match that of Couchers.org wordmark by switching the color to the secondary.main color rather than the secondary.dark


lucaslcode (approved):

lucaslcode:
Thanks!

PR #1114, opened 2021-04-20
Discussions 2/?: tidied up discussion list + add new discussion form

darrenvong:
2/? for #543. As in title, and also added tests for those.

Bit short on time so gonna be lazy on the screenshot on this one 🙈  (but I'll add some in if I have more time)


app/frontend/src/features/communities/hooks.ts
```diff
@@ -126,23 +126,24 @@ export const useListNearbyUsers = (communityId?: number) =>
...
+
+export const useNewDiscussionMutation = (onSuccess?: () => void) => {
+  const queryClient = useQueryClient();
+  return useMutation<Discussion.AsObject, GrpcError, CreateDiscussionInput>(
     ({ title, content, ownerCommunityId }) =>
       service.discussions.createDiscussion(title, content, ownerCommunityId),
     {
       onSuccess(_, { ownerCommunityId }) {
+        onSuccess?.();
         queryClient.invalidateQueries(
```
lucaslcode, line 143: We need an optimistic update here, at the moment there's a period when the new discussion box closes but the new discussion hasn't appeared yet.

app/frontend/src/features/communities/constants.ts
```diff
@@ -9,8 +9,13 @@ export const communityTabBarLabels: Record<CommunityTab, string> = {
   hangouts: "Hangouts",
 };
 
+export const getByCreator = (name: string) => `By ${name}`;
+
 export const COMMUNITY_HEADING = (name: string) => `Welcome to ${name}!`;
 export const COMMUNITY_TABS_A11Y_LABEL = "Tabs for community sub-pages";
+export const CREATE_NEW_DISCUSSION_TITLE = "Create new post";
+export const CREATE_NEW_DISCUSSION_TOPIC =
+  "Create a new discussion topic - e.g. My favourite Amsterdam moment is when...";
```
lucaslcode:
```suggestion
  "Create a new discussion topic";
```

app/frontend/src/features/communities/discussion/CreateDiscussionForm.stories.tsx
```diff
@@ -0,0 +1,63 @@
+import { Meta, Story } from "@storybook/react";
+import { mockedService } from "stories/__mocks__/service";
+
+import CreateDiscussionForm, {
+  CreateDiscussionFormProps,
+} from "./CreateDiscussionForm";
+
+export default {
+  component: CreateDiscussionForm,
+  title: "Communities/discussion/CreateDiscussionForm",
```
lucaslcode:
```suggestion
  title: "Communities/Discussions/CreateDiscussionForm",
```

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+import makeStyles from "utils/makeStyles";
+
+import {
+  CREATE_NEW_DISCUSSION_TITLE,
+  CREATE_NEW_DISCUSSION_TOPIC,
+  NEW_DISCUSSION_TITLE,
+  NEW_DISCUSSION_TOPIC,
+  POST,
+} from "../constants";
+import { CreateDiscussionInput, useNewDiscussionMutation } from "../hooks";
```
lucaslcode, line 16: How strict are we being with relative imports? I was only doing same folder

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+    },
+    display: "flex",
+    justifyContent: "flex-end",
+  },
+}));
+
+export interface CreateDiscussionFormProps {
+  communityId: number;
+  onCancel?(): void;
+  onPost?(): void;
```
lucaslcode: I did not know this

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          inputRef={register({ required: true })}
+          label={NEW_DISCUSSION_TOPIC}
+          multiline
+          placeholder={CREATE_NEW_DISCUSSION_TOPIC}
+          rows={6}
+        />
+        <div className={classes.actionButtonsContainer}>
+          <Button onClick={handlePost} type="submit">
+            {POST}
+          </Button>
```
lucaslcode: Needs loading state.
Also it's a bit opaque that clicking this calls `onSubmit` and `handlePost`. Maybe just call `onPost` in `onSubmit`?
That will also stop the form closing if you push this button without required fields

app/frontend/src/features/communities/discussion/DiscussionsListPage.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import { Meta, Story } from "@storybook/react";
+import { mockedService } from "stories/__mocks__/service";
+import community from "test/fixtures/community.json";
+import discussions from "test/fixtures/discussions.json";
+
+import DiscussionsListPage from "./DiscussionsListPage";
+
+export default {
+  component: DiscussionsListPage,
+  title: "Communities/discussion/DiscussionsListPage",
```
lucaslcode:
```suggestion
  title: "Communities/Discussions/DiscussionsListPage",
```

lucaslcode (changes requested):
Nice progress and thorough testing as usual

lucaslcode:
I'm trying to think of a clever way to have newly created discussions transition into the list. I came up with this:
```
const AnimatedMount = ({ children: ReactNode }) => {
  const [visible, setVisible] = useState(false);
  useEffect(() => setVisible(true), []);
  return <Collapse in={visible}>{children}</Collapse>;
};
```
Which works, but the problem is the original list all animate too. Any ideas?

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
darrenvong: I've been going with if it's one level (either `"./*` or `"../*"`, replace * with actual file), then it's ok, as it's still a bit annoying to do a full absolute import for things that are "nearby enough". What do you think?

app/frontend/src/features/communities/constants.ts
darrenvong: Maybe I should rename this to make clearer it is placeholder text, so I think the extra text here is fine as a prompt for people on the kind of things they should write (copied from the design)? Or do you think it's still unnecessary?

darrenvong:
> I'm trying to think of a clever way to have newly created discussions transition into the list. I came up with this:
> 
> ```
> const AnimatedMount = ({ children: ReactNode }) => {
>   const [visible, setVisible] = useState(false);
>   useEffect(() => setVisible(true), []);
>   return <Collapse in={visible}>{children}</Collapse>;
> };
> ```
> 
> Which works, but the problem is the original list all animate too. Any ideas?

Do you mean only have the new item animate in, but leave the rest fully shown as before? If so, don't have a good suggestion on top of my head at the moment without some more thoughts!

darrenvong:
http://reactcommunity.org/react-transition-group/transition-group

This kinda look promising as to what you're after? Material UI uses React Transition Group under the hood for these components so if we take care to use the same version, it shouldn't add another dependency to what we got already in the `yarn.lock` file if we need to use something explicitly from the lib itself.

app/frontend/src/features/communities/hooks.ts
darrenvong: Haven't a lot of time to test this yet, but since this is an infinite query, I wonder if we get any weird UI and/or data consistency problems temporarily if we do a "naive" optimistic update and put an extra element in the latest/last page when it should have been the next page, since React Query will then correct it soon after when it refetches the query again.

Otherwise the logic might be a bit more involved and requires checking for page boundaries etc which could get a bit fiddly. Will have to experiment and see I guess...

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
darrenvong: Good point. I wonder if it's better to collapse the form in `onSuccess` instead, since even calling in `onSubmit` closes the form too quickly for us to see the loading state

app/frontend/src/features/communities/constants.ts
lucaslcode: I thought I didn't see the example text in the design in the new discussion box...

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
lucaslcode: Yeah ok

app/frontend/src/features/communities/discussion/CreateDiscussionForm.tsx
lucaslcode: Yep. Actually, if this is just being used for closing, maybe the prop should just be `onClose` like MUI

darrenvong: I think I'm going to rename it as `onPostSuccess` and call it in `onSuccess` in the hook, since it just happens that successful posting closes the form. I think it's not unreasonable to expect we _might_ have to something extra given design keeps on tweaking this area still... so probably doesn't hurt to keep it semi-generic.

app/frontend/src/features/communities/hooks.ts
darrenvong: ^ This is exactly as I feared and is non-trivial to do an optimistic update as soon as there are more than one page of discussions for a community. This is because each new discussion will need to correctly put the new discussion at the beginning, move the last discussion on the first page to the second page and repeat this array shunting. And since the `discussionList` itself is nested within an object for each `pages`, it's a bit of a pain to do and not quite as simple as merging arrays together...

If we don't do that and do it the naive way I mentioned above, the eagled eye users will notice a temporary data inconsistency where the UI shows `pageSize + 1` discussions temporarily, then the last one will disappear and get cut off to the next page once the invalidated query gets refetched and corrected.

Not sure if it's worth the effort and it might be easier to put up a "fetch" loading state on this first pass.

lucaslcode: Yeah I think that's fine, optimistic updates aren't essential, it's more for the leading state do using `isFetching` is probably enough

app/frontend/src/features/communities/discussion/DiscussionsListPage.tsx
```diff
@@ -78,10 +79,11 @@ export default function DiscussionsListPage({
         <CreateDiscussionForm
           communityId={community.communityId}
           onCancel={() => setIsCreatingNewPost(false)}
-          onPost={() => setIsCreatingNewPost(false)}
+          onPostSuccess={() => setIsCreatingNewPost(false)}
         />
       </Collapse>
       <div className={classes.discussionsContainer}>
+        {!isDiscussionsLoading && isDiscussionsFetching && <CircularProgress />}
```
lucaslcode: Can this be moved to next to the "New post" button? That way it won't cause content shifting.

We could also only display this if `isFetchedAfterMount` is true, then it will (I think?) only be displayed on invalidation and not clicking on the page. But maybe we don't want that behaviour anyway.

lucaslcode (changes requested):
Last thing

app/frontend/src/features/communities/discussion/DiscussionsListPage.tsx
darrenvong: Tried `isFetchedAfterMount` and didn't seem to be work, it stayed as true even when it's fetching after invalidation. Moved the loading state next to the "new post" button and yeah definitely less jumping around now

lucaslcode (approved):

Issue #1082, opened 2021-04-14
Error boundary


aapeliv:
As per last frontend meeting, we should set up an error boundary for react so people don't just see blank white pages. It might be good to set up a separate error boundary for the map component? I'm not 100% sure how this all works though!

thdk:
I can try to set this up, any special requirements?

lucaslcode:
Maybe offer the options of either refreshing the page or going to the homepage.

thdk:
Something like this?
![image](https://user-images.githubusercontent.com/4201102/115160011-41897c80-a096-11eb-913f-a2f60ac6289e.png)

![image](https://user-images.githubusercontent.com/4201102/115159952-08e9a300-a096-11eb-8b67-1832ea6fa6e8.png)

lucaslcode:
Great! I just think it should somehow be clearer to try refreshing first, before trying the homepage. And maybe error opens underneath the buttons?

darrenvong:
Can we not show a stacktrace on the page at all? It's not very user friendly (even though it's hidden by default) and I rather not have it so readily available on the page. If we are concerned about losing information then we should look at setting up proper error reporting instead of this.

lucaslcode:
Good point, and it will be meaningless minified stuff in production anyway

justjooshing:
@thdk @lucaslcode @darrenvong 
Closing this because it looks done

PR #1115, opened 2021-04-20
Frontend/feat/error boundary

thdk:
Let me know if tests would be needed for some part of this PR.

I didn't find an api endpoint for logging errors to our backend, is that correct?


app/frontend/package.json
```diff
@@ -16,6 +16,7 @@
     "maplibre-gl": "^1.14.0",
     "react": "^17.0.0",
     "react-dom": "^17.0.0",
+    "react-error-boundary": "^3.1.1",
```
lucaslcode, line 19: I'm debating with myself if it's worth adding a dependency for this.

The pro is our code can be cleaner, but the con is we are then relying on this package being maintained. Considering we are hoping to support the app long-term, we shouldn't add dependencies too lightly, and it's not that difficult to manually do an [error boundary component](https://reactjs.org/docs/error-boundaries.html).

What do you think?

app/frontend/src/AppRoute.tsx
```diff
@@ -46,6 +47,11 @@ export default function AppRoute({
 
   const classes = useStyles();
 
+  const RouteContent = useMemo(
+    () => () => <ErrorBoundary>{children}</ErrorBoundary>,
+    [children]
+  );
```
lucaslcode: I don't think there's any benefit to memoizing this.

lucaslcode: I see, you switched to `component` instead of `render` below. If you do `render={() => <RouteContent>{children}</RouteContent>`, you won't need to memoize.

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,17 @@
+import ErrorFallback from "components/ErrorFallback";
+import { PropsWithChildren, useCallback } from "react";
+import { ErrorBoundary as ReactErrorBoundary } from "react-error-boundary";
+
+export default function ErrorBoundary({
+  children,
+}: PropsWithChildren<unknown>) {
```
lucaslcode: `ReactNode` is the correct type for children usually.

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const onError = useCallback(() => {
+    // TODO: log error to backend?
+  }, []);
```
lucaslcode: remove for now, we can put it in when we have proper error reporting

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,51 @@
...
+    justifyContent: "flex-end",
+    marginTop: theme.spacing(2),
+  },
+  report: {
+    marginTop: theme.spacing(2),
+  },
+}));
+
+export default function ErrorFallback() {
+  const { actions, report } = useStyles();
```
lucaslcode: We prefer not destructuring for `classes`

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const history = useHistory();
+
+  const handleRefresh = useCallback(() => history.go(0), [history]);
```
lucaslcode: No need to memoize simple functions

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <>
+      <PageTitle>{ERROR_HEADING}</PageTitle>
+      <Typography variant="body1">{ERROR_INFO}</Typography>
+      <Box className={report}>
```
lucaslcode: We used to use `Box` and there are probably some lying around, but we use `div` now

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <BugReport isResponsive={false} />
+      </Box>
+
+      <CardActions className={actions}>
```
lucaslcode: Card actions should only be in a card

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <NavLink exact to={{ pathname: baseRoute }}>
+          <Button variant="outlined">{GO_TO_HOMEPAGE}</Button>
+        </NavLink>
```
lucaslcode: 1. Use `Link` (make sure it's from `react-router-dom` not mui), `NavLink` is used where you need an "active" class.
2. This isn't valid html (button inside anchor).  use `<Link component={Button} ...`

app/frontend/src/components/ErrorFallback/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const ERROR_HEADING = "Ooops... that's an error";
+export const ERROR_INFO =
+  "Please use the 'Report a problem' button to help us prevent this error in the future.";
```
lucaslcode:
```suggestion
export const ERROR_HEADING = "Ooops... that's an unknown error";
export const ERROR_INFO =
  "Please use the 'Report a problem' button and tell us what you were doing to help us prevent this error in the future.";
```

lucaslcode (changes requested):
Thanks! 👍🏻

app/frontend/src/AppRoute.tsx
thdk: Ok I drop the useMemo. 
`const RouteContent = () => <ErrorBoundary>{children}</ErrorBoundary>;`

Not sure if I understand though if there is a difference between:
`<Route {...otherProps} render={() => <RouteContent />} />`
and
`<Route {...otherProps} component={RouteContent} />`

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
thdk: I've learned to use PropsWithChildren because the type of children can change internally in React (unlikely I know) and this code would still have correct typings. Both of them are correct to my knowledge.

If you prefere `children: ReactNode`, that's ok :)

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
thdk: Is there any other component I can use for this cause? I went looking in the designs to find something similar but didn't see much except for 'NewHostRequest.tsx' which also uses CardActions. But it should be only used within a card indeed.

I'll change it here, and maybe also on NewHostRequest?

thdk: nm I see NewHostRequest itself is being wrapped in a Card component by it's parent.
I've created an Actions component that wraps these buttons when used outside of a Card. If that's too much I can move the styling inside the ErrorFallback component.

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
thdk: except that when using PropsWithChildren type, the children property is optional while here for ErrorBoundary component the children are required.

app/frontend/package.json
thdk: It's a small dependency with many weekly downloads and no other dependencies except for @babel/runtime which I raised a question for on the react-error-boundary repo to know why they have that dependency.
Making your own error boundary is a possibility and is fairly simple indeed but doesn't mean it won't need maintenance in the future. I would say that, if necessary, we can make the ErrorBoundary ourselves once we notice problems with maintenance of react-error-boundary.

app/frontend/src/AppRoute.tsx
```diff
@@ -46,6 +47,8 @@ export default function AppRoute({
 
   const classes = useStyles();
 
+  const RouteContent = () => <ErrorBoundary>{children}</ErrorBoundary>;
```
lucaslcode: Sorry, I'm an idiot, components in a render function do need to be memoized. Better still would be moving it into a separate function. That way it doesn't get recreated and also doesn't have the (small) overhead of memoizing.

lucaslcode: Alternatively, I'd also consider moving the error boundary to `App.tsx`, plus that way it will help in case there's an error in the navigation or theme.

app/frontend/src/components/Actions.tsx
```diff
@@ -0,0 +1,19 @@
...
+    marginTop: theme.spacing(2),
+    justifyContent: "flex-end",
+    padding: theme.spacing(1),
+    "& :not(:first-child)": {
+      marginLeft: theme.spacing(1),
+    },
+  },
+}));
+
+export default function Actions({ children }: { children: ReactNode }) {
```
lucaslcode: Could you also add this to storybook by making an `Actions.stories.tsx` file?

lucaslcode (changes requested):

app/frontend/src/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <Typography variant="body1">{ERROR_INFO}</Typography>
+      <div className={classes.report}>
+        <BugReport isResponsive={false} />
+      </div>
+
+      <Actions>
+        <Link
+          to={{ pathname: baseRoute }}
+          component={() => <Button variant="outlined">{GO_TO_HOMEPAGE}</Button>}
+        />
```
lucaslcode:
```suggestion
        <Link
          to={{ pathname: baseRoute }}
          component={Button}
          variant="outlined"
        >
          {GO_TO_HOMEPAGE}
        </Link>
```

lucaslcode: Wait! I think this is also wrong

lucaslcode:
```suggestion
        <Button
          to={{ pathname: baseRoute }}
          component={Link}
          variant="outlined"
        >
          {GO_TO_HOMEPAGE}
        </Button>
```
Try that. Check the console for errors

app/frontend/src/AppRoute.tsx
thdk: Not sure if we need to move the ErrorBoundary from AppRoute to App level. Instead we can add an additional ErrorBoundary to App.

When a Route throws an error, it will still show the top navigation bar and the error will be reset when user navigates to another page. While if we just have a root ErrorBoundary, all the user will see is the ErrorFallback component.

Does that makes sense?

app/frontend/src/components/Actions.tsx
thdk: yes, shall do :)

app/frontend/src/AppRoute.tsx
lucaslcode: Yes good point

thdk (commented):
Somehow I started a review, and I can't find a button to cancel this :(

thdk:
Fixes #1082

app/frontend/src/components/Actions/Actions.tsx
```diff
@@ -0,0 +1,19 @@
...
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    display: "flex",
+    marginTop: theme.spacing(2),
+    justifyContent: "flex-end",
+    padding: theme.spacing(1),
+    "& :not(:first-child)": {
+      marginLeft: theme.spacing(1),
```
lucaslcode:
```suggestion
      marginInlineStart: theme.spacing(1),
```
`marginInlineStart` and `marginInlineEnd` (but *not* `marginInline`) have good browser compatibility these days, so we have switched to using them in case we ever want to support RtL languages (though there will probably still be plenty of other stuff to fix up!)

lucaslcode: You could also do `marginBlockStart` instead of `marginTop` if you want, but I don't think it's that important.

app/frontend/src/App.tsx
```diff
@@ -14,13 +17,17 @@ function App() {
...
-          <AuthProvider>
-            <CssBaseline />
-            <EnvironmentBanner />
-            <AppRoutes />
-          </AuthProvider>
-        </ReactQueryClientProvider>
+        <FatalErrorBoundary fallbackRender={() => <ErrorFallback isFatal />}>
+          <ReactQueryClientProvider>
+            <AuthProvider>
+              <ErrorBoundary>
```
lucaslcode: Why this one as well as the fatal one? It also seems a bit odd to have our component `ErrorBoundary` but also use the one from the package directly.

lucaslcode (changes requested):
Looks great! One or two very minor nits

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -1,15 +1,22 @@
...
+  ErrorBoundary as ReactErrorBoundary,
+  ErrorBoundaryProps,
+} from "react-error-boundary";
 import { useLocation } from "react-router";
 
-export default function ErrorBoundary({ children }: { children: ReactNode }) {
+export default function ErrorBoundary({
+  children,
+  ...otherProps
+}: PropsWithChildren<Partial<ErrorBoundaryProps>>) {
```
lucaslcode, line 12: Doesn't `ErrorBoundaryProps` already contain `children`? So maybe `PropsWithChildren` isn't needed.

lucaslcode (approved):
1 minor thing, if I'm incorrect feel free to merge.

app/frontend/src/components/ErrorBoundary/ErrorBoundary.tsx
thdk:
```javascript
declare type ErrorBoundaryProps = ErrorBoundaryPropsWithFallback | ErrorBoundaryPropsWithComponent | ErrorBoundaryPropsWithRender;
declare type ErrorBoundaryState = {
    error: Error | null;
};
declare class ErrorBoundary extends React.Component<React.PropsWithRef<React.PropsWithChildren<ErrorBoundaryProps>>, ErrorBoundaryState> {
```

thdk: These props are also being wrapped by the ErrorBoundary component (the one from react-error-boundary) using `PropsWithChildren`. The type `ErrorBoundaryProps` itself doesn't have the children prop yet.

Issue #1090, opened 2021-04-16
Gaps in layout of My connections page on desktop


CouchersBot:
Subject: Gaps in layout of My connections page on desktop
Description:
- Have one pending friend request
- Send a friend request

Results:
I see three lists on the my connections page:
1. my friends
2. friend requests
3. friend requests you sent

Expected: three lists should nicely use the available space on the screen. Using two columns on wide screens. Reducing to one columns on narrow screens.

Actual: gaps exist between the three list blocks (see screenshot)

Backend version: develop-a0c66b6e
Frontend version: develop-a0c66b6e-next
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.72 Safari/537.36
Page: https://develop--frontend.preview.couchershq.org/connections
User (spoofable): thdk (43)

![image](https://user-images.githubusercontent.com/4201102/114986252-120b2200-9e94-11eb-9389-fbbcfbfaed13.png)

![image](https://user-images.githubusercontent.com/4201102/114987366-5f3bc380-9e95-11eb-94eb-d405b9bc4380.png)

PR #1119, opened 2021-04-22
Fix weird gaps on connections page

justjooshing:
A quick fix for [#1090](https://github.com/Couchers-org/couchers/issues/1090) by 100% height to child divs means that they take the height of their respective row, whether on mobile or desktop

image
image

I'm not sure having friends and requests on the same tab is the best idea, because eventually you'll have way more friends and it'll look really awkward
image

The design suggests having Recently Added and Friend Requests as separate tabs (which I'm guessing are the same as our Sent/Received sections), but I couldn't figure out how to do that, and figured it's not the highest priority to spend too much time digging around

### DIFF
--- a/app/frontend/src/features/auth/useAuthStyles.ts
+++ b/app/frontend/src/features/auth/useAuthStyles.ts
@@ -1,7 +1,6 @@
 import makeStyles from "utils/makeStyles";
 
 import DesktopAuthBg from "./resources/desktop-auth-bg.jpg";
-import MobileAuthBg from "./resources/mobile-auth-bg.jpg";
 
 const useAuthStyles = makeStyles((theme) => ({
   button: {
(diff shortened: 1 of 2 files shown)


lucaslcode (approved):

darrenvong:
> I'm not sure having friends and requests on the same tab is the best idea, because eventually you'll have way more friends and it'll look really awkward
> image
> 
> The design suggests having Recently Added and Friend Requests as separate tabs (which I'm guessing are the same as our Sent/Received sections), but I couldn't figure out how to do that, and figured it's not the highest priority to spend too much time digging around

Hmm yeah I'm not too sure I like the look of that and if it's "tidier" than before to have this 🤔 

The whole friends page needs redoing anyway so I wonder if it's best to leave it as it is, as I think the "mosaic" like layout before would look less odd when you have more friends than friend requests

justjooshing:
Yep @darrenvong I agree. Although it'll always look weird when you have more requests sent than your number of friends, I assume that'll be the least common situation as soon as people add more than a handful of connections
Happy to close this without merge and then look into the tabs later down the line

darrenvong:
I didn't read into this issue properly... the gap does look more broken than this actually, so let's merge it still, but yeah this still look crap but it'll be redesigned so doesn't matter for now.

Issue #1079, opened 2021-04-13
Limit size of images on emails


ItsiW:
A very lovely photo of Jakub, but perhaps it doesn't need to be in such detail to extend over the margins of the email.

![image](https://user-images.githubusercontent.com/39325627/114479366-c2ceb280-9c43-11eb-90a5-814b7c3766f9.png)

aapeliv:
Should be the thumbnail size, not full size. See also #871

alexishonzik:
I'll work on this!

alexishonzik:
We noticed in friend_request.md in backend/templates that the full_url is used. We were intending on changing this to the thumbnail_url, however we realized that this will cause the tests in backend/tests/test_email.py to fail because of the assertions for making sure the full_url is in the HTML (Line 150). Assuming we are understanding this correctly, can we change the tests to assert that the thumbnail URL is in the HTML instead of the full_url? Thanks!

lucaslcode:
Yes! Tests are for guarding against unintentional changes - if that's what you mean to change, it's normal to alter the test too to reflect that

lucaslcode:
Partly closed in #1120, rest in #871

PR #1120, opened 2021-04-22
fixing issue #1079. making avatar the thumbnail size instead of full size

laurenfalvo:


lucaslcode (changes requested):
Thanks! While you're at it, can you do the same for host requests? (And the tests?)

laurenfalvo:
I've now updated the host request and tests as well !

lucaslcode (approved):
Thanks!

PR #1122, opened 2021-04-23
Use GetHostRequest to standardize query and simplify testing

ShmuelTreiger:
Realized a bunch of tests were missing. Also realized this was a simpler solution


app/backend/src/couchers/servicers/requests.py
```diff
@@ -136,7 +136,7 @@ def CreateHostRequest(self, request, context):
 
             return requests_pb2.CreateHostRequestRes(host_request_id=host_request.conversation_id)
 
-    def GetHostRequest(self, request, context):
+    def GetHostRequest(self, request, context, pb=True):
```
aapeliv: Don't do this, please. It's confusing, this does two very different things now which is not good conceptually.

Just make a new helper function to grab the host request if you want to

aapeliv (changes requested):

ShmuelTreiger:
Testing _get_host_request_query directly would require faking context and I'm not sure how to do that. Instead I tested it indirectly, through GetHostRequest, but only once, instead of for each api call which uses the query.

app/backend/src/couchers/servicers/requests.py
```diff
@@ -55,6 +55,30 @@ def message_to_pb(message: Message):
         )
 
 
+def _get_host_request_query(context, host_request_id):
+    with session_scope() as session:
```
aapeliv, line 59: pass in the `session` as a variable so you don't need to do the `merge` stuff

aapeliv (changes requested):

ShmuelTreiger:
I should have changed this from merge, as whatever I do in #1163 will largely impact what I do here. I'm fact, I'll probably scrap this branch for that one.

PR #1124, opened 2021-04-24
[WIP] Email notifications for conversations in host requests

### LINKED ISSUE #915: Email notifications for conversations in host requests
Seems the email notifications don't include messages in host requests.

ShmuelTreiger:
First attempt. Untested. Just looking for feedback/thoughts on my approach.

Other potential approaches:
1) This approach in one query, joining User to the initial subquery 3 times with a bunch of filtering
2) Starting from Users and joining messages 3 times with a *ton* of filtering
3) Collect users and host requests in its own separate query and merge the results before sending out the email (very ugly and messy)
4) Create a whole new background job for hostrequest notifications with its own, separate email


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -48,69 +61,88 @@ def process_purge_signup_tokens(payload):
 
 def process_send_message_notifications(payload):
     """
-    Sends out email notifications for messages that have been unseen for a long enough time
+    Sends out email notifications for groupchat messages that have been unseen for a long enough time #FIX
```
aapeliv, line 64: What's the `#FIX`?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -48,69 +61,88 @@ def process_purge_signup_tokens(payload):
...
+            .join(Conversation, Conversation.id == Message.conversation_id)
             .filter(Message.time < now() - timedelta(minutes=5))
-            .filter(Message.message_type == MessageType.text)  # TODO: only text messages for now
-            .all()
+            .filter(Message.time > now() - timedelta(minutes=10))
+            .filter(Message.id > User.last_notified_message_id)
+            .group_by(Message.conversation_id)
+            .group_by("user_id")
+            .subquery()
         )
```
aapeliv, line 102: This query is way too confusing.

It seems is not stateless/idempotent: if we fail to run this for 15 minutes, some notifications will not be delivered?

Also `Hosts`/`Guests` should be lowercase and don't seem to be very descriptive words.

aapeliv (commented):
I think you should do 4: make a separate background job. They're sufficiently different and othewrise the code just looks like a mess.

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: A note to myself to fix it that I forgot to take out 😂

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: Yes, it's a confusing query, which is why I stopped to get feedback.

>It seems is not stateless/idempotent: if we fail to run this for 15 minutes, some notifications will not be delivered?

That's fixed by taking out the >10 minutes filter.

>don't seem to be very descriptive words.

No, I would have renamed them at some point

aapeliv:
Closed through #1302

PR #1125, opened 2021-04-24
chore: disable environment banner in dev

darrenvong:
Short PR to disable the intrusive environment banner in development so I don't have to keep dismissing it...

### DIFF
--- a/app/frontend/src/components/EnvironmentBanner.tsx
+++ b/app/frontend/src/components/EnvironmentBanner.tsx
@@ -23,25 +23,17 @@ export function EnvironmentBanner() {
   const theme = useTheme();
   const classes = useStyles();
   const isBelowSm = useMediaQuery(theme.breakpoints.down("sm"));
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/EnvironmentBanner.tsx
```diff
@@ -25,17 +25,13 @@ export function EnvironmentBanner() {
   const isBelowSm = useMediaQuery(theme.breakpoints.down("sm"));
   const [isHidden, setIsHidden] = useState(false);
```
thdk: `const [isHidden, setIsHidden] = useState(process.env.REACT_APP_COUCHERS_ENV !== "preview");`

darrenvong: Nice suggestion, actually this gives a good opportunity to simplify the render logic too and I flipped it so it reads more naturally (true renders, false renders nothing)

thdk (approved):
Much better!

PR #1126, opened 2021-04-24
chore: rename storybook mock file

darrenvong:
Another small PR to fix the following issue/warning I noticed in our CI recently:

```
jest-haste-map: duplicate manual mock found: service
  The following files share their name; please delete one of them:
    * <rootDir>/src/stories/__mocks__/service.ts
    * <rootDir>/src/__mocks__/service.ts
```

I suspect this might be what's causing tests to fail on some of our machines locally (yet magically CI seems to know to pick the last one 🤷 ).

Since there are no good reasons for the Storybook mocks to be placed inside a `__mocks__` folder and it clashes with jest, I renamed it so it doesn't get picked up by jest during tests. Doing this meant I had to rename quite a few imports (luckily all auto picked up by VS Code).


app/frontend/src/features/connections/friends/AddFriendButton.stories.tsx
```diff
@@ -2,7 +2,7 @@ import { action } from "@storybook/addon-actions";
 import { Meta, Story } from "@storybook/react";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 
-import { mockedService } from "../../../stories/__mocks__/service";
+import { mockedService } from "../../../stories/serviceMocks";
```
thdk: import from "stories/serviceMocks" ?

darrenvong: Good catch 👍  The guilt of relying on auto import fix too much, it's not that smart sometimes 😅

thdk (approved):

PR #1127, opened 2021-04-25
test: fix timezone issues in datepicker tests

thdk:


lucaslcode (approved):
LGTM. If you're up for it, can you add a regression test? As in, whatever test was failing before, add another test which uses a previously failing timezone.

lucaslcode:
Unless it's just a test bug, in which case I guess there's no difference.

thdk:
I thought it was just a test bug, but now you made me doubt.. :)

PR #1128, opened 2021-04-25
Backend/bug/coordinate wrapping

### LINKED ISSUE #835: Coordinate wrapping
Currently the database will accept coordinates with `lat` outside `[-180,180]`, which is problematic, this requires a bit of inverstigation.

The best would be to enforce this in the database somehow

spreeni:
Closes #835 

When entering a geometry point with `create_coordinate`, it is now cast to geography type and back to ensure longitude wrapping. 

If a latitude is given outside of the valid range [-90, 90], `create_coordinate` will instead throw a `ValueError`. This was done as PostGIS wraps latitude wackily, e.g. (40, 95) -> (40, 85) instead of (-120, 85).

Added a test case in `test_api.py` and a migration script that does the above mentioned back-and-forth casting.


spreeni:
Note: We still have issues in the Frontend if somebody wants to set their location to e.g. the North or South Pole. The displaced longitude for a latitude of 90 or -90 will be infinitely large. And in general the calculated randomized location and circle points can again have invalid coordinates for a valid starting point. Maybe it makes sense to implement a guard here.

Steps to reproduce: Search for "North Pole" or "South Pole" in MapBox

aapeliv:
Just doing the cast seems to work for me? This is from postgres:

```
postgres=# select st_asewkt(st_makepoint(0, 95)::geography::geometry);
NOTICE:  Coordinate values were coerced into range [-180 -90, 180 90] for GEOGRAPHY
       st_asewkt
-----------------------
 SRID=4326;POINT(0 85)
(1 row)
```

I think this is fine: if someone does put "North Pole", at least this will be close? Or am I missing something?

Also, please add full regression tests by just creating a list of pairs of ((inpux_lat, input_lng),(output_lat,output_lng)) and running them all through the database directly to double check the output is correct for a set of inputs!

spreeni:
The cast works fine as a guard, but the results for latitude wrapping were a bit counter-intuitive to me. If I would go further North and cross the North Pole, I would expect to at some point be on the other half of the globe in terms of longitude. But it is probably right that this should be enough and close enough to the real location in most situations. I'll remove the check for a ValueError then.

And I can add the regression tests, no problem. Is there a convention where I should insert them? And would you test the SQL query directly or the use of `create_coordinate`? For a direct SQL query I would just add it in `test_db.py`, right?

aapeliv:
Yeah, `test_db.py` is good. You need to use `create_coordinate` which creates SQL so you need to run it through the database. Let me know on slack if you have any issues.

spreeni:
Removed the Exception throw and added regression tests. One weird thing that PostGIS does is that it generally won't wrap a longitude of -180, but will wrap it to 180 if there is a latitude overflow.

app/backend/src/tests/test_db.py
```diff
@@ -87,6 +88,40 @@ def test_get_parent_node_at_location(testing_communities):
...
+            for lng, lng_expected in longitudes:
+                latitude, longitude = get_coordinates(session.query(create_coordinate(lat, lng)).scalar())
+
+                assert latitude == lat_expected
+
+                # Weird interaction in PostGIS where lng flips at -180 only when there is latitude overflow
+                if lng == -180 and (lat > 90 or lat < -90):
+                    assert longitude == -lng_expected
+                else:
+                    assert longitude == lng_expected
```
aapeliv: Just make a list of `lat,lng,lat_exp,lng_exp` and test them with

```py
assert lat_exp, lng_exp == get_coordinates(...)
```

Something like this. Having this logic in the test is not good, now `lng_expected` is not really the expected lng.

app/backend/src/couchers/migrations/versions/e264cbf0ca07_coordinate_wrapping_for_points_in_users_.py
```diff
@@ -0,0 +1,38 @@
...
+depends_on = None
+
+
+def upgrade():
+    """
+    Cast points to geography and back to wrap coordinates
+    """
+    op.execute("""
+    UPDATE users
+    SET geom = (ST_SetSRID( geom, 4326)::geography )::geometry
```
aapeliv: Remove extra space in front of `geom`

aapeliv (changes requested):

app/backend/src/tests/test_db.py
spreeni: Yeah, thought about this, thought it might be ok with the comment. 

Changed it now to a few fixed coordinates.

app/backend/src/couchers/migrations/versions/f4a49acd8801_coordinate_wrapping_for_users_events_.py
```diff
@@ -0,0 +1,44 @@
...
+def upgrade():
+    """
+    Cast points to geography and back to wrap coordinates
+    """
+    op.execute(
+        """
+    UPDATE users
+    SET geom = (ST_SetSRID(geom, 4326)::geography)::geometry
+    """
+    )
```
aapeliv: You can make this single lines and it'll format them much nicer, but not critical

aapeliv (dismissed):
LGTM!

app/backend/src/couchers/migrations/versions/f4a49acd8801_coordinate_wrapping_for_users_events_.py
spreeni: True, changed this.

aapeliv (approved):

PR #1130, opened 2021-04-25
[MRG] Properly escape slugify in the migration

kalvdans:
So that it treats dash as a punctation character.


aapeliv (approved):
Looks good, thanks for spotting this!

PR #1132, opened 2021-04-25
Revert "[MRG] Properly escape slugify in the migration"

kalvdans:
Reverts Couchers-org/couchers#1130 due to failing CI tests


kalvdans:
I clicked merge too quick before checking CI results, can you approve my revert @aapeliv and I'll redo the change properly.

aapeliv (approved):

PR #1134, opened 2021-04-25
[MRG] Properly escape slugify in the migration take 2

kalvdans:
So that it treats dash as a punctation character. Re-applies #1130

### DIFF
--- a/app/backend/src/couchers/migrations/versions/2d656b6ad999_fix_slugify.py
+++ b/app/backend/src/couchers/migrations/versions/2d656b6ad999_fix_slugify.py
@@ -23,7 +23,7 @@ def upgrade():
     # e.g. slugify('Detta är ett test!') -> detta-ar-ett-test
     op.execute(
         r"""
-    REPLACE FUNCTION slugify("text" TEXT)
+    CREATE OR REPLACE FUNCTION slugify("text" TEXT)
     RETURNS TEXT AS $$
     SELECT regexp_replace(
       regexp_replace(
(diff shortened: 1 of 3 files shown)


aapeliv (dismissed):
LGTM!

aapeliv:
The rule should probably be "unaccent, lowercase, then replace every char in [^0-9a-z] with `-`, remove repeated and starting/ending `-`s". Currently I get this:

```
postgres=# select slugify('this is_a  ------  test');
        slugify
-----------------------
 this-is_a--------test
(1 row)
```

It should just be `this-is-a-test`, what do you think @kalvdans ?

kalvdans:
> The rule should probably be "unaccent, lowercase, then replace every char in [^0-9a-z] with `-`, remove repeated and starting/ending `-`s". Currently I get this:
> 
> ```
> postgres=# select slugify('this is_a  ------  test');
>         slugify
> -----------------------
>  this-is_a--------test
> (1 row)
> ```
> 
> It should just be `this-is-a-test`, what do you think @kalvdans ?

I agree, let me look into why `"  ------  "` doesn't match a single `[^0-9a-z]+`, might be something with greediness of the patterns.

aapeliv:
> I agree, let me look into why `" ------ "` doesn't match a single `[^0-9a-z]+`, might be something with greediness of the patterns.

Sorry. What you have now works. What's in the DB currently does not work. I read/misunderstood the regexp wrong.

But I think we should remove the underscore

aapeliv (approved):
LGTM

PR #1135, opened 2021-04-25
[MRG] Skip half of the tests now when migrations are proved equal to models

### LINKED ISSUE #391: Speed up CI

kalvdans:
Skip running all other migration tests after a specific test for migrations was merged in https://github.com/Couchers-org/couchers/pull/885 .

Closes #391

### DIFF
--- a/app/backend/src/tests/test_fixtures.py
+++ b/app/backend/src/tests/test_fixtures.py
@@ -68,7 +68,7 @@ def create_schema_from_models():
     Base.metadata.create_all(get_engine())
 
 
-def db_impl():
+def recreate_database():
     """
     Connect to a running Postgres database, build it using metadata.create_all()
     """
@@ -89,7 +89,7 @@ def db():
     Pytest fixture to connect to a running Postgres database and build it using metadata.create_all()
     """
 
-    db_impl()
+    recreate_database()
 
 
 def generate_user(*_, **kwargs):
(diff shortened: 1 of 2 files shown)


app/backend/src/tests/test_fixtures.py
```diff
@@ -68,37 +68,28 @@ def create_schema_from_models():
...
-        # create everything from the current models, not incrementally through migrations
-        create_schema_from_models()
+    # create everything from the current models, not incrementally through migrations
+    create_schema_from_models()
 
 
-@pytest.fixture(params=["migrations", "models"])
-def db(request):
+@pytest.fixture()
+def db():
```
aapeliv, line 87: I think this function is now unneeded, it was used somewhere in the communities tests, but I don't think this is required anymore?

I might be wrong though

kalvdans: db_impl is called from communities test, but I'm leaning towards not using a fixture but a simple Python function `recreate_database()` that all tests call in the beginning of the test function. What do you think? Since there's no teardown code we don't win anything by using a fixture.

aapeliv: Why not use fixtures?

kalvdans: They are unnecessary magical. Nothing beats pure Python code when checking what the tests do.

aapeliv: Yeah fair enough. As long as we feel we won't need that magic, then why not?

kalvdans: We might want to add magic in the future, so let's keep it as a fixture for now :)

aapeliv (approved):
LGTM

PR #1142, opened 2021-04-26
blank when no options

Jesus82:
I left it when there are no friends, let me know if you want me to add any translation key.

### DIFF
--- a/app/frontend/src/features/messages/groupchats/CreateGroupChat.tsx
+++ b/app/frontend/src/features/messages/groupchats/CreateGroupChat.tsx
@@ -14,6 +14,7 @@ import { AddIcon } from "components/Icons";
 import TextField from "components/TextField";
 import useFriendList from "features/connections/friends/useFriendList";
 import {
+  COULDNT_FIND_ANY_FRIENDS,
   CREATE,
   ERROR_USER_LOAD,
   FRIENDS,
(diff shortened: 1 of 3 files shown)


lucaslcode:
Yes I think it does need a string. Maybe "Couldn't find any friends". 
Also make sure to run `yarn format` or set up your IDE to format on save (should be double quote marks)

app/frontend/src/features/messages/groupchats/CreateGroupChat.tsx
```diff
@@ -148,7 +148,7 @@ export default function CreateGroupChat({ className }: { className?: string }) {
                     multiple={true}
                     loading={friends.isLoading}
                     options={friends.data ?? []}
-                    noOptionsText={''}
+                    noOptionsText={"Couldn't find any friends"}
```
lucaslcode: Could you put it in the nearest `constants.ts`?

lucaslcode (approved):

Jesus82:
Hi, I see that 1 check failed in the pipeline, but the PR was merged. Should I delete the branch?

lucaslcode:
The failed part was just some coverage calculator, looks like a one-off server error.
Since the PR is merged now, I've deleted the branch (would need a revert if there was a problem).

PR #1145, opened 2021-04-26
[MRG] Removing mutual friends from user proto message

### LINKED ISSUE #1143: Remove mutual friends from user API call
I haven't benchmarked it, but I'm pretty sure the mutual friends join is one of the heaviest things in the API call that gets the user.

I think we should remove it from that RPC, and just have the frontend can grab mutual friends from the dedicated RPC when needed.

What do you think @lucaslcode and @ShmuelTreiger?

ShmuelTreiger:


app/backend/src/tests/test_api.py
```diff
@@ -489,21 +488,6 @@ def test_ListMutualFriends(db):
...
-    user2, token2 = generate_user()
-    user3, token3 = generate_user()
-    user4, token4 = generate_user()
-    make_friends(user1, user2)
-    make_friends(user1, user3)
-    make_friends(user4, user2)
-    make_friends(user4, user3)
-
-    with api_session(token1) as api:
-        user4_from_db = api.GetUser(api_pb2.GetUserReq(user=user4.username))
```
aapeliv, line 503: Just change this to `ListMutualFriends` instead of removing the whole test?

aapeliv (changes requested):
Looks good other than one comment

app/backend/src/tests/test_api.py
ShmuelTreiger: There's other tests for ListMutualFriends. I wrote this test specifically to test getting it from the user protobuff message

aapeliv (approved):

PR #1148, opened 2021-04-27
Change landing page text

ItsiW:
Changing stand-in text to be in line with our marketing.

Potentially we could put a line break between the two sentences in INTRODUCTION_SUBTITLE


darrenvong:
Probably easier to break this into another header. Or remove the dark green underline? I think having the underline looks a little odd now the message is too long

lucaslcode:
This needs to be run past design team first since the text is much longer. Marta did this page

ItsiW:
I condensed it quite a bit, hope this is fine

darrenvong (approved):

PR #1152, opened 2021-04-27
Remove empty references tab

ItsiW:


app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -111,12 +111,12 @@ export default function ProfilePage() {
...
                   <Home user={user}></Home>
                 </TabPanel>
-                <TabPanel
+                {/* <TabPanel
                   classes={{ root: classes.tabPanel }}
                   value="references"
                 >
                   <References />
-                </TabPanel>
+                </TabPanel> */}
```
darrenvong: Might be neater to hide this under a feature flag like `REACT_APP_IS_REFERENCES_ENABLED` similar to how the communities stuff are hidden rather than commenting this out.

I assume this is to work around the fact we don't have users leaving references so it'd say "no references of this kind" for everyone at the moment?

lucaslcode: I think this is fine, adding leaving references back in should be in one PR and it can be uncommented there - it's not a big interlinked thing like communities.

lucaslcode (dismissed):

app/frontend/src/features/profile/view/ProfilePage.tsx
darrenvong: True... in that case, it's probably easier to delete it. Personally not a fan of having commented out code all over the place as it adds visual clutter. If we need to recover the code, that's what Git is for since we still have it in the history.

darrenvong (approved):

PR #1153, opened 2021-04-27
Update Terms of Service

ItsiW:


app/frontend/src/components/TOS.tsx
```diff
@@ -1,4 +1,4 @@
-import { Link } from "@material-ui/core";
+// import { Link } from "@material-ui/core";
```
aapeliv, line 1: Remove, don't uncomment.

aapeliv (changes requested):
I feel like it's too verbose to have all of this on the signup page. I think this should instead be moved to a different page and linked from the signup form

aapeliv:
Superseded by other PRs: #1184 (frontend), #1185 (backend), #1170 (bump TOS version)

PR #1154, opened 2021-04-28
Start search filters

lucaslcode:


lucaslcode:
This first commit is very rudimentary, but I'm having trouble with the search API @aapeliv . `search_in_area` seems to have no effect. Have a look at the console log output in the preview, the requests seems to be correct.

Currently it ignores the query string when you include coordinates, but it was the same when the search string was also included.

aapeliv:
> `search_in_area` seems to have no effect. Have a look at the console log output in the preview, the requests seems to be correct.

Thanks for spotting that. It was indeed broken. Fixed in #1179. Sorry for taking so long to fix it.

> Currently it ignores the query string when you include coordinates, but it was the same when the search string was also included.

What do you mean by this?

lucaslcode:
Nevermind, I was just referring to my current implementation

lucaslcode:
This is ready for initial review, but I still plan on adding some tests and more options to the filter dialog.

aapeliv:
The main search box doesn't work at all for me. I can see from the grpc web tools that the query is "undefined" (the string), similarly in the URL params, I've got https://frontend-feature-search-part-2--frontend.preview.couchershq.org/search?query=undefined#loc=1/0/0. (That is, when I type in "Itsi" into the search box and press enter without touching filters)

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -58,7 +65,17 @@ export default function SearchResultsList({
...
-  const { query } = useParams<{ query?: string }>();
+  const selectedUser = useUser(selectedResult);
+
+  const location = useLocation();
+
+  const searchParams = Object.fromEntries(new URLSearchParams(location.search));
+  const query = searchParams.query;
+  const lat = Number.parseFloat(searchParams.lat) || undefined;
+  const lng = Number.parseFloat(searchParams.lng) || undefined;
+  const radius = 5000;
```
aapeliv: I would suggest setting this to something bigger like 25 or 50 km to start. The density is low enough that it doesn't really matter and that way if you search for e.g. "NYC", it will show everyone within the city.

app/frontend/src/features/search/constants.ts
```diff
@@ -1,5 +1,23 @@
+import { UserSearchFilters } from "service/search";
+
+export const APPLY_FILTER = "Apply";
+export const FILTER_DIALOG_TITLE = "Filters";
+export const LOCATION = "Location";
```
aapeliv: If you type in "Finland" into the location, you get no one because the OSM point for Finland is in the middle of nowhere in the woods. (Probably similar for nay country/state/etc).

Maybe calling this "Near location" would give a hint? Otherwise something like "City"? I'm not sure

app/frontend/src/components/Autocomplete.tsx
```diff
@@ -23,6 +23,9 @@ export type AutocompleteProps<
   MuiAutocompleteProps<T, Multiple, DisableClearable, FreeSolo>,
   "renderInput"
 > & {
+  id: string;
+  /// TODO(lucas) - make sure autocompletes use this
+  error?: string;
```
darrenvong, line 28: Is this still relevant? Looks like you're using it in `helperText` of the TextField now?

darrenvong:
Screenshot 2021-05-16 at 01 27 34
Screenshot 2021-05-16 at 01 28 15

The text field label is getting clipped off on large+ breakpoints and being covered by the nav bar when I'm on the medium breakpoint

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -0,0 +1,182 @@
...
+  LAST_MONTH,
+  LAST_WEEK,
+} from "./constants";
+
+const lastActiveOptions = [
+  { label: LAST_DAY, value: 1 },
+  { label: LAST_WEEK, value: 7 },
+  { label: LAST_2_WEEKS, value: 14 },
+  { label: LAST_MONTH, value: 31 },
+  { label: LAST_3_MONTHS, value: 93 },
```
darrenvong, line 41: Pretty minor - but I am thinking we could store the value here as milliseconds so we don't need to calculate it when subtracting from `Date.now()` in the `userSearch` service helper?

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -0,0 +1,182 @@
...
+];
+
+const hostingStatusOptions = [
+  HostingStatus.HOSTING_STATUS_CAN_HOST,
+  HostingStatus.HOSTING_STATUS_MAYBE,
+  HostingStatus.HOSTING_STATUS_CANT_HOST,
+];
+
```
darrenvong, line 49: Screenshot 2021-05-16 at 01 44 36
The input label looks quite crammed together when it's filled in with a value - maybe they could do with some additional margin around each input?

darrenvong:
I just noticed that when there are no search results, the map seems to show everyone rather than nothing which is probably not the intended behaviour?
Screenshot 2021-05-16 at 01 55 58

app/frontend/src/features/search/LocationAutocomplete/LocationAutocomplete.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const submitButton = await screen.findByRole("button");
+    userEvent.click(submitButton);
+    await waitFor(() => {
+      expect(submitAction.mock.calls[0][0]).toMatchObject({
+        location: {
+          name: "test city, test country",
+          simplifiedName: "test city, test country",
+          location: { lng: 1.0, lat: 2.0 },
+        },
+      });
```
darrenvong:
```suggestion
      expect(submitAction).toHaveBeenNthCalledWith(
        1,
        expect.objectContaining({
          location: {
            name: "test city, test country",
            simplifiedName: "test city, test country",
            location: { lng: 1.0, lat: 2.0 },
          },
        }),
        expect.anything()
      );
```
I think this should also work. Slightly nicer than having to access the calls array directly I think with more explicit test matchers

app/frontend/src/utils/hooks.ts
```diff
@@ -39,4 +41,63 @@ function useSafeState<State>(
...
+    const fetchOptions = {
+      headers: {
+        Accept: "application/json",
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+      method: "GET",
+    };
+    try {
+      const response = await fetch(url, fetchOptions);
+      const nominatimResults = (await response.json()) as Array<NominatimPlace>;
```
darrenvong:
```suggestion
      const nominatimResults: NominatimPlace[] = await response.json();
```
Nitpick: Personal preference to not use type assertions if possible, but probably doesn't make a big difference in this specific case since the type is loose/bad anyway...

app/frontend/src/test/restMock.ts
```diff
@@ -0,0 +1,28 @@
...
+              address: { city: "test city", country: "test country" },
+              lon: 1.0,
+              lat: 2.0,
+              display_name: "test city, test country",
+            },
+          ])
+        );
+      }
+    ),
+  ]
```
darrenvong:
```suggestion
  rest.get(
    `${process.env.REACT_APP_NOMINATIM_URL!}search`,
    (req, res, ctx) => {
      return res(
        ctx.json([
          {
            address: { city: "test city", country: "test country" },
            lon: 1.0,
            lat: 2.0,
            display_name: "test city, test country",
          },
        ])
      );
    }
  )
```
I think this also works without the extra spread

app/frontend/src/setupTests.ts
```diff
@@ -4,16 +4,27 @@
...
+
 import user from "./test/fixtures/defaultUser.json";
 
 jest.mock("./service");
 
 global.defaultUser = user;
 global.localStorage = createLocalStorageMock();
 
+beforeAll(() => {
+  server.listen();
```
darrenvong: I imagine these will probably only be used in a few specific tests since most of our API is in gRPC still? I think it might be better to scope the set up code of this only in tests that needs the server rather than spinning up for every test suite/file

darrenvong (changes requested):
Haven't looked through everything yet, but this looks like a pretty solid start!

app/frontend/src/features/search/FilterDialog.tsx
lucaslcode: That would put milliseconds in the url - I think days is better

app/frontend/src/components/Autocomplete.tsx
lucaslcode: We are basically only using `helperText` for errors, and the implementation wasn't right. I am going to go through and check them all and make sure they use `error` unless they really intend to use the helper text.

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -61,16 +52,31 @@ export default function SearchPage() {
     undefined
   );
 
-  const { query } = useParams<{ query?: string }>();
+  //callbacks provided to the map need a ref, or they won't get the latest value of selectedResult
+  const selectedResultRef = useRef<number | undefined>(undefined);
```
darrenvong: Isn't this the same issue we solved in the other PR by syncing the callbacks in the `useEffect` instead so we don't end up with two states?

app/frontend/src/service/search.ts
```diff
@@ -1,13 +1,65 @@
...
+  query?: string;
+  lat?: number;
+  lng?: number;
+  radius?: number;
+  lastActive?: number; //within x days
+  hostingStatusOptions?: HostingStatus[];
+  numGuests?: number;
+}
+
+export async function userSearch(
```
darrenvong, line 20: Might be worth unit testing this since there's a bit more involved logic here in determining how the filters are set?

app/frontend/src/features/search/FilterDialog.tsx
darrenvong: Ahh yeah in that case, it's more user friendly to use the smaller numbers

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -0,0 +1,182 @@
...
+export default function FilterDialog({
```
darrenvong, line 58: Also might be one for the design team as well - but I wonder if it's more preferable to have the filters shown below the search input that's only revealed when the filter button is clicked as opposed to a modal dialog. The rough design for search we got at the moment doesn't really give much hint...

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -99,7 +133,7 @@ export default function SearchResultsList({
           const firstResult = resultUsers[0];
           if (!firstResult) return;
           const newBounds = new LngLatBounds([
-            [firstResult.lng, firstResult.lat],
+            [firstResult.lng - 0.0001, firstResult.lat - 0.0001],
```
darrenvong: How come this needs the extra "padding"?

app/frontend/src/setupTests.ts
```diff
@@ -4,6 +4,8 @@
 // learn more: https://github.com/testing-library/jest-dom
 import "@testing-library/jest-dom/extend-expect";
 
+import { server } from "test/restMock";
```
darrenvong:
```suggestion
```

app/frontend/src/setupTests.ts
```diff
@@ -16,6 +18,10 @@ afterEach(() => {
   jest.restoreAllMocks();
 });
 
+afterAll(() => {
+  server.close();
+});
+
```
darrenvong:
```suggestion
```
I guess this isn't needed anymore since you moved the setup code into the Autocomplete test file

app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -1,45 +1,79 @@
...
+      params.current.delete("query");
+    }
+  };
+
+  const onSubmit = handleSubmit(() => {
+    setParams();
+    history.push(`${searchRoute}?${params.current.toString()}`);
   });
 
+  //for some reason, params.current.keys.length is always 0
```
darrenvong: Apparently `URLSearchParams.keys()` returns an iterator. Would it work if you do `[...params.current.keys()].length` instead?

app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -1,45 +1,79 @@
...
+  const numParams = Object.keys(Object.fromEntries(params.current)).length;
+  //changing params doesn't cause re-render, but it's fine because
+  //closing the filter dialog box does
+  const hasFilters = params.current.has("query")
```
darrenvong, line 39: Then this can become `numParams > 0` again if the above works

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    ? searchParams.set("hostingStatus", options.join(","))
+                    : searchParams.delete("hostingStatus")
+                }
+                defaultValue={
+                  searchParams.has("hostingStatus")
+                    ? searchParams
+                        .get("hostingStatus")
+                        ?.split(",")
+                        .map((k) => Number.parseInt(k))
+                    : undefined
```
darrenvong: Since search parameters can be repeated (i.e. `?hostingStatus=1&hostingStatus=2` is valid) and `URLSearchParams` pick that up correctly, it might be worth using its API more by doing something like this in the `onChange` for the bit that's setting `hostingStatus`:
```typescript
options.forEach((option, index) => {
  if (index === 0) {
    searchParams.set("hostingStatus", option);
  } else {
    searchParams.append("hostingStatus", option);
  }
});
```

So then in `defaultValue` and in `SearchPage`, you can do `searchParams.getAll("hostingStatus")?.map((v) => Number(v))` without needing to split

darrenvong (changes requested):
A few more collection of little things I found except the design comment

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: I think it's just a commit or two behind :S

app/frontend/src/features/search/SearchResultsList.tsx
lucaslcode: I was getting weird behaviour with a single result. I think it was fixed by setting maxZoom though so I'll change this back

app/frontend/src/features/search/FilterDialog.tsx
lucaslcode: I was reading it as a dialog since it is off to the side

app/frontend/src/service/search.ts
lucaslcode: Yes - it will be done in this pr!

app/frontend/src/features/search/SearchBox.tsx
lucaslcode: surprised it has a length property if it doesn't work...

lucaslcode: oh, it's the function length *facepalm*

app/frontend/src/features/search/SearchBox.tsx
lucaslcode: This is still necessary - since the query is a `searchParam`, having one might not be a filter, it could just be a query

app/frontend/src/features/search/FilterDialog.tsx
darrenvong: Oh my bad, there is a design for it in `#5` in the design! Yeah that looks like a dialog to me with the close button on the top right. Ignore what I said there then 😅

lucaslcode:
@darrenvong I haven't had time to work on this yet, but should we merge it and I'll continue with tests in my next PR?

app/frontend/src/components/Autocomplete.tsx
darrenvong: Are you planning to follow up on this in the next PR too?

app/frontend/src/components/TextField.tsx
```diff
@@ -21,10 +21,11 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-interface AccessibleTextFieldProps extends BaseTextFieldProps {
+type AccessibleTextFieldProps = Omit<TextFieldProps, "variant"> & {
   id: BaseTextFieldProps["id"];
   onChange?: TextFieldProps["onChange"];
-}
+  variant?: "filled" | "outlined" | "standard";
```
darrenvong, line 27: I think you can add this and leave it as an interface and this will overrule the `variant` from `TextFieldProps`

darrenvong:
Screenshot 2021-05-22 at 20 34 06
The search input is being clipped under the nav bar on desktop now. But yeah, once that's fixed with the other minor comments, you can merge and do the tests in the follow up PR

app/frontend/src/components/Autocomplete.tsx
lucaslcode: Yes

app/frontend/src/components/TextField.tsx
lucaslcode: I can't, I get `An interface can only extend an object type or intersection of object types with statically known members.` when trying to extend `TextFieldProps`

darrenvong (approved):

PR #1163, opened 2021-04-29
[WIP] Backend/feat/simplifying queries

ShmuelTreiger:
Here are three different approaches for standardizing block/visibility filters

Commit 1) Method for joining the user table to any given query and filtering in the process

Commit 2) Standardizing the common queries

Commit 3) Method for filtering user table post-join

I'm partial to 2. Easy to use, very beginner friendly for any of the normal queries and most queries will be covered.

I suspect you'll be partial to 3 if any.

Big downside of 1 & 3 is having to get userblocks every time you query the User table. Could pass it in as an optional parameter (commit 4), but that further complicates the process.

### DIFF
--- a/app/backend/src/couchers/db.py
+++ b/app/backend/src/couchers/db.py
@@ -298,3 +298,12 @@ def all_blocked_or_blocking_users(user_id):
             user_block.blocking_user_id if user_block.blocking_user_id != user_id else user_block.blocked_user_id
             for user_block in relevant_user_blocks
         ]
+
+
+def filter_users_for_visibility_and_blocks(query, requesting_user_id, user_id_column, user_table=User):
+    relevant_blocks = all_blocked_or_blocking_users(requesting_user_id)
+    return (
+        query.join(user_table, user_table.id == user_id_column)
+        .filter(user_table.is_visible)
+        .filter(~user_table.id.in_(relevant_blocks))
+    )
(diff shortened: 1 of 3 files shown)


aapeliv:
3 is the way to go. Still requires a bit of cleanup and some more thinking about what that API should look like (input parameters, etc).

If we're going to be doing blocking, then maybe the way to go is to grab the list of users hidden just for this user in the auth interceptor, and attach it to the context in the same way that we attach `user_id` to context. That way we guarantee that query is only ever done once.

aapeliv:
I think we can use this: https://stackoverflow.com/a/15962062

Then the API could look like this:

```py3
host_request = (
    session.query(HostRequest)
    .join(from_users, from_users.id == HostRequest.from_user_id)
    .join(to_users, to_users.id == HostRequest.to_user_id)
    .filter_users(from_users, context) # like this
    .filter_users(to_users, context) # and this
    .filter(HostRequest.conversation_id == request.host_request_id)
    .filter(or_(HostRequest.from_user_id == context.user_id, HostRequest.to_user_id == context.user_id))
    .one_or_none()
)
```

```py3
host_request = (
    session.query(HostRequest)
    .filter_users_column(HostRequest.from_user_id, context) # like this
    .filter_users_column(HostRequest.to_user_id, context) # and this
    .filter(HostRequest.conversation_id == request.host_request_id)
    .filter(or_(HostRequest.from_user_id == context.user_id, HostRequest.to_user_id == context.user_id))
    .one_or_none()
)
```

PR #1167, opened 2021-04-30
test: simplify useUpdateHostingPreferences test

thdk:
Follow up on pr #1166 

This simplifies the tests for useUpdateHostingPreferences hook (to my knowledge).

### DIFF
--- a/app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
+++ b/app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
@@ -7,9 +7,13 @@ import { HostingPreferenceData, service } from "service";
 import wrapper from "test/hookWrapper";
 import { addDefaultUser } from "test/utils";
 
-const getUserMock = service.user.getUser as jest.Mock;
+const getUserMock = service.user.getUser as jest.Mock<
+  ReturnType<typeof service.user.getUser>
+>;
 const updateHostingPreferenceMock = service.user
-  .updateHostingPreference as jest.Mock;
+  .updateHostingPreference as jest.Mock<
+  ReturnType<typeof service.user.updateHostingPreference>
+>;
 
 describe("useUpdateHostingPreference hook", () => {
   const newHostingPreferenceData: HostingPreferenceData = {
@@ -43,16 +47,7 @@ describe("useUpdateHostingPreference hook", () => {
     addDefaultUser();
 
     updateHostingPreferenceMock.mockResolvedValue(new Empty());
-    getUserMock.mockImplementation(() => {
-      if (!updateHostingPreferenceMock.mock.calls.length) {
-        return defaultUser;
-      } else {
-        return {
-          ...defaultUser,
-          ...newHostingPreferenceData,
-        };
-      }
-    });
+
     const { result, waitFor } = renderHook(
       () => ({
         currentUser: useCurrentUser(),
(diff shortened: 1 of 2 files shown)


app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -100,21 +77,19 @@ describe("useUpdateHostingPreference hook", () => {
     //once for getCurrentUser then once for invalidation
     expect(getUserMock).toHaveBeenCalledTimes(2);
     expect(getUserMock).toHaveBeenCalledWith(`${defaultUser.userId}`);
-    // Things that have been updated are being reflected
+    // Things that have been updated are being reflected, rest of profile should be the same as before
     expect(result.current.currentUser.data).toMatchObject({
```
darrenvong: I'm a little confused why this now works... does that mean we've got rid of all of the `*Value` wrapper in the protobuf in the API contract? 🤔

darrenvong: I see why this has worked, because the API response is mocked, so really you can return whatever in the `currentUser` mock and asserting that the mock returns that same thing of course is going to work.

Since everything is so heavily mocked here, I'm thinking if it might be better off if we just delete that assertion altogether, as this isn't testing anything apart from our mock works, which is useless. The above assertions are enough for saying as long we call the function with the right data, we trust the mocked function works correctly

thdk: Yes, I agree. What need to be tested is if the api is called with the right parameters. Whether or not the api works, that's for another test. I'll remove that last part from this test.

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -6,11 +6,14 @@ import { act } from "react-test-renderer";
 import { HostingPreferenceData, service } from "service";
 import wrapper from "test/hookWrapper";
 import { addDefaultUser } from "test/utils";
-import { Primitives } from "utils/types";
 
-const getUserMock = service.user.getUser as jest.Mock;
+const getUserMock = service.user.getUser as jest.Mock<
+  ReturnType<typeof service.user.getUser>
+>;
```
darrenvong:
```suggestion
const getUserMock = service.user.getUser as MockedService<
  typeof service.user.getUser
>;
```
I made the `MockedService` utility type for this purpose in `test/utils` since we need this type assertion for all of the mocked API functions so frequently

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
```diff
@@ -6,11 +6,14 @@ import { act } from "react-test-renderer";
...
 const updateHostingPreferenceMock = service.user
```
darrenvong, line 13: Same comment as above :)

darrenvong (changes requested):

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.test.tsx
thdk: Ah yes, of course. I shall update it .

darrenvong (approved):

PR #1168, opened 2021-05-01
Update group message preview when new messages arrive

### LINKED ISSUE #1004: [bug][frontend] Message list does not show new messages in preview until refresh
When a user sends a message to another user in chats, hosting, or surfing tab, the recipient will see the number badge update in the tab list. The conversation preview does not update by showing the latest message that was received until the user refreshes the page or leaves the tab and goes back. This isn't a big issue, but if a user is actively chatting with another user and wants to see the message preview they are unable to do so.

Jakeii:
Added message count as a dependant key for the `listGroupsChats` query so it'll invalidate when it changes.

This won't work for the hosting/surfing tabs unfortunately because the counts are just the requests sent/received not the message count.

This will close #1004

### DIFF
--- a/app/frontend/src/features/messages/groupchats/GroupChatsTab.tsx
+++ b/app/frontend/src/features/messages/groupchats/GroupChatsTab.tsx
@@ -16,8 +16,12 @@ import { Link } from "react-router-dom";
 import { routeToGroupChat } from "routes";
 import { service } from "service";
 
+import useNotifications from "../../useNotifications";
+
 export default function GroupChatsTab() {
   const classes = useMessageListStyles();
+  const {data: notifications} = useNotifications();
+  const unseenMessageCount = notifications?.unseenMessageCount;
 
   const {
     data,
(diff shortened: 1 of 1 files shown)


lucaslcode:
Doing it this way is a bit of a leak, since the cache will grow and grow (since new queryKeys doesn't delete the old ones).

Using unseen message count is neat though, I think add a `useEffect` with that as a dependency which calls `queryClient.invalidateQuery(...)`

Jakeii:
@lucaslcode react-query clears the cache periodically for unused queries doesn't it?

lucaslcode:
I'm not so sure, possibly. Either way, to me it seems to go against the intention of the queryKey (it's forcing it to be a new query but it's really the same query, we just know it will have new data)

Jakeii:
@lucaslcode cool, yeah your suggestion works great thanks!

lucaslcode (approved):

PR #1170, opened 2021-05-02
Implement onboarding emails

aapeliv:
Adds a few onboarding emails that are sent out to users.


kalvdans:
What is onboarding and why can't we use cloudflare mail?

aapeliv:
> What is onboarding

Introducing people to the platform. Basically we want to send a few friendly emails to people after they sign up. The first one is just a "welcome" email. The second is sent only if the user hasn't filled in their profile enough to remind them to do so.

Currently there's a problem with a lot of empty profiles, that's why Itsi would like these emails to be sent out.

The idea is that the emails are sent out at a schedule that starts when the user signs up, etc.

> why can't we use cloudflare mail?

What's cloudflare email?

Also, I've added the mailchimp integration here so that when people sign up they are automatically subscribed to the mailing list.

kalvdans:
> > What is onboarding
> 
> Introducing people to the platform. Basically we want to send a few friendly emails to people after they sign up. The first one is just a "welcome" email. The second is sent only if the user hasn't filled in their profile enough to remind them to do so.
> 
> Currently there's a problem with a lot of empty profiles, that's why Itsi would like these emails to be sent out.
> 
> The idea is that the emails are sent out at a schedule that starts when the user signs up, etc.

ok!

> > why can't we use cloudflare mail?
> 
> What's cloudflare email?

Sorry, I meant the current provider, (Amazon)?

Ok so mailchimp is used to mass-send emails without going through the backend.

aapeliv:
> Sorry, I meant the current provider, (Amazon)?

Yes, the onboarding emails (with templates here) go through Amazon (SES: "simple email service").

> Ok so mailchimp is used to mass-send emails without going through the backend.

Yep. That's the newsletter mainly.

Sorry, I mixed two things into the same PR here: adding the onboarding emails (through the same system as other emails from backend), and adding auto-subscription to the newsletter (through Mailchimp). I can also separate them into different PRs, but they're quite alike so I merged the migration.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -114,3 +115,87 @@ def process_send_message_notifications(payload):
...
+    Sends out onboarding emails
+    """
+    logger.info(f"Sending out onboarding emails")
+
+    with session_scope() as session:
+        # first onboarding email
+        users = session.query(User).filter(User.onboarding_emails_sent == 0).all()
+
+        for user in users:
+            user.onboarding_emails_sent = 1
```
kalvdans: Send email firsr, then increment counter, in case the email sending throws exception

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -114,3 +115,87 @@ def process_send_message_notifications(payload):
...
+
+def process_add_users_to_email_list(payload):
+    if not config.config["MAILCHIMP_ENABLED"]:
+        logger.info(f"Not adding users to mailing list")
+        return
+
+    logger.info(f"Adding users to mailing list")
+
+    with session_scope() as session:
+        users = session.query(User).filter(User.added_to_mailing_list == False).limit(100).all()
```
kalvdans, line 171: Only send the user's email to a third party if they have agreed, check that the TOS version the user has agreed to contains that.

kalvdans (commented):
Looks good code-wise, although we could use an external program to sync the list of users to mailchimp to not pollute our database. It is something a fork would remove immediately anyway.

app/backend/src/couchers/jobs/handlers.py
aapeliv: :+1:

Waiting for new TOS to be finalized and added to frontend.

aapeliv:
@kalvdans: I added an env var to keep track of the TOS version number. But maybe this should be handled some other way, it's a bit icky. Should I just put the version code in markdown frontmatter in `terms_of_service.md` instead?

kalvdans:
> @kalvdans: I added an env var to keep track of the TOS version number. But maybe this should be handled some other way, it's a bit icky. Should I just put the version code in markdown frontmatter in `terms_of_service.md` instead?

Just hardcode the required tos version number it in the code, no need for an environment variable.

aapeliv:
@kalvdans: ready for review again

kalvdans (approved):
lgtm!

PR #1179, opened 2021-05-03
Fix search in area

### LINKED ISSUE #1158: Search in area does not work

aapeliv:
Closes #1158. I got the units mixed up.

This does an approximation of 1 degree = 111 111 m, which technically only holds at the equator, and gets worse as one approaches the poles. This is because postgis does the computation in a cartesian plane.

**Backend checklist**
- [ ] All tests pass


kalvdans (approved):
tricky one, lgtm!

PR #1180, opened 2021-05-04
Discussions part 3: single discussion

darrenvong:
Discussion Part 3 for #543. Includes a more tidied up single discussion page with Storybook and some initial tests. To avoid this getting too large, this is not quite feature parity with the basic version of the page - at the moment, it only lists top level replies to a discussion and not nested replies.

You also can't reply... but those functionality will return in the next PR. For that reason, it might be best to check this out through Storybook as it has some dummy data with more replies in the discussion

**Desktop**
Screenshot 2021-05-04 at 01 20 19

**Mobile**
Screenshot 2021-05-04 at 01 20 39

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  />
+                  <div className={classes.replyContent}>
+                    <Typography variant="body2">
+                      {getByCreator(user?.name ?? UNKNOWN_USER)}
+                      {` • ${postedTime}`}
+                    </Typography>
+                    <Typography variant="body1">{reply.content}</Typography>
+                  </div>
+                </Card>
+              );
```
darrenvong: This will probably be refactored into a Comment component in the next PR, but for getting something to show, it will do for now I guess...

app/frontend/src/theme.ts
```diff
@@ -86,7 +87,7 @@ const themeOptions: ThemeOptions = {
   shape: {
     borderRadius,
     navPaddingDesktop: spacing(10),
-    navPaddingMobile: spacing(navBarHeight),
+    navPaddingMobile: spacing(navBarHeight + navBarPadding),
```
darrenvong: Had to add a little of bit padding back, as the header with the back button in the single discussion was getting clipped under the nav bar. This was also a problem for the group chat view, so I reckon that's why the padding was there before. The padding is a bit less than before though and it seems fine with the map page.

app/frontend/src/features/communities/discussions/DiscussionPage.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <Route exact path="/previous-page">
+        <h1 data-testid="previous-page">Previous page</h1>
+      </Route>
+    </Switch>,
+    { wrapper }
+  );
+
+  return client;
+}
+
```
darrenvong, line 52: These don't cover much the replies section for now as that needs more work in the next PRs, so only a starting point

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+                : "sometime";
+              return (
+                <Card
+                  className={classes.replyContainer}
+                  key={reply.createdTime!.seconds}
+                >
+                  <Avatar
+                    user={user}
+                    className={classes.avatar}
+                    isProfileLink={false}
```
lucaslcode: Why is this false? They should always link to profile unless they are inside another link, which isn't the case here.

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+import {
+  COMMENTS,
+  getByCreator,
+  NO_COMMENTS,
+  UNKNOWN_USER,
+} from "../constants";
+import { useThread } from "../hooks";
+
+const useStyles = makeStyles((theme) => ({
```
lucaslcode, line 18: See my comment on naming

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+    marginInlineStart: theme.spacing(3),
+  },
+  avatar: {
+    height: "3rem",
+    width: "3rem",
+  },
+}));
+
+interface CommentTreeProps {
+  discussion: Discussion.AsObject;
```
lucaslcode: This is fine for now but we may want to generalize this later to take a `thread_id` instead, since, for example, events will have a thread id for comments etc. but they are not discussions.

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+}
+
+export default function CommentTree({ discussion }: CommentTreeProps) {
+  const classes = useStyles();
+
+  const {
+    data: discussionThread,
```
lucaslcode: maybe `discussionComments`? See my comment about naming

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <>
+      <Typography className={classes.commentsHeader} variant="h2">
+        {COMMENTS}
+      </Typography>
+      {threadError && <Alert severity="error">{threadError.message}</Alert>}
+      {isThreadLoading || isUsersLoading ? (
+        <CircularProgress />
+      ) : hasAtLeastOnePage(discussionThread, "repliesList") ? (
+        <div className={classes.repliesListContainer}>
```
lucaslcode: `commentsListContainer` since this is the outer one

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    user={user}
+                    className={classes.avatar}
+                    isProfileLink={false}
+                  />
+                  <div className={classes.replyContent}>
+                    <Typography variant="body2">
+                      {getByCreator(user?.name ?? UNKNOWN_USER)}
+                      {` • ${postedTime}`}
+                    </Typography>
+                    <Typography variant="body1">{reply.content}</Typography>
```
lucaslcode, line 96: according to the .proto, this can be markdown (but I wonder if only the discussion content should be)

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  </div>
```
lucaslcode, line 97: This needs another level of nesting for comment replies

lucaslcode: Oh sorry I just saw you're aware of this lol

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -0,0 +1,113 @@
...
+  discussion: Discussion.AsObject;
+}
+
+export default function CommentTree({ discussion }: CommentTreeProps) {
+  const classes = useStyles();
+
+  const {
+    data: discussionThread,
+    error: threadError,
+    isLoading: isThreadLoading,
```
lucaslcode: also these

app/frontend/src/features/communities/discussions/CommentTree.tsx
lucaslcode: Actually, if this is going to be generalized, maybe `comments` or `threadComments`

app/frontend/src/features/communities/discussions/DiscussionPage.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+    expect(screen.getByText("Funny Cat current User")).toBeVisible();
+    expect(screen.getByText("Created at Jan 01, 2020")).toBeVisible();
+  });
+
+  it("renders a loading skeleton if the user info is still loading", async () => {
+    getUserMock.mockImplementation(async () => new Promise(() => undefined));
+    renderDiscussion();
+    await waitForElementToBeRemoved(screen.getByRole("progressbar"));
+
+    expect(await screen.findByTestId(CREATOR_LOADING_TEST_ID)).toBeVisible();
```
lucaslcode: Can we avoid the test id by checking that the discussion info is shown, user info isn't shown and there is a progress bar? Or is that too inaccurate?

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+                    {discussionCreator?.name ?? UNKNOWN_USER}
+                  </Typography>
+                )}
+                <Typography variant="body2">
+                  Created at{" "}
+                  {dateFormatter.format(timestamp2Date(discussion.created!))}
+                </Typography>
+              </div>
+            </div>
+            <Divider />
```
lucaslcode: I know there's no mockup for this, but IMO, don't use any dividers and put the user at the top under the header (without the "Added by" text)

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+            <Markdown source={discussion.content} />
+            <Typography className={classes.addedByLabel} variant="body1">
+              {ADDED_BY}
+            </Typography>
+            <div className={classes.creatorContainer}>
+              <Avatar
+                user={discussionCreator}
+                className={classes.avatar}
+                isProfileLink={false}
```
lucaslcode:
```suggestion
```

app/frontend/src/theme.ts
lucaslcode: I think it's better for pages to handle this individually, for the same reason we are now usually avoiding putting margins on components.. Most pages do so with `<PageTitle>` which has padding. The map page now has a gap which breaks the fullscreen effect.

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+  },
+  avatar: {
+    height: "3rem",
+    width: "3rem",
+  },
+}));
+
+export const CREATOR_LOADING_TEST_ID = "creator-loading-state";
+
+export default function DiscussionPage() {
```
lucaslcode, line 60: It would be great if this could be shown within the community page - makes it less bare, and discussions are always part of a community.

Maybe that means making the route `/community/1/world/discussion/24/what-is-good-place`? That seems a tad long...
The other option would be making the backend give the community id in the discussion but that seems like code smell.

lucaslcode (changes requested):
Exciting to see this coming together!

I think we need to be strict with naming. It is pretty confusing and maybe needs to be changed. But this is my understanding:
- Thread - just a list of replies, could be nested up to two levels, no content itself (unless it is a reply)
- Reply - a thread with content
- Discussion - sort of a page which is a container for a top-level thread with an author and content
- Comment - the first level of replies from a top-level parent thread (which could be, say, a discussion or event)

This is confusing because comments are also replies, but not all replies are comments. And comments, replies and discussions are all threads.

Here is the illustration from the .proto: 
```
// Discussion threads are organized as a tree with maximum three levels depth:
  // thread1 (eg. discussion)
  // +- comment1 (this is also a reply, since comments are replies)
  // |  +- reply1
  // |  +- reply2
  // +- comment2
  //    +- reply3
  //    +- reply4
  // thread2
  // +- comment3
  // |  +- reply5
  // |  +- reply6
  // +- comment4
  //    +- reply7
  //    +- reply8
```
ALL of these have a thread id. (You probably know this, it's more for reference)

Anyway, the point is, in one or two places where I commented, the name can be clearer.

app/frontend/src/features/communities/discussions/CommentTree.tsx
lucaslcode: (What I mean is, some of these classnames should be `Comment` instead of `Reply` if it will only be used for the first level

app/frontend/src/features/communities/discussions/CommentTree.tsx
darrenvong: Good catch, I think I copied it from discussion card where it made sense to mark this false and had the same styling. Oops will update!

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
darrenvong: Yeah the trickiest part of all this has been figuring out the best way to lay this out without things looking terrible at the same time! I based it off the design for guides (also where I got the divider ideas off) which has the creator at the bottom. Do you think it's worth raising that to design and maybe it's better to have it near the top? Or are guides different enough that the current design as it is makes sense?

app/frontend/src/features/communities/discussions/DiscussionPage.test.tsx
darrenvong: Checking discussion is shown but user is not shown could work actually. If it errors, it'd not be loading and falls back to "Unknown user" instead since 1) I can't think of a good way to show an error state here without it looking terrible and 2) this could error if in the future the user deactivates/deletes their account, so falling back to "Unknown user" seems reasonable to me.

The progress bar bit is quite impossible without test ID because the `Skeleton` component doesn't actually have a `progressbar` role and MUI implements it with a span... so that's why I added a test ID as spans are so generic there's nothing else in the query hierarchy we can use to select this properly. But maybe the parts described above are sufficient, so I think I might remove it

app/frontend/src/features/communities/discussions/CommentTree.tsx
darrenvong: Good point... this is where I'm a little confused and so kinda went with the simplest option for now. Should both the discussion creation form and the reply input box (still to come) use the Markdown editor then?

darrenvong:
Having made a small start in the next part (and got a slightly clearer mental modal on this) and did a little refactoring, I agree a lot with what you said about the naming of things.

Maybe we should avoid referencing "replies" altogether and call them all "comments"? A reply (based off terminology from the ASCII diagram) feels to me like a special comment where you can't reply further, so calling it a comment still kinda accurately describes what it is?

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
darrenvong: Do you mean putting this under the community heading ("Welcome to Amsterdam" in the design example)? But yeah that URL seems a little clunky.

Isn't the idea of this page is for people to access the discussion standalone?

Although I see what you're saying, this still gives no way for the users to access the discussion in the same page of the entity it's tied to (by entity I mean a community, event etc), but the current URL structure for a single discussion as it is (`/discussion/72/some-topic`) doesn't work as the first bit is already different so the URL won't match properly for us to nest this under the community tab/page...

app/frontend/src/features/communities/discussions/CommentTree.tsx
darrenvong: I think I'm going to leave this for the next PR as this is in the process of being refactored into its own component

darrenvong:
Addressed all comments apart from the URL one which I'm not sure we can resolve without further discussion on the URL structure and maybe some design input on how that'd look on a community page?

lucaslcode (approved):
lgtm, url stuff discussed on next pr

PR #1181, opened 2021-05-04
Make profile card go full width on mobile screens

darrenvong:
As requested by Itsi in the frontend chat, so now on mobile screen it looks like this:

Screenshot 2021-05-02 at 00 28 38

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/AppRoute.tsx
+++ b/app/frontend/src/AppRoute.tsx
@@ -67,7 +67,11 @@ export default function AppRoute({
                 [classes.fullscreenContainer]: variant === "full-screen",
                 [classes.standardContainer]: variant === "standard",
               })}
-              maxWidth={variant === "full-screen" ? false : undefined}
+              maxWidth={
+                variant === "full-screen" || variant === "full-width"
+                  ? false
+                  : undefined
+              }
             >
               {isJailed ? (
                 <Redirect to={jailRoute} />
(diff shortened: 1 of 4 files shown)


app/frontend/src/AppRoute.tsx
```diff
@@ -23,17 +24,21 @@ export const useStyles = makeStyles((theme) => ({
...
+  fullWidthContainer: {
+    paddingLeft: 0,
+    paddingRight: 0,
+  },
 }));
 
 interface AppRouteProps extends RouteProps {
   isPrivate: boolean;
-  isFullscreen?: boolean;
+  variant?: "standard" | "full-screen" | "full-width";
```
darrenvong, line 38: Switched to using `variant` here as booleans aren't really a good idea when there are more than two state values, since boolean combos quickly introduce impossible states and makes things hard to reason about.

app/frontend/src/AppRoute.tsx
```diff
@@ -53,12 +58,20 @@ export default function AppRoute({
       render={({ location }) => (
         <>
           {isAuthenticated ? (
-            <Container className={classes.standardContainer}>
+            <Container
+              className={classNames({
+                [classes.fullWidthContainer]: variant === "full-width",
+                [classes.fullscreenContainer]: variant === "full-screen",
```
darrenvong, line 68: We haven't got a use case for a full screen container yet for authenticated routes, but we may as well fully support it since this doesn't break existing behaviour? Also it'd be more accurate to the behaviour described by the prop, which wasn't the case before as is only full screen when used on unauthenticated routes...

darrenvong: Never mind, looks like we do a have full screen use case since the latest merge conflict so this is a valid case to support :D

lucaslcode (approved):

PR #1184, opened 2021-05-05
Feature/replace tos

### LINKED ISSUE #1156: Replace fake Terms of Service  on signup with actual
@ItsiW has text for ToS. It's long - we should change sign-up to just be a link and checkbox rather than the full ToS inline.

Jesus82:
Closes [#1156](https://github.com/Couchers-org/couchers/issues/1156)


Jesus82:
I captured the onChange event of the checkbox, not sure if that's the way to go in React (as it doesn't really bind to the state).

aapeliv:
I've just opened a new PR, #1185 that adds the `Resources/GetTermsOfService` RPC that you can use to get the latest ToS from the backend.

I propose you add a new page, say at route `/terms`, that contains the actual ToS text.

There you need to basically have a blank page with a title "Terms of Service" that uses a markdown component to render the actual markdown for the ToS (that you can get through that backend call). This might be complicated a bit by the fact that it has to be accessible even when not logged in, or at least when jailed.

This complicates it a bit, happy to talk through it in the meeting coming up!

aapeliv:
@Jesus82: #1185 was now merged and I've deployed it onto the preview site. You can now develop against that API. (You'll need to rebase onto develop or merge develop into here and then regenerate protos.)

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -319,17 +319,16 @@ export default function CompleteSignupForm() {
...
-              <Button
-                classes={{
-                  label: authClasses.buttonText,
-                  root: authClasses.button,
-                }}
-                loading={loading}
-                onClick={() => setAcceptedTOS(true)}
-                disabled={acceptedTOS}
-              >
-                {acceptedTOS ? THANKS : ACCEPT}
```
lucaslcode, line 331: Also remove these constants from `constants.ts` if they are no longer being used

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -319,17 +319,16 @@ export default function CompleteSignupForm() {
...
-              </Button>
+              <FormControlLabel
+                control={
+                  <Checkbox
+                    onChange={() =>
+                      setAcceptedTOS((prevAcceptedTOS) => !prevAcceptedTOS)
```
lucaslcode: Wow, I can't believe I didn't know you could use the previous state 😆

lucaslcode (commented):
This looks good for the form, but maybe you can rebase onto Aapeli's branch and use the new api to display the TOS from the backend?

aapeliv:
> you can rebase onto Aapeli's branch and use the new api

Just a heads up, it's been merged into develop and is also available on the preview instance.

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
Jesus82: Stackoverflow to the rescue :D

Jesus82:
> > you can rebase onto Aapeli's branch and use the new api
> 
> Just a heads up, it's been merged into develop and is also available on the preview instance.

So should I just merge develop into this branch and then push?

aapeliv:
> > > you can rebase onto Aapeli's branch and use the new api
> > 
> > 
> > Just a heads up, it's been merged into develop and is also available on the preview instance.
> 
> So should I just merge develop into this branch and then push?

You can rebase or just merge, up to you

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
darrenvong: I'd say this is the preferred way to do set state if you need to rely on previous state (even though probably doesn't matter in this case, more getting into good habits :p), since React guarantees you always get back the correct previous state, rather than a small chance that this might read the "wrong state" if other things also set/changes the state.

app/frontend/src/components/Markdown.tsx
```diff
@@ -56,7 +59,7 @@ export default function Markdown({
     return () => viewer.current?.remove();
   }, [source, topHeaderLevel]);
 
-  return <div className={classes.root} ref={rootEl} />;
+  return <div className={classNames(className, classes.root)} ref={rootEl} />;
```
darrenvong, line 62: Extended this to allow some styling on this outer div container, mainly for adding spacing so things don't look terrible and hard to read with it stretching across the full width of the screen, so I don't need to add another wrapper div around this in the TOS

darrenvong (commented):
This should be ready now with the necessary changes to pull the terms of service text from the backend. If someone else can take a look that'd be great so I don't just cheekily approve my own changes :p

app/frontend/src/features/auth/jail/TOSSection.tsx
```diff
@@ -30,7 +30,7 @@ export default function TOSSection({
 
   return (
     <Box className={className}>
-      <TOS />
+      <TOSLink />
       <Button loading={loading} onClick={accept} disabled={completed}>
         {completed ? "Thanks!" : "Accept"}
```
aapeliv: Move to constants?

app/frontend/src/features/auth/constants.tsx
```diff
@@ -58,13 +57,11 @@ export const SIGN_UP_COMPLETE_HEADER = "Your basic details";
 export const SIGN_UP_FULL_NAME = "Full name";
 export const SIGN_UP_HEADER = "Let's get started!";
 export const SIGN_UP_LOCATION_MISSING = "Please select a valid location.";
+export const SIGN_UP_TOS_ACCEPT = "I Accept TOS";
```
aapeliv: "I accept the Terms of Service"

app/frontend/src/features/auth/constants.tsx
```diff
@@ -58,13 +57,11 @@ export const SIGN_UP_COMPLETE_HEADER = "Your basic details";
...
+export const SIGN_UP_TOS_LINK_TEXT = "Link to TOS";
```
aapeliv: "Read the Terms of Service"

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -318,18 +318,17 @@ export default function CompleteSignupForm() {
...
-                  label: authClasses.buttonText,
-                  root: authClasses.button,
-                }}
-                loading={loading}
-                onClick={() => setAcceptedTOS(true)}
-                disabled={acceptedTOS}
-              >
-                {acceptedTOS ? THANKS : ACCEPT}
-              </Button>
+              <TOSLink />
```
aapeliv: It would be good to put this in a sentence:

"To continue, please read and accept the [Terms of Service]."

Or even just put it as part of the checkbox "I accept the [Terms of Service]"?

aapeliv (changes requested):
Looks good to me, a few small comments

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
darrenvong: Uhh this is a bit of a pain in the arse, as I want the link to be reused across the few places but putting it into a sentence makes as part of the checkbox label makes extracting the text into constans awkward, so think I'll stick with the first even though it's a bit of repetition

aapeliv: Sounds good

app/frontend/src/components/TOS.tsx
```diff
@@ -1,35 +1,39 @@
...
+  if (error) {
+    // Re-throw error to trigger error boundary to encourage user to report it
+    // if they can't see the terms
+    throw error;
+  }
+
+  return isLoading ? (
+    <CircularProgress />
+  ) : data ? (
+    <Markdown className={classes.root} source={data?.termsOfService} />
```
aapeliv: I think this needs a `<PageTitle>Terms of Service</PageTitle>`, sorry for missing this earlier

aapeliv (changes requested):
Sorry, missed one thing

aapeliv (approved):
Looks good to me!

PR #926, opened 2021-04-02
Regions and languages

aapeliv:
@ShmuelTreiger did most of the work on regions

PR #1185, opened 2021-05-05
ToS RPC

aapeliv:
**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/resources/terms_of_service.md
+++ b/app/backend/resources/terms_of_service.md
@@ -34,12 +34,9 @@ We are not liable if at any time, with or without prior notice, we modify, suspe
 
 # 5. Our service providers
 
-We take utmost care to protect your personal data and choose our service providers carefully.
-
 Our platform is hosted on Amazon Web Services (AWS), and we use a variety of their products to provide this service to you, including hosting our servers, storing our backups, and sending transactional emails.
 
-We manage our mailing lists for our newsletter and general announcements using MailChimp. You agree to be added to our mailing list. You may unsubscribe at any time from within these emails. We only share minimal information with MailChimp required to send you these emails, such as your email address and display name.
-
+We manage a mailing list for our newsletter and general announcements using MailChimp. You agree to be added to our mailing list, but may unsubscribe at any time. We share minimal information with MailChimp including your email address and display name.
 
 # 6. Termination
 
--- a/app/backend/resources/terms_of_service.md
+++ b/app/backend/resources/terms_of_service.md
@@ -2,6 +2,7 @@ You must agree to these Terms to access the Couchers.org services and create an 
 
 **Note: we'll update the Terms of Service as we approach the full launch and clarify the reasons and ways in which we may terminate or ban accounts as we develop these processes and solidify our moderation process.**
 
+*This is version 2.0 of the Terms of Service, last updated on 5th May, 2021.*
 
 # 1. Foundation
 
--- a/app/backend/resources/terms_of_service.md
+++ b/app/backend/resources/terms_of_service.md
@@ -1,0 +1,45 @@
+You must agree to these Terms to access the Couchers.org services and create an account.
+
+**Note: we'll update the Terms of Service as we approach the full launch and clarify the reasons and ways in which we may terminate or ban accounts as we develop these processes and solidify our moderation process.**
+
+
+# 1. Foundation
+
+Access to the Couchers.org platform ("platform", "service") is provided to you by the **Couchers.org Foundation** ("Couchers.org", "we", "us"), a non-profit organization formed in Sydney, Australia.
+
+
+# 2. Eligibility
+
+You must be at least 18 years old to access the service.
+
+
+# 3. Account information
+
+You are responsible for any activity done using your account, as well as securing your account from inappropriate access. For example, you must use a secure password not shared across other products or services, and you cannot leave your account logged in on shared computers. If you give access to your account to others for any purpose, you are liable for any such activity.
+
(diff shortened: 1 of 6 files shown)


app/pb/resources.proto
```diff
@@ -0,0 +1,19 @@
...
+
+  // This API contains general resources needed by frontends and managed by the backend team
+
+  rpc GetTermsOfService(google.protobuf.Empty) returns (GetTermsOfServiceRes) {
+    // Returns the latest Terms of Service markdown, to be shown on signup or if jailed due to this
+  }
+}
+
+message GetTermsOfServiceRes {
+  string terms_of_service = 1; // CommonMark
```
kalvdans:
```suggestion
  string terms_of_service = 1;  // CommonMark without images
```

aapeliv:
I was thinking about the ToS versioning: at the moment it's all in the backend, that's good enough, right? Frontend doesn't need to know about what version we're up to.

Also, what do you think of this "resources" name, I think it's better than "constant data" (used in #926, I can update it there too, I think these belong in the same API)?

The ToS still needs a few tweaks, and I'll get Itsi to look over it later. E.g. it needs a "last updated" text somewhere and some brief sentence on how we will update it.

kalvdans (commented):
There is a race condition here, if we update the tos while someone is reading the old ones, they will errorneously agree to the new one. I suggest we versionfiy them even if we only allow the latest one. (i.e. pass an integer along in the reply, that should be passed back to the backend when agreeing to the tos)

aapeliv:
> There is a race condition here, if we update the tos while someone is reading the old ones, they will errorneously agree to the new one. I suggest we versionfiy them even if we only allow the latest one. (i.e. pass an integer along in the reply, that should be passed back to the backend when agreeing to the tos)

Hmmm... I see your concern, and of course ToS is more important than other things. But unfortunately this complicates the frontend quite a lot: the ToS are so lengthy that we won't display them on the signup form, instead, we'll link to a separate page with the ToS there.

Sure, we could send a ToS version that reflects the current version when the user goes to the signup form, and then use that. But that would only protect against the case that the user opened the ToS page *after* loading the signup form, but what if the user read the ToS before signing up to make sure they were happy with it?

This is the kind of race condition that we really can't protect against in general: e.g. if someone makes an event and you load the page, then they change the description and you click attend without reloading. It's of course a minor thing and ToS is more important.

kalvdans:
> Hmmm... I see your concern, and of course ToS is more important than other things. But unfortunately this complicates the frontend quite a lot: the ToS are so lengthy that we won't display them on the signup form, instead, we'll link to a separate page with the ToS there.

That makes people not read the ToS. Let's have a short ToS so that it can actually be read. Most paper contracts work that way, you have to sign each page of the contract text with your initials.

> Sure, we could send a ToS version that reflects the current version when the user goes to the signup form, and then use that. But that would only protect against the case that the user opened the ToS page _after_ loading the signup form, but what if the user read the ToS before signing up to make sure they were happy with it?

That's tricky; let's stick a version or date in the beginning of the ToS so that the user can verify it hasn't been updated since she last read it.

> This is the kind of race condition that we really can't protect against in general: e.g. if someone makes an event and you load the page, then they change the description and you click attend without reloading. It's of course a minor thing and ToS is more important.

Good comparison! There are technical tricks to do prevent that (checksumming the event text and pass the checksum along with your attend button comes to my mind).

app/backend/resources/terms_of_service.md
```diff
@@ -0,0 +1,45 @@
...
+We are not liable if at any time, with or without prior notice, we modify, suspend or discontinue our services.
+
+
+# 5. Our service providers
+
+We take utmost care to protect your personal data and choose our service providers carefully.
+
+Our platform is hosted on Amazon Web Services (AWS), and we use a variety of their products to provide this service to you, including hosting our servers, storing our backups, and sending transactional emails.
+
+We manage our mailing lists for our newsletter and general announcements using MailChimp. You agree to be added to our mailing list. You may unsubscribe at any time from within these emails. We only share minimal information with MailChimp required to send you these emails, such as your email address and display name.
```
kalvdans:
```suggestion
We manage a mailing list for our newsletter and general announcements using MailChimp. You agree to be added to our mailing list, and may unsubscribe at any time from within these emails. We share your email address and display name with MailChimp.
```

app/backend/resources/terms_of_service.md
```diff
@@ -0,0 +1,45 @@
...
+Although we are not legally responsible for user-created content, we actively moderate the platform to remove any content that is harmful, deceitful, or otherwise does not adhere to our rules. If you believe some content does not adhere to our rules, please report it to us via support@couchers.org.
+
+When posting any image, you agree to obtain consent from any individual pictured in the image.
+
+We are not liable if at any time, with or without prior notice, we modify, suspend or discontinue our services.
+
+
+# 5. Our service providers
+
+We take utmost care to protect your personal data and choose our service providers carefully.
```
kalvdans:
```suggestion

```

aapeliv:
> That makes people not read the ToS. Let's have a short ToS so that it can actually be read. Most paper contracts work that way, you have to sign each page of the contract text with your initials.

It's a tradeoff: it'd be great to have a very short ToS, but simultaneously, I think it'd be awesome to explain in the ToS how users' data is used, stored, and where it's transferred. Unfortunately you can't have both (short and comprehensive), and the UI/UX design team has said that the ToS currently here (which is already very minimal as ToSs go) is already way too long. We can only fit around 2 paragraphs + a link on the current spot.

I think the solution to that is to have a condensed version with the main points on the signup form, with a link to the full policy. But we will have to do that later.

> That's tricky; let's stick a version or date in the beginning of the ToS so that the user can verify it hasn't been updated since she last read it.

Good idea, I'll add that.

kalvdans (commented):
Some suggestions for shortening the ToS

kalvdans:
> Good idea, I'll add that.

Yes, let's leave the atomicity for later since it wasn't as simple as I thought.

aapeliv:
Shortened a bit and added a date + version.

kalvdans (approved):

PR #1188, opened 2021-05-06
Fix docker-compose restart policy

aapeliv:
The option changed between v2 and v3 docker compose files


kalvdans (dismissed):
didn't try but if you say so :)

only version 2 needs annotating in the file, google says.

aapeliv:
Thanks, what I did there was wrong. I fixed it up now

kalvdans (approved):
lgtm!

PR #1190, opened 2021-05-06
User deletion and blocking

### LINKED ISSUE #414: User blocking

### LINKED ISSUE #610: User hiding

### LINKED ISSUE #1092: Can get users by email
You shouldn't be able to do this!

https://develop--frontend.preview.couchershq.org/user/[email]

We need to review where `get_user_by_field` is used, and phase it out except for in auth. We need to figure out how the frontend should translate usernames to user IDs as well. In particular, we don't want two trips to the backend for getting a user's profile. At least it needs another function

aapeliv:
Takes over from Sam's #715, #724, #1122, and  #1163, we cleaned this up together. Closes #414, and closes #610.

This adds a new query object that adds some filtering options. For now we grab the blocked users each time, but this can later be cached per-request into `context`, or something.

We also removed `get_user_by_field`, and removed the ability to get a user by email (`GetUser`), so closes #1092.

Also updated most `*_to_pb` style functions to take full `context` instead of just `user_id`.

**Backend checklist**
- [x] All tests pass


app/pb/blocking.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package org.couchers.blocking;
+
+import "google/protobuf/empty.proto";
+
+service Blocking {
```
kalvdans, line 7: Describe that you have to be logged in to use this service. I guess we should allow jailed users to update blocking.

app/backend/src/tests/test_fixtures.py
```diff
@@ -95,7 +97,7 @@ def db():
     recreate_database()
 
 
-def generate_user(*_, **kwargs):
+def generate_user(*_, delete=False, **kwargs):
```
kalvdans: Name the flag invisible to better reflect the actual effect. Document it in the docstring.

Also, a drive-by fix is to replace `*_, ` with `*, ` to make sure everyone uses keywords when calling it.

app/backend/src/couchers/models.py
```diff
@@ -239,6 +243,60 @@ def __repr__(self):
         return f"User(id={self.id}, email={self.email}, username={self.username})"
 
 
+# CouchersQuery is placed here (instead of the more natural db) because we import stuff from models into db, so need to
+# avoid circular imports. Additionally, we need it in the relationship configuration, but it requires the User model, so
+# it's as high up as we can get it, and then we patch the User relationships in manually.
```
kalvdans: Instead of creating a CouchersQuery just create standalone functions that takes a query as the first argument. Will be a bit different/uglier when using it though, mixing yoda speak with regular speak:

```
                filter_users_column(session.query(Message.id)
                    .join(HostRequest, Message.conversation_id == HostRequest.conversation_id)
                    .outerjoin(
                        message_2, and_(Message.conversation_id == message_2.conversation_id, Message.id < message_2.id)
                    )
                    .filter(HostRequest.from_user_id == context.user_id),
                        context, HostRequest.to_user_id)
                .filter(message_2.id == None)
                .filter(HostRequest.from_last_seen_message_id < Message.id)
                .count()
```

Or introduce a third file to avoid circular imports.

app/backend/src/couchers/models.py
```diff
@@ -1258,3 +1316,37 @@ def ready_for_retry(self):
 
     def __repr__(self):
         return f"BackgroundJob(id={self.id}, job_type={self.job_type}, state={self.state}, next_attempt_after={self.next_attempt_after}, try_count={self.try_count}, failure_info={self.failure_info})"
+
+
+class UserBlocks(Base):
```
kalvdans: traditionally, the class name is singular while the table is plural. No need to stick to it though!

app/backend/src/couchers/models.py
```diff
@@ -1258,3 +1316,37 @@ def ready_for_retry(self):
...
+
+    blocking_user_id = Column(ForeignKey("users.id"), nullable=False)
+    blocked_user_id = Column(ForeignKey("users.id"), nullable=False)
+    time_blocked = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
+
+    is_blocking_user = relationship("User", backref="is_blocking_user", foreign_keys="UserBlocks.blocking_user_id")
+    is_blocked_user = relationship("User", backref="is_blocked_user", foreign_keys="UserBlocks.blocked_user_id")
+
+
+def blocked_users(session, user_id):
```
kalvdans: Move to db.py?

app/backend/src/couchers/models.py
```diff
@@ -149,6 +150,7 @@ class User(Base):
     additional_information = Column(String, nullable=True)  # CommonMark without images
 
     is_banned = Column(Boolean, nullable=False, default=False)
+    is_deleted = Column(Boolean, nullable=False, default=False)
```
kalvdans, line 152: I think we should name this column `is_invisible` to indicate that it can be resurrected at any time.

app/backend/src/couchers/servicers/blocking.py
```diff
@@ -0,0 +1,70 @@
...
+                    blocked_user_id=request.user_id,
+                )
+                session.add(user_block)
+                session.commit()
+
+        return empty_pb2.Empty()
+
+    def UnblockUser(self, request, context):
+        if context.user_id == request.user_id:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.CANT_UNBLOCK_SELF)
```
kalvdans: Redundant, will be caught by USER_NOT_BLOCKED further down anyway.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -19,7 +19,7 @@
 MAX_PAGINATION_LENGTH = 25
 
 
-def _parents_to_pb(node_id, user_id):
+def _parents_to_pb(node_id, context):
```
kalvdans: Remove argument instead, it seems to be unused?

kalvdans (changes requested):

app/pb/blocking.proto
aapeliv: Good catch

app/backend/src/couchers/models.py
aapeliv: "invisible" means that it's not visible: but you can be invisible if you've been banned or you can be invisible if you've been deleted. There's the `is_visible` helper function that combines both of these.

I think it's good to keep this as `is_deleted` to clarify the purpose of the flag.

app/backend/src/couchers/models.py
aapeliv: I really think `CouchersQuery` makes everything *a lot* nicer, cleaner, and more concise than having separate functions that you have to wrap around your query. In some of the bigger queries, they just become a huge mess otherwise, especially if you have to use multiple filters.

> Or introduce a third file to avoid circular imports.

The problem is that:
* `CouchersQuery` requires the `User` class to be defined
* `relationship` requires `CouchersQuery` to be defined
* Basically all classes require `relationship` to be defined

I'm not sure how to get around this, but it'd be really nice to find a cleaner way.

Maybe if we split `User` into another file? The split of `models.py` is long overdue, but those PRs never seem to go through and somehow become stale so the merges are a pain...

app/backend/src/couchers/models.py
aapeliv: It requires `UserBlock` from this file, so that'd cause circular imports...

app/backend/src/couchers/models.py
kalvdans: Fair enough!

app/pb/blocking.proto
```diff
@@ -0,0 +1,29 @@
...
+    //Returns all users blocked by a given account
+  }
+}
+
+message BlockUserReq {
+  int64 user_id = 1;
+}
+
+message UnblockUserReq {
+  int64 user_id = 1;
```
aapeliv: @kalvdans: I think these `user_id`s should be changed to `string username`.

There's no way for the requester to get the user_id because `GetUser` will return not found for blocked users.

Also if user A blocks user B, then user B might also want to block user A, so it's not guaranteed that a the blocker can find the user id either

kalvdans: If I block you, I should still be able to GetUser you IMHO. But indeed if we have a mutual blockage I guess I have to call you and ask which user_id number you have.

app/backend/src/couchers/query.py
```diff
@@ -0,0 +1,59 @@
+from sqlalchemy import or_
+from sqlalchemy.orm import Query, aliased
+
+from couchers.models import User, UserBlock
+from couchers.utils import is_valid_email, is_valid_user_id, is_valid_username
+
+
+class CouchersQuery(Query):
+    def _blocked_users(self, session, user_id):
```
kalvdans: move to top level, no need to have it as a member function. It doesn't chain another filter onto the query.

app/backend/src/couchers/query.py
```diff
@@ -0,0 +1,59 @@
...
+        # no fields match, this will return no rows
+        return self.filter(False)
+
+    def filter_users(self, context, table=User):
+        """
+        Filters out users that should not be visible: blocked, deleted, or banned
+
+        Filters the given table, assuming it's already joined/selected from
+        """
+        hidden_users = self._blocked_users(self.session, context.user_id)
```
kalvdans: This call does a full sql query to the database, it would be better if it could augment the current query with the right filters. But it could be an improvement to be done later.

app/pb/blocking.proto
aapeliv: I think it's better to just use username here.

Regarding getting the user with GetUser; there should maybe be an option to get some subset of information. But I don't think it's very good that the blocker has the power to see the other user but the other user can't see the blocker, could be abused.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -35,22 +35,52 @@ def _parents_to_pb(node_id, user_id):
...
-        can_moderate = can_moderate_node(session, user_id, node.id)
+        can_moderate = can_moderate_node(session, context.user_id, node.id)
+
+        member_count = (
+            session.query(ClusterSubscription)
+            .filter_users_column(context, ClusterSubscription.user_id)
+            .filter(ClusterSubscription.cluster_id == node.official_cluster.id)
+            .count()
+        )
+        member = (
```
kalvdans: is_member

app/backend/src/couchers/servicers/communities.py
```diff
@@ -35,22 +35,52 @@ def _parents_to_pb(node_id, user_id):
...
+        )
+
+        admin_count = (
+            session.query(ClusterSubscription)
+            .filter_users_column(context, ClusterSubscription.user_id)
+            .filter(ClusterSubscription.cluster_id == node.official_cluster.id)
+            .filter(ClusterSubscription.role == ClusterRole.admin)
+            .count()
+        )
+        admin = (
```
kalvdans: is_admin

app/backend/src/couchers/utils.py
```diff
@@ -14,6 +15,47 @@
...
+    """
+    Checks if it has at least one non-whitespace character
+    """
+    return re.match(r"\S+", field) is not None
+
+
+def is_valid_email(field):
+    """
+    From SO
+    """
```
kalvdans: comment instead of docstring

kalvdans (dismissed):
Small remarks, but looks good in general!

app/backend/src/couchers/query.py
aapeliv: I'll leave it for later

app/pb/blocking.proto
kalvdans: I guess one can argue either way. Symmetrical blocking is conceptually simpler indeed.

OK, please go with usernames here and in GetBlockedUsers

aapeliv:
All done

kalvdans (dismissed):
LGTM! Please merge when CI is happy!

PR #1191, opened 2021-05-07
Show map user info even without search query

lucaslcode:
Otherwise the new map search is a regression

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/features/search/SearchPage.tsx
+++ b/app/frontend/src/features/search/SearchPage.tsx
@@ -61,6 +61,12 @@ export default function SearchPage() {
     undefined
   );
 
+  //callbacks provided to the map need a ref, or they won't get the latest value of selectedResult
+  const selectedResultRef = useRef<number | undefined>(undefined);
+  useEffect(() => {
+    selectedResultRef.current = selectedResult;
+  }, [selectedResult]);
+
   const showResults = useRef(false);
 
   const { query } = useParams<{ query?: string }>();
@@ -131,10 +137,10 @@ export default function SearchPage() {
   ) => {
     //if undefined, unset
     if (!user) {
-      if (selectedResult) {
+      if (selectedResultRef.current) {
         //unset the old feature selection on the map for styling
         map.current?.setFeatureState(
-          { source: "all-objects", id: selectedResult },
+          { source: "all-objects", id: selectedResultRef.current },
           { selected: false }
         );
         setSelectedResult(undefined);
@@ -143,11 +149,11 @@ export default function SearchPage() {
     }
 
     //make a new selection if it has changed
-    if (selectedResult !== user.userId) {
-      if (selectedResult) {
+    if (selectedResultRef.current !== user.userId) {
+      if (selectedResultRef.current) {
         //unset the old feature selection on the map for styling
         map.current?.setFeatureState(
-          { source: "all-objects", id: selectedResult },
+          { source: "all-objects", id: selectedResultRef.current },
           { selected: false }
         );
       }
--- a/app/frontend/src/features/search/SearchPage.tsx
+++ b/app/frontend/src/features/search/SearchPage.tsx
@@ -1,16 +1,16 @@
 import { Collapse, Hidden, makeStyles, useTheme } from "@material-ui/core";
 import Map from "components/Map";
 import SearchBox from "features/search/SearchBox";
-import { LngLat, Map as MaplibreMap } from "maplibre-gl";
+import { EventData, LngLat, Map as MaplibreMap } from "maplibre-gl";
 import { User } from "pb/api_pb";
-import React, { useEffect, useRef, useState } from "react";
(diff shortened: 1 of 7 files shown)


app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -61,13 +52,25 @@ export default function SearchPage() {
     undefined
   );
 
+  //callbacks provided to the map need a ref, or they won't get the latest value of selectedResult
+  const selectedResultRef = useRef<number | undefined>(undefined);
+  useEffect(() => {
+    selectedResultRef.current = selectedResult;
+  }, [selectedResult]);
```
lucaslcode: This took me a while...

darrenvong: This defeats the purpose of using state and we're essentially mixing imperative and declarative ways of maintaining the same thing - we should choose one side and stick to it. Is it because we aren't consistently updating state to cause re-renders when it should?

darrenvong (changes requested):
~~I'm still trying to understand the bug behind this - is this referring to when you have nothing in the search bar, you're still seeing user pins on the map?~~

Never mind, I think I got it now - the pins weren't showing any user profiles if you just click on one without searching 👍  For some reason the PR title completely went over my ahead until I played around with it and makes sense now

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -113,16 +117,34 @@ export default function SearchPage() {
       }
       addUsersToMap(newMap, handleMapUserClick);
     });
+    newMap.on("click", (e) => {
+      if (!e.defaultPrevented) {
+        handleResultClick(undefined);
+      }
```
darrenvong: Does this only get registered once when the map is initialised? i.e. in subsequent renders, this will continue to be used until the map is destroyed? That might also explain the outdated state the callback gets as it only keeps seeing the state value from the closure it's created in 🤔

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: As you found in the other comment - the state value gets 'frozen' when the map is created because that's when the callback is set. What a good way to clean this up?

darrenvong: If relying on state, usually the way to do it would be to clean up and re-register the event handlers in a `useEffect` whenever the state updates to ensure the callback gets the latest value, but that doesn't feel that different to what you're doing and it's (marginally?) more code...

I'm thinking it might be better to switch full imperative and hold that state in a ref and update it ourselves in the handlers in places where you set state? Or would that cause problems for when there is a query? My thinking is the map itself is already imperative and Mapbox is clearly doing some state management of the map on their end (to display the right pins) and it doesn't look like it'd play well with React state without these hacks...

lucaslcode: Only problem is, the normal state is used too, to show which card is selected. If it's only in a ref, it won't re-render.

darrenvong: I see, because that part is a declarative React component... the joy of having to mix the two together ☹️ 

In that case, I think I'd prefer cleaning up and re-subscribing the event handler whenever state updates, since that's the kind of things `useEffect` is designed for and so fits closer to the React way of doing things than ["swimming against the tide"](https://overreacted.io/a-complete-guide-to-useeffect/#swimming-against-the-tide).

This also lets the state be the single source of truth in this component so we don't have to figure out in certain bits, we have to use a "different copy" of the state (when they should be the same thing).

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -62,12 +55,18 @@ export default function SearchPage() {
...
-    setTimeout(
-      () => map.current?.resize(),
-      theme.transitions.duration.standard
-    );
-  }, [query, theme.transitions.duration.standard]);
+    const shouldShowResults = !!query || !!selectedResult;
+    if (showResults.current !== shouldShowResults) {
+      showResults.current = shouldShowResults;
+      setTimeout(
+        () => map.current?.resize(),
```
darrenvong, line 65: Is this Mapbox magic that works out the right zoom level when the number of users available on the map changes?

darrenvong (commented):
Just waiting for the build to go green to have a final look, but looks good to me otherwise!

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: No - it's just necessary to call when the size of the canvas changes, otherwise you have tiles which aren't loaded

lucaslcode: And the timeout is because of the collapse animation, to make sure happens at the end

darrenvong (approved):

PR #1192, opened 2021-05-07
Clean up onboarding emails to use urls.py

aapeliv:
**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/urls.py
+++ b/app/backend/src/couchers/urls.py
@@ -1,4 +1,16 @@
+"""
+Please make sure this file stays in sync with
+//app/frontend/src/routes.ts
+"""
 from couchers.config import config
+
+
+def app_link():
+    return f"{config['BASE_URL']}/"
+
+
+def profile_link():
+    return f"{config['BASE_URL']}/user"
 
 
 def signup_link(signup_token):
--- a/app/backend/src/couchers/urls.py
+++ b/app/backend/src/couchers/urls.py
@@ -1,7 +1,8 @@
-"""
-Please make sure this file stays in sync with
-//app/frontend/src/routes.ts
-"""
+# The source of truth for URLs is
+# //docs/urls.md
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/urls.py
```diff
@@ -1,6 +1,18 @@
+"""
+Please make sure this file stays in sync with
+//app/frontend/src/routes.ts
+"""
```
kalvdans: No, //docs/urls.md is the primary source of URL information that both frontend and backend should refer to.

Please describe it with a comment instead of a docstring as it only makes sense to developers reading the source, not programmers who want to use the functions therein.

app/frontend/src/routes.ts
```diff
@@ -1,3 +1,7 @@
+/*
+Please make sure this file stays in sync with
+//app/backend/src/couchers/urls.py
+*/
```
kalvdans, line 6: No, //docs/urls.md is the primary source of URL information that both frontend and backend should refer to.

app/backend/templates/onboarding2.md
```diff
@@ -27,5 +27,5 @@ We would really appreciate it if you could add a photo and fill in at least the
 
 Thank you so much!
 
-Emily from Couchers.org
+Emily from Couchers.org
```
kalvdans: revert trailing whitespace change.

kalvdans (changes requested):
Just some small nitpicks, auto-LGTM after they are fixed!

app/backend/templates/onboarding2.md
aapeliv: It's ugly markdown stuff to get a newline

kalvdans: [confirmed.](https://spec.commonmark.org/0.29/#hard-line-breaks)

PR #1195, opened 2021-05-09
Pass number of responses along whenever we return a thread_id

### LINKED ISSUE #1144: Return the number of the replies/comments in the discussion object in the API
Screenshot 2021-04-26 at 03 10 49

In the design above, we want to show the number of replies a discussion has. This discussion card/component is used to in the discussions list page/tab.

Currently there is no way to get this information directly without an additional request to get the thread to figure this out, even when we don't need the rest of the thread's data at this stage.

Also, even if the additional request is not an issue, this is requiring the frontend to do more data wrangling than it should to figure this out, when I think it would be cleaner to do this in the backend instead. Apparently @kalvdans [agrees here](https://github.com/Couchers-org/couchers/pull/578#discussion_r573989349) so maybe this isn't a terrible idea 😉

kalvdans:
Closes https://github.com/Couchers-org/couchers/issues/1144.

**Backend checklist**
- ~~[ ] All tests pass~~ not needed since we are Backend&trade;


aapeliv:
How about adding a submessage like this:

```proto
message Thread {
  int64 thread_id = 1;
  int64 num_responses = 2;
}
```

Then in `Page`, we could do

```proto
message Page {
  // ...
  Thread thread = 20;
  // ...
}
```

That way we can have a `thread_to_pb` and reuse it everywhere and be able to add any additional things to the thread as well, as needed. In the frontend too, we can then just pass that over to whatever component needs it.

I think that would be cleaner and keep the thread object more modular?

kalvdans:
> How about adding a submessage like this:

Good idea, I'll make the suggested change and see how it looks like.

kalvdans:
Looks nicer @aapeliv ? Also I'm frontend analphabet so need help to adapt the frontend :white_flag:

app/pb/threads.proto
```diff
@@ -4,6 +4,14 @@ package org.couchers.api.threads;
...
 import "google/protobuf/timestamp.proto";
 
+// object returned by pages, discussions etc to refer to top level threads.
+message TopLevelThread {
+  // top level thread id.
+  int64 thread_id = 1;
+  // total number of comments and replies to comments.
+  int64 num_responses = 2;
+}
+
```
aapeliv: Move this under the service definition.

I would also just call it `Thread` and specify it's top level in a comment

app/pb/threads.proto
```diff
@@ -51,9 +59,6 @@ message GetThreadRes {
   // sorted in inverse time order, latest reply first.
   repeated Reply replies = 1;
 
-  // number of responses (comments and replies) associated with the thread
-  uint32 num_responses = 2;
-
```
aapeliv, line 56: So `TopLevelThread` returns total number of responses for the recursive thread, and `Reply` contains the num_replies for comments?

aapeliv (commented):
Just a few small comments, looks good in general

aapeliv:
Regarding merge: let's leave the frontend as-is and let @darrenvong merge/rebase this into his work. There's no point going to the old discussion stuff and changing that, it'll just cause pain and suffering in merge for the new discussion stuff.

app/pb/threads.proto
kalvdans: Yes, exactly. There's no way to get the number of comments (without counting replies to comments), but that could be one thing to be added to TopLevelThread in the future.

aapeliv: That's not a problem, I think. I don't think we need that (# comments without replies).

The main thing was to make sure we return number of replies for each comment as well.

kalvdans:
> Regarding merge: let's leave the frontend as-is and let @darrenvong merge/rebase this into his work. There's no point going to the old discussion stuff and changing that, it'll just cause pain and suffering in merge for the new discussion stuff.

Ok! Life is easy being a backend developer ;)

aapeliv (dismissed):
LGTM!

Let's leave it to @darrenvong to take it from here and integrate into what he's doing.

aapeliv:
> * ~[ ] All tests pass~ not needed since we are Backend™

🤣

darrenvong:
Pending rebuild, the frontend itself and the tests should be fixed now (within this branch) to read from the nested thread object, so you can merge to develop with this

aapeliv (approved):
LGTM, I'll merge if tests pass!

I thought it'd be more messy and mess up your PRs as well. Thanks for fixing!

darrenvong:
Same, glad they were only small tweaks 😌

PR #1196, opened 2021-05-09
Regions and languages v2

aapeliv:
Add regions/languages RPCs and move into resources instead of constant_data

**Backend checklist**
- [ ] All tests pass


kalvdans:
Lots of unrelated changes in the commit list of this MR. Did you mess up rebasing? Why not continue in https://github.com/Couchers-org/couchers/pull/926 ?

aapeliv:
> Did you mess up rebasing?

Merged `develop` into this



> Why not continue in #926 ?

In case Thomas had based work on it

kalvdans:
I suggest rebasing your commits on top of `backend/feature/regions-and-languages` and force-push, to get a reviewable state.

aapeliv:
@kalvdans: merged `develop` into the other branch, and added just this one commit here.

app/backend/src/couchers/resources.py
```diff
@@ -0,0 +1,65 @@
...
+    Get list of allowed languages as a set.
+    """
+    return set(get_language_dict.keys())
+
+
+def language_is_allowed(code):
+    """
+    Check a language code is valid
+    """
+    return code in get_language_set()
```
kalvdans: Will be equally fast with `code in get_language_dict()` and you can remove `get_language_set()`.

app/pb/resources.proto
```diff
@@ -12,8 +12,34 @@ service Resources {
...
+  string alpha3 = 1;
+  string name = 2;
+}
+
+message GetRegionsRes {
+  repeated Region regions = 1;
+}
+
+message Language {
+  string code = 1;
```
kalvdans:
```suggestion
  string code = 1;  // ISO639-3 lowercase language code
```

app/pb/resources.proto
```diff
@@ -12,8 +12,34 @@ service Resources {
   rpc GetTermsOfService(google.protobuf.Empty) returns (GetTermsOfServiceRes) {
     // Returns the latest Terms of Service markdown, to be shown on signup or if jailed due to this
   }
+
+  rpc GetRegions(google.protobuf.Empty) returns (GetRegionsRes) {
+    // Returns a list of allowed regions
+  }
+
+  rpc GetLanguages(google.protobuf.Empty) returns (GetLanguagesRes) {
+    // Returns a list of allowed languages
```
kalvdans:
```suggestion
    // Returns a list of allowed languages. This is a subset of the full ISO639-3 table.
```

app/backend/src/couchers/resources.py
```diff
@@ -0,0 +1,65 @@
...
+import json
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+resources_folder = Path(__file__).parent / ".." / ".." / "resources"
+
+
+@functools.cache
```
kalvdans, line 11: File will cached the filesystem caches anyway if it is frequently read, not sure if we need to cache it in memory as well. It's small though so probably doesn't matter.

app/backend/src/couchers/resources.py
```diff
@@ -0,0 +1,65 @@
...
+    Get list of allowed regions as a set.
+    """
+    return set(get_region_dict.keys())
+
+
+def region_is_allowed(code):
+    """
+    Check a region code is valid
+    """
+    return code in get_region_set()
```
kalvdans: `code in get_region_dict()` works too

kalvdans (changes requested):

app/backend/src/couchers/servicers/resources.py
```diff
@@ -1,16 +1,21 @@
...
 
 class Resources(resources_pb2_grpc.ResourcesServicer):
     def GetTermsOfService(self, request, context):
-        with open(resources_folder / "terms_of_service.md", "r") as f:
-            return resources_pb2.GetTermsOfServiceRes(terms_of_service=f.read())
+        return resources_pb2.GetTermsOfServiceRes(terms_of_service=get_terms_of_service())
+
+    def GetRegions(self, request, context):
+        return resources_pb2.GetRegionsRes(
+            regions=[resources_pb2.Region(alpha3=alpha3, name=name) for alpha3, name in get_region_dict().items()]
```
kalvdans: Throw in a sorted() as well to get a deterministic order, not depending on Python's randomized hash function.

kalvdans (approved):
just a small comment, auto-lgtm after that :)

PR #1197, opened 2021-05-09
Discussion part 4: comments, nested comments/replies + ability to comment on discussion and comments

### LINKED ISSUE #543: Discussions

darrenvong:
Covers more of #543.

Discussion feature as described in PR title.

I'll add screenshots later. Tests left to do but feel free to do an early review if you got time

**Frontend checklist**
- [x] All tests pass


lucaslcode:
With the URLs, looking at `docs/URLs.py` what you have now is correct. But the community (aka official cluster) id is in the database, so it should be trivial to include it in `GetDiscussion`, so in that case I think we can put the discussion page inside the community. Dunno if you want to do it in this PR or not?

darrenvong:
How would this work with the URL matching though, since the community pages start with `/community` and a single discussion starts with `/discussion`? I'm guessing we'll probably have to duplicate or split the common bits into a shared component so both community and discussion can use it?

Also how would things look roughly layout wise? Instead of the tabs with the different things, we replace it with the discussion in the same place if we are on a single discussion?

lucaslcode:
Yes, I guess the community page would need some refactoring so it can take a child?
The top bit of the community page (everything up to line 190) can be `CommunityBase.tsx`, and it can optionally take a `communityId` prop - if it doesn't have it, it gets the id from the url. And then put `{ children }` just below it.

Then the real community page can have `<CommunityBase>/*<Switch>s here*/</CommunityBase>` and discussions page can be `<CommunityBase communityId={discussion.communityId}>/*discussion page*/</CommunityBase>`

lucaslcode:
Oh and I didn't answer your question - I think what would make the most sense is the tab bar is still there, just none of them are selected

darrenvong:
Hmm I think I'll do the page refactor in another PR, as I imagine this one will get pretty huge otherwise with big diffs as is usually the case when refactoring/shifting things around

lucaslcode:
Didn't notice last time - the cards on the community page are slightly broken on mobile: 
image

lucaslcode:
~~I think the button to add a nested comment should move outside the card and at the bottom of any other nested comments:~~ 
image
~~Otherwise, it's not intuitive what happens when there are already nested comments (like in this image).~~

I see, nested comments are also latest first. Hmm. ~~I wonder if they should be oldest first? That might require backend changes.~~

After thinking about it, how about making the latest comments at the bottom, and displaying a "load earlier comments" link at the top? And indeed moving the button as I have in the image.

There also seems to be a bit too much bottom padding in the cards

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const postedTime = timeAgo(replyDate, false);
+
+  return (
+    <>
+      <Card
+        className={classes.commentContainer}
+        data-testid={COMMENT_TEST_ID}
+        key={comment.createdTime!.seconds}
+      >
+        <Avatar user={user} className={classes.avatar} isProfileLink={false} />
```
lucaslcode:
```suggestion
        <Avatar user={user} className={classes.avatar} />
```

app/frontend/src/features/communities/discussions/CommentForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+  const {
+    error,
+    isLoading,
+    mutate: postComment,
+    reset: resetMutation,
+  } = useMutation<PostReplyRes.AsObject, GrpcError, CommentData>(
+    ({ content }) => service.threads.postReply(threadId, content),
+    {
+      onSuccess() {
+        queryClient.invalidateQueries(threadKey(threadId));
```
lucaslcode, line 62: I can't remember what you decided last time, but this is suffering from the same problem where on a slow connection, the loading state finishes when the new comment is posted but before it actually appears.

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -0,0 +1,122 @@
...
+    height: "3rem",
+    gridArea: "avatar",
+    width: "3rem",
+  },
+  nestedCommentsContainer: {
+    "& > * + *": {
+      marginBlockStart: theme.spacing(2),
+    },
+    marginBlockStart: theme.spacing(2),
+    marginInlineStart: theme.spacing(5),
```
lucaslcode: How about
```suggestion
    marginInlineStart: theme.spacing(3),
    marginInlineStart: `clamp(${theme.spacing(2)}, 5vw, ${theme.spacing(5)})`,
```
Mobile is most likely not to support clamp, so I made the fallback a bit less

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -21,24 +21,22 @@ import { CREATED_AT, PREVIOUS_PAGE, UNKNOWN_USER } from "../constants";
 import CommentTree from "./CommentTree";
 
 const useStyles = makeStyles((theme) => ({
+  root: {
+    paddingBlockEnd: theme.spacing(5),
+    "& .tui-editor-contents": {
+      marginBottom: theme.spacing(3),
```
lucaslcode: This should be in the relevant components (or more specific) rather than here. It's appropriate for the discussion content, but it's also affecting the cards, where it's too much margin (imo)

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -96,8 +94,13 @@ export default function DiscussionPage() {
                 {discussion.title}
               </PageTitle>
             </div>
+            <Markdown source={discussion.content} />
             <div className={classes.creatorContainer}>
-              <Avatar user={discussionCreator} className={classes.avatar} />
+              <Avatar
+                user={discussionCreator}
+                className={classes.avatar}
+                isProfileLink={false}
```
lucaslcode: remove

lucaslcode (changes requested):
Haven't looked through all the code yet - these are just some design comments

darrenvong:
> I think the button to add a nested comment should move outside the card and at the bottom

Basing off Reddit and Facebook, they all have the reply button right under/near the comment you are replying to, so not too sure about putting it under the last nested comment/reply. That could be unexpected behaviour for users who are used to other sites. What's your main objection on keeping the button inside the comment card in that case?

> how about making the latest comments at the bottom, and displaying a "load earlier comments" link at the top?

I know we did for chat, but since these comments also has a reply button (i.e. something keyboard focusable), I'm not sure it's a good idea to do the same CSS trick of reversing the display order of the comments as keyboard users would probably end up going from bottom to top if they tab between buttons. Getting that to work "out of the box" as we do now might be a bit of a pain.

Reversing the data is a potential option, but feels like extra work that doesn't need to be done by the frontend if the backend gives it in the right order to start with. And the operation is linear time so it's gonna make our page slower as it grows (although unlikely to be an immediate problem for some time) so probably better done in the backend.

lucaslcode:
The difference is, reddit has infinite reply levels, and clicking reply adds another level. Here, it will only go one level deep - I think the difference is important.

Facebook does have the reply button on the comment, but when you click it, it opens the comment box underneath all the other comments. We could do that!

> keyboard users would probably end up going from bottom to top

Good point. Let's reverse it in the backend then - changing `.desc()` to `.asc()` in `threads.py` lines 67 and 90 should be enough, but it might break some tests, so leave it to another PR if you want. But treat it as if it has been done?

app/frontend/src/features/communities/discussions/CommentForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+  testId: string;
+  threadId: number;
+}
+
+interface CommentData {
+  content: string;
+}
+
+export default function CommentForm({
+  shown = false,
```
lucaslcode: Would this be better as `alwaysShow` or `toggleable`?

app/frontend/src/features/communities/discussions/CommentTree.tsx
```diff
@@ -57,52 +53,39 @@ export default function CommentTree({ threadId }: CommentTreeProps) {
...
-                  <div className={classes.commentContent}>
-                    <Typography variant="body2">
-                      {getByCreator(user?.name ?? UNKNOWN_USER)}
-                      {` • ${postedTime}`}
-                    </Typography>
-                    <Typography variant="body1">{reply.content}</Typography>
-                  </div>
-                </Card>
+                <Comment
+                  key={comment.createdTime?.seconds}
```
lucaslcode: Why not `threadId`? There could easily be comments posted in the same second.

lucaslcode (changes requested):

app/frontend/src/features/communities/discussions/CommentTree.tsx
darrenvong: That's a good point, `threadId` is much better - I remember I was scrambling to find something unique but probably overlooked this with the whole thread/comment confusion at the beginning 😅

app/frontend/src/features/communities/discussions/DiscussionPage.tsx
darrenvong: I blame the brutal merge conflict with too many changes 😤   but good catch 😄

darrenvong:
> Good point. Let's reverse it in the backend then - changing `.desc()` to `.asc()` in `threads.py` lines 67 and 90 should be enough, but it might break some tests, so leave it to another PR if you want. But treat it as if it has been done?

Wouldn't that break the pagination logic though, as the limit bit might be dependent on it? I think I'll leave this to the backend folks for that one... but yeah happy to treat it as done for the sake of this PR

>  it opens the comment box underneath all the other comments. We could do that!

I'll have a play around with that to see how much more complex that might make it than just placing it near the bottom and probably go with whatever's simpler. The extra complexity being needing to keep refs around probably to scroll to the comment box, but it might not be too bad

app/frontend/src/features/communities/discussions/CommentForm.tsx
darrenvong: We went with showing a refetch loading state instead, since optimistically updating the data in an infinite query is much more of a pain than a normal query. There is one in `Comment`, but I may have to adjust the position of it to be at the bottom if we are going with having the comment input near the last comment.

lucaslcode: I see it now in storybook. For some reason I didn't before, maybe I just missed it.

app/frontend/src/features/communities/discussions/Comment.tsx
darrenvong: That's a nice trick actually rather than needing media queries. Still had to do an annoying workaround with '&' to select the same thing again and put in the clamp since you can't have duplicate object keys :(

aapeliv:
> > Good point. Let's reverse it in the backend then - changing `.desc()` to `.asc()` in `threads.py` lines 67 and 90 should be enough, but it might break some tests, so leave it to another PR if you want. But treat it as if it has been done?
> 
> Wouldn't that break the pagination logic though, as the limit bit might be dependent on it? I think I'll leave this to the backend folks for that one... but yeah happy to treat it as done for the sake of this PR

See #1223

darrenvong:
I think I addressed all the comments now apart from needing to add the "load earlier comments" button somewhere. Feel free to have another look through in between

darrenvong:
Ok, list of messages is now reversed in the frontend, and the load more button is done now.

If you manage to check before I have more time to add Storybook (need to hack the mock a bit to deal with pagination), you'll probably have to pull this down locally and change the `pageSize` inside getThread to something smaller temporarily to see the button, but otherwise the functionality is pretty much there now

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const isCommentsRefetching = !isCommentsLoading && isCommentsFetching;
+  const showLoadMoreButton = topLevel && hasNextPage;
+
+  const [showCommentForm, setShowCommentForm] = useState(false);
+  const commentFormRef = useRef<HTMLFormElement>(null);
+
+  useEffect(() => {
+    if (showCommentForm && commentFormRef.current) {
+      if (isSafari()) {
+        commentFormRef.current.scrollIntoViewIfNeeded(true);
```
darrenvong: Safari doesn't support smooth scrolling AND scrolling element to centre of the screen even with the polyfill, but calling `scrollIntoViewIfNeeded(true)` does it but without smooth scrolling. I thought for Safari only, it's worth giving up smooth scrolling for this, otherwise the input always scroll to top (default option) of screen and so the input gets clipped under the nav bar, which is probably more annoying than not having smooth scrolling imo.

darrenvong: Also since I'm not measuring anything and probably because of the 1-2s delay caused by the animation revealing the input, I managed to get away with using `useEffect` to do this rather than with `useLayoutEffect` :D

lucaslcode: Maybe we need a better polyfill - does this one work? https://github.com/magic-akari/seamless-scroll-polyfill

darrenvong: Haven't tried yet but from looking at the code, it looks more promising and comprehensive than the original polyfill we are using now (which this is a fork of!)

darrenvong: That other polyfill worked like a charm, nice find!

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <>
+              {comments.pages
+                .flatMap((page) => page.repliesList)
+                .reverse()
+                .map((reply) => {
+                  return <Comment key={reply.threadId} comment={reply} />;
+                })}
+              {!showLoadMoreButton && isCommentsRefetching && (
+                <CircularProgress />
+              )}
```
lucaslcode: Aha! I wasn't going crazy - this doesn't appear for the first nested comment, because it is inside `hasAtLeastOnePage`

app/frontend/package.json
```diff
@@ -23,7 +23,7 @@
     "react-query": "^3.15.2",
     "react-router-dom": "^5.2.0",
     "react-scripts": "4.0.3",
-    "smoothscroll-polyfill": "^0.4.4",
```
lucaslcode, line 26: also remove the types dependency

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -0,0 +1,175 @@
...
+              {showLoadMoreButton && (
+                <Button
+                  className={classes.loadMoreRepliesButton}
+                  loading={isFetchingNextPage}
+                  onClick={() => fetchNextPage()}
+                >
+                  {LOAD_MORE_REPLIES}
+                </Button>
```
lucaslcode: I was imagining this would also be a "Load earlier replies" button, but maybe it's not so important

lucaslcode (changes requested):
It's great, just a few very minor things

app/frontend/src/features/communities/discussions/Comment.tsx
```diff
@@ -145,16 +149,13 @@ export default function Comment({ topLevel = false, comment }: CommentProps) {
                 .map((reply) => {
                   return <Comment key={reply.threadId} comment={reply} />;
                 })}
-              {!showLoadMoreButton && isCommentsRefetching && (
-                <CircularProgress />
-              )}
               {showLoadMoreButton && (
                 <Button
```
lucaslcode, line 147: Can this also be moved to above the list of replies?

lucaslcode (changes requested):
Last thing

lucaslcode (approved):
lgtm pending checks!

PR #1198, opened 2021-05-09
Idea for CouchersQuery

aapeliv:
@kalvdans: here's an idea to clean up the circular imports/definition ordering, what do you think?

### DIFF
--- a/app/backend/src/couchers/query.py
+++ b/app/backend/src/couchers/query.py
@@ -1,0 +1,65 @@
+from sqlalchemy import or_
+from sqlalchemy.orm import Query, aliased
+
+
+class CouchersQuery(Query):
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/query.py
```diff
@@ -0,0 +1,65 @@
...
+from sqlalchemy.orm import Query, aliased
+
+
+class CouchersQuery(Query):
+    # this class is required in models, but models require this
+    # so to get around this, we have these variables here that you have to patch up after the models have been defined
+    user_class = None
+    user_block_class = None
+
+    def blocked_users(self, session, user_id):
```
aapeliv: This should maybe be `_blocked_users`

PR #1200, opened 2021-05-09
Combine edit profile and edit home in one page with tabs

### LINKED ISSUE #949: Combine edit profile and edit home into one page
It was suggested for better UX that we have one edit page with tabs for "profile" and "home"

thdk:
It closes #949 at least partially and is the start of some changes to the profile pages to better match the design from figma.

There now is a single page with two tabs. One tab for 'about' and another one for 'home'. I've added routes so you can directly navigate to a specific tab. Also after saving the form, you'll be redirected to the related view with the updated information.

BREAKING CHANGE: the username is now REQUIRED in profile urls.

Example routes:
user/aapeli
user/aapeli/about (same as user/aapeli)
user/aapeli/home
user/edit/aapeli
user/edit/aapeli/home
user/edit/aapeli/about (same as user/edit/aapeli)

When current route is _user/aapeli/home_ and user clicks the 'edit' button,
then user the route is changed to _user/edit/aapeli/home_.
Next when the user clicks 'save', the route is changed again to user/aapeli/home.

So you stay on the profile section you were editing/viewing.

Let me know if this is not what we want, I have little experience with best practices for routes.
My argument for having the user id in the url is:
- you won't have issues if there is a user that uses 'edit', 'home'  or 'about' as username
- users can copy paste their url from their profile page and send it to other people.

Regarding the Account Settings button. There is an option to add the account settings as another tab on the edit profile page. However, @ItsiW said it actually belongs in the hamburger menu so I left it as is for now. On the top right of the edit profile page.

**Frontend checklist**
 => I can probably still write (or copy) the existing test that we now have for edit hosting preferences for edit profile tab.
- [x] All tests pass


app/frontend/src/routes.ts
```diff
@@ -16,10 +18,20 @@ export const referencesRoute = `${userRoute}/references`;
 export const favoritesRoute = `${userRoute}/favorites`;
 export const photosRoute = `${userRoute}/photos`;
```
lucaslcode: Since we are in this bit, do you want to change these to match what the community routes do?
ie. 
```
const userBaseRoute = "/user"; //(not exported)
export const userRoute = `${userBaseRoute}/:var1/:var2`
export const aboutRoute = `${userBaseRoute}/about`
...
```
That way, it keeps relevant code together and there doesn't have to be hardcoded variables in `AppRoutes.tsx`.

app/frontend/src/routes.ts
```diff
@@ -16,10 +18,20 @@ export const referencesRoute = `${userRoute}/references`;
...
+  {
+    edit,
+    tab,
+  }: {
+    edit?: boolean;
+    tab?: ProfileTabs;
+  } = {}
+) =>
+  // /user/edit/username/about
+  `${userRoute}/${edit ? "edit/" : ""}${username}${tab ? "/" : ""}${tab || ""}`;
```
lucaslcode:
```suggestion
  `${userRoute}/${edit ? "edit/" : `${username}/`}${tab || ""}`; //I think?
```
I don't think username should be there if you are editing - it implies you can put a different username in and edit it

lucaslcode: In any case, this means "edit" is not a valid username. Maybe after merging this, open a bug for backend to disallow it.

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -113,7 +95,34 @@ export default function Navigation() {
   const authenticated = useAuthContext().authState.authenticated;
   const [open, setOpen] = React.useState(false);
 
-  const drawerItems = (
+  const { data: user } = useCurrentUser();
+
+  const menu = useMemo(
```
lucaslcode: I could be wrong, but I don't think this needs to be memoized as the identity of the array is not important.

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -113,7 +95,34 @@ export default function Navigation() {
...
+            {
+              name: "Search",
+              route: searchRoute,
+            },
+            {
+              name: "Profile",
+              route: routeToUser(user.username),
+            },
+          ]
+        : null,
```
lucaslcode: I'm having visions of the navigation buttons disappearing if the user is offline and their cache expires or something 😆 

Maybe a better solution is, if `user` is loading (undefined), to link to "/profile" and have that page just be a minimal widget that displays a loading indicator while `useCurrentUser()` is loading, then redirect to "/profile/username". That would still satisfy your two reasons for requiring username in the url

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -160,319 +158,321 @@ export default function HostingPreferenceForm() {
...
-                    name="sleepingArrangement"
-                    className={classes.field}
-                  />
-                )}
-              />
-            )}
-          />
-          <div className={classes.checkboxContainer}>
-            <div>
+        <>
```
lucaslcode: Is it just me or would this diff be way smaller without this (it's not needed if there's only one child)

app/frontend/src/features/profile/edit/EditProfilePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const classes = useStyles();
+  const { tab = "about" } = useParams<{ tab: ProfileTabs }>();
+  const history = useHistory();
+
+  const { username } = useParams<{
+    username?: string;
+  }>();
+
+  const currentUser = useCurrentUser();
+  const { data: user, isLoading: loading, error } = useUserByUsername(
```
lucaslcode: just use `isLoading` instead of renaming

app/frontend/src/features/profile/edit/EditProfilePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+          >
+            {ACCOUNT_SETTINGS}
+          </Button>
+        </div>
+      </Grid>
+      {error && <Alert severity="error">{error}</Alert>}
+      {loading ? (
+        <CircularProgress />
+      ) : user ? (
+        <ProfileUserProvider user={user}>
```
lucaslcode: I don't think this is necessary here - the profile page uses it to avoid prop drilling of the user

app/frontend/src/features/profile/edit/EditProfilePage.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <div className={classes.root}>
```
lucaslcode: nit: maybe a different name? It's not really root here.

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
```diff
@@ -14,21 +15,27 @@ interface UpdateUserProfileVariables {
...
   const userId = useAuthContext().authState.userId;
   const { mutate: updateUserProfile, status, reset } = useMutation<
     Empty,
     Error,
     UpdateUserProfileVariables
   >(({ profileData }) => service.user.updateProfile(profileData), {
     onError: (error, { setMutationError }) => {
+      console.error({
+        error,
+      });
```
lucaslcode: remove

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -95,11 +95,12 @@ export default function Overview({ user, setIsRequesting }: OverviewProps) {
...
         {user.userId === currentUserId ? (
           <>
-            <Button component={Link} to={editProfileRoute}>
-              {EDIT_PROFILE}
-            </Button>
-            <Button component={Link} to={editHostingPreferenceRoute}>
-              {EDIT_HOME}
+            <Button
+              component={Link}
+              to={() => routeToUser(user.username, { edit: true, tab })}
```
lucaslcode: Why is this a function and not a string?

lucaslcode (changes requested):
This is much better, thanks!

ItsiW (dismissed):
Looks good, waaay cleaner now. But we need username to **not** be required in URLs. 

For your first point, we can prevent specific usernames in the backend. Good point on wanting to share profile, we can satisfy this and the requirement with:

- /user/ redirects to /user/aapeli
- /user/about redirects to /user/aapeli/about
- /user/home redirects to /user/aapeli/home

The reason for this is that we link to profile and editing pages in a number of places so that needs to be generic.

I don't think either point applies to edit now. The only use case for sharing is people showing others how to edit their own profile, so it would be better not to have the username in that URL at all. It's sufficient to just have 

- /user/edit
- /user/edit/about 
- /user/edit/home 
 
Also, and might just be a me thing, I couldn't see the About/Home tabs immediately. Could be a question for UX because I understand we want to be consistent.

darrenvong:
Agreed with what's said here already. In particular, we shouldn't have `user/edit/<username>` as that creates a lot of impossible URLs a normal user won't have access to (since the URL implies they can edit another user when they can't) so I think the best thing is to not have them altogether.

thdk:
Wow, you are all very fast with reviewing! 👍

I've already adjusted the routing issues and will continue the other remarks as well as merging develop branch tonight.

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
thdk: Oops. eslint can prevent this. Maybe I look to add a warning for console statements to the rules?

lucaslcode: I think it's okay, there are legitimate uses. Maybe once we have proper error tracking

app/frontend/src/features/profile/edit/EditProfilePage.tsx
thdk: Ah but it will be once the "account settings" has been moved to the hamburger menu.
Not a big fan of using 'root' though. It requires renaming when moving stuff arround.

aapeliv:
Hi, just a heads up, if you change the routes, please remember to fix up also `//docs/urls.md` and `//app/backend/src/couchers/urls.py`. There's a bunch of onboarding emails going out on Sunday, so I just want to make sure they don't lead to 404s!

lucaslcode:
I'm assuming this is ready for review now - remember to request so we know! :)

app/frontend/src/features/profile/edit/EditProfilePage.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { TabContext, TabPanel } from "@material-ui/lab";
+import PageTitle from "components/PageTitle";
+import TabBar from "components/TabBar";
+import {
+  ACCOUNT_SETTINGS,
+  EDIT_PROFILE,
+  SECTION_LABELS,
+  SECTION_LABELS_A11Y_TEXT,
+} from "features/constants";
+// import References from "features/profile/view/References";
```
lucaslcode: Delete - I don't think references will be necessary on this page

app/frontend/src/features/profile/hooks/useUpdateHostingPreferences.ts
```diff
@@ -33,7 +35,12 @@ export default function useUpdateHostingPreferences() {
       },
       onSuccess: () => {
         queryClient.invalidateQueries(["user", userId]);
-        history.push(userRoute);
+        if (user) {
+          history.push(routeToUser({
+            username: user.username,
+            tab: "home"
+          }));
+        }
```
lucaslcode: The user should exist here, even if the cache is stale. If for some reason it doesn't, it could make a really difficult to track bug ("Clicking save has no effect"). I think maybe better do `user!.username` - at least that way there'll be a console error (unfortunately it wouldn't trigger our error component since this is async).

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
```diff
@@ -32,10 +34,14 @@ export default function useUpdateUserProfile() {
       },
       onSuccess: () => {
         queryClient.invalidateQueries(["user", userId]);
-        history.push(userRoute);
+        if (user) {
+          history.push(routeToUser({
+            username: user.username,
+            tab: "about"
+          }));
+        }
```
lucaslcode: As above

app/frontend/src/routes.ts
```diff
@@ -15,17 +15,39 @@ export const confirmChangeEmailRoute = "/confirm-email";
...
+export type UserTab = "about" | "home" | "references" | "favorites" | "photos";
+export type EditUserTab = Extract<UserTab, "about" | "home">;
+
+export const userRoute = `${userBaseRoute}/:userId?/:tab?`;
+export const editUserRoute = `${userBaseRoute}/edit/:tab?`;
+
+type UserRouteOptions<T extends UserTab> = {
+  username?: string;
+  tab?: T;
+  edit?: T extends EditUserTab ? true : T extends undefined ? true : never;
```
lucaslcode: nice

app/frontend/src/routes.ts
```diff
@@ -15,17 +15,39 @@ export const confirmChangeEmailRoute = "/confirm-email";
...
-export const editHostingPreferenceRoute = `${userRoute}/edit-hosting`;
-
-export const routeToUser = (username: string) => `${userRoute}/${username}`;
+// user
+
+const userBaseRoute = "/user";
+export type UserTab = "about" | "home" | "references" | "favorites" | "photos";
+export type EditUserTab = Extract<UserTab, "about" | "home">;
+
+export const userRoute = `${userBaseRoute}/:userId?/:tab?`;
```
lucaslcode:
```suggestion
export const userRoute = `${userBaseRoute}/:username?/:tab?`;
```
Visiting other user profiles and tests are broken because of this line

lucaslcode (changes requested):
Nice work!

There's a variable accidental rename and just a few other nits

app/frontend/src/routes.ts
darrenvong: Why is `edit` typed as `true` if it's undefined? Doesn't that mean they are not on the edit page and so edit can't be true?

app/frontend/src/routes.ts
```diff
@@ -15,17 +15,39 @@ export const confirmChangeEmailRoute = "/confirm-email";
...
+): string;
+export function routeToUser<T extends UserTab>(
+  options?: Omit<UserRouteOptions<T>, "edit">
+): string;
+export function routeToUser<T extends UserTab>(options?: T): string;
+export function routeToUser<T extends UserTab>({
+  username,
+  tab,
+  edit,
+}: UserRouteOptions<T> = {}) {
```
darrenvong: I think it would be nice if there are some unit tests (one for each function signature would be a good start) for this as there are a fair bit of logic here. Also I find all these clever generics and combination of ternaries a little hard to follow without some concrete examples personally so the tests will serve as a nice secondary thing to make supposed usage of this function clearer

app/frontend/src/routes.ts
```diff
@@ -15,17 +15,39 @@ export const confirmChangeEmailRoute = "/confirm-email";
...
+  // edit: user/edit(/tab)
+  // view: user(/username)(/tab)
+  return `${userBaseRoute}${!edit && username ? "/" : ""}${
+    (!edit && username) || ""
+  }${edit ? "/edit" : ""}${tab ? "/" : ""}${tab || ""}`;
```
```suggestion
  return `${userBaseRoute}${edit ? "/edit" : username ? `/${username}`: ""}${tab ? `/${tab}` : ""}`;
```
I think this would also work? And reduce the need of having to do the same boolean check twice

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -93,10 +92,17 @@ export default function ProfilePage() {
...
-                  value={currentTab}
-                  setValue={setCurrentTab}
+                  value={tab}
+                  setValue={(newTab) =>
+                    history.push(
+                      routeToUser({
+                        username: user.username,
+                        tab: newTab,
+                      })
+                    )
```
darrenvong: > * /user/home redirects to /user/aapeli/home

This use case doesn't seem to be supported I think because of this. Could we support it somehow without needing the redirect, considering `/user/about` works fine to take you to your user's "about" section currently?

thdk:
@aapeliv @darrenvong @lucaslcode thank you all for reviewing.

I'm still busy with this PR. I had some trouble on finding the best routes with optional usernames. 

/User/home (for route user/username?/tab?)

Still don't know how the frontend can now for such route whether 'home' is a user or a tab on the profile page.

Except for hardcoding something like:

const user,tab = useParams()

If user === home || user === about
 tab = user
 user = undefined
end if


This long weekend I was visiting my family in law in Czechia so I pushed my changes already in case I would have some spare time here :)

This week Ill have some more time again. I'll work on your feedback and add tests.

app/frontend/src/routes.ts
thdk: If T is undefined it means caller didn't pass any value (or undefined) for the 'tab' property. So the result should be a url for the default tab for profile page which is 'about'. 

About page does have an 'edit' mode so you are allowed to pass 'true' for the 'edit' prop.

However, you can also omit the 'edit' prop for the 'about' tab. Which will return the view. 

Passing 'false' for edit prop explicitly isn't allowed by any overload as you can omit the edit prop in that case.

darrenvong:
```suggestion
  edit?: T extends EditUserTab | undefined ? true : never;
```
I think this would also work? Makes this marginally easier to read :)

thdk:
> Hi, just a heads up, if you change the routes, please remember to fix up also `//docs/urls.md` and `//app/backend/src/couchers/urls.py`. There's a bunch of onboarding emails going out on Sunday, so I just want to make sure they don't lead to 404s!

Ah didn't know about these, so good that you mentioned it. I've added the new routes to both these files. But not sure if they should support the 'tab' parameter as well. Let me know if changes are required.

app/frontend/src/routes.ts
thdk: Haha, definitly true! I'll change it of course :)

app/frontend/src/features/profile/view/ProfilePage.tsx
thdk: Not sure if I understand what I should do with this. 
I think this is what you want: 

- current page: `/user`
- click on _about_ or _home_
- url changes to: `/user/about` or `/user/home`
- while with my code it will change to `/user/username/about` and `/user/username/home`

Maybe a follow issue can further look into optimising the routes?

darrenvong: Yes, the not needing redirect is the less important part, it's more the lack of the consistency this introduces if we push this out, so I'd prefer that we fix it here rather than pushing this out and leave for later.

darrenvong:
Screenshot 2021-05-17 at 22 44 12
Visiting other profiles is completely broken - the URL says "aapeli" but it's still showing the test user's profile and it seems to the same for all profiles

app/frontend/src/features/profile/view/ProfilePage.tsx
thdk: Ok, no problem. I just don't immediately know how can I prevent 'about' and 'home' to be seen as a username for the urls:
`/user/about` and `/user/home` when path is configured with two optional parameters like `/user/:username?/:tab?`

We can maybe cover this tonight in the dev meeting?

thdk: image

thdk: One approach could be to have two base routes. One for the profile of the current user and one for viewing profiles of other users:

```
/me
/me/(edit/)about
/me/(edit/)home
```

and
```
/user/username
/user/username/about
/user/username/home
```

('me' can be anything else of course such as 'profile')

darrenvong: I see, I missed part of the discussion near the beginning but I think we agreed (?) what you proposed there is probably the best alternative that makes sense and more resilient to change than hardcoding a list of special words that can't be usernames (which I'm almost certain we'll keep on changing as we keep on adding tabs to that list...)

thdk: Ooops.. I thought we would go for the option where we would look on the profilepage (frontend) if the username parameter (the first parameter) is either "home" or "about". If that's the case we use the following parameters:
username: undefined
tab: home (or about)

But now I read your comment, I believe you would actually vote for having two seperate base routes?

app/frontend/src/features/profile/view/ProfilePage.test.tsx
```diff
@@ -95,6 +108,26 @@ describe("Profile page", () => {
...
+
+    describe("and a tab is opened", () => {
+      it("updates the url with the chosen tab value", async () => {
+        renderProfilePage();
+
+        expect(testLocation.pathname).toBe("/user");
+
+        await act(async () => {
+          userEvent.click(await screen.findByText(SECTION_LABELS.home));
+        });
```
darrenvong:
```suggestion
        userEvent.click(await screen.findByText(SECTION_LABELS.home));
```
`userEvent` already wrap things in `act` with all of its methods so we don't need to explicitly do it.

app/frontend/src/features/profile/view/ProfilePage.test.tsx
```diff
@@ -95,6 +108,26 @@ describe("Profile page", () => {
...
+
+        expect(testLocation.pathname).toBe("/user/home");
+
+        await act(async () => {
+          userEvent.click(await screen.findByText(SECTION_LABELS.about));
```
darrenvong: Same comment as above

darrenvong (changes requested):
Only some final minor comments about the tests from me!

> But now I read your comment, I believe you would actually vote for having two seperate base routes?

Since the other file is now outdated it won't let me reply directly, but yes I'd prefer having two routes like you propose tbh. In fact, that was how we started out - users were under `/user/<username>` and the current user was under `/profile` before a previous contributor eagerly merged them because they were too similar and now we're paying the cost of over DRYing the code.

Considering user profiles are mostly going to be read only and the current user'a profile eventually is going to allow you to edit parts of it within the profile cards/tiles if we follow the designs, trying to share it and make it DRY was clearly a bad idea. I should have seen that coming and veto'ed against the change but overlooked it on my part 🙃 

Anyway with the context out of the way, I think this has dragged for a while and since you fixed the URL consistency issue with this workaround (which was my main issue of not wanting it to be merged), I'm happy for this to be merged and leave that issue to be addressed in a separate PR. Ideally the invalid username should be addressed in the backend imo, as if the user is not found, but then I guess that won't be an issue if we go back to using two routes in future 🤷

app/frontend/src/components/TabBar/TabBar.tsx
```diff
@@ -29,20 +31,23 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
   return (
     <TabList
       aria-label={ariaLabel}
+      value={value}
```
darrenvong: Why was this change needed? Is it to allow `undefined` values so you can have no tabs selected?

app/frontend/src/features/profile/edit/EditProfilePage.tsx
darrenvong: This could be deleted like what Lucas said since it's not relevant to this page?

app/frontend/src/features/profile/view/ProfilePage.test.tsx
```diff
@@ -102,6 +154,24 @@ describe("Profile page", () => {
...
+      it("updates the url with the chosen tab value", async () => {
+        expect(testLocation.pathname).toBe("/user/funnydog");
+
+        await act(async () => {
+          userEvent.click(await screen.findByText(SECTION_LABELS.home));
+        });
+
+        expect(testLocation.pathname).toBe("/user/funnydog/home");
+
+        await act(async () => {
```
darrenvong: And another on `act`

app/frontend/src/features/userQueries/useUserByUsername.ts
```diff
@@ -11,15 +11,21 @@ import { service } from "service";
...
 ) {
   //We look up the userId first from the username.
   //This causes a duplicate query, but it is not made stale for a long time
   //and ensures no duplication of users in the queryCache.
-  const usernameQuery = useQuery<{ username: string; userId: number }, Error>({
+  const usernameQuery = useQuery<
+    { username: string; userId: number } | null,
+    Error
+  >({
     cacheTime: username2IdStaleTime,
```
darrenvong: Still feels a bit weird to allow `undefined` in the query key as they are technically pointing to "ghost entities" in the cache which feels a bit of a code smell... but I'm guessing this is the problem we've created by badly overloading our URL routes.

Maybe if `username` is undefined, we should have a super low `cacheTime` so at least React Query clears it from the cache so we don't have invalid data sitting around with something like this?
```suggestion
    cacheTime: username ? username2IdStaleTime : 0,
```

app/frontend/src/components/TabBar/TabBar.tsx
thdk: Ah yes, I had seen your commit where you simplified the tabs component. However, I was unable to open a specific tab after that change.

(Example value is 'home' when url is user/home)

So I added the value prop again. Or is there another way to render the tabbar component with a specific tab opened instead of the first one (default)?

app/frontend/src/features/profile/edit/EditProfilePage.tsx
thdk: Strange, I remember me already deleting this line. Maybe I forgot to commit it.

app/frontend/src/features/userQueries/useUserByUsername.ts
thdk: Yes, this can be avoided with seperating the routes. Rules of hooks don't allow to call hooks conditionally so I had to allow undefined here which does look weird indeed.

app/frontend/src/features/profile/view/ProfilePage.tsx
thdk: Will look at separating routes in #1272

app/frontend/src/components/TabBar/TabBar.tsx
darrenvong: That should be controllable through the value prop passed to `TabContext`. I just tried it on `develop` by changing the initial state passed to it in the current code without this change and it works, so you should be able to do something similar by deriving the initial state based on the route?

darrenvong: That should be controllable through the value prop passed to `TabContext`. I just tried it on `develop` by changing the initial state passed to it in the current code without this change and it works, so you should be able to do something similar by deriving the initial state based on the route?

app/frontend/src/components/TabBar/TabBar.tsx
```diff
@@ -29,20 +31,23 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
...
-      {Object.entries(labels).map(([value, label]) => (
-        <Tab
-          key={value}
-          label={label}
-          value={value}
-          className={classes.messagesTab}
-        />
-      ))}
+      {Object.entries(labels)
+        .filter(([isDefined]) => !!isDefined)
```
darrenvong:
```suggestion
```
I'm not sure there's a good case for supporting undefined labels, since we can always add to the tab label later when we actually have content for it.

darrenvong (changes requested):

aapeliv:
@thdk: how're you doing on this one?

thdk:
I have no computer with me this weekend as I'm on a family trip.

I'll finish it beginning of this week.

Op zo 23 mei 2021 22:13 schreef Aapeli Vuorinen ***@***.***>:

app/frontend/src/components/TabBar/TabBar.tsx
thdk: Yes, that works indeed! :)

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
```diff
@@ -34,7 +32,7 @@ export default function useUpdateUserProfile() {
         setMutationError(null);
       },
       onSuccess: () => {
-        queryClient.invalidateQueries(["user", userId]);
```
darrenvong, line 35: This invalidation needs to also be kept in addition to the `accountInfoQueryKey`, otherwise user data is gonna be outdated after they updated their profile and introduces an unintended bug...

app/frontend/src/components/TabBar/TabBar.tsx
```diff
@@ -23,11 +23,13 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
   labels,
 }: TabBarProps<T>) {
   const classes = useStyles();
+  const tabContext = useTabContext();
   const handleChange = (event: any, newValue: keyof T) => {
     setValue(newValue);
   };
   return (
     <TabList
+      value={tabContext?.value}
```
darrenvong:
```suggestion
```
The custom tab select still works for me when I tried this branch locally so I don't think this is necessary.

darrenvong (changes requested):
The `TabBar` thing is minor really and I was happy to leave it if there wasn't anything else... but since your merge here is introducing a bug by replacing the user invalidation on save (by accident I hope!), thought I'll point it out as you might as well fix both together

app/frontend/src/components/TabBar/TabBar.tsx
thdk: Ah, thanks for pointing that out! Now I understand how this component works :)

app/frontend/src/features/profile/hooks/useUpdateUserProfile.ts
thdk: My fault, I thought this new line overrides the previous line. I should have verified that before committing this merge.

app/frontend/src/features/constants.ts
```diff
@@ -36,6 +37,9 @@ export const WHY = "Why I use Couchers";
 export const SECTION_LABELS = {
   about: ABOUT_ME,
   home: HOME,
+  references: undefined, // REFERENCES,
+  photos: undefined, // PHOTOS,
+  favorites: undefined, // FAVORITES,
```
darrenvong:
```suggestion
```
I think these are leaving behind "ghost tabs" with empty labels and causing errors (since there aren't users called "references", "photos" etc) if the user accidentally clicked on what looks like whitespace next to the "home" tab.

darrenvong (changes requested):
Sorry I didn't review this as quickly, there's also a merge conflict with develop now but hopefully this should be it once fixed 🤞

app/frontend/src/features/constants.ts
thdk: Ah that's possible, that was why I filtered out undefined values in the tabbar. I checked the tabbar after and didn't see a problem, but it makes sense the tabs are still there but invisible. I'll remove them.

darrenvong (dismissed):

lucaslcode (commented):
There shouldn't have been prettier issues - I just updated prettier. Try running `yarn && yarn format`

thdk:
> There shouldn't have been prettier issues - I just updated prettier. Try running `yarn && yarn format`

Auch, now I have even more changes.. I'll commit them so you can see if that's normal.

lucaslcode (approved):

PR #1218, opened 2021-05-10
Events part 2

aapeliv:
WIP. Missing tests. From #744:

* Proper timezone handling
* Notifications
* ~~Corresponding `ListEvents` in communities, groups~~
* ~~Search functionality~~
* Automatically repeating events (currently need to schedule each instance manually) — not so urgent

Other:

* ~~Integrate with user filtering~~

### DIFF
--- a/app/backend/src/couchers/servicers/communities.py
+++ b/app/backend/src/couchers/servicers/communities.py
@@ -6,8 +6,20 @@ from sqlalchemy.sql import func, or_
 
 from couchers import errors
 from couchers.db import can_moderate_node, get_node_parents_recursively, session_scope
-from couchers.models import Cluster, ClusterRole, ClusterSubscription, Discussion, Node, Page, PageType, User
+from couchers.models import (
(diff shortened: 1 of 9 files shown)


lennartvanlaake:
I was just going through some of your PRs to get a sense of the code style you are using.

I did notice that there are multiple implementations of ListEvents in different files that seem to have the same business logic. Do you have a reason for doing so and not encapsulating that logic on a separate component/function so it can be reused?

aapeliv:
> I did notice that there are multiple implementations of ListEvents in different files that seem to have the same business logic. Do you have a reason for doing so and not encapsulating that logic on a separate component/function so it can be reused?

I'm not sure what the reasoning was back when this grouping was initially proposed: maybe the thought was that e.g. communities have some pretty involved logic about what content is shown there. But I tend to agree, we should probably move all `ListEvents` into `events` and combine into one.

kalvdans (approved):
LGTM!

PR #1219, opened 2021-05-10
Timezone areas

aapeliv:
@kalvdans: what do you think about this?

We need to figure out some way of managing resources that need to be synced from files into the database. It's important for the timezone stuff: the timezone polygons need to be stored in the database but loaded from a source of truth file.

In this PR, resource tables are updated in migrations: the table data is fully replaced, and if there is any inconsistencies (e.g. referring to a language that is deleted), then the migration will fail like any other migration failure.


app/backend/src/couchers/resources.py
```diff
@@ -38,12 +41,56 @@ def get_language_dict():
...
+        regions = [(region["alpha3"], region["name"]) for region in json.load(f)]
+
+    with open(resources_folder / "languages.json", "r") as f:
+        languages = [(language["code"], language["name"]) for language in json.load(f)]
+
+    timezone_areas_file = resources_folder / "timezone_areas.sql"
+
+    if not timezone_areas_file.exists():
+        timezone_areas_file = resources_folder / "timezone_areas.sql-fake"
+        logger.info(f"Using fake timezone areas")
```
aapeliv: How should this be implemented?

Through an env var? Like this but raise an exception if in prod?

kalvdans:
We discussed on a meeting https://github.com/Couchers-org/couchers/blob/develop/docs/meetings/backend/Meeting%20020.md but reached no consensus... Checking in the timezone areas in the git repo is probably the easiest, even if it is big. Then we can also bundle updates to the timezone area dataset with an appropriate migration file.

If test times gets too long, we could flag tests that can run without timezone areas, and skip the loading for those.

aapeliv:
> We discussed on a meeting

This is roughly what I was suggesting in that meeting. The file I have at <https://bigstuff.a19e.net/couchers/timezone_areas.sql.zst> is just exported from <https://github.com/evansiroky/timezone-boundary-builder> and re-exported using QGIS to get it into an easy format for inserting into pg.

> Checking in the timezone areas in the git repo is probably the easiest, even if it is big.

The file is 200 MB uncompressed, ~80 with gzip, ~35 with zstd (max compression). I was thinking of putting it in a sibling repository, then we can grab the file by git hash.

> If test times gets too long, we could flag tests that can run without timezone areas, and skip the loading for those.

Adding the full thing dataset the DB at the start of the test is way too slow. I added a `-fake` version where I drew a couple of manual polygons that approximate the real timezones (and UTC for everything else). That seems to be fast and pretty good for testing and for local development.

> Then we can also bundle updates to the timezone area dataset with an appropriate migration file.

Well that's kind of the idea here: whenever we would change the dataset, we'd add that little snippet to the end of the migration to replace the constant table with the version from a certain git hash and it would work like usual.

kalvdans:
>  I was thinking of putting it in a sibling repository, then we can grab the file by git hash.

I like that idea, then we get version control too. Would you like to use `git submodule` for this? I'm pretty fluent in that having worked with it quite recently.

> Adding the full thing dataset the DB at the start of the test is way too slow.

Ok, maybe only the migration test need to load the whole dataset.

aapeliv:
> I like that idea, then we get version control too. Would you like to use `git submodule` for this? I'm pretty fluent in that having worked with it quite recently.

We'll need to come up with something that doesn't add complexity to the initial setup with new devs. I think it probably suffices to just link to the asset through the git hash like this:

<https://raw.githubusercontent.com/Couchers-org/grpc/b48b45ca3ea9869d0efe85bdb36b4bf9a0c8c060/Dockerfile>

But I'm also happy to add a submodule that can be optionally cloned to get the full dataset. Though if we have multiple versions, the git repo will get huge very quickly, whereas this only allows grabbing one file.

I'm happy with anything as long as it keeps it easy to get started. Development doesn't require the full timezone area definitions.

kalvdans:
> > I like that idea, then we get version control too. Would you like to use `git submodule` for this? I'm pretty fluent in that having worked with it quite recently.
> 
> We'll need to come up with something that doesn't add complexity to the initial setup with new devs. I think it probably suffices to just link to the asset through the git hash like this:
> 
> https://raw.githubusercontent.com/Couchers-org/grpc/b48b45ca3ea9869d0efe85bdb36b4bf9a0c8c060/Dockerfile

Sounds good, we can always switch later.

> But I'm also happy to add a submodule that can be optionally cloned to get the full dataset. Though if we have multiple versions, the git repo will get huge very quickly, whereas this only allows grabbing one file.

True, if we commit compressed data git can't do any delta-coding.

> I'm happy with anything as long as it keeps it easy to get started. Development doesn't require the full timezone area definitions.

It does in order to run the migration test.

aapeliv:
> It does in order to run the migration test.

We should modify that test to also check for database contents then? Currently it just checks the schema itself, and that matches regardless of the data inserted.

kalvdans:
> > It does in order to run the migration test.
> 
> We should modify that test to also check for database contents then? Currently it just checks the schema itself, and that matches regardless of the data inserted.

I didn't think about that. You are right, some changes are needed to also check timezone data migrations are correct.

app/backend/src/couchers/resources.py
kalvdans: Hard error if timezone_areas.sql doesn't exists, with a pointer to the README that describes how to download it.

aapeliv: How about a hard error if not running with `DEV=1`?

kalvdans: Sure!

aapeliv:
> I didn't think about that. You are right, some changes are needed to also check timezone data migrations are correct.

Do you want to work on that in another PR? 🙃

kalvdans:
> > I didn't think about that. You are right, some changes are needed to also check timezone data migrations are correct.
> 
> Do you want to work on that in another PR?

No :) But I'm fine leaving it to later.

aapeliv:
I created https://github.com/Couchers-org/resources with a script to generate the areas and compressed output (github wouldn't let me upload the uncompressed one as it's over their file limit of 100 MB).

aapeliv:
Could you re-review now @kalvdans. The frontend tests are failing but backend should be fine.

app/backend/src/couchers/resources.py
```diff
@@ -38,12 +42,59 @@ def get_language_dict():
...
+        if not config["DEV"]:
+            raise Exception("Missing timezone_areas.sql and not running in dev")
+
+        timezone_areas_file = resources_folder / "timezone_areas.sql-fake"
+        logger.info(f"Using fake timezone areas")
+
+    with open(timezone_areas_file, "r") as f:
+        tz_sql = f.read()
+
+    session.execute("SET CONSTRAINTS ALL DEFERRED")
```
kalvdans, line 89: Hmm, shouldn't we update all constraints in models.py to be deferred as well?

aapeliv: I'm not quite sure what you mean.

This only sets the constraints that have been marked as deferrable as deferred for this transaction only.

If the constraints aren't deferrable, then this will fail as we wipe the tables. What we're doing now allows us to wipe and recreate because the constraints are only evaluated after the fact. So we still catch it if we removed some data but didn't migrate correctly (before this fragment), but it'll fail at the `COMMIT` of the tx

kalvdans: Aha, I understand! Thanks for explaining. LGTM!

kalvdans (approved):

app/backend/src/couchers/resources.py
aapeliv: I added a comment to clarify, thanks

PR #1222, opened 2021-05-11
API tracing

aapeliv:
Add request logging, both for debug diagnostics as well as better analytics on waht features people are using, etc.

Eventually we surely want a better way to store these traces, but for now, I think this is fine.

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/interceptors.py
```diff
@@ -89,56 +94,64 @@ def intercept_service(self, continuation, handler_call_details):
         return continuation(handler_call_details)
 
 
-class LoggingInterceptor(grpc.ServerInterceptor):
```
kalvdans, line 92: Keep them as two separate chainable interceptors, instead of one that does two things.

app/backend/src/couchers/interceptors.py
```diff
@@ -89,56 +94,64 @@ def intercept_service(self, continuation, handler_call_details):
...
+                    method=method,
+                    status_code=status_code,
+                    duration=duration,
+                    user_id=user_id,
+                    request=req_bytes,
+                    response=res_bytes,
+                    traceback=traceback,
+                )
+            )
+        logger.debug(f"{user_id=}, {method=}, {duration=} ms, req={len(req_bytes or b'')}, res={len(res_bytes or b'')}")
```
kalvdans, line 129: The length of the sanitized request and response doesn't add any statistics or debugging aid.

app/pb/annotations.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package org.couchers;
+
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FieldOptions {
+  // whether the field is sensitive and should not be logged
+  bool sensitive = 50000;
```
kalvdans, line 7: What does the number indicate here? (Too lazy to look up docs)

app/backend/src/couchers/interceptors.py
```diff
@@ -89,56 +94,64 @@ def intercept_service(self, continuation, handler_call_details):
...
-    """
+    def _sanitized_bytes(self, proto):
+        """
+        Remove fields marked sensitive and return serialized bytes
+        """
+        if not proto:
+            return None
+        new_proto = deepcopy(proto)
+        for name, descriptor in new_proto.DESCRIPTOR.fields_by_name.items():
+            if descriptor.GetOptions().Extensions[annotations_pb2.sensitive]:
```
kalvdans, line 110: An easier alternative would be `if "token" in name:`

kalvdans (commented):
Do you have any tool to make use of the saved data, or is it just "good to have" for a future tool?

Somehow it feels like traffic logging should go into a separate database.

app/backend/src/couchers/interceptors.py
aapeliv: Ah, instead of the annotations? What about passwords, etc? Better to annotate the protos, I think

app/pb/annotations.proto
aapeliv: It's the field number like elsewhere in protos. The "internal extension" fields should be between 50000-59999 according to docs.

(There's some common/public annotations e.g. the http transcoding stuff that needs to be unique, so there's some registry of those as well)

app/backend/src/couchers/interceptors.py
aapeliv: I'll remove them

app/backend/src/couchers/interceptors.py
aapeliv: OK

app/pb/annotations.proto
kalvdans: But it is never seen on the wire?

aapeliv: We never see it on the wire. But protobuf eats protobuf: all the code generation API surfaces (it's all plugin based) pass around these descriptor protos, etc.

kalvdans: And it doesn't use strings for that? silly. Anyway, not our fault :) resolving.

aapeliv:
> Do you have any tool to make use of the saved data, or is it just "good to have" for a future tool?

Yes and no. I do connect to the db and do some simple queries quite frequently. It'd be good to compute simple error rates and response times. We need to keep an eye out on the response times, we're slowly starting to grow in user numbers and need to make sure we've the right indexes, etc; so having the timings per method and a way to diagnose slowness will be great. I've also helped Itsi connect to the database through a jupyter notebook environment with a restricted read-only user. We'll write some hacky scripts and go from there.

> Somehow it feels like traffic logging should go into a separate database.

Do you mean in just a separate instance or separate software (something more suited for this kind of task)? I'd tend to agree, it probably wouldn't be too difficult to add another container running alongside everything to store the traces. I'm just trying to keep things as simple as possible until they need to be made more complicated.

In particular I want to add absolutely no extra friction to getting a dev setup running, but for that we could just turn logging on/off with an environment variable.

app/backend/src/couchers/interceptors.py
kalvdans: yeah, it's easy to miss when renaming fields otherwise. resolving.

kalvdans:
> > Do you have any tool to make use of the saved data, or is it just "good to have" for a future tool?
> 
> Yes and no. [...]

Great thanks for giving the background! Historical protobuf messages would be impossible to parse though since we keep updating the format, so I suggest skipping them. rpc function name should be enough.

> > Somehow it feels like traffic logging should go into a separate database.
> 
> Do you mean in just a separate instance or separate software (something more suited for this kind of task)?

I meant a separate postgresql database.

> In particular I want to add absolutely no extra friction to getting a dev setup running, but for that we could just turn logging on/off with an environment variable.

At work I've written a log parser to extract telemetry data. It keeps the actual production code very simple and pushes complexity (for example parsing old log formats) to the log extractor. A simple awk script can summarize execution times for different rpc calls.

app/backend/src/couchers/interceptors.py
kalvdans: keep LoggingInterceptor separate from TracingInterceptor

app/pb/account.proto
```diff
@@ -50,14 +52,14 @@ message GetAccountInfoRes {
 
 message ChangePasswordReq {
   // a null value indicates that the password wasn't set before
-  google.protobuf.StringValue old_password = 1;
+  google.protobuf.StringValue old_password = 1 [(sensitive) = true];;
```
kalvdans:
```suggestion
  google.protobuf.StringValue old_password = 1 [(sensitive) = true];
```

app/pb/account.proto
```diff
@@ -50,14 +52,14 @@ message GetAccountInfoRes {
...
   // a null value indicates that the password should be unset (removed)
   // the frontend should ask for the password twice and whatnot
-  google.protobuf.StringValue new_password = 2;
+  google.protobuf.StringValue new_password = 2 [(sensitive) = true];;
```
kalvdans:
```suggestion
  google.protobuf.StringValue new_password = 2 [(sensitive) = true];
```

app/pb/account.proto
```diff
@@ -50,14 +52,14 @@ message GetAccountInfoRes {
...
 }
 
 message ChangeEmailReq {
-  google.protobuf.StringValue password = 1;
+  google.protobuf.StringValue password = 1 [(sensitive) = true];;
```
kalvdans:
```suggestion
  google.protobuf.StringValue password = 1 [(sensitive) = true];
```

app/pb/annotations.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package org.couchers;
```
kalvdans:
```suggestion
```

app/backend/src/couchers/interceptors.py
kalvdans: As discussed on the software meeting, let's keep them clumped up together since they share lot of data. We can always restructure later.

aapeliv:
Added some tests and a `version` column so we can figure out which protos were used.

kalvdans (approved):
Looks good to me now! Thanks for adding tests!

PR #1223, opened 2021-05-11
Reverse pagination order of threads

aapeliv:


darrenvong (approved):
Looks good to me based on the tests, unless you want to wait for @kalvdans' stamp instead :P

aapeliv:
This is the wrong thing... See #1224

PR #1224, opened 2021-05-11
threads: newest page first, oldest msg in page first

lucaslcode:
@darrenvong 

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/tests/test_threads.py
+++ b/app/backend/src/tests/test_threads.py
@@ -57,30 +57,30 @@ def test_threads_basic(db):
         ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=PARENT_THREAD_ID))
         assert len(ret.replies) == 3
         assert ret.next_page_token == ""
-        assert ret.replies[0].thread_id == dog_id
-        assert ret.replies[0].content == "dog"
+        assert ret.replies[0].thread_id == bat_id
+        assert ret.replies[0].content == "bat"
         assert ret.replies[0].author_user_id == user1.id
-        assert ret.replies[0].num_replies == 3
+        assert ret.replies[0].num_replies == 0
 
         assert ret.replies[1].thread_id == cat_id
         assert ret.replies[1].content == "cat"
         assert ret.replies[1].author_user_id == user1.id
         assert ret.replies[1].num_replies == 3
 
-        assert ret.replies[2].thread_id == bat_id
-        assert ret.replies[2].content == "bat"
+        assert ret.replies[2].thread_id == dog_id
+        assert ret.replies[2].content == "dog"
         assert ret.replies[2].author_user_id == user1.id
-        assert ret.replies[2].num_replies == 0
+        assert ret.replies[2].num_replies == 3
 
         ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=cat_id))
         assert len(ret.replies) == 3
         assert ret.next_page_token == ""
-        assert [reply.thread_id for reply in ret.replies] == cats[::-1]
+        assert [reply.thread_id for reply in ret.replies] == cats[::1]
 
         ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=dog_id))
         assert len(ret.replies) == 3
         assert ret.next_page_token == ""
-        assert [reply.thread_id for reply in ret.replies] == dogs[::-1]
+        assert [reply.thread_id for reply in ret.replies] == dogs[::1]
 
 
 def test_threads_errors(db):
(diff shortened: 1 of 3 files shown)


app/backend/src/tests/test_threads.py
```diff
@@ -120,7 +120,7 @@ def pagination_test(api, parent_id):
     token = ""
     import textwrap
 
-    for expected_page in textwrap.wrap(string.ascii_lowercase, 5):
+    for expected_page in [x[::-1] for x in textwrap.wrap(string.ascii_lowercase, 5)]:
```
aapeliv, line 123:
```suggestion
    for expected_page in textwrap.wrap(reversed(string.ascii_lowercase), 5):
```

app/backend/src/tests/test_threads.py
```diff
@@ -57,30 +57,30 @@ def test_threads_basic(db):
...
+        assert ret.replies[2].content == "dog"
         assert ret.replies[2].author_user_id == user1.id
-        assert ret.replies[2].num_replies == 0
+        assert ret.replies[2].num_replies == 3
 
         ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=cat_id))
         assert len(ret.replies) == 3
         assert ret.next_page_token == ""
-        assert [reply.thread_id for reply in ret.replies] == cats[::-1]
+        assert [reply.thread_id for reply in ret.replies] == cats[::1]
```
aapeliv: This is the same as without `[::1]` at the end

```suggestion
        assert [reply.thread_id for reply in ret.replies] == cats
```

app/backend/src/tests/test_threads.py
```diff
@@ -57,30 +57,30 @@ def test_threads_basic(db):
...
 
         ret = api.GetThread(threads_pb2.GetThreadReq(thread_id=dog_id))
         assert len(ret.replies) == 3
         assert ret.next_page_token == ""
-        assert [reply.thread_id for reply in ret.replies] == dogs[::-1]
+        assert [reply.thread_id for reply in ret.replies] == dogs[::1]
```
aapeliv: Ditto
```suggestion
        assert [reply.thread_id for reply in ret.replies] == dogs
```

aapeliv (changes requested):
My PR was probably wrong

app/backend/src/tests/test_threads.py
lucaslcode: This would give "zyxwv", "ustrq"... but we want "edcba", "jihgf" ...

lucaslcode:
Sorry I didn't know you were doing this too

app/backend/src/tests/test_threads.py
aapeliv: Ah yes, you're right, sorry.

aapeliv:
> Sorry I didn't know you were doing this too

Nah that's fine, I was doing the wrong thing lol

aapeliv:
Could you add a comment in the proto file explaining the ordering as well?

aapeliv (approved):

app/pb/threads.proto
```diff
@@ -25,7 +25,9 @@ service Threads {
 
   // The same API is used regardless of level. All posts are using the same id namespace.
 
-  // Get non-deleted posts belonging to a thread, sorted in reverse time order.
+  // Get non-deleted posts belonging to a thread.
+  // For ease of display, the pagination order is most recent first, but within a page, replies are oldest first.
+  // For example, if A is the most recent reply, the first page would have `[C, B, A]` and the second page `[F, E, D]`.
```
darrenvong, line 30: Crap, so with the paginated query on the frontend, that means we're gonna get [C, B, A, F, E, D] when we do combine the lists before we render it on the page. But that's not gonna be correct right as we'd want [F, E, D, C, B, A]?

To get the correct thing, we'd have to go through those pages backwards but since you can't write for-loops in JSX, that means the frontend would still need to reverse the pages first before this will be right, defeating the point of what this PR is trying to achieve in the first place... 🤔

lucaslcode: True. Maybe reversing in the frontend is not so bad?
I don't think we're likely to have over 1000 replies on screen. 

Maybe we can do some kind of custom iterator later if there are performance issues?

darrenvong: Yeah thinking about this more, I think doing things on the backend will only work if it's returning data consistently in the same direction as we go through pages, but I don't think that's possible without breaking the pagination logic in this case. The only way I can see this work is you'd have to sort by ascending order and somehow know to return the last X items, which would do a full table scan and would kill performance faster than the frontend lol so that's not really an option. (unless there is something I overlooked?)

So yeah maybe it is easier to just leave it and do it on the frontend. I reckon we will run into more performance issues with 1000s of replies on the screen itself (i.e. in the DOM) rather than a couple of linear time list operations tbh, but that's a different problem to solve for us (and a good problem to have in the famous words of Aapeli 😂 )

Issue #1234, opened 2021-05-12
Develop is broken after flask-2.0.0 was released


aapeliv:
I think Flask 2.0 came out and that broke it...

Test failure from https://gitlab.com/couchers/couchers/-/jobs/1258479036:
```
src/tests/test_server.py .................F                              [100%]
=================================== FAILURES ===================================
______________________________ test_cache_headers ______________________________
client_with_secrets = (<FlaskClient <Flask 'media.server'>>, b"\x01Vh\xbdj\xe2~Z\x01O=\xdb\x9c\x03\x94\xefV0\x99'\xeen\x1frL%p\xeb\xf0\xf3\xd5\x90", '0683105f9afe27896facd49b71d96c2b945581f9609a150664e811f87200dbb3')
    def test_cache_headers(client_with_secrets):
        client, secret_key, bearer_token = client_with_secrets
    
        key, request = create_upload_request()
        upload_path = generate_upload_path(request, secret_key)
    
        f = io.BytesIO(one_pixel_bytes())
    
        with mock_main_server(bearer_token, lambda x: True):
            rv = client.post(upload_path, data={"file": (f, "f")})
        assert rv.status_code == 200
        jd = json.loads(rv.data)
    
        assert jd["ok"]
        assert jd["key"] == key
        assert jd["filename"] == f"{key}.jpg"
        assert jd["full_url"] == f"https://testing.couchers.invalid/img/full/{key}.jpg"
        assert jd["thumbnail_url"] == f"https://testing.couchers.invalid/img/thumbnail/{key}.jpg"
    
        rv = client.get(f"/img/full/{key}.jpg")
        assert rv.status_code == 200
>       assert "max-age=43200" in rv.headers["Cache-Control"].split(", ")
E       AssertionError: assert 'max-age=43200' in ['no-cache']
E        +  where ['no-cache'] = <built-in method split of str object at 0x7feb0a6035f0>(', ')
E        +    where <built-in method split of str object at 0x7feb0a6035f0> = 'no-cache'.split
src/tests/test_server.py:536: AssertionError
------------------------------ Captured log call -------------------------------
WARNING  media.server:server.py:44 Notifying main server about new upload at localhost:8088
WARNING  media.server:server.py:49 Connecting to main server insecurely!
```

kalvdans:
How about pinning our dependencies?

aapeliv:
That would work too. The downside would be that we'd need someone to regularly check them so we get any security patches, etc.

[...]

PR #1235, opened 2021-05-12
Add cache headers back in after flask 2

aapeliv:
Fixes #1234

### DIFF
--- a/app/media/requirements.txt
+++ b/app/media/requirements.txt
@@ -1,5 +1,5 @@
 backoff
-Flask<2
+Flask>=2,<3
 grpcio
 Pillow
 protobuf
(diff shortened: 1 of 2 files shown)


kalvdans:
It would be cleaner if we merge https://github.com/Couchers-org/couchers/pull/1236 first and then we can explicitly bump Flask version in this PR.

aapeliv:
So what should I change the requirements line to here?

kalvdans:
Flask>=2,<3 seems to work :heart:

kalvdans:
nodejs have notations like ^2.0.0 but it seems pip doesn't.

kalvdans (approved):
If CI is happy, I am!

aapeliv:
CI sometimes spawns two pipelines, I killed the one with non-media jobs, but the media stuff went through

https://gitlab.com/couchers/couchers/-/pipelines/302012371

PR #1236, opened 2021-05-12
Our code is only compatible with Flask 1.x so specify that.

kalvdans:
**Backend checklist**
- [x] All tests pass

Fixes: https://github.com/Couchers-org/couchers/issues/1234


aapeliv:
I did fix it in #1235

aapeliv (approved):

PR #1237, opened 2021-05-13
Increase max_workers for ThreadPoolExecutors

aapeliv:
Currently the frontend does a lot of parallel calls to the backend, so this increases the number of workers to increase parallelism to speed it up a bit.


kalvdans (approved):
Looks good to me, I didn't know our frontend was that sophisticated!

PR #1248, opened 2021-05-15
Remove coordinates in guide data that was breaking the dummy_data script

darrenvong:
As in title. Before this change I was getting this error:

```
(psycopg2.errors.CheckViolation) new row for relation "page_versions" violates check constraint "ck_page_versions_geom_iff_address"
backend_1   | DETAIL:  Failing row contains (72, 172, 2, A Guide to Sightseeing in Sydney, # A Guide to Sightseeing in Sydney
```

**Backend checklist**
- [ ] Formatted my code by running `isort . && black .` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [x] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

PR #1249, opened 2021-05-16
Add tests to double-check that our dummy data could be loaded

### LINKED ISSUE #931: Test dummy data
Dummy data is used to populate a new database in dev mode. Currently it goes untested. Would be good to check that all data properly inserts to database.

kalvdans:
closes: https://github.com/Couchers-org/couchers/issues/931
Test loading our dummy data so that issues like https://github.com/Couchers-org/couchers/pull/1248 won't happen again.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/dummy_data.py
+++ b/app/backend/src/dummy_data.py
@@ -33,7 +33,6 @@ from couchers.models import (
 from couchers.servicers.api import hostingstatus2sql
 from couchers.utils import create_coordinate, create_polygon_lng_lat, geojson_to_geom, now, to_multi
 from pb.api_pb2 import HostingStatus
-from tests.test_fixtures import db
 
 logger = logging.getLogger(__name__)
 
@@ -353,8 +352,3 @@ def add_dummy_communities():
 def add_dummy_data():
     add_dummy_users()
     add_dummy_communities()
-
-
-def test_add_dummy_data(db, caplog):
-    add_dummy_data()
-    assert len(caplog.records) == 0
--- a/app/backend/src/dummy_data.py
+++ b/app/backend/src/dummy_data.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 from datetime import date
 
 from dateutil import parser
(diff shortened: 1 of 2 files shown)


app/backend/src/dummy_data.py
```diff
@@ -348,3 +353,8 @@ def add_dummy_communities():
 def add_dummy_data():
     add_dummy_users()
     add_dummy_communities()
+
+
+def test_add_dummy_data(db, caplog):
+    add_dummy_data()
+    assert len(caplog.records) == 0
```
aapeliv: Why not put this in `tests/`?

aapeliv (commented):
Thanks for doing this!

There was also an issue explicitly for this #931

app/backend/src/dummy_data.py
kalvdans: Good question! I somehow thought that `tests/` should only test the `couchers/` code.

(But one could equally well say that `dummy_data.py` should run without the `tests/` code.)

What do you think?

aapeliv: I was just thinking of `tests/` as where the tests go :)

Or maybe just move it into `test_dummy_data.py` next to `dummy_data.py`? I just feel mixing testing code with the main code is a bit icky?

aapeliv:
Also, it seems this test isn't being picked up by CI?

kalvdans:
> Also, it seems this test isn't being picked up by CI?

True! I wonder why that is. We run pytest on the `src` folder, so it should search it recursively according to documentation.

Found it! It is the [python_files = *.py](https://docs.pytest.org/en/latest/reference/reference.html?highlight=python_files#confval-python_files) setting we have at work that fooled me to that pytest by default searches everything.

app/backend/src/dummy_data.py
kalvdans: Ok, I both made an own file and moved it to tests :) Let's see if CI picks it up.

aapeliv (approved):

PR #1250, opened 2021-05-16
First steps towards SMS verification of profiles

### LINKED ISSUE #896: Phone verification
Using Twilio or AWS SNS (this is probably preferred due to us already using AWS heavily)

kalvdans:
Closes  #896

**Backend checklist**
- [x] All tests pass

- [x] check uniqueness of verified phone numbers
- [x] random Luhn token generation
- [x] hide unverified phone numbers from API


app/backend/src/couchers/errors.py
```diff
@@ -61,6 +61,7 @@
 INVALID_HOST_REQUEST_STATUS = "You can't set the host request status to that."
 INVALID_MESSAGE = "Invalid message."
 INVALID_NAME = "Name not supported."
+INVALID_PHONE = "Phone number not in E.164 format."
```
lucaslcode: These errors are displayed to the user - they won't know what E.164 is (neither do I)

app/pb/account.proto
```diff
@@ -77,3 +98,49 @@ message GetContributorFormInfoRes {
 message MarkContributorFormFilledReq {
   bool filled_contributor_form = 1;
 }
+
+message ChangePhoneReq {
+  // New phone number in E.164 format, or empty string to remove phone number.
+  // The phone number will be visible to moderators in your area, verified or not.
```
lucaslcode: Do you mean site admins? Moderators sounds like community moderators, which would be a bad idea

app/pb/account.proto
```diff
@@ -77,3 +98,49 @@ message GetContributorFormInfoRes {
...
+  // If you have a previously verified number, changing your number and losing
+  // the verified status will fail if false is passed.
+  bool remove_verification = 2;
+}
+
+message SendVerificationSMSReq {
+}
+
+message SendVerificationSMSRes {
+  enum SendVerificationSMSStatus {
```
lucaslcode:
```suggestion
  enum SMSResultCode {
```

app/pb/api.proto
```diff
@@ -189,7 +189,7 @@ message User {
   double lng = 34;
   double radius = 35; // meters
 
-  double verification = 5;
+  double verification = 5;  // 1.0 if phone number verified, 0.0 otherwise
```
lucaslcode, line 192: I think this value is an algorithmic value between 0 and 1, of which phone number verification is just a part?

app/pb/account.proto
kalvdans: Reading https://community.couchers.org/t/verification-process/480/18 it sounds like the phone number should be shown to the accepted hosts and accepted guests. I will change the description to that, but don't know how the API should look like yet. Probably a new API call to get all verified data.

app/pb/api.proto
kalvdans: Sounds good. Will adjust the comment when we implement other verification methods.

app/backend/src/couchers/errors.py
kalvdans: No, don't display this error to the user.

app/pb/account.proto
```diff
@@ -33,6 +34,19 @@ service Account {
...
     // Marks the contributor form as filled
   }
+
+  rpc ChangePhone(ChangePhoneReq) returns (google.protobuf.Empty) {
+    // Set/Change/Remove phone number
+  }
+
+  rpc SendVerificationSMS(SendVerificationSMSReq) returns (SendVerificationSMSRes) {
+    // Send phone number verification sms. The code in the SMS should be send back in VerifyPhone()
+  }
```
aapeliv: Why have these as different RPCs? I think it overly complicates the state machine, easier to just have one RPC that sets the pending number and sends a verification code, then another one to verify the sent code itself and replace the current number with the now verified one.

app/pb/account.proto
```diff
@@ -77,3 +98,49 @@ message GetContributorFormInfoRes {
...
+
+message SendVerificationSMSRes {
+  enum SendVerificationSMSStatus {
+    SMS_RESULT_CODE_SENT = 0;
+    SMS_RESULT_ALREADY_VERIFIED = 1;  // No code sent since the same number is already verified, and a renewal can't be done yet
+    SMS_RESULT_RATELIMIT = 2;   // Hold your horses, contact admin to send more than one SMS in a short period
+    SMS_RESULT_UNSUPPORTED_OPERATOR = 3;  // Sorry, we don't know how to send SMS to your number
+    SMS_RESULT_OPERATOR_ERROR = 4;  // Unknown, see operator_error_message.
+    SMS_RESULT_CONSENT_NEEDED = 5;  // Accept a later ToS version to allow couchers to use external services for sms delivery
+    SMS_RESULT_NO_PHONE_NUMBER = 6;  // User hasn't inputed a phone number yet
```
aapeliv: Error codes should be returned through gRPC StatusCodes. Also, `SMS_RESULT_CONSENT_NEEDED` is not needed, this API is "secure" so not having agreed to latest TOS means the user can never get this far.

app/pb/account.proto
```diff
@@ -77,3 +98,49 @@ message GetContributorFormInfoRes {
...
+
+  // Empty string in all cases except when status == SMS_RESULT_OPERATOR_ERROR
+  // Might not be english.
+  string operator_error_message = 2;
+}
+
+message VerifyPhoneReq {
+  // Should be a numerical string with Luhn algorithm mod 10 checksum equals 0.
+  // Frontends could do that pre-check.
+  string token = 1 [(sensitive) = true];
```
aapeliv, line 113: Hmm, why is a checksum needed? When I get SMS codes they're these days 6 digits mostly; I think it'd be good to keep them short at the expense of not having a checksum digit.

app/pb/account.proto
```diff
@@ -77,3 +98,49 @@ message GetContributorFormInfoRes {
...
+}
+
+message VerifyPhoneRes {
+  enum VerifyPhoneStatus {
+    VERIFY_PHONE_SUCCESS = 0;
+    VERIFY_PHONE_TOO_MANY_ATTEMPTS = 1;  // No brute-forcing please.
+    VERIFY_PHONE_NO_SMS_SENT = 2;   // Please do a succesfull call to SendVerificationSMS first.
+    VERIFY_PHONE_WRONG_CODE = 3;    // Either Luhn checksum error or incorrect code, try again.
+  }
```
aapeliv: Error codes should be handled through gRPC status codes

aapeliv (commented):
I don't think the user needs to re-verify once a year, this would also incur quite a big cost to us. But I do think there should be a rate limit in how often they can update their phone number

app/pb/account.proto
kalvdans: Since you only have one try per year, I thought it would be neat to see your own phone number shown by the app before pushing the verify button. But you are right it complicates the state machine a bit.

app/pb/account.proto
kalvdans: Ok, I will spread them out evently over INVALID_ARGUMENT, FAILED_PRECONDITION, RESOURCE_EXHAUSTED and ALREADY_EXISTS.

app/pb/account.proto
kalvdans: I thought it would be fun to have. Frontends are free to ignore that though. 5 digits + checksum with three attempts still gives you only 0.003 % chance of guessing it.

Let's hardcode six digits in the interface.

app/backend/src/couchers/errors.py
lucaslcode: That would require specially handling only this error compared to the others on the frontend. 

I agree in theory, but until we switch to error codes, it will make the frontend code very messy

app/pb/account.proto
aapeliv: > Since you only have one try per year, I thought it would be neat to see your own phone number shown by the app before pushing the verify button.

We should do that in the frontend!

kalvdans: Yay!

kalvdans:
Ready for review of simplified API and the implementation.

Will do migrations once you are happy with the database model!

app/backend/src/couchers/phone/sms.py
```diff
@@ -0,0 +1,20 @@
+import secrets
+
+import luhn
+
+
+def generate_random_code():
+    """Return a random 6-digit string with correct Luhn checksum"""
+    # For simplicity we force the string to not have leading zeros
+    return luhn.append(str(10000 + secrets.randbelow(90000)))
```
aapeliv: I know it's a bit of a pain, but I think it'd be better to include 0s in the first digit as well (and just treat it as text), here you could zero-pad it. That way we don't loose the 10% from leading 0s.

app/backend/src/couchers/phone/sms.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def format_message(token):
+    return "Couchers.org code: " + token + ". If not requested by you, ignore this sms. -- the couchers team"
```
aapeliv: I don't think there's a need to reduce the length, I think SMSs are 140 chars? Also, I've seen the first sentence be backwards so it shows up in a notification banner at the front, e.g.

"{token} is your Couchers.org verification code. If you did not requested this, please ignore this message. Best, the Couchers.org team."

Also, f-string?

app/backend/src/couchers/servicers/account.py
```diff
@@ -55,17 +59,16 @@ def GetAccountInfo(self, request, context):
             user = session.query(User).filter(User.id == context.user_id).one()
 
             if not user.hashed_password:
-                return account_pb2.GetAccountInfoRes(
+                auth_info = dict(
```
aapeliv, line 61: I change this around, you'll have to rebase/merge develop here

app/pb/account.proto
```diff
@@ -33,6 +34,23 @@ service Account {
   rpc MarkContributorFormFilled(MarkContributorFormFilledReq) returns (google.protobuf.Empty) {
     // Marks the contributor form as filled
   }
+
+  rpc ChangePhone(ChangePhoneReq) returns (google.protobuf.Empty) {
+    // Set/Change/Remove phone number. An verification sms with a 6-digit code will be sent out,
+    // And returned in a call to in VerifyPhone() to complete the phone number change.
+    // An INVALID_ARGUMENT is returned if the phone number is not formatted according to E.164.
+    // A RESOURCE_EXHAUSTED error code indicates that you have done too many changes lately.
+    // An UNIMPLEMENTED error code indicates that sms to your operator is not supported.
```
aapeliv, line 42: Would FAILED_PRECONDITION also be appropriate here? I'm not sure

app/pb/account.proto
```diff
@@ -33,6 +34,23 @@ service Account {
...
+  }
+
+  rpc VerifyPhone(VerifyPhoneReq) returns (google.protobuf.Empty) {
+    // Enter the code you got in the sms. Other accounts using the same phone number will
+    // lose their verification status.
+
+    // A RESOURCE_EXHAUSTED error code indicates too many tries has been performed,
+    // and FAILED_PRECONDITION means ChangePhone was not called before this method.
+    // An incorrect code generates NOT_FOUND.
```
aapeliv, line 51: Maybe FAILED_PRECONDITION would be more appropriate for this as well? NOT_FOUND feels a bit funky?

I tend to use FAILED_PRECONDITION in a lot of places...

app/pb/account.proto
```diff
@@ -47,7 +65,14 @@ message GetAccountInfoRes {
   // user's current email address
   string email = 3;
 
-  // e.g. bool account_public = 3;
+  // User's verified phone number, or empty string if no successful verification happened yet.
+  // in E.164 format (including country code without the + sign)
```
aapeliv: I think we should include the `+` sign; it's how people normally use phone numbers

app/pb/account.proto
```diff
@@ -77,3 +102,16 @@ message GetContributorFormInfoRes {
 message MarkContributorFormFilledReq {
   bool filled_contributor_form = 1;
 }
+
+message ChangePhoneReq {
+  // New phone number in E.164 format (international format without spaces and punctuation,
+  // for example "46701740605". Pass empty string to remove a verified phone number.
+  // The phone number will only be visible to accepted guests and accepted hosts.
+  string phone = 1;
```
aapeliv, line 107: I think the API should take phone numbers starting with `+`

app/backend/src/couchers/servicers/account.py
```diff
@@ -149,3 +152,63 @@ def MarkContributorFormFilled(self, request, context):
...
+            user.phone_verification_verified = None
+            user.phone_verification_token = None
+
+            session.commit()
+
+            if not phone:
+                return empty_pb2.Empty()
+
+            if not is_known_operator(user.phone):
+                context.abort(grpc.StatusCode.UNIMPLEMENTED, errors.UNSUPPORTED_OPERATOR)
```
aapeliv: This seems more like `is_valid_number` than `is_known_operator`?

I don't think libphonenumber checks operators, just whether it fits the known number patterns, which are different.

app/backend/src/couchers/servicers/account.py
```diff
@@ -149,3 +152,63 @@ def MarkContributorFormFilled(self, request, context):
...
+
+            if user.phone_verification_sent and now() - user.phone_verification_sent < timedelta(days=180):
+                context.abort(grpc.StatusCode.RESOURCE_EXHAUSTED, errors.REVERIFICATION_TOO_EARLY)
```
aapeliv: This should happen earlier, not after the phone has been removed from the user.

aapeliv (changes requested):
API looks pretty good to me, see a few small comments!

I think we should prefix all numbers (everywhere, API and in DB) with the plus: that's what AWS expects also when we send the text, I don't see any reason to leave it out (and it also conveniently makes sure we keep them as strings everywhere).

Sorry, I ended up writing a bunch of comments! Hope they're not too pedantic and too chaotic.

app/pb/account.proto
kalvdans: FAILED_PRECONDITION is already taken by the "ChangePhone was not called before this method". Maybe PERMISSION_DENIED?

app/backend/src/couchers/servicers/account.py
kalvdans: Our phonenumber database might not be up-to-date and it is a bit rude to say that my shiny new phone number is invalid.

app/pb/account.proto
aapeliv: The codes don't need to be unique: the frontends will just display the text anyway. There's many cases in the API where the status codes are reused but with a different message

app/backend/src/couchers/servicers/account.py
kalvdans: Why? Clearly you have changed phone number so the earlier one is not in your possession anymore.

app/pb/account.proto
kalvdans: I went through the list at https://github.com/grpc/grpc/blob/master/doc/statuscodes.md and thought "The operation is not implemented or is not supported/enabled in this service." sounds just about right even if it is a third party that doesn't implement it.

app/pb/account.proto
kalvdans: I disagree, I'd would see FAILED_PRECONDITION as an programming error in the frontend logic here, calling methods in the wrong order.

aapeliv: I don't think there's anything that fits cleanly, INVALID_ARGUMENT implies that you supplied arguments incorrectly, FAILED_PRECONDITION indeed sounds like something is failing before the arguments are even evaluated, but I also think NOT_FOUND feels wrong too!

I think in a lot of places elsewhere I've used INVALID_ARGUMENT when the arguments are... well, invalid, e.g. if an email is not a recognized format. I've used FAILED_PRECONDITION where the arguments are correctly formatted etc but something else is wrong/missing, and NOT_FOUND when a resource was not found.

Anyway, it doesn't really matter, not a very important thing; just take your pick!

app/backend/src/couchers/servicers/account.py
aapeliv: Good point, I had a look at phonenumber.

Just change the error message to something that implies that the error may be fixable: "Your mobile phone number is invalid or unrecognized. Please double check it and contact support if this error persists."

app/backend/src/couchers/servicers/account.py
aapeliv: Some people have multiple (active) numbers.

If the user wants to remove it because they're not in possession of it, they can do so by calling this with an empty number, we shouldn't try to "catch them" like this (even if they're not in possession of that number!).

app/pb/account.proto
kalvdans: I'll just stick with NOT_FOUND for now.

app/pb/account.proto
kalvdans: Let's add that such API later!

kalvdans:
Just realized that my test file wasn't committed, but now it is included in the branch! Ready for another review and I will do migrations tomorrow unless someone does it before me :)

app/backend/src/couchers/models.py
```diff
@@ -252,6 +249,11 @@ def display_last_active(self):
         """
         return self.last_active.replace(minute=(self.last_active.minute // 15) * 15, second=0, microsecond=0)
 
+    def phone_is_verified(self):
+        return self.phone_verification_verified is not None and now() - self.phone_verification_verified < timedelta(
+            days=356
+        )
```
aapeliv: Could you move the `timedelta` into `constants.py`, and I think make it 2 years for now.

app/backend/src/couchers/phone/sms.py
```diff
@@ -0,0 +1,20 @@
+import secrets
+
+import luhn
+
+from couchers import errors
+
+
+def generate_random_code():
+    """Return a random 6-digit string with correct Luhn checksum"""
+    return luhn.append("%05d" % secrets.randbelow(100000))
```
aapeliv: Please move this function into `crypto.py` (or just `"%05d" % secrets.randbelow(100000)` if you want to keep the luhn stuff here).

I want to keep all the cryptography functions in one place so they're easy to verify and audit!

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+            if not phone:
+                user.phone = None
+                user.phone_verification_verified = None
+                user.phone_verification_token = None
+                return empty_pb2.Empty()
+
+            if not is_known_operator(phone):
+                context.abort(grpc.StatusCode.UNIMPLEMENTED, errors.UNRECOGNIZED_PHONE_NUMBER)
+
+            if user.phone_verification_sent and now() - user.phone_verification_sent < timedelta(days=180):
```
aapeliv: Move 180 to constants too. I think we'll need to tweak these and it'll be easier to have them there

app/backend/src/couchers/models.py
```diff
@@ -194,6 +185,12 @@ class User(Base):
     new_email_token_created = Column(DateTime(timezone=True), nullable=True)
     new_email_token_expiry = Column(DateTime(timezone=True), nullable=True)
 
+    # for verifying their phone number
+    phone_verification_token = Column(String(6), nullable=True)  # randomly generated Luhn 6-digit string
+    phone_verification_sent = Column(DateTime(timezone=True), nullable=True)
+    phone_verification_verified = Column(DateTime(timezone=True), nullable=True)
+    phone_verification_attempts = Column(Integer, nullable=False, default=0)
```
aapeliv: Could you write a brief comment here for each as to when each is null and when not and what it's used for (e.g. `phone_verification_sent` is not reset when the phone is changed, it's kept to keep track of the last time a message was sent, etc)

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+                context.abort(grpc.StatusCode.RESOURCE_EXHAUSTED, errors.REVERIFICATION_TOO_EARLY)
+
+            token = sms.generate_random_code()
+            result = sms.send_sms(phone, sms.format_message(token))
+
+            if result == "success":
+                user.phone = phone
+                user.phone_verification_sent = now()
+                user.phone_verification_token = token
+                user.phone_verification_attempts = 0
```
aapeliv, line 189: I think you also need to set `user.phone_verification_verified = None`? If the user has a phone verified before, it should be unverified now, right?

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+                return empty_pb2.Empty()
+
+        context.abort(grpc.StatusCode.UNIMPLEMENTED, result)
+
+    def VerifyPhone(self, request, context):
+        if len(request.token) != 6:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.WRONG_SMS_CODE)
```
aapeliv: Good to also check the Luhn code here? If some frontend doesn't implement it, it's nice to not run out of attempts by writing the check digit wrong

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+
+        with session_scope() as session:
+            user = session.query(User).filter(User.id == context.user_id).one()
+            if user.phone_verification_token is None:
+                context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.NO_PENDING_VERIFICATION)
+
+            if user.phone_verification_attempts > 3:
+                context.abort(grpc.StatusCode.RESOURCE_EXHAUSTED, errors.TOO_MANY_SMS_CODE_ATTEMPTS)
+
+            if user.phone_verification_token != request.token:
```
aapeliv: This should be a secure compare, as insignificant as it is. (This is prone to timing attacks, though the 3 verification attempts mitigates that, still good to use a constant time compare). You can do it with `secure_compare`: that takes bytes though, you can either encode it here or add a helper function in crypto.py.

I would maybe just create `generate_sms_code()` and `verify_sms_code(request_token, right_token)` in `crypto.py`

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+                user.phone_verification_attempts += 1
+                session.commit()
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.WRONG_SMS_CODE)
+
+            # Delete verifications from everyone else that has this number
+            session.query(User).filter(User.phone == user.phone).update(dict(phone_verification_verified=None))
```
aapeliv: This looks risky; are you sure it leaves the variables in the correct states? I think you should also wipe the `user.phone` and reset all other `phone_verified_*` fields as well.

aapeliv (changes requested):
Thanks for adding tests! Just a few small comments left, but I think the design and API is great.

One thing that I think would be good would be to document the phone verification state machine a bit more clearly (in models.py) and then enforce that with database CHECK constraints and/or a few sanity assertions in the functions.

app/backend/src/couchers/servicers/account.py
```diff
@@ -150,3 +157,61 @@ def MarkContributorFormFilled(self, request, context):
...
+            if result == "success":
+                user.phone = phone
+                user.phone_verification_sent = now()
+                user.phone_verification_token = token
+                user.phone_verification_attempts = 0
+                return empty_pb2.Empty()
+
+        context.abort(grpc.StatusCode.UNIMPLEMENTED, result)
+
+    def VerifyPhone(self, request, context):
```
aapeliv, line 194: The could should also expire, say after 24 hours.

kalvdans: ?SYNTAX ERROR
[edit]
You mean the verification code? It just adds complexity.

aapeliv: Ahahahah, sorry.

I mean: the verification code should expire if the user doesn't `VerifyPhone` within some reasonable amount of time.

(That way we don't end up with this... https://bugzilla.mozilla.org/show_bug.cgi?id=1630870 [summary: GlobalSign issued a certificate with completely obsolete and non-(today's-)standards compliant cert because someone clicked a 9 year old email])

aapeliv: (I meant this one: https://bugzilla.mozilla.org/show_bug.cgi?id=1690807 wrong cert bug lol)

app/backend/src/couchers/servicers/account.py
kalvdans: Yes, thanks! Really need that state diagram :D

app/backend/src/couchers/models.py
kalvdans: Ascii art state chart pushed! See if you can understand it and if you believe it is possible to make database constraints that checks that we are in one of the boxes. (or if you have any suggested simplification).

We could initialize times to 1970-01-01 instead of None and make them nullable=False.

aapeliv: Yes, it makes sense... I need a moment to digest this :P

One thing that comes to mind is; I think `ChangePhone("")` should maybe be its own RPC (`ClearPhone`?) since it's completely decoupled?

kalvdans: `ChangePhone("")` could even be removed completely. Even if I decide to be phoneless, why wouldn't I enjoy a higher verification score for 2 years for free? :D

app/backend/src/couchers/errors.py
kalvdans: I changed the message to "Phone number must be in international format without punctuation." But I'd expect the frontend to do some text processing before passing it to the backend so it could as well check the format itself :)

The phonenumbers Python library could do local-to-international conversion, so I assume there is one for javascript as well. :)

app/backend/src/couchers/models.py
aapeliv: I think this is correct. Some small nitpicks:

* One missing thing is that `someone else verifies xx` should lead from any state to `Start`.
* There should maybe be some difference in the diagram between the variables in states `Code expired` and `Code sent` and similarly with `Verified` and `Verification expired`
* `Code sent` should lead to `Code sent` through `VerifyPhone(wrong)` as well

aapeliv: I think it'd be good to add a db constraint to basically say `(verified is not null) => ((phone is not null) and (token is null) and (sent is not null))` and a `unique index on phone WHERE verified is not null`

aapeliv: Okay, I checked the code and it seems correct. Could you add those db constraints, then I'm happy to approve this PR! (And also if you feel like tweaking the diagram, but not so important.)

kalvdans: I managed the nonimportant stuff, but the constraints I failed to do. I don't think postgres has any UNIQUE .. WHERE construction.

aapeliv: You can create a partial index that's also a unique index. This should work, if I got my logic right:

```py
    __table_args__ = (
        # If the phone is verified, must have a phone, and a sent time but no token
        # (verified is not null) => ((phone is not null) and (token is null) and (sent is not null))
        # note (p => q) iff (p OR ~q)
        CheckConstraint(
            "(verified IS NOT NULL) OR ((phone IS NULL) OR (token IS NOT NULL) OR (sent IS NULL))",
            name="phone_verified_conditions",
        ),
        # Verified phone numbers should be unique
        Index(
            "ix_users_unique_phone",
            phone,
            unique=True,
            postgresql_where=(verified != None),
        ),
    )
```

aapeliv: Also thanks for fixing up the diagram and other comments!

kalvdans: Your logic was incorrect, (p => q) iff (~p OR q), but I got the idea how to write these things. I came up with a compact expression that checks many things at once. Thanks for the secret into using an index for unique-where constraints!

app/backend/src/couchers/models.py
```diff
@@ -194,6 +185,52 @@ class User(Base):
...
+
+    phone_verification_sent = Column(DateTime(timezone=True), nullable=False, server_default=text("to_timestamp(0)"))
+    phone_verification_verified = Column(DateTime(timezone=True), nullable=True, server_default=text("NULL"))
+    phone_verification_attempts = Column(Integer, nullable=False, server_default=text("0"))
+
+    __table_args__ = (
+        # Whenever a phone number is set, it must either be pending verification or already verified.
+        # Exactly one of the following must always be true: not phone, token, verified.
+        CheckConstraint(
+            "(phone IS NULL)::int + (phone_verification_verified IS NOT NULL)::int + (phone_verification_token IS NOT NULL)::int = 1",
```
aapeliv, line 229: I think it's cleaner to write this as you wrote in the comment:

`(phone IS NULL) OR (phone_verification_verified IS NOT NULL) OR (phone_verification_token IS NOT NULL)`

app/backend/src/couchers/models.py
```diff
@@ -194,6 +185,52 @@ class User(Base):
...
+    # '-----------------'                   '-------------------'                      '-----------------------'
+
+    # randomly generated Luhn 6-digit string
+    phone_verification_token = Column(String(6), nullable=True, server_default=text("NULL"))
+
+    phone_verification_sent = Column(DateTime(timezone=True), nullable=False, server_default=text("to_timestamp(0)"))
+    phone_verification_verified = Column(DateTime(timezone=True), nullable=True, server_default=text("NULL"))
+    phone_verification_attempts = Column(Integer, nullable=False, server_default=text("0"))
+
+    __table_args__ = (
```
aapeliv, line 225: Please move `__table_args__` just after relationships, before functions for consistency, i.e. just after `blocked_user`

aapeliv (commented):
Looks good! Please see the last two minor styling nitpicks and add migrations and it's good to go! Also let me know if the constraint migrations (or anything else) are giving you grief, I can also look into that.

app/backend/src/couchers/models.py
kalvdans: It is not as powerful, doesn't catch (!phone && verified)

aapeliv: Ah yeah, you're right. Sorry!

app/backend/src/couchers/migrations/versions/80d30951919d_phone_verifications.py
```diff
@@ -0,0 +1,75 @@
...
+        sa.Column(
+            "phone_status",
+            postgresql.ENUM("unverified", "verified", name="phonestatus"),
+            autoincrement=False,
+            nullable=True,
+        ),
+    )
+    op.add_column(
+        "users",
+        sa.Column("verification", postgresql.DOUBLE_PRECISION(precision=53), autoincrement=False, nullable=True),
```
aapeliv, line 68: That's a lot of precision :P

aapeliv (dismissed):
LGTM!

kalvdans:
> LGTM!

... as in merge despite that I broke frontend? Just double-checking.

aapeliv:
No, sorry, I didn't notice. I'll fix it up and push to this branch.

kalvdans:
Thanks for a quick helping hand! Merging!

But I need another review approval, prettyplease :sad_dog_face:

aapeliv (approved):

app/backend/src/couchers/models.py
```diff
@@ -196,11 +187,57 @@ class User(Base):
...
+    phone_verification_sent = Column(DateTime(timezone=True), nullable=False, server_default=text("to_timestamp(0)"))
+    phone_verification_verified = Column(DateTime(timezone=True), nullable=True, server_default=text("NULL"))
+    phone_verification_attempts = Column(Integer, nullable=False, server_default=text("0"))
+
     avatar = relationship("Upload", foreign_keys="User.avatar_key")
 
     blocking_user = relationship("UserBlock", backref="blocking_user", foreign_keys="UserBlock.blocking_user_id")
     blocked_user = relationship("UserBlock", backref="blocked_user", foreign_keys="UserBlock.blocked_user_id")
 
+    __table_args__ = (
```
kalvdans, line 225: This definition was overridden by a later assignment to `__table_args__`. Fixed in https://github.com/Couchers-org/couchers/pull/1428 .

app/backend/src/couchers/models.py
```diff
@@ -196,11 +187,57 @@ class User(Base):
...
+        # Whenever a phone number is set, it must either be pending verification or already verified.
+        # Exactly one of the following must always be true: not phone, token, verified.
+        CheckConstraint(
+            "(phone IS NULL)::int + (phone_verification_verified IS NOT NULL)::int + (phone_verification_token IS NOT NULL)::int = 1",
+            name="phone_verified_conditions",
+        ),
+        # Verified phone numbers should be unique
+        Index(
```
kalvdans, line 233: Index should not be inside `__table_args__`. Fixed in https://github.com/Couchers-org/couchers/pull/1428 .

PR #1251, opened 2021-05-17
Discussion part 5: Wrap discussion page around community base

### LINKED ISSUE #543: Discussions

darrenvong:
This should be final part for discussion and thus closes #543.

Main changes involve refactoring/splitting the community page into a community base and the rest, so the base can be re-used on the discussion page to include the outer community page styling.

**Desktop**
Screenshot 2021-05-18 at 01 12 00

**Mobile**
Screenshot 2021-05-18 at 01 12 26

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/communities/CommunityBase.tsx
+++ b/app/frontend/src/features/communities/CommunityBase.tsx
@@ -43,12 +43,14 @@ interface CommunityBaseProps {
     community: Community.AsObject;
     communitySlug?: string;
   }): React.ReactElement;
-  communityId?: string;
+  communityId?: number;
+  defaultTab?: CommunityTab;
 }
 
 export default function CommunityBase({
   children,
   communityId,
+  defaultTab,
 }: CommunityBaseProps) {
   const classes = useCommunityBaseStyles();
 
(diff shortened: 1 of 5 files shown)


app/frontend/src/features/communities/CommunityBase.tsx
```diff
@@ -0,0 +1,131 @@
...
+    data: community,
+  } = useCommunity(communityId ?? +communityIdFromUrl);
+
+  const history = useHistory();
+
+  const { page = "overview" } = useParams<{ page: string }>();
+
+  // Use default tab from prop, otherwise derived state from community URL
+  const tab =
+    defaultTab ??
```
darrenvong, line 74: Material UI was throwing errors when a tab is unselected by passing it `undefined` (type error on `TabContext`) or an unknown value like `unselected` (works, but throws console error on `unselected` being not in the list of tabs), which kinda makes sense as I think tabs aren't designed to be unselected. Since the discussion page won't have a slug, I decided to allow this to be configurable by props too to select a "default tab" to be chosen

lucaslcode (approved):

PR #1252, opened 2021-05-17
autorotate - issue 1162

### LINKED ISSUE #1162: Image upload orientation data not taken into account
Subject: Insert Profil Picture
Description:
I chose a picture as my profile picture. I'm using Firefox on an android device

Results:
There was no preview picture, and when I accepted the picture, it often was orientated in apperently random directions. 

Backend version: develop-bf9caa27
Frontend version: develop-bf9caa27
User Agent: Mozilla/5.0 (Android 11; Mobile; rv:88.0) Gecko/88.0 Firefox/88.0
User (spoofable): leo (506)

Hassan-Mallah:
**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/media/src/media/server.py
+++ b/app/media/src/media/server.py
@@ -110,6 +110,9 @@ def create_app(
         if scale < 1:
             img = img.resize(scale)
 
+        #  autorotate before stripping the data from photo
+        img = img.autorot()
+
         # strip removes EXIF (e.g. GPS location) and other metadata
         img.jpegsave(path, strip=True, interlace=True, Q=75)
 


aapeliv:
Are you able to come up with a simple test for this? If it seems very tricky (might be), then no problem

Hassan-Mallah:
Hey,

Is this what we need? :)
I wasn't able to recreate the error on Mozilla, but i applied autorotate before removing EXIF data.
cheers

Hassan-Mallah:
> Are you able to come up with a simple test for this? If it seems very tricky (might be), then no problem

it looks like the test should be here:
`def test_strips_exif(client_with_secrets):`
i'll check after work :)

aapeliv:
The way to go would be to create a small test image (maybe a 3x3 image) with some given EXIF rotation info, and then creating a new test to make sure the image was rotated correctly, etc. I'm just not sure how annoying it is to create such a test image!

Hassan-Mallah:
hmmm, that's not clear from the issue :)
so we need to autorotate the image? the error is that we rotated the image, so we shouldn't rotate it, right?

aapeliv:
If I understand correctly; it's that a lot of software stores the image's orientation instead of actual 'moving' the pixels. So when you for example upload a portrait oriented photo from an iPhone it probably assumes the viewer (browser) will rotate it, but we strip that data away. So we need to rotate it ourselves to truly move the pixels in order for it to display correctly in the browser after exif stripping.

So you'd need to make a test image that "displays" correctly in full viewers, but whose data was not 'moved' and just has a roatiton exif tag; and then making sure that image shows up correctly after we save it.

Does that make sense?

Hassan-Mallah:
Oh!, ya, now it's clear. thnks for the explanation.

aapeliv:
@Hassan-Mallah: I know you're busy for now, so I'll merge this so that the bug fix can be deployed. You can write a new PR with a test, or just leave it; it's not so important.

PR #1263, opened 2021-05-17
Reduce massive gap between logo and couchers.org logo text

darrenvong:
As in title - the massive gap caused by the rotated chip taking up space when it doesn't need to was annoying me. Here's what it looks like now:
Screenshot 2021-05-18 at 00 55 23

**Frontend checklist**


thdk (approved):
LGTM

PR #1269, opened 2021-05-20
Use dropdown without auto completion for simple lists in edit hosting preferences form

### LINKED ISSUE #1268: Use select component from MUI instead of autocomplete for fields with limited options
For the following dropdown fields on the edit hosting preferences form, having autocomplete doesn't make much sense.

Accept smoking
Parking details
Sleeping arangement

Suggestion is to use a Select component (MUI) instead of Autocomplete component.

Note: currently tests and dev-tools are throwing warnings like:

```
console.warn
      Material-UI: The value provided to Autocomplete is invalid.
      None of the options match with `1`.
      You can use the `getOptionSelected` prop to customize the equality test.
      at useAutocomplete (node_modules/@material-ui/lab/useAutocomplete/useAutocomplete.js:267:17)
      at Autocomplete (node_modules/@material-ui/lab/Autocomplete/Autocomplete.js:417:56)
      at renderWithHooks (node_modules/react-dom/cjs/react-dom.development.js:14985:18)
      at updateForwardRef (node_modules/react-dom/cjs/react-dom.development.js:17044:20)
      at beginWork (node_modules/react-dom/cjs/react-dom.development.js:19098:16)
      at beginWork$1 (node_modules/react-dom/cjs/react-dom.development.js:23940:14)
      at performUnitOfWork (node_modules/react-dom/cjs/react-dom.development.js:22779:12)
      at workLoopSync (node_modules/react-dom/cjs/react-dom.development.js:22707:5)
 ```
 
 This should be fixed with this issue.

### PR DESCRIPTION
**Pull request by thdk**
Closes #1268 

Sleeping arrangement (or 'private space'), smoking location and parking details fields on the edit hosting preferences form now no longer allow typing text with autocompletion.

These fields now display the valid options in a dropdown (MUI select component)

Users may have 0 (unspecified) or 1 (unknown) as value for these fields. 
Before this issue, this would log warnings in the console because they were not listed as an option for the autocomplete control.

I have now added the unknown (ask me) option to the dropdown for each of these fields.
If the value of the user was 0 (unspecified) it will fallback in the UI to 1 (unknown) and be saved as 1 (unknown) if the user would submit the form.

The user may now also change a previous saved value to 'ask me'.

If this is not the expected behavior, please let me know.

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -5,7 +5,7 @@ import {
   SelectProps,
 } from "@material-ui/core";
 import classnames from "classnames";
-import React, { useMemo } from "react";
+import React from "react";
 import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
@@ -36,21 +36,6 @@ export default function Select<T extends Record<string | number, string>>({
 }) {
   const classes = useStyles();
 
-  const SelectOptions = useMemo(
-    () => () => {
-      return (
-        <>
-          {options.map((option) => (
-            <option value={option} key={option}>
-              {optionLabelMap[option]}
-            </option>
-          ))}
-        </>
-      );
-    },
-    [options, optionLabelMap]
-  );
-
   return (
     <FormControl
       variant={variant}
@@ -67,7 +52,11 @@ export default function Select<T extends Record<string | number, string>>({
           id,
         }}
       >
-        <SelectOptions />
+        {options.map((option) => (
+          <option value={option} key={option}>
+            {optionLabelMap[option]}
+          </option>
+        ))}
       </MuiSelect>
     </FormControl>
   );
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -1,0 +1,53 @@
+import {
+  FormControl,
+  InputLabel,
+  Select as MuiSelect,
+  SelectProps,
+} from "@material-ui/core";
+import classnames from "classnames";
+import React from "react";
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  formControl: {
+    "& .MuiOutlinedInput-root": {
+      borderRadius: theme.shape.borderRadius * 3,
+    },
+    "& .MuiInputBase-input": {
+      height: "auto",
+    },
+    display: "block",
+  },
+}));
+
+export default function Select({
+  id,
+  children,
+  className,
+  label,
+  variant = "outlined",
+  ...otherProps
+}: SelectProps) {
+  const classes = useStyles();
+  return (
+    <FormControl
+      variant={variant}
+      id={id}
+      className={classnames(className, classes.formControl)}
+      margin="normal"
+    >
+      <InputLabel htmlFor={id}>{label}</InputLabel>
+      <MuiSelect
+        native
+        label={label}
+        {...otherProps}
+        inputProps={{
+          name: id,
+          id,
+        }}
+      >
+        {children}
+      </MuiSelect>
+    </FormControl>
+  );
+}
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -27,12 +27,13 @@ export default function Select({
   label,
   variant = "outlined",
   ...otherProps
-}: SelectProps) {
+}: SelectProps & {
+  id: string;
+}) {
   const classes = useStyles();
   return (
     <FormControl
       variant={variant}
-      id={id}
       className={classnames(className, classes.formControl)}
       margin="normal"
     >
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -26,9 +26,11 @@ export default function Select({
   className,
   label,
   variant = "outlined",
+  options,
   ...otherProps
 }: SelectProps & {
   id: string;
+  options?: { label: string; value: string | number }[];
 }) {
   const classes = useStyles();
   return (
@@ -47,6 +49,12 @@ export default function Select({
           id,
         }}
       >
+        {options &&
+          options.map(({ value, label }) => (
+            <option value={value} key={value}>
+              {label}
+            </option>
+          ))}
         {children}
       </MuiSelect>
     </FormControl>
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -5,7 +5,7 @@ import {
   SelectProps,
 } from "@material-ui/core";
 import classnames from "classnames";
-import React from "react";
+import React, { useMemo } from "react";
 import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
@@ -20,40 +20,36 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function Select<
-  T extends { [key: string]: string } | { [key: number]: string }
->({
+export default function Select<T extends Record<string | number, string>>({
   id,
-  children,
   className,
-  data,
+  optionLabelMap,
   label,
   variant = "outlined",
   options,
   ...otherProps
-}: SelectProps & {
+}: Omit<SelectProps, "children"> & {
   id: string;
-  options?: Extract<keyof T, string | number>[];
+  options: Extract<keyof T, string | number>[];
   value?: T extends undefined ? string | number : keyof T;
-  data?: T;
+  optionLabelMap: T;
 }) {
   const classes = useStyles();
 
-  const SelectOptions = () => {
-    if (!options || !data) {
-      return null;
-    }
-
-    return (
-      <>
-        {options.map((option) => (
-          <option value={option} key={option}>
-            {data[option]}
-          </option>
-        ))}
-      </>
-    );
-  };
+  const SelectOptions = useMemo(
+    () => () => {
+      return (
+        <>
+          {options.map((option) => (
+            <option value={option} key={option}>
+              {optionLabelMap[option]}
+            </option>
+          ))}
+        </>
+      );
+    },
+    [options, optionLabelMap]
+  );
 
   return (
     <FormControl
@@ -72,7 +68,6 @@ export default function Select<
         }}
       >
         <SelectOptions />
-        {children}
       </MuiSelect>
     </FormControl>
   );
--- a/app/frontend/src/components/Select.tsx
+++ b/app/frontend/src/components/Select.tsx
@@ -20,19 +20,41 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function Select({
+export default function Select<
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: app/frontend/src/components/Select.tsx [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+  return (
+    <FormControl
+      variant={variant}
+      id={id}
+      className={classnames(className, classes.formControl)}
+      margin="normal"
+    >
+      <InputLabel htmlFor={id}>{label}</InputLabel>
+      <MuiSelect
+        native
```
darrenvong, line 47: I wonder if it's better to not use the `native` option so we get a consistent look and feel between operating systems and also against our other components

darrenvong (commented):
LGTM! Maybe the native point I raised should be relayed to Slack to ask for some input from the design team...

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -53,7 +54,7 @@ import {
   SleepingArrangement,
   SmokingLocation,
 } from "pb/api_pb";
-import { useState } from "react";
+import React, { useState } from "react";
```
lucaslcode: revert

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -95,6 +96,7 @@ const useStyles = makeStyles((theme) => ({
     paddingTop: theme.spacing(1),
   },
   field: {
+    display: "block",
```
lucaslcode: Is this necessary? It is also in your new `<Select>` component

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -250,31 +252,25 @@ export default function HostingPreferenceForm() {
...
-                forcePopupIcon
-                freeSolo={false}
-                getOptionLabel={(option) => smokingLocationLabels[option]}
-                multiple={false}
-                options={[
+              <Select
+                defaultValue={user.smokingAllowed || SmokingLocation.SMOKING_LOCATION_UNKNOWN}
+                onChange={onChange}
+                label={ACCEPT_SMOKING}
+                name="smokingAllowed"
```
lucaslcode:
```suggestion
```

Name is set in the controller

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -250,31 +252,25 @@ export default function HostingPreferenceForm() {
...
-                    label={ACCEPT_SMOKING}
-                    name="smokingAllowed"
-                    className={classes.field}
-                  />
-                )}
-              />
+                ].map((o) => (
+                  <option value={o} key={o}>
+                    {smokingLocationLabels[o]}
+                  </option>
```
lucaslcode: Instead of doing this every time, what do you think of giving `<Select>` an `options: {label: string, value: T}[]` prop and doing the map in the component?

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -250,31 +252,25 @@ export default function HostingPreferenceForm() {
...
-              <Autocomplete
-                disableClearable={false}
-                defaultValue={user.smokingAllowed}
-                forcePopupIcon
-                freeSolo={false}
-                getOptionLabel={(option) => smokingLocationLabels[option]}
-                multiple={false}
-                options={[
+              <Select
+                defaultValue={user.smokingAllowed || SmokingLocation.SMOKING_LOCATION_UNKNOWN}
```
lucaslcode: Seems like code smell to have `defaultValue` on both the controller and the rendered component. Maybe the most correct way would be to set `value` on the component from `render={({ onChange, value }) => (` ? Not totally sure

lucaslcode (changes requested):

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
thdk: yes :( seems it didn't work indeed. I've added tests now to check if the correct value from the database is set in the form.

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
thdk: will do :)

app/frontend/src/components/Select.tsx
thdk: "As the user experience can be improved on mobile using the native select of the platform, we allow such pattern."
https://material-ui.com/components/selects/#native-select

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -249,28 +248,55 @@ export default function HostingPreferenceForm() {
...
+                        SmokingLocation.SMOKING_LOCATION_OUTSIDE
+                      ],
+                  },
+                  {
+                    value: SmokingLocation.SMOKING_LOCATION_NO,
+                    label:
+                      smokingLocationLabels[
+                        SmokingLocation.SMOKING_LOCATION_NO
+                      ],
+                  },
```
lucaslcode: Haha this is worse...
What if you do `Object.entries(smokingLocationLabels).map(([key, value]) => ({label: value, value: +key}))`?
However that may be  less type safe and one could argue it's more opaque.

You decide if you want to leave it like you have done now, how it was before, or that one-liner.

lucaslcode (changes requested):
One comment - if you want to leave as is, re-request and I'll accept

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
thdk: I agree it looks worse, but it is more 'solid' because at least you don't have to write html tags (but still can using the children prop of select). I've tried it with object.entries but then we are less flexibile in the sort order of options and you can't have entries in the labels which you don't want to display in the dropdown. 
Eg. the labels contains both a label for unknown and unspecified while now I only show unknown in the dropdown.

Another option is to pass the select component an array of values and a dictionary (or a getLabel function) from which to retrieve the label for each value.

lucaslcode: I think that's a good idea - our usual use case is the protobuf enums, and in other cases, it probably won't be much more verbose than any other way

app/frontend/src/components/Select.tsx
```diff
@@ -20,19 +20,41 @@ const useStyles = makeStyles((theme) => ({
...
   id: string;
-  options?: { label: string; value: string | number }[];
+  options?: Extract<keyof T, string | number>[];
+  value?: T extends undefined ? string | number : keyof T;
+  data?: T;
 }) {
   const classes = useStyles();
+
+  const SelectOptions = () => {
+    if (!options || !data) {
```
lucaslcode: If this is the case, those props shouldn't be optional and you won't need this

app/frontend/src/components/Select.tsx
```diff
@@ -20,19 +20,41 @@ const useStyles = makeStyles((theme) => ({
...
   label,
   variant = "outlined",
   options,
   ...otherProps
 }: SelectProps & {
   id: string;
-  options?: { label: string; value: string | number }[];
+  options?: Extract<keyof T, string | number>[];
+  value?: T extends undefined ? string | number : keyof T;
+  data?: T;
```
lucaslcode: If you make the type `T extends Record<string | number, string>`, maybe you don't need `options` at all?

And what do you think about making the name `labelledOptions` or `optionsAndLabels` or something?

lucaslcode (changes requested):

app/frontend/src/components/Select.tsx
thdk: You can still provide options using the `children `prop. In that case `options `and `data `can be undefined.

app/frontend/src/components/Select.tsx
thdk: The reason I need both `options` and `data` is because the labels object contains more items than I want to display in the dropdown. I don't see how that would be different when using Record (which seems a better option here indeed) than an index signature.

lucaslcode: Oh, I see. Can you think of a better name for `data`? I wouldn't know the difference between it and `options` without looking at a previous usage or the source.

app/frontend/src/components/Select.tsx
lucaslcode: Is that necessary?

app/frontend/src/components/TextField.tsx
```diff
@@ -37,7 +36,6 @@ export default function TextField({
     <MuiTextField
       {...otherProps}
       variant={variant}
-      id={id}
```
darrenvong, line 40: Why was this removed? We need this previously so [MUI correctly links label up with the right ID](https://material-ui.com/components/text-fields/#accessibility). Removing it will break this

app/frontend/src/components/Select.tsx
```diff
@@ -0,0 +1,79 @@
...
+  ...otherProps
+}: SelectProps & {
+  id: string;
+  options?: Extract<keyof T, string | number>[];
+  value?: T extends undefined ? string | number : keyof T;
+  data?: T;
+}) {
+  const classes = useStyles();
+
+  const SelectOptions = () => {
```
darrenvong: Can we lift this out of this component? Creating a component inside a component is an anti-pattern as it creates a new function reference each time so React won't know it's the same component and so screws up its reconciliation and ends up recreating the component tree and replacing more things in the DOM than necessary

app/frontend/src/components/Select.tsx
thdk: no, but it's cool :) (I'll remove it)

app/frontend/src/components/TextField.tsx
thdk: I've also removed  `id` from the destructured props. So `otherProps` will still have the `id` prop. The spread operator used on line 38 will nicely pass all these props, including the `id` prop, to the MuiTextField component.

So this change is just a 'code style' issue.

app/frontend/src/components/Select.tsx
thdk: I can lift this out of the component. But I can also use useMemo for this. Any preference?

app/frontend/src/components/Select.tsx
thdk: what about 'optionLabelMap' ?

lucaslcode: Sounds good

app/frontend/src/components/Select.tsx
lucaslcode: Moving it out is better, but in this case you can just inline the options instead of having another component

lucaslcode (approved):
Great! Pretty sure all Darren's comments are addressed too so I'll approve :)

PR #1275, opened 2021-05-22
Add next.couchershq.org to CORS allowlist for API

aapeliv:
I've created a new site, https://next.couchershq.org that's the same as https://develop--frontend.preview.couchershq.org but easier to point non-technical folks to who still need to poke around on the staging area. This adds it to the allowlist to allow requests from there for CORS.

### DIFF
--- a/app/proxy/envoy.yaml
+++ b/app/proxy/envoy.yaml
@@ -50,6 +50,7 @@ static_resources:
                 # main site
                 - exact: https://app.couchers.org
                 # deploy preview sites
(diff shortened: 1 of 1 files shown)


darrenvong (approved):
Is this to allow the new site to access the API? LGTM if it works :D

aapeliv:
> Is this to allow the new site to access the API? LGTM if it works :D

Yes it is. Otherwise the proxy fails the requests for not coming from the right origin.

PR #1281, opened 2021-05-23
Fix login issues with email

### LINKED ISSUE #1279: Can't log in with e-mail
Subject: Can't log in with e-mail
Description:
I tried logging in with my email in the Email/Username field and my password (which has worked before in the beta), but I got a "Cannot Find this User" error message.  I could log in with my username.

Results:
I should be able to log in using either an email or a username... right?

Backend version: develop-7f962c9e
Frontend version: develop-7f962c9e
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.212 Safari/537.36
User (spoofable): natalia (198)

aapeliv:
Add user_id to AuthRes so that frontend doesn't have to do an extra query to get the user right after logging in.

Ought to fix #1279 

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/features/auth/useAuthStore.ts
+++ b/app/frontend/src/features/auth/useAuthStore.ts
@@ -73,12 +73,8 @@ export default function useAuthStore() {
         setLoading(true);
         try {
           const auth = await service.user.passwordLogin(username, password);
+          setUserId(auth.userId);
 
-          if (!auth.jailed) {
-            const user = await service.user.getUser(username);
-            setUserId(user.userId);
-            queryClient.setQueryData(["user", user.userId], user);
-          }
           //this must come after setting the userId, because calling setQueryData
           //will also cause that query to be background fetched, and it needs
           //userId to be set.
@@ -94,11 +90,7 @@ export default function useAuthStore() {
         setLoading(true);
         try {
           const auth = await service.user.completeSignup(signupArguments);
-          if (!auth.jailed) {
-            const user = await service.user.getCurrentUser();
-            setUserId(user.userId);
-            queryClient.setQueryData(["user", user.userId], user);
-          }
+          setUserId(auth.userId);
           setJailed(auth.jailed);
           setAuthenticated(true);
         } catch (e) {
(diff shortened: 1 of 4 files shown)


app/frontend/src/features/auth/useAuthStore.ts
```diff
@@ -73,12 +73,8 @@ export default function useAuthStore() {
         setLoading(true);
         try {
           const auth = await service.user.passwordLogin(username, password);
+          setUserId(auth.userId);
 
-          if (!auth.jailed) {
-            const user = await service.user.getUser(username);
-            setUserId(user.userId);
-            queryClient.setQueryData(["user", user.userId], user);
```
darrenvong, line 80: I just looked at how this differs from `getCurrentUser` - `getCurrentUser` calls the `Ping` rpc to get the current user, so I think we could do the same here and we wouldn't even need to change the API?

lucaslcode (commented):
How does this fix the problem? And how was frontend showing the wrong message?

darrenvong (approved):

aapeliv:
> How does this fix the problem? And how was frontend showing the wrong message?

It was calling `API.GetUser` as soon as the user had logged in (to get the user's ID): but it was using `username` from the login input which was really the user's email. `GetUser` no longer supports getting users by email (even the current user), so this was failing.

I saw that it was doing two requests when logging in because the frontend wants the user id in its state, so I thought it would be better to just return it in one request.

darrenvong:
> How does this fix the problem?

There's a semi-recent (?) change to the API that stops you from getting a user using an email and when the user logs in with username/email and password, in the frontend it was using the email to get the current user as soon as they are authenticated and so triggering the "cannot find this user" error message from the backend

darrenvong:
Lol looks like we were typing that at around the same time and Aapeli beats me to it 😄

PR #1282, opened 2021-05-23
SMS sending

aapeliv:
This is the code required to send SMSs using AWS SNS.

I think it needs a bit more error handling...

Also missing migrations.


kalvdans:
I think backend/feature/sms-verification should be merged to develop first. It will do the right thing reporting "unsupported operator" for any number.

aapeliv:
> I think backend/feature/sms-verification should be merged to develop first. It will do the right thing reporting "unsupported operator" for any number.

It doesn't matter much which order it's done in; it'll take a bit for frontend to implement it anyway...

But if we merge this into yours first, then we can do a single migration instead of two.

It doesn't matter either way, up to you!

kalvdans:
It is good with two migrations, so it is visible what technical dept debug information in the database will add.

app/backend/src/couchers/models.py
```diff
@@ -598,6 +598,23 @@ class Email(Base):
...
+    """
+
+    __tablename__ = "smss"
+
+    id = Column(String, primary_key=True)
+
+    # timezone should always be UTC
+    time = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
+
+    sender = Column(String, nullable=False)
```
kalvdans: sender_user_id maybe you meant?

app/backend/src/couchers/phone/sms.py
```diff
@@ -17,4 +20,33 @@ def send_sms(number, message):
...
     success, "unsupported operator" on unsupported operator, and any other
     string for any other error."""
-    return "unsupported operator"
+
+    if not config["ENABLE_SMS"]:
+        logger.info(f"SMS not emabled, need to send to {number}: {message}")
+        return
+
+    if len(request.message) > 140:
+        return "message too long"
```
kalvdans:
```suggestion
    assert len(request.message) <= 140
```

app/backend/src/couchers/phone/sms.py
```diff
@@ -17,4 +20,33 @@ def send_sms(number, message):
...
+
+    sns = boto3.client("sns")
+    sns.set_sms_attributes(
+        attributes={"DefaultSenderID": config.get("SMS_SENDER_ID"), "DefaultSMSType": "Transactional"}
```
kalvdans: Hardcode DefaultSenderID to "Couchers.org", no need for a config variable.

kalvdans: I guess you need some API key as config variable though.

aapeliv:
We've been in touch with a Twilio engineer (SMS sending API, competitor to this); so we're waiting to talk to them and see if they can help us out with their non-profit program, etc.

app/backend/src/couchers/models.py
aapeliv: This is the SMS sender ID that the SMS is sent from. I think it's a phone number.

app/backend/src/couchers/phone/sms.py
aapeliv: I think it's good to use asserts for asserting things that should be true. E.g. making sure an invariant is right, but it's not good to use asserts to verify business logic. Asserts can be turned off with a command line switch "for performance", etc.

app/backend/src/couchers/phone/sms.py
aapeliv: I think it's good to be able to change it e.g. on the "next" instance so one can disambiguate between the production deployment and staging

kalvdans: Good point. The API key is needed though I guess.

app/backend/src/couchers/phone/sms.py
```diff
@@ -18,5 +21,36 @@ def format_message(token):
...
 def send_sms(number, message):
     """Send SMS to a E.164 formatted phone number with leading +. Return "success" on
-    success, on failure return an error string for any other error."""
-    return errors.UNSUPPORTED_OPERATOR
+    success, "unsupported operator" on unsupported operator, and any other
+    string for any other error."""
+
+    if not config["ENABLE_SMS"]:
+        logger.info(f"SMS not emabled, need to send to {number}: {message}")
+        return
```
kalvdans: Move as late as possible, so we get the length check always.

app/backend/src/couchers/phone/sms.py
```diff
@@ -18,5 +21,35 @@ def format_message(token):
 
 def send_sms(number, message):
     """Send SMS to a E.164 formatted phone number with leading +. Return "success" on
-    success, on failure return an error string for any other error."""
-    return errors.UNSUPPORTED_OPERATOR
+    success, "unsupported operator" on unsupported operator, and any other
+    string for any other error."""
+
+    if not config["ENABLE_SMS"]:
+        logger.info(f"SMS not emabled, need to send to {number}: {message}")
```
kalvdans:
```suggestion
        logger.info(f"SMS not enabled, need to send to {number}: {message}")
```

app/backend/src/couchers/phone/sms.py
kalvdans: But it is a programming error if we generate a too long message. Whatever user input should not cause the sms to become too long.

aapeliv:
Hmmmm, good point. Though I’m worried if we ever have a reason to send arbitrary SMSs everywhere in the codebase?

kalvdans:
> Hmmmm, good point. Though I’m worried if we ever have a reason to send arbitrary SMSs everywhere in the codebase?

Let's add unit tests on all sms formatting code that tries with a super long username, largest possible integer etc.

aapeliv:
@kalvdans: I've cleaned this up and addressed review comments + added a test and migrations. I think it's ready to merge after review!

aapeliv:
@kalvdans: would you have a chance to look at this one? It's ready to go

app/backend/src/couchers/phone/sms.py
```diff
@@ -18,5 +22,37 @@ def format_message(token):
...
+    string for any other error."""
+
+    assert len(message) <= 140, "message too long"
+
+    if not config["ENABLE_SMS"]:
+        logger.info(f"SMS not enabled, need to send to {number}: {message}")
+        return
+
+    sns = boto3.client("sns")
+    sender_id = config.get("SMS_SENDER_ID")
```
kalvdans:
```suggestion
    sender_id = config["SMS_SENDER_ID"]
```

Will give an easier to understand traceback, than passing None in the dict later on, in case one forgets to set this env variable.

app/backend/src/tests/test_fixtures.py
```diff
@@ -680,6 +680,10 @@ def testconfig():
     config["VERSION"] = "testing_version"
     config["BASE_URL"] = "http://localhost:3000"
     config["COOKIE_DOMAIN"] = "localhost"
+
+    config["ENABLE_SMS"] = False
+    config["SMS_SENDER_ID"] = "invalid"
```
kalvdans, line 685: The SMS_SENDER_ID config variable is never used in tests, remove.

kalvdans (dismissed):
Sorry for a late follow-up review!

Two minor comments, LGTM with or without fixing them. Great with the extensive test stubbing out boto3.

app/backend/src/couchers/phone/sms.py
aapeliv: Thanks! Good point

PR #1284, opened 2021-05-24
Default profile headings

### LINKED ISSUE #1246: Default headings for profile fields
see #410 for the default headings we want. These can just be in a constants file.

This should have the following behaviour:

- On page load, if a particular field is blank, it should be preloaded with the default headings for that field
- On save, if a particular field contains the default, it should be saved blank (not with the defaults)

darrenvong:
Closes #1246 

The headings in #410 seems a bit outdated? I chose the ones that made the most sense for the sections we got now.

This is what it looks like if the "Who I am" / "What I do in my free time" fields are empty:

Screenshot 2021-05-24 at 02 42 32

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/components/Markdown.tsx
+++ b/app/frontend/src/components/Markdown.tsx
@@ -15,16 +15,14 @@ const useStyles = makeStyles((theme) => ({
   root: {
     fontSize: theme.typography.fontSize,
     fontFamily: theme.typography.fontFamily,
-    "& h1": {
+    "& h1, & h2, & h3, & h4, & h5, & h6": {
       borderBottom: "none",
       paddingBottom: 0,
-      ...theme.typography.h1,
+      marginBottom: 0,
+      marginTop: theme.spacing(2),
     },
-    "& h2": {
-      borderBottom: "none",
-      paddingBottom: 0,
-      ...theme.typography.h2,
-    },
+    "& h1": theme.typography.h1,
+    "& h2": theme.typography.h2,
     "& h3": theme.typography.h3,
     "& h4": theme.typography.h4,
     "& h5": theme.typography.h5,
--- a/app/frontend/src/components/Markdown.tsx
+++ b/app/frontend/src/components/Markdown.tsx
@@ -17,10 +17,12 @@ const useStyles = makeStyles((theme) => ({
     fontFamily: theme.typography.fontFamily,
     "& h1": {
       borderBottom: "none",
+      paddingBottom: 0,
       ...theme.typography.h1,
     },
     "& h2": {
       borderBottom: "none",
+      paddingBottom: 0,
       ...theme.typography.h2,
     },
     "& h3": theme.typography.h3,
--- a/app/frontend/src/components/MarkdownInput.tsx
+++ b/app/frontend/src/components/MarkdownInput.tsx
@@ -14,10 +14,8 @@ const useStyles = makeStyles((theme) => ({
       "& h1, & h2, & h3, & h4, & h5, & h6": {
         borderBottom: "none",
         paddingBottom: 0,
-        marginBottom: theme.spacing(1),
-        "&:not(:first-of-type)": {
-          marginTop: theme.spacing(3),
-        },
+        marginBottom: 0,
+        marginTop: theme.spacing(2),
       },
       "& h1": {
         ...theme.typography.h1,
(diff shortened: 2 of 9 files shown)


app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
```diff
@@ -47,10 +44,43 @@ describe("EditHostingPreference", () => {
...
+  it(`should not submit the default headings for the '${ABOUT_HOME}'section`, async () => {
+    jest.isolateModules(() => {
+      getUserMock.mockImplementation(async (user) => ({
+        ...(await getUser(user)),
+        aboutPlace: "",
+      }));
+      jest.unmock("components/MarkdownInput");
+      const EditHostingPreference = require("./EditHostingPreference").default;
+      renderPage(EditHostingPreference);
+    });
```
darrenvong: I was mocking this before as the toast UI editor was quite heavy to render and since interaction with it is a bit dodgy in the tests (keyboard events were firing out of order for some bizarre reasons that I still haven't figured out, but this doesn't happen in the browser...), I thought it's not worth automate testing it and so better to mock it to make the tests run faster.

However, by mocking the input, I introduced a different problem... the mock (it's a literal swap to using a bog standard `<input>`) is introducing a different behaviour when the submit button is clicked that doesn't happen with the markdown input (it's wrongly submitting the default title with the mocked input).

So for this test only, I needed to unmock the markdown component to get the same behaviour. But doing `jest.unmock` has no effect when the component is imported at the top level, since it's already loaded with the mocked version, hence this funky looking setup with `jest.isolateModules`.

app/frontend/src/features/profile/edit/constants.ts
```diff
@@ -0,0 +1,19 @@
+export const DEFAULT_ABOUT_ME_HEADINGS = `# Current mission
+
```
lucaslcode: Can you add a `<br>` to all of these so people can just click in the space and type

lucaslcode (changes requested):
Just some minor usability comment. 
Could you also fix up the margins in the markdown renderer and editor? Right now, when there's a heading, paragraph, and heading, the margin looks the same above and below the paragraph. It should be much smaller between the heading and it's paragraph

app/frontend/src/features/profile/edit/EditHostingPreference.test.tsx
darrenvong: Some weird shit is happening here, the tests seem to work again with the mocks (as I'd expect in the first place) locally. I hate frontend tests sometimes, let's see if CI agrees... 🤦

app/frontend/src/components/MarkdownInput.tsx
```diff
@@ -11,12 +11,18 @@ const useStyles = makeStyles((theme) => ({
     "& .tui-editor-contents": {
       fontSize: theme.typography.fontSize,
       fontFamily: theme.typography.fontFamily,
-      "& h1": {
+      "& h1, & h2, & h3, & h4, & h5, & h6": {
         borderBottom: "none",
+        paddingBottom: 0,
```
lucaslcode, line 16: How about adding a `margin: 0` here, there are still some weird ones of a few pixels. But maybe it won't work, I don't think jss respects object order?

app/frontend/src/components/Markdown.tsx
```diff
@@ -17,10 +17,12 @@ const useStyles = makeStyles((theme) => ({
     fontFamily: theme.typography.fontFamily,
     "& h1": {
```
lucaslcode: This should probably be have the same `"& h1, & h2, & h3, & h4, & h5, & h6"`as the other one

lucaslcode (changes requested):

app/frontend/src/components/MarkdownInput.tsx
darrenvong: Haven't time to look into properly yet (maybe tomorrow...), but the only place where that might not work is for the `first-of-type` rule I got below due to higher specificity. It's actually why I wrote that so it matches the highly specific CSS Toast UI sets on titles by default...

When I implemented your suggested change with the clamp/margin fallback in discussions, since objects can't have the same key, doing the `& { key: ... }` workaround after `key: <fallback vale>` works so it might be respecting object order somehow? 🤷

darrenvong:
@lucaslcode Have another look and see what you think. The rendering of paragraphs seem to differ between the input and the read only view - in the input, the paragraphs are rendered as `<div>`s whereas they are actually shown as `<p>`s in the read only view.

lucaslcode (approved):

PR #1290, opened 2021-05-24
Change profile medium screens

### LINKED ISSUE #1019: Change style sheet for profile page on medium-sized screens?
Profile layout looks good on large PC screens and on phone, but the "phone" layout also seems to apply for medium-sized screens (below ~960 px), such that the profile picture now takes up the entire page width and looks humongous. 
image

I think it would be better if the intermediate-screen-size had a more similar layout to the big-screen size (profile picture + box on the left, and text on the right). If there is a concern with fitting things width-wise that way, maybe the "About Home" section could turn into single-column if needed.

Feel free to close this issue if this type of design is intentional.

Jesus82:
Closes #1019


lucaslcode:
Thanks! It's better but it still looks rather odd with a width of 700px - 950px. Maybe the left section needs a max width?
Also, the profile pic is too large in the sm breakpoint. I think it needs a maximum size.

Jesus82:
I set the left column to be half the size of the right column in the medium breakpoint. Also, I set a max width for the avatar image at 288px. In the last commit, I made some spacing fixes in order to make the profile page look like Figma, (basically spacing and paddings), as I was messing around with them when making the commit, let me know if that's OK.

app/frontend/src/components/Divider.tsx
```diff
@@ -4,8 +4,8 @@ import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
   root: {
-    marginBottom: theme.spacing(1),
-    marginTop: theme.spacing(1),
+    marginBottom: theme.spacing(2),
+    marginTop: theme.spacing(3),
```
lucaslcode: I don't think it makes sense for the divider not to have the same margin either side. If there's unevenness somewhere, it should be fixed in the relevant spots.

We don't want margins on components as a rule (set the in layouts using `className` instead), because it affects things "outside" the component and can subtly break things. (Obviously here there has to be margin since it's more like setting the height)

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -37,8 +37,9 @@ import makeStyles from "utils/makeStyles";
 const useStyles = makeStyles((theme) => ({
   card: {
     flexShrink: 0,
-    margin: theme.spacing(2),
-    padding: theme.spacing(2),
+    borderRadius: theme.spacing(1),
+    margin: "1.5rem 0.75rem 1.5rem 1rem",
```
lucaslcode:
```suggestion
    margin: theme.spacing(3, 1.5, 3, 2),
```

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -84,12 +97,14 @@ export default function Overview({ user, setIsRequesting }: OverviewProps) {
   return (
     <Card className={classes.card}>
       <Avatar user={user} className={classes.grow} />
```
lucaslcode:
```suggestion
      <Avatar user={user} grow />
```

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -37,8 +37,9 @@ import makeStyles from "utils/makeStyles";
 const useStyles = makeStyles((theme) => ({
   card: {
     flexShrink: 0,
-    margin: theme.spacing(2),
-    padding: theme.spacing(2),
+    borderRadius: theme.spacing(1),
```
lucaslcode:
```suggestion
    borderRadius: theme.shape.borderRadius * 2,
```

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -49,26 +50,38 @@ const useStyles = makeStyles((theme) => ({
...
+    paddingBottom: theme.spacing(2),
     "& > *": {
       marginBlockStart: theme.spacing(1),
     },
   },
+
   grow: {
-    paddingTop: "100%",
+    paddingTop: "min(288px, 100%)",
   },
```
lucaslcode: Wait, I know this wasn't you, but this class shouldn't be here!

app/frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -14,16 +14,27 @@ const useStyles = makeStyles({
...
+
   defaultSize: {
     height: "3rem",
     width: "3rem",
   },
+
   grow: {
     height: 0,
-    paddingTop: "100%",
+    paddingTop: "min(288px, 100%)",
```
lucaslcode: clever

app/frontend/src/components/Avatar/Avatar.tsx
```diff
@@ -14,16 +14,27 @@ const useStyles = makeStyles({
     position: "absolute",
     top: 0,
     width: "100%",
+    maxWidth: "288px",
```
lucaslcode: How did you pick 288? Maybe leave a comment above it

lucaslcode (changes requested):
It looks great now! Just a few small things.
Also, I noticed that the bottom of the cards don't line up at all widths.

app/frontend/src/components/Avatar/Avatar.tsx
Jesus82: It is 12 times our 24px grid. I could leave it as `theme.spacing(36) `

Jesus82:
**About the divider:**
 I agree with the point of not spacing the components. I changed it to equal spacing (`16px` top and bottom). The point is that the divider is often used with text above and/or below. As text always have extra spacing due to `line-height`, there is often need for adjusting. Here in this screen, 2 out of 3 lines would need extra `margin-top` or extra `margin-bottom` in order to keep even spacing.

**Needs extra `margin-top`**

Captura de pantalla de 2021-05-26 10-45-56

**Needs extra `margin-bottom`**

Captura de pantalla de 2021-05-26 11-31-34

This can be adjusted with an extra `div` and writing an extra class on each component. But in the long run we will end up with hundreds of classes that will basically be either `margin-top: 0.5rem` or `margin-bottom 0.5rem` (as this is a very common use-case). That doesn't scale well and it slows down CSS rendering.

Instead, I would propose to use pure html for this. Right now we have a dummy `divider` component just with styles wrapping a MUI component that is dummy as well.

I would instead use `hr` as plain html, so:

`<hr class="o-divider" >`

could be used the "default" divider instead of `<Divider />`, and

`<hr class="o-divider o-divider--extra-top-1" >`
`<hr class="o-divider o-divider--extra-bottom-1" >`
`<hr class="o-divider o-divider--extra-top-1 o-divider--extra-bottom-1" >`

would be used when needed to add `8px` to `margin-top`, `8px` to  `margin-bottom` or `8px` to both. In CSS we would just need to add 3 classes that would be non-scoped:

```
.o-divider : {
    border: none;
    height: 1px;
    margin: 1rem 0;
    flex-shrink: 0;
    background-color: rgba(0, 0, 0, 0.12);
}

.o-divider--extra-top-1 {
    margin-top: 1.5rem
}

.o-divider--extra-bottom-1 {
    margin-bottom: 1.5rem
}
```

(here `o-` would be a namespacing for objects, I'm following the BEMIT convention in order to avoid clashes)

Let me know what you think!

app/frontend/src/components/Avatar/Avatar.tsx
lucaslcode: Ok! Actually maybe make it `18rem` - that way it will respond to user don't size preferences, and won't break if we want to change our spacing later (We try to keep that function for margins/padding only)

lucaslcode:
Yes, it is a common use case. However it is also used in other places like the search filter dialog. 

I think you've got the right idea with a easy way to add it, but let's not introduce another customisation system. Instead maybe you want to add two more props to `<Divider>` like `padTop` and `padBottom` (or some other name) which add another jss `className`

Or even better, until we have need of bottom padding (maybe we won't), just one prop - `text` - for text dividers which gives the extra top margin. 

So in this page they would be `<Divider text />`

Jesus82:
I set both props `margTop` and `margBottom` as I already needed, and added them in the html. Let me know if you see anything else! 

Edit: Only thing missing right now is the background color, but for that we should solve the `height:100%` issue in the `body`

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -129,17 +135,21 @@ export default function Overview({ user, setIsRequesting }: OverviewProps) {
...
-        description={COMMUNITY_STANDING_DESCRIPTION}
-      />
-      <BarWithHelp
-        value={user.verification || 0}
-        label={VERIFICATION_SCORE}
-        description={VERIFICATION_SCORE_DESCRIPTION}
-      />
+      <Divider margBottom />
+      <div className={classes.bar}>
+        <BarWithHelp
```
lucaslcode: If this component doesn't support `className` yet, better add it in rather than wrap it with a div

app/frontend/src/components/Divider.tsx
```diff
@@ -1,16 +1,38 @@
...
+  },
+
+  marginBottomExtra1: {
+    marginBottom: theme.spacing(3),
   },
 }));
 
-export default function Divider() {
+export interface DividerProps {
+  margBottom?: boolean;
```
lucaslcode: If there were no other changes I wouldn't have mentioned, but can this just be `marginTop` and `marginBottom`

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -37,38 +37,42 @@ import makeStyles from "utils/makeStyles";
...
+
   intro: {
     display: "flex",
     justifyContent: "center",
   },
+
+  wrapper: {
+    marginTop: theme.spacing(1),
+    "& h1": {
+      lineHeight: "3rem",
```
lucaslcode: Hmm, I'm a bit worried that this will subtly break if we want to change h1's size in the theme. Is there a way around that?

app/frontend/src/features/profile/view/ProfilePage.tsx
```diff
@@ -39,16 +38,25 @@ const useStyles = makeStyles((theme) => ({
...
+      margin: theme.spacing(0, 3),
+      padding: 0,
+      paddingTop: theme.spacing(3),
+      paddingBottom: theme.spacing(12),
+    },
     [theme.breakpoints.up("md")]: {
-      display: "flex",
-      maxWidth: theme.breakpoints.values.lg,
+      gridTemplateColumns: "1fr 3fr",
+      maxWidth: theme.spacing(123),
```
lucaslcode: Use rem for widths

lucaslcode (changes requested):
Just nitpicks now! I can investigate the #root problem more tonight or tomorrow

app/frontend/src/components/Divider.tsx
Jesus82: I didn't want it to have the exact same name as the css property to avoid confussion.

app/frontend/src/features/profile/view/Overview.tsx
Jesus82: If we change the h1 size we will have also to review this spacing.

app/frontend/src/components/Divider.tsx
Jesus82: I did the following: as I had to set another prop for margins in barWithHelp, I used the same syntax (borrowed from Tailwind), so here the props are `mt3` and `mb3` (they stand for `margin-top-3` and `margin-bottom-3`) and in `barWithHelp` the prop will be `mb2`.

app/frontend/src/features/profile/view/Overview.tsx
lucaslcode: Can you use a unitless line height instead (which is a multiple of font size)?

app/frontend/src/components/Divider.tsx
```diff
@@ -1,16 +1,38 @@
...
+
+  marginBottom3: {
+    marginBottom: theme.spacing(3),
   },
 }));
 
-export default function Divider() {
+export interface DividerProps {
+  mb3?: boolean;
+  mt3?: boolean;
```
lucaslcode: After thinking about it a bit more, I think just use `extraMarginTop` or bottom like your class name. We want to avoid using utility classes so I don't think this naming makes sense - especially considering there is no `mb2` or `mb1`, and it isn't clear for people who haven't used TailwindCSS before.

app/frontend/src/components/Bar/BarWithHelp.tsx
```diff
@@ -7,28 +8,39 @@ import ScoreBar from "./ScoreBar";
...
+    marginBottom: theme.spacing(2),
+  },
 }));
 
 interface BarWithHelpProps {
   value: number;
   label: string;
   description: string;
+  mb2?: boolean;
 }
```
lucaslcode, line 27: Use `className` - we aren't using utility classes in general and `className` is more flexible

lucaslcode (changes requested):

app/frontend/src/components/Bar/BarWithHelp.tsx
Jesus82: unitless line-height is pretty messy to use https://stackoverflow.com/questions/12303169/line-height-without-units#answer-12304722 and an lead to lots of layouts problems difficult to debug. When set to a fixed unit it makes things easier. In this context, the line-height is fixed by design, and it will change if design changes, imo it's way better to set it with units.

Jesus82:
unitless line-height is pretty messy to use https://stackoverflow.com/questions/12303169/line-height-without-units#answer-12304722 and an lead to lots of layouts problems difficult to debug. When set to a fixed unit it makes things easier. In this context, the line-height is fixed by design, and it will change if design changes, imo it's way better to set it with units.

app/frontend/src/components/Divider.tsx
Jesus82: We're actually using them as utility (atomic) classes whether if we reflect that on naming or not. 

I can name them here `extraMarginTop`, but in the future we may have to add more margins if the design requires so. Also, in order to get spacing right, we'll have to use lots of these in different components. So in the end we'll have tens of atomic classes with basically the same properties (duplicated code) but different names (not systematized) and scattered all around.

lucaslcode: In that case we should use `className`.

I know there are lots of benefits to utility classes, but we have decided earlier on that in this app it makes more sense (or maybe, it was the overall preference) to stick with JSS classes.

If you feel strongly about it, we can see what people think on slack. It's certainly not very clear-cut in this case.

lucaslcode:
What kind of layout problems? In the link, they quite MDN saying unitless is the preferred way to set line height.

app/frontend/src/components/Divider.tsx
Jesus82: So are we just talking about naming, isn't it? Would it be OK marginBottom1, marginBottom2, marginBottom3, etc?

Jesus82:
It was more about the second comment (the linked one)

> 
> Yes, it is a ratio: 1.5 means 1.5 times the font size of the element. So it means the same as 1.5em or 150%, but with one important exception: in inheritance, when a pure number is used, the number is inherited, not the computed value.
> 
> So if you have an element with font size 24pt, line-height: 1.5 sets the line height to 36pt. But if the element has a child, i.e. an inner element, with font size of 10pt, and without any line height set on it, then the child inherits the line-height value of 1.5, which means 15pt for that element. If, on the other hand, the line height were set to 1.5em or 150%, then the child would inherit the computed value of 36pt, creating grotesque line spacing.
> 
> Technically, this is hidden under a formulation that says. for a pure number used as line-height value: “The computed value is the same as the specified value.” So, nominally, the child inherits the “computed” value of 1.5, which will then be interpreted in the context of the child (1.5 times its font size).

A fixed unit is far more consistent, with no drawbacks. Also, take in account that our spaces are "fixed blocks" of 8px, while the unitless tend to give as a result things like semipixels and values out of grid. So if I set unitless to a 24px h1, I will have to change the line height from 3rem to 2. Then if somebody changes h1 to 25px, line height will then be 50px that is out of the grid. In cases where you have to set fractions in unitless the result could lead to subpixels.

In summary, as a good practice I try to avoid unitless (and I would say that is rarely used in good layouts), but if you feel is needed let me know and I'll add that.

app/frontend/src/components/Divider.tsx
lucaslcode: Just `extraMarginBottom` and if more options are needed, use `className`

lucaslcode:
There is a drawback - if in the future the font size is changed in the theme, all valued line-heights need to be found and recalculated, whereas a unitless line height will be fine. 

It's a good point about the grid, but we tend to be very loose with the grid anyway, so I think it's not worth the extra maintenance burden. 

I don't think inheritance is a factor for us, since we are only setting it on header elements which never have children of other sizes.

app/frontend/src/components/Divider.tsx
Jesus82: Sorry if I'm missing your point here :S, but what do you mean as "use `className`"? Could you give an example?

Jesus82:
I'm ok with setting is a unitless, but tbh I don't get what would be the less manteinance burden. Could you maybe ellaborate that, so I have something to stick on?

lucaslcode:
Sure - the design team decides h1, now 1.5rem, should be 2rem and we change it in the theme, if we have used a rem or px value, we will have to find every instance of `lineHeight` and recalculate it for the new font size (eg. 3rem would become 4rem). If we use a unitless size (here would be `lineHeight: 2`), no change would be needed.

Unless I'm misunderstanding why you added line height here?

app/frontend/src/components/Divider.tsx
lucaslcode: I haven't been very clear, sorry. I think there's two things:

Components should be self-contained, and part of this is not usually setting their own margin. This is because when building layouts, if there's any variation needed in margins, this will have to happen outside the component anyway.

The other part is that when we are building page layouts, we use jss styling, which is the `const useStyles = makeStyles({ exampleClass: { marginTop: ...` and then in the page layout, `const classes = useStyles(); return ...<MyComponent className={classes.exampleClass}>...`. There are things equivalent to utility classes in Material-UI we could use (the `Box` component), but we previously decided against using them because it was the overall preference not to have styling in two places (even if it's the same file).

Taking those two things into account, I'm okay with having `<Divider extraMarginBottom>` if we don't usually need to customize the extra margin amount (seems you all used the same amount in this PR), but if we start needing `<Divider extraMarginBottom={2}>` etc, to me that's just building layouts and should use jss classes like `<Divider className={classes.myDivider}>` (or more likely, a class on the components *around* the divider).

Jesus82: I'd rather put the class outside as you say, but it gives me an error (don't know if I'm doing something wrong here)
Captura de pantalla de 2021-05-27 08-42-04

lucaslcode: You'll need to change the custom `Divider` component so it accepts a `className` prop and passes it through

Jesus82: Is there any working example of this in another component? I tried to do that first thing and I wasn't able to accomplish that.

lucaslcode: https://github.com/Couchers-org/couchers/blob/48594f38354b19cca66ae5f983b80ed9b77ebd6b/app/frontend/src/components/Alert.tsx#L35 

Note in this component, the type of the props extends `MuiAlertProps`, you can just add `className` directly to the interface instead.

darrenvong:
> A fixed unit is far more consistent, with no drawbacks.

Since you are using `rem` here, this is not really an issue, but in general the biggest drawback with using a fixed unit `line-height` (especially when used against `px`) is that it's bad for accessibility, as it won't scale correctly if the user changes their browser's default font size. So MDN recommends using unitless line height for good reasons.

I'd rather our app be usable than chasing for "pixel perfection" when our grid isn't very precise anyway. If this flexibility is breaking designs, this is something we need to work with designs to get better at as a team so we don't come up with brittle designs that will break easily. This will be important especially as we consider having the app across multiple languages further down the line.

Jesus82:
I think I fixed all the issues. Finally I used margins to fix the name spacing, so no custom lineHeight

lucaslcode (dismissed):
Looks good!

aapeliv:
@Jesus82: are you merging this?

Jesus82:
> @Jesus82: are you merging this?

Sorry, I didn't know I had the permissions to merge. I've just seen there are conflicts now. Will check that tomorrow.

lucaslcode:
I will merge this to save you dealing with the conflict which I caused

lucaslcode (approved):

PR #1293, opened 2021-05-24
Search testing

lucaslcode:
Testing follow-up to #1154 

**Frontend checklist**
- [x] All tests pass


app/frontend/src/utils/hooks.test.ts
```diff
@@ -26,3 +28,79 @@ describe("useSafeState hook", () => {
...
+  it("works with expected loading state and result", async () => {
+    const { result, waitFor } = renderHook(() => useGeocodeQuery());
+    expect(result.current).toMatchObject({
+      isLoading: false,
+      error: undefined,
+      results: undefined,
+      query: expect.anything(),
+    });
+    await act(() => result.current.query("test"));
+    expect(result.all[1]).toMatchObject({
```
lucaslcode, line 52: @darrenvong  `result.current` here is already the final result, any ideas? Or is this fine?

darrenvong: Maybe don't await on the act and use `await waitForNextUpdate()`?

Otherwise a little obscure for reading, but it's probably ok as it at least shows the states the hook went through from start to finish in `result.all`

lucaslcode: Didn't work :( I'll just leave it

app/frontend/src/features/search/FilterDialog.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+        }}
+      />
+    </>,
+    {
+      wrapper: hookWrapper,
+    }
+  );
+};
+
+describe("FilterDialog", () => {
```
lucaslcode, line 51: @darrenvong need your testing wisdom again, I'm getting `An update to FilterDialog inside a test was not wrapped in act` warnings (I thought act is not necessary when using `userEvent`?)

darrenvong: After some digging (since there's nothing obviously wrong in your test), it looks like this is a quirk with `react-hook-form` that others have hit as well, namely the validation happens async which falls outside of the synchronous `act` boundary that `userEvent` would have wrapped around things like `type` - https://github.com/testing-library/user-event/issues/128#issuecomment-720119616

So it looks like the recommended thing to do by the linked comment is to "wait a bit" for the async operations from `react-hook-form` to complete properly. Wrapping the assertions inside an `await waitFor(() => { // assertions }` seem to be enough

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -76,6 +77,34 @@ export default function FilterDialog({
     onClose();
   });
 
+  //because we are abusing URLSearchParams as a kind of global mutable state, we need
+  //to calculate the default values and put them in a ref so they don't change
+  //otherwise we get console warnings about defaultValue changing
+  const defaultValues = useRef({
```
darrenvong, line 79: This `URLSearchParams` thing feels "wrong", but it's pretty neat at the same time since the UI is still updating without any React state updates by embracing uncontrolled elements and using the URL as the eventual single source of state. I feel like I'm still wrapping my head around this 😄

app/frontend/src/features/search/LocationAutocomplete/LocationAutocomplete.test.tsx
```diff
@@ -71,4 +72,65 @@ describe("LocationAutocomplete component", () => {
...
+        name: "test location",
+        simplifiedName: "test location",
+        location: new LngLat(1, 2),
+      },
+      onChange
+    );
+
+    const input = (await screen.findByLabelText(LOCATION)) as HTMLInputElement;
+    expect(input).toBeVisible();
+    expect(input.value).toBe("test location");
```
darrenvong:
```suggestion
    expect(input).toHaveValue("test location");
```
In this case, the readability is negligible, but I think the custom matcher from React Testing Library is supposed to give clearer error messages when things fail so more that it's slightly better for debugging tests.

app/frontend/src/features/search/FilterDialog.tsx
lucaslcode: Yeah I completely agree.

I could change it to use a provider in `SearchPage`. It would cause a big re-render but I could keep the "pending" filters in a ref and "active" filters in state so it only re-renders when the dialog closes...

I think I will do that - it will be a lot more maintainable actually because the current way isn't very type-safe

lucaslcode: Unfortunately I still needed this bit to prevent the warning, but I think the refactor was worth it anyway :P

app/frontend/src/features/search/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -27,19 +27,19 @@ export default function LocationAutocomplete({
...
-  const handleChange = (value: GeocodeResult | string) => {
+  const handleChange = (value: GeocodeResult | string | null) => {
     //workaround - autocomplete seems to call onChange with the string value on mount
     //this line prevents needing to reselect the location even if there are no changes
     if (value === controller.field.value.simplifiedName) return;
 
     controller.field.onChange(value);
-    if (value === "") {
-      onChange(value);
+    if (value === "" || value === null) {
```
darrenvong, line 36: How come you have to handle `null` now?

app/frontend/src/features/search/FilterDialog.tsx
```diff
@@ -59,73 +58,67 @@ const useStyles = makeStyles((theme) => ({
...
           <LocationAutocomplete
             control={control}
-            defaultValue={defaultValues.current.location}
+            defaultValue={defaultValues.location}
             onChange={(value) => {
               if (value === "") {
-                searchParams.delete("location");
-                searchParams.delete("lat");
-                searchParams.delete("lng");
+                searchFilters.change("location", undefined);
```
darrenvong: Probably a personal preference/style thing, but I find using `undefined` explicitly as a function argument a bit strange to read, considering it's meant to represent values that are still to be set/initialised, rather than an intentionally missing one (what I'd expect `null` for).

The intent of this might be clearer if we also have a `searchFilters.clear` or `searchFilters.remove` function for this?

app/frontend/src/features/search/useSearchFilters.ts
```diff
@@ -0,0 +1,88 @@
...
+export interface SearchFilters extends ServiceUserSearchFilters {
+  location?: string;
+}
+
+export default function useSearchFilters(route: string) {
+  const location = useLocation();
+  const [active, setActive] = useState<SearchFilters>(
+    locationToFilters(location)
+  );
+  const pending = useRef<SearchFilters>(locationToFilters(location));
```
darrenvong:
```suggestion
  const pending = useRef<SearchFilters>(active);
```
Since I believe `active` will only be used for initialising the ref only and so you don't have to calculate it again?

app/frontend/src/features/search/useSearchFilters.ts
```diff
@@ -0,0 +1,88 @@
...
+  }, []);
+
+  return { active, change, apply };
+}
+
+export function locationToFilters(location: Location) {
+  const searchParams = new URLSearchParams(location.search);
+  const filters: SearchFilters = {};
+  Array.from(searchParams.keys()).forEach((untypedkey) => {
+    const key = untypedkey as unknown as keyof SearchFilters;
```
darrenvong:
```suggestion
    const key = untypedkey as keyof SearchFilters;
```
Would this alone not work since `untypedkey` would be a string to start with?

darrenvong (changes requested):
Haven't played around with the actual thing yet, but looks like a good refactor for better type safety indeed 🚀

app/frontend/src/features/search/useSearchFilters.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export interface SearchFilters extends ServiceUserSearchFilters {
+  location?: string;
+}
+
+export default function useSearchFilters(route: string) {
+  const location = useLocation();
+  const [active, setActive] = useState<SearchFilters>(
+    locationToFilters(location)
+  );
```
darrenvong:
```suggestion
  const [active, setActive] = useState<SearchFilters>(
    () => locationToFilters(location)
  );
```
So `locationToFilters` is lazily evaluated and run once for initialising the state only

app/frontend/src/features/search/LocationAutocomplete/LocationAutocomplete.tsx
lucaslcode: I made it clearable

app/frontend/src/features/search/useSearchFilters.ts
lucaslcode: Weird, I thought I tried this.

app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -4,59 +4,47 @@ import { FilterIcon } from "components/Icons";
...
 import { OPEN_FILTER_DIALOG, USER_SEARCH } from "features/search/constants";
 import FilterDialog from "features/search/FilterDialog";
+import useSearchFilters from "features/search/useSearchFilters";
 import React, { useRef, useState } from "react";
-import { useForm } from "react-hook-form";
-import { useHistory, useLocation } from "react-router-dom";
-import { searchRoute } from "routes";
 
-export default function SearchBox({ className }: { className?: string }) {
+export default function SearchBox({
```
darrenvong, line 10: Are there any reasons why you didn't write tests at this level? I think it would be good to have some and check that the filters get updated correctly. And maybe one to check you can open the filter dialog after clicking the button if it's not too fiddly.

app/frontend/src/features/search/LocationAutocomplete/LocationAutocomplete.test.tsx
darrenvong: Got a straggler to update, considering you've done it in the other tests :p

app/frontend/src/features/search/SearchResultsList.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+        selectedResult={1}
+        handleResultClick={mockHandleResultClick}
+        map={mockMapRef}
+        searchFilters={mockSearchFiltersFactory()}
+      />,
+      { wrapper: hookWrapper }
+    );
+
+    expect(screen.getByRole("progressbar")).toBeVisible();
+    expect(await screen.findByRole("heading", { name: users[0].name }));
```
darrenvong:
```suggestion
    expect(await screen.findByRole("heading", { name: users[0].name })).toBeVisible();
```

app/frontend/src/features/search/SearchResultsList.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  it("Is blank with no selection or query", () => {
+    render(
+      <SearchResultsList
+        selectedResult={undefined}
+        handleResultClick={mockHandleResultClick}
+        map={mockMapRef}
+        searchFilters={mockSearchFiltersFactory()}
+      />,
+      { wrapper: hookWrapper }
```
darrenvong:
```suggestion
      { wrapper }
```
If you import the `hookWrapper` named `wrapper` instead. Same comment for the rest rather than repeating.

app/frontend/src/features/search/SearchResultsList.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+          handleResultClick={mockHandleResultClick}
+          map={mockMapRef}
+          searchFilters={mockSearchFiltersFactory({ query: "test query" })}
+        />,
+        { wrapper: hookWrapper }
+      );
+    });
+    it("displays the result", async () => {
+      expect(screen.getByRole("progressbar")).toBeVisible();
+      expect(await screen.findByRole("heading", { name: users[0].name }));
```
darrenvong:
```suggestion
      expect(await screen.findByRole("heading", { name: users[0].name })).toBeVisible();
```
Or maybe even remove the `expect(...)` around it. I guess it's ok if it's only used as a handle to wait for something first before proceeding with the next assertion.

app/frontend/src/features/search/useSearchFilters.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const history = useHistory();
+
+  useEffect(() => {
+    const entries: [string, string][] = [];
+    Object.entries(active).forEach((entry) => {
+      if (typeof entry[1] === "object") {
+        Object.values(entry[1]).forEach((value) => {
+          entries.push([entry[0], `${value}`]);
+        });
```
darrenvong:
```suggestion
    Object.entries(active).forEach(([key, filterValues]) => {
      if (Array.isArray(filterValues)) {
        filterValues.forEach((value) => {
          entries.push([key, `${value}`]);
        });
```
Based off the type, looks like the only thing that's an "object" is `hostingStatusOptions` so the Array check is more accurate and also means you don't have to wrap it around `Object.values`

app/frontend/src/features/search/useSearchFilters.ts
```diff
@@ -0,0 +1,94 @@
...
+  }, []);
+
+  return { active, change, remove, apply };
+}
+
+export function locationToFilters(location: Location) {
+  const searchParams = new URLSearchParams(location.search);
+  const filters: SearchFilters = {};
+  Array.from(searchParams.keys()).forEach((untypedkey) => {
+    const key = untypedkey as keyof SearchFilters;
```
darrenvong:
```suggestion
  Array.from(searchParams.keys()).forEach((key) => {
```
Since you have exhaustively handled all cases anyway with the fallback to a console warn if the type assertion there turns out to be false, is there any point in doing it anyway?

darrenvong (changes requested):
Couple more minor comments, but mostly looking good! Sorry for completely forgetting to finish it off and let it sat there for a while 😅

app/frontend/src/features/search/useSearchFilters.ts
lucaslcode: I do want to keep it for the autocomplete, but users can also type anything in the url so I feel I should handle default cases.

darrenvong: Like you said, since the user can type anything, then `untypedkey`/`key` having the `string` type is correct and so it'd be wrong to type assert here then. It's not needed for the assignment in the `case`s later since the cases themselves seems to correctly narrow the type of `key` to be just that string, which matches `keyof SearchFilters` and so no errors even without the type assert.

darrenvong (approved):

PR #1296, opened 2021-05-25
Add paragraph about events on dashboard

aapeliv:


lucaslcode (approved):

lucaslcode:
Note, I suspect no one will read this text unless it grabs attention somehow

lucaslcode:
Though NEW is big, maybe that's enough

aapeliv:
Yeah the dashboard probably needs to be reorganized so that the information is clear and not like a series of "more attention grabbing" stuff

PR #1299, opened 2021-05-25
Feature/prometheus poc

lennartvanlaake:
Minor POC on how we could use prometheus metrics. Prometheus can scrape these metrics from port 8000 - you can check the result when running the application locally.

Not sure whether the two basic counters I made make sense, that is why it is a draft and a POC :)

**Backend checklist**
- [x ] Formatted my code by running `isort . && black .` in `app/backend`
- [ ] All tests pass
- [x ] Run the backend locally and it works


aapeliv (commented):
This looks very cool! It'd be awesome to have something like this integrated. I also had a (very) brief look at the prometheus docs.

A few general questions:

1. What's required to scrape all of this from the backend? Do we need a prometheus container sitting alongside this? I guess we'd add grafana that would then grab stats from prometheus?
2. Where are the stats stored long-term (what kind of db? etc)?
3. What's the storage (long term), memory (short term), and processing cost per metric? There's a lot of metrics I could think of that would be interesting (e.g. bytes in/out per user or handling time per user per method, which have very high cardinality). I know you mentioned that the counters themselves can be implemented efficiently (1 byte per counter?), but any idea on how it is on the python world we live in? Is there sparse storage (e.g. we don't want to keep track of each user * method if they don't use those methods)?

Python questions:

1. How does the metric exposing server (`start_http_server`) work? Does it run in a separate thread?
2. Should we have another file, `src/couchers/stats.py` or similar with all the counters/histograms/etc?
3. It seems there are some problems with multiprocessing (see `//app/backend/src/couchers/jobs/worker.py`), does that cause issues.

lennartvanlaake:
@aapeliv 

Thanks!

1. It would be an option to run our own stack (at a minimum a grafana container and a prometheus container). However, I would suggest having a look at the free tier of grafana cloud (https://grafana.com/products/cloud/pricing/). You would still need a prometheus container that would forward dat to them, but grafana would do the actual storage of data and dashboards. Their product also does tracing and infra monitoring, so it sounds like a pretty nice solution for having one place where we ship all our metrics and logs.
2. The stats are stored in Prometheus itself, it is a scraper and database in one. If we go for grafana cloud we wouldn't have to worry about storage, if we do it ourself the storage is pretty efficient. I have hosted it quite a few times with hundreds of metrics and never had any issues with storage space.
3. Prometheus is mostly used for short-term monitoring. Usually you would have a retention period after which you scrap your data. If an incident happened a year ago it is probably not necessary to look at all the details and graphs anymore. You can store a massive amount of time series in 1 GB, so if we would be hosting it ourselves costs would remain pretty low unless we go absolutely crazy on the things we monitor.
Memory-wize the thing being scraped is just a dict with strings as keys and numbers as values. So roughly (number of series * 1 string + 1 number) would give you the memory required.
 
You want to have separate series for things that translate to a graph. I think this would be the wrong tool to do anything on a per-user basis. Having a line-graph where each line is a user would look terrible ;)

Counters per job and histograms per servicer (see https://prometheus.io/docs/practices/histograms/) for response times would be examples of things that I think are sensible to measure, since they translate user experience. If you would do count successful/failed jobs + histograms of servicer timings you would still only get several hundred series, which would be minimal in terms of memory

Python stuff:

1. Yes, it runs in a deamon thread.
2. That sounds sensible, I will update the PR
3. You are right, multiprocessing does not work with a 'regular' setup. There are options to work around this by writing metrics to files and having different threads look those metrics up, but I think for now it would make the most sense to not give forked processes access to counters, since that setup would add a significant amount of complexity. I will refactor this code to ensure the main process increases the counter, not the forked process.

aapeliv:
>     1. It would be an option to run our own stack

Yes, this is what I meant; what does it take to host it ourselves? I'm happy to use a hosted solution at the start when trying it out, but in the medium/long term want to run+host it ourselves.

>     2. The stats are stored in Prometheus itself, it is a scraper and database in one. If we go for grafana cloud we wouldn't have to worry about storage, if we do it ourself the storage is pretty efficient. I have hosted it quite a few times with hundreds of metrics and never had any issues with storage space.

Where is the database stored though? Space shouldn't be an issue but just need to properly manage the data. Currently the backend container has no persistent storage, it's stateless in the sense that it just interfaces with the SoT postgres DB. So if prometheus expects it to have some persistence, that would complicate things: it seems to me this is fine in the sense that another service with persistence will scrape the metrics regularly. I guess it's stored in grafana then?

> Prometheus is mostly used for short-term monitoring. Usually you would have a retention period after which you scrap your data. If an incident happened a year ago it is probably not necessary to look at all the details and graphs anymore. You can store a massive amount of time series in 1 GB, so if we would be hosting it ourselves costs would remain pretty low unless we go absolutely crazy on the things we monitor.

Hmmmm, okay, that makes sense!

> I think this would be the wrong tool to do anything on a per-user basis.

I was just thinking that it'd be useful to (be able to) segment by user: that way you can see how e.g. response time grows as the "user" grows (more friends/requests/messages) or similar? It might help in figuring out why stuff is slow and diagnosing choke points. But maybe I'm thinking of this wrong.

>     3. You are right, multiprocessing does not work with a 'regular' setup.

I think we only have one (or two?) extra processes that do background jobs: maybe they should run their own copy on 8081/8082 and be scraped from there?

lennartvanlaake:
> Yes, this is what I meant; what does it take to host it ourselves? I'm happy to use a hosted solution at the start when trying it out, but in the medium/long term want to run+host it ourselves.
> 
> >

For a self-hosted version we would need to have two additional containers coupled with persistent storage. You would need a prometheus container that does scraping and stores the metrics, and a grafana container which would contain contain the dashboards.

My suggestion for the longer term would be to go toward kubernetes. Using kubernetes would simplfy hosting our own sidecar applications for things like monitoring and would make it a lot easier to say: prometheus, go monitor all the containers that we have.

Our API does not need to be stateful for it to export prometheus metrics: it keeps the metrics in memory and then you discount all decreases of metrics in your prometheus queries. For example: say you have a counter with value 40 that increases by 1 every minute that resets to 0 after a restart of your application. You would then have a query that only looks at the increase, so it would only see the increase going from 39 to 40 and next minute going from 0 to 1, showing a consistent increase of 1 per minute.

> I was just thinking that it'd be useful to (be able to) segment by user: that way you can see how e.g. response time grows as the "user" grows (more friends/requests/messages) or similar? It might help in figuring out why stuff is slow and diagnosing choke points. But maybe I'm thinking of this wrong.

To run with that example: If you want to do know if a 'large' user would get a high response time and when that would become problematic, I would do create a user locally with a stupid amount of friends/request/messages and measure when the response times become unreasonable. If you want to know if this happens in production, I would use response time percentile-buckets. Say 1% of users has a crazy amount of friends/messages and want to know if for them the application still works well you can check if for the 99th percentile the response times are still reasonable.

> I think we only have one (or two?) extra processes that do background jobs: maybe they should run their own copy on 8081/8082 and be scraped from there?

My mistake was trying to increase the counter from inside the forked job. When looking at the code again I think it is perfecly possible to increase the counter from the main process and get potential errors when the forked process is joined again. That way you keep the counter and the prometheus client in the same process and you don't need to worry about multiprocessing stuff.

aapeliv:
Good points.

> My suggestion for the longer term would be to go toward kubernetes.

Yeah, would love to go for k8s at some point, but it's a bit of a pain to manage and get set up properly. Eventually though... maybe for now we can still live with this!

> To run with that example: If you want to do know if a 'large' user would get a high response time and when that would become problematic, I would do create a user locally with a stupid amount of friends/request/messages and measure when the response times become unreasonable. If you want to know if this happens in production, I would use response time percentile-buckets. Say 1% of users has a crazy amount of friends/messages and want to know if for them the application still works well you can check if for the 99th percentile the response times are still reasonable.

My thinking was that getting metrics per user_id allows you to do the splitting afterwards to figure out whether the culprit was number of friends or number of messages, etc.

It seems to me that these metrics have to be thought about beforehand; which I imagine is great for general monitoring once you've set up the monitoring (e.g. just general per-method response times), but that's not very useful for debugging or trying to figure out rare performance regressions?

Let's discuss more in the meeting in a bit, if you can make it!

lennartvanlaake:
Dicussed this with @aapeliv during the meeting yesterday, we'll go for per request/job timings and errors for the scope of this PR. If we like how it all works and have the need for it we can definitely add more metrics and things.

On kubernetes: I recently set up a cluster to host my travel blog on (I know, a little bit overkill) and it was pretty easy to do. With relevatively little knowledge it took me about 2 evenings to have it running, and it feels so much nicer now. For reference you can check out the `kube` dir in my repo https://github.com/lennartvanlaake/bigbikeride for all the config needed to have a (minimal) cluster running.  If you already do everything in Docker and have a reasonably complex setup (you have a few containers, want to have some monitoring, a dev environment, make sure the application keeps on running, have a couple of secrets, do backups for a couple of stateful things, maybe self-host some things like Sentry, etc) I think the benefits outweigh the costs of "manually" doing all that well.

I'd be happy to make a POC-environment when we have a period someday in which we don't have a lot of high-priority things to do :)

app/backend/src/couchers/interceptors.py
```diff
@@ -123,35 +127,45 @@ def _store_log(self, method, status_code, duration, user_id, request, response,
...
-        logger.debug(f"{user_id=}, {method=}, {duration=} ms")
+        logger.debug(f"{user_id=}, {method=}, {grpc_details=}, {duration=} ms")
 
     def intercept_service(self, continuation, handler_call_details):
         handler = continuation(handler_call_details)
         prev_func = handler.unary_unary
         method = handler_call_details.method
 
         def tracing_function(request, context):
+            user_id = getattr(context, "user_id", None)
```
aapeliv, line 142: This needs to happen *after* the call; as the auth interceptor is after this one... `context` is mutated in `prev_func`.

app/backend/src/couchers/interceptors.py
```diff
@@ -123,35 +127,45 @@ def _store_log(self, method, status_code, duration, user_id, request, response,
...
+                    if code_object != None:
+                        code = code_object.name
+                    binary_details = context._state.details
+                    if binary_details != None:
+                        details = binary_details.decode('utf-8')
+                traceback = "".join(format_exception(exception_type, e, e.__traceback__))
+                exception = e
+                exception_type = type(e)
+            finally:
                 finished = perf_counter_ns()
```
aapeliv, line 160: I had the perf counter end asap to try to only time the function, not exception handling, etc

lennartvanlaake:
Made a new brach taking into account given comments and renamed branch to match naming conventions.

PR #1302, opened 2021-05-26
Hosting request emails

### LINKED ISSUE #915: Email notifications for conversations in host requests

aapeliv:
Implements emails for when hosting requests change state and when they get messages.

Closes #915, but more of a stopgap before we get a full push notification system built.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -117,6 +117,8 @@ class User(Base):
 
     # id of the last message that they received a notification about
     last_notified_message_id = Column(BigInteger, nullable=False, default=0)
+    # same as above for host requests
+    last_notified_request_message_id = Column(BigInteger, nullable=False, server_default=text("0"))
 
     # display name
     name = Column(String, nullable=False)
@@ -1377,6 +1379,8 @@ class BackgroundJobType(enum.Enum):
     send_onboarding_emails = 5
     # payload: google.protobuf.Empty
     add_users_to_email_list = 6
+    # payload: google.protobuf.Empty
+    send_request_notifications = 7
 
 
 class BackgroundJobState(enum.Enum):
--- a/app/backend/src/couchers/urls.py
+++ b/app/backend/src/couchers/urls.py
@@ -30,8 +30,12 @@ def password_reset_link(password_reset_token):
     return f"{config['BASE_URL']}/password-reset/{password_reset_token}"
 
 
-def host_request_link():
+def host_request_link_host():
     return f"{config['BASE_URL']}/messages/hosting/"
+
+
+def host_request_link_guest():
+    return f"{config['BASE_URL']}/messages/surfing/"
 
 
 def messages_link():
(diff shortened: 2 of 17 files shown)


app/backend/src/couchers/models.py
```diff
@@ -117,6 +117,8 @@ class User(Base):
 
     # id of the last message that they received a notification about
     last_notified_message_id = Column(BigInteger, nullable=False, default=0)
+    # same as above for host requests
+    last_notified_request_message_id = Column(BigInteger, nullable=False, server_default=text("0"))
```
kalvdans, line 121:
```suggestion
    last_notified_request_message_id = Column(BigInteger, nullable=False, server_default=0)
```

app/backend/src/couchers/urls.py
```diff
@@ -30,10 +30,14 @@ def password_reset_link(password_reset_token):
...
 
-def host_request_link():
+def host_request_link_host():
     return f"{config['BASE_URL']}/messages/hosting/"
 
 
+def host_request_link_guest():
+    return f"{config['BASE_URL']}/messages/surfing/"
+
+
```
kalvdans, line 40: Update //docs/urls.md with these urls. Later on we want to have the conversation id in there too I guess.

app/backend/src/tests/test_references.py
```diff
@@ -66,7 +66,7 @@ def create_host_request(
     )
     session.add(host_request)
     session.commit()
-    # send_host_request_email(host_request)
+    # send_new_host_request_email(host_request)
```
kalvdans:
```suggestion
```

kalvdans (commented):
Some small comments, in general it looks good!

app/backend/src/couchers/models.py
aapeliv: SQLAlchemy won't take that... It needs to be an SQL text fragment! It's pretty dumb... I know...

kalvdans: I saw 0 on the line above, but that's not a server_default. Are you saying we should change all default to server_default?

kalvdans: LOL, "there's an [app](https://sqlalchemy-defaults.readthedocs.io/en/latest/) for that"

aapeliv: > LOL, "there's an [app](https://sqlalchemy-defaults.readthedocs.io/en/latest/) for that"

🤦

aapeliv: > I saw 0 on the line above, but that's not a server_default. Are you saying we should change all default to server_default?

Yeah, we probably should... Depending on it, `default` is python side, `server_default` is postgres side.

Probably most should be `server_default`s, but it's just, yeah, this text stuff is silly

kalvdans (approved):
There are no test for this, but all email formatting code is run by the unit tests I presume, and added to the outgoing queue. (Only the worker process is affected by the ENABLE_EMAIL config variable)

So the code should be free from syntax errors at least, LGTM!

PR #1306, opened 2021-05-26
Fix references bug

aapeliv:
If user 1 surfed with user 2 and either wrote a reference, then the other couldn't. 

**Backend checklist**
- [x] All tests pass


kalvdans (approved):
Looks good with tests and everything! Add a closes: tag before merge if it was reported there.

PR #1310, opened 2021-05-28
Frontend/feature/profile route

### LINKED ISSUE #1272: Separate profile route from user routes
Follow up issue for #1200 

We now have one /user route with optional `username` parameter.

If no username is provided we display the profile for the current logged in user.

This causes difficulties in our code. Suggestion is to use two separate routes:

/profile/:tab?
/user/:username/:tab?

Note: don't forget to edit the `urls.py` and `urls.MD` accordingly

thdk:
Separate profile route (view and edit your own profile) from user route (view other users profile)

This is a follow up for #1200 and simplifies our code on some places.

Closes #1272 

**Frontend checklist**
- [x] All tests pass


app/frontend/src/AppRoutes.tsx
```diff
@@ -136,12 +138,15 @@ export default function AppRoutes() {
       {
         // PROFILE
       }
-      <AppRoute isPrivate path={editUserRoute}>
+      <AppRoute isPrivate path={editProfileRoute}>
         <EditProfilePage />
       </AppRoute>
-      <AppRoute variant="full-width" isPrivate path={userRoute}>
+      <AppRoute isPrivate path={profileRoute}>
```
darrenvong:
```suggestion
      <AppRoute variant="full-width" isPrivate path={profileRoute}>
```
So it's consistent with the user page - it's mainly there so the cards stretches the full width of the screen when on breakpoints below `md`.

app/frontend/src/features/user/UserCard.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onTabChange,
+}: {
+  user: User.AsObject;
+  top?: ReactNode;
+  onTabChange: (tab: UserTab) => void;
+}) {
+  const classes = useStyles();
+  let { tab = "about" } =
+    useParams<{
+      tab: UserTab;
```
darrenvong: I guess this is not quite correct and `tab` actually has type `string` at this point. If the user goes to `/profile/some-giberrish` or `/user/aapeli/some-gibberish`, they're gonna land on an empty tab with nothing selected. It'd be good to a check to see if `tab` is really `UserTab` and if not, fallback to the default tab as if `tab` is undefined?

app/frontend/src/features/user/UserCard.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+      table: {
+        disable: true,
+      },
+    },
+    top: {
+      table: {
+        disable: true,
+      },
+    },
+  },
```
darrenvong: What is this `{ table: { disable: true } }` arg type supposed to do? Are they needed since you aren't using `UserCardProps` as the argument type on line 33?

app/frontend/src/features/profile/view/UserPage.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <CircularProgress />
+      ) : user ? (
+        <ProfileUserProvider user={user}>
+          <div className={classes.root}>
+            <Overview user={user} setIsRequesting={setIsRequesting} />
+            <UserCard
+              onTabChange={(newTab) => {
+                history.push(routeToUser(user.username, newTab));
+              }}
+              user={user}
```
darrenvong: Seeing you've shuffled a few things around from working on this, it might be a good opportunity here to remove explicitly passing the `user` around as props and simplify things a bit, since you can access it only in the component that needs the `user` with the `useProfileUser` hook that reads from the context. I made the provider exactly for this purpose for `References` as the prop drilling of `user` in this part of the tree was getting deep and unwieldy.

darrenvong: Although feel free to leave this in a separate PR/noted as a refactor task/issue if you like

darrenvong (changes requested):
Only couple of small comments, but generally looking good 👍

aapeliv:
When I currently go to another user's profile, the "Profile" tab is highlighted as active. I think that bug is related to this PR.

thdk:
> When I currently go to another user's profile, the "Profile" tab is highlighted as active. I think that bug is related to this PR.

Ah yes, I can fix that with this issue

app/frontend/src/features/user/UserCard.stories.tsx
thdk: I've added this so no controls would be automatically generated by storybook for these. I agree it's verbose and their might be better options to accomplish this?

app/frontend/src/features/profile/view/UserPage.tsx
thdk: I'm not sure if I can remove the user prop here as the UserCard component is used both on ProfilePage and on UserPage. So the parent provided the user object here. One page will pick the user by username, the other page will provide the current authenticated user.

app/frontend/src/features/user/UserCard.tsx
thdk: ah that's correct. I've created a new component that wraps the TabContext and adds validation for the tab provided from the url.

app/frontend/src/features/user/UserCard.stories.tsx
thdk: Ah, when I remove the component prop of the default export these props arent generated either. So I can remove the argTypes indeed here! :)

app/frontend/src/features/user/UserTabContext.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </UserTabContext>
+      </Route>
+    </MemoryRouter>
+  );
+}
+
+describe("UserTabContext", () => {
+  it("accepts 'home' as tab parameter from url path", () => {
+    renderTest("profile/home");
+    screen.getByText("home");
```
darrenvong:
```suggestion
    expect(screen.getByText("home")).toBeInTheDocument();
```
Same comment for the others so we get better error messages for when the tests fail in future for whatever reason (e.g. we changed something or broken something unintentionally)

darrenvong (dismissed):
Should be the last minor comment on the tests, looks good to me otherwise!

darrenvong:
Also the renames that had to be done here has broken the `EditProfilePage` tests and broken CI. Do you mind fixing that as well please? 😬  Hopefully should just be updating the variable that test file is referencing...

app/frontend/src/features/user/UserTabContext.test.tsx
thdk: Is that needed? The jest docs write that getByText throws an error with a descriptive error message:
https://testing-library.com/docs/queries/about/#types-of-queries

> getBy...: Returns the matching node for a query, and throw a descriptive error if no elements match or if more than one match is found (use getAllBy instead if more than one element is expected).

thdk: screen.getByText("foobar")
```javascript
 FAIL  src/features/user/UserTabContext.test.tsx
  UserTabContext
    ✓ accepts 'home' as tab parameter from url path (40 ms)
    ✓ accepts 'about' as tab parameter from url path (5 ms)
    ✓ replaces invalid tab parameters from url path with default 'about' (5 ms)
    ✕ uses the default 'about' when there is no tab parameter (9 ms)

  ● UserTabContext › uses the default 'about' when there is no tab parameter

    TestingLibraryElementError: Unable to find an element with the text: foobar. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    <body>
      <div>
        about
      </div>
    </body>

      42 |   it("uses the default 'about' when there is no tab parameter", () => {
      43 |     renderTest("profile");
    > 44 |     screen.getByText("foobar");
         |            ^
      45 |   });
      46 | });
      47 |

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:37:19)
      at node_modules/@testing-library/dom/dist/query-helpers.js:90:38
      at node_modules/@testing-library/dom/dist/query-helpers.js:62:17
      at getByText (node_modules/@testing-library/dom/dist/query-helpers.js:111:19)
      at Object.<anonymous> (src/features/user/UserTabContext.test.tsx:44:12)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 3 passed, 4 total
Snapshots:   0 total
Time:        4.077 s
```

expect(screen.getByText("foobar").toBeInTheDocument())
```javascript
 FAIL  src/features/user/UserTabContext.test.tsx
  UserTabContext
    ✓ accepts 'home' as tab parameter from url path (33 ms)
    ✓ accepts 'about' as tab parameter from url path (5 ms)
    ✓ replaces invalid tab parameters from url path with default 'about' (6 ms)
    ✕ uses the default 'about' when there is no tab parameter (8 ms)

  ● UserTabContext › uses the default 'about' when there is no tab parameter

    TestingLibraryElementError: Unable to find an element with the text: foobar. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    <body>
      <div>
        about
      </div>
    </body>

      42 |   it("uses the default 'about' when there is no tab parameter", () => {
      43 |     renderTest("profile");
    > 44 |     expect(screen.getByText("foobar")).toBeInTheDocument();
         |                   ^
      45 |   });
      46 | });
      47 |

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:37:19)
      at node_modules/@testing-library/dom/dist/query-helpers.js:90:38
      at node_modules/@testing-library/dom/dist/query-helpers.js:62:17
      at getByText (node_modules/@testing-library/dom/dist/query-helpers.js:111:19)
      at Object.<anonymous> (src/features/user/UserTabContext.test.tsx:44:19)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 3 passed, 4 total
Snapshots:   0 total
Time:        3.87 s
```

app/frontend/src/features/profile/view/UserPage.tsx
darrenvong: It should be ok, since the same context can be reused in multiple places. React will know to pick the relevant one in the component tree, so the important thing is both `ProfilePage` and `UserPage` here wrap this part of the UI with the same `Provider` component.

app/frontend/src/features/user/UserTabContext.test.tsx
darrenvong: Ahh interesting, it looks like the error comes from the same place anyway so that's good to know!

In that case, I think it's more of a style nitpick for me, since imo it's unclear on its own that we are meant to be asserting something with that statement as I could also write that on its own to wait for something if I don't care about the result. Putting things in `expect(...).<assertion>` makes it clearer it's a test result we care about. Although in succinct tests like these, it might not make a big difference, but I think it's better to be consistent across all our tests if we can

thdk: I thought I had once read that wrapping it with expect was bad practice. However, [this article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) - although the author agrees it's opiniated - claims the opposite. And since the other remarks of this article are already taken care of very well here in the couchers repo, I'll hapilly add the assertions back.

(Except for the one regarding the discussions test because that one really always fails the test for me :( )

thdk: I thought I had once read that wrapping it with expect was bad practice. However, [this article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) - although the author agrees it's opiniated - claims the opposite. And since the other remarks of this article are already taken care of very well here in the couchers repo, I'll hapilly add the assertions back.

(Except for the one regarding the discussions test because that one really always fails the test for me :( )

darrenvong: It's a good article since React Testing Library is written/started by Kent! The tests in this repo have intentionally followed the medium to high importance tips he's given there as after all, who knows better how to use the library than the library author himself :)

This diff here still says you haven't update these though? 🤔

app/frontend/src/features/profile/view/UserPage.tsx
thdk: Ah I'm an idiot! Of course all that's needed is that the components are somewhere wrapped with the `ProfileUserProvider` and this is already the case on `ProfilePage` and `UserPage`.

thdk: Hopefully I didn't overdo it now  🐰

darrenvong (approved):

darrenvong:
Nice, thanks for fixing the references stuff as well 🎉

lucaslcode:
Hope you don't mind if I merge this!

Issue #1318, opened 2021-05-29
Community tab


lucaslcode:
Until we have the dashboard, a simple community tab with links to your current community and a list of all communities

aapeliv:
I'm modifying `ListCommunities` to return a list of all communities if not given a parent community ID.

aapeliv:
> links to your current community

A user can have multiple communities, so this should be a list as well. Sorry, that's my bad for saying otherwise earlier.

thdk:
This tab would need to be added to the user/profile page?

aapeliv:
No, just as a tab in the navbar, next to "Dashboard", "messages", etc

thdk:
Just one more question to make this clear for new contributors:

So in the nav bar we get something like: (?)

```
Dashboard
Messages
Communities
- my current community
- my other community 1
- my other community b
Profile
```

aapeliv:
I think Communities should just be a "communities landing page", and the page content is those lists so they're not in the navbar.

It's just a quick and dirty way to discover communities for now: we'll definitely change it as we go, but we need some page to just list all the communities so people can discover them and get to their pages.

thdk:
Ok, I can have a look at this one then. I'll not assign myself yet until I effectively start working on this issue.

So if anyone is faster, feel free! :)

lucaslcode:
Sorry I started before I saw your comment and also I didn't do what Aapeli said :P

PR #1319, opened 2021-05-29
Implement listing of all communities, not just given a parent

aapeliv:
In preparation for #1318.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/communities.py
+++ b/app/backend/src/couchers/servicers/communities.py
@@ -100,7 +100,7 @@ class Communities(communities_pb2_grpc.CommunitiesServicer):
             next_node_id = int(request.page_token) if request.page_token else 0
             nodes = (
                 session.query(Node)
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/servicers/communities.py
```diff
@@ -100,7 +100,7 @@ def ListCommunities(self, request, context):
             next_node_id = int(request.page_token) if request.page_token else 0
             nodes = (
                 session.query(Node)
-                .filter(Node.parent_node_id == request.community_id)
+                .filter(or_(Node.parent_node_id == request.community_id, request.community_id == 0))
```
kalvdans, line 103: `(request.community_id == 0) or (Node.parent_node_id == request.community_id)` should work fine as well since the or can be evalated before the query takes place.

kalvdans (approved):
Looks good with or without my small note.

PR #1320, opened 2021-05-29
NotificationBadge added to navbar

### LINKED ISSUE #942: there's no green circle with a number to indicate new messages on the top navbar
Subject: there's no green circle with a number to indicate new messages on the top navbar
Description:
The messages page has a green circle with a number next to each chat to show when there is new messages (and how many). There should be a green circle by "messages" on the top navbar (between "events" and "map") so people can see they have new messages even when they are on other pages.

Steps:
Visited the website

Results:
There should be a green circle by "messages" on the top navbar (between "events" and "map") whenever people get a new message, so people can see they have new messages even when they are on other pages of the website.

Backend version: develop-a6cdeee4
Frontend version: develop-a6cdeee4
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:87.0) Gecko/20100101 Firefox/87.0
Page: https://app.couchers.org/messages
User ID: 57

vrouilhac:
**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass


vrouilhac:
I added a notification badge so when we receive a message we can see it on the navigation bar even if we are on other tabs than messages. The issue though is that on desktop the badge override the right button as you can see with the pictures. I can add a prop to NotificationBadge as optional to place it a little upper to solve this ?

vrouilhac:
I think reducing the size of the badge on desktop should be a good idea here's what I get 

 For this I added a prop to NavigationBadge named 'up' optional and boolean that is set to true for desktop version

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -10,8 +10,10 @@ import {
 import classNames from "classnames";
 import { CloseIcon, MenuIcon } from "components/Icons";
 import ExternalNavButton from "components/Navigation/ExternalNavButton";
+import NotificationBadge from "components/NotificationBadge";
 import { useAuthContext } from "features/auth/AuthProvider";
 import useAuthStyles from "features/auth/useAuthStyles";
+import useNotifications from "features/useNotifications";
```
aapeliv: I think you forgot to add these to git!

aapeliv (commented):
I think you forgot to add a few files to git; also there's an extra `.Navigation.tsx.un~` file that shouldn't be added!

vrouilhac:
> I think you forgot to add a few files to git; also there's an extra `.Navigation.tsx.un~` file that shouldn't be added!

I'll remove those

app/frontend/src/components/Navigation/Navigation.tsx
vrouilhac: You mean those lines wasn't added to the commit ? @aapeliv

aapeliv: Sorry I'm mistaken here, please ignore the comment! I didn't realize these files were already in the tree, I though they were files you added.

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -115,13 +117,22 @@ export default function Navigation() {
   const authClasses = useAuthStyles();
   const authenticated = useAuthContext().authState.authenticated;
   const [open, setOpen] = React.useState(false);
+  const { data } = useNotifications();
+    
+  let notificationNumber = data?.unseenMessageCount;
```
lucaslcode: Make this `const` and also make it the sum of the unread messages and also host requests, since they are under this page too.

lucaslcode (changes requested):
In `NotificationBade.tsx`, the styling only really works in the place where it was used. Try changing it to this:
```
  badge: {
    right: "-0.5rem", //changed from "theme.spacing(3)" because it's too far for the general case, and also it isn't related to spacing, just offsetting a specific amount that isn't spacing-related, so we should use rem directly
    //delete: top: "50%",
  },
}));
```

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -115,13 +117,22 @@ export default function Navigation() {
...
         {menu.map(({ name, route }) => (
           <ListItem button key={name}>
-            <NavButton route={route} label={name} labelVariant="h2" />
+            {route === messagesRoute ? (
+              <NotificationBadge count={notificationNumber}>
+                <NavButton route={route} label={name} labelVariant="h2" />
+              </NotificationBadge>
+            ) : (
+              <NavButton route={route} label={name} labelVariant="h2" />
+            )}
```
lucaslcode: This doesn't seem ideal but I can't think of a better way. Later we will have a dedicated notifications dropdown like facebook so I think it's fine for now.

vrouilhac: Dedicated notifications seems a good idea, will the message tab be remaining in the navbar then?

vrouilhac:
> In `NotificationBade.tsx`, the styling only really works in the place where it was used. Try changing it to this:
> 
> ```
> const useStyles = makeStyles((theme) => ({
>   badge: {
>     right: "-0.5rem", //changed from "theme.spacing(3)" because it's too far for the general case, and also it isn't related to spacing, just offsetting a specific amount that isn't spacing-related, so we should use rem directly
>     //delete: top: "50%",
>   },
> }));
> ```

This still seem to be on top of search button, because of the padding of the message NavButton i guess

vrouilhac:
Here is the result 
-0.7rem : 


-0.5rem :

lucaslcode:
Hmm. How about this. In `Navigation.tsx`, change the menu definition at the top to be:
```
const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
  ...
  {
    name: "Messages",
    route: messagesRoute,
    notificationCount: data.whatever,
  },
  ...
```

Then you can do `menu(data).map(...`, and then pass `notificationCount` to `NavButton`, and put the badge in there (it's optional) somewhere that can avoid the padding.

Does that make sense? That will also solve the part I said isn't ideal.

PS. remember to run `yarn lint --fix` to sort the imports, or the CI won't build the preview

vrouilhac:
> Hmm. How about this. In `Navigation.tsx`, change the menu definition at the top to be:
> 
> ```
> const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
>   ...
>   {
>     name: "Messages",
>     route: messagesRoute,
>     notificationCount: data.whatever,
>   },
>   ...
> ```
> 
> Then you can do `menu(data).map(...`, and then pass `notificationCount` to `NavButton`, and put the badge in there (it's optional) somewhere that can avoid the padding.
> 
> Does that make sense? That will also solve the part I said isn't ideal.
> 
> PS. remember to run `yarn lint --fix` to sort the imports, or the CI won't build the preview

Yeah look definitely much better but it think we should add right spacing to the message button. 



( I also added restructuring because we were using .map((item) .., changed to .map(({ route, name, notificationCount })... )

vrouilhac:
I tried to add marginLeft to the label ( inside useNavLinkStyle ) which move a bit to the left and help keep the badge tied to the message button, tell me if this is a good way to solve this I'm not sure about the overall impact.


 
PS : the marginLeft is set to 0.5rem

lucaslcode:
How does it look adding that margin to both sides, and only if the notification badge is shown (if it makes a difference)?

vrouilhac:
I don't know if this is a good way of doing so but I conditionally added spacing ( `className={[class1, condition ? class2 : null].join(' ')}` ), it doesn't change that much from the previous version but unnecessary space was removed.

En result :


PS : `class2` contains margin only if the badge is present ( notificationCount exists )

app/frontend/src/components/Navigation/NavButton.tsx
```diff
@@ -8,24 +9,31 @@ interface NavButtonProps {
...
   return (
     <NavLink
       activeClassName={classes.selected}
       exact={route === baseRoute || route === userRoute}
       to={{ pathname: route }}
-      className={classes.link}
+      className={[
+        classes.link,
+        notificationCount ? classes.notification : null,
+      ].join(" ")}
```
lucaslcode:
```suggestion
      className={classNames(classes.link, [classes.notification]: !!notificationCount},
```
And import `classNames`

lucaslcode (changes requested):
Looks good. See below for correct way to do multiple/conditional class names

app/frontend/src/components/Navigation/NavButton.tsx
vrouilhac: I did not know that classnames was installed, I added it to the last commit. Also I think I get what !!value do but I struggle to understand it's deepest implementation, I mean how it really works.

lucaslcode: It's the same as `Boolean(value)`, converts to a Boolean. 

If value is truthy (for a number, not zero), !value gives false, so !!value gives true. And the reverse for a falsey value.

lucaslcode (approved):

app/frontend/src/components/Navigation/NavButton.tsx
lucaslcode: You don't need to do this conversion in `if (value)`, but in some places, you have to have either true or false, not just truthy or falsey

PR #1324, opened 2021-05-30
Signup flow v2

aapeliv:
This revamps the signup flow. Rationale:

1. This allows users to fill in all the forms without having to check their email first: this is nice because email can take some time and people get distracted very easily! So users now first fill in all the pages, then are asked to finally confirm their email.
1. This adds the (optional) volunteering form into the signup flow, encouraging more contributor signups
1. This stores information about incomplete signups in the database so we can then encourage users to fill in the rest of their profile and complete their signup (in a later PR).
1. Preparations for invitations feature and signup attribution (where people came from).

To be done in future PRs:

1. Replace current ContributorForm (and backend) with new form and get rid of the Lambda function completely (currently there's a bit of cruft leftover and duplication in this PR)
1. Invitations
1. Remind people to continue signing up if they only got halfway through
1. Signup attribution (UTM, referrer, aka where people came from, so we can improve our marketing)

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [ ] All tests pass


kalvdans:
Why can't the frontend keep track of all the details until the required parts are finished? (That's basically what will happen anyway if we encode everything in the flow_token)

aapeliv:
> Why can't the frontend keep track of all the details until the required parts are finished?

A few reasons. One of the main ones is that it allows this to be asynchronous: we can send the email verification token before the user has submitted the full form; which alleviates problems with slow email delivery. It also allows the form to be validated incrementally by the backend (which simplified logic in frontends). And finally also allows us to encourage users to sign up who only signed up halfway.

> That's basically what will happen anyway if we encode everything in the flow_token

The flow token is just an identifier for the backend object, I'm not sure what you mean by encoding something in the token?

aapeliv:
This is kinda ready for frontend review. I'm not sure if the frontend is complete crap, also I'm having trouble with tests as I was explaining on Slack. There's one failing test that I couldn't make work in a sensible way.

kalvdans:
> The flow token is just an identifier for the backend object, I'm not sure what you mean by encoding something in the token?

I was thinking about concatenating the partial profile data into a string and pass it back to the frontend, without storing anything in the backend.

aapeliv:
> > The flow token is just an identifier for the backend object, I'm not sure what you mean by encoding something in the token?
> 
> I was thinking about concatenating the partial profile data into a string and pass it back to the frontend, without storing anything in the backend.

Why not just store it in the backend? It's much simpler and allows us to do data migrations easily.

It will also really help to have visibility and the ability to prod people to sign up who only got halfway through the process or got distracted while checking their email, etc.

kalvdans:
> > > The flow token is just an identifier for the backend object, I'm not sure what you mean by encoding something in the token?
> > 
> > I was thinking about concatenating the partial profile data into a string and pass it back to the frontend, without storing anything in the backend.
> 
> Why not just store it in the backend? It's much simpler and allows us to do data migrations easily.

Some people might put unnecessary data in the database just to enjoy doing migrations. I'm not into self-inflicting pain but I can only speak for myself ;)

> It will also really help to have visibility and the ability to prod people to sign up who only got halfway through the process or got distracted while checking their email, etc.

That's still visible in the email log. Email without existing user => someone who gave up halfway.

aapeliv:
I am not sure what you're objecting to @kalvdans?

If we're going to allow the "asynchronous" signup of 1. enter email, 2. fill in and submit form, 3. confirm email; then we need to store the results from step 2 in the database in some form. It's very likely a user starts filling in 1 and 2 on their desktop and clicks the email link on their phone, so we need to store that data in the backend.

kalvdans:
> If we're going to allow the "asynchronous" signup of 1. enter email, 2. fill in and submit form, 3. confirm email; then we need to store the results from step 2 in the database in some form. It's very likely a user starts filling in 1 and 2 on their desktop and clicks the email link on their phone, so we need to store that data in the backend.

Can't object with that. I first thought step 1 and 2 were simultaneous but now I read through all :) How about putting them in the real database directly and treating the email verification like just like phone verification? (and allow you to log in using password if you didn't provide an email address)

aapeliv:
> How about putting them in the real database directly and treating the email verification like just like phone verification?

It would have its upsides; but I think the downsides are bigger. We need to validate users somehow, and we need to have some way of getting in touch with them and delivering notifications. Otherwise we'd have to check basically everywhere if the user was fully signed up or not, which is too complex.

kalvdans:
> > How about putting them in the real database directly and treating the email verification like just like phone verification?
> 
> It would have its upsides; but I think the downsides are bigger. We need to validate users somehow, and we need to have some way of getting in touch with them and delivering notifications. Otherwise we'd have to check basically everywhere if the user was fully signed up or not, which is too complex.

They could be jailed initially if their profile is not fully set up. And getting in touch could be done by logging in and checking if you have any pending messages. (rss feed?)

aapeliv:
>  How about putting them in the real database directly and treating the email verification like just like phone verification?

Maybe I misunderstood what you meant here. I thought you just meant we'd use the same table to store incomplete users as well as complete users, but I see now you meant that email verification would not be mandatory for a full account.

The answer is that email verification will be mandatory for using the platform.

kalvdans:
> > How about putting them in the real database directly and treating the email verification like just like phone verification?
> 
> Maybe I misunderstood what you meant here. I thought you just meant we'd use the same table to store incomplete users as well as complete users, but I see now you meant that email verification would not be mandatory for a full account.

Sorry to be unclear. I meant to store incomplete users in the "users" table, the same way as we do with users that has not filled out their home location or not accepted the latest terms.

> The answer is that email verification will be mandatory for using the platform.

This is orthogonal to where to store the incomplete users but thanks for the clarification.

aapeliv:
> Sorry to be unclear. I meant to store incomplete users in the "users" table, the same way as we do with users that has not filled out their home location or not accepted the latest terms.

Yeah, but they're different. Home location and not accepted latest ToS are still full users and should be visible on e.g. the map. Jailed means full user but requires some more action or has notices to ack before signing in.

This would create yet another class of users that should be completely ignored by the vast majority of backend functionality, and we'd probably have to call it something else too. I think it's easier technically to just separate them into different database tables and deal with the other issues (uniqueness of emails and usernames being prime) in other ways, I think.

kalvdans:
> I think it's easier technically to just separate them into different database tables

Yeah, thanks a lot for the motivation, I'll go on commenting the code :)

app/pb/auth.proto
```diff
@@ -80,53 +73,69 @@ service Auth {
...
-  SignupStep next_step = 1;
+  // email token is reused for both verifying a signup's email, but also for continuing a signup
+  string email_token = 5 [(sensitive) = true];
 }
 
-message UsernameValidReq {
-  string username = 1;
+message SignupFlowRes {
+  // a token used to link the user's information across the signup flow
+  string flow_token = 1 [(sensitive) = true];
```
kalvdans: Use a cookie instead to prevent cross-scripting.

app/pb/auth.proto
```diff
@@ -80,53 +73,69 @@ service Auth {
...
+  // signup is fully successful, can go to app now
+  bool success = 2;
+  // this is included iff success is true
+  AuthRes auth_res = 3;
```
kalvdans: Remove `success` as it is redunant.

app/pb/account.proto
```diff
@@ -112,3 +112,22 @@ message VerifyPhoneReq {
...
 }
+
+enum ContributeOption {
+  CONTRIBUTE_OPTION_UNSPECIFIED = 0;
+  CONTRIBUTE_OPTION_YES = 1;
+  CONTRIBUTE_OPTION_MAYBE = 2;
+  CONTRIBUTE_OPTION_NO = 3;
+}
+
+message ContributorForm {
```
kalvdans: Move to auth.proto where it is used.

app/pb/account.proto
```diff
@@ -112,3 +112,22 @@ message VerifyPhoneReq {
...
+  // all fields are optional
+  string ideas = 1;
+  string features = 2;
+  string experience = 3;
+
+  // if contribute is set to yes or no, the frontend should make the ones after this required
```
kalvdans:
```suggestion
  // if contribute is set to yes or maybe, then contribute, contribute_ways and expertise are required.
```

app/pb/auth.proto
```diff
@@ -80,53 +73,69 @@ service Auth {
...
+  bool accept_tos = 10;
+}
 
-  string name = 3;
-  string birthdate = 5; // in YYYY-MM-DD format
-  string gender = 6;
-  org.couchers.api.core.HostingStatus hosting_status = 7;
+message SignupFeedback {
+  org.couchers.api.account.ContributorForm form = 1;
+}
```
kalvdans: Unnecessary indirection, just use ContributorForm instead of this message.

app/pb/auth.proto
```diff
@@ -35,22 +28,22 @@ service Auth {
...
-    // returns info about a signup token
-  }
-
-  rpc CompleteSignup(CompleteSignupReq) returns (AuthRes) {
-    // Complete the signup form
-  }
-
   rpc Login(LoginReq) returns (LoginRes) {
     // First step of login flow
   }
```
kalvdans: Document what is returned if the username is invalid?

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+    __tablename__ = "signup_flows"
+
+    id = Column(BigInteger, primary_key=True)
+
+    # housekeeping
     created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
-    expiry = Column(DateTime(timezone=True), nullable=False)
+    flow_token = Column(String, nullable=False, unique=True)
+    email_verified = Column(Boolean, nullable=False, default=False)
+    email_sent = Column(Boolean, nullable=False, default=False)
```
kalvdans, line 469: Always true so no need to store in the database.

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+    email_token = Column(String, nullable=True)
+    email_token_created = Column(DateTime(timezone=True), nullable=True)
+    email_token_expiry = Column(DateTime(timezone=True), nullable=True)
```
kalvdans, line 471: Reduntant, could be calculated from email_token_created or vice versa.

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+
+    ## Basic
+    name = Column(String, nullable=False)
+    # TODO: unique across both tables
+    email = Column(String, nullable=False, unique=True)
```
kalvdans, line 476: No, we should allow duplicates here. First to verify gets the account. I don't want someone to be able to block me from creating an account just because they know my email.

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+    contribute_ways = Column(ARRAY(String), nullable=True)
+    expertise = Column(String, nullable=True)
 
     @hybrid_property
-    def is_valid(self):
-        return (self.created <= func.now()) & (self.expiry >= func.now())
+    def token_is_valid(self):
+        return (
+            (self.email_token != None)
+            & (self.email_token_created <= func.now())
```
kalvdans: no need to check created date, we don't time travel.

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+
+    id = Column(BigInteger, primary_key=True)
+
+    # housekeeping
     created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
-    expiry = Column(DateTime(timezone=True), nullable=False)
+    flow_token = Column(String, nullable=False, unique=True)
+    email_verified = Column(Boolean, nullable=False, default=False)
+    email_sent = Column(Boolean, nullable=False, default=False)
+    email_token = Column(String, nullable=True)
```
kalvdans, line 470: nullable=False as we always sent out an email the first thing that happens.

app/backend/src/couchers/models.py
```diff
@@ -341,26 +342,111 @@ class FriendRelationship(Base):
...
+    email_token_expiry = Column(DateTime(timezone=True), nullable=True)
+
+    ## Basic
+    name = Column(String, nullable=False)
+    # TODO: unique across both tables
+    email = Column(String, nullable=False, unique=True)
+    # TODO: invitation, attribution
+
+    ## Account
+    filled_account = Column(Boolean, nullable=False, default=False)
```
kalvdans: Redundant, could be calculated from the other columns.

app/backend/src/couchers/servicers/auth.py
```diff
@@ -170,102 +153,194 @@ def _username_available(self, username):
...
+                    existing_user = session.query(User).filter(User.email == request.basic.email).one_or_none()
+                    if existing_user:
+                        context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.SIGNUP_FLOW_EMAIL_TAKEN)
+                    existing_flow = (
+                        session.query(SignupFlow).filter(SignupFlow.email == request.basic.email).one_or_none()
+                    )
+                    if existing_flow:
+                        send_signup_email(existing_flow)
+                        session.commit()
+                        context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.SIGNUP_FLOW_EMAIL_STARTED_SIGNUP)
```
kalvdans, line 149: I think we should allow several people starting to sign up using the same email address, and the email that gets clicked will "win" and the other partially signed up will be discarded.

app/backend/templates/signup_continue.md
```diff
@@ -0,0 +1,24 @@
...
+{% from "macros.html" import button %}
+Hi {{ escape(flow.name) }}!
+
+Please finish signing up for Couchers.org.
+
+{% if html %}
+
+{{ button("Continue signing up", signup_link) }}
+
+Please click on the above button to continue, or click the following link: <{{ signup_link }}>.
```
kalvdans, line 14: Maybe a blurb about "if this was not you, please ignore this message."

kalvdans (changes requested):

app/pb/account.proto
aapeliv: The idea in the future would be that people can re-fill the form through the account API. But I'll move it at least for now

aapeliv:
Thanks @kalvdans: really good comments. I'll fix this up in the next couple of days.

app/backend/src/couchers/models.py
kalvdans: I got consistency errors when I set `nullable=False` and it is a bit of spagetti code to follow. Will try to restructure.

app/backend/src/couchers/servicers/auth.py
kalvdans: If we allow duplicate usernames, your username might be stolen in front of your eyes although it was was ok in the beginning of the signup flow. How should we handle usernames?

aapeliv: I think it'd be good to check in the backend to make sure the usernames are approximately unique across both tables (though there will still be a slight race condition), just to prevent this as a courtesy to the user.

But otherwise I think it's okay to just have it unique in both tables and the database can enforce that.

app/backend/src/couchers/models.py
aapeliv: Did you figure it out?

kalvdans: Not yet, I think it has to do with refresh() setting it to null temporarily before the transaction is committed.

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -103,47 +97,44 @@ export default function CompleteSignupForm() {
...
-      signupToken: urlToken,
-      username: sanitizeName(data.username),
-    });
+    try {
+      if (!authState.flowState) {
+        authActions.authError(
+          "Submitting account form without current signup flow in progress"
+        );
+      }
+      // authActions catches errors here
```
lucaslcode: You are now catching errors here instead - I think it's better to do it in `authActions` so all the business logic is in one place.

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -154,7 +145,7 @@ export default function CompleteSignupForm() {
         <>
           <form
             className={`${authClasses.form} ${classes.firstForm}`}
-            onSubmit={completeSignup}
```
lucaslcode: The loading indicator above this isn't necessary anymore and now causes all the fields to unmount when clicking submit, so if there is an error like "You must be 18", all the things you filled out disappear.

app/backend/src/couchers/tasks.py
```diff
@@ -4,18 +4,39 @@
...
-    signup_link = urls.signup_link(signup_token=token.token)
+def send_signup_email(flow):
+    logger.info(f"Sending signup email to {flow.email=}:")
+
+    # whether we've sent an email at all yet
+    email_sent_before = flow.email_sent
+    if flow.email_verified:
+        # we just send a link to continue, not a verification link
+        signup_link = urls.signup_link(token=flow.token)
+    elif flow.email_token and flow.token_is_valid:
```
lucaslcode, line 25:
```
2021-06-25 21:49:04,314: couchers.interceptors: Probably an unknown error! Sanitizing...

2021-06-25 21:49:06,796: couchers.tasks: Sending signup email to flow.email='lucas1@test.com':

2021-06-25 21:49:06,821: couchers.interceptors: unsupported operand type(s) for &: 'bool' and 'BinaryExpression'

Traceback (most recent call last):

File "/app/src/couchers/interceptors.py", line 235, in sanitizing_function

res = prev_func(req, context)

File "/app/src/couchers/interceptors.py", line 212, in tracing_function

raise e

File "/app/src/couchers/interceptors.py", line 197, in tracing_function

res = prev_func(request, context)

File "/app/src/couchers/interceptors.py", line 117, in user_unaware_function

return user_aware_function(req, context)

File "/app/src/couchers/servicers/auth.py", line 147, in SignupFlow

send_signup_email(existing_flow)

File "/app/src/couchers/tasks.py", line 24, in send_signup_email

elif flow.email_token and flow.token_is_valid:

File "/usr/local/lib/python3.8/site-packages/sqlalchemy/ext/hybrid.py", line 900, in __get__

return self.fget(instance)

File "/app/src/couchers/models.py", line 489, in token_is_valid

return (self.email_token != None) & (self.email_token_expiry >= func.now())

TypeError: unsupported operand type(s) for &: 'bool' and 'BinaryExpression'
```

lucaslcode: This occurred when I stopped signing up mid-way, deleted my cookies, and used the same email.

After I successfully signed up, I tried logging out and signing up with the same email and got this error again instead of "Email taken". Are you sure signup flows are being deleted on successful signup?

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+    flexDirection: "row",
+  },
+}));
+
+interface ContributorFormProps {
+  // Return true for success, false for failure. Note that this component does not display errors on failure, do it in the parent.
+  processForm: (form: ContributorFormPb) => Promise<boolean>;
+}
+
+export default function ContributorForm({ processForm }: ContributorFormProps) {
```
lucaslcode, line 68: There are now 2 contributor forms. Did you mean to delete the old one in `features/contributorForm`?

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const [success, setSuccess] = useState(false);
+
+  const { control, register, handleSubmit, errors } =
+    useForm<ContributorInputs>({
+      mode: "onBlur",
+      shouldUnregister: false,
+    });
+
+  const submit = handleSubmit(async (data: ContributorInputs) => {
+    setLoading(true);
```
lucaslcode: Put this in a react-query mutation and you won't need to set your own loading state and errors

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+          <Controller
+            id="contributeWays"
+            control={control}
+            name="contributeWays"
+            defaultValue=""
+            render={({ onChange, value }) => (
+              <FormControl>
+                <FormLabel>{CONTRIBUTE_WAYS_LABEL}</FormLabel>
+                <FormGroup aria-label={CONTRIBUTE_WAYS_ARIA_LABEL}>
+                  {CONTRIBUTE_WAYS_OPTIONS.map(({ name, description }) => (
```
lucaslcode: I think not destructuring is clearer here - `name` is an important prop for react-hook-form which is a bit confusing
```suggestion
                  {CONTRIBUTE_WAYS_OPTIONS.map((option) => (
```

app/frontend/src/components/ContributorForm/constants.ts
```diff
@@ -0,0 +1,80 @@
+export const CONTRIBUTE = "Sign up to volunteer with Couchers.org";
+export const JOIN_THE_TEAM =
+  "Join the worldwide team of couch surfers bringing this community-run platform to life";
+export const FILL_IN_THE_FORM =
+  "Fill in the form below to sign up to volunteer. We'll get in touch if we need your skills.";
```
lucaslcode: This is in the signup form now so some of these labels need to be different. Can probably remove this one completely.

app/frontend/src/components/ContributorForm/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+  {
+    name: "web",
+    description: "Development, web (TS/React)",
+  },
+  {
+    name: "other",
+    description: "Other",
+  },
+];
+export const CONTRIBUTE_WAYS_ARIA_LABEL = "contribute ways";
```
lucaslcode:
```suggestion
export const CONTRIBUTE_WAYS_ARIA_LABEL = "Ways I can contribute";
```

app/frontend/src/components/ContributorForm/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+  "Please share any ideas you have that would improve the couch surfing experience for you and for the community.";
+export const IDEAS_HELPER =
+  "Feel free to describe any problems you've had or experienced with other platforms, and what you'd like to see done about them.";
+
+export const FEATURES_LABEL =
+  "What feature would be most important to you? How could we make that feature as good as possible for your particular use?";
+export const FEATURES_HELPER =
+  "Do you care about hosting or surfing? Events or hangouts? Wish there was a better messaging system? Do you like mobile apps or prefer to use a computer?";
+
+export const CONTRIBUTE_LABEL = "Would you like to contribute to ...";
```
lucaslcode:
```suggestion
export const CONTRIBUTE_LABEL = "Would you like to contribute to Couchers.org?";
```

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+            id="contribute"
+            control={control}
+            name="contribute"
+            defaultValue=""
+            render={({ onChange }) => (
+              <FormControl>
+                <FormLabel component="legend">{CONTRIBUTE_LABEL}</FormLabel>
+                <RadioGroup
+                  className={classes.contributeRadio}
+                  aria-label={CONTRIBUTE_ARIA_LABEL}
```
lucaslcode:
```suggestion
```
This will have an aria label from the above FormLabel

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -0,0 +1,242 @@
...
+          />
+          <Controller
+            id="contributeWays"
+            control={control}
+            name="contributeWays"
+            defaultValue=""
+            render={({ onChange, value }) => (
+              <FormControl>
+                <FormLabel>{CONTRIBUTE_WAYS_LABEL}</FormLabel>
+                <FormGroup aria-label={CONTRIBUTE_WAYS_ARIA_LABEL}>
```
lucaslcode:
```suggestion
                <FormGroup>
```

app/frontend/src/components/ContributorForm/constants.ts
lucaslcode: Actually these can just be removed if they're not being used anymore

app/frontend/src/components/ContributorForm/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+export const CONTRIBUTE_OPTIONS = ["Yes", "Maybe", "No"];
+export const CONTRIBUTE_ARIA_LABEL = "contribute";
```
lucaslcode: remove

app/frontend/src/components/ContributorForm/index.ts
```diff
@@ -0,0 +1,3 @@
+// depending on the page the form is on, we want to style these differently, so leave up to the embedding page
```
lucaslcode: I don't understand this

app/frontend/src/features/auth/constants.tsx
```diff
@@ -56,23 +54,31 @@ export const RESET_PASSWORD_SUCCESS =
   "Your password has been reset successfully! You will now be emailed a magic link when you log in with your username or email.";
 export const SIGN_UP = "Sign up";
 export const SIGN_UP_AGREEMENT =
-  "By signing up, you agree with the T&Cs of using the platform and confirm to adhere to our Code of Conduct.";
+  "By continuing, you agree to our Terms of Service, including our cookie, email, and data handling policies.";
```
lucaslcode: No link to these?

app/frontend/src/features/auth/signup/BasicForm.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+describe("signup form (basic part)", () => {
+  it("cannot be submitted empty", async () => {
+    const { result } = renderHook(() => useAuthContext(), { wrapper });
+    expect(result.current.authState.authenticated).toBe(false);
+    expect(result.current.authState.flowState).toBe(null);
+
+    render(<BasicForm />, { wrapper });
+    userEvent.click(await screen.findByRole("button", { name: CONTINUE }));
+
+    await waitFor(() => {
```
lucaslcode, line 36: Do you get warnings without this `waitFor`? Doesn't make much sense to wait for something not to be called.

app/frontend/src/features/auth/signup/BasicForm.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+      expect(startSignupMock).not.toBeCalled();
+    });
+
+    expect(result.current.authState.authenticated).toBe(false);
+    expect(result.current.authState.flowState).toBe(null);
+  });
+
+  it("cannot be submitted without name", async () => {
```
lucaslcode:
```suggestion
  it("cannot be submitted without email", async () => {
```

app/frontend/src/features/auth/signup/BasicForm.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    await waitFor(() => {
+      expect(startSignupMock).not.toBeCalled();
+    });
+
+    expect(result.current.authState.authenticated).toBe(false);
+    expect(result.current.authState.flowState).toBe(null);
+  });
+
+  it("cannot be submitted without email", async () => {
```
lucaslcode:
```suggestion
  it("cannot be submitted without name", async () => {
```

app/frontend/src/features/auth/signup/BasicForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+        await service.auth.startSignup(data.name, sanitizedEmail)
+      );
+    } catch (err) {
+      authActions.authError(err.message);
+    }
+    setLoading(false);
+  });
+
+  return (
+    <>
```
lucaslcode, line 54: Unnecessary fragment

app/frontend/src/features/auth/signup/BasicForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const { register, handleSubmit, errors } = useForm<SignupBasicInputs>({
+    mode: "onBlur",
+    shouldUnregister: false,
+  });
+
+  const onSubmit = handleSubmit(async (data: SignupBasicInputs) => {
+    setLoading(true);
+    authActions.clearError();
+    try {
```
lucaslcode: Handle errors in the action

app/frontend/src/features/auth/signup/BasicForm.tsx
```diff
@@ -0,0 +1,109 @@
...
+    } catch (err) {
+      authActions.authError(err.message);
+    }
+    setLoading(false);
+  });
+
+  return (
+    <>
+      <form className={authClasses.form} onSubmit={onSubmit}>
+        <InputLabel className={authClasses.formLabel} htmlFor="name">
```
lucaslcode, line 59: This doesn't work for accessibility. I think the TextField has to be an Input and they have to be wrapped with FormControl.

lucaslcode: I just check this one, but probably the same applies to all the signup and login ones. That's probably why the forum user said we aren't accessible! Would be good to fix this now.

app/frontend/src/features/auth/signup/FeedbackForm.tsx
```diff
@@ -0,0 +1,31 @@
...
+export default function FeedbackForm() {
+  const { authActions, authState } = useAuthContext();
+
+  const handleSubmit = async (form: ContributorFormPb) => {
+    authActions.clearError();
+    try {
+      if (!authState.flowState) {
+        authActions.authError(
+          "Submitting feedback form without current signup flow in progress"
+        );
```
lucaslcode: If you return here, you won't need the ? in authState?.flowState

app/frontend/src/features/auth/signup/Signup.tsx
```diff
@@ -47,52 +55,89 @@ const useStyles = makeStyles((theme) => ({
...
+        </Typography>
+      </>
+    );
+  } else if (state.needAccount) {
+    return <AccountForm />;
+  } else if (state.needFeedback) {
+    return <FeedbackForm />;
+  } else if (state.needVerifyEmail) {
+    return (
+      <>
```
lucaslcode: Typography

app/frontend/src/features/auth/signup/Signup.tsx
```diff
@@ -47,52 +55,89 @@ const useStyles = makeStyles((theme) => ({
...
-            </Route>
-          </Switch>
+          <AuthHeader>{SIGN_UP_HEADER}</AuthHeader>
+          {error && (
+            <Alert className={authClasses.errorMessage} severity="error">
+              {error}
+            </Alert>
+          )}
+          {loading ? <CircularProgress /> : <CurrentForm />}
+          <Typography className={classes.logIn}>
```
lucaslcode: The desktop one you only show this if there's no flow state

app/frontend/src/features/auth/useAuthStore.test.ts
```diff
@@ -116,27 +115,78 @@ describe("tokenLogin action", () => {
   });
 });
 
-describe("signup action", () => {
-  //not testing for the user itself, as that is handled by getUser
-  //and getCurrentUser
-  it("sets authenticated correctly", async () => {
-    signupMock.mockResolvedValue({ jailed: false });
-    getCurrentUserMock.mockResolvedValue(defaultUser);
+describe("firstLogin action", () => {
```
lucaslcode, line 118: Needs error testing

lucaslcode (changes requested):

app/frontend/src/features/auth/signup/BasicForm.tsx
lucaslcode: I was wrong about this, it's fine

app/backend/src/couchers/models.py
kalvdans: I am travelling abroad and will not have time to work more. Looks good to merge and do cleanups later!

aapeliv: OK, thanks. I'll take it from here!

app/backend/src/couchers/tasks.py
aapeliv: Fixed it!

PR #1330, opened 2021-05-31
Regions and languages as maps

aapeliv:
What do you think?

The ergonomics of using this through `toObject` isn't particularly different in the frontend; on the backend it's a bit nicer.

The other thing that I have in mind is that I'm not sure what we'll do with this once we do translations, but I don't know what we're doing the other way either... We can cross that bridge when we get there.

### DIFF
--- a/app/frontend/src/features/profile/edit/EditProfile.tsx
+++ b/app/frontend/src/features/profile/edit/EditProfile.tsx
@@ -35,6 +35,8 @@ import {
   MEETUP,
   NO_MEETUP,
   NOT_ACCEPTING,
+  UNKNOWN_LANGUAGE,
+  UNKNOWN_REGION,
 } from "features/profile/constants";
 import { useLanguages } from "features/profile/hooks/useLanguages";
 import { useRegions } from "features/profile/hooks/useRegions";
@@ -152,7 +154,7 @@ export default function EditProfileForm() {
     register("radius");
   }, [register]);
 
-  const { regions, regionsLookup } = useRegions();
+  const { data: regions } = useRegions();
   const { data: languages } = useLanguages();
 
   const onSubmit = handleSubmit(
@@ -162,12 +164,8 @@ export default function EditProfileForm() {
         {
           profileData: {
             ...data,
-            regionsVisited: regionsVisited.map(
-              (region) => (regionsLookup || {})[region]
-            ),
-            regionsLived: regionsLived.map(
-              (region) => (regionsLookup || {})[region]
-            ),
+            regionsVisited,
+            regionsLived,
             languageAbilities: {
               valueList: fluentLanguages.map((code) => ({
                 code,
@@ -368,24 +366,26 @@ export default function EditProfileForm() {
             {languages && (
               <Controller
                 control={control}
-                defaultValue={user.languageAbilitiesList.map((ability) =>
-                  languages.get(ability.code)
+                defaultValue={user.languageAbilitiesList.map(
+                  (ability) => ability.code
                 )}
                 name="fluentLanguages"
-                render={({ onChange, value }) => (
-                  <ProfileTagInput
-                    onChange={(_, value) => onChange(value)}
-                    value={value}
-                    getOptionLabel={(languageCode) =>
-                      languages.get(languageCode)!
-                    }
-                    options={Array.from(
-                      languages.keys() as IterableIterator<string>
-                    )}
-                    label={LANGUAGES_SPOKEN}
-                    id="fluentLanguages"
-                  />
-                )}
+                render={({ onChange, value }) => {
+                  return (
+                    <ProfileTagInput
+                      onChange={(_, value) => onChange(value)}
+                      value={value}
+                      getOptionLabel={(languageCode) =>
+                        languages.get(languageCode) || UNKNOWN_LANGUAGE
+                      }
+                      options={Array.from(
+                        languages.keys() as IterableIterator<string>
+                      )}
+                      label={LANGUAGES_SPOKEN}
+                      id="fluentLanguages"
+                    />
+                  );
+                }}
               />
             )}
             <ProfileTextInput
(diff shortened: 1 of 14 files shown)


aapeliv:
Okay, I've no clue what those type signatures are in typescript... Maybe some frontend person wants to have a look.

kalvdans:
It is a bit ugly to get a non-deterministic wire format, but it is indeed shorter code on the backend side. I'm indifferent... you can decide :)

aapeliv:
I would probably favour the other one; but I'm pretty indifferent. Let's see what frontend people think!

thdk:
I don't think this will make much of a difference for frontend. The type of languagesMap and regionsMap still seems to be an array of tuples instead of a map/dictionary so we still have to convert it to a map object. (Luckily this is cached with react-query so it's only done once per user session).

I don't mind this to be merged, but I don't think it's what @darrenvong meant with his comment.

lucaslcode:
@thdk I believe if you change the service function `getLanguages` to *not* call `toObject()`, you will be able to use it like a map with `get` etc.

darrenvong:
Yeah, this only changed how we destructure things in the `reduce` which is no better than what we had for the frontend.

I was thinking the `languages` and `languagesLookup` object/map can be returned directly from the backend rather than having to build it on the frontend... will this be possible now with this `map` type?

aapeliv:
We can’t return a bidirectional map through protobuf. We could do a map, but underneath it’s the same thing, and since we’re using AsObject, this is basically adding nothing.

I propose we close this and go with what’s in the main PR.

>

lucaslcode:
Why don't we not use `toObject`?

aapeliv:
> Why don't we not use `toObject`?

I don't mind if we did. Let me update this PR to do that and see how it looks.

aapeliv:
Actually, I spoke too soon. I don't think there's a way to grab the map as a dictionary even if not using `AsObject`

lucaslcode:
What about this? https://github.com/protocolbuffers/protobuf/blob/bd42fcc7a3e04504df895ce2fd0782c0e84b68a5/js/map.js#L391

darrenvong:
Yeah the `jsonb.Map` thing looks promising, then all we need is the backend to do the reverse map computation and return that as a separate map object?

aapeliv:
> Yeah the `jsonb.Map` thing looks promising, then all we need is the backend to do the reverse map computation and return that as a separate map object?

We can't have the backend return both the forward and the backward map. It's better to return pairs and reconstruct in the frontend.

thdk:
Just posting in case you weren't aware but I also saw that instead of toObject(), we can call getRegionsMap() and getLanguagesMap() that returns an object of that jsonb.Map type.

But if backend only returns a single map, we still need to loop each item to create the reverse/lookup map like we now already do to create both maps.

```javascript
export async function getRegions() {
  const result = await client.resources.getRegions(new Empty());
  return result.getRegionsMap();
}

export async function getLanguages() {
  const languages = await client.resources.getLanguages(new Empty());
  return languages.getLanguagesMap();
}
```

darrenvong:
I just had a quick play around with the language map thing, for some reason the `getOptionLabel` doesn't work and using the map crashes the app when you go straight to `/user/edit` since it doesn't create the `jsonb.Map` object consistently for some very strange reasons. Maybe this is not worth it if we have to re-map things on the frontend...

> We can't have the backend return both the forward and the backward map. It's better to return pairs and reconstruct in the frontend.

Why's that? Is it a protobuf limitation or are you just trying to avoid doing extra computation on the backend?

kalvdans:
> > We can't have the backend return both the forward and the backward map. It's better to return pairs and reconstruct in the frontend.
> 
> Why's that? Is it a protobuf limitation or are you just trying to avoid doing extra computation on the backend?

It is just very silly to send the same data twice over the network.

darrenvong:
> > > We can't have the backend return both the forward and the backward map. It's better to return pairs and reconstruct in the frontend.
> > 
> > 
> > Why's that? Is it a protobuf limitation or are you just trying to avoid doing extra computation on the backend?
> 
> It is just very silly to send the same data twice over the network.

Ok that's a fair point. Still, this feels unsatisfactory as we have to build this up again twice in the frontend. Maybe if we restructure the json/object returned to be more like `{"abk": "Abkhazian", ...}` etc, then at least the frontend only has to do the work once to build the reverse map. But maybe that should be resolved in a separate issue...

lucaslcode:
> I just had a quick play around with the language map thing, for some reason the getOptionLabel doesn't work and using the map crashes the app

@darrenvong can you push your code, I want to try

darrenvong:
Lol shouldn't have deleted the code... but I think I managed to recover what I had so maybe you'll find something I overlooked 🤷 

Only did it for languages but should be enough to figure out why this wouldn't work...

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -368,15 +368,20 @@ export default function EditProfileForm() {
...
                 render={({ onChange, value }) => (
                   <ProfileTagInput
                     onChange={(_, value) => onChange(value)}
                     value={value}
-                    options={Object.values(languages)}
+                    getOptionLabel={(languageCode) =>
+                      languages.get(languageCode)!
+                    }
+                    options={Array.from(
+                      languages.keys() as IterableIterator<string>
```
darrenvong: Not sure if this is the culprit... but without this type assertion, the types of `Array.from` won't accept this as a valid iterator, probably because Google rolled their crappier version with the generated code and it's missing something. But it still seems to work at first (when clicking from profile, then the edit button) until you go to the route directly...

darrenvong:
> Maybe if we restructure the json/object returned to be more like {"abk": "Abkhazian", ...} etc

After more thoughts, never mind, as I'm not sure this is possible since the data is semi dynamic? If we do this, we'd basically have to write all the languages field in the proto contract? That sounds like a terrible idea to me just as I suggested sending the map over twice as we're now duplicating the data in the proto and the resource file. I guess this kind of "JavaScript optimal" structure is not really feasible with a statically typed schema and perhaps what we got is the best at the moment.

darrenvong:
@lucaslcode Pushed the code. Feel free to take a look when you have some time. CI failed because I changed the languages hook but the preview built fine so at least you can see the user profile is completely f**ked since I added "English" to the test user because:
> it doesn't create the `jsonb.Map` object consistently for some very strange reasons. 

And even when we do get a `jsonb.Map` object, the `getOptionLabel` prop seems to have no effect at all as you can see it's still showing the three-letter language code... (you'll have to go to a normal preview site and remove the language first otherwise the profile is broken/not accessible because of this)

lucaslcode:
Figured it out. Maps are not json serializable, so it was being stored as an empty object in the query cache.

The last commit seems to be all working - however persistence is commented out. I'm sure there's a way to add custom serialization for a Map.

darrenvong:
Ahh nice find! I think we'd have to create a custom persistor on top of the default one for this to work probably

aapeliv:
Are we happy to close this?

darrenvong:
Yeah maybe make a low priority issue out of this, as a custom persistor will need a bit of work (probably not too difficult, but not exactly trivial either) and there are probably more important issues to work on first... 😅

PR #1332, opened 2021-05-31
Refactor map layout

Jesus82:
Refactors map layout, removes general 100% height chain of rules and fixes spacing


app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -65,14 +63,13 @@ export default function SearchResult({
     >
       <CardActionArea>
         <CardContent>
-          <UserSummary user={user} avatarIsLink={false} compact>
+          <UserSummary user={user} avatarIsLink={false}>
             <div className={classes.statusLabelWrapper}>
               <div>
                 <Hidden smDown>
-                  <CouchIcon className={classes.statusIcon} />
+                  <CouchIcon className={classes.statusIcon}/>
```
lucaslcode: Remember to run `yarn format`

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -15,14 +15,18 @@ const useStyles = makeStyles((theme) => ({
...
-    height: "100%",
+    position: "fixed",
+    top: "4rem",
+    left: 0,
+    right: 0,
+    bottom: 0,
     [theme.breakpoints.up("md")]: {
-      flexDirection: "row",
+      display: "grid",
+      gridTemplateColumns: "30rem 1fr",
```
lucaslcode: For some reason, the maplibre gl canvas doesn't fill the full space anymore until a user is selected or the window resized. Seems like it might be a maplibre gl bug? Might have to leave it as flex.

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -27,16 +27,14 @@ const useStyles = makeStyles((theme) => ({
     ...theme.typography.body1,
   },
   statusLabelWrapper: {
-    display: "flex",
+    display: "grid",
+    gridTemplateColumns: "1fr 1fr",
```
lucaslcode, line 42: I think flex is better for 1-dimensional layouts (with `justifyContent: "space-around"` to fix the spacing as you have done), but it's not a big deal

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -65,14 +63,13 @@ export default function SearchResult({
     >
       <CardActionArea>
         <CardContent>
-          <UserSummary user={user} avatarIsLink={false} compact>
+          <UserSummary user={user} avatarIsLink={false}>
```
lucaslcode: `compact` makes the avatar smaller and displays only the name. I think we don't need age and location here because there is a map to show location, and the age is listed below. People have annoyingly long location names (which we should maybe fix by not including state and country by default in the location editing...)

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: Also, why grid? You prefer setting the widths in one location?

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -15,14 +15,18 @@ const useStyles = makeStyles((theme) => ({
     display: "flex",
     alignContent: "stretch",
     flexDirection: "column-reverse",
-    height: "100%",
+    position: "fixed",
+    top: "4rem",
```
lucaslcode: this should be `theme.shape.navPaddingXs` and then `theme.shape.navPaddingSmUp` at the right breakpoint.

lucaslcode (changes requested):
Thanks for tidying up the spacing!

app/frontend/src/features/search/SearchResult.tsx
Jesus82: I see. Problem here is that compact outputs the small-sized avatar, when here the size should be the default. Should I write a method for this? (maybe a compact-text prop)?

app/frontend/src/features/search/SearchPage.tsx
Jesus82: I will take a look to that. It seems to me a JS initializing bug of the maplibre canvas.

app/frontend/src/features/search/SearchResult.tsx
lucaslcode: Maybe another `nameOnly` prop?

lucaslcode: Or rather separate `compact` into `nameOnly` and `smallAvatar`

app/frontend/src/features/search/SearchResult.tsx
Jesus82: The _one dimension-flex many dimensions-grid_ rule is kind of popular, but I would say that is not very accurate. I recommend the beginning of this (great) course where that is discussed https://cssgrid.io/. Here grid let's you set the spacing and gutters in the same place (avoiding setting different margins), and also it is useful in order to control word break.

app/frontend/src/features/search/SearchPage.tsx
Jesus82: Finally I used flex for the main layout, as it was the only way for the map to fit its space. Don't know why tbh (some JS black magic).

Jesus82:
Last commit addresses some issues in the mobile view. I set the scrolling to vertical (as now it is possible to do so, and makes cards more readable in mobile taking full width), but I could revert it back to horizontal easily.

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -16,13 +16,14 @@ const useStyles = makeStyles((theme) => ({
...
+    top: theme.shape.navPaddingXs,
     left: 0,
     right: 0,
     bottom: 0,
-    [theme.breakpoints.up("md")]: {
-      display: "grid",
-      gridTemplateColumns: "30rem 1fr",
+    [theme.breakpoints.up("sm")]: {
+      top: theme.shape.navPaddingSmUp,
+      display: "flex",
```
lucaslcode:
```suggestion
```

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -22,19 +22,28 @@ const useStyles = makeStyles((theme) => ({
...
   searchDesktop: {
+    "& > div > *": {
+      flexShrink: 0,
+      width: "100%",
+    },
   },
   scroller: {
     marginTop: theme.spacing(3),
-    "&&": { alignItems: "flex-start" },
+    alignItems: "flex-start",
```
lucaslcode: Sometimes the && is needed to increase the specificity to override the Material-UI style. At the moment this line doesn't get applied (but it doesn't matter for vertical scrolling)

lucaslcode (changes requested):
It should be horizontal scrolling on mobile - the whole page should scroll if we're doing vertical scroll, otherwise it's a bit unusual. (Also that component is called `HorizontalScroller`, so if we did want vertical scrolling we would need some refactoring.)

Jesus82:
Shame on me I didn't realize the name of the component :D I've revert it back to horizontal and fixed spacing issues and added scroll snapping.

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -44,11 +49,17 @@ const useStyles = makeStyles((theme) => ({
...
   searchResult: {
-    [theme.breakpoints.down("sm")]: {
-      maxWidth: "80%",
-    },
-    [theme.breakpoints.up("md")]: {
-      margin: theme.spacing(2),
+    marginBottom: theme.spacing(3),
+    borderRadius: theme.shape.borderRadius * 2,
+    backgroundColor: theme.palette.primary.contrastText,
+    boxShadow: "0 0 4px rgba(0,0,0,0.25)",
```
lucaslcode, line 50: I think we should either use `elevation` on the card, or set this in the theme somewhere

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -44,11 +49,17 @@ const useStyles = makeStyles((theme) => ({
...
+    [theme.breakpoints.down("md")]: {
+      "& > div > *": {
+        padding: 0,
+        margin: theme.spacing(2),
+        height: "calc(100% - 2rem)",
+        overflow: "hidden",
+      },
```
lucaslcode: I don't understand this - there doesn't seem to be any div directly under the search result.

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -21,21 +21,26 @@ import hasAtLeastOnePage from "utils/hasAtLeastOnePage";
...
     },
   },
   baseMargin: { margin: theme.spacing(2) },
   searchDesktop: {
-    margin: theme.spacing(0, 2),
-    marginTop: theme.spacing(2),
+    "& > div > *": {
+      flexShrink: 0,
+      width: "100%",
+    },
```
lucaslcode: The field should accept a `fullWidth` prop which is probably clearer than doing this

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -44,11 +49,17 @@ const useStyles = makeStyles((theme) => ({
...
   },
   searchResult: {
-    [theme.breakpoints.down("sm")]: {
-      maxWidth: "80%",
-    },
-    [theme.breakpoints.up("md")]: {
-      margin: theme.spacing(2),
+    marginBottom: theme.spacing(3),
+    borderRadius: theme.shape.borderRadius * 2,
+    backgroundColor: theme.palette.primary.contrastText,
```
lucaslcode: `theme.palette.background.paper`?

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -8,23 +8,29 @@ import useOnVisibleEffect from "../utils/useOnVisibleEffect";
...
   root: {
     "& > *": {
-      flex: "0 0 auto",
-    },
-    "& > * + *": {
-      marginInlineStart: theme.spacing(2),
+      flexShrink: 0,
+      height: "calc(100% - 2rem)",
+      width: "85vw",
+      maxWidth: "33rem",
```
lucaslcode: I thought it's better to let the children decide their width

lucaslcode (changes requested):
This is great, and I love the scroll snapping!

However, the horizontal scroller is also used on the community page and it looks a bit broken now (links: [this pr](https://frontend-refactor-full-container-in-maps--frontend.preview.couchershq.org/community/1/world), [develop](https://develop--frontend.preview.couchershq.org/community/1/world))

app/frontend/src/features/search/SearchResultsList.tsx
Jesus82: Don't know what happened as I had it working, but you're right. Switched it to button and it works.

app/frontend/src/components/HorizontalScroller.tsx
Jesus82: Without a set width, the card would grow as large as the text inside, making it a really long div.

app/frontend/src/features/search/SearchResultsList.tsx
Jesus82: Box shadow in MUI works only accepting parameters. So we cannot just set that boxShadow in the theme file, as it gives an error. The alternative would be to store it as a CSS variable (we can place that in the index.css file).

Jesus82:
I fixed when in community it scrolled the text when no places. But I don't know if that works with actual places, as there are none created.

app/frontend/src/features/search/SearchResultsList.tsx
lucaslcode: I see. Can you use the MUI class instead of this selector - it's clearer what you're targeting. So it would be `"& MuiCardContent-root:`

app/frontend/src/components/HorizontalScroller.tsx
lucaslcode: This component takes `children` and the children can set their width. For example, the events in the community page fit a few in one screen, even on mobile, but now there is only one fully visible

lucaslcode:
> I fixed when in community it scrolled the text when no places.

Sorry - I was referring to the events

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -8,23 +8,35 @@ import useOnVisibleEffect from "../utils/useOnVisibleEffect";
...
 
 const useStyles = makeStyles((theme) => ({
-  padder: {
-    height: 1,
-    width: 0.1,
-  },
   root: {
-    "& > *": {
-      flex: "0 0 auto",
+    "& > div": {
```
lucaslcode: Why did you change from * to div? What if the children are not divs?

app/frontend/src/features/search/SearchResultsList.tsx
lucaslcode: You can put what you like in the theme file (even just this whole string) if you extend the type definition (as is being done at the top of `theme.ts` for `theme.shape.borderRadius`)

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -44,11 +45,20 @@ const useStyles = makeStyles((theme) => ({
...
+    backgroundColor: theme.palette.background.paper,
+    boxShadow: "0 0 4px rgba(0,0,0,0.25)",
+    [theme.breakpoints.down("md")]: {
+      "& > button": {
+        height: "100%",
+      },
+      "& > button > *": {
+        padding: 0,
+        margin: theme.spacing(2),
+        height: "calc(100% - 2rem)",
```
lucaslcode:
```suggestion
        height: `calc(100% - ${theme.spacing(1)})`,
```
However I wonder if it's easier for the parent to have a padding, and the height to just be 100%. Or even better, `alignItems: "stretch"`?

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -8,23 +8,29 @@ import useOnVisibleEffect from "../utils/useOnVisibleEffect";
...
     flexDirection: "row",
-    overflowX: "auto",
-    padding: theme.spacing(2),
+    height: "100%",
+    width: "100%",
+    padding: theme.spacing(0, 2),
     WebkitOverflowScrolling: "touch",
+    overflowX: "scroll",
+    scrollSnapType: "x mandatory",
+    scrollPadding: theme.spacing(1.5),
```
darrenvong, line 21: I think we probably want to avoid using `0.5` units in the spacing value, so we should go with 1 or 2? Otherwise we are just moving off the 8px scale we agreed with design (even though I know they haven't been the best at sticking to it in the Figma designs) and defeats the point of having one.

Jesus82: I totally agree to follow the grid, but in some cases, the `4px` unit will be the only way to do that (for example if you apply `margin-left` and `margin-right` to a set of elements in order to achieve an `8px` gutter while still having them centered. Here it is more a visual matter. If we apply `scroll-padding: 16px`, `the box-shadow` of the previous card enters in the screen (in the left part of the scroller), which is not intended.
Captura de pantalla de 2021-06-06 13-01-43

But if we apply `scroll-padding: 12px`, we get rid of that piece of shadow, and the proportion is still nice (3:4, which is used all over the design).
Captura de pantalla de 2021-06-06 13-01-21

Jesus82:
Made it so the width is determined just for the map scroller. The only pending thing would be the box shadow (I couldn't make that work in the theme file).

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -19,33 +19,16 @@ export const useCommunityPageStyles = makeStyles((theme) => ({
...
-    alignItems: "flex-start",
-    [theme.breakpoints.down("xs")]: {
-      //break out of page padding
-      left: "50%",
-      marginLeft: "-50vw",
-      marginRight: "-50vw",
-      position: "relative",
-      right: "50%",
-      width: "100vw",
-    },
```
lucaslcode: Not a fan of this? It allowed the horizontal scroller to go all the way to the edge

app/frontend/src/features/search/SearchResultsList.tsx
```diff
@@ -44,11 +45,27 @@ const useStyles = makeStyles((theme) => ({
...
+      scrollSnapAlign: "start",
+      "&:last-child": {
+        marginRight: 0,
+      },
+      "& MuiCardContent-root": {
+        padding: 0,
+        height: `calc(100% - ${theme.spacing(4)})`,
+        overflow: "hidden",
+        margin: theme.spacing(2),
+      },
```
lucaslcode: Invalid selector, looks fine though so remove this?

lucaslcode (changes requested):
All looks good! Just some tiny things, and can you also fix the storybook example for `HorizontalScroller` which is broken on mobile somehow?

lucaslcode:
With the box shadow, I saw there's a few other places with a custom shadow so let's leave it for now.

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
Jesus82: ooops, missed that. Will do another commit later this night.

app/frontend/src/components/HorizontalScroller.tsx
darrenvong: Hmm maybe we should review the scale to be `4px` based instead. I'm seeing more and more cases where we have to break out of it because things don't look right otherwise, which suggests to me it's not granular enough or designs need to adjust to having more spaces around things...

Or stick to what we have and allow half units, but no more than that? Like you said, for anything that needs centering with margin/padding, the half unit seems like a necessary evil.

lucaslcode: I think allow half in special circumstances. Having 4px grid would mean we would have lots of `theme.spacing(4)` which seems too high for the most common spacing.

app/frontend/src/features/search/SearchResultsList.tsx
Jesus82: It was a fix for overlap happening in narrow screens. Should be working with last commit.

app/frontend/src/components/HorizontalScroller.tsx
```diff
@@ -13,7 +13,6 @@ const useStyles = makeStyles((theme) => ({
     display: "inline-flex",
     flexDirection: "row",
     height: "100%",
-    width: "100%",
```
lucaslcode: scroll snap stops working with this.
I think you have to leave this in and do `"& > *": { flexShrink: 0 }`

lucaslcode (changes requested):

lucaslcode:
Sorry about the merge conflicts - I tried to keep changes as minimal as possible. I'm happy to do it if you want.

Jesus82:
> Sorry about the merge conflicts - I tried to keep changes as minimal as possible. I'm happy to do it if you want.

No worries, I think right now everything should be OK (I also adapted the spacing to fit the button in all cases.

app/frontend/src/components/HorizontalScroller.tsx
lucaslcode: @Jesus82 just needs the flex shrink 0 to fix storybook

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -39,9 +44,24 @@ const useStyles = makeStyles((theme) => ({
     "& > div": {
       display: "flex",
     },
+    [theme.breakpoints.down("md")]: {
+      "& > div": {
+        display: "grid",
+        gridTemplateColumns: "1.25rem 1fr",
+        gridGap: theme.spacing(1),
+      },
```
lucaslcode, line 53:
```suggestion
        alignItems: "center",
      },
```

app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -79,34 +102,19 @@ export default function SearchResult({
...
-              <Hidden smDown>
-                <LocationIcon className={classes.statusIcon} />
-              </Hidden>
-              <Typography
-                className={classes.statusLabel}
-                display="inline"
-                variant="body1"
-              >
+              <LocationIcon className={classes.statusIcon} />
+              <Typography display="inline" variant="body1">
```
lucaslcode, line 118: Thanks for fixing the wrapping, but I think we should still hide the icons for xsDown. We want to support 320px width displays and with the icons at that width, you can't see much label. (That might mean this needs to be flex rather than grid.)

lucaslcode (changes requested):
Pinky promise this is the last one!

app/frontend/src/features/search/SearchResult.tsx
Jesus82: This one is really tricky.
If you get rid of the icons, the labels lose their context, and in some cases is hard to guess what are they referring to

Image 1

Making it flex is also problematic, as when both texts don't fit, they push all the elements down, breaking the layout

Image 2

This could be even worse with languages that are not as economic as English when it comes to hosting.

Image 3

My suggestion would be to have semantic icons (someting like a cross over the place icon meaning I can't host), and leave the text with ellipsis as a complement.

lucaslcode: Good idea. Let's leave this for now then and ask design for some icons

Jesus82: Fine, let me know if there's something pending here

lucaslcode (approved):

PR #1333, opened 2021-05-31
Feature/frontend/communities tab

### LINKED ISSUE #1318: Community tab

lucaslcode:
~~Start of #1318 but still need a listing of all communities. Probably in a dialog box.~~ Closes #1318 

I put it on the home page and made the home neater, since the home will eventually be a proper dashboard and that's where this belongs.

Also no tests since there's no new logic, really.

**Frontend checklist**
- [ ] All tests pass


app/frontend/src/features/dashboard/Home.tsx
```diff
@@ -0,0 +1,120 @@
...
+  },
+  accordionTitle: {
+    flexBasis: "30vw",
+    flexGrow: 1,
+    marginInlineEnd: theme.spacing(4),
+  },
+  accordionSubtitle: {
+    ...theme.typography.h3,
+    color: theme.palette.grey[600],
+    flexGrow: 100,
```
lucaslcode: Is there a better way to do this? I want the subtitles to all be aligned, but I also want the title to fill the entire space if the subtitle is pushed to the next line.

aapeliv:
Merged that listing PR and upgraded the preview so that should all work now.

ItsiW:
Could we flesh out the welcome message a bit? I understand we want to make it cleaner, but for many users this may be the first page they read/interact with if they skip the landing page.

> Being in Beta means that we are still in the process of building the platform. Thank you for being one of the first people to join in this exciting journey. The Beta is designed so early adopters like you can try out our new features and help with **finding and reporting bugs** by clicking the red button on the top right of the page.
> 
> We are always in need of new Contributors. If you are interested in **volunteering for Couchers.org**, please open and fill out the "Sign up to volunteer" tab below. [Read about the team and becoming a contributor](https://couchers.org/volunteer). We are looking for everything from developers (check out our [open source github repo](https://github.com/Couchers-org/couchers)), to design to marketing to community and event organising.
> 
> If you are interested in helping this project succeed more generally, please tell your friends, communities and the wonderful people you have met while couch surfing. We have a [community forum](https://community.couchers.org) to discuss and have your say on everything to do with Couchers.org. We also would love for you to join our **weekly online social events** that you can read about below.

lucaslcode:
Aybe ask the design team. My limited UX knowledge says that the way I have done it, users are actually *more* likely to read, because they can just scan the headings and click if interested, whereas they'd probably ignore a text block completely. 

Or maybe I should add another expansion panel? "Bugs"?

aapeliv:
I don't think that bug paragraph adds anything, it just clutters the page.

ItsiW:
Yeah fair enough. We were brainstorming ways we can encourage people to help out by sharing with friends, social media, FB groups etc. But we can revisit this later.

Also, the volunteer signup section has been split into two columns. Not sure if that is intentional.

lucaslcode:
image
The order of all communities is pretty arbitrary

aapeliv:
It’s creation order. Everything currently is ordered by ID, which is a big thing that needs to be fixed at some point.

>

app/frontend/src/features/dashboard/Home.tsx
darrenvong: I think setting the `flexBasis` for both the title and subtitle to `50%` and remove the `flexWrap` on the `accordion` class would work? That'd also allow us to avoid more "magic values" like "30vw" on the title.

Might want to reduce the `marginInlineEnd` on smaller screen with that clamp trick you showed me? The gap looks a bit too wide to me on those screens otherwise.

app/frontend/src/features/dashboard/Home.tsx
```diff
@@ -0,0 +1,151 @@
...
+        >
+          <Typography variant="h2" className={classes.accordionTitle}>
+            {CONTRIBUTE}
+          </Typography>
+          <Typography className={classes.accordionSubtitle}>
+            {JOIN_THE_TEAM}
+          </Typography>
+        </AccordionSummary>
+        <AccordionDetails>
+          <ContributorForm />
```
darrenvong, line 142: Screenshot 2021-06-02 at 01 14 17
The alignment of the button looks broken somehow it's now in the accordion...

darrenvong (changes requested):
Mostly looking good - I guess this part of the app is still subject to major changes so ok to let the tests slide for now :p

app/frontend/src/features/dashboard/Home.tsx
lucaslcode: I do want it to wrap, though!

darrenvong: You mean for the title and subtitle to wrap without them keep going to the rest of the line right? I think what I suggested works with that too

lucaslcode: What I want is:
- if there's enough room for both on one line when the title width is 30vw, but them both in the same line (and have the title take up 30vw, subtitle takes the rest)
- if not, have title and subtitle on their own line, taking up the whole line

darrenvong: Wouldn't we end up with cases where one long title would put things onto two lines, but the other accordion headers with a short title are still on one line with two columns if we do the wrap thing? So then the titles and subtitles are not lined up vertically anymore in those cases

lucaslcode: Yes, but I think it's fine?

darrenvong: Personally I think it looks better if all titles/subtitles are in two columns at some large screen sizes, and goes to two rows in smaller screens. Having a mix where each titles do their own thing doesn't feel as tidy to me. Maybe show a screenshot of that scenario and see what design thinks too?

ItsiW:
Looks like the text margins need to be corrected

app/frontend/src/features/dashboard/Home.tsx
lucaslcode: Fiiiine I'll do a media query 😠

lucaslcode:
> Looks like the text margins need to be corrected

The MUI fields are not designed/intended to have labels that wrap, and there's no easy way to get it to work. So I just moved them above.

app/frontend/src/features/contributorForm/ContributorForm.tsx
```diff
@@ -257,63 +258,95 @@ export default function ContributorForm() {
...
             id="ideas"
             margin="normal"
             name="ideas"
-            label={IDEAS_LABEL}
             helperText={IDEAS_HELPER}
             fullWidth
             multiline
             rows={4}
             rowsMax={6}
+            aria-labelledby="ideas-label"
```
darrenvong, line 329: What's the behaviour with this as well as setting the `id` prop (which prompts MUI to do this usually)? Will this win and override the internal behaviour? Maybe we should remove setting `id` for all these fields that have an explicit `aria-labelledby` value

app/frontend/src/features/dashboard/Home.tsx
```diff
@@ -0,0 +1,146 @@
...
+const useStyles = makeStyles((theme) => ({
+  button: { display: "block", marginTop: theme.spacing(1) },
+  divider: { marginTop: theme.spacing(3) },
+  accordion: {
+    marginBlockStart: theme.spacing(2),
+    "& .MuiAccordionSummary-content": {
+      display: "grid",
+      alignItems: "baseline",
+      gridTemplateColumns: "repeat(auto-fit, 100%)",
+      [theme.breakpoints.up("md")]: {
```
darrenvong, line 44: I think we'd want to set a `columnGap` between the columns in case the accordion title ever gets super long? Currently there is a (small) chance the title would be right next to the subtitle.
Screenshot 2021-06-04 at 00 58 31

app/frontend/src/features/dashboard/Home.tsx
```diff
@@ -0,0 +1,146 @@
...
+        gridTemplateColumns: "30vw 1fr",
```
darrenvong:
```suggestion
        gridTemplateColumns: "1fr 1fr",
```
Is there a good reason not to use `fr` unit for the title too? It looks alright to me when I played around in the dev tools, then it's less "magic value"

darrenvong: But nice to see grid fixed this generally without the awkward `flexGrow: 100` thing!

darrenvong (changes requested):

app/frontend/src/features/contributorForm/ContributorForm.tsx
lucaslcode: Seems to work, I checked the accessibility tree

app/frontend/src/features/dashboard/Home.tsx
lucaslcode: I want the title to be narrower. I'll do `1fr 2fr`

darrenvong (approved):

PR #1264, opened 2021-05-18
Giving References

javiwq:
A big part of #929

**Frontend checklist**
- [x] Formatted my code with `yarn format && yarn lint --fix`
- [x] There are no warnings from `yarn lint`
- [x] There are no console warnings when running the app
- [x] Added any new components to storybook
- [x] Added tests where relevant
- Has some initial tests, but needs a followup to add more
- [x] All tests pass
- [x] Clicked around my changes running locally and it works
- see below! 
- [x] Checked Desktop, Mobile and Tablet screen sizes

**There is a problem with the host request references,** which I think might come from the backend but otherwise I need another set of eyes to look at this. After a confirmed host request has occurred, the reference is available for both parties. After the first person submits a reference, the other person involved with that host request no longer has a reference available. I've tried myself and can't seem to fix it. Depending on what comes up during the review process we might have to raise another Issue to deal with the backend bug (if that's where the problem lies).

Otherwise, clicking around locally works. Here is leaving a friend reference on Desktop: 

https://user-images.githubusercontent.com/35485667/118649311-a6dea380-b7e3-11eb-876b-fc7f2bafbbbf.mov

And with a mobile screen size: 

https://user-images.githubusercontent.com/35485667/118649826-3be19c80-b7e4-11eb-9d5a-4a27132ea784.mov

PR #1345, opened 2021-06-02
Revert "Giving References"

javiwq:
Reverts Couchers-org/couchers#1264


javiwq:
Not sure what happened, the tests were all passing and now suddenly the build failed cause of a storybook file :/

lucaslcode:
Let's just fix it instead of reverting, I will have a look

PR #1347, opened 2021-06-02
Fix mock user properties

lucaslcode:
@ineswq problem was a merge just before yours that changed the user type


thdk (approved):

javiwq:
Thanks for taking a look!

darrenvong:
I'm going to hit the merge button since this is blocking off further branching off develop for development...

PR #1350, opened 2021-06-03
Backend/feature/prometheus poc

### LINKED ISSUE #1298: Make prometheus POC
- Add a prometheus exporter to the API
- Create an example metric for prometheus to scrape

First step towards being able to visualise metrics on a dashboard

lennartvanlaake:
Basic monitoring for the couchers backend.

Note: before this runs in production a file callled `prometheus.prod.env` should exist on our server(s) that sets the variables:
- PROMETHEUS_ENVIRONMENT to a sensible name describing the environment it is running on (could be 'next' for the preview environment and 'production' for the production environment). This will be added as a label to all metrics so we can distinguish metrics from various runtimes.
- GRAFANA_CLOUD_TOKEN to a token I sent to @aapeliv 

closes #1298

**Backend checklist**
- [x] All tests pass


app/docker-compose.prod.yml
```diff
@@ -60,7 +61,18 @@ services:
     networks:
       - media_nginx
       - envoy_nginx
+  prometheus:
+    build:
+      context: prometheus
```
aapeliv: Needs to be an `image: ` in prod, no building there

app/prometheus/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM prom/prometheus
+ADD prometheus.yml /etc/prometheus/
+ADD entrypoint.sh /etc/prometheus
```
aapeliv: Inconsistency in trailing slashes

app/prometheus/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM prom/prometheus
```
aapeliv: We should probably pin to at least a major version here?

app/prometheus/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+/bin/sed -i 's/{GRAFANA_CLOUD_TOKEN}/'$GRAFANA_CLOUD_TOKEN'/' /etc/prometheus/prometheus.yml
+/bin/sed -i 's/{PROMEHTEUS_ENVIRONMENT}/'$PROMEHTEUS_ENVIRONMENT'/' /etc/prometheus/prometheus.yml
+/bin/prometheus --config.file=/etc/prometheus/prometheus.yml \
+             --storage.tsdb.path=/prometheus \
+             --web.console.libraries=/usr/share/prometheus/console_libraries \
+             --web.console.templates=/usr/share/prometheus/consoles
```
aapeliv, line 7: What are these `web.console` things?

app/prometheus/entrypoint.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+/bin/sed -i 's/{GRAFANA_CLOUD_TOKEN}/'$GRAFANA_CLOUD_TOKEN'/' /etc/prometheus/prometheus.yml
+/bin/sed -i 's/{PROMEHTEUS_ENVIRONMENT}/'$PROMEHTEUS_ENVIRONMENT'/' /etc/prometheus/prometheus.yml
+/bin/prometheus --config.file=/etc/prometheus/prometheus.yml \
+             --storage.tsdb.path=/prometheus \
```
aapeliv, line 5: So just to check: it doesn't matter if prometheus dies and this tsdb gets nuked? The actual data is pushed to grafana(?)

app/prometheus/readme.md
```diff
@@ -0,0 +1,3 @@
+# What is prometheus
+
+Prometheus is a monitoring tool that we use to get data into our dashboards and alerts on [https://couchers.grafana.net/]. Prometheus works by scraping metrics from its data sources. In our case, the app respondes to GET requests on port 8000 and 8001 with a list of numbers and tags. Prometheus regularly calls these endpoints and then forwards the data to a centralized database we use (for FREE) at Grafana Cloud. From there we can use the data for monitoring and alerting.
```
aapeliv: responds*

Not sure if it's necessary to tout the freeness... It won't be free for very long, it's not like we get some special treatment here?

app/backend/src/tests/test_interceptors.py
```diff
@@ -57,6 +58,21 @@ def interceptor_dummy_api(
             server.stop(None).wait()
 
 
+def _get_count_from_histogram():
+    metrics = servicer_duration_histogram.collect()
+    servicer_histogram = next(m for m in metrics if m.name == "servicer_duration")
+    return next(s for s in servicer_histogram.samples if s.name == "servicer_duration_count")
```
aapeliv: Are these `next()`s the most obvious way of doing this? Are you basically just doing `[comprehension][0]` here?

aapeliv (changes requested):
This is looking pretty clean and good. Please also run `isort . && black .` in backend to fix formatting, and see a few comments + questions

app/prometheus/Dockerfile
aapeliv: Also, I'd add some extra spaces for formatting here, up to you

lennartvanlaake: added newlines and removed first trailing slash

app/prometheus/entrypoint.sh
lennartvanlaake: Copied from the original entrypoint of the prom/prometheus image. Not sure what they do, but it is what image would otherwise do if we did not have to make our own.

app/prometheus/entrypoint.sh
lennartvanlaake: That is correct. The full path is: application hosts metrics on port 8000/8001 -> scraped by our prometheus -> our prometheus forwards them to the prometheus instance hosted by the folks at Grafana labs -> the grafana application in our Grafana labs setup shows graphs from the Grafana labs central prometheus.

app/backend/src/tests/test_interceptors.py
lennartvanlaake: Yeah, I'm not in love with they python syntax here (I am used to being able to do` listOfThings.first( lambda )` ). Happy to change it to `[comprehension][0]`.

app/prometheus/readme.md
lennartvanlaake: Will fix the typo and remove the freeness thing. Not relevant to the readme. 

The setup is claimed to be 'free forever' within (reasonably generous) usage limits.

app/docker-compose.prod.yml
lennartvanlaake: good find, fixing this

lennartvanlaake:
> This is looking pretty clean and good. Please also run `isort . && black .` in backend to fix formatting, and see a few comments + questions

Sorry, forgot the formatting run. Fixes are pushed.

app/prometheus/Dockerfile
lennartvanlaake: Fixed, they did not have separate major version tags, so pinned to the current latest exact version.

app/docker-compose.prod.yml
```diff
@@ -60,7 +61,17 @@ services:
     networks:
       - media_nginx
       - envoy_nginx
+  prometheus:
+    image: registry.gitlab.com/couchers/couchers/prometheus
+    env_file: prometheus.prod.env
```
aapeliv, line 67: needs a restart always

aapeliv:
How do we differentiate between prod and staging in grafana?

lennartvanlaake:
> How do we differentiate between prod and staging in grafana?

Setting the PROMETHEUS_ENVIRONMENT to a different value. That value will be added as a label to all metrics in that environment so we can separate them in our dashboards and alerts.

app/docker-compose.prod.yml
lennartvanlaake: Fixing now

lennartvanlaake: fixed

aapeliv (approved):
LGTM

PR #1351, opened 2021-06-03
Backend/feature/1035 setup sentry be

lennartvanlaake:
PR is draft because it builds upon on the prometheus PR (both make changes to the same sections of the code)

Adds exception logging in sentry to the backend process.
NB: for sentry logging to take effect on production the following enironment variables need to be added to `backend.prod.env`:

- SENTRY_ENABLED=1
- SENTRY_URL=<URL I sent to @aapeliv via Slack>

- [ x] Formatted my code by running `isort . && black .` in `app/backend`
- [ -] Added tests for any new code or added a regression test if fixing a bug
- [x] All tests pass
- [ x] Run the backend locally and it works


app/backend/src/app.py
```diff
@@ -60,7 +61,12 @@
 
 logging.getLogger("couchers.jobs.worker").setLevel(logging.INFO)
 
-start_http_server(port = 8000, registry = main_process_registry)
+if config.config["SENTRY_ENABLED"]:
+    """Sends exception tracebacks to Sentry, a cloud service for collecting exceptions"""
+    sentry_sdk.init(config.config["SENTRY_URL"], traces_sample_rate=0.0)
```
aapeliv, line 66: What's the `traces_sample_rate`?

app/backend/src/app.py
```diff
@@ -69,6 +75,7 @@ def log_unhandled_exception(exc_type, exc_value, exc_traceback):
         # call the default excepthook saved at __excepthook__
         sys.__excepthook__(exc_type, exc_value, exc_traceback)
         return
+    sentry_sdk.capture_exception(exc_value)
```
aapeliv: Are you sure exc_value is the right thing to put in here?

app/backend/src/app.py
```diff
@@ -157,6 +164,7 @@ def log_unhandled_exception(exc_type, exc_value, exc_traceback):
 if config.config["ROLE"] in ["worker", "all"]:
     worker = start_jobs_worker()
 
+
```
aapeliv: Formatting

app/backend/src/couchers/config.py
```diff
@@ -45,6 +45,9 @@
     ("BUG_TOOL_GITHUB_REPO", str),
     ("BUG_TOOL_GITHUB_USERNAME", str),
     ("BUG_TOOL_GITHUB_TOKEN", str),
+    # Sentry
+    ("SENTRY_ENABLED", bool, "0"),
+    ("SENTRY_URL", str, ""),
```
aapeliv: Better to not have default in here and just add it to the `backend.*.env`s like others?

aapeliv (commented):
Looks good, few small comments

app/backend/src/app.py
lennartvanlaake: It is a setting for sending traces to Sentry for their performance monitoring. We already have our own traces and monitoring in grafana, so I would suggest not using that feature for now and setting it to 0. We would also probably generated too many traces to fit in the free tier, so that would make the Sentry performance monitoring somewhat useless.

app/backend/src/app.py
lennartvanlaake: I thought that was the exception itself. I will check now to see if we might be able to skip this completely, as Sentry itself also generates an unhandled exception hook.

app/backend/src/couchers/config.py
lennartvanlaake: I think having defaults is nicer, as you prevent code from breaking if we miss this value in one of the .env files. In this case the default is also sensible semantically, since if nothing is configured, it should just not use Sentry.

app/backend/src/app.py
lennartvanlaake: Checked, don't need this bit of code, removed it.

app/backend/src/app.py
```diff
@@ -60,7 +61,12 @@
 
 logging.getLogger("couchers.jobs.worker").setLevel(logging.INFO)
 
+if config.config["SENTRY_ENABLED"]:
+    """Sends exception tracebacks to Sentry, a cloud service for collecting exceptions"""
```
aapeliv: These should be comments, not docstrings

app/backend/src/app.py
```diff
@@ -60,7 +61,12 @@
...
+    sentry_sdk.init(config.config["SENTRY_URL"], traces_sample_rate=0.0)
+
 start_http_server(port=8000, registry=main_process_registry)
+"""Starts a prometheus metrics endpoint on port 8000"""
```
aapeliv: Ditto, but also you probably want this above the line?

aapeliv (commented):
Sorry, missed a few tiny bits

app/backend/src/couchers/config.py
aapeliv: The original thinking was to have basically all config options mandatory: that way you are forced to make sure they're all set correctly, config files will have the whole thing there, and the app will just fail on startup if everything isn't set up correctly. There are two main exceptions to this: `VERSION` and `IN_TEST` because they have special handling.

I guess we could move to having sensible defaults for most config options, but I think it'd be better to stick to one convention and stay consistent.

app/backend/src/app.py
lennartvanlaake: was my attempt to follow the google docstring format, will change back to regular comments

app/backend/src/couchers/config.py
lennartvanlaake: cool, changed it. But beware: not setting these prometheus variables in the environment will crash the app now.

aapeliv (approved):
LGTM

app/backend/src/couchers/config.py
aapeliv: There's some env vars in the test config too btw, maybe these should be there too

PR #1353, opened 2021-06-03
[WIP] Combining into one ConfirmChangeEmail

ShmuelTreiger:
@aapeliv Do you want something like this?

(Tests aren't passing, just did this quickly)

[Skip CI]

### DIFF
--- a/app/backend/src/couchers/servicers/auth.py
+++ b/app/backend/src/couchers/servicers/auth.py
@@ -424,18 +424,30 @@ class Auth(auth_pb2_grpc.AuthServicer):
             else:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
+    def ConfirmChangeEmail(self, request, context):
+        with session_scope() as session:
+            query = session.query(User).filter(User.id == context.user_id)
+            if (
+                query.filter(User.old_email_token == request.change_email_token)
+                .filter(User.old_email_token_created <= func.now())
+                .filter(User.old_email_token_expiry >= func.now())
+                .one()
+            ):
+                return self.ConfirmChangeEmailWithOldAddress(request, context)
+            elif (
+                query.filter(User.new_email_token == request.change_email_token)
+                .filter(User.new_email_token_created <= func.now())
+                .filter(User.new_email_token_expiry >= func.now())
+                .one()
+            ):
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/servicers/auth.py
```diff
@@ -424,19 +424,31 @@ def CompletePasswordReset(self, request, context):
...
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
-    # Only called if user doesn't have password
-    def ConfirmChangeEmailWithOldAddress(self, request, context):
+    def ConfirmChangeEmail(self, request, context):
         with session_scope() as session:
-            user = (
-                session.query(User)
-                .filter(User.old_email_token == request.change_email_token)
+            query = session.query(User).filter(User.id == context.user_id)
```
aapeliv: The user should be able to verify the token without logging in (e.g. did something on computer and then clicks on email on the phone).

In general in open servicers (like auth is), we don't have `context.user_id` defined.

app/backend/src/couchers/servicers/auth.py
```diff
@@ -424,19 +424,31 @@ def CompletePasswordReset(self, request, context):
...
+                .filter(User.new_email_token_created <= func.now())
+                .filter(User.new_email_token_expiry >= func.now())
+                .one()
+            ):
+                return self.ConfirmChangeEmailWithNewAddress(request, context)
+            else:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
 
+    # Only called if user doesn't have password
+    def ConfirmChangeEmailWithOldAddress(self, request, context):
```
aapeliv: You should remove this, right?

aapeliv (commented):
Yeah, this is what I was thinking in terms of API

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: Ya, that was one of the bugs which failed

app/backend/src/couchers/servicers/auth.py
ShmuelTreiger: I need to rejigger a few things. Probably reword, just to emphasize it's not called every time

PR #1359, opened 2021-06-04
bugfixes prometheus

lennartvanlaake:


aapeliv (dismissed):
LGTM

app/prometheus/entrypoint.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 /bin/sed -i 's/{GRAFANA_CLOUD_TOKEN}/'$GRAFANA_CLOUD_TOKEN'/' /etc/prometheus/prometheus.yml
```
aapeliv, line 2: Should this also needs curly braces?

aapeliv (approved):

app/prometheus/entrypoint.sh
lennartvanlaake: it should not have curly braces, the PROMETHEUS_ENVIRONMENT one is wrong

lennartvanlaake: (fixed in last commit)

PR #1361, opened 2021-06-05
Cleanup for community v1

### LINKED ISSUE #1339: Community page cleanup before release
- [ ] Add ability to join/leave community
- [ ] Remove event placeholders
- [ ] Remove Pages section
- [ ] Fix more tips link
- [ ] Make "Find host" link to search with correct location filter
- [ ] Change tab order: Overview, Find host, Local info, Discussions, Events, Hangouts (Last two, the page just says Coming soon),

lucaslcode:
closes #1339

**Frontend checklist**
- [ ] All tests pass
  - I need help with the tests :angel:


app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  member = false;
+});
+function View() {
+  const community = useCommunity(2);
+  return community.data ? (
+    <JoinCommunityButton community={community.data} />
+  ) : null;
+}
+
+describe("JoinCommunityButton", () => {
```
lucaslcode, line 32: These don't work. I'm not sure how to properly mock, because the button just takes the `community` object directly, but modifies it by calling the api and invalidating

app/frontend/src/components/Button/Button.tsx
```diff
@@ -64,11 +74,15 @@ export default function Button<D extends ElementType = "button", P = {}>({
...
-        <CircularProgress size={theme.typography.button.fontSize} />
-      ) : (
-        children
+      {(loading || waiting) && (
+        <CircularProgress
+          size={theme.typography.button.fontSize}
+          className={classes.spinner}
+        />
       )}
+      <div className={loading || waiting ? classes.transparent : undefined}>
```
darrenvong, line 83: Use `classNames` for this so you don't need the `undefined` in the end?

Also why does `children` now need to be hidden with `opacity`? Is it for the leave button's white background?

app/frontend/src/features/communities/CommunityPage/LocalInfoSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Community } from "pb/communities_pb";
+import React, { useMemo } from "react";
+import { Link } from "react-router-dom";
+import { routeToCommunity } from "routes";
+import stripMarkdown from "utils/stripMarkdown";
+
+import { useCommunityPageStyles } from "./CommunityPage";
+import SectionTitle from "./SectionTitle";
+
+export default function LocalInfoSection({
```
darrenvong: Guessing this is mostly a placeholder and we don't need this once some of my local info page stuff gets merged?

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default function JoinCommunityButton({
+  community,
+}: {
+  community: Community.AsObject;
+}) {
+  const queryClient = useQueryClient();
+  const join = useMutation<void, Error>(async () => {
+    await service.communities.joinCommunity(community.communityId);
+    await queryClient.refetchQueries(communityKey(community.communityId));
```
darrenvong: Shouldn't we do this in `onSuccess` like we do in the other mutations? And also `invalidate` instead of directly refetching, or is there a good reason to something differently here?

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  });
+  const leave = useMutation<void, Error>(async () => {
+    await service.communities.leaveCommunity(community.communityId);
+    await queryClient.refetchQueries(communityKey(community.communityId));
```
darrenvong: Same as above here

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.test.tsx
darrenvong: Does that mean when the mutation succeeds, some properties in the `community` object updates?

Yeah the mocking gets a bit fiddly - I did something similar in the post reply tests in `DiscussionPage` so that might have the closest example for what you need here, where you need to update the `joinCommunity` mock just before you click the button so the `invalidate`/`refetch` gets the updated result.

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+import community from "test/fixtures/community.json";
+
+import JoinCommunityButton from "./JoinCommunityButton";
+
+export default {
+  component: JoinCommunityButton,
+  title: "Communities/CommunityPage/JoinCommunityButton",
+} as Meta;
+
+const Template: Story<any> = (args) => {
```
darrenvong: Type the `any` here since you are using `args` with `succeed` and `joined` so it's known what the type is?

darrenvong (changes requested):

app/frontend/src/components/Button/Button.tsx
lucaslcode: It's just so the button doesn't resize when it's loading. Less content shift, it was annoying me.

app/frontend/src/features/communities/CommunityPage/LocalInfoSection.tsx
lucaslcode: Yeah

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
lucaslcode: I tried that first, but it's the same old problem where the button stops loading and still displays the old state while it's invalidating. Is there a better way to avoid that than this?

darrenvong: Ohh because by the time line 19 is done, `isLoading` is false? Can we get the query's refetching/loading state out from the query client? Otherwise can't think of anything better without introducing another spinner like I did in other places, which kinda feels a bit suboptimal imo now I think more about this :/

darrenvong: This looks promising - https://react-query.tanstack.com/reference/useIsFetching

Then maybe we can combine that on `isLoading` with another `||` ?

lucaslcode: How about I do an optimistic update in `onSuccess`

lucaslcode: How about I do an optimistic update in `onSuccess`

darrenvong: Hmm yeah that'd also work, since this isn't an infinite query, it's a lot easier to optimistic update and so seems reasonable

darrenvong: Maybe that `isFetching` thing is a good workaround for infinite queries though, good to have looked that up :D

lucaslcode: It might appear every 5 seconds though for the ping! 😱

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    mockConsoleError();
+    joinCommunityMock.mockRejectedValueOnce(Error("generic error"));
+    render(<View />, { wrapper });
+    const joinButton = await screen.findByRole("button", {
+      name: JOIN_COMMUNITY,
+    });
+    userEvent.click(joinButton);
+    const errorAlert = await screen.findByRole("alert");
+    expect(errorAlert).toBeVisible();
+    expect(errorAlert).toHaveTextContent("generic error");
```
lucaslcode, line 81: I tried using the `assertErrorAlert` function and I got weird behaviour - either the test continuously retried, or just crashed with `Cannot read property 'createEvent' of null`

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -15,14 +15,42 @@ export default function JoinCommunityButton({
...
   const queryClient = useQueryClient();
-  const join = useMutation<void, Error>(async () => {
-    await service.communities.joinCommunity(community.communityId);
-    await queryClient.refetchQueries(communityKey(community.communityId));
-  });
-  const leave = useMutation<void, Error>(async () => {
-    await service.communities.leaveCommunity(community.communityId);
-    await queryClient.refetchQueries(communityKey(community.communityId));
-  });
+  const join = useMutation<void, Error>(
```
darrenvong: `Error` type should be from `grpc-web` instead?

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -15,14 +15,42 @@ export default function JoinCommunityButton({
...
+                  ...prevData,
+                  member: true,
+                }
+              : undefined
+        );
+        queryClient.invalidateQueries(communityKey(community.communityId));
+      },
+    }
+  );
+  const leave = useMutation<void, Error>(
```
darrenvong: Same comment as above

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+      onSuccess() {
+        queryClient.setQueryData<Community.AsObject | undefined>(
+          communityKey(community.communityId),
+          (prevData) =>
+            prevData
+              ? {
+                  ...prevData,
+                  member: true,
+                }
+              : undefined
```
darrenvong, line 31:
```suggestion
          (prevData) =>
            ({
                ...prevData,
                member: true,
            })
```
The spread operator is safe to use on `undefined` on objects so don't need this extra check. Formatting on this might be a bit off so might be best to not take the patch.

(Although it errors on arrays which I didn't know before (lol js) - [there's a good Stackoverflow answer on why that's the case](https://stackoverflow.com/questions/47155141/spreading-undefined-in-array-vs-object))

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+      onSuccess() {
+        queryClient.setQueryData<Community.AsObject | undefined>(
+          communityKey(community.communityId),
+          (prevData) =>
+            prevData
+              ? {
+                  ...prevData,
+                  member: false,
+                }
+              : undefined
```
darrenvong, line 49:
```suggestion
          (prevData) =>
             ({
                ...prevData,
                member: false,
              })
```

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      getCommunityMock.mockResolvedValue({
+        ...mockCommunity,
+        member: true,
+      });
+    });
+    leaveCommunityMock.mockImplementation(async () => {
+      getCommunityMock.mockResolvedValue({
+        ...mockCommunity,
+        member: false,
+      });
```
darrenvong, line 48: Nice! 🚀

darrenvong (changes requested):
A couple more minor comments around the button mainly, but looks good mostly!

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.tsx
lucaslcode: I originally tried this and got a type error, because the object might be just {member: true} which is not a Community

darrenvong: Ohh the return type has to be a strictly all or nothing (undefined), fair enough 👍

app/frontend/src/features/communities/CommunityPage/JoinCommunityButton.test.tsx
darrenvong: A bit strange inlining fixes this. Was this a problem on CI too? Anyway happy to leave it for this as the helper function is more for reducing repetition than the fact this is unclear.

darrenvong (approved):
Have we screwed up the data insertion for the global community in the test environment? Even the markdown renderer is showing the text as `\n` rather than putting things into a new paragraph...

PR #1362, opened 2021-06-05
[WIP/MRG] Backend/feat/no password second email confirmation added test

ShmuelTreiger:
Adding a test to check that the tokens successfully expire. Not super thrilled with this solution:
1) I would have preferred to mock func.now(), but couldn't get it to work
2) Would prefer to be able to use now() instead of couchers.utils.now() (and not have to import couchers), but again, couldn't get it to work.

Creating this as a branch to see if anyone else knows how to fix either of the above two issues

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/tests/test_account.py
+++ b/app/backend/src/tests/test_account.py
@@ -442,7 +442,7 @@ def test_ChangeEmail_tokens_expire(db):
         old_email_token = user.old_email_token
         new_email_token = user.new_email_token
 
-    with patch("couchers.utils.now", two_hours_one_minute_in_future):
+    with patch("couchers.servicers.auth.now", two_hours_one_minute_in_future):
         with auth_api_session() as (auth_api, metadata_interceptor):
             with pytest.raises(grpc.RpcError) as e:
                 auth_api.ConfirmChangeEmail(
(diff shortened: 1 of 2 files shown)


app/backend/src/tests/test_account.py
```diff
@@ -422,6 +423,46 @@ def test_ChangeEmail_wrong_token(db, fast_passwords):
...
+                new_email=new_email,
+            )
+        )
+
+    with session_scope() as session:
+        user = session.query(User).filter(User.id == user.id).one()
+        old_email_token = user.old_email_token
+        new_email_token = user.new_email_token
+
+    with patch("couchers.utils.now", two_hours_one_minute_in_future):
```
aapeliv: To fix the import, you should be able to do something like `with patch("couchers.servicers.auth.now")`, and then you can import normally.

aapeliv (commented):
Yes, this is the right way to do these tests.

I've been looking for a while for a way to mock time for the db tests; maybe we should go for something along these lines (we could create a `now` function that works similar to `func.now` in an sqlalchemy context).

Anyway, this is looking good; just fix the imports to be as usual, you should be able to mock it as long as you mock the `now` in `couchers.servicers.auth` (see the other `with patch(...)` blocks in tests).

lucaslcode:
I seem to remember mocking `now` in tests for calculating the user's age.

ShmuelTreiger:
@lucaslcode Did you mean in test_models.py? It looks like you mocked date, not now, but maybe I'm missing something?

Issue #927, opened 2021-04-02
Community info page


lucaslcode:
Pages/guides are postponed for now, except for the main page of a community. It is basically a title, and edit button, and a big body of markdown content.

When approaching this, keep in mind we will want to re-use it to display any page (not just main community pages) in the future.

darrenvong:
Hasn't this been done in #737 and refactored recently in my discussion PRs?

lucaslcode:
No, this is the page with information about the community, the "local info" page

PR #1363, opened 2021-06-05
Community info page

darrenvong:
Half of #927. Community edit page to come in the next PR as this is getting bigger than expected (although a good 1/3 of the changes are tests/Storybook code).

**Desktop**
Screenshot 2021-06-05 at 04 04 17

**Mobile**
Screenshot 2021-06-05 at 04 03 53

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/communities/CommunityInfoPage.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    beforeEach(async () => {
+      listAdminsMock.mockResolvedValue({
+        adminUserIdsList: [2, 3],
+        nextPageToken: "3",
+      });
+      render(<CommunityInfoPage community={community} />, { wrapper });
+      await waitForElementToBeRemoved(screen.getByRole("progressbar"));
+
+      userEvent.click(screen.getByRole("button", { name: SEE_ALL_MODERATORS }));
+      await screen.findByRole("presentation");
```
lucaslcode: Does this (and similar) need to be more specific? Match the title or something?

app/frontend/src/features/communities/CommunityInfoPage.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+      // Close dialog by clicking the backdrop
+      userEvent.click(document.querySelector(".MuiBackdrop-root")!);
+      await waitForElementToBeRemoved(screen.getByRole("presentation"));
+
+      expect(
+        screen.queryByRole("button", { name: LOAD_MORE_MODERATORS })
+      ).not.toBeInTheDocument();
+    });
+
+    it("closes the dialog by pressing the escape key", async () => {
```
lucaslcode, line 169: Isn't this testing mui?

app/frontend/src/features/communities/CommunityInfoPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+  title: {
+    alignItems: "center",
+    display: "flex",
+    "& > * + *": {
+      marginInlineStart: theme.spacing(1),
+    },
+  },
+  content: {
+    "& p": {
+      margin: 0,
```
lucaslcode: Why is this needed here and not the markdown component in general?

app/frontend/src/features/communities/CommunityLeadersSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+        <SectionTitle icon={<CommunityLeadersIcon />}>
+          {COMMUNITY_LEADERS}
+        </SectionTitle>
+        <Typography variant="body1">{COMMUNITY_LEADERS_DESCRIPTION}</Typography>
+        <Typography variant="h3">{COMMUNITY_MODERATORS}</Typography>
+        {error ? (
+          <Alert severity="error">{error.message}</Alert>
+        ) : isLoading ? (
+          <CircularProgress />
+        ) : adminIds && adminIds.length > 0 ? (
```
lucaslcode: Time to stop destructuring react-query? Or you prefer this way?

app/frontend/src/features/communities/CommunityLeadersSection.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: CommunityLeadersSectionProps) {
+  const classes = useStyles();
+  const { adminIds, adminUsers, error, isLoading, hasNextPage } = useListAdmins(
+    community.communityId,
+    "summary"
+  );
+  const [isModeratorsDialogOpen, setIsModeratorsDialogOpen] = useState(false);
+
+  return (
+    <>
```
lucaslcode: Unneeded fragment

app/frontend/src/features/communities/CommunityPage/EventsSection.tsx
```diff
@@ -15,7 +15,7 @@ import { useCommunityPageStyles } from "./CommunityPage";
 import EventCard from "./EventCard";
 import SectionTitle from "./SectionTitle";
 
-export default function PlacesSection({
+export default function EventsSection({
```
lucaslcode, line 18: oops

app/frontend/src/features/communities/constants.ts
```diff
@@ -2,6 +2,7 @@ import { CommunityTab } from "routes";
 
 export const communityTabBarLabels: Record<CommunityTab, string> = {
   overview: "Overview",
+  info: "General",
```
lucaslcode:
```suggestion
  info: "Local info",
```

app/frontend/src/features/communities/constants.ts
```diff
@@ -14,6 +15,10 @@ export const CLOSE = "Close";
 export const COMMENT = "Comment";
 export const COMMENTS = "Comments";
 export const COMMUNITY_HEADING = (name: string) => `Welcome to ${name}!`;
+export const COMMUNITY_LEADERS = "Community leaders";
+export const COMMUNITY_LEADERS_DESCRIPTION =
+  "Below is a list of current leaders within this community.";
```
lucaslcode: Is this from the desktop mockup? "Leaders" sounds pretty weird to me...

lucaslcode (changes requested):

app/frontend/src/features/communities/CommunityInfoPage.test.tsx
darrenvong: I tried and it didn't work, ends up with a cannot find element with `role="presentation"` :( Might be because the dialog itself also a role "none" on it that interferes with the *ByRole stuff...

app/frontend/src/features/communities/CommunityInfoPage.test.tsx
darrenvong: I thought about this, but in some places like the bug report dialog we explicitly disable some things like no clicking background for closing, so I guess this is checking this hasn't been disabled by accident

app/frontend/src/features/communities/CommunityInfoPage.tsx
darrenvong: I haven't checked if this will break in other places and if we want to disable the margin too, my guess is probably yes? I'll move it into `Markdown` if that's the case 👍

app/frontend/src/features/communities/CommunityLeadersSection.tsx
darrenvong: I modified the `useListAdmins` hook slightly so in this case it's returning more than the `listAdmins` query object, and since `adminIds` isn't from the main query object, it doesn't make a difference in this case.

But yeah, definitely worth trying out the type narrowing trick we read on the other post more for these kind of pattern more for a bog standard query object!

app/frontend/src/features/communities/constants.ts
darrenvong: Yeah, and then it has another heading on "Community moderators" right underneath that. Maybe I should move "Community moderators" into the h2 title and remove this description? It's kinda obvious from looking at the UI with what the title says right?

lucaslcode: Yes definitely

app/frontend/src/features/communities/CommunityPage/InfoPageSection.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <SectionTitle icon={<InfoIcon />}>{GENERAL_INFORMATION}</SectionTitle>
+      <Typography className={classes.summaryText} variant="body1">
+        {getContentSummary(community.mainPage?.content)}
+      </Typography>
+
+      <div className={classes.loadMoreButton}>
+        <Link
+          to={routeToCommunity(community.communityId, community.slug, "info")}
+        >
+          <Button component="span">{SEE_MORE_INFORMATION}</Button>
```
lucaslcode: Do you think this makes more sense as an MuiLink? (And I guess also the other "See more" buttons)

darrenvong: Hmm yeah, since it's changing the user's URL (regardless of with or without React Router) so a link makes more sense to me. Maybe I'll start changing it here...

app/frontend/src/features/communities/CommunityInfoPage.test.tsx
darrenvong: Never mind, doing it on role `dialog` works instead, so yes better to match with the title so we know it's the right one 👍

app/frontend/src/components/Markdown.tsx
```diff
@@ -27,7 +27,7 @@ const useStyles = makeStyles((theme) => ({
     "& h4": theme.typography.h4,
     "& h5": theme.typography.h5,
     "& h6": theme.typography.h6,
-    "& p": theme.typography.body1,
+    "& p": { ...theme.typography.body1, margin: 0 },
```
lucaslcode: On second thought, it's better with the paragraph margin. Why did you want to get rid of it in the local info?

lucaslcode (changes requested):
Looks good, one tiny thing

app/frontend/src/components/Markdown.tsx
darrenvong: I thought it made the gap between the paragraph and the title a little too big, and the fact it's closer to the moderators section makes it looks like it belongs more to that section rather than the "general information" heading immediately below.

**With margin**
Screenshot 2021-06-06 at 16 18 16

**Without margin (current code)**
Screenshot 2021-06-06 at 16 18 23

darrenvong: ~~Maybe a better solution is to push the "community moderators" further down?~~ Never mind, pushing the section down doesn't really help after more glancing at it, I think the thing that bothers me is the lack of symmetry between the distance of the paragraph from the page header compared to the Mui parapgraph under "community moderators"

The problem is the page title already has some padding, so that's why having the paragraph also with the margin makes the gap look massive when compared to the normal "body1" paragraph under "community moderators", so actually I think the specific component override is fine? I just checked again and removing the margin makes the markdown paragraph a bit too close to the back button on a discussion page description now...

darrenvong:
By the way if you want to get this in soon, feel free to hit merge if you are happy as I might not get around doing it today

lucaslcode (changes requested):
Failing test

app/frontend/src/components/Markdown.tsx
lucaslcode: I don't mind either way for the headings, but I'm thinking about when you have two paragraphs together, there should be some spacing. If it's set to 0, then it's equivalent to `<br>` and it shouldn't be

lucaslcode: @darrenvong ☝️

lucaslcode (changes requested):
Merge conflicts, and did you see my comment about the `<p>` margin?

darrenvong:
> did you see my comment about the `<p>` margin?

Yes, I ended up adding `& p` selector so it shares the same `marginTop` that we get in other heading component as well so should fix the issue of two paragraphs being right below each other you mentioned? I didn't need to hack around the margin stuff as much in the end by using `SectionTitle` and making it more flexible to accept different heading levels so it's still semantically correct depending on where it's used.

lucaslcode (approved):

PR #1364, opened 2021-06-05
Add schedule to run enforce community memberships

### LINKED ISSUE #1335: Run enforce_community_memberships regularly
Every day in a bg job or something

lennartvanlaake:
Add schedule to run 'enforce_community_memberships' every 15 minutes.
closes #1335

- [ x] Formatted my code by running `isort . && black .` in `app/backend`
- [ x] Added tests for any new code or added a regression test if fixing a bug
- [ x ] All tests pass
- [x ] Run the backend locally and it works

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -1476,7 +1476,7 @@ class BackgroundJobType(enum.Enum):
     # payload: google.protobuf.Empty
     send_request_notifications = 7
     # payload: google.protobuf.Empty
-    enforce_community_membership = 7
+    enforce_community_membership = 8
 
 
 class BackgroundJobState(enum.Enum):
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -1475,6 +1475,8 @@ class BackgroundJobType(enum.Enum):
     add_users_to_email_list = 6
     # payload: google.protobuf.Empty
     send_request_notifications = 7
+    # payload: google.protobuf.Empty
+    enforce_community_membership = 7
 
 
 class BackgroundJobState(enum.Enum):
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -1462,21 +1462,21 @@ class Reply(Base):
 
 class BackgroundJobType(enum.Enum):
     # payload: jobs.SendEmailPayload
-    send_email = 1
+    send_email = enum.auto()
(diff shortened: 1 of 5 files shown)


lennartvanlaake:
For some reason test_db.py on the method pg_dump(), even before adding the migration. I don't really get what is going on there, and it seems not obviously related to the code here. 

Do you @aapeliv have an idea what is happening there? Other branches still work, so doesn't seem to be an issue related to unpinned dependencies. For reference: https://gitlab.com/couchers/couchers/-/jobs/1321192120.

lennartvanlaake:
> For some reason test_db.py on the method pg_dump(), even before adding the migration. I don't really get what is going on there, and it seems not obviously related to the code here.
> 
> Do you @aapeliv have an idea what is happening there? Other branches still work, so doesn't seem to be an issue related to unpinned dependencies. For reference: https://gitlab.com/couchers/couchers/-/jobs/1321192120.

Nevermind, seems to work now for reasons I don't understand :)

app/backend/src/couchers/models.py
```diff
@@ -1475,6 +1475,8 @@ class BackgroundJobType(enum.Enum):
     add_users_to_email_list = 6
     # payload: google.protobuf.Empty
     send_request_notifications = 7
+    # payload: google.protobuf.Empty
+    enforce_community_membership = 7
```
aapeliv: Repeated enum value. Please change all of these to `enum.auto()` as these are categorical, not ordinal. (I think they were left as non-auto due to some merge timing between the PR that added bg jobs and where we switched to auto)

aapeliv:
> For some reason test_db.py on the method pg_dump(), even before adding the migration. I don't really get what is going on there, and it seems not obviously related to the code here.

Was this locally or in CI? Which job on gitlab?

app/backend/src/couchers/models.py
lennartvanlaake: done

aapeliv (approved):
LGTM if tests pass!

lennartvanlaake:
> > For some reason test_db.py on the method pg_dump(), even before adding the migration. I don't really get what is going on there, and it seems not obviously related to the code here.
> 
> Was this locally or in CI? Which job on gitlab?

Broke both locally and in CI. Maybe had something to do with inserting a job with an enum value that didn't exist in the DB? 

PS: I hate SQL enums in case you haven't noticed 👯‍♀️ 👯‍♂️ 👯

PR #1368, opened 2021-06-05
Make search results new-tabbable

### LINKED ISSUE #1227: On map page, there is no right-click option to "open in new tab" the profile listed in the sidebar
Subject: On map page, there is no right-click option to "open in new tab" the profile listed in the sidebar
Description:
I clicked on a pin on the map, then right-clicked on the profile picture to try to open the profile in a new tab (instead of being taken away from the map) but there is no option to open it in a new tab.

Results:
I should have the option to open the profile in a new tab when right-clicking on the profile in the sidebar after clicking on a pin on the map.

Backend version: develop-e5978b3b
Frontend version: develop-e5978b3b
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:88.0) Gecko/20100101 Firefox/88.0
Page: https://app.couchers.org/search/
User (spoofable): jesse (57)

lucaslcode:
closes #1227 

I considered a few options. It is difficult to have both a real anchor link to the profile (not a button, which can't open a new tab), and also be able to go to select the user on the map.

desktop view
mobile view

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/search/SearchResult.tsx
```diff
@@ -1,117 +1,132 @@
...
-                  className={classes.statusLabel}
-                  display="inline"
-                  variant="body1"
-                >
-                  {meetupStatusLabels[user.meetupStatus]}
-                </Typography>
-              </div>
+    <Card id={id} className={className} elevation={highlight ? 4 : undefined}>
+      <CardContent>
+        <UserSummary user={user} avatarIsLink titleIsLink compact>
```
darrenvong: Is there much point in duplicating the link on the avatar as well when the title is now a link to the user's profile? Just thinking for a non-visual user on screen readers, might be a bit annoying to have back-to-back links that both basically do the same thing as they go through each search result. And for visual users, I think the title link is enough and probably better with an easier to find/ larger click area.
```suggestion
        <UserSummary user={user} titleIsLink compact>
```

darrenvong (changes requested):
Screenshot 2021-06-06 at 02 27 23

The "show X on map" button looks a bit cut off to the side of the card on small desktop screens if someone decides to have a really long user name...

Otherwise this is a nice improvement!

darrenvong (dismissed):

darrenvong (approved):

PR #1370, opened 2021-06-05
Fix email rendering

aapeliv:


app/backend/src/couchers/email/__init__.py
```diff
@@ -12,16 +12,25 @@
...
+env.filters["couchers_safe"] = couchers_safe
+
+plain_base_template = env.get_template("email_base_plain.md")
+html_base_template = env.get_template("email_base_html.html")
 
 
-def _escape_html(text):
-    return text.replace("_", "\\_")
+def render_html(text):
+    stripped_paragraphs = text.strip().split("\n\n")
```
lennartvanlaake, line 32: Not immediately obvious to me why we need this, maybe a short doc string or comment here?

lennartvanlaake (approved):

PR #1373, opened 2021-06-06
Implement auth interceptor using protobuf options

aapeliv:
This uses protobuf `option`s to describe the auth level for each service instead of having them as comments and then syncing the comments with the multi-server system.

I know this won't run as is; and it's missing tests. I wanted to seek out opinions before I put a lot of time into tests and fixing the rest of everything to work with this.

Pros:
* `.proto`s are the source of truth for required auth level, which reduces
* Only need to have the server listening on one port instead of one for each auth level
* Much easier to implement new auth levels (such as `admin`, implemented here)
* Can do more complicated auth/interceptor annotations; e.g. per-method, or other per-service/method interceptor things
* Potentially allows frontends to attach their own logic to services depending on annotations

Cons:
* Silly protobuf ~~reflection~~ descriptor stuff
* Need to have a generated `DescriptorFileSet` proto in backend too (but we already do codegen, etc; so I don't think this is big)

I've got to say: I do personally think these protobuf annotations/options are pretty nifty...

Missing:
* ~~Add `is_admin` (superuser) field to `User`~~ removed from this PR
* ~~Fix envoy proxy to use single-port thing~~
* Add tests

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/app.py
+++ b/app/backend/src/app.py
@@ -1,58 +1,16 @@
 import logging
 import signal
 import sys
-from concurrent import futures
 
-import grpc
 import sentry_sdk
 from prometheus_client import start_http_server
 
 from couchers import config
 from couchers.db import apply_migrations, session_scope
-from couchers.interceptors import ErrorSanitizationInterceptor, TracingInterceptor
 from couchers.jobs.worker import start_jobs_scheduler, start_jobs_worker
 from couchers.metrics import main_process_registry
-from couchers.servicers.account import Account
-from couchers.servicers.api import API
-from couchers.servicers.auth import Auth
-from couchers.servicers.blocking import Blocking
-from couchers.servicers.bugs import Bugs
-from couchers.servicers.communities import Communities
-from couchers.servicers.conversations import Conversations
-from couchers.servicers.discussions import Discussions
-from couchers.servicers.events import Events
-from couchers.servicers.gis import GIS
-from couchers.servicers.groups import Groups
-from couchers.servicers.jail import Jail
-from couchers.servicers.media import Media, get_media_auth_interceptor
-from couchers.servicers.pages import Pages
-from couchers.servicers.references import References
-from couchers.servicers.requests import Requests
-from couchers.servicers.resources import Resources
-from couchers.servicers.search import Search
-from couchers.servicers.threads import Threads
+from couchers.server import create_main_server, create_media_server
 from dummy_data import add_dummy_data
-from proto import (
-    account_pb2_grpc,
-    api_pb2_grpc,
-    auth_pb2_grpc,
-    blocking_pb2_grpc,
-    bugs_pb2_grpc,
-    communities_pb2_grpc,
-    conversations_pb2_grpc,
-    discussions_pb2_grpc,
-    events_pb2_grpc,
-    gis_pb2_grpc,
-    groups_pb2_grpc,
-    jail_pb2_grpc,
-    media_pb2_grpc,
-    pages_pb2_grpc,
-    references_pb2_grpc,
-    requests_pb2_grpc,
-    resources_pb2_grpc,
-    search_pb2_grpc,
-    threads_pb2_grpc,
-)
 
 config.check_config()
 
@@ -97,46 +55,11 @@ if config.config["ADD_DUMMY_DATA"]:
 logger.info(f"Starting")
 
 if config.config["ROLE"] in ["api", "all"]:
-    server = grpc.server(
-        futures.ThreadPoolExecutor(64),
-        interceptors=[
-            ErrorSanitizationInterceptor(),
-            TracingInterceptor(),
-            AuthValidatorInterceptor(),
-        ],
-    )
-    server.add_insecure_port("[::]:1751")
-
-    account_pb2_grpc.add_AccountServicer_to_server(Account(), server)
-    api_pb2_grpc.add_APIServicer_to_server(API(), server)
-    auth_pb2_grpc.add_AuthServicer_to_server(Auth(), open_server)
-    blocking_pb2_grpc.add_BlockingServicer_to_server(Blocking(), server)
-    bugs_pb2_grpc.add_BugsServicer_to_server(Bugs(), open_server)
-    communities_pb2_grpc.add_CommunitiesServicer_to_server(Communities(), server)
-    conversations_pb2_grpc.add_ConversationsServicer_to_server(Conversations(), server)
-    discussions_pb2_grpc.add_DiscussionsServicer_to_server(Discussions(), server)
-    events_pb2_grpc.add_EventsServicer_to_server(Events(), server)
-    gis_pb2_grpc.add_GISServicer_to_server(GIS(), server)
-    groups_pb2_grpc.add_GroupsServicer_to_server(Groups(), server)
-    jail_pb2_grpc.add_JailServicer_to_server(Jail(), jailed_server)
-    pages_pb2_grpc.add_PagesServicer_to_server(Pages(), server)
-    references_pb2_grpc.add_ReferencesServicer_to_server(References(), server)
-    requests_pb2_grpc.add_RequestsServicer_to_server(Requests(), server)
-    resources_pb2_grpc.add_ResourcesServicer_to_server(Resources(), open_server)
-    search_pb2_grpc.add_SearchServicer_to_server(Search(), server)
-    threads_pb2_grpc.add_ThreadsServicer_to_server(Threads(), server)
-
+    server = create_main_server(port=1751)
     server.start()
-
-    media_server = grpc.server(
-        futures.ThreadPoolExecutor(8),
-        interceptors=[TracingInterceptor(), get_media_auth_interceptor(config.config["MEDIA_SERVER_BEARER_TOKEN"])],
-    )
-    media_server.add_insecure_port("[::]:1753")
-    media_pb2_grpc.add_MediaServicer_to_server(Media(), media_server)
+    media_server = create_media_server(port=1753)
     media_server.start()
-
-    logger.info(f"Serving on 1751 (secure), 1752 (auth), 1753 (media), and 1754 (jailed)")
+    logger.info(f"Serving on 1751 (secure) and 1753 (media)")
 
 if config.config["ROLE"] in ["scheduler", "all"]:
     scheduler = start_jobs_scheduler()
--- a/app/backend/src/app.py
+++ b/app/backend/src/app.py
@@ -97,51 +97,32 @@ if config.config["ADD_DUMMY_DATA"]:
 logger.info(f"Starting")
 
 if config.config["ROLE"] in ["api", "all"]:
-    auth = Auth()
-    open_server = grpc.server(
-        futures.ThreadPoolExecutor(8), interceptors=[ErrorSanitizationInterceptor(), TracingInterceptor()]
-    )
-    open_server.add_insecure_port("[::]:1752")
-    auth_pb2_grpc.add_AuthServicer_to_server(auth, open_server)
-    bugs_pb2_grpc.add_BugsServicer_to_server(Bugs(), open_server)
-    resources_pb2_grpc.add_ResourcesServicer_to_server(Resources(), open_server)
-    open_server.start()
-
-    jailed_server = grpc.server(
-        futures.ThreadPoolExecutor(8),
-        interceptors=[
-            ErrorSanitizationInterceptor(),
-            TracingInterceptor(),
-            auth.get_auth_interceptor(allow_jailed=True),
-        ],
-    )
-    jailed_server.add_insecure_port("[::]:1754")
-    jail_pb2_grpc.add_JailServicer_to_server(Jail(), jailed_server)
-    jailed_server.start()
-
-    servicer = API()
     server = grpc.server(
         futures.ThreadPoolExecutor(64),
         interceptors=[
             ErrorSanitizationInterceptor(),
             TracingInterceptor(),
-            auth.get_auth_interceptor(allow_jailed=False),
+            AuthValidatorInterceptor(),
         ],
     )
     server.add_insecure_port("[::]:1751")
 
     account_pb2_grpc.add_AccountServicer_to_server(Account(), server)
-    api_pb2_grpc.add_APIServicer_to_server(servicer, server)
+    api_pb2_grpc.add_APIServicer_to_server(API(), server)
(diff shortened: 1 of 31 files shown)


lennartvanlaake (commented):
Big fan, love it :)

kalvdans:
Looks like the way to go!

Maybe do the admin in a later PR, so that this one is a pure refactoring step.

Tell me when you want to get flooded with code comments :D

aapeliv:
Yep, ready for review now @kalvdans.

Happy to take suggestions for how to test this other than the existing tests.

app/backend/src/couchers/interceptors.py
```diff
@@ -34,26 +70,43 @@ class AuthValidatorInterceptor(grpc.ServerInterceptor):
...
-            return unauthenticated_handler()
+        assert auth_level in [
+            annotations_pb2.AUTH_LEVEL_OPEN,
+            annotations_pb2.AUTH_LEVEL_JAILED,
+            annotations_pb2.AUTH_LEVEL_SECURE,
+        ]
+
+        token = parse_session_cookie(dict(handler_call_details.invocation_metadata))
+
+        res = _try_get_and_update_user_details(token)
```
kalvdans, line 96: Since I'm allergic to "try" or "check" functions, I'd like to get this function reworked in the following ways:

1. Don't do any database lookup when using the open API yet providing a token.
2. Fail directly by returning the unauthenticated_handler() instead of proxying it first returning None from _try_get_and_update_user_details.

The easiest is probably to inline _try_get_and_update_user_details() here and try to break out some other part.

aapeliv: I would like the token/user data to be validated even on open APIs, so we can make a (verified) `user_id` available in the context.

This is useful in many cases; currently for example, the bug reporter has the frontend relay the user id (in the request) to the backend which is attached to bug reports. I would much rather have this be checked so we could be sure the user wasn't trivially spoofed.

That being said, I do think the code could be reworked. I'm trying to keep this section easy to read + understand.

kalvdans: It is important to know for the frontend developer which calls it needs to send the auth cookie for. Perhaps we should have another enum value for the "grayzone" methods where the auth cookie makes a difference? Or split those into two different methods?

kalvdans:
> Silly protobuf reflection stuff

I'm trying to wrap my head around it but I can't find any documentation, neither on [protobuf](https://developers.google.com/protocol-buffers/docs/proto3#options) nor [grpc](https://grpc.github.io/grpc/python/)? Where did you learn about user-defined service options?

aapeliv:
> > Silly protobuf reflection stuff
> 
> I'm trying to wrap my head around it but I can't find any documentation, neither on [protobuf](https://developers.google.com/protocol-buffers/docs/proto3#options) nor [grpc](https://grpc.github.io/grpc/python/)? Where did you learn about user-defined service options?

There's some stuff here:

* Proto3 bit on it: https://developers.google.com/protocol-buffers/docs/proto3#customoptions
* Proto2 better bit: https://developers.google.com/protocol-buffers/docs/proto#customoptions
* Edit: also autogenerated docs here: https://googleapis.dev/python/protobuf/latest/

The protobuf (and especially grpc) docs are pretty awful. A lot of the proto docs are for proto2 only, and you have to extrapolate to figure out how to do it in proto3. I also figured out a lot of the actual implementation by reading the source code for protobuf, grpc, envoy, and a bunch of other codebases that use it... It's a bit of a pain, hence the "con".

aapeliv:
Also, it's *not* really "reflection", just using the info in the descriptor protos (edit: the "not")

aapeliv:
Ping @kalvdans: the communities stuff requires this (and then a basic admin API to create a new community) before we can push it out

app/backend/src/couchers/interceptors.py
aapeliv: As discussed in the backend meeting: I think we should send the token always when possible. It's super useful for debugging, and will be in the future in general.

I think we just need to catch the problems here in tests.

app/backend/src/couchers/descriptor_pool.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@functools.cache
+def get_descriptor_pool():
+    """
+    Generates a protocol buffer object descriptor pool which allows looking up info about our proto API, such as options
+    for each servicer, method, or message.
+    """
+    # this needs to be imported so the annotations are available in the generated pool...
+    from proto import annotations_pb2
```
kalvdans: Side-effects on import really sucks, and autoflake wants to remove this unused import. Good with a comment.

app/generate_protos.sh
```diff
@@ -24,6 +24,10 @@ find proto -name '*.proto' | protoc -I proto \
   \
   $(xargs)
 
+# protoc only allows passing --descriptor_set_out once...
+cp proto/protos.pb proxy/protos.pb
+cp proto/protos.pb backend/src/protos.pb
+
```
kalvdans: We should only have a single copy of this file and let each one dig it up using lots of ../../..

Does the proxy even use the file?

aapeliv: > We should only have a single copy of this file and let each one dig it up using lots of ../../..

They go in different containers; and the container build system won't fetch anything from parent folders. I think it's fine. (Another option might be to symlink, but it's not much better in my opinion).

> Does the proxy even use the file?

Yes, it's required for the GIS stuff to return non-protobuf stuff (i.e. the map layers that are GeoJSON). (So that it knows what the input and return types are; though they're always the same in our use case)

kalvdans: Symlinks won't work if a subtree is mounted inside the container. The last cp could be an mv so we get one less. And I think backend/src/proto/descriptors.pb is a more descriptive name.

kalvdans: Maybe I'm blind, but I don't see proxy/protos.pb referenced anywhere? [EDIT] Aha, it was there even before this PR.

aapeliv: > Symlinks won't work if a subtree is mounted inside the container.

Docker is stupid. I don't think it is mounted. From memory I think symlinks work! But I'm not 100% sure.


> I think backend/src/proto/descriptors.pb is a more descriptive name.

I agree, will change it.

aapeliv: > Maybe I'm blind, but I don't see proxy/protos.pb referenced anywhere? [EDIT] Aha, it was there even before this PR.

Yep, it's in `//app/proxy/envoy.yaml`

aapeliv: Btw, I imagine you know this already, but that `protos.pb` (soon to be `descriptors.pb`) is just a serialized `google.protobuf.FileDescriptorSet` message from here `https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto`.

app/backend/src/couchers/interceptors.py
kalvdans: If you stop force-pushing maybe I can help refactor the function.

aapeliv: Okay! Done force pushing for a while. I just had to rebase to update for the other proto changes.

app/generate_protos.sh
kalvdans: I'd like it to end up inside src/proto/ to keep everything proto-related there, and thus it falls under the generic .gitignore as well.

app/backend/src/couchers/interceptors.py
kalvdans: You could merge :) but we've had this discussion before, both ways have its pros and cons.

app/generate_protos.sh
aapeliv: Yes, that's a good idea. I didn't notice, sorry. Want me to fix that or do you have the mutex on this branch now?

app/backend/src/couchers/interceptors.py
aapeliv: I'm a rebase guy 🙃

app/generate_protos.sh
kalvdans: Please fix it :) Nah, we don't need a mutex, should be easy to rebase my stuff even if you force-push, I just need to learn the comands :D

app/backend/src/couchers/interceptors.py
```diff
@@ -34,26 +70,43 @@ class AuthValidatorInterceptor(grpc.ServerInterceptor):
...
+                return unauthenticated_handler()
+            user_id = None
+        else:
+            # a valid user session was found
+            user_id, is_jailed = res
 
-        if not self._allow_jailed and jailed:
-            return unauthenticated_handler("Permission denied")
+            # if the user is jailed and this is isn't a jailed service, fail
+            if is_jailed and auth_level != annotations_pb2.AUTH_LEVEL_JAILED:
```
kalvdans: Jailed users should be able to call the open service right?

app/generate_protos.sh
aapeliv: Fixed (I hope)!

app/backend/src/couchers/interceptors.py
aapeliv: Yes... you've uncovered a heisenbug before it entered prod

aapeliv:
I guess one question here is whether the auth levels are ordinal or not?

If they're ordinal, then we can do increasingly stringent checks to make sure the token (if any) satisfies them.

If they're not; then we should really just basically have a different code path for each to check for each of the different behaviours, or something like that.

kalvdans:
> I guess one question here is whether the auth levels are ordinal or not?

As long as we only have three levels we can manage with comparing for equality or inequality to get all needed subsets of auth levels.

aapeliv:
After this PR I'll add the `admin` level

kalvdans:
> After this PR I'll add the `admin` level

True. I guess we can depend on the numerical values of the enum as long as we document that it matters in annotations.proto .

app/backend/src/couchers/descriptor_pool.py
```diff
@@ -0,0 +1,21 @@
...
+
+    pool = descriptor_pool.DescriptorPool()
+    with open(Path(__file__).parent / ".." / "protos.pb", "rb") as descriptor_set_f:
+        desc = descriptor_pb2.FileDescriptorSet.FromString(descriptor_set_f.read())
+    for file_descriptor in desc.file:
+        pool.Add(file_descriptor)
```
kalvdans, line 20: I get a cryptic error when running test_api.py (without running any other test file first)
```
        pool = descriptor_pool.DescriptorPool()
        with open(Path(__file__).parent / ".." / "proto" / "descriptors.pb", "rb") as descriptor_set_f:
            desc = descriptor_pb2.FileDescriptorSet.FromString(descriptor_set_f.read())
        for file_descriptor in desc.file:
>           pool.Add(file_descriptor)
E           TypeError: Couldn't build proto file into descriptor pool!
E           Invalid proto descriptor for file "annotations.proto":
E             google/protobuf/descriptor.proto: Import "google/protobuf/descriptor.proto" has not been loaded.
E             sensitive: ".google.protobuf.FieldOptions" is not defined.
E             auth_level: ".google.protobuf.ServiceOptions" is not defined.

../couchers/descriptor_pool.py:20: TypeError
```

app/backend/src/couchers/interceptors.py
```diff
@@ -34,26 +70,43 @@ class AuthValidatorInterceptor(grpc.ServerInterceptor):
...
-        if not token:
-            return unauthenticated_handler()
+        service_options = self._pool.FindServiceByName(service_name).GetOptions()
+        auth_level = service_options.Extensions[annotations_pb2.auth_level]
 
-        # None or (user_id, jailed)
-        res = self._get_session_for_token(token=token)
+        # if unknown auth level, then it wasn't set and something's wrong
+        if auth_level == annotations_pb2.AUTH_LEVEL_UNKNOWN:
+            raise Exception("Service not annotated with auth_level")
```
kalvdans, line 86: We should make a unit test that checks that all functions are annotated and remove AUTH_LEVEL_UNKNOWN and this run-time check.

app/backend/src/couchers/interceptors.py
kalvdans: Basically I couldn't find any good way to refactor, so I'll be happy with how it looks now.

app/backend/src/couchers/interceptors.py
aapeliv: Protobuf says that each enum should have an `_UNKNOWN` or equivalent.

Happy to add unit tests, but I'd be more comfortable with this in place just as a sanity check.

app/backend/src/couchers/descriptor_pool.py
aapeliv: That's indeed very cryptic.

lennartvanlaake (approved):

app/backend/src/app.py
```diff
@@ -97,51 +97,32 @@ def log_unhandled_exception(exc_type, exc_value, exc_traceback):
...
-            auth.get_auth_interceptor(allow_jailed=False),
+            AuthValidatorInterceptor(),
         ],
     )
     server.add_insecure_port("[::]:1751")
 
     account_pb2_grpc.add_AccountServicer_to_server(Account(), server)
-    api_pb2_grpc.add_APIServicer_to_server(servicer, server)
+    api_pb2_grpc.add_APIServicer_to_server(API(), server)
+    auth_pb2_grpc.add_AuthServicer_to_server(Auth(), open_server)
```
lennartvanlaake: I am using this PR as a base, and my code crashes on `open_server` not being defined, while being used here.

lennartvanlaake: Auth() also seems to no longer be defined while still being used.

lennartvanlaake (changes requested):

app/backend/src/couchers/descriptor_pool.py
aapeliv: I'm not able to replicate this. Are you sure you generated the protos correctly, including with the `--include_imports` and `--include_source_info` flags?

kalvdans: Thanks for testing! It was indeed the `--include_imports` that was missing from my homemade proto generator script. Resolving!

aapeliv:
Did I miss something or happy to merge @kalvdans ?

kalvdans (approved):
LGTM!

*Ceterum autem censeo grpc-python esse delendam*

PR #1385, opened 2021-06-07
Proto formatting

aapeliv:
...fight!

I don't expect everyone to be setting up clang format, etc; but I can run this occasionally to keep the protos roughly in order.


kalvdans (approved):
Good to have a coding style on the protocol definitions!

PR #1386, opened 2021-06-07
Move protos from //app/pb to //app/proto and fix import paths

aapeliv:
This has been irking me for quite some time. Does two things:

* move protos from `//app/pb` to `//app/proto` like the rest of the world
* fixes the imports in the protos to make sense so they're relative to their folder, and not all prefixed with `pb/`. This requires a bit of import path hacking in python to be able to keep the protos inside `proto/` (see [this eternal mess](https://github.com/protocolbuffers/protobuf/issues/1491))


app/frontend/.gitignore
```diff
@@ -23,9 +23,7 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-src/pb/**
```
darrenvong, line 26: Don't we still need `src/proto/**` instead to stop people accidentally commit the generated js files?

Edit: Oh I see you got a more specific `gitignore` in the folder instead. Do we want to commit that effectively empty folder (apart from the gitignore file) considering we've been leaving it out all along?

aapeliv: We need to include the folder, because `protoc` won't generate it now; or we could add it to the top of the script. (But the python side needs a script in that folder, so for consistency I'd prefer to have them all)

darrenvong (dismissed):
Thank god for TypeScript when you seem to like a rippling change every now and then. Trying to stop us catching up to your million lines diff count? 😛  jk, it builds, and it looks like only import updates so looks fine to me on frontend.

I guess up to you if you want to wait for @kalvdans's backend approval first before merging

kalvdans:
>     * This requires a bit of import path hacking in python

Not fond of that hack, earlier system was better. I like the rename from pb to proto though.

Perhaps we could keep the Python generated code as before, and do the javascript generated code in the new way.

app/generate_protos.sh
```diff
@@ -1,26 +1,27 @@
...
+  --python_out=../media/src/proto \
+  --grpc_python_out=../media/src/proto \
   \
-  --js_out="import_style=commonjs,binary:frontend/src" \
-  --grpc-web_out="import_style=commonjs+dts,mode=grpcweb:frontend/src" \
+  --js_out="import_style=commonjs,binary:../frontend/src/proto" \
+  --grpc-web_out="import_style=commonjs+dts,mode=grpcweb:../frontend/src/proto" \
   \
-  $(xargs)
+  $(xargs) && cd ..
```
kalvdans: Split into two commands, one for Python and one for Javascript. Keep the old style for Python.

aapeliv:
> Not fond of that hack, earlier system was better. I like the rename from pb to proto though.

Fair enough, sure. I'll leave that hack out for the time being and will make this PR do what it says on the tin and not smuggle in this little change in ;).

It's just that the import hack allows importing the protos from protos properly (using e.g. `import "api.proto";` instead of `import "proto/api.proto";`). I find it really annoying and it also screws up my editor's analysis, so I get lots of red squiggly lines everywhere.

aapeliv:
I'm going to paste this here so I find it later if need be; I'll be rewriting the history to squash these commits into one.

```py3
import sys
import os

sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
```

kalvdans (dismissed):
LGTM!

kalvdans:
Sorry to be the party pooper!

> It's just that the import hack allows importing the protos from protos properly (using e.g. `import "api.proto";` instead of `import "proto/api.proto";`). I find it really annoying and it also screws up my editor's analysis, so I get lots of red squiggly lines everywhere.

Ah, somehow I missed that part. Red squiggly lines are never good. Let's make another try after this rename is in.

kalvdans:
> Ah, somehow I missed that part. Red squiggly lines are never good. Let's make another try after this rename is in.

See https://github.com/Couchers-org/couchers/pull/1397

PR #1391, opened 2021-06-08
Update community info page

### LINKED ISSUE #927: Community info page

darrenvong:
Closes #927

**Desktop (community info page)**
With new "edit" link near the top right underneath the tabs if user has permission. Same for mobile.
Screenshot 2021-06-12 at 02 37 36

**Mobile (edit page)**
Screenshot 2021-06-12 at 02 37 21

**Desktop (edit page)**
Screenshot 2021-06-12 at 02 37 06

**Mobile (edit page)**
Screenshot 2021-06-12 at 02 36 53

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/communities/EditCommunityPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+    error,
+    isLoading,
+    isSuccess,
+    mutate: updatePage,
+  } = useMutation<Page.AsObject, GrpcError, UpdatePageData>(
+    ({ content, pageId }) =>
+      service.pages.updatePage({ content, pageId: +pageId }),
+    {
+      onSuccess(newPageData, { communityId, slug }) {
+        queryClient.setQueryData<Community.AsObject | undefined>(
```
darrenvong: Optimistic update in `onSuccess` for some reason doesn't work here and I'm not quite sure what I've done wrong compared to other similar queries... on slow network, even the `history.push` redirect you can see the content is rolling back to the old state in the editor before it redirects you back to the actual page with the new content displaying correctly.

lucaslcode: It works if this is added to `MarkdownInput.tsx`

```
const {
    ref: fieldRef,
    onBlur: fieldOnBlur,
    onChange: fieldOnChange,
    value,
  } = field;

  useEffect(() => (initialDefaultValue.current = value), [value]);
```

The useEffect below this that creates the ToastEditor is triggered when you press the update button, which destroys the editor and re-creates it with the original default value.
It would be more correct to stop `useEffect` being triggered, but I have no idea why it is... I tried memoizing stuff to no avail

darrenvong: Good catch! It looks like `onBlur` and `onChange` from `useController` doesn't have a stable identity and gets re-created on every re-render, and we can't really memoise with `useCallback` since the exhaustive dep rule will just ask `fieldonBlur`/`fieldonChange` to be a dependency which will lead to the same problem as in that `useEffect`...

The other workaround I came up with is storing `fieldOnBlur`/`fieldOnChange` in refs and call that in the `useEffect` that creates the ToastEditor instead. Since neither `fieldOnBlur` nor `fieldOnChange` depends on state, it's probably ok?

darrenvong:
@lucaslcode This basic version of edit page should be ready to be reviewed now

app/frontend/src/components/MarkdownInput.tsx
```diff
@@ -61,16 +61,22 @@ export default function MarkdownInput({
   });
 
   const initialDefaultValue = useRef(defaultValue);
-  const { ref: fieldRef, onBlur: fieldOnBlur, onChange: fieldOnChange } = field;
+  const { ref: fieldRef } = field;
 
   const rootEl = useRef<HTMLDivElement>(null);
+
```
lucaslcode, line 67: Comment here why this was necessary

app/frontend/src/features/communities/constants.ts
```diff
@@ -16,12 +16,15 @@ export const COMMENT = "Comment";
...
 export const COMMUNITY_MODERATORS = "Community moderators";
+export const COMMUNITY_PAGE_UPDATED =
+  "The community page has been successfully updated.";
 export const COMMUNITY_TABS_A11Y_LABEL = "Tabs for community sub-pages";
 export const CREATED_AT = "Created at ";
 export const CREATE_NEW_DISCUSSION_TITLE = "Create new post";
 export const DISCUSSIONS_EMPTY_STATE = "No discussions at the moment.";
 export const DISCUSSIONS_LABEL = "Discussions";
 export const DISCUSSIONS_TITLE = "Discussions";
+export const EDIT_COMMUNITY_PAGE = "Edit community page";
```
lucaslcode: I wonder if this should be Edit local info? Can probably get away with changing this only and not all the components 😆

app/frontend/src/routes.ts
```diff
@@ -127,3 +127,9 @@ export const routeToCommunity = (
   slug: string,
   page?: CommunityTab
 ) => `${communityBaseRoute}/${id}/${slug}${page ? `/${page}` : ""}`;
+export const editCommunityPageRoute = `${communityRoute}/edit`;
```
lucaslcode: With this setup, when editing the page, the "Overview" tab is selected. Would be better if local info was.

lucaslcode (changes requested):
Can you also do one more thing - remove the description from under the title and use that as the General Info section instead (and render it as markdown instead of stripping). 

I don't think using the local info page excerpt will look good, and community mods can just make sure the markdown for the description is appropriate.

app/frontend/src/routes.ts
darrenvong: That's not the case if you look at the screenshot, because `routeToEditCommunityPage` sets the `page` param in the route to "info", so if you click from the link, it works fine.

~~However it doesn't stop people from going `/community/1/world/some-gibberish/edit` and still hitting the edit info page form. Hmm maybe rendering the "not found page" in those cases instead is best~~ Lol ignore this, literally making the route more specific to match on `/community/<id>/<slug>/info/edit` solved this

app/frontend/src/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -11,9 +11,7 @@ import InfoPageSection from "./InfoPageSection";
 
 export const useCommunityPageStyles = makeStyles((theme) => ({
   title: {
-    marginBottom: 0,
-    marginTop: 0,
-    ...theme.typography.h1Large,
+    marginTop: theme.spacing(3),
   },
   description: {
```
lucaslcode: Remove

app/frontend/src/features/communities/CommunityPage/InfoPageSection.tsx
```diff
@@ -1,12 +1,12 @@
...
+import { Link as MuiLink } from "@material-ui/core";
 import { InfoIcon } from "components/Icons";
+import Markdown from "components/Markdown";
 import { Community } from "proto/communities_pb";
 import { Link } from "react-router-dom";
 import { routeToCommunity } from "routes";
 import makeStyles from "utils/makeStyles";
 
 import { GENERAL_INFORMATION, SEE_MORE_INFORMATION } from "../constants";
-import getContentSummary from "../getContentSummary";
```
lucaslcode, line 9: Is this still used anywhere?

lucaslcode (changes requested):
Nits

app/frontend/src/features/communities/CommunityPage/InfoPageSection.tsx
darrenvong: Yeah looks like this might be unnecessary now as `PlaceCard` uses the `LineEllipsis` for doing the ...

I'll remove it!

darrenvong: Oh wait, discussion still uses it

lucaslcode (approved):

PR #1392, opened 2021-06-08
[MRG] Small changes

ShmuelTreiger:
Fixing a typo, plus attempting to clarify some documentation

### DIFF
--- a/app/backend/src/tests/test_db.py
+++ b/app/backend/src/tests/test_db.py
@@ -162,7 +162,7 @@ def test_migrations(testconfig):
 
     from_scratch = pg_dump()
 
-    def massage(s):
+    def message(s):
         s = sort_pg_dump_output(s)
 
         # filter out alembic tables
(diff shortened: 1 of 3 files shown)


ItsiW (approved):
Good changes, thanks for looking over it

app/backend/src/tests/test_db.py
```diff
@@ -162,7 +162,7 @@ def test_migrations(testconfig):
 
     from_scratch = pg_dump()
 
-    def massage(s):
+    def message(s):
```
kalvdans, line 165: This was not a typo, but I realised I could have used a better verb.

ShmuelTreiger: Oof 🤦‍♂️ Sorry about that. Happy to fix unless you have already, just lmk

kalvdans: Nothing to worry about :)

PR #1394, opened 2021-06-08
[MRG] Adapted test I previously wrote to Aapeli's notification processor

ShmuelTreiger:
I'd already written a test when I started working on the notification emails, so I just adapted it to what Aapeli came up with

**Backend checklist**
- [x] All tests pass


aapeliv (approved):
LGTM if tests pass

PR #1397, opened 2021-06-09
Use relative imports in protocol definitions

kalvdans:
Fixup the broken generated code in a postprocessing step.

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/generate_protos.sh
+++ b/app/generate_protos.sh
@@ -1,6 +1,11 @@
 #!/bin/sh
 set -e
 
+# create the directories if they don't exist
+mkdir -p backend/src/proto/
+mkdir -p media/src/proto/
+mkdir -p frontend/src/proto/
+
 # generate API protos and grpc stuff
 find proto -name '*.proto' | protoc -I proto \
   --plugin=protoc-gen-grpc_python=$(which grpc_python_plugin) \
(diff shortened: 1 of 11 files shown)


kalvdans:
Since I'm rolling my own generate_protos.sh I need help to adapt the one in the repository. (don't ask why, long story)

aapeliv:
Thanks! This is a great start; I'd love to get the proto imports fixed. Some comments:

* If you generate it straight into `.../proto`, then you have to have those folders in the repo already. Won't be an issue for you since they were generated before, but will fail on first run. You'll need to add a `mkdir -p` or then just add an empty folder with a `.gitignore` (I'd prefer this, then the structure is clear before running the script)
* I think you need to also change `.../src` to `.../src/proto` for the other ones, not just python

aapeliv:
@kalvdans: are you happy for me to push into this repo to fix this up for the non-python part? Would you prefer `mkdir -p` or empty dirs with a `.gitkeep`/`.gitignore`; I'd prefer the latter (to have the right directory structure in git, maybe even a readme that says "the contents of this folder are generated by the protocol buffer compiler...").

kalvdans:
> @kalvdans: are you happy for me to push into this repo to fix this up for the non-python part?

Yes please!

> Would you prefer `mkdir -p` or empty dirs with a `.gitkeep`/`.gitignore`; I'd prefer the latter (to have the right directory structure in git, maybe even a readme that says "the contents of this folder are generated by the protocol buffer compiler...").

An `mkdir` will give a easier gitignore expression so I think I prefer that. But I'm not going to complain if you let git create the directory.

aapeliv:
@kalvdans: fixed now; have a look

aapeliv (approved):
Feel free to merge if you're happy with it

PR #1401, opened 2021-06-09
Require python>3.8 instead of 3.9

kalvdans:
It is just silly to depend on a too new Python just to use functools.cache instead of functools.cache_lru. Use a Python that is installed by default in stable distros.

**Backend checklist**
- [x] All tests pass


aapeliv (approved):
LGTM!

PR #1406, opened 2021-06-11
[MRG] Added new pb folders to their respective .gitignores

ShmuelTreiger:
Title


aapeliv:
These are from the old generated code and you should remove the files and folders, see the message on slack!

PR #1408, opened 2021-06-11
Map refinement

### LINKED ISSUE #1206: Map tweaks
New map looks great!

For the next iteration can we add:
- [ ] Button to clear search
- [ ] More obvious highlighting of user card when selecting user icon
- [ ] Re-introduce clusters of people (like (100) symbol if 100 in an area)
- [ ] Add location to user card
- [ ] Render markdown in user card into plain text

### LINKED ISSUE #1228: Exclude users with no location from map centering
Users with no location don't appear on the map but are still returned in search results. That's correct, but they should be excluded from map effects like centering (as their location is 0, 0 which causes strange behaviour). 

Original report:

Subject: Map search does not center map on primary result(s)
Description:
When I search "Sydney" the map moved from the location I was zoomed in on in the USA to the middle of the Indian Ocean.

It appears the map only moves until the result is just located within the frame of the map window, but it does not center the map on the primary result(s). See screenshot below.

Results:
The map should have moved so that the primary search result(s) are located in the center of the map (not on the edge of the map!)

Backend version: develop-e5978b3b
Frontend version: develop-e5978b3b
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:88.0) Gecko/20100101 Firefox/88.0
Page: https://app.couchers.org/search/sydney
User (spoofable): jesse (57)

lucaslcode:
adds clusters back
adds search clear button
don't zoom to results when back is pressed
close #1228 closes #1206

Did it ~~Am also going to~~ change the icon to a format that supports colour change and scaling.

**Frontend checklist**
- [ ] All tests pass


lucaslcode:
Lots of failing tests in things I didn't think I changed...

lucaslcode:
@darrenvong  `SearchResultsList.test.tsx` tests are failing and I haven't been able to figure it out after staring for 45 mins.
The progress bar just doesn't want to disappear. The user hooks are always loading, despite using the same mocks and tests as before this PR when it was working...

app/frontend/src/features/search/SearchBox.tsx
```diff
@@ -16,41 +22,72 @@ export default function SearchBox({
...
                     setIsFiltersOpen(!isFiltersOpen);
                   }}
+                  size="small"
                 >
                   <FilterIcon />
                 </IconButton>
+                <IconButton
+                  aria-label={CLEAR_SEARCH}
+                  onClick={() => {
+                    setValue("query", "", { shouldDirty: true });
```
darrenvong: Does setting the field value and making it "dirty" cause re-renders or something?

app/frontend/src/features/search/SearchPage.tsx
```diff
@@ -169,7 +190,8 @@ export default function SearchPage() {
         //addPlacesToMap(newMap);
         //addGuidesToMap(newMap);
       }
-      addUsersToMap(newMap);
+      //addUsersToMap(newMap);
```
darrenvong: Will we still need this back later if we are swapping this with the clustered user layer instead?

app/frontend/src/features/search/users.ts
```diff
@@ -68,38 +60,39 @@ export const layers: Record<LayerKeys, AnyLayer> = {
...
+      "icon-halo-blur": 2,
     },
-    source: "all-objects",
+    source: "clustered-users",
     type: "symbol",
   },
 };
 
 const addPinImages = (map: MaplibreMap) => {
+  if (map.hasImage("user-pin")) return;
```
darrenvong, line 99: I'm guessing leaving this out would work, but ends up doing a request to get the image still if later on it's found to be loaded already?

What was happening before with the old code? 🤔

app/frontend/src/features/search/users.ts
```diff
@@ -128,24 +121,38 @@ export const addClusteredUsersToMap = (
...
+    map.off("click", layers.clusterLayer.id, zoomCluster);
   }
+  map.removeLayer(layers.clusterLayer.id);
+  map.removeLayer(layers.clusterCountLayer.id);
+  map.removeLayer(layers.unclusteredPointLayer.id);
+  map.removeSource("clustered-users");
+
+  if (ids) {
+    //@ts-ignore - type definition incorrect
+    sources["clustered-users"].filter = ["in", ["get", "id"], ["literal", ids]];
```
darrenvong, line 164: All these Mapbox GL specific config stuff reads like black magic to me 🙈

darrenvong (commented):
Mostly questions tbh, but looks fine to me.

As for the tests, yeah it's a little weird. I managed to fix it by setting the `userSearchMock` in a `beforeEach` with a default success implementation. I'm guessing it's somehow hanging the React Query client when that is not resolving to anything (if no default mock is set)? 🤷  Although that'd be worrying if that actually happens... but I suspect it's a thing specific to test environment only (hopefully)

app/frontend/src/features/search/SearchBox.tsx
lucaslcode: Yes, it re-renders. Although I changed the mode to `onChange` (so clicking apply in the filter dialog would use the search query) so I can probably get rid of this now.

app/frontend/src/features/search/SearchPage.tsx
lucaslcode: No, so I just got rid of it everywhere.

app/frontend/src/features/search/users.ts
lucaslcode: I was getting console warnings about the image already being added to the map. 
It's weird because this function is only called once as far as I could see.

app/frontend/src/features/search/users.ts
lucaslcode: I can put comments in. This is just `ids contains(sourceItem.id)`

darrenvong (approved):

PR #1418, opened 2021-06-13
Fix google/api/... imports

aapeliv:

PR #1419, opened 2021-06-13
Test that gis could import its protobuf files properly

kalvdans:
Tests the issue that was solved by https://github.com/Couchers-org/couchers/pull/1418


aapeliv (approved):
LGTM! Good to go if tests pass. That ended up being simpler than I expected...

PR #1428, opened 2021-06-14
Fix phone verification constraints

kalvdans:
Adds back phone verification constraint accidentally removed before merge of https://github.com/Couchers-org/couchers/pull/1250

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -213,6 +213,14 @@ class User(Base):
     phone_verification_verified = Column(DateTime(timezone=True), nullable=True, server_default=text("NULL"))
     phone_verification_attempts = Column(Integer, nullable=False, server_default=text("0"))
 
+    # Verified phone numbers should be unique
+    Index(
+        "ix_users_unique_phone",
+        phone,
+        unique=True,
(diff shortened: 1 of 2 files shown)


ShmuelTreiger (approved):

app/backend/src/couchers/models.py
```diff
@@ -213,6 +213,14 @@ class User(Base):
     phone_verification_verified = Column(DateTime(timezone=True), nullable=True, server_default=text("NULL"))
     phone_verification_attempts = Column(Integer, nullable=False, server_default=text("0"))
 
+    # Verified phone numbers should be unique
+    Index(
+        "ix_users_unique_phone",
+        phone,
+        unique=True,
+        postgresql_where=phone_verification_verified != None,
+    ),
```
aapeliv, line 222: Should `Index`es be moved out of `__table_args__`? There's a few more instances :S

kalvdans: Yeah, we could move the rest in a new PR. sqlalchemy documentation is sparse on this metaprogramming style of defining tables.

PR #1432, opened 2021-06-14
Backend/feature/1431 monitoring fix up

### LINKED ISSUE #1431: Monitoring fix-up
Both sentry and prometheus work now for monitoring backend. However, we still need a few fixes to make it work as well as it can:

- Sentry errors should be labeled per job/servicer and contain the name of the environment (prod/next)
- Caught exceptions with status code should not be sent to Sentry (this clogs our quota for free Sentry really quickly)
- Prometheus should monitor background jobs
- Servicer timing buckets should make sense for miliseconds (currently geared towards seconds)
- Exception classes do not add any value and should not be logged in prometheus servicer monitoring

lennartvanlaake:
closes #1431

NB: This PR makes prometheus manage multiple processes by having all these processes write things to a directory. The location fo that directory needs to be set in the env var `PROMETHEUS_MULTIPROC_DIR`. This directory is automatically created and emptied, so it can be set as to an arbitrary subdirectory of any dir to which the user has write access.

We could use something like `/tmp/prometheus` for all environments it doens't really matter since that directory will be whiped at every restart of the application.

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/metrics.py
```diff
@@ -11,11 +16,19 @@
 main_process_registry = CollectorRegistry()
 jobs_process_registry = CollectorRegistry()
 
+# ensures the directory used for prometheus multiprocessing exists and is clean
+multiproc_dir = pathlib.Path(config["PROMETHEUS_MULTIPROC_DIR"])
+if multiproc_dir.parent.exists():
+    shutil.rmtree(str(multiproc_dir), ignore_errors=True)
+    pathlib.Path(multiproc_dir).mkdir()
+    multiprocess.MultiProcessCollector(jobs_process_registry)
```
aapeliv: Not a fan of this. `/tmp` won't work on all OSes. Is there any other way to do this other than through the filesystem? If not, maybe there's something convenient, OS-independent in the standard lib to create a temporary directory?

lennartvanlaake: I also don't like this. The problem is that this whole thing is necessary to have multiple separate processes talk to eachother. Because the processes don't have access to shared memory you need to set this var from something outside the process. 

I have checked it and it only works for me when I set an env var before starting the application. Even programatically setting an env var in the application doesn't do the trick. 

Because this whole functionality is not really relevant for local development I thought just not having metrics on jobs would be a decent solution to people running Windows.

Anyhow, tests are failing, so I need to do something about it anyway.

lennartvanlaake:
Don't get why tests are failing, the env variable required to run is not set in the CI env. Will probably have to chat to @aapeliv about this setup anyway, so changing PR to draft for now.

app/backend/src/couchers/metrics.py
lennartvanlaake: Ignore what I said, figured out a way to do it by programatically setting an os env-var in an init function :)

lennartvanlaake: ARGH false alarm, doing it programtically didn't work. Spent the last two hours going into a whirlpool of madness on why this isn't working.

lennartvanlaake: Alright, found a way to manually clean up the http server. Still not 100% sure if this will fix the problem, since I cannot replicate the problem with a lingering server on port 8001 on my machine. The lingering http server is however a known issue that can occur, and the solution with manually killing the server on interrupt should work.

app/backend/src/couchers/interceptors.py
```diff
@@ -155,7 +156,12 @@ def tracing_function(request, context):
                 user_id = getattr(context, "user_id", None)
                 self._store_log(method, code, duration, user_id, request, None, traceback)
                 self._observe_in_histogram(method, code or "", type(e).__name__, duration)
-                sentry_sdk.capture_exception(e)
+
+                if code not in (grpc.StatusCode.NOT_FOUND, grpc.StatusCode.UNAUTHENTICATED):
```
aapeliv: I think we should only send exceptions that were not raised by us in `context.abort`. Those will have `code` be `None` in this bit here. So you could change this line to `if not code`

aapeliv (commented):
Looks good otherwise except one comment

PR #1433, opened 2021-06-14
Fix slider click not setting value

### LINKED ISSUE #1427: Unstable reference value
Subject: Unstable reference value
Description:
I was giving a reference to a host as a surfer. I clicked the 'amazing' icon on the slider. 

Results:
I expected the 'amazing' icon to show up when I got to preview my reference. I saw the 'negative' icon in the preview of my reference.

I then re-did the slider of my reference to set it to 'amazing', and again the preview showed the 'negative' icon. When I re-did the slider again, it showed the 'amazing' icon and I was able to submit my reference.

(this is Lennart from the backend team btw, contact me on slack if you have questions)

Backend version: develop-762ac12b
Frontend version: develop-762ac12b
User Agent: Mozilla/5.0 (X11; CrOS x86_64 13904.41.0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.81 Safari/537.36
Page: https://app.couchers.org/messages#loc=1/0/0
User (spoofable): lennart_the_cyclist (438)

lucaslcode:
closes #1427 
I'm not too sure why this works and not the previous version. Seems like before it was one value behind?

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/components/RatingsSlider/RatingsSlider.tsx
+++ b/app/frontend/src/components/RatingsSlider/RatingsSlider.tsx
@@ -8,7 +8,6 @@ import {
 } from "components/RatingsSlider/constants";
 import { getSliderColor } from "components/RatingsSlider/getSliderColor";
 import SliderLabel from "components/RatingsSlider/SliderLabel";
-import React, { useState } from "react";
 import makeStyles from "utils/makeStyles";
 
 interface ColorProps {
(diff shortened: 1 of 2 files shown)


app/frontend/src/components/RatingsSlider/RatingsSlider.tsx
```diff
@@ -91,16 +89,15 @@ export default function RatingsSlider({ defaultValue, onChange }: SliderProps) {
...
+      value={value}
       min={0}
       max={1}
       step={0.01}
       marks={marks}
       valueLabelDisplay="on"
       valueLabelFormat={(value) => <SliderLabel value={value} />}
       onChange={(event, value) => {
-        typeof value === "number" && setCurrentValue(value);
-        onChange?.(currentValue);
```
darrenvong, line 103: I think this was probably why it's one value behind - probably should be `onChange(value)` instead? But looks like you got something simpler working here and letting React Hook Form manage state seems like a better idea as it manages re-renders more cleverly.

darrenvong (approved):

app/frontend/src/components/RatingsSlider/RatingsSlider.tsx
lucaslcode: Does setting state not set it immediately and mark the component as dirty?

darrenvong: Nope, the state value doesn't change for that closure/particular render the component is in, so would still be referring to the old value. Also the state setting is batched by React so even if it does update, I guess it wouldn't update fast enough for the `onChange` here to read the correct value.

PR #1443, opened 2021-06-16
Release communities

lucaslcode:
Make sure global community is there in production with the correct description set.

Also added a choice of content or bug reporting to report dialog:
image
image

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/AppRoutes.tsx
+++ b/app/frontend/src/AppRoutes.tsx
@@ -183,35 +183,31 @@ export default function AppRoutes() {
       {
         // COMMUNITIES
       }
-      {process.env.REACT_APP_IS_COMMUNITIES_ENABLED && (
-        <>
-          <AppRoute isPrivate exact path={communityRoute}>
-            <CommunityPage />
-          </AppRoute>
-          <AppRoute isPrivate path={editCommunityPageRoute}>
-            <EditCommunityInfoPage />
-          </AppRoute>
+      <AppRoute isPrivate exact path={communityRoute}>
+        <CommunityPage />
+      </AppRoute>
+      <AppRoute isPrivate path={editCommunityPageRoute}>
+        <EditCommunityInfoPage />
+      </AppRoute>
 
-          <AppRoute isPrivate path={newPlaceRoute}>
-            <NewPlacePage />
-          </AppRoute>
-          <AppRoute isPrivate path={placeRoute}>
-            <PagePage pageType={PageType.PAGE_TYPE_PLACE} />
-          </AppRoute>
-          <AppRoute isPrivate path={newGuideRoute}>
-            <NewGuidePage />
-          </AppRoute>
-          <AppRoute isPrivate path={guideRoute}>
-            <PagePage pageType={PageType.PAGE_TYPE_GUIDE} />
-          </AppRoute>
-          <AppRoute isPrivate path={discussionRoute}>
-            <DiscussionPage />
-          </AppRoute>
-          <AppRoute isPrivate path={groupRoute}>
-            <GroupPage />
-          </AppRoute>
-        </>
-      )}
+      <AppRoute isPrivate path={newPlaceRoute}>
+        <NewPlacePage />
+      </AppRoute>
+      <AppRoute isPrivate path={placeRoute}>
+        <PagePage pageType={PageType.PAGE_TYPE_PLACE} />
+      </AppRoute>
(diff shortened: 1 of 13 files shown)


app/frontend/src/AppRoutes.tsx
```diff
@@ -183,35 +183,31 @@ export default function AppRoutes() {
...
+      </AppRoute>
+      <AppRoute isPrivate path={placeRoute}>
+        <PagePage pageType={PageType.PAGE_TYPE_PLACE} />
+      </AppRoute>
+      <AppRoute isPrivate path={newGuideRoute}>
+        <NewGuidePage />
+      </AppRoute>
+      <AppRoute isPrivate path={guideRoute}>
+        <PagePage pageType={PageType.PAGE_TYPE_GUIDE} />
+      </AppRoute>
```
darrenvong: We might still want to keep these disabled in prod? They still got Aapeli's minimal stuff in there with bad styling...

app/frontend/src/AppRoutes.tsx
```diff
@@ -183,35 +183,31 @@ export default function AppRoutes() {
...
+      <AppRoute isPrivate path={discussionRoute}>
+        <DiscussionPage />
+      </AppRoute>
+      <AppRoute isPrivate path={groupRoute}>
+        <GroupPage />
+      </AppRoute>
```
darrenvong: I think this one too

app/frontend/src/features/BugReport.tsx
```diff
@@ -68,6 +85,7 @@ export default function BugReport({
 
   const classes = useStyles();
   const [isOpen, setIsOpen] = useState(false);
+  const [type, setType] = useState<undefined | "bug" | "content">();
```
darrenvong:
```suggestion
  const [type, setType] = useState<"initial" | "bug" | "content">("initial");
```
Probably nitpick, but I think it's clearer to have explicit state value rather than abusing `undefined` as a state value too, similar-ish to React Query's `status` string for representing different possible states.

app/frontend/src/features/BugReport.tsx
```diff
@@ -129,53 +151,95 @@ export default function BugReport({
...
-          </DialogContent>
-          <DialogActions>
-            <Button type="submit" loading={isLoading} onClick={onSubmit}>
-              {SUBMIT}
-            </Button>
-            <Button onClick={handleClose}>{CANCEL}</Button>
-          </DialogActions>
-        </form>
+        <DialogTitle id="bug-reporter">{REPORT}</DialogTitle>
+        {type === undefined ? (
```
darrenvong:
```suggestion
        {type === "initial" ? (
```
If following the above suggestion

darrenvong (changes requested):

darrenvong:
Also I wonder if it's better to rename this to `Report` since it's showing more than bug reports. Tests are failing because of that (since it's a functionality change)

app/frontend/src/AppRoutes.tsx
aapeliv: They still got Aapeli's *bad* stuff in there with *minimal* styling...

FTFY

lucaslcode: There are no links to them though, so does it matter?

darrenvong: We probably don't have enough users yet for it to matter that much, but it doesn't seem like a good idea to me to leave half done stuff accessible to the public either, even if it'd involve some URL hacking...

app/frontend/src/features/ReportButton.tsx
```diff
@@ -0,0 +1,246 @@
...
+      {bug && (
+        <Snackbar severity="success">
+          <>
+            {BUG_REPORT_SUCCESS}
+            <Link href={bug.bugUrl}>{bug.bugId}</Link>.
+          </>
+        </Snackbar>
+      )}
+      <Button
+        aria-label="Report a bug"
```
darrenvong:
```suggestion
        aria-label={REPORT}
```
For small screens so it matches the label we hide and shrinks to the icon only. I'm guessing the label of that button must have changed in between...

darrenvong (changes requested):
One last thing! But that means some find/replace for places where the button is found using the old label...

darrenvong (approved):

PR #1450, opened 2021-06-18
Backend/feature/894 add admin api

### LINKED ISSUE #894: Minimal Admin API
We need an admin (superuser) API. Users table needs a boolean "is_superuser" or similar. Admin APIs are things like creating a new community, etc.

Waiting for a list of minimal absolutely required API calls from @ItsiW.

lennartvanlaake:
Blocked by: 
- me cleaning up the code and adding docs where necessary

closes #894

- [ ] All tests pass
- [] Run the backend locally and it works


app/backend/src/tests/test_fixtures.py
```diff
@@ -559,6 +589,14 @@ def resources_session():
     yield resources_pb2_grpc.ResourcesStub(channel)
 
 
+@contextmanager
+def admin_session(token):
```
lennartvanlaake: note to self: should delete this

app/proxy/envoy.yaml
```diff
@@ -29,7 +29,7 @@ static_resources:
                       in building the next-generation couch surfing platform with us, or if you're
                       just interested in API access, please contact us through https://couchers.org
                       or come help out on GitHub at https://github.com/Couchers-org.
-              - match: { prefix: "/" }
+              - match: { prefix: "/org.couchers.auth" }
```
lennartvanlaake: dunno what happened here, should not have changed

app/backend/src/couchers/servicers/admin.py
```diff
@@ -0,0 +1,60 @@
...
+            except Exception as e:
+                logging.error(f"Error occured while parsing geojson for creating community: {e}")
+                context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_MULTIPOLYGON)
+            else:
+                create_cluster(
+                    session, node.id, request.name, request.description, context.user_id, request.admin_ids, [], True
+                )
+                return community_to_pb(node, context)
+
+    def BlockUser(self, request, context):
```
aapeliv: This should be called `BanUser` I think: blocking is something one user does to another

app/backend/src/couchers/servicers/admin.py
```diff
@@ -0,0 +1,60 @@
...
+
+        with session_scope() as session:
+            try:
+                geom = shape(json.loads(request.geojson))
+                assert geom.type == "MultiPolygon"
+                parent_node_id = request.parent_node_id if request.parent_node_id != 0 else None
+                node = create_node(session, geom, parent_node_id)
+            except Exception as e:
+                logging.error(f"Error occured while parsing geojson for creating community: {e}")
+                context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_MULTIPOLYGON)
```
aapeliv: Catch just the exception you're expecting; and move the other stuff out of the try ... except block.

app/backend/src/couchers/interceptors.py
```diff
@@ -102,7 +103,10 @@ def intercept_service(self, continuation, handler_call_details):
             user_id = None
         else:
             # a valid user session was found
-            user_id, is_jailed = res
+            user_id, is_jailed, is_superuser = res
+
+            if auth_level == annotations_pb2.AUTH_LEVEL_ADMIN and not is_superuser:
+                return unauthenticated_handler("Permission denied", grpc.StatusCode.PERMISSION_DENIED)
```
aapeliv, line 109: Why did you change this to `PERMISSION_DENIED`? I think it's good to keep the interceptor-level status code the same. It's a good idea to give it different text though, maybe "You need to be a superuser to perform that action".

`PERMISSION_DENIED` is often returned by the underlying API. I know in this case the waters get muddled a bit; but I think it's still better to use `UNAUTHENTICATED` iff it's from the interceptor.

app/backend/src/couchers/helpers/clusters.py
```diff
@@ -0,0 +1,73 @@
...
+from couchers.models import Cluster, ClusterRole, ClusterSubscription, Node, Page, PageType, PageVersion, Thread
+
+DEFAULT_PAGE_CONTENT = "There is nothing here yet..."
+DEFAULT_PAGE_TITLE_TEMPLATE = "Main page for the {name} {type}"
+
+
+def create_node(session, geom, parent_node_id):
+    node = Node(geom=from_shape(geom), parent_node_id=parent_node_id)
+    session.add(node)
+    session.commit()
```
aapeliv: You shouldn't `commit` here. You can `flush` if you need to flush the SQL to the db e.g. to get the node's ID (autogenerated by db).

Committing will commit the SQL transaction which might lead to an inconsistent state if you commit halfway through creating an object that consists of multiple db objects (e.g. an orphan node without a main page).

app/backend/src/couchers/helpers/clusters.py
```diff
@@ -0,0 +1,73 @@
...
+):
+    type = "community" if is_community else "group"
+    cluster = Cluster(
+        name=name,
+        description=description,
+        parent_node_id=parent_node_id,
+        is_official_cluster=is_community,
+    )
+    session.add(cluster)
+    session.commit()
```
aapeliv: Ditto

app/backend/src/couchers/helpers/clusters.py
```diff
@@ -0,0 +1,73 @@
...
+    main_page = Page(
+        parent_node=cluster.parent_node,
+        creator_user_id=creator_user_id,
+        owner_cluster=cluster,
+        type=PageType.main_page,
+        thread=Thread(),
+    )
+    session.add(main_page)
+    session.commit()
```
aapeliv: Ditto

app/backend/src/tests/test_admin.py
```diff
@@ -0,0 +1,167 @@
...
+                        name=COMMUNITY_NAME,
+                        slug=COMMUNITY_SLUG,
+                        description=COMMUNITY_DESCRIPTION,
+                        admin_ids=[],
+                        geojson=VALID_GEOJSON_MULTIPOLYGON,
+                    )
+                )
+                community = session.query(Cluster).filter(Cluster.name == COMMUNITY_NAME).one()
+                assert community.description == COMMUNITY_DESCRIPTION
+                assert community.slug == COMMUNITY_SLUG
```
aapeliv: Your testing boilerplate is quite different to the rest of the testing boilerplate. I think it'd be good to stay consistent with the rest of the tests.

app/backend/src/tests/test_admin.py
```diff
@@ -0,0 +1,167 @@
...
+"""
+INVALID_GEOJSON = """
+    {
+        "bla": "blah"
+    }
+"""
+
+
+def _session(token: str):
+    return real_session(token, admin_pb2_grpc.add_AdminServicer_to_server, Admin(), admin_pb2_grpc.AdminStub)
```
aapeliv: `_session` is a confusing name for this, I first thought it was a db session. Why not use `admin_session` or `real_admin_session`?

aapeliv (changes requested):

app/backend/src/couchers/interceptors.py
lucaslcode: Doesn't `UNAUTHENTICATED` cause the frontend to log out?

aapeliv: Or redirect you to the jail page. It depends on the error message.

I'm not sure where we're putting these sudo functions in the frontend: maybe the auth state should track that as another state in addition to jailed so it can show an extra "sudo" tab or similar? Then it'd be good to have the (frontend) interceptor catch that and un-sudo you?

lennartvanlaake: Using UNAUTHENTICATED when you are authenticated, but do not have permissions do to a thing seems very confusing. I would rather move the whole thing to a different interceptor then to return a status code that is functionally incorrect.

app/backend/src/tests/test_admin.py
lennartvanlaake: I don't think it is that structurally different besides replacing "magic"/duplicated strings with constants (which I think is generally a good practice in testing) and using more semantically precise methods to get the tokens needed for the tests we are executing here.

If there are reasons that these things are not good pratices I am happy to change them, otherwise I would not like introducing more "bad" practices because of consistency.

app/backend/src/couchers/interceptors.py
aapeliv: Okay; leave as is!

app/backend/src/tests/test_admin.py
aapeliv: I just mean that normally we have the tests as free functions that do the setup and teardown all in one test. The exception is the community tests that have their own setup etc, but that's because it's quite involved and time consuming so we wanted to reuse the same setup across tests.

lennartvanlaake: Ah, that makes sense, I copied the stetup from the communities tests. Will change it to the setup in one of the others.

app/backend/src/couchers/servicers/admin.py
lennartvanlaake: Will fix

app/backend/src/couchers/helpers/clusters.py
```diff
@@ -0,0 +1,73 @@
...
+                role=ClusterRole.admin,
+            )
+        )
+    for member_id in members_ids:
+        cluster.cluster_subscriptions.append(
+            ClusterSubscription(
+                user_id=member_id,
+                role=ClusterRole.member,
+            )
+        )
```
aapeliv: You can get rid of the `member_ids` here; users can just join the community/group later if they want.

app/backend/src/couchers/servicers/admin.py
```diff
@@ -0,0 +1,59 @@
...
+            user = session.query(User).filter(User.id == request.user_id).one()
+            return admin_pb2.GetUserEmailResponse(user_id=user.id, email=user.email)
+
+    def GetUserEmailByUserName(self, request, context):
+        with session_scope() as session:
+            user = session.query(User).filter(User.username == request.username).one()
+            return admin_pb2.GetUserEmailResponse(user_id=user.id, email=user.email)
+
+    def CreateCommunity(self, request, context):
+
```
aapeliv: tiny nit: extra newline

app/backend/src/couchers/servicers/admin.py
```diff
@@ -0,0 +1,59 @@
...
+                session, node.id, request.name, request.description, context.user_id, request.admin_ids, [], True
+            )
+
+            return community_to_pb(node, context)
+
+    def BanUser(self, request, context):
+        with session_scope() as session:
+            user = session.query(User).filter(User.id == request.user_id).one()
+            user.is_banned = True
+            session.add(user)
```
aapeliv: `session.add(...)` emits an `INSERT ...`, we want an update. SA will automatically do that on commit if you assign to an attribute like here.

app/backend/src/couchers/servicers/admin.py
```diff
@@ -0,0 +1,59 @@
...
+            session.commit()
+            return empty_pb2.Empty()
+
+    def DeleteUser(self, request, context):
+        with session_scope() as session:
+            user = session.query(User).filter(User.id == request.user_id).one()
+            user.is_deleted = True
+            session.add(user)
```
aapeliv: Ditto

app/backend/src/tests/test_admin.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def _cleanup(session):
+    normal_user = _get_normal_user(session)
+    if normal_user:
+        session.delete(normal_user)
+    super_user = _get_super_user(session)
+    if super_user:
+        session.delete(super_user)
+    session.flush()
```
aapeliv: No need to do this: the database is nuked and recreated for each test

app/proto/admin.proto
```diff
@@ -0,0 +1,60 @@
...
+package org.couchers.api.admin;
+
+import "google/protobuf/empty.proto";
+import "annotations.proto";
+import "communities.proto";
+
+service Admin {
+  option (auth_level) = AUTH_LEVEL_ADMIN;
+
+  rpc GetUserEmailByUserName(GetUserEmailByUserNameRequest) returns (GetUserEmailResponse) {
```
aapeliv: Please rename `*Request` and `*Response` to `*Req` and `*Res` for consistency

app/proto/blocking.proto
```diff
@@ -9,7 +9,7 @@ import "annotations.proto";
 service Blocking {
   option (auth_level) = AUTH_LEVEL_SECURE;
 
-  rpc BlockUser(BlockUserReq) returns (google.protobuf.Empty);
+  rpc BanUser(BanUserReq) returns (google.protobuf.Empty);
```
aapeliv: This should be `BlockUser`

aapeliv (changes requested):
See comments. Also, you accidentally checked in `app/proxy/protos.pb`.

PR #1453, opened 2021-06-18
Frontend/bugfix/datepicker tz

### LINKED ISSUE #1068: Won't let me request for current date
Subject: Won't let me request for current date
Description:
 I requested to stay with Aapeli on 12/04/2021, my current time is 2:59pm 12/04/2021 AEST

Results:
I got an error that I have to select a date after the current one. Given that is today, it should have been allowed

Backend version: develop-2cba9bcf
Frontend version: develop-2cba9bcf
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:87.0) Gecko/20100101 Firefox/87.0
Page: https://app.couchers.org/user/aapeli
User (spoofable): itsi (2)

lucaslcode:
closes #1068

This seems to be working but it's impossible to do tests because you can't mock the local timezone.

The problem was `Date.toISOString()` always gives the iso string in UTC time.
`dayjs().format()` gives the local iso string so I changed everything to that.

**Frontend checklist**
- [ ] All tests pass


app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
```diff
@@ -204,8 +205,14 @@ export default function CompleteSignupForm() {
             <Datepicker
               className={authClasses.formField}
               control={control}
-              error={!!errors?.birthdate?.message}
-              helperText={errors?.birthdate?.message ?? " "}
+              error={
+                //@ts-ignore Dayjs type breaks this
```
darrenvong, line 209: What errors do you get with this? Would have thought it'd be ok with this since you are converting the value to boolean with `!!`

app/frontend/src/components/Datepicker.test.tsx
```diff
@@ -37,17 +29,36 @@ const Form = ({ submitForm }: { submitForm: (data: unknown) => void }) => {
...
-        datefield: new Date("2021-03-23"),
-      });
+      expect(date?.format()).toEqual(dayjs("2021-03-23").format());
+    });
+    mockdate.reset();
+  });
+
+  //Doesn't work because it seems impossible to mock the local timezone
+  //so date.format() gives the time in utc
+  it.skip.each`
```
darrenvong: Will using this work? https://www.npmjs.com/package/timezone-mock

app/frontend/src/features/auth/signup/CompleteSignupForm.tsx
lucaslcode: From memory, something like `DeepMap<Dayjs, FieldError> does not contain member 'message'`

app/frontend/src/components/Datepicker.test.tsx
lucaslcode: Didn't work, I think because mockdate already replaces the date object. I can try it without setting a mock date, though, but it will be flakey. Or maybe I can fork it and add support for a mock date as well as a mock tz.

darrenvong: Probably worth pulling this in and for tests that don't need time zone mocking, mock the relevant Date methods with jest directly? They aren't as hard to mock compared to time zones

lucaslcode:
@darrenvong still can't get it working. You're welcome to have a go, otherwise this might have to be untested...

darrenvong:
I managed to fix the tests! You were pretty close, the thing it was failing on was on the `mockDate`, since `new Date('2021-03-20')` (without time) always defaults to UTC no matter what timezone you are in, so the fix was literally adding the time at the end like "00:00" to force it to use the local time zone

darrenvong (approved):
This works when I did another sanity test for real on the browser, so let's get this in once CI passes 🎉

lucaslcode:
Thanks!!

PR #1456, opened 2021-06-20
Map Circle Warning 

JGibbsWork:
Updated signup to include a confirmation dialogue wrapper warning about the circle surrounding the address. Not sure if we want to create a barrier for signing up, but it did seem like a more direct way to solve the problem for new users. Also added text below the map when editing your profile. More than happy to change either accordingly.

**Frontend checklist**
- [ ] All tests pass


lucaslcode:
@AllhandsNoFeet thoughts?

lucaslcode (changes requested):
Seems good to me for now.

Could you change `ConfirmationDialogWrapper.tsx` so it has a prop that lets you choose between "Okay" and "Yes" as the 'accept button'. Like `confirmButtonLabel`

I also noticed that if there's a server error ("you must be 18 yo to sign up"), it is hard to notice because the submit button is at the bottom of a long form.

Could you add this to `signup.tsx`

```
  useEffect(() => {
    if (authState.error) window.scroll({ top: 0, behavior: "smooth" });
  }, [authState.error]);
```

app/frontend/src/components/EditLocationMap.tsx
```diff
@@ -363,6 +364,9 @@ function RadiusSlider({ commit, initialRadius, redrawMap }: RadiusSliderProps) {
   const [radius, setRadius] = useState(initialRadius);
   return (
     <>
+      <Typography color="secondary" gutterBottom={true}>
```
lucaslcode:
```suggestion
      <Typography variant="body2" gutterBottom>
```
I think that's sufficient

JGibbsWork:
Cool on it!

app/frontend/src/components/ConfirmationDialogWrapper.tsx
```diff
@@ -37,7 +39,9 @@ export default function ConfirmationDialogWrapper({
           <DialogContentText>{message}</DialogContentText>
         </DialogContent>
         <DialogActions>
-          <Button onClick={handleConfirm}>Yes</Button>
+          <Button onClick={handleConfirm}>
+            {confirmButtonLabel ? confirmButtonLabel : "Yes"}
```
lucaslcode: Better move the default to a constant (sorry)

lucaslcode (changes requested):
Tiny thing but looks good!

lucaslcode (approved):
👍

aapeliv:
Is this ready to merge? If so, feel free to merge it yourself @JGibbsWork

JGibbsWork:
Sorry about that! All set now!

Issue #903, opened 2021-04-01
SQLAlchemy 2.0 migration/refactor


aapeliv:
SQLAlchemy is making some large changes from 1.3 to 1.4 (a kind of 2.0 migration release). This is causing a bunch of annoying warnings in the console, and we should in general refactor how database queries are done. I.e. read the SQLA docs about the recommended patterns, decide how much to rely on the ORM vs Core and then standardize across the codebase.

ShmuelTreiger:
Warnings taken care of; reading docs, standardizing, and refactoring have not.

ShmuelTreiger:
Here's the guideline for migrating to SQLA 2.0*: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-core-connection-transaction

I'm sure there's more to be done. In particular, I think some of the new/changed autocommit stuff may effect us.

There's also this warning message when you run tests, but I don't understand locations stuff well enough to solve (the other warning message seems safe to ignore):

  /couchers/couchers/app/backend/src/couchers/models.py:277: SAWarning: Multiple rows returned with uselist=False for lazily-loaded attribute 'User.timezone_area'
    return self.timezone_area.tzid if self.timezone_area else "Etc/UTC"
	
That being said, I think I've done what I can here. Someone else is welcome to take over.

*The steps above the specific link location have 100% been taken care of, excepting the warning included here.

aapeliv:
> There's also this warning message when you run tests, but I don't understand locations stuff well enough to solve (the other warning message seems safe to ignore):

I think that's only in the test using test data which has overlapping timezone areas. Shouldn't be an issue in prod.

ShmuelTreiger:
That makes sense. Should we adjust the test? Seems weird we run a test that could never be in prod?

aapeliv:
It would be good to update the sample timezone regions, but it's a fair bit of effort. Basically there is a UTC region that covers the whole world, and then a few low-poly regions that I've drawn just to have some more real ones. It would require cutting out those low-poly regions from the UTC polygon.

ShmuelTreiger:
Low priority issue?

aapeliv:
Yep

aapeliv:
This is done! :tada:

PR #1473, opened 2021-06-22
[MRG] Updating SQLAlchemy

ShmuelTreiger:
Updating our codebase to work with sqlalchemy 1.4 (and 2.0) #903 

This pull request deals with all issues caused by upgrading 1.3 --> 1.4 --> 2.0 (as well as one warning which I believe was there before)

It does not, however, "refactor how database queries are done. I.e. read the SQLA docs about the recommended patterns, decide how much to rely on the ORM vs Core and then standardize across the codebase."

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/servicers/api.py
```diff
@@ -525,12 +526,9 @@ def ListMutualFriends(self, request, context):
                 .filter(FriendRelationship.status == FriendStatus.accepted)
             )
 
-            mutual_friends = (
-                session.query(User)
-                .filter_users(context)
-                .filter(User.id.in_(q1.union(q2).intersect(q3.union(q4)).subquery()))
-                .all()
-            )
+            mutual = select(intersect(union(q1, q2), union(q3, q4)).subquery())
```
ShmuelTreiger, line 528: Both extracting this onto it's own line as well as using intersection and union in this manner are not strictly necessary, but I find it clearer and prettier

Adding the `select(...)` was necessary

app/backend/src/couchers/models.py
```diff
@@ -1581,8 +1591,8 @@ class UserBlock(Base):
     blocked_user_id = Column(ForeignKey("users.id"), nullable=False)
     time_blocked = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
 
-    is_blocking_user = relationship("User", backref="is_blocking_user", foreign_keys="UserBlock.blocking_user_id")
-    is_blocked_user = relationship("User", backref="is_blocked_user", foreign_keys="UserBlock.blocked_user_id")
+    blocking_user = relationship("User", foreign_keys="UserBlock.blocking_user_id")
+    blocked_user = relationship("User", foreign_keys="UserBlock.blocked_user_id")
```
ShmuelTreiger, line 1621: Removed the backref, since it seems to me we don't use user.is_blocking/blocked_user and there's no real need for it (especially since it returns a list of all UserBlocks where the User is the blocking/blocked user, which is, IMO, rather confusing and unwieldly). Trivial to add it back in and reinstantiate the test I removed from test_blocking.py.

app/backend/src/couchers/models.py
```diff
@@ -926,17 +925,22 @@ class Cluster(Base):
         primaryjoin="and_(Cluster.parent_node_id == Node.id, Cluster.is_official_cluster)",
         backref=backref("official_cluster", uselist=False),
         uselist=False,
+        viewonly=True,
```
ShmuelTreiger, line 954: I don't 100% understand why viewonly=True is needed here/fixes the warning

app/backend/src/couchers/models.py
```diff
@@ -896,12 +894,13 @@ class Node(Base):
 
     contained_users = relationship(
         "User",
-        lazy="dynamic",
         primaryjoin="func.ST_Contains(foreign(Node.geom), User.geom).as_comparison(1, 2)",
         viewonly=True,
         uselist=True,
     )
 
+    user_ids = association_proxy("contained_users", "id")
```
ShmuelTreiger: Not 100% certain this works as intended, as it's empty the only time it's tested

ShmuelTreiger (commented):
The biggest cause of warnings caused by upgrading sqlalchemy to 1.4 (before adapting for 2.0) were relationships using association tables (i.e. `secondary="table_name"`). Adding `viewonly=True` to these relationships removes these warnings. If we ever want to be able to use the relationships to add associations [i.e. `discussions.subscribers.append(user_1)`], we will need to create association_proxies with creator parameters.

See documentation here:

https://docs.sqlalchemy.org/en/14/orm/extensions/associationproxy.html#proxying-dictionaries

ShmuelTreiger:
Other shortcomings:

1) ~~I can't figure out how to [turn on RemovedIn20](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings). I don't know if the warnings I dealt with were these warnings or if there's still more to warnings to resolve.~~
Got this working. Only 4150 warnings 🙃

2) Alembic didn't pick up any migrations and it seems test_migrations is passing. I assume this means the changes I made to models are python-exclusive/superficial and don't require any migrating, but it's possible I'm mistaken?

ShmuelTreiger:
~~After figuring out how to run this with warnings turned on, I propose merging, upgrading sqlalchemy, and allowing for slowly closing warnings over time.~~ Warnings dealt with

image

ShmuelTreiger:
Surprisingly easy to close. All warnings and errors dealt with except for:

1) ~~995 warnings all associated with one line of geoalchemy2, so hopefully they fix that soon~~ Just needed update geoalchemy library

2) A warning which I think is mistaken and which it's safe to ignore
[Warnings.txt](https://github.com/Couchers-org/couchers/files/6693079/Warnings.txt)

ShmuelTreiger:
Project updated for SQLAlchemy 2.0 and ready for review and merge

The one warning which I believe is safe to ignore still remains

Final note: It's possible some of the untested stuff needs updating and will fail once used post-merge

app/backend/requirements.txt
```diff
@@ -20,4 +20,4 @@ pyyaml
 requests
 sentry-sdk
 Shapely
-sqlalchemy<1.4
+sqlalchemy>=1.4
```
aapeliv: I feel like this should be `<1.5`

app/backend/src/app.py
```diff
@@ -4,6 +4,7 @@
 
 import sentry_sdk
 from prometheus_client import start_http_server
+from sqlalchemy import text
```
aapeliv: Import it from `sqlalchemy.sql` please

app/backend/src/couchers/resources.py
```diff
@@ -3,6 +3,8 @@
 import logging
 from pathlib import Path
 
+from sqlalchemy import text
```
aapeliv: Same

app/backend/src/couchers/servicers/api.py
```diff
@@ -4,8 +4,9 @@
 
 import grpc
 from google.protobuf import empty_pb2
+from sqlalchemy import intersect, union
```
aapeliv: can you import these from `.sql` too or are they not there?

app/backend/src/couchers/models.py
aapeliv: Call this `contained_user_ids`

app/backend/src/tests/test_fixtures.py
```diff
@@ -7,6 +7,7 @@
 
 import grpc
 import pytest
+from sqlalchemy import text
```
aapeliv: .sql

aapeliv (commented):
It's looking good, see small comments. Also the version pinning

app/backend/requirements.txt
ShmuelTreiger: There are changes which will break on 1.3

aapeliv: So maybe `>=1.4,<1.5` or something?

app/backend/src/couchers/models.py
aapeliv: It's because it has a complicated `primaryjoin` (so it's not controlled by SA: SA isn't able to tell what to insert if you tell it to add something into this relationship)

app/backend/src/couchers/models.py
ShmuelTreiger: Done. Still planning to write a test for it

app/backend/requirements.txt
ShmuelTreiger: So... =1.4? 😂

app/backend/src/couchers/models.py
ShmuelTreiger: That makes sense, thanks!

app/backend/requirements.txt
aapeliv: This ought to also match `1.4.2`, etc

aapeliv: It's the "half open interval [1.4,1.5)" as us mathematicians would call it :P

So including all minor versions of 1.4 but not major 1.5 or higher

aapeliv (approved):
LGTM

PR #1484, opened 2021-06-23
Frontend/feature/email confirmation

aapeliv:
Frontend bit. Also changed the RPC to return an enum to better represent output state.

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
+++ b/app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
@@ -25,12 +25,12 @@ export default {
 
 interface ConfirmChangeEmailArgs {
   isLoading?: boolean;
-  resultState?: EmailConfirmationState | false;
+  resultState?: EmailConfirmationState | null;
 }
 
 const Template: Story<ConfirmChangeEmailArgs> = ({
   isLoading = false,
-  resultState = false,
+  resultState = null,
 } = {}) => {
   setMocks({ isLoading, resultState });
   return <ConfirmChangeEmail />;
@@ -60,7 +60,7 @@ NeedOld.args = {
 
 export const Failed = Template.bind({});
 Failed.args = {
-  resultState: false,
+  resultState: null,
 };
 
 function setMocks({
--- a/app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
+++ b/app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
@@ -1,5 +1,5 @@
 import { Meta, Story } from "@storybook/react";
-import { ConfirmChangeEmailRes } from "proto/auth_pb"
+import { ConfirmChangeEmailRes, EmailConfirmationState } from "proto/auth_pb";
 import { MemoryRouter, Route } from "react-router-dom";
 import { confirmChangeEmailRoute } from "routes";
 import { mockedService } from "stories/serviceMocks";
@@ -25,14 +25,14 @@ export default {
 
 interface ConfirmChangeEmailArgs {
   isLoading?: boolean;
-  shouldSucceed?: boolean;
+  resultState?: EmailConfirmationState | false;
 }
 
 const Template: Story<ConfirmChangeEmailArgs> = ({
   isLoading = false,
-  shouldSucceed = true,
+  resultState = false,
 } = {}) => {
-  setMocks({ isLoading, shouldSucceed });
+  setMocks({ isLoading, resultState });
   return <ConfirmChangeEmail />;
 };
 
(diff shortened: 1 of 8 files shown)


app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
```diff
@@ -26,14 +25,14 @@ export default {
 
 interface ConfirmChangeEmailArgs {
   isLoading?: boolean;
-  shouldSucceed?: boolean;
+  resultState?: EmailConfirmationState | false;
```
aapeliv: Not really sure how to do this

app/frontend/src/features/auth/email/ConfirmChangeEmail.tsx
```diff
@@ -18,25 +20,39 @@ export default function ConfirmChangeEmail() {
...
+        return "Error";
+    }
+  }
 
   return isLoading ? (
     <Typography variant="body1">{CHANGE_EMAIL_PROGRESS}</Typography>
   ) : isSuccess ? (
     <>
-      <Alert severity="success">{CHANGE_EMAIL_SUCCESS}</Alert>
+      <Alert severity="success">{successMsg(data!.state)}</Alert>
```
aapeliv, line 55: Should this not be `success` if it still requires confirming new/old email? Or should it not show the login button?

aapeliv:
@ShmuelTreiger: I changed the return type to an enum to better capture the "one of three" return type. Please review that bit!

The rest is just frontend stuff.

app/proto/auth.proto
```diff
@@ -184,10 +184,13 @@ message ConfirmChangeEmailReq{
   string change_email_token = 1 [ (sensitive) = true ];
 }
 
+enum EmailConfirmationState {
+  EMAIL_CONFIRMATION_STATE_UNKNOWN = 0;
```
ShmuelTreiger, line 188: Is this ever used/when would this be used?

aapeliv: It would show up if the value is missing, which is recommended in the protobuf docs. Protocol buffers don't track presence, so this allows determining if this field is missing

ShmuelTreiger: It's automatic? Got it.

app/frontend/src/features/auth/email/ConfirmChangeEmail.tsx
```diff
@@ -18,25 +20,39 @@ export default function ConfirmChangeEmail() {
...
+
+  function successMsg(state: EmailConfirmationState) {
+    switch (state) {
+      case EmailConfirmationState.EMAIL_CONFIRMATION_STATE_SUCCESS:
+        return CHANGE_EMAIL_SUCCESS;
+      case EmailConfirmationState.EMAIL_CONFIRMATION_STATE_REQUIRES_CONFIRMATION_FROM_NEW_EMAIL:
+        return CHANGE_EMAIL_NEED_NEW;
+      case EmailConfirmationState.EMAIL_CONFIRMATION_STATE_REQUIRES_CONFIRMATION_FROM_OLD_EMAIL:
+        return CHANGE_EMAIL_NEED_OLD;
+      default:
```
darrenvong, line 46: Does the backend ever return `UNKNOWN` in practice (which is when this code path is hit here)? Maybe worth throwing an error here if it does as it'd be exceptional/unexpected and it shouldn't be a success message?

app/frontend/src/features/auth/email/ConfirmChangeEmail.tsx
darrenvong: Maybe hide the login link unless the state is `EMAIL_CONFIRMATION_STATE_SUCCESS`? Otherwise they'd still have to use their old email right?

app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
darrenvong: Yeah, I see this is no longer the straightforward succeed or fail binary state since there are multiple states now. Maybe `null` better represents something has gone wrong (even though this is technically not an object, but with the type name it kinda still reads like one)?

app/frontend/src/features/auth/email/ConfirmChangeEmail.tsx
aapeliv: Yeah, just with `throw Error("Invalid ...")`?

darrenvong: Yup 👍

darrenvong: This is only testable locally I assume since the staging API is not updated until this is merged?

aapeliv: Yeah that's right

app/frontend/src/features/auth/email/ConfirmChangeEmail.stories.tsx
aapeliv: Yeah: I just wasn't sure how to use null :P

darrenvong (approved):
Frontend looks good to me 👍

PR #1485, opened 2021-06-23
The communites are now listed sorted

### LINKED ISSUE #1471: Alphabetize communities list
Subject: Alphabetize communities list
Description:
"Browse all communities"

Results:
The community list will be hard to navigate. It would be easier if it was alphabetized

Backend version: develop-c1e24cd2
Frontend version: develop-c1e24cd2
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:89.0) Gecko/20100101 Firefox/89.0
User (spoofable): itsi (2)

sadikekin:
This PR closes #1471. I think it is more user-friendly to have communities sorted as well. There can be two different solutions to achieve this. One is from the backend and one from the frontend. I decided to go with the backend since it makes sense to give sorted communities to every single frontend request.

We can further discuss it if you have any comments/questions/etc.

**Backend checklist**
- [X] Formatted my code by running `autoflake -r -i --remove-all-unused-imports src && isort . && black .` in `app/backend`
- [X] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [X] Run the backend locally and it works
- [X] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

It looks like the following in my local.

Screenshot 2021-06-23 at 20 37 00


aapeliv (changes requested):
The sorting needs to happen in the SQL query: this is now only sorting within each page.

So e.g. with page size 3, you could get this kind of thing:

Page 1:
Australia
Brisbane
Zimbabwe

Page 2:
Annerley
New York City
New York State

And then in the frontend they wouldn't be sorted anymore, giving you




Australia
Brisbane
Zimbabwe
Annerley
New York City
New York State

sadikekin:
Makes sense

sadikekin:
Hey @aapeliv, The last change should do the trick. 

I want to raise one question. It seems like pagination does not affect the popup section. When I decreased the `MAX_PAGINATION_LENGTH` variable, I got the following. Is this the behavior that we want? 
Screenshot 2021-06-23 at 21 48 54

app/backend/src/couchers/servicers/communities.py
```diff
@@ -99,9 +96,10 @@ def ListCommunities(self, request, context):
             next_node_id = int(request.page_token) if request.page_token else 0
```
aapeliv: The pagination needs to be based on the name as well: this will still sort within each page only

app/backend/src/couchers/servicers/communities.py
```diff
@@ -99,9 +96,10 @@ def ListCommunities(self, request, context):
...
             nodes = (
                 session.query(Node)
+                .join(Cluster)
```
aapeliv: I think you need to `.join(Cluster, Cluster.parent_node_id == Node.id)`

aapeliv (changes requested):

aapeliv:
> I want to raise one question. It seems like pagination does not affect the popup section. When I decreased the `MAX_PAGINATION_LENGTH` variable, I got the following. Is this the behavior that we want?

That might be a frontend bug

sadikekin:
I think the front end of the popup needs some updates. Because, it really doesn't make any sense to show only one page if we have pagination. I am planning to check it.

lucaslcode:
My bad

lucaslcode:
I take it page - pagination is working correctly on the frontend. [Try it here](https://develop--storybook.preview.couchershq.org/?path=/story/dashboard-communitieslist--with-more) or do `req.setPageSize(2)` in `communities.tsx:listCommunities()`

sadikekin:
Hey @aapeliv, What do you think about this version?

aapeliv (changes requested):
Just order by cluster name in the db and filter on cluster.name >= next page token. You can make next page token the name of the next cluster in alphabetical order.

That way you don't have to pull all the communities into the backend

sadikekin:
It sounds great. However, the behavior of the SQLalchmy is not the way I wanted. So, When I change it to the way you described it. I get the following SQL translation. Further, I tried this SQL in the PostgreSQL container and I get the following results.

Screenshot 2021-06-27 at 13 51 15

So, we have `East Coast Hitchhikers`. It is a group. I don't receive it in the python code. So, I say `limit by 2`. However, I receive only one community (aka Australia). I don't know why I receive it in the SQL and I don't know how it gets filtered out in the python code.

In the second technique, I tried the following code.
`
            page_size = request.page_size if request.page_size else MAX_PAGINATION_LENGTH
            next_node_name = request.page_token if request.page_token else ""

            nodes = (
                session.query(Node)
                .select_from(Cluster)
                .join(Node, Cluster.parent_node_id == Node.id)
                .filter(or_(Node.parent_node_id == request.community_id, request.community_id == 0))
                .filter(Cluster.name >= next_node_name)
                .filter(Cluster.is_official_cluster == True)
                .order_by(Cluster.name)
                .limit(page_size + 1)
                .all()
            )

            communities = [community_to_pb(node, context) for node in nodes]
            return communities_pb2.ListCommunitiesRes(
                communities=communities[:page_size],
                next_page_token=str(communities[-1].name) if len(nodes) > page_size else None,
            )
`

At this time, I correctly received the `limit` specification. However, I am not quite sure if this is the right way. 

I have limited knowledge of database configuration for this project. Hence, I am having some trouble.

aapeliv:
> So, we have `East Coast Hitchhikers`. It is a group. I don't receive it in the python code.

That's because of `.filter(Cluster.is_official_cluster == True)` in Python. A group is also a cluster but a non-official cluster.

Your code fragment seems pretty good, what's the problem with that?

sadikekin:
Since I don't know the model interaction, I wasn't sure is using `Cluster.is_official_cluster == True` is a good idea or not. However, it seems like it is working. Do you have any additional comments?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -107,19 +104,25 @@ def GetCommunity(self, request, context):
 
     def ListCommunities(self, request, context):
         with session_scope() as session:
-            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
-            next_node_id = int(request.page_token) if request.page_token else 0
+            page_size = request.page_size if request.page_size else MAX_PAGINATION_LENGTH
```
aapeliv: This should have the minimum

app/backend/src/couchers/servicers/communities.py
```diff
@@ -107,19 +104,25 @@ def GetCommunity(self, request, context):
...
+            next_node_name = request.page_token if request.page_token else ""
```
aapeliv: I think `request.page_token` should default to an empty string, so I don't think this is necessary

app/backend/src/couchers/servicers/communities.py
```diff
@@ -107,19 +104,25 @@ def GetCommunity(self, request, context):
...
+
             nodes = (
                 session.query(Node)
+                .select_from(Cluster)
```
aapeliv: What does `select_from` do? We don't use it elsewhere, so it'd be good to not introduce it here.

aapeliv: You probably want something along the lines of

```py3
session.query(Node, Cluster)
...
```

app/backend/src/couchers/servicers/communities.py
```diff
@@ -107,19 +104,25 @@ def GetCommunity(self, request, context):
...
                 .limit(page_size + 1)
                 .all()
             )
+
+            communities = [community_to_pb(node, context) for node in nodes]
             return communities_pb2.ListCommunitiesRes(
-                communities=[community_to_pb(node, context) for node in nodes[:page_size]],
-                next_page_token=str(nodes[-1].id) if len(nodes) > page_size else None,
+                communities=communities[:page_size],
+                next_page_token=str(communities[-1].name) if len(nodes) > page_size else None,
```
aapeliv: You're turning all communities into the proto format, including the last one, which is just there to figure out if we need to fetch more. It's also not good to use `communities[-1].name` for the name: that's an implementation detail of the proto format, and you should just use the `cluster.name`.

With the snippet I had before, you could do `communities=[community_to_pb(node, context) for node, _ in node_and_clusters[:page_size]]` (assuming you rename `nodes` to `node_and_clusters`) and then `node_and_clusters[-1][1].name if ...` in the second bit (also the cast to `str` is unnecessary since it's already a string).

app/backend/src/couchers/servicers/communities.py
```diff
@@ -305,10 +308,7 @@ def JoinCommunity(self, request, context):
                 context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.ALREADY_IN_COMMUNITY)
 
             node.official_cluster.cluster_subscriptions.append(
-                ClusterSubscription(
-                    user_id=context.user_id,
-                    role=ClusterRole.member,
-                )
+                ClusterSubscription(user_id=context.user_id, role=ClusterRole.member,)
```
aapeliv: Please run the auto-formatters in `//app/backend` (so it picks up the config)

aapeliv (changes requested):

app/backend/src/couchers/servicers/communities.py
sadikekin: When I run this in `//ap/backend` it reformats a bunch of files. I run the following command as stated in the documentation.

`autoflake -r -i --remove-all-unused-imports src && isort . && black . in the //app/backend`

aapeliv: Which ones?

sadikekin:
``` 
        modified:   src/couchers/email/__init__.py
        modified:   src/couchers/jobs/enqueue.py
        modified:   src/couchers/jobs/handlers.py
        modified:   src/couchers/jobs/worker.py
        modified:   src/couchers/migrations/env.py
        modified:   src/couchers/migrations/versions/128252798bc4_add_new_email_pattern_constraint.py
        modified:   src/couchers/migrations/versions/3b8d963e0b7d_regions_and_languages.py
        modified:   src/couchers/migrations/versions/c035cdcabd40_add_new_events_tables.py
        modified:   src/couchers/models.py
        modified:   src/couchers/server.py
        modified:   src/couchers/servicers/account.py
        modified:   src/couchers/servicers/api.py
        modified:   src/couchers/servicers/auth.py
        modified:   src/couchers/servicers/blocking.py
        modified:   src/couchers/servicers/conversations.py
        modified:   src/couchers/servicers/events.py
        modified:   src/couchers/servicers/groups.py
        modified:   src/couchers/servicers/jail.py
        modified:   src/couchers/servicers/media.py
        modified:   src/couchers/servicers/pages.py
        modified:   src/couchers/servicers/requests.py
        modified:   src/couchers/servicers/search.py
        modified:   src/couchers/servicers/threads.py
        modified:   src/couchers/tasks.py
        modified:   src/dummy_data.py
        modified:   src/tests/test_account.py
        modified:   src/tests/test_api.py
        modified:   src/tests/test_auth.py
        modified:   src/tests/test_bg_jobs.py
        modified:   src/tests/test_communities.py
        modified:   src/tests/test_discussions.py
        modified:   src/tests/test_email.py
        modified:   src/tests/test_events.py
        modified:   src/tests/test_fixtures.py
        modified:   src/tests/test_groups.py
        modified:   src/tests/test_jail.py
        modified:   src/tests/test_pages.py
        modified:   src/tests/test_references.py
        modified:   src/tests/test_search.py
```

aapeliv (commented):
This looks good now. But I presume the tests don't pass given the change in behaviour. Have you been able to get the tests running locally?

aapeliv:
@sadikekin: how are you doing on this? The tests still need to be updated, and we've merged in the SQLalchemy 2 refactor, so this requires some changes before merging.

app/backend/src/couchers/servicers/communities.py
ShmuelTreiger: If you're still having this issue, try entering backend then running it. I think it's grabbing its settings from app (which has no settings, so it's using default) instead of from backend, where we use length 120.

aapeliv:
Abandoned.

PR #1496, opened 2021-06-25
Rename Search tab to Map Search

aapeliv:
Up for debate, what do you think @lucaslcode?

Screen Shot 2021-06-24 at 9 34 06 pm

**Frontend checklist**
- [ ] All tests pass


lucaslcode:
Imo map Search kind of implies a non-map search which we don't have yet and will probably be the same page anyway

lucaslcode (approved):
But I approve the code :p

PR #1499, opened 2021-06-25
Return user's timezone in User object

aapeliv:
…and use it to display local time on profile

Still needs to dayjs timezone stuff!

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass


kalvdans:
web browsers and Android phones know their timezone already?

app/backend/src/couchers/db.py
```diff
@@ -235,3 +235,9 @@ def can_moderate_node(session, user_id, node_id):
     return _can_moderate_any_cluster(
         session, user_id, [cluster.id for _, _, _, cluster in get_node_parents_recursively(session, node_id)]
     )
+
+
+def timezone_at_coordinate(session, geom):
+    area = session.query(TimezoneArea.tzid).filter(func.ST_Contains(TimezoneArea.geom, geom)).one_or_none()
+    if area:
+        return area.tzid
```
kalvdans, line 243:
```suggestion
        return area.tzid
    return None
```
It is good to be explicit if the function is returning something.

app/backend/src/couchers/models.py
```diff
@@ -263,6 +275,10 @@ class User(Base):
         ),
     )
 
+    @property
+    def timezone(self):
+        return self.timezone_area.tzid if self.timezone_area else "Etc/UTC"
```
kalvdans, line 280: Don't the timezone areas cover the whole earth's surface?

app/proto/account.proto
```diff
@@ -71,6 +71,9 @@ message GetAccountInfoRes {
   // User's verified phone number, or empty string if no successful verification happened yet.
   // in E.164 format (including country code with the + sign)
   string phone = 5;
+
+  // the user's time zone identifier, derived from their coordinates
```
kalvdans:
```suggestion
  // the user's time zone name derived from their coordinates, for example "Australia/Melbourne"
```
Or refer to https://en.wikipedia.org/wiki/Tz_database#Names_of_time_zones ?

app/proto/api.proto
```diff
@@ -198,6 +198,8 @@ message User {
   string city = 4;
   string hometown = 100;
 
+  string timezone = 122;
```
kalvdans, line 202:
```suggestion
  // the user's time zone name derived from their coordinates, for example "Australia/Melbourne"
  string timezone = 122;
```

app/backend/src/couchers/models.py
aapeliv: They ought to. But the main thing is that some users still (!!) don't have a `geom`/coordinate. So this mostly covers that case.

aapeliv:
> web browsers and Android phones know their timezone already?

They do. But we can't really take that timezone and siphon it to the backend and use that to base e.g. hosting request information on. I think it's good for displaying time locally; and we should add some small debug thing to compare the stored timezone with the one derived from the location and flag it if they're different.

kalvdans (dismissed):
Backend LGTM!

app/frontend/src/features/auth/timezone/Timezone.tsx
```diff
@@ -0,0 +1,19 @@
+import { Typography } from "@material-ui/core";
+import PageTitle from "components/PageTitle";
+import { GetAccountInfoRes } from "proto/account_pb";
+
+import { TIMEZONE, TIMEZONE_HELPER, YOUR_TIMEZONE } from "../constants";
+
+export default function Timezone(accountInfo: GetAccountInfoRes.AsObject) {
+  return (
+    <>
+      <PageTitle>{TIMEZONE}</PageTitle>
```
lucaslcode: This is H1 and there should only be one h1 at the top of each page

app/frontend/src/features/auth/timezone/Timezone.tsx
```diff
@@ -0,0 +1,19 @@
...
+      <>
```
lucaslcode: Unnecessary fragment

app/frontend/src/features/user/UserTextAndLabel.tsx
```diff
@@ -33,6 +34,7 @@ export const LabelsReferencesLastActive = ({ user }: Props) => (
           : LAST_ACTIVE_FALSE
       }
     />
+    <LabelAndText label={LOCAL_TIME} text={user.timezone} />
```
lucaslcode: if you create a file `src/utils/dayjs.ts` which is this:
```
import dayjs from "dayjs";
import Timezone from "dayjs/plugin/timezone";
import LocalizedFormat from "dayjs/plugin/timezone";
dayjs.extend(timezone).extend(LocalizedFormat);
export default dayjs;
```

and change all our dayjs imports to use that file, then I believe you should be able to do this:
```suggestion
    <LabelAndText label={LOCAL_TIME} text={dayjs().tz(user.timezone).format("LT")} />
```
And it will display the correct time in that timezone, in the current user's localized time display format.

lucaslcode (changes requested):

app/frontend/src/features/auth/timezone/Timezone.tsx
aapeliv: It's the same for each fragment in the Account Settings screen, I just did the same as the others

app/frontend/src/features/user/UserTextAndLabel.tsx
aapeliv: Awesome, thanks! I thought you or Darren would figure out how to do it really quickly so I wouldn't have to spend three hours on it :P

aapeliv:
Settings page:

Screen Shot 2021-06-25 at 6 23 44 pm

Profile page:

Screen Shot 2021-06-25 at 6 23 32 pm

app/frontend/src/features/auth/timezone/Timezone.tsx
lucaslcode: Can you change them to h2 Typography and put a pagetitle on the page? 👼🏻

app/frontend/src/features/user/UserTextAndLabel.tsx
```diff
@@ -34,7 +35,10 @@ export const LabelsReferencesLastActiveLocalTime = ({ user }: Props) => (
           : LAST_ACTIVE_FALSE
       }
     />
-    <LabelAndText label={LOCAL_TIME} text={user.timezone} />
+    <LabelAndText
+      label={LOCAL_TIME}
+      text={dayjs().tz("America/New_York").format("LT")}
```
lucaslcode: Double check this one ;)

app/frontend/src/features/auth/timezone/Timezone.tsx
```diff
@@ -1,19 +1,25 @@
...
-      <>
-        <Typography variant="body1">
-          {YOUR_TIMEZONE} <b>{accountInfo.timezone}</b>.
-        </Typography>
-        <Typography variant="body1">{TIMEZONE_HELPER}</Typography>
-      </>
+      <Typography variant="body1" paragraph>
+        {YOUR_TIMEZONE} <b>{accountInfo.timezone}</b>
+        {YOUR_LOCAL_TIME_IS}{" "}
+        <b>{dayjs().tz("America/New_York").format("LT")}</b>.
```
lucaslcode: Also this one

lucaslcode (changes requested):

app/frontend/src/features/auth/timezone/Timezone.tsx
aapeliv: Yep

app/frontend/src/features/user/UserTextAndLabel.tsx
aapeliv: Thanks :D

lucaslcode (approved):

PR #1503, opened 2021-06-25
Changed and moved "Your answer will remain private and anonymous" tex…

### LINKED ISSUE #1420: Change and move "Your answer will remain private and anonymous" text in references
Currently the text reads "Your answer will remain private and anonymous" but it should be reworded to make it clear that the answer will not be seen by the other person at all. It's also disingenuous to say it is anonymous because it is not (the user submitting the information is stored in the backend, and thus is not anonymous).

The text should be changed from "Your answer will remain private and anonymous"
to
"Your answer will not be seen by the other person, and will remain private."

Also, the text should be moved to the bottom of the card, so it is the last thing people see before submitting the information (hopefully encouraging them to be as honest as possible):

image

helizaga:
…t in references

Changed and moved "Your answer will remain private and anonymous" text in references.

Also, the text is moved to the bottom of the card.

closes #1420

### DIFF
--- a/app/frontend/src/features/profile/constants.ts
+++ b/app/frontend/src/features/profile/constants.ts
@@ -82,7 +82,7 @@ export const INVALID_STEP = "Invalid form step";
 export const NEXT = "Next";
 export const OKAY = "Okay";
 export const PREVIOUS_STEP = "Previous step";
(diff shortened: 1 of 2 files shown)


lucaslcode (approved):
Thanks for your contribution!

PR #1507, opened 2021-06-25
Move api_calls to a different schema and don't back it up

### LINKED ISSUE #1313: Move traces out of the main database into something better
They're just polluting the main DB and should be in an OLAP db or something.

See also #1312.

aapeliv:
Closes #1313.

I've decided not to move these to a different database for now. That would come with more pain of migrations, and a mess in local and prod dev environments. Instead, we can put them in a different schema and eventually move them to a different [tablespace](https://www.postgresql.org/docs/current/manage-ag-tablespaces.html) if we want them on different media.

**Backend checklist**
- [x] All tests pass


kalvdans (approved):
That was easy :)

You might want to move the Email table as well since it is also just debug.

aapeliv:
This reduced our compressed backups from 551 MB to 83 MB! Great :D

PR #1510, opened 2021-06-26
Frontend/feature/signup v2

lucaslcode:
**Frontend checklist**
- [ ] All tests pass


darrenvong:
Are there more details on how this new sign up flow works?

lucaslcode:
There are comments in the proto files

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
```diff
@@ -82,19 +83,18 @@ describe("contributor form", () => {
   });
 
   it("it shows the form again if processing the form fails", async () => {
```
darrenvong:
```suggestion
  it("shows the form again if processing the form fails", async () => {
```
Nitpick, but it bothers me to read `it` twice in the sentence 😅

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -55,51 +53,57 @@ const useStyles = makeStyles((theme) => ({
...
+  label: { display: "block" },
+  textbox: {
+    marginBlockEnd: theme.spacing(3),
+    marginBlockStart: theme.spacing(1),
+  },
+  radioLabel: { ...theme.typography.body1, color: theme.palette.text.primary },
 }));
 
 interface ContributorFormProps {
   // Return true for success, false for failure. Note that this component does not display errors on failure, do it in the parent.
```
darrenvong: This comment is no longer accurate if its return type is void? does it have the same behaviour but without returning any result somehow?

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
```diff
@@ -82,19 +83,18 @@ describe("contributor form", () => {
...
-    render(<ContributorForm processForm={processForm} />);
+    render(<ContributorForm processForm={processForm} />, { wrapper });
     userEvent.type(
       await screen.findByLabelText(IDEAS_LABEL),
       "I have great ideas"
     );
 
-    userEvent.click(await screen.findByRole("button", { name: SUBMIT }));
-
+    userEvent.click(await screen.findByRole("button"));
```
darrenvong: What happened to the button's specific label here?

app/frontend/src/features/auth/signup/FeedbackForm.test.tsx
```diff
@@ -34,12 +34,14 @@ const stateAfterFeedback = {
 
 describe("signup form (feedback part)", () => {
   it("works", async () => {
-    signupFlowFeedbackMock.mockResolvedValue(stateAfterFeedback);
+    signupFlowFeedbackMock.mockResolvedValue({ ...stateAfterFeedback });
```
darrenvong: The object itself should be the same and so don't need the extra spread?

app/frontend/src/features/auth/signup/Signup.tsx
```diff
@@ -64,7 +71,11 @@ function CurrentForm() {
       <>
         <BasicForm />
         <Typography variant="body1" className={classes.agreement}>
-          {SIGN_UP_AGREEMENT}
+          {SIGN_UP_AGREEMENT[0]}
+          <Link to={tosRoute} component={MuiLink}>
+            {SIGN_UP_AGREEMENT[1]}
+          </Link>
+          {SIGN_UP_AGREEMENT[2]}
```
darrenvong: How come this constant is now in an array? Might be clearer to give them each a more meaningful constant name

app/frontend/src/features/auth/signup/Signup.test.tsx
```diff
@@ -0,0 +1,235 @@
...
+          break;
+        case "email":
+          expect(screen.getByText(SIGN_UP_AWAITING_EMAIL)).toBeVisible();
+          break;
+        case "redirect":
+          expect(await screen.findByText(SIGN_UP_REDIRECT)).toBeVisible();
+          break;
+      }
+    }
+  );
```
darrenvong: Considering the label will say something like "display basic form given state: false false false true false", it is a bit opaque to read and follow. Also there's a fair bit of logic in the test itself which themselves can't be tested, so not sure this is best use case for `it.each` here.

I think it'd be a lot clearer for these to be split out into separate test cases, even if it means a bit of repetition, but at least the test label will read more clearly than a list of booleans...

app/frontend/src/features/auth/signup/FeedbackForm.test.tsx
```diff
@@ -62,9 +64,15 @@ describe("signup form (feedback part)", () => {
       expect(params[1].getExpertise()).toBe("I have lots of expertise!");
     });
 
-    expect(result.current.authState.authenticated).toBe(false);
-    expect(result.current.authState.flowState).toStrictEqual(
-      stateAfterFeedback
-    );
+    const { result: result2 } = renderHook(() => useAuthContext(), {
+      wrapper,
+    });
```
darrenvong, line 69: How come we need to render the hook again from scratch? Will using the `rerender` function in the object returned by `renderHook` not work?

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -94,231 +97,235 @@ export default function AccountForm() {
...
-              error={
-                //@ts-ignore Dayjs type breaks this
-                !!errors?.birthdate?.message
-              }
-              helperText={
-                //@ts-ignore
-                errors?.birthdate?.message ?? " "
+      {errors.location && (
+        //@ts-ignore - we register "location" but rhf thinks the error should be
+        //under location.address
```
darrenvong, line 154: I wonder if updating to v7 of react-hook-form will fix this issue 🤔   it seems to infer the type wrong whenever the field has an object type apart from Date...

app/frontend/src/features/auth/signup/AccountForm.test.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+      userEvent.click(screen.getByLabelText(FEMALE));
+      userEvent.click(screen.getByLabelText(SIGN_UP_TOS_ACCEPT));
+    });
+
+    it("submits correctly", async () => {
+      userEvent.click(screen.getByRole("button", { name: SIGN_UP }));
+
+      await waitFor(() => {
+        expect(signupFlowAccountMock).toHaveBeenCalledWith(
```
darrenvong: The `signupFlowAccount` service function should take an object here really considering the number of arguments it has! At least then this would become `toHaveBeenCalledWith({ flowToken: 'token', ... })` etc where we are forced to be explicit, which would make the meaning of each parameter clearer for those that aren't self-explanatory. E.g. what the heck is 1, 2, 5 supposed to mean at the end there without looking up the file? :p

darrenvong (changes requested):
Big change this, but some very solid testing going on here! Think I've been through 90% of it so will have another look and probably a quick test locally when I get more time.

darrenvong:
Also the tests are failing in CI crashing on `createEvent` - perhaps some async stuff never got returned when you expect them to or maybe forgot an await on a `findBy*` somewhere? I see that error sometimes for the latter case.

app/frontend/src/features/auth/signup/FeedbackForm.test.tsx
lucaslcode: I tried it, it always gave the same thing unless I did a new one

app/frontend/src/features/auth/signup/Signup.test.tsx
lucaslcode: How strongly do you feel about it? I know it seems a bit weird with the label but I thought if there's a problem with the test, you will be looking at this part of the file anyway, and the label makes it very clear which case is being tested. It's going to be at least 100 extra lines to set up each of these cases individually

app/frontend/src/features/auth/signup/Signup.tsx
lucaslcode: Because it's "Please check the \<a>terms of service\</a> here" and I don't want to split that up into three constants, nor do I want a function returning a react node in the constants file.

Once we implement an i18n library, this will be much easier (all the options being considered can handle cases like this).

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: I refactored to use `useMutation` so the promise throws if there's an error and it is now displayed.

lucaslcode:
Ugh these are super hard to debug. I will try running with `-i` to see what is causing the issue

lucaslcode:
When I run tests sequentially, I don't get any error. When I run `--all`, I get failures and the `createEvent` issue but at the end, all tests have passed. At a bit of a loss now.

app/frontend/src/features/auth/signup/Signup.test.tsx
darrenvong: I think the extra lines because of repetition are a good sacrifice for better readability in test code. If any of these fail, there's big productivity value in being able to figure out which one failed from the test output without having to come back here to work out which branch caused the problem imo.

Aside from the labels, the extra logic within the test adds a layer of abstraction which makes them less straightforward to follow, as this is essentially multiple tests squashed into one when each really deserve its own test. [This blog post](https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html) sums up my thinking more coherently than what I can say here I feel :p

app/frontend/src/features/auth/signup/Signup.tsx
```diff
@@ -64,7 +71,11 @@ function CurrentForm() {
       <>
         <BasicForm />
         <Typography variant="body1" className={classes.agreement}>
-          {SIGN_UP_AGREEMENT}
+          {SIGN_UP_AGREEMENT[0]}
+          <MuiLink to={tosRoute} component={Link}>
+            {SIGN_UP_AGREEMENT[1]}
+          </MuiLink>
```
darrenvong: This is probably [one of the few exceptions](https://css-tricks.com/use-target_blank/#a-good-reason-the-user-is-working-on-something-on-the-page-that-might-be-lost-if-the-current-page-changed) where I'd say it's worth making the link open in a new tab so they don't lose what they typed if they clicked on the link to read the TOS.

darrenvong:
Says the tests failed again... any idea which tests are being flaky?

app/frontend/src/features/auth/signup/AccountForm.test.tsx
```diff
@@ -0,0 +1,202 @@
...
+          lat: 1,
+          lng: 2,
+          address: event.target.value,
+          radius: 5,
+        });
+      }}
+    />
+  ),
+}));
+
```
darrenvong, line 50: This is missing a test case for when the mutation errored

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -87,238 +89,259 @@ export default function AccountForm() {
...
+          birthdate: birthdate.format().split("T")[0],
+          gender,
+          acceptTOS,
+          hostingStatus,
+          city: location.address,
+          lat: location.lat,
+          lng: location.lng,
+          radius: location.radius,
+        }
+        // TODO: password
```
darrenvong, line 124: Is this a backend limitation at the moment with the sign up RPC (that you can't supply password during sign up)?

app/frontend/src/features/auth/signup/Signup.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+          lat: 1,
+          lng: 2,
+          address: event.target.value,
+          radius: 5,
+        });
+      }}
+    />
+  ),
+}));
+
```
darrenvong, line 78: There are two missing cases here:
- When the component is rendered in a route with a `urlToken`
- When using the `urlToken` to continue the sign up throws an error

The code coverage report has actually been pretty useful for spotting these cases in this instance, as I usually don't obsess over 100% coverage if they are low value things that's gonna be a pain to test. But they seem reasonable to do here and important use cases!

darrenvong (changes requested):
Should be the final round, mostly comments on tests. Otherwise this new sign up flow is looking slick and makes more sense when I start playing around with it! 😄

lucaslcode:
I really have no idea. Is there a way we can proxy the api to find out what is not being mocked?

app/frontend/src/features/auth/signup/AccountForm.tsx
lucaslcode: No we just didn't put it in frontend yet

darrenvong:
Do you mean similar to what you did with the `msw`? Might be tricky with grpc. But everything interacts with the API is done through `service` and that is mocked, so everything should be mocked? Maybe there is some async update we aren't waiting for in the tests before letting it finish? Seems to be the most common answer when searching for this error...

lucaslcode:
I mean proxy the service later so we know what is being called. Like in storybook, if you make an API call that doesn't have a mock implementation, you get a console warning

darrenvong:
The best I can think of is probably iterating through all of the service functions and give it a default mock implementation that log the warning if it's then not replaced by the individual test files later on.

darrenvong:
This is a bit of a weird one as I still get the warning (using `yarn test-ci`) even on develop, but it's fine on CI...

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
```diff
@@ -47,63 +50,59 @@ describe("contributor form", () => {
 
   it("can be submitted filled", async () => {
```
lucaslcode: I am 90% sure this test is causing the issue but I don't know why @darrenvong

app/frontend/src/setupTests.ts
```diff
@@ -8,7 +8,42 @@ import mediaQuery from "css-mediaquery";
 
 import user from "./test/fixtures/defaultUser.json";
 
-jest.mock("./service");
+const mockedService: any = {};
+
+jest.mock("./service", () => ({
```
lucaslcode: I added this and "fixed" three instances where api calls weren't being mocked, but it didn't make any difference and I'm not sure if we should actually keep this.

app/frontend/src/features/auth/useAuthStore.test.ts
```diff
@@ -39,9 +39,9 @@ describe("usePersistedState hook", () => {
 describe("useAuthStore hook", () => {
   it("sets and clears an error", async () => {
     const { result } = renderHook(() => useAuthStore(), { wrapper });
-    act(() => result.current.authActions.authError("error1"));
+    void act(() => result.current.authActions.authError("error1"));
     expect(result.current.authState.error).toBe("error1");
-    act(() => result.current.authActions.clearError());
+    void act(() => result.current.authActions.clearError());
```
lucaslcode: I will remove these - I used `no-floating-promises` to find missing awaits in tests (didn't help)

app/frontend/src/setupTests.ts
darrenvong: Probably not, I remember it made `console.log`ing to figure out jest mock specific things a pain in the arse since all you get is `Proxy` in the output, and jest generates a mock stub automatically anyway (sort of, but with no implementation) with the code before the change.

How does `mockedService` even get populated here?

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
darrenvong: ~~The main difference I noticed is in this component, we are putting the `pb` stuff directly in the component where that stuff usually lives in `service` which gets mocked completely in tests, so I wonder if that's screwed up the mocking setup in some way...~~

Maybe never mind, looks like it's just a standard object. Does commenting this test out fix CI then?

app/frontend/src/setupTests.ts
darrenvong: Actually, the pain might not be a problem here, but still, probably shouldn't write more code than needed :p

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
lucaslcode: You're right, its a pb, I refactored it but same issue.
I am skipping it now - spent way too long on this. It's tested anyway in `Signup.test.tsx`

app/frontend/src/features/auth/signup/Signup.test.tsx
```diff
@@ -327,4 +333,54 @@ describe("Signup", () => {
...
+    const wrapper = getHookWrapperWithClient({
+      initialRouterEntries: [`${signupRoute}/fakeEmailToken`],
+    }).wrapper;
+    render(<View />, {
+      wrapper,
+    });
+    expect(await screen.findByLabelText(USERNAME)).toBeVisible();
+    expect(signupFlowEmailTokenMock).toBeCalledWith("fakeEmailToken");
+    const { result } = renderHook(() => useAuthStore(), { wrapper });
+    expect(result.current.authState.flowState?.needVerifyEmail).toBe(false);
```
darrenvong, line 362: Couldn't we look into `localStorage` and assert the result there instead?

app/frontend/src/components/ContributorForm/ContributorForm.test.tsx
```diff
@@ -22,97 +21,82 @@ import ContributorForm from "./ContributorForm";
...
-    const processForm = jest.fn();
-    processForm.mockReturnValue(true);
-    render(<ContributorForm processForm={processForm} />);
-    userEvent.type(
-      await screen.findByLabelText(IDEAS_LABEL),
-      "I have great ideas"
-    );
+  //this test intermittantly fails due to async issues
+  //don't know the root cause
+  it.skip("can be submitted filled", async () => {
```
darrenvong, line 44: Somehow this still passes locally or when run using the `runInBand` mode? Yeah this is so weird... let's hope this is the only one for the foreseeable future 😬

darrenvong (approved):
Looks good from a final quick test of the app for real locally! The last question I had is not really blocking I guess as it's still testing/asserting on the right thing which is the important bit. Nice work on battling through all the testing quirks 👍

app/frontend/src/features/auth/signup/Signup.test.tsx
lucaslcode: I thought about that, but local storage is basically hacked in so thought it's not a great idea, and this works fine

PR #1514, opened 2021-06-27
fix default headings persisting

### LINKED ISSUE #1356: More robust check for unedited "default" profile fields
I noticed some people have the blank defaults in their profile. Not sure why.

lucaslcode:
closes #1356 
Visual editor seems to strip `<br>` and \n at the end of the content, so the strings didn't match.
Unfortunately we can't test for this since we mock out the visual editor

@aapeliv you can probably clean these from the db (a bit dangerous though :P)


aapeliv (approved):
Yep, I'll clean up the DB manually

PR #1516, opened 2021-06-27
comment count break fix

### LINKED ISSUE #1483: Spacing of "Comments | XX" in bottom right of discussion container is off
Subject: Spacing of "Comments | XX" in bottom right of discussion container is off
Description:
See screenshot below

Results:
The word "Comments" and the digit should be on the same line separated by a pipe. 

Eg:    Comments | 3

Backend version: develop-d6943b34
Frontend version: develop-d6943b34
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:89.0) Gecko/20100101 Firefox/89.0
Page: https://app.couchers.org/community/4/americas
User (spoofable): jesse (57)

sadikekin:
closes #1483 

**Frontend checklist**
- [X] Formatted my code with `yarn format && yarn lint --fix`
- [X] Added any new components to storybook
- [X] All tests pass

Screenshot 2021-06-27 at 16 52 15


lucaslcode:
Hi @sadikekin!
I think we should take this opportunity to make it look acceptable on mobile.
Right now, it has it's own column on the right and on narrow screens, the main text of the card gets really squished up
Can you move the comment count text so it is on it's own line underneath everything else? Also make sure the margins are suitable for the new position. And the count text can be the smaller "body2" instead of "body1" variant.

Now: 
image

Instead, something like this:
image

lucaslcode (changes requested):
see above

sadikekin:
Screenshot 2021-06-30 at 21 56 55

app/frontend/src/features/communities/discussions/DiscussionCard.tsx
```diff
@@ -30,11 +30,12 @@ const useStyles = makeStyles((theme) => ({
     display: "flex",
     flexDirection: "column",
     flexGrow: 1,
-    marginInlineEnd: theme.spacing(2),
+    marginInlineEnd: theme.spacing(1),
```
lucaslcode:
```suggestion
```

lucaslcode (changes requested):
Looks great! I think we don't need this right margin at all now, since we have the card padding.

lucaslcode (approved):
Thanks!!

PR #1517, opened 2021-06-27
Coarsen last active to an hour and display appropriate text in frontend

### LINKED ISSUE #938: Last active should be different for <15 min
Last active is rounded to 15 min or something, so for less than 15 min ago it should say "<15 min ago" or "just now", or similar?

aapeliv:
Closes #938

Prompted by confusion on the forum: <https://community.couchers.org/t/last-active/1313>

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/utils/timeAgo.ts
+++ b/app/frontend/src/utils/timeAgo.ts
@@ -12,13 +12,20 @@ export const twoMonthMillis = weekMillis * 8;
 export const yearMillis = 31557600000;
 export const twoYearMillis = 31557600000 * 2;
 
-export function timeAgo(input: Date | string, fuzzy: boolean = false) {
+export const lessThanHour = "Less than an hour ago";
+
+export function timeAgo(
+  input: Date | string,
+  fuzzyMillis?: number,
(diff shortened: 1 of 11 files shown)


app/frontend/src/utils/timeAgo.ts
```diff
@@ -12,13 +12,20 @@ export const twoMonthMillis = weekMillis * 8;
 export const yearMillis = 31557600000;
 export const twoYearMillis = 31557600000 * 2;
 
-export function timeAgo(input: Date | string, fuzzy: boolean = false) {
+export const lessThanHour = "Less than an hour ago";
+
+export function timeAgo(
+  input: Date | string,
+  fuzzyMillis?: number,
+  fuzzyText?: string
```
lucaslcode: I think it's safer to have this as one optional `fuzzy` parameter, an object with required `millis` and `text`, so it's impossible to accidentally set only one.

lucaslcode (changes requested):

lucaslcode (approved):

PR #1528, opened 2021-06-29
[MRG] Adding fake context to test filters independently of api calls

ShmuelTreiger:
@aapeliv does this work for fake context?

**Backend checklist**
- [x] All tests pass


aapeliv (approved):
LGTM

PR #1532, opened 2021-06-29
[MRG] Backend/refactor/emails and tokens

### LINKED ISSUE #1338: Refactor tasks.py
There are all these `send_*_email` functions in `tasks.py` that take a token and expiry text that are generated in another function. The expiry durations should also just be in constants and not be passed in at the call site. I think these should just be merged: and because we don't actually use the expiry text, we should just get rid of it.

ShmuelTreiger:
Attempt at #1338 . Is this sort of what you had in mind?

My opinion:
I'm not sure how I feel about getting rid of the token generation methods. Getting rid of set_email_change_tokens is probably a good move, but I'm not sure getting rid of the other ones improves the codebase

**Backend checklist**
- [x] All tests pass


aapeliv (commented):
Yes, this is along the lines of what I was thinking.

We could still keep the token setting and email queueing code in a separate function, e.g. before there was a

```py
                token, expiry_text = new_signup_token(session, request.email)
                send_signup_email(request.email, token, expiry_text)
```

I think the signup token stuff could just be moved into the send_signup_email to get like `send_signup_email(session, request.email)`.

As is is better as well though

ShmuelTreiger:
Upside of the current version is I've reconcentrated token generation in one place

Downside, we now have return values only used for testing (Could remove and query independently, but seems a pain?)

aapeliv (approved):
This looks good to me if you're happy with it

ShmuelTreiger:
Only concern is the return statements, but it you're good with it in good with it

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -98,8 +97,8 @@ def test_purge_login_tokens(db):
     user, api_token = generate_user()
 
     with session_scope() as session:
-        login_token, expiry_text = new_login_token(session, user)
-        login_token.expiry = func.now()
+        login_token = LoginToken(token=urlsafe_secure_token(), user=user, expiry=now())
```
ShmuelTreiger, line 100: Should fix this. Missed it earlier

Issue #1393, opened 2021-06-08
Add testing for each constraint in the models


ShmuelTreiger:
Add tests to test_models.py which check that each constraint is being followed.

Important, for example, because I discovered that my restraints weren't being picked up due to two separate \_\_table_args\_\_ in the same model.

austinburnett:
Hey @ShmuelTreiger can I work on this issue? I'm new to open source

ShmuelTreiger:
Double check if @aapeliv has anything to add, but yes! Go for it! 

Welcome! Always looking for new contributors :)

aapeliv:
Hi @austinburnett, yes, you're welcome to work on this issue. I'm not sure if it's the most friendly issue to start off with, but you should be fine.

I'll also email you in a bit.

aapeliv:
This is fine for now

PR #1533, opened 2021-06-29
[MRG] Create test_model_constraints.py and pull out two existing constraint tests

ShmuelTreiger:
As a start on #1393 I pulled out the existing constraint tests into their own file. IMO we should aim for a one-to-one test for each constraint, and putting them in their own file will help keep track. This will also make it easier for whoever writes future tests to copy the existing model.

**Backend checklist**
- [x] All tests pass


aapeliv (approved):
LGTM

PR #1534, opened 2021-06-29
Remove unimplemented tabs

### LINKED ISSUE #1523: Breadcrumb Menu Navigation Clarification
The current design for the breadcrumb navigation shows each community and sub-community as one color (teal). 
To clarify which page the user is on, only the present community should be highlighted. For example...

Before: 
---------
Screen Shot 2021-06-27 at 7 36 07 PM

After:
----------
Screen Shot 2021-06-27 at 7 37 36 PM

(Note: each "breadcrumb" (in this case, each community) should still act like a link)

### LINKED ISSUE #1525: Remove non-developed tabs from community pages
We should remove:

- Find host
- Events
- Places
- Hangouts

### LINKED ISSUE #1530: Add "New post" to community overview
Users have been confused about how to create new discussion posts. Until we have more on the overview page it might be a good idea to add this option to encourage use.

lucaslcode:
**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/communities/CommunityPage/EventsSection.tsx
+++ b/app/frontend/src/features/communities/CommunityPage/EventsSection.tsx
@@ -1,17 +1,27 @@
+import Button from "components/Button";
 import HorizontalScroller from "components/HorizontalScroller";
 import { CalendarIcon } from "components/Icons";
 import TextBody from "components/TextBody";
 import {
   EVENTS_EMPTY_STATE,
   EVENTS_TITLE,
+  SEE_MORE_EVENTS_LABEL,
 } from "features/communities/constants";
+import { Community } from "proto/communities_pb";
+import { useHistory } from "react-router-dom";
+import { routeToCommunity } from "routes";
 
 import { useCommunityPageStyles } from "./CommunityPage";
 import EventCard from "./EventCard";
 import SectionTitle from "./SectionTitle";
 
-export default function EventsSection() {
+export default function EventsSection({
+  community,
+}: {
+  community: Community.AsObject;
+}) {
   const classes = useCommunityPageStyles();
+  const history = useHistory();
 
   return (
     <>
@@ -38,7 +48,7 @@ export default function EventsSection() {
             />
           ))
         )}
-        {/*eventsHasNextPage && (
+        {true && ( //eventsHasNextPage && (
           <div className={classes.loadMoreButton}>
             <Button
               onClick={() =>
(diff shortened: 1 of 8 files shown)


app/frontend/src/features/communities/CommunityPage/EventsSection.tsx
```diff
@@ -1,27 +1,17 @@
...
 import { useCommunityPageStyles } from "./CommunityPage";
 import EventCard from "./EventCard";
 import SectionTitle from "./SectionTitle";
 
-export default function EventsSection({
-  community,
-}: {
-  community: Community.AsObject;
-}) {
+export default function EventsSection() {
```
darrenvong: Is this being used anywhere at the moment once you removed from `CommunityPage`? If not, might be better to leave this one to avoid merge conflicts as I'm making changes to this file soon to hook in real event data 😅

app/frontend/src/features/communities/discussions/DiscussionsSection.tsx
```diff
@@ -54,6 +56,17 @@ export default function DiscussionsSection({
...
+        to={`${routeToCommunity(
+          community.communityId,
+          community.slug,
+          "discussions"
+        )}#${composingDiscussionHash}`}
+      >
+        <Button size="small" className={classes.newPostButton}>
+          {NEW_POST_LABEL}
+        </Button>
+      </Link>
```
darrenvong: I think this will lead to the warning about invalid nesting of "interactive elements" under an anchor and so leads to invalid HTML. Probably should make this a link only since it goes to a new URL?

darrenvong (changes requested):

app/frontend/src/routes.ts
```diff
@@ -130,3 +123,4 @@ export const routeToCommunity = (
 export const editCommunityPageRoute = `${communityBaseRoute}/:communityId/:communitySlug/info/edit`;
 export const routeToEditCommunityPage = (id: number, slug: string) =>
   `${routeToCommunity(id, slug, "info")}/edit`;
+export const composingDiscussionHash = "new";
```
darrenvong, line 126: Do we need to update the URL document on this? Creating a new guide is currently under `guide/new`, so I wonder if those needs updating 🤔 
Or maybe not because they will actually be on a new page (although still to be decided as they haven't been implemented properly yet) unlike discussions?

app/frontend/src/features/communities/discussions/DiscussionsSection.tsx
lucaslcode: Link looks super weird. I made it a span

app/frontend/src/routes.ts
lucaslcode: Yes my logic was it's not an actual page, so a hash is fine.

darrenvong (approved):

PR #1535, opened 2021-06-29
[MRG] Added test for node.contained_user_ids association proxy

### LINKED ISSUE #1474: Test new association proxy
Not 100% certain this works as intended, as it's empty the only time it's tested

_Originally posted by @ShmuelTreiger in https://github.com/Couchers-org/couchers/pull/1473#r655908681_

ShmuelTreiger:
Testing the association proxy I recently added. Closes #1474

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/tests/test_communities.py
+++ b/app/backend/src/tests/test_communities.py
@@ -680,6 +680,15 @@ class TestCommunities:
                 assert d.thread.thread_id > 0
                 assert d.thread.num_responses == 0
 
+    @staticmethod
+    def test_node_contained_user_ids_association_proxy(testing_communities):
+        with session_scope() as session:
+            c1_id = get_community_id(session, "Country 1")
+            node = session.query(Node).filter(Node.id == c1_id).one_or_none()
+            assert node.contained_user_ids == [1, 2, 3, 4, 5, 9, 10, 11]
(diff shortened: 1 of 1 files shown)


app/backend/src/tests/test_communities.py
```diff
@@ -680,6 +680,14 @@ def test_ListDiscussions(testing_communities):
                 assert d.thread.thread_id > 0
                 assert d.thread.num_responses == 0
 
+    @staticmethod
+    def test_node_contained_user_ids_association_proxy(testing_communities):
+        with session_scope() as session:
+            c1_id = get_community_id(session, "Country 1")
+            node = session.query(Node).filter(Node.id == c1_id).one_or_none()
+            assert node.contained_user_ids == [1, 2, 3, 4, 5, 9, 10, 11]
```
ShmuelTreiger: I don't understand where users 3, 4, and 5 come from home/how they got in there. 4 and 5 I think are from child nodes, but 3?

aapeliv: This seems right to me, "Country 1" has geom from 0 to 50, so any user with point value between those two should show up

aapeliv (dismissed):

aapeliv (approved):

PR #1537, opened 2021-06-30
Fix broken imports for ListEvents

darrenvong:
**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/communities.py
+++ b/app/backend/src/couchers/servicers/communities.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 
 import grpc
@@ -22,7 +23,7 @@ from couchers.servicers.discussions import discussion_to_pb
 from couchers.servicers.events import event_to_pb
 from couchers.servicers.groups import group_to_pb
 from couchers.servicers.pages import page_to_pb
-from couchers.utils import Timestamp_from_datetime, dt_from_millis, now
+from couchers.utils import Timestamp_from_datetime, dt_from_millis, millis_from_dt, now
 from proto import communities_pb2, communities_pb2_grpc, groups_pb2
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/servicers/communities.py
```diff
@@ -260,18 +261,14 @@ def ListEvents(self, request, context):
                 .filter(Event.owner_cluster == node.official_cluster)
             )
 
-            if not request.past:
```
darrenvong: The communities `ListEvent` request object doesn't seem to have this `past` field so I removed it, but this still doesn't seem to be working...

darrenvong: In my local db I got an online event happening next week, but this is giving me an empty list on the frontend... so this only fix the broken import but probably not the underlying issue with the query 🙃

app/backend/src/couchers/servicers/communities.py
```diff
@@ -1,3 +1,4 @@
+import datetime
```
ShmuelTreiger: Instead of importing datetime, just import timedelta

app/backend/src/couchers/servicers/communities.py
ShmuelTreiger: I'm guessing that ListEventReq in communities.proto was supposed to have a past boolean. LMK if you want me to add it for you.

ShmuelTreiger (dismissed):
Not sure about the actual query. Still trying to understand it. In the meantime, here are two things.

app/backend/src/couchers/servicers/communities.py
darrenvong: Yea probably... I added it to the proto and put the deleted code back

ShmuelTreiger:
If it's still not working IDK. Maybe @aapeliv can help you, I believe he wrote it.

Also make sure to update the front end where the API is actually called

darrenvong:
This is not being used anywhere yet, still in a WIP branch I'm on at the moment, so nothing to update for the current frontend

aapeliv (approved):

ShmuelTreiger:
I'm confused, is the query working or not?

darrenvong:
Haven't had time to look again to see if it's something funky in my DB, but last time I checked it wasn't working :(

PR #1540, opened 2021-06-30
Fix GetAccountInfo bug

### LINKED ISSUE #1539: Bug in prod
```
2021-06-30 22:06:59,054: couchers.interceptors: object of type 'NoneType' has no len()
Traceback (most recent call last):
  File "/app/src/couchers/interceptors.py", line 244, in sanitizing_function
    res = prev_func(req, context)
  File "/app/src/couchers/interceptors.py", line 221, in tracing_function
    raise e
  File "/app/src/couchers/interceptors.py", line 201, in tracing_function
    res = prev_func(request, context)
  File "/app/src/couchers/interceptors.py", line 121, in user_unaware_function
    return user_aware_function(req, context)
  File "/app/src/couchers/servicers/account.py", line 74, in GetAccountInfo
    profile_complete=user.has_completed_profile,
  File "/usr/local/lib/python3.8/site-packages/sqlalchemy/ext/hybrid.py", line 928, in __get__
    return self.fget(instance)
  File "/app/src/couchers/models.py", line 282, in has_completed_profile
    return self.avatar_key is not None and len(self.about_me) >= 20
TypeError: object of type 'NoneType' has no len()
2021-06-30 22:06:59,106: couchers.interceptors: Probably an unknown error! Sanitizing...
```

aapeliv:
Closes #1539 

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -279,7 +279,7 @@ class User(Base):
 
     @hybrid_property
     def has_completed_profile(self):
-        return self.avatar_key is not None and len(self.about_me) >= 20
+        return self.avatar_key is not None and self.about_me and len(self.about_me) >= 20
 
     @has_completed_profile.expression
     def has_completed_profile(cls):
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -279,7 +279,7 @@ class User(Base):
 
     @hybrid_property
     def has_completed_profile(self):
-        return self.avatar_key is not None and self.about_me and len(self.about_me) >= 20
+        return self.avatar_key is not None and self.about_me is not None and len(self.about_me) >= 20
 
     @has_completed_profile.expression
     def has_completed_profile(cls):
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/models.py
```diff
@@ -279,7 +279,7 @@ def timezone(self):
 
     @hybrid_property
     def has_completed_profile(self):
-        return self.avatar_key is not None and len(self.about_me) >= 20
+        return self.avatar_key is not None and self.about_me and len(self.about_me) >= 20
```
ShmuelTreiger: if were doing self.avatar_key is not None, should we do the same for self.about_me?

ShmuelTreiger (commented):
Other than the one question, looks good

I didn't realize an account wasn't complete without a picture. My account was incomplete for a long time 😂

app/backend/src/couchers/models.py
aapeliv: Yeah, I guess that'd be good

ShmuelTreiger (approved):

PR #1541, opened 2021-06-30
Update message when map is not supported on a user's browser

aapeliv:
Requested by @AllhandsNoFeet 

**Frontend checklist**
- [ ] All tests pass


lucaslcode (approved):
Do users know what WebGL is though?

aapeliv:
> Do users know what WebGL is though?

Probably not most of them. Some might get this because they've disabled it though, and it gives people something to hang on to.

PR #1552, opened 2021-07-02
Clean up bug reporter

aapeliv:
The new auth interceptor validates the user id if the cookie is present, so the username is no longer spoofable. This uses that info and cleans up the bug reporter in the process.

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass


lucaslcode (commented):
frontend ok

ShmuelTreiger (approved):

PR #1553, opened 2021-07-02
Feature/donations

Jesus82:
This is the layout with no functionality. I will add that in next commits.


app/frontend/src/features/messages/requests/HostRequestListItem.tsx
```diff
@@ -102,7 +102,7 @@ export default function HostRequestListItem({
                 )}
               </Typography>
             </div>
-            <Typography component="div" display="inline" variant="h3">
+            <Typography component="div" display="inline" variant="h4">
```
lucaslcode: Are all these h3 to h4 changes because you updated the theme?

I think you can keep them all as they are even with the theme update. We have been doing the headers structurally (h1 to h6) rather than following the size very precisely from the mockup.

Some of the changes (not this one, the ones which don't specify a component) mean that there is an h4 without an h3 now.

app/frontend/src/pb/account_grpc_web_pb.d.ts
```diff
@@ -0,0 +1,104 @@
+import * as google_protobuf_empty_pb from 'google-protobuf/google/protobuf/empty_pb';
```
lucaslcode: You've accidentally committed all the old pb files. You can delete the pb folder, the protobuf files are in `src/proto` now

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const classes = useStyles();
+
+  return (
+    <div>
+      <div className={classes.donationsWrapper}>
+        <div className={classes.donationsLogoHeader}>
+          <CouchersLogo className={classes.donationsLogo} />
+          <div className={classes.donationsLogoText}>
+            <Typography variant="h2">{DONATIONS_BANNER_TITLE}</Typography>
+            <Typography variant="subtitle1">{DONATIONS_BANNER_TEXT}</Typography>
```
lucaslcode: This is just body1 I think

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,203 @@
...
+              {`${value.currency}${value.amount}`}
+            </Button>
+          ) : null
+        )}
+        <div className={classes.inputWrapper}>
+          <input type="number" className={classes.inputNumber} />
+        </div>
+      </div>
+      <div className={classes.marginBottom2}>
+        <Typography variant="subtitle1">{DONATIONSBOX_TEXT}</Typography>
```
lucaslcode: body1?

app/frontend/src/theme.ts
```diff
@@ -191,6 +193,9 @@ const themeOptions: ThemeOptions = {
         fontSize: "1rem", //14px
       },
     },
+    subtitle1: {
+      fontSize: "1rem", //16px
+    }
```
lucaslcode: If the two uses are indeed body1 you can remove this

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+import CouchersLogo from "resources/CouchersLogo";
+
+import DonationsBox from "./DonationsBox";
+
+const useStyles = makeStyles((theme) => ({
+  donationsImage: {
+    width: "100%",
+    height: "100%",
+    objectFit: "cover",
+    opacity: "0.3",
```
lucaslcode:
```suggestion
    opacity: 0.3,
```

app/frontend/src/components/Navigation/Navigation.tsx
```diff
@@ -175,7 +175,7 @@ export default function Navigation() {
   }
   return (
     <AppBar
-      position="fixed"
+      position="sticky"
```
lucaslcode, line 192: What does this do differently?
The width needs fixing after this

app/frontend/src/routes.ts
```diff
@@ -17,6 +17,7 @@ export const couchersURL = "https://couchers.org";
 export const forumURL = "https://community.couchers.org";
 
 export const contributeRoute = "/contribute";
+export const donationsRoute = "/donations";
```
lucaslcode:
```suggestion
export const donationsRoute = "/donate";
```

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+      fontSize: "1rem",
+    },
+  },
+}));
+
+export default function Donations() {
+  const classes = useStyles();
+
+  return (
+    <div>
```
lucaslcode: If this is just because you can only return a single child, use a fragment `<>`

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+  },
+
+  donationsText: {
+    "&& p": {
+      fontSize: "1rem",
+    },
+    "&& li": {
+      fontSize: "1rem",
+    },
+  },
```
lucaslcode: Instead of doing this, do you want to change `theme.ts` so the default body size is 1rem on desktop and 0.875rem on mobile? The mockup text styles have changed a while ago

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <div className={classes.donationsLogoHeader}>
+          <CouchersLogo className={classes.donationsLogo} />
+          <div className={classes.donationsLogoText}>
+            <Typography variant="h2">{DONATIONS_BANNER_TITLE}</Typography>
+            <Typography variant="subtitle1">{DONATIONS_BANNER_TEXT}</Typography>
+          </div>
+        </div>
+        <img className={classes.donationsImage} 
+             src={Landscape} 
+             alt={DONATIONS_IMAGE_ALT}/>
```
lucaslcode: This is a decorative image, it should have a blank alt text `alt=""`

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,203 @@
...
+      borderRadius: "0.5rem",
+      boxShadow: "initial",
+    },
+    "&:hover": {
+      border: `2px solid ${theme.palette.primary.main}`,
+      backgroundColor: theme.palette.background.paper,
+    },
+    "&:hover > .MuiButton-label": {
+      color: theme.palette.primary.main,
+      transition: "color 250ms cubic-bezier(0.4, 0, 0.2, 1)",
```
lucaslcode: try to use [the theme](https://next.material-ui.com/customization/transitions/#durations):
```suggestion
      transition: `color ${theme.transitions.duration.short} ${theme.transitions.easing.easeInOut}`,
```

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,203 @@
...
+    },
+  },
+
+  buttonSecondaryText: {
+    color: theme.palette.grey[600],
+    fontWeight: 700,
+    transition: "color 250ms cubic-bezier(0.4, 0, 0.2, 1)",
```
lucaslcode: ditto, and all the other `transition`s

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,203 @@
...
+  },
+
+  buttonMain: {
+    backgroundColor: theme.palette.primary.main,
+    "&&": {
+      borderRadius: "0.5rem",
+      boxShadow: "initial",
+    },
+    "&:hover": {
+      opacity: "0.4",
```
lucaslcode:
```suggestion
      opacity: 0.4,
```
nit

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,203 @@
...
+                root: classes.buttonSecondary,
+                label: classes.buttonSecondaryText,
+              }}
+            >
+              {`${value.currency}${value.amount}`}
+            </Button>
+          ) : null
+        )}
+        <div className={classes.inputWrapper}>
+          <input type="number" className={classes.inputNumber} />
```
lucaslcode: Use our TextField component, which by default is `variant="outlined"` and you may not need any custom styling (or much less):
```
<TextField
  type="number"
  InputProps={{startAdornment: <InputAdornment position="start">{DONATIONSBOX_CURRENCY}</InputAdornment>}}
/>

File: app/frontend/src/features/donations/DonationsBox.tsx [hunk 16]
```diff
@@ -0,0 +1,203 @@
...
+          }}
+        >
+          {DONATIONSBOX_ONETIME}
+        </Button>
+      </div>
+
+      <Divider className={classes.marginY2} />
+
+      <div className={classes.donationsBoxRow}>
+        {Array.from(DONATIONSBOX_VALUES?.values() ?? []).map((value) =>
```
lucaslcode: Can the rows be centered if they don't fill the whole space?
Also, it would be great if there was never the custom input by itself on a row. Maybe by having each grid cell have 2 buttons instead of one?

app/frontend/src/features/donations/constants.ts
```diff
@@ -0,0 +1,50 @@
...
+  "Your donation will support the future of couchsurfing";
+export const DONATIONS_IMAGE_ALT = "Couchers banner" 
+export const DONATIONS_TITLE = "You can build the future of couchsurfing";
+export const DONATIONS_TEXT = `[Couchers.org](https://couchers.org) is a non-profit that is run on donations, similar to the model used by Wikipedia.
+
+We believe that there is enough good will in the couch surfing community to build a platform that will easily outcompete Couchsurfing, and restore ownership to all of us.
+
+Couchsurfing was built by the community, and we fully intend to put it back in their hands by building an amazing product. We need your help to do so.
+
+[Couchers.org](https://couchers.org) is a place to connect with others, not a place to run advertising.
```
lucaslcode: There are only 2 actual links on the desktop mockups, having too many looks too busy imo

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,123 @@
...
+    width: "100%",
+    height: "100%",
+    objectFit: "cover",
+    opacity: "0.3",
+  },
+
+  donationsLogo: {
+    "&&": {
+      height: "72px",
+      width: "initial",
```
lucaslcode, line 31: From memory, I should have used this trick in `Avatar` somewhere related to the `grow` prop.....

lucaslcode (changes requested):
Looks great!

app/frontend/src/features/donations/Donations.tsx
Jesus82: Body 1 is 0.875rem/14px, while the font size here is 1rem/16px. There was not any class in the theme for that, that's why I had to use subtitle1 (I tried to create a custom body3 with no luck).

app/frontend/src/features/donations/Donations.tsx
Jesus82: Ok, I will do and then remove this classes (not the li one)

app/frontend/src/features/donations/Donations.tsx
lucaslcode: Yes, I realized that our `theme.ts` is out of date, so you can change body 1 to 1rem for desktop (:

app/frontend/src/features/donations/constants.ts
Jesus82: Here I just copy pasted the text from figma, let me know if you want to modify from that.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: I'm checking that out and it seems I have to overwrite lots of stuff there. Is it OK, if I leave the syled input?

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: I grouped them in groups of two. They don't appear centered, but IMO that's more intuitive. If not I would have to rewrite everything with flex using some hacks.

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: Yeah ok, I guess this one is very different

app/backend/src/pb/account_pb2.py
```diff
@@ -0,0 +1,519 @@
+# -*- coding: utf-8 -*-
```
lucaslcode: And the backend ones :D

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: In that case add an aria-label

ItsiW (dismissed):
This looks really great! Thanks for doing this.

A few suggestions:
1. Logo should be without the beta sticker
2. Make the benefactor text its own variable
3. Add a header donations_title2 "Where your donation will go" before the rest of donations_text2

For mobile view, the order should be reorganised
1. donation box
2. benefactor text
3. donations_title
4. donations_text
5. donations_title2
6. donations_text2

I'll also send through updated copy for the text sections

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+  },
+
+  donationsPage: {
+    display: "grid",
+    gridTemplateColumns: "624px 408px",
+    gridGap: theme.spacing(8),
+    position: "relative",
+    left: "50%",
+    transform: "translateX(-50%)",
```
darrenvong, line 63:
```suggestion
```
Don't think this is needed as the section appears in the same position without these rules on large screens?

app/frontend/src/features/donations/constants.ts
```diff
@@ -0,0 +1,49 @@
+export const DONATIONS_BANNER_TITLE =
+  "Your help is needed to support our mission";
+export const DONATIONS_BANNER_TEXT =
+  "Your donation will support the future of couchsurfing";
+export const DONATIONS_TITLE = "You can build the future of couchsurfing";
```
darrenvong: I think couch-surfing here needs a hyphen? Otherwise we might get into trouble with Couchsurfing.org 🤔

app/frontend/src/features/donations/constants.ts
```diff
@@ -0,0 +1,49 @@
...
+export const DONATIONS_TEXT = `[Couchers.org](https://couchers.org) is a non-profit that is run on donations, similar to the model used by Wikipedia.
+
+We believe that there is enough good will in the couch surfing community to build a platform that will easily outcompete Couchsurfing, and restore ownership to all of us.
+
+Couchsurfing was built by the community, and we fully intend to put it back in their hands by building an amazing product. We need your help to do so.
```
darrenvong: Same comment as above, or I think we need to add the TM superscript if we are referring to Couchsurfing.org?

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,117 @@
...
+      maxWidth: "672px",
+    },
+    [theme.breakpoints.down("sm")]: {
+      margin: theme.spacing(0, 3),
+    },
+  },
+
+  donationsPage: {
+    display: "grid",
+    gridTemplateColumns: "624px 408px",
```
darrenvong: These seems like highly specific "magic numbers" - could this not be done with `fr` units instead?

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,117 @@
...
+    alignItems: "center",
+    [theme.breakpoints.down("sm")]: {
+      height: "8rem",
+    },
+  },
+
+  donationsLogoHeader: {
+    position: "absolute",
+    zIndex: 1,
+    maxWidth: "1100px",
```
darrenvong: Use the `rem` equivalent of this for consistency with the rest of the code

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,117 @@
...
+    display: "flex",
+    alignItems: "center",
+    width: "100%",
+    [theme.breakpoints.down("md")]: {
+      maxWidth: "672px",
```
darrenvong: Same comment about using `rem` anywhere that has `px` values for consistency

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,270 @@
+// import { Divider, makeStyles, Typography, InputAdornment } from "@material-ui/core";
+import { Divider, makeStyles, Typography } from "@material-ui/core";
+import Button from "components/Button";
+// import TextField from "components/TextField";
```
darrenvong: Delete unused code?

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,270 @@
...
+            <input type="number" className={classes.inputNumber} />
+          </div>
+        </div>
+
+        {/*<TextField
+          id="'custom donation'"
+          className={classes.inputNumber}
+          type="number"
+          InputProps={{startAdornment: <InputAdornment position="start">{DONATIONSBOX_CURRENCY}</InputAdornment>}}
+        />*/}
```
darrenvong: Unused code

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,270 @@
...
+                <Button
+                  classes={{
+                    root: classes.buttonSecondary,
+                    label: classes.buttonSecondaryText,
+                  }}
+                >
+                  {`${value.currency}${value.amount}`}
+                </Button>
+              ) : null
+            )}
```
darrenvong: I think it's better to inline and repeat the `Button` JSX markup here instead. This array mapping thing is not really adding any value and makes the code more obscure to read imo, as you are hardcoding the different currency values by repeating the rows anyway, so this is adding an unnecessary layer of indirection with the JSX being generated from the array.

darrenvong (changes requested):

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+  DONATIONSBOX_VALUES,
+} from "features/donations/constants";
+import { Error as GrpcError } from "grpc-web";
+import React, { useEffect, useState } from "react";
+import { Controller, useForm } from "react-hook-form";
+import { useMutation } from "react-query";
+import { useLocation } from "react-router-dom";
+import { service } from "service";
+
+const stripePromise = loadStripe(process.env.REACT_APP_STRIPE_KEY!);
```
lucaslcode: Can this be done in the component - I believe it injects iframes and stuff which we don't want in the rest of the app.

app/frontend/src/features/donations/DonationPage.tsx
```diff
@@ -88,6 +88,7 @@ export default function DonationPage() {
           id="recurring"
           control={control}
           name="recurring"
+          rules={{ required: "This field is required." }}
```
lucaslcode: Currently the value of the field is a boolean, true if recurring. So selecting one-time fails with "This field is required".
You can change it so the onChange line just does `onChange(e.target.value)` and adjust how `onSubmit` works accordingly, or have a custom validation rule that differentiates between undefined and false.

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -53,30 +52,24 @@ const useStyles = makeStyles((theme) => ({
   buttonSecondaryText: {
     color: theme.palette.grey[600],
     fontWeight: 700,
+    fontSize: theme.typography.button.fontSize,
     transition: `color ${theme.transitions.duration.short} ${theme.transitions.easing.easeInOut}`,
```
lucaslcode:
```suggestion
    transition: `color ${theme.transitions.duration.short}ms ${theme.transitions.easing.easeInOut}`,
```
sorry looking at the generated css looks like "ms" is necessary when you use the shorthand transition property.

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -166,8 +158,14 @@ export default function Donations() {
         </Button>
         <Button
           classes={{
-            root: classes.buttonSecondaryActive,
-            label: classes.buttonSecondaryTextActive,
+            root: classNames(
+              classes.buttonSecondary,
+              classes.buttonSecondaryActive
```
lucaslcode: confused about the naming. Is active for ":hover"? In that case don't use a separate class, just do `"&:hover": {...}`

lucaslcode: also you can use `color="secondary"` directly on the `Button` and not worry about css, if that's what you're trying to do

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+export interface DonationFormData {
+  amount: number;
+  recurring: boolean;
+}
+
+export default function DonationsBoxMixed() {
+  const classes = useStyles();
+
+  const [success, setSuccess] = useState<boolean>(false);
+  const [cancelled, setcancelled] = useState<boolean>(false);
```
lucaslcode:
```suggestion
  const [success, setSuccess] = useState(false);
  const [cancelled, setCancelled] = useState(false);
```
Type inference is sufficient here, also snakeCase

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+              </FormHelperText>
+            </FormControl>
+          )}
+        />
+
+        <Divider className={classes.marginY2} />
+
+        <div className={classes.donationsBoxRow}>
+          <div className={classes.donationsBoxSubRow}>
+            <button
```
lucaslcode: all these buttons need `type="button"`, they currently submit the form

If you feel really excited, I think you could reduce some boilerplate but having all these buttons as `<radio>`s, but that's totally optional :P

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+  amount: number;
+  recurring: boolean;
+}
+
+export default function DonationsBoxMixed() {
+  const classes = useStyles();
+
+  const [success, setSuccess] = useState<boolean>(false);
+  const [cancelled, setcancelled] = useState<boolean>(false);
+  const [singleAmount, setSingleAmount] = useState(0);
```
lucaslcode: Instead of doing this, you can watch the react-hook-form value (`watch` is from `useForm`):
`const watchedAmount = watch("amount");`

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+              {DONATIONSBOX_VALUES[0].amount}
+            </button>
+            <button
+              className={`${classes.buttonSecondary} ${
+                singleAmount === DONATIONSBOX_VALUES[1].amount
+                  ? "buttonSecondaryActive"
+                  : ""
+              }`}
+              value={DONATIONSBOX_VALUES[1].amount}
+              onClick={() => setSingleAmount(DONATIONSBOX_VALUES[1].amount)}
```
lucaslcode: You can remove this line on all the buttons if you use the watch suggestion

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+            >
+              {DONATIONSBOX_VALUES[0].currency}
+              {DONATIONSBOX_VALUES[0].amount}
+            </button>
+            <button
+              className={`${classes.buttonSecondary} ${
+                singleAmount === DONATIONSBOX_VALUES[1].amount
+                  ? "buttonSecondaryActive"
+                  : ""
+              }`}
```
lucaslcode: `buttonSecondaryActive` won't be the actual class name. classNames is good for this:
```suggestion
              className={classNames(classes.buttonSecondary, { [classes.buttonSecondaryActive]: watchedAmount === DONATIONSBOX_VALUES[1].amount })}
```

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+              value={DONATIONSBOX_VALUES[1].amount}
+              onClick={() => setSingleAmount(DONATIONSBOX_VALUES[1].amount)}
+              key={DONATIONSBOX_VALUES[1].amount}
```
lucaslcode: Not needed if you are no longer using a map

lucaslcode (changes requested):
Also there's still 20000 LoC in this commit so check the pb files again :D

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+        <div className={classes.donationsBoxRow}>
+          <div className={classes.donationsBoxSubRow}>
+            <button
+              className={`${classes.buttonSecondary} ${
+                singleAmount === DONATIONSBOX_VALUES[0].amount
+                  ? "buttonSecondaryActive"
+                  : ""
+              }`}
+              value={DONATIONSBOX_VALUES[0].amount}
```
lucaslcode: These buttons currently won't set the value in `react-hook-form` because they aren't input components (value doesn't do anything on buttons).
You could wrap them all, and the custom input, in a `<Control render={({onChange, value}) => ...}>` and then call `onChange(amount)` when you want to change the amount (ie. the buttons `onClick` or the custom input's `onChange`), and then you can also have access to the current value from `value` without needing `watch`

aapeliv:
> Also there's still 20000 LoC in this commit so check the pb files again :D

Could we squash so that's purged from the history too?

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: Active is meant for when the option is selected (hovered or not). It is the same behaviour as hover, but they are two different selectors.

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: That wasn't my code, I'll change it if it doesn't break anything :)

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: I also thought about making them radio, but then I realized that we have the "extra" text input. As I understand it, when we set a custom value, the rest of the values should be unselected (if not it would be very confusing), so that is kind of not consistent with them being radio (as radio buttons should always have one of them checked).

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: True that

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: Here you mean `Controller`, isn't it?

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: This is the box component. Am I misunderstanding what a component is in React?

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: I'm a little bit lost on how to do the onclick event, I made a push so you can take a look.

app/frontend/src/features/donations/DonationsBoxMixed.tsx
lucaslcode: This line is top level (not actually in the functional component function) so `loadStripe` will be run when the file is imported (somewhere in `AppRoutes.tsx` I guess, which is itself imported straight away)

app/frontend/src/features/donations/DonationsBoxMixed.tsx
lucaslcode: Yes I mean controller. I will take a look tonight, but it's something like this:
```
<Controller control={control} name="amount" render={({ onChange, value }) => (
  <>
    <MyButton onClick={() => onChange(5)} isActive={value === 5} ...>
    <MyButton onClick={() => onChange(10)} isActive={value === 10} ...>
  </>
)} />
```

app/frontend/src/features/donations/DonationsBoxMixed.tsx
darrenvong: Also don't need the non-null assertion on the stripe key if you update the type in `react-app-env.d.ts`

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,440 @@
...
+    <>
+      <form onSubmit={onSubmit} className={classes.donationsBox}>
+        {error && <Alert severity="error">{error.message}</Alert>}
+        {success && (
+          <Alert severity="success">
+            Thank you for donating to Couchers.org! We appreciate your support!
+          </Alert>
+        )}
+        {cancelled && (
+          <Alert severity="warning">The payment was cancelled.</Alert>
```
darrenvong: Put messages here in constants

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,440 @@
...
+                <FormControlLabel
+                  className={classes.buttonSecondaryRadio}
+                  value="one-time"
+                  control={<Radio />}
+                  label={DONATIONSBOX_ONETIME}
+                />
+              </RadioGroup>
+              <FormHelperText error={!!errors?.recurring?.message}>
+                {errors?.recurring?.message ??
+                  "A monthly gift helps us plan ahead and build our operations on a stable footing."}
```
darrenvong: Move to constants

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,440 @@
...
+          <Alert severity="warning">The payment was cancelled.</Alert>
+        )}
+        <Typography className={classes.marginBottom2} variant="h3">
+          {DONATIONSBOX_TITLE}
+        </Typography>
+        <Controller
+          id="recurring"
+          control={control}
+          name="recurring"
+          rules={{ required: "This field is required." }}
```
darrenvong: Move to constants

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,440 @@
...
+          defaultValue="monthly"
+          render={({ onChange, value }) => (
+            <FormControl className={classes.formGroup}>
+              <RadioGroup
+                className={classes.donationsBoxRow}
+                aria-label="recurring"
+                name="recurring-radio"
+                onChange={(e) => onChange(e.target.value === "monthly")}
+                value={value ? "monthly" : "one-time"}
```
darrenvong: Move to constants

app/frontend/src/routes.ts
```diff
@@ -16,8 +16,9 @@ export const baseRoute = "/";
 export const couchersURL = "https://couchers.org";
 export const forumURL = "https://community.couchers.org";
 
-export const contributeRoute = "/contribute";
-export const donationsRoute = "/donate";
+export const donationsRoute = "/donatemore";
+export const donationRoute = "/donate";
+export const donationsMixedRoute = "/donatemixed";
```
darrenvong: Why do we need three routes here? Shouldn't `/donate` be enough since this is all on one page?

app/frontend/src/AppRoute.tsx
```diff
@@ -60,28 +56,34 @@ export default function AppRoute({
...
+            <>
               {isJailed ? (
                 <Redirect to={jailRoute} />
               ) : (
-                <>
-                  {variant !== "full-screen" && <Navigation />}
-                  <ErrorBoundary>{children}</ErrorBoundary>
-                </>
+                <>{variant !== "full-screen" && <Navigation />}</>
               )}
```
darrenvong: I'm guessing this is necessary so the navigation bar isn't restricted by the container?

Also, you don't need this `Redirect` since there is already a redirect below, so you can remove it and do:
```typescript
{variant !== "full-screen" && <Navigation />}
```

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
darrenvong: Remove this - I think you may have run `yarn install` in the wrong folder.

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
+  "systemParams": "linux-x64-72",
+  "modulesFolders": [],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
darrenvong: Remove this file - same comment as in `yarn.lock`

darrenvong (commented):
I don't think `app/proxy/protos.pb` should be committed to the repo either (in addition to the random empty yarn artifacts) but won't let me comment directly on the file so writing a note here.

app/frontend/src/routes.ts
lucaslcode: He mentioned to me that this is just temporary while its WIP

app/frontend/src/features/donations/DonationsBoxMixed.tsx
```diff
@@ -0,0 +1,440 @@
...
+          render={({ onChange, value }) => (
+
+          <div className={classes.donationsBoxRow}>
+            <div className={classes.donationsBoxSubRow}>
+              <button
+                type="button"
+                onChange={(amount) => onChange(amount)}
+                className={classNames(classes.buttonSecondary, { [classes.buttonSecondaryActive]: watchedAmount === DONATIONSBOX_VALUES[0].amount })}
+                value={DONATIONSBOX_VALUES[0].amount}
+              >
```
lucaslcode:
```suggestion
              <button
                type="button"
                onChange={() => onChange(DONATIONSBOX_VALUES[0].amount)}
                className={classNames(classes.buttonSecondary, { [classes.buttonSecondaryActive]: value === DONATIONSBOX_VALUES[0].amount })}
              >
```

aapeliv:
When we eventually merge, let's do a squash merge to get rid of the extra stuff in history.

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: Done that but it doesn't seem to trigger the class change (see latest push).

lucaslcode: Sorry, it's my typo - should of course be `<button onClick={() => onChange(...`
Styles are applying but being overriden because the non-active styles have more specificity.

Jesus82: Ok, no worries, I'll look into that :)

Jesus82:
Did another tentative. Buttons are working OK. The only problem is that I'm not able to access the input number value in order to make it active where a number is inserted manually and inactive when a button is selected (I just can access the general form value).

lucaslcode:
You could have a `const isCustom = useState(false)` which is set appropriately in each `onClick/onChange`

Jesus82:
> You could have a `const isCustom = useState(false)` which is set appropriately in each `onClick/onChange`

Got that working. There are still some things I'd like to make cleaner (cleaning the input when pressing the button, and some suggested CSS issues), will do that tomorrow. Cheers!

app/frontend/src/AppRoute.tsx
Jesus82: Yes, with `display:sticky`, the menu is fixed but also "occupies" its space, so it's no longer needed to have different paddings in order to fit the container. For that I had to put it out of the container. Now if the menu changes height, everything will adapt without further change.

app/frontend/src/features/donations/DonationsBoxMixed.tsx
Jesus82: Not sure about how to handle the type issue, just give a look to that in the final commit :)

app/frontend/src/features/donations/Donations.tsx
Jesus82: 624px equals to 27 columns of 24px and 408px to 17 columns of 24px. The donations box width was set by the design mockup. The width of the text was adjusted by me in order to have an optimal text length (between 65 to 80 characters per line) taking in account the font size (16px).

app/frontend/src/features/donations/Donations.tsx
Jesus82: It centers the text between `md` and `lg` breakpoints

app/frontend/src/features/donations/constants.ts
Jesus82: > This looks really great! Thanks for doing this.
> 
> A few suggestions:
> 
>     1. Logo should be without the beta sticker
> 
>     2. Make the benefactor text its own variable
> 
>     3. Add a header donations_title2 "Where your donation will go" before the rest of donations_text2
> 
> 
> For mobile view, the order should be reorganised
> 
>     1. donation box
> 
>     2. benefactor text
> 
>     3. donations_title
> 
>     4. donations_text
> 
>     5. donations_title2
> 
>     6. donations_text2
> 
> 
> I'll also send through updated copy for the text sections

@ItsiW done the reordering in mobile. About the logo, I used our logo component as if I put an image and then the logo updates, we would also have to update it here manually, and that's not ideal. Maybe we could hide the beta sticker with a prop, what do you think, @lucaslcode ?

Jesus82:
> This looks really great! Thanks for doing this.
> 
> A few suggestions:
> 
>     1. Logo should be without the beta sticker
> 
>     2. Make the benefactor text its own variable
> 
>     3. Add a header donations_title2 "Where your donation will go" before the rest of donations_text2
> 
> 
> For mobile view, the order should be reorganised
> 
>     1. donation box
> 
>     2. benefactor text
> 
>     3. donations_title
> 
>     4. donations_text
> 
>     5. donations_title2
> 
>     6. donations_text2
> 
> 
> I'll also send through updated copy for the text sections

@ItsiW done the reordering in mobile, but I don't know which part do you mean with "benefactor text". About the logo, I used our logo component as if I put an image and then the logo updates, we would also have to update it here manually, and that's not ideal. Maybe we could hide the beta sticker with a prop, what do you think, @lucaslcode ?

Jesus82:
I think I have addressed all the issues. Just one usability detail pending imo. When a custom amount is inserted via the input, and then another button is clicked afterwards, should we maybe clear the amount from the input? @lucaslcode

ItsiW:
> @ItsiW done the reordering in mobile, but I don't know which part do you mean with "benefactor text".

By benefactor text I mean "*If you wish to contribute over $1000, please contact us at benefactor@couchers.org for us to arrange a lower fee transfer."

This should go before the "Where your donations will go" heading on desktop, and immediately after the donation box on mobile.

yarn.lock
darrenvong: This and `//app/proxy/protos.pb` still needs purging

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+  },
+
+  buttonSecondary: {
+    "&&": {
+      boxShadow: "initial",
+      color: theme.palette.grey[600],
+      cursor: "pointer",
+      fontWeight: 700,
+      fontSize: theme.typography.button.fontSize,
+      transition: `color ${theme.transitions.duration.short} ${theme.transitions.easing.easeInOut}`,
```
lucaslcode: transitions all need `ms` after the duration (it is stored as a number, jss adds the `ms` for you when you use it directly in the duration property, but not in the shorthand property)

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+      border: `2px solid ${theme.palette.primary.main}`,
+      backgroundColor: theme.palette.background.paper,
+      color: theme.palette.primary.main,
+      transition: `color ${theme.transitions.duration.short} ${theme.transitions.easing.easeInOut}`,
+    },
+
+    "&.buttonSecondaryActive": {
+      border: `2px solid ${theme.palette.primary.main}`,
+      backgroundColor: theme.palette.background.paper,
+    },
```
lucaslcode: This won't be doing anything since that class name doesn't exist - jss renames all the classes.

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+  },
+}));
+
+export interface DonationFormData {
+  amount: number;
+  recurring: boolean;
+}
+
+export default function DonationsBoxMixed() {
+  const stripePromise = loadStripe(process.env.REACT_APP_STRIPE_KEY!);
```
lucaslcode: This calls `loadStripe` on every render. Try this:
`const stripe = useRef(loadStripe(process.env.REACT_APP_STRIPE_KEY));`

Then use `stripe.current` when you need it. Also if you change the type in `REACT_APP_ENV.d.ts` so the key is always defined, you don't need the non-null assertion.

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+  const classes = useStyles();
+
+  const [success, setSuccess] = useState(false);
+  const [cancelled, setCancelled] = useState(false);
+
+  const location = useLocation();
+
+  const [isAmount, setIsAmount] = useState(false);
```
lucaslcode: `isCustomAmount`?

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+  useEffect(() => {
+    const query = new URLSearchParams(location.search);
+    setSuccess(query.get("success") ? true : false);
+    setCancelled(query.get("cancelled") ? true : false);
+  }, [location]);
```
lucaslcode: I not 100% sure about this but maybe @darrenvong can comfirm, but you can avoid one frame of the wrong value if you move this out of `useEffect` and into the default value, like this:

```
const location = useLocation();
const query = new URLSearchParams(location.search);
const [success, setSuccess] = useState(query.get("success") ? true : false);
const [cancelled, setCancelled] = useState(query.get("cancelled") ? true : false);
```

(Recreating query every render isn't a big deal, it just parses the url)

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+
+  const {
+    control,
+    handleSubmit,
+    reset: resetForm,
+    errors,
+  } = useForm<DonationFormData>({ defaultValues: { recurring: true } });
```
lucaslcode:
```suggestion
  } = useForm<DonationFormData>();
```
You've already set the default value on the controller

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+  inputNumberActive: {
+    "&&": {
+      border: `2px solid ${theme.palette.primary.main}`,
+    },
+  },
+}));
+
+export interface DonationFormData {
+  amount: number;
+  recurring: boolean;
```
lucaslcode:
```suggestion
  recurring: "monthly" | "one-off";
```

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+  } = useForm<DonationFormData>({ defaultValues: { recurring: true } });
+
+  const {
+    error,
+    isLoading,
+    mutate: initiateDonation,
+  } = useMutation<void, GrpcError, DonationFormData>(
+    async (formData) => {
+      const stripe = (await stripePromise)!;
+      const session_id = await service.donations.initiateDonation(formData);
```
lucaslcode:
```suggestion
    async ({ amount, recurring }) => {
      const stripe = (await stripePromise)!;
      const session_id = await service.donations.initiateDonation({ amount, recurring: recurring === "monthly" });
```
This way, it won't be broken from the `required` rule

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+        <Typography className={classes.marginBottom2} variant="h3">
+          {DONATIONSBOX_TITLE}
+        </Typography>
+        <Controller
+          id="recurring"
+          control={control}
+          name="recurring"
+          rules={{
+            validate: (value) =>
+              typeof value === undefined ? DONATIONSBOX_REQUIRED : "",
```
lucaslcode: My previous suggestions allow this to just be `required: DONATIONSBOX_REQUIRED`

I think it's better because react-hook-forms discourages undefined as a value (it gives console warnings)

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+          }}
+          defaultValue="monthly"
+          render={({ onChange, value }) => (
+            <FormControl className={classes.formGroup}>
+              <RadioGroup
+                className={classes.donationsBoxRow}
+                aria-label="recurring"
```
lucaslcode: This needs to also be in a constant

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+                name="recurring-radio"
+                onChange={(e) => onChange(e.target.value === "monthly")}
+                value={value ? "monthly" : "one-time"}
```
lucaslcode: If you use my typing suggestions, this can just be `value={value}`

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+              >
+                <FormControlLabel
+                  className={classes.buttonSecondaryRadio}
+                  value="monthly"
```
lucaslcode, line 294: This matches my typing suggestions, not boolean

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: I see, you have the opposite way. Maybe then `isPredefinedAmount`

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+                    {DONATIONSBOX_VALUES[6].amount}
+                  </button>
+                  <div className={classes.inputWrapper}>
+                    <input
+                      type="number"
+                      onChange={(e) => {
+                        onChange(
+                          typeof e.target.valueAsNumber === "number"
+                            ? e.target.valueAsNumber
+                            : DONATIONSBOX_VALUES[0].amount
```
lucaslcode, line 459: Isn't this always a number? Is `NaN` a number? Anyway i think `isNan(...)` is clearer.
I don't think it's good to set a arbitrary amount if they don't type a valid number.
Maybe you could just directly do `onChange(e.target.valueAsNumber)` and then have a rule in the controller that checks if the number is valid? (And add an error text if not?)

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+                        );
+                        setIsAmount(false);
+                      }}
+                      className={classNames(classes.inputNumber, {
+                        [classes.inputNumberActive]: !isAmount,
+                      })}
+                      id="amount"
+                      name="amount"
```
lucaslcode:
```suggestion
```
This is set on the controller

app/frontend/src/resources/CouchersLogo.tsx
```diff
@@ -18,13 +19,20 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function CouchersLogo(props: SvgIconProps) {
+export interface CouchersLogoProps {
```
lucaslcode, line 22:
```suggestion
export interface CouchersLogoProps extends SvgIconProps {
```
For that sweet autocomplete

lucaslcode: Actually, if the extra props aren't used anywhere, better not include them at all and get rid of the `{...SVgIconProps}` below

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+              </FormHelperText>
+            </FormControl>
+          )}
+        />
+
+        <Divider className={classes.marginY2} />
+
+        <Controller
+          control={control}
+          name="amount"
```
lucaslcode, line 316:
```suggestion
          name="amount"
          defaultValue={DONATIONSBOX_VALUES[0].amount}
```

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+              <div className={classes.donationsBoxSubRow}>
+                <button
+                  type="button"
+                  onClick={() => {
+                    onChange(DONATIONSBOX_VALUES[0].amount);
+                    setIsAmount(true);
+                  }}
+                  className={classNames(classes.buttonSecondary, {
+                    [classes.buttonSecondaryActive]:
+                      value === DONATIONSBOX_VALUES[0].amount,
```
lucaslcode: So the buttons aren't highlighted if you enter the same amount in the box, change all these to:
```
value === DONATIONSBOX_VALUES[0].amount && isAmount
```

darrenvong:
I think while using the `order` CSS to re-shuffle the ordering of elements on the page to fit narrower screens seem like a clever idea, we might be better off not doing that and restructure our HTML and rely on the order it appears in the DOM instead. Then for larger screens, I don't think this is affected at the grid level anyway since you are placing items explicitly on the grid.

This is because re-ordering via CSS doesn't actually re-order the elements in the DOM, so for keyboard users, the tab order is a bit confusing on tablet screen sizes where it jumps from the text back to the donation form then to the email (see video below).

The donations amount ("one-time" and "monthly" buttons) should also be reachable via tabbing on the keyboard, which currently doesn't seem possible.

https://user-images.githubusercontent.com/13669362/127240842-551d4dc1-26df-4231-86eb-586d6296aeed.mov

Using the `order` property to [re-order things as opposed to having it ordered logically in the DOM is also something MDN warns against of doing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items#the_order_property_and_accessibility).

Considering this is the page where we are asking people to give us money... I think it's important to get this right and make things as easy as possible for users, even if they happen to be not using a mouse!

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+          control={control}
+          name="amount"
+          render={({ onChange, value }) => (
+            <div className={classes.donationsBoxRow}>
+              <div className={classes.donationsBoxSubRow}>
+                <button
+                  type="button"
+                  onClick={() => {
+                    onChange(DONATIONSBOX_VALUES[0].amount);
+                    setIsAmount(true);
```
lucaslcode: To clear the custom amount box when clicking a button, add a `const customAmountInput = useRef<HTMLInputElement>();` up with the hooks, add it to the `<input ref={...}` and then here (and every button), do `customAmountInput.current.clear()` or whatever it is

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,191 @@
...
+  donationsWrapper: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+    height: "120px",
+  },
+
+  donationsLogoHeader: {
+    position: "absolute",
+    zIndex: 1,
```
lucaslcode, line 44: I would prefer making the image a background image than messing with z-indexes, but this works so if you want just leave as is

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,474 @@
...
+          control={control}
+          name="recurring"
+          rules={{
+            validate: (value) =>
+              typeof value === undefined ? DONATIONSBOX_REQUIRED : "",
+          }}
+          defaultValue="monthly"
+          render={({ onChange, value }) => (
+            <FormControl className={classes.formGroup}>
+              <RadioGroup
```
lucaslcode, line 285: There might need to be a `<FormLabel>` here to be tabbable

lucaslcode (changes requested):
Good stuff, it's mostly all cleaned up.

However it's not submitting at the moment. I suspect it's the typing/validation of the recurring field.

Jesus82:
> I think while using the `order` CSS to re-shuffle the ordering of elements on the page to fit narrower screens seem like a clever idea, we might be better off not doing that and restructure our HTML and rely on the order it appears in the DOM instead. Then for larger screens, I don't think this is affected at the grid level anyway since you are placing items explicitly on the grid.
> 
> This is because re-ordering via CSS doesn't actually re-order the elements in the DOM, so for keyboard users, the tab order is a bit confusing on tablet screen sizes where it jumps from the text back to the donation form then to the email (see video below).
> 
> The donations amount ("one-time" and "monthly" buttons) should also be reachable via tabbing on the keyboard, which currently doesn't seem possible.
> Screen.Recording.2021-07-28.at.00.32.44.mov
> 
> Using the `order` property to [re-order things as opposed to having it ordered logically in the DOM is also something MDN warns against of doing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items#the_order_property_and_accessibility).
> 
> Considering this is the page where we are asking people to give us money... I think it's important to get this right and make things as easy as possible for users, even if they happen to be not using a mouse!

@darrenvong  The point here is that as @ItsiW requested, the order in mobile is different than in the desktop. So this is not to make it fit, but actually to change the order in mobile vs the order in desktop while still having the same html. About the tabs, I will look into it.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: I cannot find REACT_APP_ENV.d.ts anywhere, so I changed the line you proposed and nothing else (tbh I have no idea how stripe is working here, as this part was done by Aapeli).

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: You can enter text in the input, even if the input is type=number. What I do is to highlight the first button if what is entered is not a valid amount. This is in order to avoid having all buttons deselected.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: What do you mean with that, should I change something here?

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: Clever!

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: No, this is right

darrenvong:
> So this is not to make it fit, but actually to change the order in mobile vs the order in desktop while still having the same html.

Yeah that's what I meant, poor wording on my initial comment! But I think that is also causing the tabbing order issue, as the order in the HTML is designed for desktop, and since you are using the same HTML without re-ordering in mobile/tablet screen sizes (and only doing so visually with the CSS `order`), it still has the tabbing order as if we are displaying things in two columns.

I think if you base the ordering of the HTML off the mobile layout, it's "good enough" for desktop too as the user will at least go to the donation box first when they tab, which imo is the main call to action area that we want them to reach to easily first.

app/frontend/src/features/donations/DonationsBox.tsx
darrenvong: Yeah I can't see why this wouldn't work 👍 

You could pass `useState` an initialiser function instead, it would be lazily evaluated and only do so once. i.e.

```typescript
const location = useLocation();
const query = new URLSearchParams(location.search);
// Use initialiser function as opposed to passing in the ternary expression directly
const [success, setSuccess] = useState(() => query.get("success") ? true : false);
```

But yeah, not sure it makes a huge difference in this case anyway, since I think most of the work is still done on the line where you re-create the `URLSearchParams` and unless it's proven to be slow, I probably wouldn't bother optimising it.

app/frontend/src/features/donations/DonationsBox.tsx
darrenvong: It's in `//app/frontend/src`

Jesus82:
I did a push following @lucaslcode suggestion, but I hit a wall. There is one error with the `setRecurring` that comes from `proto/donations_pb`. Let me know how to handle this.

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -206,42 +202,39 @@ const useStyles = makeStyles((theme) => ({
...
   const {
     error,
     isLoading,
     mutate: initiateDonation,
   } = useMutation<void, GrpcError, DonationFormData>(
-    async (formData) => {
+    async ({ amount, recurring }) => {
       const stripe = (await stripePromise)!;
-      const session_id = await service.donations.initiateDonation(formData);
+      const session_id = await service.donations.initiateDonation({ amount, recurring: recurring = "monthly" });
```
lucaslcode: You've accidentally done assignment instead of a boolean test `recurring === "monthly"`. In the form, we changed the type of `recurring` from a boolean to a "state string", but `initiateDonation` is still expecting a boolean.

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: @Jesus82 Did you try this?

app/frontend/src/react-app-env.d.ts
```diff
@@ -11,5 +11,6 @@ declare namespace NodeJS {
     REACT_APP_VERSION: string;
     REACT_APP_IS_VERIFICATION_ENABLED?: "true";
     REACT_APP_IS_COMMUNITIES_PART2_ENABLED?: "true";
+    REACT_APP_STRIPE_KEY?="";
```
lucaslcode: Should be `REACT_APP_STRIPE_KEY: string;`

lucaslcode (changes requested):

app/frontend/src/features/donations/DonationsBox.tsx
aapeliv: Adding a `<FormLabel>` above `RadioGroup` didn't help, but I might've done it wrong

Jesus82: Solved it. The problem was that the "original" radio button was hidden with `display:none` and that prevented tabbing. Changed it to a fixed position outside the viewport, now tabbing is working fine.

Jesus82:
Fixed tabbing and removed flex order and set the html to mobile version (mantaining grid placement in desktop). Only thing missing is to clear the input when an amount button is pressed (couldn't make that work, probably some typing error).

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -83,44 +85,30 @@ const useStyles = makeStyles((theme) => ({
       fontSize: "1rem",
     },
   },
+  benefactorText: {},
```
lucaslcode: remove

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -225,7 +227,7 @@ export default function DonationsBoxMixed() {
     errors,
   } = useForm<DonationFormData>();
 
-  const customAmountInput = useRef<HTMLInputElement | null>(null)
+  const customAmountInput = useRef<HTMLInputElement | null>(null);
```
lucaslcode:
```suggestion
  const customAmountInput = useRef<HTMLInputElement>(null);
```

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -321,7 +327,8 @@ export default function DonationsBoxMixed() {
                   }}
```
lucaslcode, line 329: to clear I guess it should be 
```
onClick={() => {
  if (customAmountInput.current) customAmountInput.current.value = "";
  onChange(DONATIONSBOX_VALUES[0].amount);
  setisPredefinedAmount(true);
}}

**Review by lucaslcode — CHANGES_REQUESTED** [review]
All looking good, suggested how to clear

File: app/frontend/src/components/Navigation/Navigation.tsx [hunk 74]
```diff
@@ -257,6 +269,16 @@ export default function Navigation() {
               label={FORUM}
               labelVariant="h3"
             />
+            {/* <NavButton
+              route={contributeRoute}
+              label={VOLUNTEER}
+              labelVariant="h3"
+            /> */}
```
darrenvong: Is this meant to be removed?

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,481 @@
...
+    "&:hover > .MuiFormControlLabel-label": {
+      border: `2px solid ${theme.palette.primary.main}`,
+      backgroundColor: theme.palette.background.paper,
+      color: theme.palette.primary.main,
+      transition: `color ${theme.transitions.duration.short} ${theme.transitions.easing.easeInOut}`,
+    },
+    "& > .MuiRadio-root": {
+      position: "fixed",
+      top: "-99999px",
+      left: "-99999px",
```
darrenvong:
```suggestion
```
Don't think this is necessary once opacity is already at 0 (presumably so keyboard arrows between them still works), since `position: fixed` already takes the radio button out of the normal document flow and so won't impact the positioning of the rest of the "button"'s content

Also it keeps the focus trap indicator on screen readers (an additional focus box they draw around UI elements) when you use something like VoiceOver, so I think better to have that visible for those users.

app/frontend/src/components/Navigation/Navigation.tsx
darrenvong: I wonder if this is why the report a bug button now triggers an annoying horizontal scroll bar. At least it's better than before where it was cut off in certain widths? 🤷 

Might be worth making a separate bug issue to fix than polluting this PR more.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: ...but it has the drawback that the "real" radio button would probably still be somewhere in the viewport, so any user could accidentally click this invisible button when trying to click something else, which is not desirable imo.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: That was it!

app/frontend/src/components/Navigation/Navigation.tsx
Jesus82: I introduced this here in order to avoid keep having the custom paddings. This shouldn't affect how the scrollbars displays in itself (in this context, sticky is just a fixed position that "makes room for itself"

app/frontend/src/features/donations/DonationsBox.tsx
darrenvong: I don't think it's that big of a deal, since the radio is still within the bounds of the button when I looked. This looks quite hacky and shifts the focus on visual users only, which isn't a very inclusive solution

app/frontend/src/components/Navigation/Navigation.tsx
aapeliv: It's supposed to be uncommented when this is merged with develop (I added the `contributeRoute` (/contribute) back in the release notes PR)

app/frontend/src/features/donations/DonationsBox.tsx
lucaslcode: https://webaim.org/techniques/css/invisiblecontent/ suggests `position: absolute; left: -10000px;` so I will change it to that. It's probably 10ns more performant than 0 opacity.

lucaslcode (commented):
I did a minor clean, @darrenvong if you approve this can be merged.
Make sure to squash the commits! (You can use the arrow next to the usual github merge button)

app/frontend/src/features/donations/Donations.tsx
```diff
@@ -0,0 +1,192 @@
...
+      width: "initial",
+    },
+  },
+
+  donationsWrapper: {
+    display: "flex",
+    justifyContent: "center",
+    alignItems: "center",
+    height: "120px",
+  },
```
darrenvong, line 40: I think this might need the break out parent container trick at least for up to but not including the lg breakpoint, as the couchers logo and the text is overflowing the edge of the image...
Screenshot 2021-08-05 at 01 54 49

app/frontend/src/features/donations/DonationsBox.tsx
```diff
@@ -0,0 +1,495 @@
...
+    "&:hover > .MuiFormControlLabel-label": {
+      border: `2px solid ${theme.palette.primary.main}`,
+      backgroundColor: theme.palette.background.paper,
+      color: theme.palette.primary.main,
+      transition: `color ${theme.transitions.duration.short}ms ${theme.transitions.easing.easeInOut}`,
+    },
+    "& > .MuiRadio-root": {
+      position: "absolute",
+      left: "-10000px",
+    },
```
darrenvong, line 97: > https://webaim.org/techniques/css/invisiblecontent/ suggests position: absolute; left: -10000px; so I will change it to that. It's probably 10ns more performant than 0 opacity.

I was less concerned about the performance, but more that this removes the focus trap state for screen readers, which the `position: "fixed"; opacity: 0;` combo still retain (shows a box around an invisible thing, not ideal but I think still better than nothing) without disrupting the text that's inside the button.

Also we should add some focus states (the blue outline) on these "radio buttons" like what we got on the $ buttons. Since they are fake buttons but styled to look like one, currently it doesn't show any which makes them pretty unusable with keyboard only at the moment without a lot of guess work.

darrenvong (dismissed):
Sorry, some final comments mainly on the usability side of things! Should be good to go after that as I trust you've looked through the code enough times already (even though I haven't) 😅

app/frontend/src/features/donations/Donations.tsx
Jesus82: Fixed this adding the "full-width" variant in AppRoutes that got lost in the merge.

app/frontend/src/features/donations/DonationsBox.tsx
Jesus82: If having "fake" radio buttons is a problem, I can refactor and make them "normal" buttons. What I don't see is to have an invisible radio button just over the fake button, as this can confuse users in some edge situations. Also, they could be moved unintentionally leading to a situation where the radio button could be placed anywhere else, causing people clicking them when wanting to click another item. This is a situation we should avoid having in a "money page", imo.

darrenvong: > Also, they could be moved unintentionally leading to a situation where the radio button could be placed anywhere else, causing people clicking them when wanting to click another item.

It's inside the button at the moment anyway and since it's relative to the body and out of normal flow, I don't see how that will change and hence clicking the radio button outside of the rectangular button itself is a moot point here.

Anyway, I think a "normal" button is probably better since it gives us the same focus styling when tabbing with keyboard like on the dollar buttons for free, so if that's easier I'd go with that. The radio button is a clever solution if we were sticking to using the normal radio buttons but clearly not the case here with the design...

Jesus82: Ok, I'll take care of that. This weekend I will not be able to work in this, so I'd suggest to merge this in order to have the functionality live, and then I'll make a commit with the refactor.

lucaslcode (approved):

PR #1556, opened 2021-07-03
[MRG] SQLA 14 migration continued

ShmuelTreiger:
Working on #903 

In SQLA 2.0, using session.query is being replaced by session.execute(select()). They aren't deprecating it, but they won't be supporting it in the future. This branch will attempt to remove session.query from the codebase. It is as of yet, incomplete.

**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/couchers_select.py
```diff
@@ -38,20 +54,43 @@ def filter_by_username_or_id(self, field):
...
 
-    def filter_users(self, context, table=User):
+    def filter_users(self, session, context, table=User):
+        """
+        Filters out users that should not be visible: blocked, deleted, or banned
+
+        Filters the given table, assuming it's already joined/selected from
+        """
+        hidden_users = _blocked_users(session, context.user_id)
+        return self.filter(table.is_visible).filter(~table.id.in_(hidden_users))
```
kalvdans: It would be fantastic if we could use subqueries for this to make the select query construction independent of the actual session.

app/backend/src/couchers/models.py
```diff
@@ -618,11 +619,14 @@ class GroupChatSubscription(Base):
...
-            .count()
+            .execute(
+                select(func.count())
+                .select_from(Message)
+                .join(GroupChatSubscription, GroupChatSubscription.group_chat_id == Message.conversation_id)
+                .filter(GroupChatSubscription.id == self.id)
+                .filter(Message.id > GroupChatSubscription.last_seen_message_id)
+            )
+            .scalar_one()
         )
```
kalvdans: Move this function out of models.py to eliminate the circular import problem (i.e. do no select in models.py)

kalvdans:
I'm a huge fan of disconnecting the select query creation from the session, I hope we could make this to work :)

app/backend/src/couchers/couchers_select.py
ShmuelTreiger: Makes way more sense. Done.

app/backend/src/couchers/models.py
ShmuelTreiger: LMK what you think of the new method

kalvdans: hmm, was this property unused? Even easier :+1: 

I like it a lot that models.py does not deal with any session now.

ShmuelTreiger: It was used, but only in conversations, which is where I put it.

ShmuelTreiger: >I like it a lot that models.py does not deal with any session now.

Ya, that was a cool added benefit

app/backend/src/couchers/couchers_select.py
```diff
@@ -0,0 +1,90 @@
+from sqlalchemy import union
```
aapeliv: Maybe we could call this file `couchers/sql.py`?

ShmuelTreiger:
The only potential issue I know of regarding this update/migration is that it seems relationships still use Query objects, so if we ever want to do something like `cluster.members.where_users_visible()` we may run into some trouble. Other than that, this is ready to go.

ShmuelTreiger:
@aapeliv @kalvdans any chance of getting this merged today?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -6,13 +6,15 @@
 from datetime import timedelta
 
 import requests
+from sqlalchemy import delete
```
aapeliv: Is this in `.sql`?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -6,13 +6,15 @@
...
 from sqlalchemy.sql import func, or_
 
 from couchers import config, email, urls
 from couchers.db import session_scope
 from couchers.email.dev import print_dev_email
 from couchers.email.smtp import send_smtp_email
 from couchers.models import GroupChat, GroupChatSubscription, HostRequest, LoginToken, Message, MessageType, User
+from couchers.sql import couchers_select as select
```
aapeliv, line 16: This seems to be imported everywhere like this. Should we instead just rename it as `select` in `couchers.sql`. I understand the thinking here, but I feel importing `select` directly would be better probably?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -219,7 +230,11 @@ def process_add_users_to_email_list(payload):
     logger.info(f"Adding users to mailing list")
 
     with session_scope() as session:
-        users = session.query(User).filter(User.is_visible).filter(User.added_to_mailing_list == False).limit(100).all()
+        users = (
+            session.execute(select(User).where(User.is_visible).where(User.added_to_mailing_list == False).limit(100))
+            .scalars()
```
aapeliv, line 234: What does the `scalars()` do in these queries? I'd think it'd return either a single row, e.g. a single `User` (equivalent to `one()` before?) or then just one column (e.g. the primary key?)?

app/backend/src/couchers/jobs/worker.py
```diff
@@ -82,7 +87,7 @@ def service_jobs():
     t = create_prometheus_server(job_process_registry, 8001)
     try:
         while True:
-            # if no job was found, sleep for a second, otherwise query for another job straight away
+            # if no job was found, sleep for a second, otherwise search for another job straight away
```
aapeliv: I think the old wording is better, but up to you

app/backend/src/couchers/servicers/api.py
```diff
@@ -743,26 +760,27 @@ def user_model_to_pb(db_user, session, context):
...
-            session.query(FriendRelationship)
-            .filter(
+        friend_relationship = session.execute(
+            select(FriendRelationship)
+            .where(
                 or_(
                     and_(
-                        FriendRelationship.from_user_id == context.user_id, FriendRelationship.to_user_id == db_user.id
+                        FriendRelationship.from_user_id == context.user_id,
+                        FriendRelationship.to_user_id == db_user.id,
```
aapeliv, line 769: Not important, but did anything change here? Why did the formatter break this line?

app/backend/src/couchers/servicers/events.py
```diff
@@ -3,6 +3,7 @@
 import grpc
 from google.protobuf import empty_pb2
 from psycopg2.extras import DateTimeTZRange
+from sqlalchemy import func, update
```
aapeliv: Are we normally importing these from `.sql `too?

app/backend/src/couchers/servicers/groups.py
```diff
@@ -3,6 +3,7 @@
 
 import grpc
 from google.protobuf import empty_pb2
+from sqlalchemy import delete, func
```
aapeliv: import from `.sql`?

app/backend/src/couchers/servicers/references.py
```diff
@@ -7,12 +7,14 @@
 * TODO: Get bugged about writing reference 1 day after, 1 week after, 2weeks-2days
 """
 import grpc
+from sqlalchemy import union_all
```
aapeliv: from `.sql`?

app/backend/src/couchers/servicers/references.py
```diff
@@ -231,48 +232,52 @@ def AvailableWriteReferences(self, request, context):
...
-                .filter(HostRequest.to_user_id == context.user_id)
+                .where(Reference.id == None)
+                .where(HostRequest.can_write_reference)
+                .where(HostRequest.from_user_id == request.to_user_id)
+                .where(HostRequest.to_user_id == context.user_id)
             )
 
-            host_request_references = q1.union_all(q2).order_by(HostRequest.end_time_to_write_reference.asc()).all()
+            union = union_all(q1, q2).order_by(HostRequest.end_time_to_write_reference.asc()).subquery()
+            host_request_references = session.execute(select(union.c[0], aliased(HostRequest, union))).all()
```
aapeliv: Is `.c[0]` the first column? Are you able to do something like naming it (I think it's `label('whatever')`?), then `.c.whatever`?

app/backend/src/couchers/servicers/requests.py
```diff
@@ -137,33 +140,30 @@ def CreateHostRequest(self, request, context):
...
-                session.query(Message.time)
-                .filter(Message.conversation_id == host_request.conversation_id)
+            initial_message = session.execute(
+                select(Message)
+                .where(Message.conversation_id == host_request.conversation_id)
                 .order_by(Message.id.asc())
                 .limit(1)
-                .one()
-            )
+            ).scalar_one()
```
aapeliv, line 159: You can refactor these to user `first`, but not critical

app/backend/src/couchers/servicers/requests.py
```diff
@@ -436,36 +437,38 @@ def GetHostRequestUpdates(self, request, context):
             if request.newest_message_id == 0:
                 context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_MESSAGE)
 
-            newest_message = session.query(Message).filter(Message.id == request.newest_message_id).one_or_none()
+            if not session.execute(select(Message).where(Message.id == request.newest_message_id)).scalar_one_or_none():
+                context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_MESSAGE)
```
aapeliv, line 441: I'm not sure what's going on here. But I think this RPC is being cut anyway.

app/backend/src/couchers/servicers/threads.py
```diff
@@ -32,11 +33,15 @@ def total_num_responses(database_id):
...
-            + session.query(func.count(Reply.id))
-            .join(Comment, Comment.id == Reply.comment_id)
-            .filter(Comment.thread_id == database_id)
-            .scalar()
+            session.execute(
+                select(func.count()).select_from(Comment).where(Comment.thread_id == database_id)
+            ).scalar_one()
+            + session.execute(
+                select(func.count())
+                .select_from(Reply)
```
aapeliv, line 41: Is a `select_from` the recommended way to do a `count`? Is it just shorthand for a subquery? Just wondering.

app/backend/src/tests/test_auth.py
```diff
@@ -3,6 +3,7 @@
 import grpc
 import pytest
 from google.protobuf import empty_pb2
+from sqlalchemy import delete, func
```
aapeliv: sql?

aapeliv (commented):
Yes, very close to merging. We can definitely push it out in the next couple of hours (though I'm in UTC+3 now so going to sleep in ~3 hours).

aapeliv:
@ShmuelTreiger: there is an option to squash these commits into one so the sqlalchemy so the history is a bit cleaner with this big change. Not necessary to do it, but it might be a good idea if you'd like

app/backend/src/couchers/jobs/handlers.py
kalvdans: I think being explicit using couchers_select everywhere makes it easier to read the  code without having to scroll up.

app/backend/src/couchers/jobs/handlers.py
kalvdans: No, it is in sqlalchemy? I don't understand your comment.

ShmuelTreiger: He wants me to import it from `sqlalchemy.sql`. No big deal, sure

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: I chose to name it `couchers_select` instead of `select` so that it's clear to a future developer we are using our own construct, and so that they don't accidentally import normal `select` and wonder why it's not working

I chose to rename it `select` so that our sql statements are as similar to normal SQLA syntax as possible, 1) for clarity and 2) for troubleshooting

Essentially I wanted it to look like normal SQLA, but I wanted to leave easy breadcrumbs so that some dev in 5 years can pick it up and figure it out

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: `.scalars()` returns an SQLA scalar object, which contains objects of the model type. So here, once you append `.all()`, it gives you a list of all the `User` objects.

https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=scalars#sqlalchemy.engine.Result.scalars

https://docs.sqlalchemy.org/en/14/core/connections.html?highlight=scalars#sqlalchemy.engine.ScalarResult

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: Looks like nothing changed. I'm not sure why it did

app/backend/src/couchers/servicers/requests.py
ShmuelTreiger: `scalar_one` raises an exception if more than one is detected (or none), same as `one`. I didn't really do any reimagining of queries, just translation. Happy to change this and any others to `.first` if that's more fitting

app/backend/src/couchers/servicers/requests.py
ShmuelTreiger: I interpreted it as .one_or_none() would raise an exception if the wrong result was found, so I added an explicit abort. Couldn't figure out what else this was doing here.

app/backend/src/couchers/servicers/threads.py
ShmuelTreiger: This is their recommended format for count. Here's one place they mention it, though I definitely saw it somewhere else as well:

https://github.com/sqlalchemy/sqlalchemy/discussions/6334#:~:text=Correct%20-%20use%20select(func.count()).select_from(Pets)%20for%20that.

ShmuelTreiger:
>@ShmuelTreiger: there is an option to squash these commits into one so the sqlalchemy so the history is a bit cleaner with this big change. Not necessary to do it, but it might be a good idea if you'd like

Either way is fine. I personally appreciate the history to see how things develop, but the squashing makes it cleaner for future users, so I see both ways.

app/backend/src/couchers/servicers/references.py
ShmuelTreiger: Yes, it's the first column

`union_all` returns a tuple of values, no objects, hence the need for aliased. I can perhaps add a line to add a name? It doesn't add much, but it makes it a little clearer:

```
            union = union_all(q1, q2).order_by(HostRequest.end_time_to_write_reference.asc()).subquery()
            union = select(union.c[0].label("surfed"), aliased(HostRequest, union))
            host_request_references = session.execute(union).all()
```

app/backend/src/couchers/servicers/threads.py
ShmuelTreiger: I believe this is where I first got it:

https://stackoverflow.com/questions/65754023/what-is-the-equivalent-of-query-count-in-the-sqlalchemy-1-4-orm

Here it is in the docs:

https://docs.sqlalchemy.org/en/14/tutorial/data_select.html#:~:text=Another%20example%20where%20we%20might%20want%20to%20use

app/backend/src/couchers/jobs/handlers.py
aapeliv: OK, if you both think it's better, let's do this!

app/backend/src/couchers/jobs/handlers.py
aapeliv: Thanks!

app/backend/src/couchers/servicers/api.py
aapeliv: Fair enough. Was wondering if your formatter config was wrong (e.g. line length not 120). Seems it's an isolated thing though, so no problem

app/backend/src/couchers/servicers/references.py
aapeliv: Yeah I think that's a bit clearer.

Also, what's the `aliased` doing here? I'm not sure I understand

app/backend/src/couchers/servicers/requests.py
aapeliv: That's fine

aapeliv: I didn't realise that difference!

app/backend/src/couchers/servicers/threads.py
aapeliv: Thanks!

aapeliv:
I'm happy for this to be merged if you add that one `label` as in the remaining comment, and if you fix the tests!

> > @ShmuelTreiger: there is an option to squash these commits into one so the sqlalchemy so the history is a bit cleaner with this big change. Not necessary to do it, but it might be a good idea if you'd like
> 
> Either way is fine. I personally appreciate the history to see how things develop, but the squashing makes it cleaner for future users, so I see both ways.

Up to you!

app/backend/src/couchers/servicers/references.py
ShmuelTreiger: It reconstitutes the object

https://github.com/sqlalchemy/sqlalchemy/discussions/6714

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: I think @kalvdans means to just import couchers_select and do session.execute(couchers_select(User))

So there's three options

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: image

🤷

app/backend/src/couchers/servicers/references.py
aapeliv: Thanks!

aapeliv (approved):
LGTM if tests go green!

Once we merge, I'll push to next and then we'll have some people manually testing a bit just to make sure none of the more exotic parts are failing.

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: We never really resolved this. But we can discuss and if we want to rename, it's no big deal

aapeliv: Could you remind us to discuss it in the next backend meeting?

PR #1563, opened 2021-07-03
Requests refactor RFC

lucaslcode:
See comments.

Also renamed GroupChats to Chats

### DIFF
--- a/app/proto/conversations.proto
+++ b/app/proto/conversations.proto
@@ -147,6 +147,7 @@ message Chat {
   bool is_dm = 6;
   google.protobuf.Timestamp created = 7;
   uint32 unseen_message_count = 8;
+  uint64 last_seen_message_id = 9;
   Message latest_message = 10;
 }
 
--- a/app/proto/conversations.proto
+++ b/app/proto/conversations.proto
@@ -80,27 +80,27 @@ message MessageContentChatCreated {}
 message MessageContentChatEdited {}
 
 message MessageContentUserInvited {
-  int64 target_user_id = 1;
+  int64 target_user_id = 2;
 }
 
 message MessageContentUserLeft {}
 
 message MessageContentUserMadeAdmin {
-  int64 target_user_id = 1;
+  int64 target_user_id = 2;
 }
 
 message MessageContentUserRemovedAdmin {
-  int64 target_user_id = 1;
+  int64 target_user_id = 2;
 }
 
 // field in Message, when a user is removed by admin
 message MessageContentUserRemoved {
-  int64 target_user_id = 1;
+  int64 target_user_id = 2;
 }
 
 // the host request id is the same as this message id
 message MessageContentHostRequestCreated {
-  int64 target_user_id = 1;
+  int64 target_user_id = 2;
 }
 
 enum HostRequestStatus {
@@ -113,14 +113,14 @@ enum HostRequestStatus {
 }
 
 message MessageContentHostRequestStatusChanged {
-  int64 host_request_id = 1;
+  int64 host_request_id = 2;
   // this is the new host request status
-  HostRequestStatus status = 2;
+  HostRequestStatus status = 3;
 }
 
 message Message {
-  int64 message_id = 1;
-  int64 author_user_id = 2;
+  int64 message_id = 101;
+  int64 author_user_id = 102;
   google.protobuf.Timestamp time = 3;
   oneof content {
     MessageContentText text = 4; // plain text
@@ -165,7 +165,7 @@ message ListChatsRes {
 }
 
 message GetDirectMessageReq {
-  int64 user_id = 1;
+  int64 user_id = 2;
 }
 
 message GetChatMessagesReq {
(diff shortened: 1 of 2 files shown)


app/proto/conversations.proto
```diff
@@ -11,63 +11,59 @@ import "annotations.proto";
...
-    // Retrieves group chat info for the unique (if any) direct message with a
+  rpc GetDirectMessage(GetDirectMessageReq) returns (Chat) {
+    // Retrieves chat info for the unique (if any) direct message with a
     // given user
     //
     // Raises: NOT_FOUND if there is no such direct message thread
   }
 
-  rpc GetUpdates(GetUpdatesReq) returns (GetUpdatesRes) {
-    // Returns all new messages (aka updates) since a given message id
```
lucaslcode, line 30: This wasn't being used since we switched to react-query

app/proto/conversations.proto
```diff
@@ -102,17 +98,25 @@ message MessageContentUserRemoved {
...
   HOST_REQUEST_STATUS_PENDING = 0;
   HOST_REQUEST_STATUS_ACCEPTED = 1;
   HOST_REQUEST_STATUS_REJECTED = 2;
   HOST_REQUEST_STATUS_CONFIRMED = 3;
   HOST_REQUEST_STATUS_CANCELLED = 4;
+  HOST_REQUEST_STATUS_MAYBE_ACCEPTED = 5;
 }
 
 message MessageContentHostRequestStatusChanged {
+  uint64 host_request_id = 1;
```
lucaslcode: Since a thread may have multiple host requests in it

lucaslcode (commented):
I did not attempt to number message fields for backwards compatibility since I don't think we are planning to make this backwards compatible with old clients.

app/proto/conversations.proto
```diff
@@ -108,16 +110,18 @@ enum HostRequestStatus {
   HOST_REQUEST_STATUS_REJECTED = 2;
   HOST_REQUEST_STATUS_CONFIRMED = 3;
   HOST_REQUEST_STATUS_CANCELLED = 4;
+  HOST_REQUEST_STATUS_MAYBE_ACCEPTED = 5;
```
ShmuelTreiger, line 112: What is maybe accepted?

app/proto/conversations.proto
```diff
@@ -84,22 +80,28 @@ message MessageContentChatCreated {}
 message MessageContentChatEdited {}
 
 message MessageContentUserInvited {
-  uint64 target_user_id = 1;
+  int64 target_user_id = 1;
```
ShmuelTreiger: Can these ever be negative? Why the switch to int?

app/proto/requests.proto
```diff
@@ -90,47 +78,5 @@ message ListHostRequestsReq {
...
-}
-
-message GetHostRequestUpdatesRes {
-  repeated HostRequestUpdate updates = 1;
-  bool no_more = 2;
-}
-
-message MarkLastSeenHostRequestReq {
-  int64 host_request_id = 1;
-  int64 last_seen_message_id = 2;
```
ShmuelTreiger, line 135: I assume this was all deleted because the front end doesn't use it? Or was this just wrapped up into the general message stuff? In either event, do we need to remove backend functions and tests?

ShmuelTreiger (commented):
Why rename from group chat? Or I guess, why did we have them as group chat in the first place? I seem to recall a whole long discussion about it, but it's been a while.

app/proto/conversations.proto
lucaslcode: If the host says "maybe" - I thought HOST_REQUEST_STATUS_MAYBE isn't clear if the host or surfer is saying maybe

app/proto/conversations.proto
lucaslcode: More recent proto files like communities use int64 so I changed it to match

app/proto/conversations.proto
ShmuelTreiger: I didn't know hosts could say maybe. Seems weird to me. Was this a requested feature? In any event, code looks good.

lucaslcode: Yes, both from design and people on the forum

lucaslcode:
Rename because I don't know why they were called group chats in the first place. Right now, DMs are group chats with only 2 participants.

app/proto/conversations.proto
lucaslcode: Also, I think the reason was js doesn't support uints so int64 is closer?

kalvdans:
I like the unification and also the rename from groupchat to chat :+1:

app/proto/conversations.proto
```diff
@@ -108,16 +109,18 @@ enum HostRequestStatus {
...
   HOST_REQUEST_STATUS_CONFIRMED = 3;
   HOST_REQUEST_STATUS_CANCELLED = 4;
+  HOST_REQUEST_STATUS_MAYBE_ACCEPTED = 5;
 }
 
 message MessageContentHostRequestStatusChanged {
+  int64 host_request_id = 2;
   // this is the new host request status
-  HostRequestStatus status = 1;
+  HostRequestStatus status = 3;
```
aapeliv: This one is fine to keep as `1`

PR #1578, opened 2021-07-05
[MRG] Backend/refactor/filter to where

ShmuelTreiger:
In SQLA 2.0, `select.where()` is recommended over `select.filter()`

@aapeliv specifically looking for your feedback since you feel strongly about the `couchers_select.where` names

Also, wasn't 100% sure about renaming `filter_`, but I think it's appropriate

"Filter" is used as a verb in a few places, I mostly left it for lack of a better word

**Backend checklist**
- [x] All tests pass


kalvdans (approved):
looks good to me, but await @aapeliv's comments too :)

app/backend/src/couchers/couchers_select.py
```diff
@@ -29,62 +29,62 @@ def couchers_select(*expr):
 
 
 class CouchersSelect(Select):
-    def filter_by_username_or_email(self, field):
+    def where_valid_username_or_email(self, field):
```
aapeliv: This should be called `where_username_or_email`, the `valid` bit is just used to check which column to filter by. Same below for username/id

app/backend/src/couchers/couchers_select.py
```diff
@@ -29,62 +29,62 @@ def couchers_select(*expr):
...
         Filters out users that should not be visible: blocked, deleted, or banned
 
         Filters the given table, assuming it's already joined/selected from
         """
         hidden_users = _relevant_user_blocks(context.user_id)
-        return self.filter(table.is_visible).filter(~table.id.in_(hidden_users))
+        return self.where(table.is_visible).where(~table.id.in_(hidden_users))
 
-    def filter_users_column(self, context, column):
+    def where_users_column_visible(self, context, column):
```
aapeliv: Looking at how this is used, I think it might make sense to rename this to something like `where_user_visible`, e.g. 
```.where_users_column_visible(context, HostRequest.from_user_id)``` would become ```.where_user_visible(context, HostRequest.from_user_id)```

aapeliv (changes requested):

app/backend/src/couchers/couchers_select.py
ShmuelTreiger: So same method name, different parameters?

ShmuelTreiger: Nevermind, Python doesn't support overloading. Having two method calls, `select(...).where_user_visible(...)` vs `select(...).where_users_visible(...)` seems quite confusing to me. I'm not sure what a better name is, but this seems a poor solution. I'll think on it.

ShmuelTreiger: I don't have a better name. But other than that, this is ready

kalvdans (approved):
The rename looks good to me. Merge it to develop after #1585 has been merged to develop.

aapeliv (approved):
LGTM

PR #1582, opened 2021-07-06
PyPI package

aapeliv:
This adds a basic pypi package for calling the APIs from python.

The main use for now is two-fold: debugging (e.g. @darrenvong is implementing events stuff, but he's doing displaying/viewing first and only then a dialog to create them, so this helps him call the APIs easier), and administration/moderation (e.g. allows us to create admin APIs that don't need a frontend if mods/admins can easily enough call them from a jupyter notebook or similar; also allows mods to create events while they're not there yet).

I've published this to pypi: https://pypi.org/project/couchers/

**Backend checklist**
- [ ] All tests pass


app/client/src/couchers/__init__.py
```diff
@@ -0,0 +1 @@
+from .client import get_client  # noqa
```
kalvdans, line 1: Why noqa on this line?

app/client/src/couchers/client.py
```diff
@@ -0,0 +1,59 @@
...
+import os
+
+import grpc
+
+from couchers.proto import auth_pb2, auth_pb2_grpc
+from couchers.services import get_all_stubs
+
+SERVER_ADDRESS = os.getenv("COUCHERS_SERVER_ADDRESS", "api.couchers.org:8443")
+# whether to use TLS/SSL or not. CONNECT_INSECURE only works on localhost
+ENABLE_TLS = os.getenv("CONNECT_INSECURE", "0") != "1"
```
kalvdans: Put these as keyword parameters to get_client() instead. Don't read anything from the environment.

app/client/src/couchers/client.py
```diff
@@ -0,0 +1,59 @@
...
+        creds = grpc.composite_channel_credentials(creds, cookie_creds)
+
+    return grpc.secure_channel(SERVER_ADDRESS, creds)
+
+
+def get_client(username, password):
+    api_key = get_api_key(username, password)
+    channel = create_channel(api_key)
+    stubs = get_all_stubs(channel)
+    return type("__Client", (), stubs)
```
kalvdans, line 64: Too much magic? What does the resulting object look like?

app/client/requirements.txt
```diff
@@ -0,0 +1,7 @@
+autoflake
+black
+build
+grpcio==1.38.0
+isort
+protobuf
+twine
```
kalvdans, line 7: What is twine and what is it used for?

kalvdans (dismissed):
Looks great with diversity in the client space :) I've some small comments.

app/client/src/couchers/__init__.py
aapeliv: `autoflake` will otherwise drop the import

app/client/requirements.txt
aapeliv: `twine` is a tool for uploading the package to pypi

app/client/src/couchers/client.py
aapeliv: Fixed

app/client/src/couchers/client.py
```diff
@@ -0,0 +1,64 @@
+import http.cookies
+
+import grpc
+
+from couchers.proto import auth_pb2, auth_pb2_grpc
+from couchers.services import get_all_stubs
+
+DEFAULT_SERVER_ADDRESS = "api.couchers.org:8443"
```
lucaslcode, line 8: Should this be configurable?

lucaslcode (commented):
Can you add instructions on how to use this in `readme.md`

app/client/src/couchers/client.py
aapeliv: You can pass it in to `get_client`

kalvdans (approved):
Some example usage would be nice but code-wise ok!

app/client/src/couchers/__init__.py
kalvdans, line 1: post-merge comment: Having an `__init__.py` in `couchers` prevents us from using `couchers.foo` in the backend since it is no longer a PEP420 namespace. I'd rather avoid painitng ourselves into that corner and remove this file.

I can't test when I'm on vacation but a simple `pip install couchers` while in the backend venv will show the problems I'm talking about.

Issue #541, opened 2021-01-30
Events MVP


aapeliv:
To-do:
- [x] Create event form/page
  - [x] v1 - allows user to add title, time, date, event details, location, photo
- [x] Event section in community page
- [x] Events listing page for a community (under events tab)
- [x] Single event page
  - [x] v1 - Basic event details - title, time, date, attendees, organisers, photo
  - [x] v2 - event discussions, updating attendance state
- [x] Edit event page (should hopefully be more or less identical to create event)

lucaslcode:
Needs field/filter for online

kalvdans:
Backend part done in https://github.com/Couchers-org/couchers/pull/744, removing backend tag

darrenvong:
Updated the issue description mainly as a check list for me to work through it based on looking at the designs. Have I missed any important pages @lucaslcode @aapeliv?

lucaslcode:
Also editing

darrenvong:
Wrapped up by Couchers-org/couchers#1862 and further non critical enhancements spun off to Couchers-org/couchers#4183

PR #1584, opened 2021-07-07
Events part 2: page for a single, particular event

darrenvong:
Covers more of #541. ~~Still need to do some tests and maybe add a few components to Storybook...~~ now all done!

~~but the main code itself should be ready for a review while I find more time to wrap up for the rest of this week...~~

**Desktop**
Screenshot 2021-07-07 at 02 41 30

**Mobile**
Screenshot 2021-07-07 at 02 44 13

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -0,0 +1,211 @@
...
+        queryClient.invalidateQueries(eventKey(eventId));
+        queryClient.invalidateQueries({
+          predicate: ({ queryKey }) =>
+            queryKey[0] === eventAttendeesBaseKey &&
+            isEventUsersInputType(queryKey[1]) &&
+            queryKey[1].eventId === eventId,
+        });
+      },
+    }
+  );
```
darrenvong: This is not working as intended at the moment... when you flick between "going" and "not going", a new user shows up in the attendees list instead of the current user.

I can't see how the frontend code has messed up here so likely a bug in the backend? 🤷

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -0,0 +1,211 @@
...
+              <Button
+                className={classes.attendanceButton}
+                loading={isSetEventAttendanceLoading}
+                onClick={() => setEventAttendance(event.attendanceState)}
+              >
+                {event.attendanceState ===
+                AttendanceState.ATTENDANCE_STATE_GOING
+                  ? NOT_GOING
+                  : GOING}
+              </Button>
```
darrenvong: Probably should extend this to a dropdown to include the "Maybe" option later, but I think ok for now for MVP/V1 rollout of this page?

app/frontend/src/features/communities/events/EventPage.tsx
lucaslcode: Lol this is so much fun, keep clicking and cycle through users 😆

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -0,0 +1,211 @@
...
+      {(eventError || setEventAttendanceError) && (
+        <Alert severity="error">
+          {eventError?.message || setEventAttendanceError?.message || ""}
+        </Alert>
+      )}
+      {isLoading ? (
+        <CircularProgress />
+      ) : (
+        event && (
+          <>
```
lucaslcode, line 190: Header image?

app/frontend/src/features/communities/events/EventPage.tsx
lucaslcode: To me this isn't clear if I'm attending or not and what clicking the button does.
I think you should make the labels "Join event" and "Leave event", and make the latter an outlined variant.
Could even optionally have a line somewhere "You are attending this event." if so

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -0,0 +1,211 @@
...
+                <CalendarIcon className={classes.calendarIcon} />
+                <div>
+                  <Typography variant="body1">
+                    {dayjs(timestamp2Date(event.startTime!)).format("LLLL")}
+                  </Typography>
+                  <Typography variant="body1">{TO}</Typography>
+                  <Typography variant="body1">
+                    {dayjs(timestamp2Date(event.endTime!)).format("LLLL")}
+                  </Typography>
+                </div>
```
lucaslcode: How does this work with timezones at the moment?
I think you should:
- have all the text inline
- make the width smaller so it's not just one line on wide screens
- make the "to" date only show the time if the date is the same

app/frontend/src/features/communities/events/EventAttendees.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <>
+      <EventUsers
+        emptyState={NO_ATTENDEES}
+        error={error}
+        hasNextPage={hasNextPage}
+        isLoading={isLoading}
+        users={attendees}
+        userIds={attendeesIds}
+        title={ATTENDEES}
```
lucaslcode, line 34: missing prop to open the dialog

app/frontend/src/features/communities/events/hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+    query.data?.pages.flatMap((data) => data.attendeeUserIdsList) ?? [];
+  const { data: attendees, isLoading: isAttendeesLoading } =
+    useUsers(attendeesIds);
+
+  return {
+    ...query,
+    attendeesIds,
+    attendees,
+    isLoading: query.isLoading || isAttendeesLoading,
+  };
```
lucaslcode: I'm not sure about including `useUsers` here, because it is tying user loading to attendee loading. As a result, this query can revert back to `isLoading` if the new list has not-loaded users, which doesn't seem correct to me - once loaded, a query should only ever be fetching if it's stale.

On the other hand, I only caught this because of the backend bug so maybe it doesn't matter. But I don't see why not just use `useUsers` where it is needed instead of bundling it.

app/frontend/src/features/communities/events/EventAttendees.tsx
```diff
@@ -0,0 +1,35 @@
...
+import EventAttendeesDialog from "./EventAttendeesDialog";
+import EventUsers from "./EventUsers";
+import { useEventAttendees } from "./hooks";
+
+interface EventAttendeesProps {
+  eventId: number;
+}
+
+export default function EventAttendees({ eventId }: EventAttendeesProps) {
+  const { attendees, attendeesIds, error, isLoading, hasNextPage } =
```
lucaslcode: How about not destructuring the query and passing that in?

lucaslcode (changes requested):
Haven't finished looking yet

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -0,0 +1,211 @@
...
+          eventKey(eventId),
+          updatedEvent
+        );
+        queryClient.invalidateQueries(eventKey(eventId));
+        queryClient.invalidateQueries({
+          predicate: ({ queryKey }) =>
+            queryKey[0] === eventAttendeesBaseKey &&
+            isEventUsersInputType(queryKey[1]) &&
+            queryKey[1].eventId === eventId,
+        });
```
lucaslcode: I think we should standardize keys as `["name", id, { ...filters }]` - not putting id in an object. That way you can just do `queryClient.invalidateQueries(eventAttendeesKey(eventId));` and not worry about the filter (`["eventAttendees", 1]` will match all queries that start with those two elements)

app/frontend/src/features/communities/events/EventUsers.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { Error as GrpcError } from "grpc-web";
+import makeStyles from "utils/makeStyles";
+
+import { SEE_ALL } from "./constants";
+
+const useStyles = makeStyles((theme) => ({
+  cardSection: {
+    padding: theme.spacing(2),
+  },
+  organisers: {
```
lucaslcode: users

app/frontend/src/features/communities/events/EventUsers.tsx
```diff
@@ -0,0 +1,80 @@
...
+    display: "grid",
+    marginBlockStart: theme.spacing(2),
+    rowGap: theme.spacing(1),
```
lucaslcode, line 18: It doesn't matter, but why use grid here?

app/frontend/src/queryKeys.ts
```diff
@@ -79,6 +81,32 @@ export const communityEventsKey = (communityId: number) => [
...
+export function isEventUsersInputType(
+  queryKey: unknown
+): queryKey is EventUsersInput {
+  return (
+    typeof queryKey === "object" &&
+    queryKey !== null &&
+    queryKey.hasOwnProperty("eventId") &&
+    queryKey.hasOwnProperty("type")
+  );
+}
```
lucaslcode: not a fan :P

app/frontend/src/service/events.ts
```diff
@@ -18,3 +25,55 @@ export async function listCommunityEvents(
...
+
+interface ListEventUsersInput {
+  eventId: number;
+  pageSize?: number;
+}
+
+export async function listEventOrganisers({
+  eventId,
+  pageSize,
+}: ListEventUsersInput) {
```
lucaslcode, line 46: You haven't included page token with these rpc calls so pagination won't work

lucaslcode (changes requested):

app/frontend/src/features/communities/events/hooks.ts
darrenvong: Good point. Although the full user objects are kinda related to the attendees query though, as in, we can't display anything in the UI without both pieces. So for organisation purposes, I prefer to have it as part of this custom hook. 

Maybe I can return the loading states (`isAttendeesLoading` and `isUsersLoading` or something similar) separately to reduce the problem you mentioned there, since in the case of more new users being added, I don't think we can avoid not having both hooks going back to the loading state (which is probably how I'm thinking of showing the loading indicator then)?

app/frontend/src/features/communities/events/EventPage.tsx
darrenvong: I believe it uses the user computer's timezone at the moment as opposed to what's set on their profile, as I believe that will need to be set explicitly on the `dayjs` object if we want to prefer using what we got in our database instead. I will have to double check. Do we want to always to stick to their profile's timezone instead I assume?

How do you think this should look on wider screens if it's not all going to be on one line? Wrap on the date like what I got now? That could be tricky to get right with everything all inline as the length could vary a lot depending on the localised format of how the date should look.

app/frontend/src/features/communities/events/EventAttendees.tsx
darrenvong: Hmm I thought about it, but it makes rendering the `EventUsers` on its own in Storybook a pain as I'd have to mock the full object `useInfiniteQuery` returns and there will be a tonne of unused props.

app/frontend/src/features/communities/events/hooks.ts
lucaslcode: In the case of new users added, it should remain "fetching" without changing what is displayed until the new users are loaded

app/frontend/src/features/communities/events/EventPage.tsx
lucaslcode: Not sure about timezones. 

I think letting it wrap anywhere is fine, just make the width smaller

app/frontend/src/features/communities/events/EventAttendees.tsx
lucaslcode: Don't you mock the service?
Well, whatever works

app/frontend/src/features/communities/events/hooks.ts
darrenvong: I'm not sure there's an easy way to tell whether something is refetching at the moment with `useUsers`, since I believe `isLoading` and `isFetching` will both be true regardless of if it's on first load or on refetch. Might have to refactor the hook slightly to compute this

app/frontend/src/queryKeys.ts
darrenvong: It's the only way to do it with more type safety but I probably won't need this function after going with your suggestion

app/frontend/src/service/events.ts
darrenvong: Good catch, building this with a broken API has really made these obviously broken things harder to spot! 🙈

app/frontend/src/features/communities/events/EventUsers.tsx
darrenvong: Less code to write for creating spacing between users (`rowGap` as opposed to the `& > * + *: { marginTop: theme.spacing(1) }`). I think I'm slowly being converted to the grid way of doing things in these cases 😂

app/frontend/src/features/communities/events/EventAttendees.tsx
darrenvong: `EventUsers` take in these props and since it's purely presentational (so I can share it between attendees and organisers), it only needs static data passed in and no React Query involved. So if I pass the whole query down as a prop, I'd have to write out that whole object rather only the props it needs.

app/frontend/src/features/communities/events/EventPage.tsx
darrenvong: I'll change the label, those were copied from the design. I think the "you are attending" subtext may not be needed as it will hopefully be more obvious when the API bug is fixed so your user is actually shown correctly under attendees!

app/frontend/src/features/communities/events/hooks.ts
darrenvong: Oh I lied, apparently `isLoading` remains false if data exists already, should be easier than expected then hopefully 🤞

darrenvong:
Well the lines count really blown up more than I thought... but most of this should have been reviewed already, this is what's actually changed since last review - https://github.com/Couchers-org/couchers/pull/1584/files/82288ddf7b3feb7b3126251f1d3e0a1b544cf93b..bc00d3ca90366b5fa6dd7816d4f8294b4bc805d9 (most additions are test code)

app/frontend/src/components/UserSummary.tsx
```diff
@@ -79,25 +84,30 @@ export default function UserSummary({
       noWrap={nameOnly}
     >
       {!user ? (
-        <Skeleton />
+        <Skeleton
+          className={classes.titleSkeleton}
```
lucaslcode, line 88: I usually just do `width={x}` for skeletons but doesn't matter

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -172,27 +184,30 @@ export default function EventPage() {
...
-                    {dayjs(timestamp2Date(event.startTime!)).format("LLLL")}
-                  </Typography>
-                  <Typography variant="body1">{TO}</Typography>
-                  <Typography variant="body1">
-                    {dayjs(timestamp2Date(event.endTime!)).format("LLLL")}
-                  </Typography>
-                </div>
+                <Typography align="center" variant="body1">
+                  {dayjs(timestamp2Date(event.startTime!)).format("LLLL")} {TO}{" "}
+                  {dayjs(timestamp2Date(event.endTime!)).format("LLLL")}
```
lucaslcode: Can this be LLLL if the dates are different, but just LT if they are the same?

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -172,27 +184,30 @@ export default function EventPage() {
...
-                <div>
-                  <Typography variant="body1">
-                    {dayjs(timestamp2Date(event.startTime!)).format("LLLL")}
-                  </Typography>
-                  <Typography variant="body1">{TO}</Typography>
-                  <Typography variant="body1">
-                    {dayjs(timestamp2Date(event.endTime!)).format("LLLL")}
-                  </Typography>
-                </div>
+                <Typography align="center" variant="body1">
```
lucaslcode:
```suggestion
                <Typography align="left" variant="body1">
```

lucaslcode (changes requested):
Minor stuff. Also, do you want to make the changes to the Event card in this PR?

darrenvong:
> do you want to make the changes to the Event card in this PR?

Uhh just found I might need some design input on this, as the gap problem you mentioned doesn't really work with the new design having the address/"view details for link" text right under the event title if it only takes up one line. Or I could just do it and have the gap in for the time being 🤷

darrenvong:
Ok updated the code to hide the attendees behind feature flag for now 👍

app/frontend/src/features/communities/events/EventCard.tsx
```diff
@@ -95,60 +112,42 @@ export default function EventCard({ event, className }: EventCardProps) {
...
-              </Typography>
-            </li>
-            <li>
-              <ClockIcon className={classes.icon} />
-              <Typography variant="body2" noWrap>
-                {`${dayjs(startTime).format("LT")} - ${dayjs(endTime).format(
-                  "LT"
+              <Typography variant="body1" noWrap>
+                {`${startTime.format("LLL")} - ${endTime.format(
+                  endTime.diff(startTime, "day") >= 1 ? "LLL" : "LT"
```
lucaslcode: Logic is not correct, 11pm to 1am the next day is 0 days difference.
```suggestion
                  endTime.isSame(startTime, "day") ? "LT" : "LLL"
```

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -97,6 +98,18 @@ export const useEventPageStyles = makeStyles((theme) => ({
...
+function getEventTimeString(
+  startTime: Timestamp.AsObject,
+  endTime: Timestamp.AsObject
+) {
+  const start = dayjs(timestamp2Date(startTime));
+  const end = dayjs(timestamp2Date(endTime));
+
+  return `${start.format("LLLL")} ${TO} ${end.format(
+    end.diff(start, "day") >= 1 ? "LLLL" : "LT"
+  )}`;
```
lucaslcode: See above comment, but you didn't use this function there?

lucaslcode (changes requested):

app/frontend/src/features/communities/events/EventCard.tsx
darrenvong: Ahh nice catch, I misread this in the doc and didn't realise you can also pass in a unit in these comparison functions 👍

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -63,6 +69,12 @@ export const useEventPageStyles = makeStyles((theme) => ({
   eventTitle: {
     gridArea: "eventTitle",
   },
+  onlineInfoContainer: {
+    display: "grid",
+    columnGap: theme.spacing(2),
+    gridAutoFlow: "column",
+    gridTemplateColumns: "max-content max-content",
```
lucaslcode, line 76: Is this the same as 1fr 1fr?

lucaslcode (approved):
Flakey tests :(

lucaslcode:
I re-ran the test CI and it passed...

app/frontend/src/features/communities/events/EventPage.tsx
darrenvong: Not quite, `1fr 1fr` would stretch that across the width of the title container and so leaves a big gap between "virtual event" and the "event link" text. I could have probably use flexbox for this... but I think I'm slowly becoming a fan of using `columnGap`/`rowGap` for defining spacing between child elements in a container.

darrenvong:
Which test was it that failed? Is it one of the new ones I wrote? 😅

lucaslcode:
I don't know, it was `createEvent` as usual

PR #1585, opened 2021-07-07
Final change "query" to "select"

ShmuelTreiger:
Some final changing of "query" to "select"

The first commit needs frontend changes

The second commit is gis untested stuff, so @aapeliv should probably look it over

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/proto/search.proto
+++ b/app/proto/search.proto
@@ -40,7 +40,7 @@ message SearchReq {
   // * `or` to search for one or another search term
   // * `-` (minus) to ignore a search term
   // in the fuzzy search these have no meaning
-  string query = 1;
+  string statement = 14;
 
   bool include_users = 4;
   bool include_communities = 5;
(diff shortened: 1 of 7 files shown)


app/proto/search.proto
```diff
@@ -40,7 +40,7 @@ message SearchReq {
   // * `or` to search for one or another search term
   // * `-` (minus) to ignore a search term
   // in the fuzzy search these have no meaning
-  string query = 1;
+  string statement = 14;
```
kalvdans: I don't think statement is a good name. Also I think we should change the protocol separately from internal changes in the backend.

kalvdans (changes requested):
Revert all .proto changes and make them as a separate pull request.

app/proto/search.proto
ShmuelTreiger: "statement" is what SQLA generally uses. SQL_statement?

kalvdans: There's no sql in our protocol between frontend and backend.

ShmuelTreiger: How about `search_string` or `search_term`?

kalvdans (approved):
Looks good to me! Thanks for switching to proper sql nomeclature :)

aapeliv (approved):

aapeliv:
LGTM

PR #1586, opened 2021-07-07
Renaming "query" within the api

ShmuelTreiger:
@kalvdans how's this

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/proto/conversations.proto
+++ b/app/proto/conversations.proto
@@ -239,7 +239,7 @@ message LeaveGroupChatReq {
 }
 
 message SearchMessagesReq {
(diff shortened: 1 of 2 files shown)


kalvdans (changes requested):
Use the develop branch as target please.
Keep the same numerical ID:s

I think query is good, most search engines call it query. But I'm happy with search_string too. Tag some english native speaker maybe for review :)

ShmuelTreiger:
You're right, leaving it as is makes more sense

PR #1593, opened 2021-07-07
Add meeting 27 notes

aapeliv:


kalvdans (approved):
Thanks for keeping notes!

PR #1594, opened 2021-07-07
No longer need to save profile to update avatar

lucaslcode:
Hopefully this will close the rest of the image issues by eliminating the need to save the form after changing the profile pic

I did not actually see if it works yet but the tests did!

**Frontend checklist**
- [ ] All tests pass


aapeliv (approved):
LGTM

PR #1596, opened 2021-07-07
Implement more admin APIs for moderation

aapeliv:
Requested by @AllhandsNoFeet 

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/query.py
+++ b/app/backend/src/couchers/query.py
@@ -38,6 +38,17 @@ class CouchersQuery(Query):
         # no fields match, this will return no rows
         return self.filter(False)
 
+    def filter_by_username_or_email_or_id(self, field):
+        # Should only be used for admin APIs, etc.
+        if is_valid_username(field):
+            return self.filter(User.username == field)
+        elif is_valid_email(field):
+            return self.filter(User.email == field)
+        elif is_valid_user_id(field):
+            return self.filter(User.id == field)
+        # no fields match, this will return no rows
+        return self.filter(False)
+
     def filter_users(self, context, table=User):
         """
         Filters out users that should not be visible: blocked, deleted, or banned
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/query.py
```diff
@@ -38,6 +38,17 @@ def filter_by_username_or_id(self, field):
         # no fields match, this will return no rows
         return self.filter(False)
 
+    def filter_by_username_or_email_or_id(self, field):
```
ShmuelTreiger, line 41: Should it just call the other two methods? Or we can refactor out "username" "id" and "email"? Seems kinda weird to have three overlapping methods like this.

app/backend/src/tests/test_admin.py
```diff
@@ -88,59 +88,96 @@ def test_AccessByNormalUser(db):
...
 
 
-def test_GetDeleteUser(db):
+def test_DeleteUser(db):
     with session_scope() as session:
         _generate_normal_user(session)
         _generate_super_user(session)
+        normal_user = _get_normal_user(session)
+
         with _admin_session(_get_super_token()) as api:
```
ShmuelTreiger, line 173: Is there anyway for an outside user to abuse this by generating a super user and using the token to make admin requests?

aapeliv: These are just testing helpers, making someone a superuser requires changing it directly in the database

app/backend/src/couchers/query.py
aapeliv: I can't reuse the other ones because this is doing first some python stuff to pick which `WHERE` to add and then adding that one. So you can't compose the other two to do this.

ShmuelTreiger (approved):

PR #1597, opened 2021-07-08
Add link to user's name in messages

### LINKED ISSUE #1046: Can't go to profile of chat partner
Subject: Can't go to profile of chat partner
Description:
I have written a message to someone. He didn't reply yet.
I clicked on his name on the top.

Results:
Nothing happened. I expected to be redirected to his profile.

Backend version: develop-d732f04e
Frontend version: develop-d732f04e
User Agent: Mozilla/5.0 (Linux; Android 11; Pixel 3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/89.0.4389.105 Mobile Safari/537.36
Page: https://app.couchers.org/messages/chats/64
User (spoofable): adventurebackpacker (168)

kerrylaval:
- Add .tsx extension to utils to support jsx
- Make each individual name its own link
- Refactor with if statement instead of ternary

closes #1046 

**Frontend checklist**
- [x] All tests pass **(date tests are failing, unrelated to this)**

### DIFF
--- a/app/frontend/src/features/messages/utils.tsx
+++ b/app/frontend/src/features/messages/utils.tsx
@@ -1,6 +1,8 @@
 import { hostRequestStatusLabels } from "features/messages/constants";
 import useUsers from "features/userQueries/useUsers";
 import { GroupChat, Message } from "proto/conversations_pb";
+import { Link } from "react-router-dom";
+import { routeToUser } from "routes";
 import { firstName } from "utils/names";
 
 export function isControlMessage(message: Message.AsObject) {
@@ -49,12 +51,17 @@ export function groupChatTitleText(
   groupChatMembersQuery: ReturnType<typeof useUsers>,
   currentUserId: number
 ) {
-  return groupChat.title
-    ? groupChat.title
-    : groupChatMembersQuery.isLoading
-    ? "Chat"
-    : Array.from(groupChatMembersQuery.data?.values() ?? [])
-        .filter((user) => user?.userId !== currentUserId)
-        .map((user) => firstName(user?.name))
-        .join(", ");
+
+
+  if (groupChat.title) {
+    return groupChat.title;
+  } else if (groupChatMembersQuery.isLoading) {
+    return "Chat";
+  }
+
+  return Array.from(groupChatMembersQuery.data?.values() ?? [])
+    .filter((user) => user?.userId !== currentUserId)
+    .map((user) => <Link to={routeToUser(user?.username || "")}>{firstName(user?.name)}</Link>)
+    .reduce((arr: any[] | null, element: any) => arr === null ? [element] : [...arr, ', ', element], null);
+  
 }


kerrylaval:
@aapeliv

app/frontend/src/features/messages/utils.tsx
```diff
@@ -49,12 +51,17 @@ export function groupChatTitleText(
...
+  if (groupChat.title) {
+    return groupChat.title;
+  } else if (groupChatMembersQuery.isLoading) {
+    return "Chat";
+  }
+
+  return Array.from(groupChatMembersQuery.data?.values() ?? [])
+    .filter((user) => user?.userId !== currentUserId)
+    .map((user) => <Link to={routeToUser(user?.username || "")}>{firstName(user?.name)}</Link>)
+    .reduce((arr: any[] | null, element: any) => arr === null ? [element] : [...arr, ', ', element], null);
```
kerrylaval, line 65: I know this reduce in here is a little funky, found it here https://stackoverflow.com/questions/33577448/is-there-a-way-to-do-array-join-in-react/44738250 maybe someone knows of a better way to join the list of links?

aapeliv:
I've tagged Lucas (frontend lead) and the frontend team to have a look :)

app/frontend/src/features/messages/utils.tsx
lucaslcode, line 65: It's good to avoid `any` if possible. Here we can avoid it by typing reduce: `.reduce<ReactNode[]>`, replace your `null` initial value with `[]` and then `arr.length === 0` instead of `arr === null`

lucaslcode (changes requested):
Small nit, and also could you run `yarn format`?

I am very sad that the date picker tests still aren't working in positive timezones :'(

darrenvong:
The fact this is displaying user names just like text, but are in fact clickable links, seems like a bad idea to me as that's introducing surprising behaviours, especially for people who are used to similar apps like Fb Messenger where people's names (as text) aren't clickable. 

They should really be styled as links, but I imagine that'd make the title look quite ugly / inconsistent to sometimes have it displayed as green and other times not. Is this really a good idea to implement? 🤔  They can click on the user as long as they've received a message from them from their avatar anyway...

app/frontend/src/features/messages/utils.tsx
darrenvong: I just noticed this would also render people's names as clickable links for chats with no title inside a chat item of the group chat list view. This would mean if the user clicked on the chat item but have their cursor too close to someone's name text (possible with a long list of names in large group chats), they'd end up going someone's profile instead of opening the chat which would be very confusing behaviour.

I think it's best to implement this inside `GroupChatView` if we were to keep this behaviour? Personally not a fan of having JSX outside of component files as that means you can't figure out what will be rendered by looking at the component alone.

lucaslcode: Good point, I missed that!!

lucaslcode:
We discussed elsewhere we aren't going to implement this

PR #1601, opened 2021-07-08
invalidate profile after changing avatar

lucaslcode:
also take the opportunity to move user query key

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/connections/friends/AddFriendButton.tsx
+++ b/app/frontend/src/features/connections/friends/AddFriendButton.tsx
@@ -3,6 +3,7 @@ import { PersonAddIcon } from "components/Icons";
 import { ADD_FRIEND } from "features/connections/constants";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { User } from "proto/api_pb";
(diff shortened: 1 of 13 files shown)


app/frontend/src/features/connections/friends/AddFriendButton.tsx
```diff
@@ -28,15 +29,15 @@ export default function AddFriendButton({
     onMutate: async ({ setMutationError }) => {
       setMutationError("");
 
-      await queryClient.cancelQueries(["user", userId]);
+      await queryClient.cancelQueries(userKey(userId));
 
       const cachedUser = queryClient.getQueryData<User.AsObject>([
         "user",
         userId,
```
darrenvong: Do `userKey` here as well?

darrenvong (changes requested):

app/frontend/src/features/connections/friends/AddFriendButton.tsx
lucaslcode: Line break got me

darrenvong (approved):

Issue #1560, opened 2021-07-03
Request pagination bug


aapeliv:
I think there's a bug in the request pagination where it's missing one request between pages.

aapeliv:
Fixed in #1604

PR #1604, opened 2021-07-09
Request ordering fix

aapeliv:
Start on fix for #1560

The requests are cut off based on request id but ordered according to last message id, which causes it to miss requests from the middle when paginated

**Backend checklist**
- [ ] All tests pass


lucaslcode (dismissed):

lucaslcode (approved):
Should probably be renamed to last_message_id but ok if you want to get it in

aapeliv:
> Should probably be renamed to last_message_id but ok if you want to get it in

Just didn't want to change the protocol layer since you've got the RFC going

PR #1619, opened 2021-07-12
Add functionality to see and set a user's birthdate in the admin API

aapeliv:
As requested by @AllhandsNoFeet 

**Backend checklist**
- [ ] All tests pass


app/backend/src/tests/test_admin.py
```diff
@@ -145,6 +149,26 @@ def test_ChangeUserGender(db):
...
+        assert res.banned == False
+        assert res.deleted == False
+
+
+def test_ChangeUserBirthdate(db):
+    with session_scope() as session:
+        _generate_normal_user(session)
+        _generate_super_user(session)
+        normal_user = _get_normal_user(session)
+
```
ShmuelTreiger, line 97: Should you add a check for the user's BD before changing it?

aapeliv: My thinking was that the admin is able to set it to anything, but maybe that's error prone (e.g. if they set a date in the future)

ShmuelTreiger: That's completely separate, but also a good point, there should probably be a check in there (i.e. no less than 18 years old)

What I'm asking, is should it be:
-check age == default
-set age
-check age == new age

Not strictly necessary, but I think good for clarity

aapeliv:
@ShmuelTreiger: could you do another review of this, please?

ShmuelTreiger (approved):
Sorry. Thought I looked at this already.

Your last commit should probably have been its own PR. I don't understand why you did it the way you did before and then changed it, but the way it is makes sense to me and I don't see any obvious bugs.

PR #1620, opened 2021-07-12
Fix ListEventAttendees

### LINKED ISSUE #1617: ListEventAttendees doesn't always return the current user correctly even when they are attending an event
Here are some rough reproduction steps:
1. Create an event
2. Transfer the event to a community (although this step probably not necessarily, only needed for it to show up in UI/`ListEvents` from `communities.proto`
3. Change the current user's attendance state to "not going" (with `SetEventAttendance`)
4. Change the current user's attendance state to "going" again

**Expected behaviour**
`ListEventAttendees` should return the current user's ID

**Actual/current behaviour**
`ListEventAttendees` returns another user's ID. This ID seems to be determined from the row's auto increment ID in `event_occurence_attendees` in the database

Basically can't release the event feature until this is fixed...

Other notes: I suspect `ListEventOrganizers` and `ListEventSubscribers` could also have this same problem, so probably worth double checking that too...

aapeliv:
This will eventually close #1617

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/events.py
+++ b/app/backend/src/couchers/servicers/events.py
@@ -530,8 +530,8 @@ class Events(events_pb2_grpc.EventsServicer):
                 .all()
             )
             return events_pb2.ListEventAttendeesRes(
-                attendee_user_ids=[attendee.id for attendee in attendees[:page_size]],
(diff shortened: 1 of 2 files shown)


kalvdans (approved):
Well done! Great with a test!

PR #1621, opened 2021-07-12
Add a way to skip CI checks and run only build & release

### LINKED ISSUE #1607: Add way to skip CI tests and checks in an emergency

aapeliv:
Closes #1607.

One can now trigger a CI run without running tests (that take around 15 min) by setting the `DO_CHECKS` variable to something other than `true`, e.g. from the gitlab "run pipeline" page.


kalvdans:
Somehow github doesn't understand it is mostly a rename, so in order to review the changes I looked at the seven first commits: https://github.com/Couchers-org/couchers/pull/1621/files/0a54b51100d717894f29f6b54d3f129f7d55dcdf..2857a10a85a9b5455843656d77668c0bf8737551

What's the reason behind this change? (You could already add  [skip ci] to the commit message to do it)

aapeliv:
> What's the reason behind this change? (You could already add [skip ci] to the commit message to do it)

The use case is when there is a bug we need to urgently fix, and we can easily run one test case locally to make sure the bug is fixed. We would then want to skip the ~15 min checks and just run the ~5 min build + release to get the containers built and re-deployed.

Adding `[skip ci]` will skip the whole CI process, so we don't release and can't deploy.

kalvdans:
Where it the settings page in gitlab CI? It would be good to understand where it is located even if I might not have access to it.

Will it affect all branches?

aapeliv:
> Where it the settings page in gitlab CI? It would be good to understand where it is located even if I might not have access to it.

To launch a new pipeline without checks, one can go to https://gitlab.com/couchers/couchers/-/pipelines/new, and override the var, the screen looks like this:

Screen Shot 2021-07-12 at 11 07 52 pm

(I can give you access if you wish.)

It's also possible to do it with git push options (which I never knew about), if you have a `gitlab` origin: `git push -o ci.variable="DO_CHECKS=no" gitlab`

> Will it affect all branches?

Yes, you can trigger a build while skipping checks on any branch, but you always need to have push/admin access on gitlab.

kalvdans:
So the workflow is this:

1. push to develop
2. cancel the started pipeline
3. manually start new pipeline from the gitlab frontend

or 

1. push to develop using the -o option

kalvdans (approved):
LGTM!

aapeliv:
> So the workflow is this:
> 
>     1. push to develop
>     2. cancel the started pipeline
>     3. manually start new pipeline from the gitlab frontend

It's probably good to just do 1 and 3, then 2 can run the checks in its own time just to double check.

> or
> 
>     1. push to develop using the -o option

Yes, but need to push to `gitlab`, I imagine push options are not sent from github to gitlab.

PR #1622, opened 2021-07-12
Add user to right communities on signup

### LINKED ISSUE #1606: Users should be added to communities straight after signing up
Currently it's every 15 min in a background thread.

I'm not sure how important this is, but I think it's quite important for the first "render" when they land on the dashboard. Opinions @ItsiW, @AllhandsNoFeet, @lucaslcode?

aapeliv:
Closes #1606

**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/tasks.py
```diff
@@ -300,3 +300,19 @@ def enforce_community_memberships():
...
+    nodes = session.execute(select(Node).where(func.ST_Contains(Node.geom, user.geom))).scalars().all()
+    for node in nodes:
+        logger.info(node.id)
+        node.official_cluster.cluster_subscriptions.append(
+            ClusterSubscription(
+                user=user,
+                role=ClusterRole.member,
+            )
+        )
+    session.commit()
```
ShmuelTreiger, line 345: Is there a way to combine this with `enforce_community_memberships`? For maintainability, I think it would be good to add all users to communities with the same method.

Alternatively, implement a test to make sure that both work the same (i.e. each adds one user to the same community)

app/backend/src/couchers/servicers/auth.py
```diff
@@ -237,8 +243,6 @@ def SignupFlow(self, request, context):
 
             # finish the signup if done
             if flow.is_completed:
-                # TODO: storing feedback forms
```
ShmuelTreiger, line 249: Why removal of todo? Leftover and unrelated?

app/backend/src/tests/test_communities.py
```diff
@@ -869,6 +881,51 @@ def test_LeaveCommunity_regression(db):
...
+        res = auth_api.SignupFlow(
+            auth_pb2.SignupFlowReq(
+                basic=auth_pb2.SignupBasic(name="testing", email="email@couchers.org.invalid"),
+                account=auth_pb2.SignupAccount(
+                    username="frodo",
+                    birthdate="1970-01-01",
+                    gender="Bot",
+                    hosting_status=api_pb2.HOSTING_STATUS_CAN_HOST,
+                    city="Country 1, Region 1, City 2",
+                    # == create_coordinate(8)
```
ShmuelTreiger: ?

app/backend/src/tests/test_communities.py
```diff
@@ -869,6 +881,51 @@ def test_LeaveCommunity_regression(db):
...
+                    accept_tos=True,
+                ),
+                feedback=auth_pb2.ContributorForm(),
+            )
+        )
+    with session_scope() as session:
+        email_token = (
+            session.execute(select(SignupFlow).where(SignupFlow.flow_token == res.flow_token)).scalar_one().email_token
+        )
+    with auth_api_session() as (auth_api, metadata_interceptor):
```
ShmuelTreiger, line 913: parentheses necessary?

app/backend/src/tests/test_communities.py
```diff
@@ -869,6 +881,51 @@ def test_LeaveCommunity_regression(db):
...
+    with session_scope() as session:
+        w_id = get_community_id(session, "Global")
+        c1_id = get_community_id(session, "Country 1")
+        c1r1_id = get_community_id(session, "Country 1, Region 1")
+        c1r1c2_id = get_community_id(session, "Country 1, Region 1, City 2")
+
+    token = get_session_cookie_token(metadata_interceptor)
+
+    with communities_session(token) as api:
+        res = api.ListUserCommunities(communities_pb2.ListUserCommunitiesReq())
```
ShmuelTreiger, line 927: Why don't you need like empty here?

ShmuelTreiger (changes requested):

app/backend/src/couchers/tasks.py
aapeliv: > Is there a way to combine this with `enforce_community_memberships`? For maintainability, I think it would be good to add all users to communities with the same method.

We kind of could, I suppose. The other one is `for each node ... for each user not yet added ...`, whereas this is `for each node ...`, so there's not really any way to run this without running for each user (regardless of if added or not), so it'd do a tonne of extra work.

app/backend/src/couchers/servicers/auth.py
aapeliv: Yeah, leftover and unrelated

app/backend/src/tests/test_communities.py
aapeliv: Meaning that the next two `lat=...,lng=...` are equivalent to creating the point with `create_coordinate(8)`

app/backend/src/tests/test_communities.py
aapeliv: Yep, otherwise it's expecting a second `with` block

app/backend/src/tests/test_communities.py
aapeliv: `ListUserCommunitiesReq` optionally takes pagination stuff

app/backend/src/couchers/tasks.py
ShmuelTreiger: You could pull out the guts and have nodes and users as input. Or even just users

app/backend/src/tests/test_communities.py
ShmuelTreiger: Could you flesh that out a bit better? I didn't understand at all.

app/backend/src/tests/test_communities.py
ShmuelTreiger: Isn't it still empty? Just trying to understand, because I've had api calls fail for lack of a request before.

aapeliv: I think it needs something inside `ListUserCommunities(...)`, the grpc python client might be picky towards what you put in there, but on the wire; if this is empty, it'd be equivalent to `api.ListUserCommunities(empy_pb2.Empty())`.

app/backend/src/tests/test_communities.py
aapeliv: I'll add a comment

app/backend/src/couchers/tasks.py
aapeliv: I think it's good as is; they're right next to each other and it's pretty easy to maintain. Our tests basically verify that both are correct.

ShmuelTreiger (dismissed):
LGTM

PR #1623, opened 2021-07-12
Implement reference reminder emails

### LINKED ISSUE #893: Email reminders to write a reference
Emails to remind a host/surfer to write a reference about the experience.

aapeliv:
Closes #893

**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -221,6 +231,90 @@ def process_send_onboarding_emails(payload):
...
+                .outerjoin(
+                    Reference,
+                    and_(
+                        Reference.host_request_id == HostRequest.conversation_id,
+                        # if no reference is found, then HostRequest.from_user_id has not written a ref (i.e. the surfer)
+                        Reference.from_user_id == HostRequest.from_user_id,
+                    ),
+                )
+                .where(user.is_visible)
+                .where(other_user.is_visible)
```
ShmuelTreiger, line 271: Why not the normal filters? If user has since been banned or deleted, since we have no recovery, there's no use leaving a reference for a non-existent user?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -221,6 +231,90 @@ def process_send_onboarding_emails(payload):
...
+                union.c[0].label("surfed"),
+                aliased(HostRequest, union),
+                aliased(user, union),
+                aliased(other_user, union),
+            )
+            reference_reminders = session.execute(union).all()
+
+            for surfed, host_request, user, other_user in reference_reminders:
+                # checked in sql
+                assert user.is_visible
```
ShmuelTreiger, line 310: Seems redundant?

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -35,6 +36,7 @@
     requests_session,
     testconfig,
 )
+from tests.test_references import create_host_reference, create_host_request
```
ShmuelTreiger, line 40: Should we move to utils?

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -649,6 +651,105 @@ def test_process_send_onboarding_emails(db):
...
+    for _ in range(4):
+        process_job()
+
+    with session_scope() as session:
+        emails = [
+            (email.recipient, email.subject)
+            for email in session.execute(select(Email).order_by(Email.recipient.asc())).scalars().all()
+        ]
+
+        print(emails)
```
ShmuelTreiger, line 731: Should this be removed?

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -649,6 +651,105 @@ def test_process_send_onboarding_emails(db):
         )
 
 
+def test_process_send_reference_reminders(db):
+    # need to test:
+    # case 1: bidirectional (no emails)
+    # case 2: host left ref (surfer needs an email)
+    # case 3: surfer left ref (host needs an email)
+    # case 4: neither left ref (host & surfer need an email)
```
ShmuelTreiger, line 660: Seems to me this should be broken up into 4 tests, each of which would tell us if its one case is broken.

app/backend/templates/reference_reminder.md
```diff
@@ -0,0 +1,17 @@
...
+
+{% from "macros.html" import button, link, support_email %}
+
+Hi {{ user.name|couchers_escape }}!
+
+You have {{ time_left_text|couchers_escape }} left to write a reference for {{ other_user.name|couchers_escape }} from when you {% if surfed %} surfed with {% else %} hosted {% endif %} them.
+
+To write a reference, go to the Messaging tab and locate the request, then click on the "Write Reference" button at the bottom of the conversation.
+
+It's a nice gesture to write references and helps us build a community together! References will become visible 2 weeks after the stay, or when you've both written a reference for each other.
```
ShmuelTreiger: ", whichever happens first."

app/backend/templates/reference_reminder.md
```diff
@@ -0,0 +1,17 @@
...
+subject: "You have {{ time_left_text|couchers_escape }} to write a reference for {{ other_user.name|couchers_escape }}!"
+---
+
+{% from "macros.html" import button, link, support_email %}
+
+Hi {{ user.name|couchers_escape }}!
+
+You have {{ time_left_text|couchers_escape }} left to write a reference for {{ other_user.name|couchers_escape }} from when you {% if surfed %} surfed with {% else %} hosted {% endif %} them.
+
+To write a reference, go to the Messaging tab and locate the request, then click on the "Write Reference" button at the bottom of the conversation.
```
ShmuelTreiger: Link to messaging (or the correct message even better if possible)?

Also, should available references eventually get its own page on the website?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -221,6 +231,90 @@ def process_send_onboarding_emails(payload):
...
+                .join(user, user.id == HostRequest.to_user_id)
+                .join(other_user, other_user.id == HostRequest.from_user_id)
+                .outerjoin(
+                    Reference,
+                    and_(
+                        Reference.host_request_id == HostRequest.conversation_id,
+                        # if no reference is found, then HostRequest.to_user_id has not written a ref (i.e. the host)
+                        Reference.from_user_id == HostRequest.to_user_id,
+                    ),
+                )
```
ShmuelTreiger: I'm having trouble understanding this. Would love if you could explain it to me.

app/backend/src/couchers/jobs/handlers.py
aapeliv: The normal filters check if one user blocked the other, but since there is no signed in user running background jobs, we can't use those filters.

Also just noticed a bug here: this does not respect blocking.

app/backend/src/couchers/jobs/handlers.py
aapeliv: Yeah, just double checking

app/backend/src/tests/test_bg_jobs.py
aapeliv: Do we have a utils for testing?

I don't really mind, the testing code isn't that much of a mess, and i have a lower bar for its quality

app/backend/src/tests/test_bg_jobs.py
aapeliv: Yes, sorry, leftover. Not so critical though since it's just in testing

app/backend/src/tests/test_bg_jobs.py
aapeliv: Part of the test is to make sure that it does it fine when there's a set of emails to send and not just one-off.

Anyway, it's faster to combine them into one unit test, and not too difficult if it breaks to figure out which one broke

app/backend/src/couchers/jobs/handlers.py
aapeliv: You take all host requests, then you try to join references on it where the reference host request is this host request, and the user who left the reference is the host (`HostRequest.to_user_id`). Then you get rid of the rows where the join found a reference (since they need no reminders)

app/backend/templates/reference_reminder.md
aapeliv: > Link to messaging (or the correct message even better if possible)?

It's possible but a bit annoying. I'm hoping soon we'll have pending references to write on the dashboard, so we can just link there.

ShmuelTreiger: I think a link to just messaging for now then. It's kind of annoying "Reminder to do this thing" and not even provide a general link to do it

app/backend/src/tests/test_bg_jobs.py
ShmuelTreiger: test fixtures

app/backend/src/tests/test_bg_jobs.py
ShmuelTreiger: Not critical, but a leftover print threw me off for a bit once

app/backend/src/tests/test_bg_jobs.py
ShmuelTreiger: Theoretically should be 5 tests then, but OK

app/backend/src/couchers/sql.py
```diff
@@ -16,6 +16,20 @@ def _relevant_user_blocks(user_id):
...
+        couchers_select(UserBlock.blocked_user_id)
+        .where(UserBlock.blocking_user_id == user1_id)
+        .where(UserBlock.blocked_user_id == user2_id)
+    )
+    blocking_users = (
+        couchers_select(UserBlock.blocking_user_id)
+        .where(UserBlock.blocking_user_id == user2_id)
+        .where(UserBlock.blocked_user_id == user1_id)
+    )
+    return session.execute(couchers_select(union(blocked_users, blocking_users).subquery())).first() is not None
```
ShmuelTreiger: Utils maybe?

app/backend/src/couchers/jobs/handlers.py
ShmuelTreiger: ># if no reference is found, then HostRequest.to_user_id has not written a ref (i.e. the host)

Is confusing. Instead, I would maybe add a note by `.where(Reference.id == None)` something like "excludes HostRequests where host already left Reference"

I would also change it to a normal (left) join

aapeliv: Added a slightly better comment. `outerjoin` is left join in SA: https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.outerjoin

app/backend/templates/reference_reminder.md
```diff
@@ -8,7 +8,17 @@ Hi {{ user.name|couchers_escape }}!
 
 You have {{ time_left_text|couchers_escape }} left to write a reference for {{ other_user.name|couchers_escape }} from when you {% if surfed %} surfed with {% else %} hosted {% endif %} them.
 
-To write a reference, go to the Messaging tab and locate the request, then click on the "Write Reference" button at the bottom of the conversation.
+Leavea a reference here:
```
ShmuelTreiger: Typo "Leavea"

ShmuelTreiger (commented):
I still find the other comment confusing, but if you change `# surfing reqs` to something like `# surfers who haven't written a reference` it's clearer for future users.

Alternatively, what if we somehow reverse the join and remove `.where(Reference.id == None)`? Then you could change the comment to be a positive statement `# if reference is found in this join, then the surfer (HostRequest.from_user_id) has already written a ref`

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -221,6 +232,91 @@ def process_send_onboarding_emails(payload):
...
+                    Reference,
+                    and_(
+                        Reference.host_request_id == HostRequest.conversation_id,
+                        # if no reference is found in this join, then the surfer (HostRequest.from_user_id) has not written a ref
+                        Reference.from_user_id == HostRequest.from_user_id,
+                    ),
+                )
+                .where(user.is_visible)
+                .where(other_user.is_visible)
+                .where(Reference.id == None)
```
ShmuelTreiger, line 272: It's fine as is, but I would add here "# i.e. no reference found"

app/backend/src/couchers/servicers/blocking.py
```diff
@@ -8,6 +9,20 @@
 from proto import blocking_pb2, blocking_pb2_grpc
 
 
+def are_blocked(session, user1_id, user2_id):
```
ShmuelTreiger, line 12: If you're moving this here, should we move `_relevant_user_blocks` as well?

ShmuelTreiger: We might also move `are_friends` to api.py

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -649,6 +652,88 @@ def test_process_send_onboarding_emails(db):
...
+    # case 5: neither left ref, but host blocked surfer, so neither should get an email
+    # host
+    user9, token9 = generate_user(email="user9@couchers.org.invalid", name="User 9")
+    # surfer
+    user10, token10 = generate_user(email="user10@couchers.org.invalid", name="User 10")
+
+    make_user_block(user9, user10)
+
+    with session_scope() as session:
+        # note host requests have one day longer age
```
ShmuelTreiger, line 695: I don't understand this whole commit: the comment, the date changes, or why you removed all the check/asserts

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -649,6 +652,88 @@ def test_process_send_onboarding_emails(db):
...
+        pass
+
+    with session_scope() as session:
+        emails = [
+            (email.recipient, email.subject)
+            for email in session.execute(select(Email).order_by(Email.recipient.asc())).scalars().all()
+        ]
+
+        print(emails)
+        print(expected_emails)
```
ShmuelTreiger, line 732: Leftover prints

ShmuelTreiger (changes requested):

app/backend/src/couchers/servicers/blocking.py
aapeliv: Import errors. Can't import models in utils, because models imports utils; this is the next best place. I just want this code to be nearby to other uses of `UserBlock` so it gets the right treatment when refactoring.

app/backend/src/tests/test_bg_jobs.py
aapeliv: > the comment

`create_host_reference` creates a reference where the actual request is one day older than the input `reference_age`

> the date changes

Noticed a bug where the reference was put in the db as having been created before the host request was created. Oops.

> why you removed all the check/asserts

I was making sure there were 4 emails sent. But my `expected_emails` is of length 4, so it gets checked anyway.

app/backend/src/tests/test_bg_jobs.py
aapeliv: I left them on purpose this time, it's useful for debugging.

app/backend/src/couchers/servicers/blocking.py
ShmuelTreiger: That all makes sense. I Think we should move the other two for consistency. Doesn't really matter though.

app/backend/src/tests/test_bg_jobs.py
ShmuelTreiger: "Note, host requests are one day older than the reference"

Something like that?

aapeliv:
I can remove the comments if they add more confusion than they add clarity.

ShmuelTreiger:
The one day older is confusing, up to you if you think it's necessary. The others I would leave/clarify

Issue #1627, opened 2021-07-12
Account deletion


jesseallhands:
For better or for worse, one of the main reasons people contact support is to ask for their account to be deleted. Most of these requests are due to duplicate accounts. We should add a "delete account" option under account settings so people can request a deletion instead of e-mailing us (which takes up a lot of support's time).

On the account settings page below the password change section, there should be the following text:

`If you would like to delete you account click **here**.`

When you click on "here" then there should be a dialog box that appears with the following:

Delete account?

Are you sure you want to delete your account? This cannot be undone.

Reason (optional):
[ Free text input box where user can input reason for cancelling] 

[Delete account forever]    [Cancel]

Text input into the box for the reason should be sent to notify@couchers.org so we can gather data on why users are deleting their account.

ShmuelTreiger:
I prefer typing a confirmation text message instead of a button (i.e. please type 'allhandsnofeet', case insensitive)

dommers-ux:
design idea to spark more ideas 

a **Delete Account** section at the bottom of account settings 
<img width="313" alt="Screen Shot 2021-07-25 at 6 39 43 PM" src="https://user-images.githubusercontent.com/83376913/126918995-ddb96883-5446-40ab-ab09-ae324b6ef0fa.png">


followed by a modal pop up 
<img width="387" alt="Screen Shot 2021-07-25 at 6 39 57 PM" src="https://user-images.githubusercontent.com/83376913/126919001-27f4a8eb-ad49-4440-8b24-cf0c5d60206d.png">

dommers-ux:
followed by 
<img width="484" alt="Screen Shot 2021-07-25 at 7 07 47 PM" src="https://user-images.githubusercontent.com/83376913/126920173-908e0a87-a8fa-47e8-8910-b5469b8cd37e.png">

lucaslcode:
Please keep GitHub issues for bugs and features which are requested/ready for devs to implement. I have added the "awaiting design" label for now and when we have a finalized design let's remove it and add a link to a finalized mock-up.

aapeliv:
Also have a look at what the procedure will be, as discussed in last backend meeting:

https://github.com/Couchers-org/couchers/blob/develop/docs/meetings/backend/Meeting%20028.md

aapeliv:
Closed in #1630

PR #1630, opened 2021-07-13
Implement self-service account deletion

ShmuelTreiger:
For issue #1627, adds basic self deletion to the API


aapeliv:
This should take some message that confirms the deletion. (E.g. just a boolean, or as you said, the user's username)

aapeliv:
Actually, we should confirm this by email.

ShmuelTreiger:
That's a good call. Type their username into the front end and then use a deletion token in the backend

aapeliv (changes requested):
We should do a few things for this feature:

1. The backend needs to do some verification of the user's intent. I think the user should pass in their username that the backend then checks, ignoring case.
2. There should be an email confirmation
3. If the user wishes to delete their account, there should be a 48 hour delay before it actually happens, so that they can cancel it if it was done in haste, etc
4. There should be a question + field as to why the user deleted their account (optional)
5. We should have a think about data exports as well so people can download their data before deleting their user

ShmuelTreiger:
\1) I figured the frontend would do that, but sure
\4) Do we want to store that in the backend?
\5) Yes, but separate issue

kalvdans:
1. "are you sure" questions belong in the frontend.
2+3. Send an email with a revival link, valid for 48 hours. Or email admin until we have revival in place.

ShmuelTreiger:
>2+3. Send an email with a revival link, valid for 48 hours. Or email admin until we have revival in place.

I like that much better

aapeliv:
2- If we're going to be friendly towards use of our APIs through non-official frontend, then we need to include an email confirmation. If that's not something we care about, then we can do it without email confirmation.

1- Either way, you need to at least include a `bool` "confirmed" or similar to trigger account deletion. I don't want just POSTing to this address to delete the account, it's too risky.


>  \4) Do we want to store that in the backend?

Yes, we store everything in the backend.

kalvdans:
> 2- If we're going to be friendly towards use of our APIs through non-official frontend, then we need to include an email confirmation.

To help develop non-official frontends, we should add a parameter that simulates a deletion, i.e. responds with the same value and kicks you out from the session.

aapeliv:
> To help develop non-official frontends, we should add a parameter that simulates a deletion, i.e. responds with the same value and kicks you out from the session.

That's for developing them. But there are some actions, such as account deletion, that we want the user to positively confirm outside of the third-party use, hence the email.

ShmuelTreiger:
Need to add undelete functionality (as an email and as an admin tool), will have to wait for next week

app/backend/src/couchers/servicers/account.py
```diff
@@ -267,3 +271,80 @@ def VerifyPhone(self, request, context):
...
+                session.delete(existing_token)
+                session.flush()
+
+            token = send_account_deletion_confirmation_email(user)
+            session.add(token)
+
+            regex = re.compile("[^a-zA-Z]")
+
+            if len(regex.sub("", request.reason)) > 0:  # Discount any messages without real text
+                reason_for_deletion = ReasonForDeletion(user_id=user.id, reason=request.reason)
```
ShmuelTreiger: Is there a chance here for injection? What's the proper way to sanitize? Or does the front end do that?

app/backend/src/couchers/servicers/account.py
```diff
@@ -267,3 +271,80 @@ def VerifyPhone(self, request, context):
...
+            account_deletion_token = session.execute(
+                select(AccountDeletionToken)
+                .where(AccountDeletionToken.user_id == context.user_id)
+                .where(AccountDeletionToken.token == request.token)
+                .where(AccountDeletionToken.expiry > now())
+                .where(AccountDeletionToken.end_time_to_recover < now())
+            ).scalar_one_or_none()
+
+            if not account_deletion_token:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.INVALID_TOKEN)
```
ShmuelTreiger: Should this return success=False instead of aborting?

app/backend/src/couchers/servicers/account.py
```diff
@@ -267,3 +271,80 @@ def VerifyPhone(self, request, context):
...
+
+            user = account_deletion_token.user
+            if not user.is_deleted:
+                context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.USER_NOT_DELETED)
+
+            if user.has_password and not verify_password(user.hashed_password, request.password):
```
ShmuelTreiger: Wasn't sure how this should work for a user without a password. Current behavior is to not check anything (i.e. anyone can recover). Since this is already coming from an email, I thought it silly to do another additional email.

app/backend/src/couchers/models.py
```diff
@@ -628,6 +628,29 @@ def __repr__(self):
...
+
+    token = Column(String, primary_key=True)
+
+    user_id = Column(ForeignKey("users.id"), nullable=False, index=True, unique=True)
+
+    created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
+    expiry = Column(DateTime(timezone=True), nullable=False, server_default=func.now() + text("interval '2 hours'"))
+    end_time_to_recover = Column(
+        DateTime(timezone=True), nullable=True, default=func.now()
+    )  # stars off as now, set to 48 hours when account is deleted
```
ShmuelTreiger: I'm essentially double-using the one token, since it's all part of one process. We could instead create an account recovery token after an account is successfully deleted, but I thought this more efficient.

app/backend/templates/signup_verify.md
```diff
@@ -19,6 +19,6 @@ To finish setting up your account, please click on this link to confirm your ema
 
 {% endif %}
 
-See you in a bit :).
+See you in a bit :)
```
ShmuelTreiger, line 22: This was supposed to be in my next pull request. Oops

ShmuelTreiger (commented):
Apparently migrations are broken. I'll fix tomorrow

app/backend/src/couchers/servicers/account.py
aapeliv: Just do a `request.reason.strip()`; no need to pull in regex. Other than that, SA takes care of sql injection, so no worries

app/proto/account.proto
```diff
@@ -52,6 +52,19 @@ service Account {
...
+  }
+
+  rpc DeleteAccount(DeleteAccountReq) returns (DeleteAccountRes){
+  // Deletes active user's account
+  // Cannot be used to delete any other user's account
+  }
+
+  rpc RecoverAccount(RecoverAccountReq) returns (RecoverAccountRes){
+    // Password required only if user has password
+  }
```
kalvdans:
```suggestion
  rpc RequestAccountDeletion(RequestAccountDeletionReq) returns (google.protobuf.Empty) {
    // Sends email with confirmation link containing token
  }

  rpc DeleteAccount(DeleteAccountReq) returns (DeleteAccountRes) {
    // Deletes active user's account
    // Cannot be used to delete any other user's account
  }

  rpc RecoverAccount(RecoverAccountReq) returns (RecoverAccountRes) {
  }
```

aapeliv:
Screenshot from 2022-02-07 00-53-58

app/backend/src/couchers/models.py
```diff
@@ -164,6 +164,9 @@ class User(Base):
     is_deleted = Column(Boolean, nullable=False, server_default=text("false"))
     is_superuser = Column(Boolean, nullable=False, server_default=text("false"))
 
+    undelete_token = Column(String, nullable=True)
+    undelete_until = Column(DateTime(timezone=True), nullable=True)
```
kalvdans: Document these. Actually I like storing all necessary information in the link better, like you did in https://github.com/Couchers-org/couchers/pull/2475 . Less state to keep track of in the database.

app/backend/src/tests/test_account.py
```diff
@@ -857,24 +840,23 @@ def test_RequestAccountDeletion(db):
 
         # first two asserts also imply created < now() and expiry > now()
         assert deletion_token.is_valid
-        assert deletion_token.end_time_to_recover < now()
```
kalvdans: Update comment

app/proto/auth.proto
```diff
@@ -102,6 +102,11 @@ service Auth {
     // Triggered when the user goes to the link sent in the either email_changed_confirmation_*_email
   }
 
+  rpc ConfirmDeleteAccount(ConfirmDeleteAccountReq) returns (google.protobuf.Empty) {
+    // Deletes active user's account
+    // Cannot be used to delete any other user's account
```
kalvdans: No, the user indicated by `token` is deleted, regardless of who is logged in.

app/web/features/auth/deletion/DeleteAccount.tsx
```diff
@@ -0,0 +1,121 @@
...
+        lowercaseAndTrimField(confirmUsername) ===
+        lowercaseAndTrimField(username);
+      if (!confirm) {
+        throw Error(t("auth:delete_account.request.username_mismatch"));
+      }
+      return service.account.deleteAccount(confirm, reason);
+    },
+    {
+      onSuccess: () => {
+        resetForm();
```
lucaslcode, line 61: redirect to `logoutRoute`? Or what happens next time they try to make an api call?

lucaslcode (changes requested):
Except for this question, frontend lgtm

app/backend/src/tests/test_account.py
aapeliv: Yes, sorry; old comments

app/backend/src/couchers/models.py
aapeliv: These things are good to keep in the DB: they're quite a bit more critical and we need these for debugging and visibility!

app/web/features/auth/deletion/DeleteAccount.tsx
aapeliv: Yes, good idea. Though that's in `ConfirmDeleteAccount.tsx`

app/backend/src/couchers/models.py
aapeliv: Documented

PR #1631, opened 2021-07-13
[MRG] Add DeleteAccount to backend

ShmuelTreiger:
My thinking is that the front end should verify the identity before requesting account deletion and that a request only adds in a possibility of passing in the wrong information. However, it would be trivial to add a request with username and a check that `request.username.lower() == user.username.lower()`

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/account.py
+++ b/app/backend/src/couchers/servicers/account.py
@@ -266,3 +266,16 @@ class Account(account_pb2_grpc.AccountServicer):
             user.phone_verification_attempts = 0
 
         return empty_pb2.Empty()
+
+    def DeleteAccount(self, request, context):
+        """
+        Cannot be used to delete any account, except for the user's own
+
+        Extensive checking should be confirmed by the frontend
+        """
+        with session_scope() as session:
+            user = session.execute(select(User).where(User.id == context.user_id)).scalar_one()
+            user.is_deleted = True
+            session.commit()
+
+        return empty_pb2.Empty()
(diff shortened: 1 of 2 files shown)


aapeliv:
Please combine into one PR

app/backend/src/couchers/servicers/account.py
```diff
@@ -266,3 +266,16 @@ def VerifyPhone(self, request, context):
...
+    def DeleteAccount(self, request, context):
+        """
+        Cannot be used to delete any account, except for the user's own
+
+        Extensive checking should be confirmed by the frontend
+        """
+        with session_scope() as session:
+            user = session.execute(select(User).where(User.id == context.user_id)).scalar_one()
+            user.is_deleted = True
+            session.commit()
```
kalvdans, line 279: The commit is automatic when the `with` context ends, but it might be good to do it explicitly like you do here anyway.

kalvdans (approved):
Looks good with or without my little comment.

aapeliv:
This still needs the confirmation.

ShmuelTreiger:
> Please combine into one PR

@kalvdans said API changes should be on their own?

kalvdans:
> Please combine into one PR

With frontend implementation or?

ShmuelTreiger:
@kalvdans I think he means with this one

https://github.com/Couchers-org/couchers/tree/api/feat/delete-account

aapeliv:
Yes, I meant combine the backend change with the API change proposal.

kalvdans:
Oh, I thought the API was already there and this PR targeted develop. Yes please join the two branches into one PR against develop.

In my comment about API earlier, the changes were independent of each other, here they are dependent and should be done in one go.

PR #1632, opened 2021-07-13
[WIP] Sort regions alphabetically

ShmuelTreiger:
This has been bothering me

Downside: extra work on the backend for every user call [O(n) where n is number of countries ~200, so minimal work, but work nonetheless] for essentially no real gain

Upside: Looks neater and satisfies my OCD

Potential future improvement: Store them sorted in the database. People don't tend to visit new countries *too* often, and it's even rarer that they unvisit them, so one time work per user/insert for permanent sorting. Honestly, every user could have a bit vector of size ~200. Might even be more space efficient than the current built-out model, though would require non-trivial dev work every time a country is added to/removed from existence.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -840,8 +840,8 @@ def user_model_to_pb(db_user, session, context):
             api_pb2.LanguageAbility(code=ability.language_code, fluency=fluency2api[ability.fluency])
             for ability in db_user.language_abilities
         ],
-        regions_visited=sorted([region.region_code for region in db_user.regions_visited]),
-        regions_lived=sorted([region.region_code for region in db_user.regions_lived]),
+        regions_visited=[region.region_code for region in db_user.regions_visited],
+        regions_lived=[region.region_code for region in db_user.regions_lived],
         additional_information=db_user.additional_information,
         friends=friends_status,
         pending_friend_request=pending_friend_request,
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/servicers/api.py
```diff
@@ -840,8 +840,8 @@ def user_model_to_pb(db_user, session, context):
             api_pb2.LanguageAbility(code=ability.language_code, fluency=fluency2api[ability.fluency])
             for ability in db_user.language_abilities
         ],
-        regions_visited=[region.region_code for region in db_user.regions_visited],
-        regions_lived=[region.region_code for region in db_user.regions_lived],
+        regions_visited=sorted([region.region_code for region in db_user.regions_visited]),
```
kalvdans: You could omit the [], avoiding creating an intermediate array.
```
         regions_visited=sorted(region.region_code for region in db_user.regions_visited),
```

kalvdans (dismissed):
Looks good, with or without my comment!

aapeliv:
Doing it in the backend is fine since the number of regions is upper bounded. We could also just do it in the database (you can define the sort order in the SA relationship)

ShmuelTreiger:
> Doing it in the backend is fine since the number of regions is upper bounded. We could also just do it in the database (you can define the sort order in the SA relationship)

Oh cool, I'm going to do that. I'll get to learn something new.

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: Yup! Good call

aapeliv (approved):

kalvdans:
> O(n) where n is number of countries

sorting is O(n log n)

sorry my OCD just needed to get this corrected :p

ShmuelTreiger:
😂 Fair point.

But really, it's constant time, since n ~200

PR #1633, opened 2021-07-13
Renaming HostRequest to_user --> host and from_user --> surfer

ShmuelTreiger:
This got me all confused when reviewing Aapeli's pull request, so I went ahead and refactored it (as briefly mentioned in the slack channel)

This will need frontend work as well

Also, unsure if the migrations will work as the test is broken on my machine, so well see

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [ ] All tests pass


aapeliv:
Could we merge https://github.com/Couchers-org/couchers/pull/1623 first?

ShmuelTreiger:
No problem. It will need frontend changes anyway.

aapeliv:
Btw, there's no need to change the tag number in protos if the meaning and type stay the same and you're just changing the name

ShmuelTreiger:
Had no idea, thanks!

ShmuelTreiger:
Done/ready for front end post #1623

kalvdans (dismissed):
Very nice clarification of the roles. redo the migrations after #1623 has been merged.

app/backend/src/couchers/migrations/versions/d1a68f873ee3_rename_hostrequest_user_columns.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_constraint("fk_host_requests_from_user_id_users", "host_requests", type_="foreignkey")
+    op.drop_constraint("fk_host_requests_to_user_id_users", "host_requests", type_="foreignkey")
+    op.create_foreign_key(op.f("fk_host_requests_host_id_users"), "host_requests", "users", ["host_id"], ["id"])
+    op.create_foreign_key(op.f("fk_host_requests_surfer_id_users"), "host_requests", "users", ["surfer_id"], ["id"])
+    op.drop_index("ix_host_requests_from_user_id", table_name="host_requests")
+    op.drop_index("ix_host_requests_to_user_id", table_name="host_requests")
+    op.create_index(op.f("ix_host_requests_host_id"), "host_requests", ["host_id"], unique=False)
+    op.create_index(op.f("ix_host_requests_surfer_id"), "host_requests", ["surfer_id"], unique=False)
```
aapeliv: Are you not able to just rename the indexes/constraints? This is fine-ish, but creating constraints/indexes can take quite a long time on larger databases and might take a whole table write lock

ShmuelTreiger: It looks like no and that its an sql thing, not an sqla thing:

https://stackoverflow.com/questions/44631175/alembic-how-to-alter-a-foreigenkey-column 
https://stackoverflow.com/questions/13606469/cannot-change-column-used-in-a-foreign-key-constraint

And yes it requires a table lock (didn't even think of this). Will sqla do that automatically? Or do I need to add it?

ShmuelTreiger:
Ready for review and then frontend implementation

app/backend/src/couchers/migrations/versions/d1a68f873ee3_rename_hostrequest_user_columns.py
aapeliv: You can rename them in postgres:

https://www.postgresql.org/docs/current/sql-alterindex.html

https://www.postgresql.org/docs/current/sql-altertable.html

app/proto/requests.proto
```diff
@@ -40,7 +40,7 @@ service Requests {
 }
 
 message CreateHostRequestReq {
-  int64 to_user_id = 1;
+  int64 host_id =1;
```
aapeliv: Space

aapeliv (commented):
Did you fix the new instances?

ShmuelTreiger:
From #1623? Yes

app/backend/src/couchers/migrations/versions/d1a68f873ee3_rename_hostrequest_user_columns.py
ShmuelTreiger: I renamed them just now without thinking. It's not about renaming them, it's about applying the same constraint to a newly named column. I don't see a way to do that?

aapeliv: What do you mean? The column didn't change, just the name. The constraint still applies to that renamed column

ShmuelTreiger: Will the index/fk just track the column post-rename?

aapeliv: Yep. The names are just labels, the data/indexes/constraints are stored separately

aapeliv:
This seems fine, just needs frontend work

aapeliv (changes requested):
Actually, it seems to me that our convention has been to use `host_user_id` (not `host_id`), so we should stick to that. (See `Page` model, `Complaint`, etc. There's some cases where it's not followed though.)

aapeliv:
I can then do the frontend bit

ShmuelTreiger:
I'm not sure why gitlab isn't running? But I updated the names @aapeliv

aapeliv (commented):
Backend LGTM

PR #1639, opened 2021-07-14
Add user location to new line with better formatting

### LINKED ISSUE #1626: Friends list looks busy
Looking at #1568, the friends list looks very busy. Maybe best to just have "city, country"?

kerrylaval:
Put user location on new line with smaller/muted text 
closes #1626 

**Frontend checklist**
- [x] All tests pass

@Couchers-org/frontend

### DIFF
--- a/app/frontend/src/components/UserSummary.tsx
+++ b/app/frontend/src/components/UserSummary.tsx
@@ -81,10 +81,25 @@ export default function UserSummary({
       {!user ? (
         <Skeleton />
       ) : (
-        <>{nameOnly ? user.name : `${user.name}, ${user.age}, ${user.city}`}</>
+        <>{nameOnly ? user.name : `${user.name}, ${user.age}`}
+        </>
       )}
     </Typography>
   );
+  const location = (
+    <Typography
+      color="textSecondary"
+      variant="body1"
+      noWrap={nameOnly}
+    >
+      {!user ? (
+        <Skeleton />
+      ) : (
+        <>{user.city}
+        </>
+      )}
+    </Typography>
+  )
 
   return (
     <div className={classes.root}>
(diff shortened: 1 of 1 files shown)


kerrylaval:
image
image

app/frontend/src/components/UserSummary.tsx
```diff
@@ -81,10 +81,25 @@ export default function UserSummary({
...
+      noWrap={nameOnly}
+    >
+      {!user ? (
+        <Skeleton />
+      ) : (
+        <>{user.city}
+        </>
+      )}
+    </Typography>
+  )
```
darrenvong, line 102: It doesn't look like this is being used elsewhere, so we can inline the JSX instead so the code is closer to where it's being used

app/frontend/src/components/UserSummary.tsx
```diff
@@ -81,10 +81,25 @@ export default function UserSummary({
...
+      color="textSecondary"
+      variant="body1"
+      noWrap={nameOnly}
+    >
+      {!user ? (
+        <Skeleton />
+      ) : (
+        <>{user.city}
+        </>
+      )}
```
darrenvong, line 100:
```suggestion
      ) : (
        user.city
      )}
```
Don't think you need a fragment for this since it doesn't generate an extra DOM element.

app/frontend/src/components/UserSummary.tsx
```diff
@@ -81,10 +81,25 @@ export default function UserSummary({
       {!user ? (
         <Skeleton />
       ) : (
-        <>{nameOnly ? user.name : `${user.name}, ${user.age}, ${user.city}`}</>
+        <>{nameOnly ? user.name : `${user.name}, ${user.age}`}
+        </>
```
darrenvong, line 85: I think the same here - you can probably remove the fragment altogether and inline the ternary

darrenvong (changes requested):
Mostly minor code styling remarks, but looks good otherwise!

darrenvong:
Taken over by #1843

PR #1645, opened 2021-07-14
Use formState from useForm to detect user changes

### LINKED ISSUE #1635: Add warning when leaving a page with a dirty form
Is there a way to do this easily with our react-hook-form?

lucaslcode:
closes #1635

**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/features/profile/edit/EditProfile.tsx
+++ b/app/frontend/src/features/profile/edit/EditProfile.tsx
@@ -109,7 +109,13 @@ type FormValues = Omit<UpdateUserProfileData, "languageAbilities"> & {
   fluentLanguages: string[];
 };
 
-export default function EditProfileForm() {
+export interface EditProfileFormProps {
+  setFormIsDirty: (value: boolean) => void;
+}
+
+export default function EditProfileForm({
+  setFormIsDirty,
+}: EditProfileFormProps) {
   const classes = useStyles();
   const {
     updateUserProfile,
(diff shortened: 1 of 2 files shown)


app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -124,17 +130,35 @@ export default function EditProfileForm() {
...
+        regionsVisited: user?.regionsVisited,
+        regionsLived: user?.regionsLived,
+        additionalInformation: user?.additionalInformation,
+        avatarKey: user?.avatarKey,
       },
       shouldFocusError: true,
     });
 
+  const { isDirty } = formState;
+  setFormIsDirty(isDirty);
```
lucaslcode, line 160: Use `useEffect` so this isn't called every time (doesn't matter that much, but it's cleaner)

app/frontend/src/features/profile/edit/EditProfile.tsx
```diff
@@ -124,17 +130,35 @@ export default function EditProfileForm() {
...
+
   //Although the default value was set above, if the page is just loaded,
   //user will be undefined on first render, so the default values will be undefined.
   //So make sure to set values when user finshes loading
```
lucaslcode, line 164: This is the problem - this `useEffect` below always calls `setValue` and the form is then dirty.

The problem is these below useEffects are workarounds for the fact that `city`, `lat`, `lng`, and `radius` are not fields in the form, but are set by the `location` field.

There's a much better way to do it: firstly, delete the below `useEffects`, then change the location controller to something like this:
```
          <Controller
            defaultValue={{
              address: user.city,
              lat: user.lat,
              lng: user.lng,
              radius: user.radius,
            }}
            name="location"
            control={control}
            render={(onChange) => (
              <EditLocationMap
                showRadiusSlider
                initialLocation={{
                  address: user.city,
                  lat: user.lat,
                  lng: user.lng,
                  radius: user.radius,
                }}
                updateLocation={(location) => {
                  if (location) {
                    onChange({
                      city: location.address,
                      lat: location.lat,
                      lng: location.lng,
                      radius: location.radius,
                    });
                  }
                }}
              />
            )}
          />
```

Then change `handleSubmit` so all the fields are passed to `updateUserProfile`. Also get rid of all the `defaultValues` in `useForm` (remember, user might be `null` on first render anyway so it is currently setting all default values to null. When `defaultValue` is defined on the fields, user already exists since the fields aren't displayed until user is defined)

That's a bit complex, let me know on slack if you need further explanation.

PR #1646, opened 2021-07-14
Signup tweaks

### LINKED ISSUE #1625: Signup flow v2.1 (tweaks)
General:

* The "Couchers is not dating" bit somewhere
* "Let's get started" should be changed to something else on subsequent pages (maybe "Let's get started", "Your details", "Feedback and volunteering")
* Volunteer form seems to load scrolled halfway down

Feedback form:

* Order of "Would you like to contribute..." and "Briefly describe  your experience..." should be switched.
* "How you like to help..." and "What expertise..." should only show up if you select "yes" or "maybe" for "Would you like to contribute..."
* Add spacing between "How you like to help..." and first checkbox as it often goes to two  lines and becomes cramped
* "How you like to help..." edit to "How would you like to help..."

lucaslcode:
closes #1625 

- [x] Better titles
- [x] No pre-scrolled forms
- [x] Feedback form fixes
- [x] Community guidelines

**Backend checklist**
  - I don't have autoflake
- [x] All tests pass
- [x] Added migrations if there are any database changes, ~~rebased onto `develop` if necessary for linear migration history~~

**Frontend checklist**
- [x] All tests pass


lucaslcode:
This can be reviewed ~~except for Jail-related files~~

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -218,60 +199,94 @@ export default function ContributorForm({ processForm }: ContributorFormProps) {
...
+                </RadioGroup>
               </FormControl>
             )}
           />
-          <Typography
-            variant="body1"
-            htmlFor="expertise"
-            component="label"
-            className={classes.label}
+          <Collapse
```
lucaslcode, line 238: This makes the background image zoom in when the page body gets larger because it is set to `background-size: cover`. Any ideas?

darrenvong:
Build failed - there's some import error. Can this be reviewed through the preview site (assuming the new sign up backend changes has been merged)?

app/frontend/src/components/ContributorForm/ContributorForm.tsx
darrenvong: Hmm sound strange... is this CSS rule set by MUI somewhere?

darrenvong (commented):
Code looks fine to me so far, although can't see the app due to compile error so don't fully know what's up with the background zoom thing from just looking at the code :p

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: No its just for the background image to cover the whole area

ShmuelTreiger (commented):
Can't unaccept -->must accept? Can't reject? Can't refuse?

Out of curiosity, what's the difference between TOS and community guidelines?

app/backend/src/couchers/models.py
```diff
@@ -181,6 +181,7 @@ class User(Base):
     camping_ok = Column(Boolean, nullable=True)
 
     accepted_tos = Column(Integer, nullable=False, default=0)
+    accepted_community_guidelines = Column(Boolean, nullable=False, server_default="false")
```
aapeliv: Could we also make the CG an integer, for same reasons as TOS? Otherwise backend looks fine to me.

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: Fixed it by making background fixed

app/backend/src/couchers/models.py
lucaslcode: I thought it was pointless because unlike TOS, we aren't getting it from the backend

app/proto/jail.proto
```diff
@@ -26,20 +26,25 @@ service Jail {
...
-    // Accept to TOS
+    // Accept TOS
   }
 
   rpc SetLocation(SetLocationReq) returns (JailInfoRes) {
     // Set the user's location
     // If you're not jailed, do this with api.UpdateProfile!
   }
+
+  rpc AcceptCommunityGuidelines(AcceptCommunityGuidelinesReq) returns (JailInfoRes) {
```
kalvdans, line 37: Please document how I, as a third-party app developer, is going to use this method, for example by giving a link to the text to be accepted and how new versions are announced.

app/backend/src/couchers/models.py
kalvdans: We need to handle versions of guidelines so make it an int.

app/frontend/src/features/auth/signup/Signup.tsx
```diff
@@ -181,6 +181,16 @@ export default function Signup() {
...
+        <div className={authClasses.formCenter}>
+          <div className={authClasses.formWrapper}>
+            {error && (
+              <Alert className={authClasses.errorMessage} severity="error">
+                {error}
+              </Alert>
+            )}
+            {loading ? <CircularProgress /> : <CurrentForm />}
+          </div>
+        </div>
```
darrenvong: How come the markup for the form needs moving up here? Seems a bit hacky to need z-index to raise this up the page when it would have appeared at the right order where it was before based on the element order.

Also this kinda screws the natural left-right, top-down tabbing order when using a keyboard since the "Log in" link is only reachable after going through the whole form. Might be a minor accessibility concern for someone who made a mistake and got here but realise they just want to log in, but now still have to go through the whole form before they can reach the log in link.

app/frontend/src/features/auth/jail/constants.ts
```diff
@@ -1,6 +1,27 @@
-export const COMMUNITY_GUIDELINE_1 = "Be kind to one another";
-export const COMMUNITY_GUIDELINE_LABEL = "Be kind to one another";
-export const COMMUNITY_GUIDELINE_TITLE_1 = "Be kind to one another";
+import { ReactComponent as ghost } from "./resources/ghost.svg";
+import { ReactComponent as handshake } from "./resources/handshake.svg";
+import { ReactComponent as handWithHeart } from "./resources/handWithHeart.svg";
+import { ReactComponent as shield } from "./resources/shield.svg";
+
+export const COMMUNITY_GUIDELINES = [
+  "Couchers.org is an inclusive community where people of all cultures and backgrounds come together. Treat everyone with mutual respect and dignity",
```
darrenvong:
```suggestion
  "Couchers.org is an inclusive community where people of all cultures and backgrounds come together. Treat everyone with mutual respect and dignity.",
```
Missed full stop?

app/frontend/src/features/auth/jail/CommunityGuidelinesSection.tsx
```diff
@@ -0,0 +1,149 @@
...
+    <div className={classes.root}>
+      <Typography variant="h2" gutterBottom>
+        {COMMUNITY_GUIDELINES_SECTION_HEADING}
+      </Typography>
+      <form onSubmit={submit} className={className}>
+        {error && <Alert severity="error">{error}</Alert>}
+
+        <div className={classes.grid}>
+          {COMMUNITY_GUIDELINES.map((_, index) => (
+            <React.Fragment key={`guideline-${index}`}>
```
darrenvong: Minor nitpick: I don't see really how adding a string prefix before it makes a difference than just use `index` alone, since it's still going to be the value of `index` that makes the keys unique, so may as well use it directly?

app/frontend/src/features/auth/jail/CommunityGuidelinesSection.tsx
```diff
@@ -0,0 +1,149 @@
...
+            </React.Fragment>
+          ))}
+        </div>
+
+        <Button
+          onClick={submit}
+          disabled={completed || !formState.isValid}
+          className={classes.button}
+        >
+          {completed ? "Thanks!" : "Continue"}
```
darrenvong: Use existing values from constants file or pull this to constants if they don't already exist?

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -218,61 +209,100 @@ export default function ContributorForm({ processForm }: ContributorFormProps) {
...
+            in={
+              watchContribute !== CONTRIBUTE_OPTIONS[2] &&
```
darrenvong, line 240:
```suggestion
              watchContribute !== "No" &&
```
I feel like in this case, this is clearer as opposed to having to check out what the third option is and an ok compromise to have the value in the file

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -218,61 +209,100 @@ export default function ContributorForm({ processForm }: ContributorFormProps) {
...
-          <TextField
-            inputRef={register}
-            id="expertise"
-            margin="normal"
-            name="expertise"
-            helperText={errors?.expertise?.message ?? EXPERTISE_HELPER}
-            error={!!errors?.expertise?.message}
+            <Controller
+              id="contributeWays"
+              control={control}
```
darrenvong, line 246: I'm noticing some weird focus state styling, where the section heading changes colour when any of the checkboxes get focused:
Screenshot 2021-07-17 at 02 18 46

I don't think the title should change colour to be consistent with other input headings/labels. Instead, the focused checkbox option should show up/change etc depending on which checkbox is being focused on, which isn't the case at the moment. You have to blindly tab (not knowing where you are in the list of checkboxes) until the focus goes to the "what expertise do you have..." input field.

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -218,61 +209,100 @@ export default function ContributorForm({ processForm }: ContributorFormProps) {
             className={classes.textbox}
           />
           <Controller
-            id="contributeWays"
+            id="contribute"
             control={control}
```
darrenvong, line 211: Same thing as what I described for `contributeWays` seems to be happening here too

app/frontend/src/features/auth/jail/resources/ghost.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100%" height="100%" clipboard-id="document_1bs6v2b"><defs>
```
darrenvong: These SVG icons don't look very optimised and way more complicated than needed from a glance. I think it might be worth optimising it with a tool like [SVGOMG](https://jakearchibald.github.io/svgomg/) before using them. Only tried it on this file and it's reduced the file size by almost 1/3, so helps to speed up downloading them as well!

darrenvong (changes requested):

app/frontend/src/features/auth/signup/Signup.tsx
lucaslcode: I was experimenting, I'll put it back underneath

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: I don't think that will work once we have translations. I'd rather leave this. (I did `CONTRIBUTE_OPTIONS as const` so you can just hover to see the values)

lucaslcode:
Added `--runInBand` to prevent multiple tests running at once. This is bad but I don't even know which test is causing failures

aapeliv:
Still failing :/

app/frontend/src/components/CommunityGuidelines.tsx/CommunityGuidelines.tsx
```diff
@@ -0,0 +1,146 @@
+import {
```
darrenvong: The folder name for this component and the rest of the files shouldn't have `.tsx` in 🤔

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -139,7 +151,11 @@ export default function ContributorForm({ processForm }: ContributorFormProps) {
             {IDEAS_LABEL}
           </Typography>
           <TextField
-            inputRef={register}
+            inputRef={(el: HTMLInputElement) => {
+              if (!ideasInputRef.current && autofocus) el.focus();
+              if (el) ideasInputRef.current = el;
+              register(el);
+            }}
```
darrenvong: I think it might be safer to put an if check on `el` to see if it exists first before calling `el.focus`, as [callback refs could get called with `null` first before receiving the `HTMLInputElement` instance](https://reactjs.org/docs/refs-and-the-dom.html#caveats-with-callback-refs) when using an inline function, so `el` should be really be `HTMLInputElement | null` (the MUI type here is a bit crap as it marks `el` with `any`...)

app/frontend/src/components/ContributorForm/ContributorForm.tsx
darrenvong: We'll probably have to change this again anyway when we do translations, but if hovering gives the value in VS Code then let's leave it 👍

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -170,17 +168,21 @@ export default function AccountForm() {
...
-              value: usernameValidationPattern,
-            },
-            required: USERNAME_REQUIRED,
-            validate: async (username) => {
-              const valid = await service.auth.validateUsername(username);
-              return valid || USERNAME_TAKEN;
-            },
-          })}
+          inputRef={(el: HTMLInputElement) => {
+            if (!usernameInputRef.current) el.focus();
```
darrenvong: Same comment as in contributor form on `el` check

app/frontend/src/features/auth/signup/BasicForm.tsx
```diff
@@ -65,13 +73,17 @@ export default function BasicForm() {
...
           variant="standard"
-          inputRef={register({
-            pattern: {
-              message: NAME_EMPTY,
-              value: nameValidationPattern,
-            },
-            required: NAME_REQUIRED,
-          })}
+          inputRef={(el: HTMLInputElement) => {
+            if (!nameInputRef.current) el.focus();
```
darrenvong: Same comment as in contributor form

app/frontend/src/features/auth/signup/Signup.test.tsx
```diff
@@ -216,10 +250,51 @@ describe("Signup", () => {
...
+      await screen.findByText(COMMUNITY_GUIDELINE_TITLES[0])
+    ).toBeVisible();
+  });
+
+  it("displays the guidelines form when only feedback is pending", async () => {
+    const state: SignupFlowRes.AsObject = {
+      needBasic: false,
+      needAccount: false,
+      needAcceptCommunityGuidelines: true,
+      needFeedback: false,
```
darrenvong:
```suggestion
      needFeedback: true,
```
Otherwise doesn't match test description?

app/frontend/src/theme.ts
```diff
@@ -37,11 +37,9 @@ const themeOptions: ThemeOptions = {
         },
       },
     },
-    MuiListItem: { gutters: { paddingLeft: 0, paddingRight: 0 } },
-    MuiTabPanel: {
+    MuiFormLabel: {
       root: {
-        padding: 0,
-        paddingTop: spacing(1),
+        "&$focused": { color: "inherit" },
```
darrenvong, line 42: What does "$" do here?

darrenvong (changes requested):

app/frontend/src/theme.ts
lucaslcode: Some MUI magic. This was to fix the colour change on focus. I tried `MuiFormLabel: { focused: { color: "inherit" } }` and I got a helpful warning telling me to do what is here now

PR #1665, opened 2021-07-19
Fix frontend tests

### LINKED ISSUE #1656: Fix frontend tests

lucaslcode:
Will close #1656


aapeliv:
You can uncomment the blocks in app/.gitlab-ci.yaml to get front end builds and previews back.

lucaslcode:
See if this works...

app/frontend/src/components/ContributorForm/ContributorForm.tsx
```diff
@@ -103,7 +103,13 @@ export default function ContributorForm({
         .setExperience(data.experience)
         .setContribute(contribute)
         .setContributeWaysList(
-          data.contributeWays.split(",").filter((v) => !!v)
+          Object.entries(data.contributeWays).reduce<string[]>(
+            //contributeWays is an object of "ways" as keys, and "checked" booleans as values
+            //this reduces it to an array of the "ways" which were keys with "true" as a value
+            (previous, current) =>
+              current[1] ? [...previous, current[0]] : previous,
+            []
```
darrenvong: How is this that much faster than what we had before? It's still going through the list and building up a list each time a checkbox is marked/unmarked so the code still has a linear runtime?

I reckon things might run a bit quicker if we just let checkboxes be its own individual fields and collect the results at the end at `onSubmit`, at least then this linear time (apparently slow) operation only runs once. We are basically abusing this one field to do multiple things when web forms aren't really designed for that and we shouldn't really be coupling the UI against the backend data structure. 

Otherwise, if we are commenting out that UI interaction in the test, I think the old code is a bit easier to read so maybe best to keep that?

app/frontend/src/features/auth/signup/AccountForm.tsx
```diff
@@ -291,7 +291,7 @@ export default function AccountForm() {
                 row
                 aria-label="gender"
                 name="gender-radio"
-                onChange={onChange}
+                onChange={(e, value) => onChange(e, value)}
```
darrenvong: How is this different to what it was before? I think it'd call the function with those arguments anyway with `onChange={onChange}`?

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -255,7 +255,7 @@ export default function HostingPreferenceForm() {
             name="smokingAllowed"
             render={({ onChange, value }) => (
               <Select
-                onChange={onChange}
+                onChange={(event) => onChange(event.target.value)}
```
darrenvong, line 258: How was this working before if `onChange` needs the value explicitly here? 🤔

darrenvong:
Nice work figuring this and get it back to green again! How did you find it was those tests that were slow? Git bisect?

We might also want to remove `-w 1` in the `test-ci` command if this fix works to speed up the test runs again? Or is it still causing too much problems using the jest max workers default?

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: I don't think performance is the issue for this one, I don't see how this nor the previous version are particularly taxing. I thought it might be some obscure thing that would be eliminated by refactoring, and indeed it did seem to fix the test in the commit this first appeared, but didn't anymore when I applied it to develop. 

Anyway I prefer doing a map than explicitly listing every checkbox, but if you think the previous one is cleared we could put it back. (I personally prefer it this way, react-hook-form intends form fields to be used like this imo)

app/frontend/src/features/auth/signup/AccountForm.tsx
lucaslcode: Whoops, it's meant to be `onChange(value)`

app/frontend/src/features/profile/edit/EditHostingPreference.tsx
lucaslcode: No idea. Are we sure it's not broken? Seems like it should be broken.

lucaslcode:
I actually read somewhere that -w 1 or 2 can make it faster on CI. Maybe I'll remove and see.

lucaslcode:
I managed to reproduce reliably with -w 10 and then did a very long git bisect...

app/frontend/src/components/ContributorForm/ContributorForm.tsx
darrenvong: >  I don't see how this nor the previous version are particularly taxing.

I thought the same and so didn't mention it the first time I reviewed it as I didn't want to pre-optimise things too much... until the tests now somehow choke on this so it is too slow? 😅 

> react-hook-form intends form fields to be used like this imo

I'm not sure I agree with that, especially after looking at their more [complex example with double level nested fields](https://codesandbox.io/s/react-hook-form-usefieldarray-nested-arrays-m8w6j?file=/src/index.js), they still register values into individual fields in a nested data structure rather than collapsing everything into one.

The performance of react-hook-form comes from preventing unnecessary re-renders from state updates from React normally, but it still runs `onChange` on every change so ideally the operation is a simple one, which is usually the case in most use cases... until we broke it by wrangling multiple fields into one with "complex" logic.

darrenvong:
> I actually read somewhere that -w 1 or 2 can make it faster on CI. Maybe I'll remove and see.

I think you could be right on this... in fact https://github.com/facebook/jest/issues/1524#issuecomment-262366820 suggests `--runInBand` is actually better for CI environments 🤔

app/frontend/src/components/ContributorForm/ContributorForm.tsx
lucaslcode: > until we broke it by wrangling multiple fields into one

This version is multiple fields though - there's a controller for each checkbox so I suppose only one will re-render at a time. Before it was all in one controller.

I don't love this reduce but I don't think it's more opaque than the logic that was previously in `toggleCheckbox`. Some similar logic will still be necessary even if there was no `map`. (Converting true/false fields into a list of strings.)

Do you want me to explicitly list every checkbox? (And change `CONTRIBUTE_WAYS_OPTIONS` to individual strings?)

darrenvong: My bad, I misread it completely and thought it was still like the old code where everything was in one field 😅 

In that case, no, there won't be much point explicitly listing the checkboxes as the resultant code will be the same 👍

darrenvong:
```suggestion
          Object.entries(data.contributeWays).reduce<string[]>(
            //contributeWays is an object of "ways" as keys, and "checked" booleans as values
            //this reduces it to an array of the "ways" which were keys with "true" as a value
            (previous, [contributeWay, checked]) =>
              checked ? [...previous, contributeWay] : previous,
            []
```
Minor nitpick, but I like having explicit variable names to make things clearer over index access

lucaslcode:
`-w 1` had no significant difference - I imagine it correctly detected the environment and runs one process only anyway.

darrenvong (commented):
Should `app/proxy/protos.pb` be commited? I thought that's generated stuff right?

aapeliv:
> Should `app/proxy/protos.pb` be commited? I thought that's generated stuff right?

Yeah, and then please do a squash merge so it doesn't take space in history as well.

lucaslcode:
@aapeliv should that be in .gitignore or just deleted?
Do you want to do it? I wont have time until tonight.

aapeliv:
> @aapeliv should that be in .gitignore or just deleted?

Deleted. We moved the file and removed the entry from .gitignore, so people keep on accidentally re-adding it (would've probably been better to leave in the .gitignore).

>  Do you want to do it? I wont have time until tonight.

Yep, will do.

aapeliv:
Merged in 7e352576f767ef61b9ea4ad569687f8b849d60ee

PR #1677, opened 2021-07-21
Backend/refactoring/small changes inspired by delete account

ShmuelTreiger:
Various minor additions/changes. Biggest is a password_reset_token purge

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/jobs/definitions.py
+++ b/app/backend/src/couchers/jobs/definitions.py
@@ -7,6 +7,7 @@ from couchers.jobs.handlers import (
     process_enforce_community_membership,
     process_purge_account_deletion_tokens,
     process_purge_login_tokens,
+    process_purge_password_reset_tokens,
     process_send_email,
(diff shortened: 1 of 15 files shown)


kalvdans:
I'm unsure what the workflow is with PR into a feature branch. Please do preparation cleanups as a PR against `develop`.

ShmuelTreiger:
@kalvdans I did it against PR b/c I change `make_invisible`--which is used in PR--to `delete_user`. Plan would be to merge this once base PR has been merged.

Sorry, should have made a note.

kalvdans:
Please mark it as draft until it can be merged.

aapeliv:
We should just merge all the token deletion bg jobs into one. No need to run them separately.

ShmuelTreiger:
@aapeliv I was wondering that. I was also wondering about making a parent class and inheritance class for tokens. Thoughts on that? Might be worth doing both at once if we do.

PR #1686, opened 2021-07-21
Pull community guidelines from backend

### LINKED ISSUE #1682: Get Community Guidelines text + logos from backend

aapeliv:
Closes #1682

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [ ] All tests pass


kalvdans (dismissed):
Looks ok backendwise!

aapeliv:
I think the frontend tests require some tweaks, waiting for #1665 to merge before I can fix those.

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -83,22 +103,21 @@ export default function CommunityGuidelines({
...
             <React.Fragment key={index}>
               <Avatar className={classes.avatar}>
-                <SvgIcon
-                  component={COMMUNITY_GUIDELINE_ICONS[index]}
-                  fontSize="large"
-                  className={classes.icon}
-                />
+                <SvgIcon fontSize="large" className={classes.icon}>
+                  <svg
+                    dangerouslySetInnerHTML={{ __html: guideline.iconSvg }}
```
lucaslcode: Does this make  `<svg><svg>...`?

lucaslcode (changes requested):
Will check properly later

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
aapeliv: I'm not sure. I had some issues getting this to work, if you know what the right way to do this is, please let me know!

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -83,22 +103,21 @@ export default function CommunityGuidelines({
...
+                    xmlns="http://www.w3.org/2000/svg"
+                  />
+                </SvgIcon>
               </Avatar>
```
lucaslcode: This alongside the style changes works:
```

              <Avatar
                className={classes.avatar}
                src={`data:image/svg+xml,${encodeURIComponent(
                  guideline.iconSvg
                )}`}
              />
```

app/backend/resources/icons/ghost.svg
```diff
@@ -1,4 +1,4 @@
 <svg xmlns="http://www.w3.org/2000/svg">
```
lucaslcode:
```suggestion
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```

lucaslcode: Do this for them all

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -11,16 +12,18 @@ import {
...
 import makeStyles from "utils/makeStyles";
 
 import {
-  COMMUNITY_GUIDELINE_ICONS,
   COMMUNITY_GUIDELINE_LABEL,
-  COMMUNITY_GUIDELINE_TITLES,
-  COMMUNITY_GUIDELINES,
   COMMUNITY_GUIDELINES_REQUIRED,
   COMMUNITY_GUIDELINES_SECTION_HEADING,
 } from "./constants";
```
lucaslcode, line 28: I can't reach the styles but get rid of icon and use this for avatar: 
```

  avatar: {
    backgroundColor: theme.palette.grey[300],
    "& img": {
      fill: "none",
      width: "2rem",
      objectFit: "unset",
    },
  },
```

lucaslcode (changes requested):

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -58,6 +62,15 @@ export default function CommunityGuidelines({
   const [completed, setCompleted] = useSafeState(isMounted, false);
   const [error, setError] = useState("");
```
darrenvong, line 63:
```suggestion
```
This is probably redundant since you are now using React Query to fetch stuff, which has error state built in already.

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -83,22 +104,19 @@ export default function CommunityGuidelines({
         {error && <Alert severity="error">{error}</Alert>}
```
darrenvong, line 104: This is also redundant since the error gets thrown early above, so this never gets displayed.
```suggestion
```

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
```diff
@@ -83,22 +104,19 @@ export default function CommunityGuidelines({
...
+              />
               <div>
                 <Typography variant="h3" color="primary">
-                  {COMMUNITY_GUIDELINE_TITLES[index]}
-                </Typography>
-                <Typography variant="body1">
-                  {COMMUNITY_GUIDELINES[index]}
+                  {guideline.title}
                 </Typography>
+                <Typography variant="body1">{guideline.guideline}</Typography>
```
darrenvong: Nitpick - seeing the same variable name repeated twice always feels a bit odd (reminds me of that whole `PagePage` thing that's still in our code 🙃 ). Maybe destructure the outer guideline object into `{ iconSvg, title, guideline }`, then you only need to `guideline` once here and likewise for other properties above without needing the `guideline.` prefix

darrenvong (changes requested):

app/frontend/src/components/CommunityGuidelines/CommunityGuidelines.tsx
aapeliv: It uses react query to fetch the guidelines; but not to submit the form. This component is used in both signup and in jail. In signup it doesn't do an API call, I don't think.

lucaslcode (approved):

PR #1698, opened 2021-07-23
Fix frontend tests on develop and various mui deprications

lucaslcode:
Sorry, I didn't pick up on the failing tests in #1686

### DIFF
--- a/app/frontend/src/features/ReportButton.tsx
+++ b/app/frontend/src/features/ReportButton.tsx
@@ -105,7 +105,11 @@ export default function ReportButton({
     }
   );
 
-  const handleClose = () => {
+  const handleClose = (
+    event: {},
+    reason: "backdropClick" | "escapeKeyDown" | "button"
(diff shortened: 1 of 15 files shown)


app/frontend/src/features/ReportButton.tsx
```diff
@@ -146,8 +150,6 @@ export default function ReportButton({
         aria-labelledby="bug-reporter"
         open={isOpen}
         onClose={handleClose}
-        // won't close when clicking outside
-        disableBackdropClick
```
darrenvong, line 150: Is this also a deprecated prop? Lol wonder how many things we have to re-learn with their billions of deprecations and changes coming in v5 😅

darrenvong (approved):

app/frontend/src/features/ReportButton.tsx
lucaslcode: yeah

Issue #1515, opened 2021-06-27
Updates to regions lists


CouchersBot:
We need to add some languages and countries to the lists.

Currently missing languages:

* Taiwanese?
* Sicilian
* Sign languages are completely missing
* ???

Missing regions:
* Scotland, and other bits of the UK

For languages, it might be easier to just look at the -2 list and prune out any obviously old/dead/way too small languages.

With regions, I do think we should include Scotland, but that requires creating a non-standard code for it.

---
Subject: Language: Taiwanese
Description:
Hello Couchers,

I am so happy to find this website! Building my profile now, I am wondering whether in the "language" option, there can be "Taiwanese" included? That is my mother tongue. 

Thank you so much!

Warm hugs,

Vivi

Results:

Backend version: develop-7b502b12
Frontend version: develop-7b502b12
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.106 Safari/537.36
Page: https://app.couchers.org/
User (spoofable): vnloya (1353)

lucaslcode:
ISO 693-3 only has [min nan Chinese](https://iso639-3.sil.org/code/nan) which is the ~~same as Taiwanese~~ covers a few mutually intelligible dialects, though it is called Taiwanese when spoken in Taiwan. So maybe difficult to label

ShmuelTreiger:
Seems UK/GB/scotland is a pretty common issue. Seems we either need to make an exception or add an explicit note.

Could we add scotland and map it to the same country code?

aapeliv:
> Seems UK/GB/scotland is a pretty common issue. Seems we either need to make an exception or add an explicit note.

We should add them with made up codes.

> Could we add scotland and map it to the same country code?

No, it's desigend to be an bidirectional map so this would break the contract.

aapeliv:
We should get this done. I'll chase up itsi

ShmuelTreiger:
@aapeliv @ItsiW The list itsi gave me removes some languages we currently support. Should I add in all current languages for backward compatibility, or should I create a script to delete all now-invalid entries?

https://docs.google.com/spreadsheets/d/1QtNlZoLa0KR6t75SUNBGpyl3TgP41osvy3xFmzIxIh4/edit?usp=sharing

kalvdans:
> The list itsi gave me removes some languages we currently support. 

[...]

PR #1699, opened 2021-07-23
Update language list

ShmuelTreiger:
First part of #1515

I'm not sure how to test my changes to copy_resources_to_database

**Backend checklist**
- [x] All tests pass


app/backend/resources/languages.json
```diff
@@ -1,7 +1,19 @@
 [
```
aapeliv, line 1: Sort this file by code, please.

app/backend/src/couchers/resources.py
```diff
@@ -93,10 +93,22 @@ def copy_resources_to_database(session):
     Truncating and recreating guarantees the data is fully in sync.
     """
```
aapeliv: Instead of deleting these here, do it in the migration. For two reasons:

1. Mainly this function is by nature a bit more involved with the deferred constraints, and so I'd like this to stay very clear and specialised in what it does (replacing the resources tables in the db)
2. It's good to be explicit in the migration as to what's going on.

aapeliv (changes requested):
Thanks for doing this!

Based on having looked at the database, let's make the following changes to the languages:

1. Let's not remove: Abkhazian (abk), Icelandic (isl), Latin (lat), Komi (kom)
2. migrate Macro Arabic (ara) to Standard Arabic (arb)
3. migrate Chinese (zho) to Mandarin (cmn)

Those are the only languages to be deleted that exist in the db.

You should be able to add this just before the commit in op.execute in the migration:

```sql
UPDATE language_abilities SET language_code = 'arb' WHERE language_code = 'ara';
UPDATE language_abilities SET language_code = 'cmn' WHERE language_code = 'zho';

DELETE FROM language_abilities WHERE language_code in ('ara', 'arg', 'ave', 'bis', 'bre', 'cha', 'cor', 'cos', 'cre', 'div', 'dzo', 'fao', 'fij', 'gla', 'gle', 'glv', 'her', 'hmo', 'ido', 'iku', 'ipk', 'kua', 'mah', 'mri', 'nau', 'nav', 'oci', 'oji', 'pli', 'roh', 'sag', 'san', 'sme', 'tah', 'ton', 'vol', 'zho');
```

app/backend/resources/languages.json
```diff
@@ -1,39 +1,103 @@
...
   {
-    "code": "abk",
-    "name": "Abkhazian"
+    "code": "_tw",
```
ShmuelTreiger, line 3: Not sure why taiwanese starts with "_"

app/backend/resources/languages.json
```diff
@@ -695,12 +1139,28 @@
...
   {
     "code": "zul",
     "name": "Zulu"
+  },
+  {
+    "code": "zza",
+    "name": "Zaza"
   }
-]
+]
```
ShmuelTreiger: Missing line, I'll add it before we commit

app/backend/resources/languages.json
lucaslcode: I think it's because it's an "internal" couchers code - Taiwanese technically comes under southern min, but there was a request to add Taiwanese specifically

aapeliv (dismissed):
Looks good to me!

aapeliv (approved):
LGTM

app/backend/resources/languages.json
kalvdans: Please don't invent our own languages that doesn't exist in ISO 639-3. Where's the request to add Taiwanese so I can complain at the source? :)

Also update the API docs if we are not ISO 639-3 compatible.

PR #1700, opened 2021-07-23
Image upload button in TUI Editor

lucaslcode:
**Frontend checklist**
- [x] All tests pass


app/frontend/src/components/MarkdownInput/UploadImage.tsx
```diff
@@ -0,0 +1,65 @@
...
+  emitter,
+  open,
+  onClose,
+}: UploadImageProps) {
+  const alt = useRef("");
+  const { control } = useForm();
+  const handleSuccess = async (data: ImageInputValues) => {
+    emitter?.emit("command", "addImage", {
+      imageUrl: data.full_url,
+      altText: alt.current,
```
darrenvong, line 38: How are these being used by toast UI (or our code?) to display the image after?

app/frontend/src/components/MarkdownInput/MarkdownInput.tsx
```diff
@@ -75,6 +80,31 @@ export default function MarkdownInput({
   const fieldOnChange = useRef<typeof field.onChange>(field.onChange);
 
   useEffect(() => {
+    const uploadButton = imageUpload ? document.createElement("button") : null;
```
darrenvong, line 90: Is adding a custom button to the editor a thing since their v3 editor? That's pretty neat!

app/frontend/src/components/MarkdownInput/MarkdownInput.tsx
```diff
@@ -75,6 +80,31 @@ export default function MarkdownInput({
...
+
+    if (imageUpload) {
+      uploadButton!.type = "button";
+      uploadButton!.innerHTML = "Pic";
+      uploadButton!.style.margin = "0";
+      uploadButton!.style.background = "transparent";
+      uploadButton!.addEventListener("click", () => {
+        setImageDialogOpen(true);
+      });
```
darrenvong: I think we may want a `removeListener` in the `useEffect` cleanup function to remove this event handler to avoid memory leaks

app/frontend/src/components/MarkdownInput/UploadImage.tsx
lucaslcode: Magic, that's how

lucaslcode: (It just inserts the correct markdown `alt text`)

app/frontend/src/components/Markdown.test.tsx
```diff
@@ -40,37 +47,68 @@ describe("Markdown widget", () => {
...
       </div>
     );
-    expect(screen.getByTestId("allowed")).toContainHTML(
-      `<img src="${mediaURL}/image.png" alt="image">`
+    expect(
+      getByAltText(screen.getByTestId("allowed"), "image")
+    ).toHaveAttribute("src", "https://mymedia.com/image.png");
+    expect(
+      queryByRole(screen.getByTestId("allowed"), "link", { name: "image" })
+    ).toBeNull();
```
darrenvong:
```suggestion
    ).not.toBeInTheDocument();
```
Slightly more explicit with the `jest-dom` matchers. Same comment for all places here using `toBeNull`

app/frontend/src/components/MarkdownInput/MarkdownInput.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {
+  IMAGE_DESCRIPTION,
+  INSERT_IMAGE,
+} from "components/MarkdownInput/constants";
+import { useForm } from "react-hook-form";
+import { service } from "service";
+import wrapper from "test/hookWrapper";
+import { MockedService } from "test/utils";
+
+import MockedMarkdownInput from "./MarkdownInput";
```
darrenvong: Does this import alone not work? I thought the mocked input only gets pulled in if you explicitly do `jest.mock("./MarkdownInput")` since this is a relative import so [jest doesn't auto mock them](https://jestjs.io/docs/26.x/manual-mocks#mocking-user-modules)?

lucaslcode:
@darrenvong I have somehow broken all the MarkdownInput mocks and can't figure out why :(

app/frontend/src/components/MarkdownInput/UploadImage.tsx
darrenvong: Is said black magic even documented anywhere? Or is it a bit of digging into their source code?

darrenvong (changes requested):
Only a few minor things in test, generally looks good despite all the Toast UI black magic 🧙 🤯

darrenvong:
> I have somehow broken all the MarkdownInput mocks

It's probably because you moved the `MarkdownInput` into its own folder, so I think you'll have to move the `__mocks__` folder that was under `components` folder to be under the `components/MarkdownInput` folder instead

lucaslcode:
@darrenvong it hates me, want to give it a try?

app/frontend/src/components/MarkdownInput/UploadImage.tsx
lucaslcode: Indeed it is source code digging, the docs are pretty poor and worse since they are mostly not updated since v3 was released

darrenvong:
~~Hmm this is so weird, the MarkdownInput test only seems to fail when run together with everything else but not on its own... (after I fixed the other broken tests in `CommunityInfoPage`)~~

~~Doing `--runInBand` seems to have fixed it though~~

Never mind, somehow re-jigging where I do the await (on the dialog) and searching for inputs within the dialog seems to have fixed it for me when tried locally. Let's see if CI agrees...

lucaslcode:
Wait, so I had correctly moved the mocks folder? So what was making everything fail?

darrenvong:
> Wait, so I had correctly moved the mocks folder? So what was making everything fail?

I'm not 100% sure tbh... but my guess is re-rendering to show the dialog for real is slow, and somehow the `findBy*` queries when done on one of the fields in the dialog like the "image description" (as opposed to await for the dialog to appear first) wasn't waiting long enough or something.

darrenvong (approved):
Anyway tested on the preview after Aapeli granted me community mod on my other account and it works, so looks good to me :D

PR #1703, opened 2021-07-24
Events part 3: events listing page/tab

darrenvong:
Covers a bit more of #541. Swaps out the events section to a new events list under the "events" tab that will actually let user see all events under a community (with a "see more events" button), as opposed to limited to 3 like before, since that was only meant to be a preview of the latest events, temporarily used under the events tab as a placeholder so it doesn't say "events coming soon" when we said we've released it.

**Desktop**
Screenshot 2021-07-26 at 00 16 04

**Mobile**
Screenshot 2021-07-26 at 00 16 55

**Frontend checklist**
- [x] All tests pass


app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+    alignItems: "center",
+  },
+  attendeesAndInfo: {
+    gridArea: "attendees",
+    display: "flex",
+    flexDirection: "column",
+    alignItems: "center",
+  },
+  image: {
+    objectFit: "contain",
```
lucaslcode: I am partial to "cover" myself

app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+import dayjs from "utils/dayjs";
+import makeStyles from "utils/makeStyles";
+
+import { getAttendeesCount, ONLINE } from "../constants";
+import getContentSummary from "../getContentSummary";
+import { getExtraAvatarCountText } from "./constants";
+import eventImagePlaceholder from "./eventImagePlaceholder.svg";
+import { useEventAttendees } from "./hooks";
+
+const useStyles = makeStyles((theme) => ({
```
lucaslcode: I think this isn't the latest iteration of the design - the one  to the right of the Events title in figma (labelled "events iteration") looks much more refined to me.

app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+  overviewRoot: {
+    "&:hover": {
+      backgroundColor: theme.palette.grey[50],
+    },
```
lucaslcode, line 25: Can we not use `<CardActionArea>`?

app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  const { attendees, attendeesIds, isAttendeesLoading } = useEventAttendees({
+    eventId: event.eventId,
+    type: "summary",
+  });
+  const extraAvatarCount = event.goingCount - Math.min(3, attendeesIds.length);
+
+  return (
+    <Card className={classes.overviewRoot}>
+      <Link
```
lucaslcode, line 103: Better if we can have the link outside `<Card>` or the padding won't be clickable

app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+          >
+            {event.offlineInformation
+              ? event.offlineInformation.address
+              : ONLINE}
+          </Typography>
+          <Typography variant="body1">
+            {getAttendeesCount(event.goingCount)}
+          </Typography>
+          {isAttendeesLoading ? (
+            <CircularProgress />
```
lucaslcode: I prefer a skeleton for this kind of element. Also, there seems to be a slight pause before the loading indicator appears and there is nothing there. Is there a possible state where `isAttendeesLoading` is false or undefined but `attendees` is still undefined?

Anyway the new design only has number of attendees.

app/frontend/src/features/communities/events/LongEventCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+            attendees &&
+            attendeesIds.length > 0 && (
+              <div
+                className={classes.avatarGroup}
+                data-testid={AVATAR_GROUP_TEST_ID}
+              >
+                {attendeesIds.slice(0, 3).map((attendeeUserId) => {
+                  const attendee = attendees.get(attendeeUserId);
+                  return (
+                    <Avatar
```
lucaslcode: Next time, there is an `AvatarGroup` mui component

app/frontend/src/features/communities/events/EventsList.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { data, error, hasNextPage, fetchNextPage, isLoading } =
+    useListCommunityEvents({
+      communityId: community.communityId,
+      pageSize: 5,
+      type: "all",
+    });
+
+  return (
+    <>
+      <SectionTitle icon={<CalendarIcon />}>{EVENTS_TITLE}</SectionTitle>
```
lucaslcode, line 52: I'm confused: There is `SectionTitle` in this component, but it is not a section, it is the events tab, right? Yet it does appear to correctly be h1.

lucaslcode (changes requested):

app/frontend/src/features/communities/events/LongEventCard.tsx
darrenvong: I used that to start with, but they hardcode the circle so that for events with a lot of attendees (like +100), the text gets clipped off. Also had to return "dummy avatars" JSX just so the components get the count correctly so I thought it was easier to write my own here in the end 😅

app/frontend/src/features/communities/events/LongEventCard.tsx
darrenvong: Oh crap, that would have been much easier to implement as well 🙃  I'll change it since I don't really like the avatar thing at the preview level tbh, it's fetching a lot of data for very little gain imo

app/frontend/src/features/communities/events/LongEventCard.tsx
darrenvong: I should probably update this to use a different placeholder... since "cover" doesn't work well with the current SVG placeholder atm

app/frontend/src/features/communities/events/EventsList.tsx
darrenvong: It defaults to h1 unless it's changed by the `component` prop I think from memory. Maybe we should rename this component? Otherwise I'll be doing the same thing with the icon and `Typography` as `h1` anyway

app/frontend/src/features/communities/events/LongEventCard.tsx
darrenvong: It sets the area as a `<button>` and I'm not sure there's any way to override it with that component from memory, and all it seems to be doing is some padding and this background colour change on hover, so seems easier to just implement it in CSS instead

app/frontend/src/features/communities/events/EventsList.tsx
darrenvong: Renamed to `TitleWithIcon`

lucaslcode (approved):

PR #1704, opened 2021-07-24
Fix text not wrapping if text had no spaces, such as link with url, CSS

CamArturo:
**Frontend checklist**
- [ x] Formatted my code with `yarn format && yarn lint --fix`
- [ x] There are no warnings from `yarn lint`
- [ ] All tests pass
- [ x] Checked Desktop, Mobile and Tablet screen sizes

The is a warning from another component. 
Warning: Failed prop type: The prop `onBackdropClick` of `ForwardRef(Modal)` is deprecated. Use the onClose prop with the `reason` argument to handle the `backdropClick` events.
I think it's on the Drawer component.


lucaslcode (approved):
Thanks! There are some new deprication warnings which we'll handle soon.

PR #1707, opened 2021-07-25
Fix: wrapping of discussion comment + clearing input when user submits comment

### LINKED ISSUE #1708: text wrap breaks words
Subject: text wrap breaks words
Description:
look at my profile

Results:
text is wrapped breaking words at any character. 

text should be wrapped without splitting words?

Backend version: develop-f191b66f
Frontend version: develop-f191b66f
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:90.0) Gecko/20100101 Firefox/90.0
Page: https://app.couchers.org/user/nolo
User: nolo (14)

darrenvong:
Various small improvements/bug fix on discussion, most notably:
- Words don't break on weird spots for the sake of wrapping (closes #1708)

**Before fix (see highlighted text)**
Screenshot 2021-07-25 at 15 02 08

**After fix**
Screenshot 2021-07-25 at 15 04 02

- The comment form is now cleared when user submits a comment/reply comment to the discussion!
- The discussion comment cards are now edge to edge (takes up full screen width) on mobile screens
- The comment also takes up more of the right hand side space on small screens
Screenshot 2021-07-25 at 15 09 35

(The above screenshot also shows the new fix hasn't broken the wrapping of super long link)

**Frontend checklist**
- [x] All tests pass


app/frontend/src/components/MarkdownInput.tsx
```diff
@@ -100,6 +102,11 @@ export default function MarkdownInput({
         "link",
       ],
     });
+
+    if (resetInputRef) {
+      resetInputRef.current = fieldRef.current.reset.bind(fieldRef.current);
+    }
```
darrenvong, line 108: This is a bit of an annoying workaround, since react-hook-form's `fieldRef` through `controller` at the parent level doesn't seem to expose the same `fieldRef` as we set here - it only exposes an object with a `focus` method :/

lucaslcode: Horrific, but yeah doesn't seem like a way around it

lucaslcode (approved):

PR #1709, opened 2021-07-25
fix: lowercases usernames/emails before sending to backend

### LINKED ISSUE #1657: Signup form lowercasing
Right now it doesn't accept mixed-case when it requires lowercase. E.g. typing in `Aapeli28` (as one does automatically on iPhone), it's not accepted.

This should lowercase as one is typing.

### LINKED ISSUE #1728: Signup page considers emails with periods in the username as invalid.
The signup page will not let you create accounts using emails with periods in the name.
Here is an example of me trying to enter in an email (granted, it's not my real email address) with a period in it. If you remove the period, the registration process can successfully go to the next stage.
invalid email
This is a problem, because some folks have emails with periods in them, and they wouldn't be able to sign up with them in this current state.

darrenvong:
Closes #1657. Closes #1728. Lowercasing as the user type requires more code changes (making the input specifically into a controlled component) and not sure it's worth it if lowercasing at submission also works.

Updated error message and validation regexp so it doesn't say the username can only have lowercase letter anymore.
Screenshot 2021-07-25 at 17 59 34

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/auth/constants.tsx
+++ b/app/frontend/src/features/auth/constants.tsx
@@ -87,7 +87,7 @@ export const SIGN_UP_REDIRECT =
 export const SIGN_UP_TOS_ACCEPT = "I Accept the Terms of Service.";
 export const SIGN_UP_TOS_TEXT = "To continue, please read and accept the ";
 export const SIGN_UP_USERNAME_ERROR =
-  "Username can only have lowercase letters, numbers or _, starting with a letter.";
+  "Username can only have numbers or _, starting with a letter.";
 export const CONTINUE = "Continue";
 export const SUBMIT = "Submit";
 export const TERMS_OF_SERVICE = "Terms of Service";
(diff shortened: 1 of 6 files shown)


aapeliv:
Emails should also be lowercased!

darrenvong:
Does the backend currently error when user types in email with capitals? I'll fix it here too 👍

aapeliv:
> Does the backend currently error when user types in email with capitals? I'll fix it here too 👍

Yep

lucaslcode (changes requested):
Looks good, how about signup email?

darrenvong:
> Looks good, how about signup email?

It's already done here. Unless there's another spot I overlooked?
https://github.com/Couchers-org/couchers/blob/7f0db6b1583a0a7db2d081f2e2af8297fa036c61/app/frontend/src/features/auth/signup/BasicForm.tsx#L44

lucaslcode:
Did you see #1728 ? I tried using upper case on preview and it didn't work.

lucaslcode (dismissed):
Ah I see you added the i flag

lucaslcode (changes requested):
The email regex validation needs the i flag too

lucaslcode (approved):

PR #1724, opened 2021-07-26
added link to write refence in emails

ghost:
**Backend checklist**
- [ ] All tests pass

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/references.py
+++ b/app/backend/src/couchers/servicers/references.py
@@ -221,7 +221,7 @@ class References(references_pb2_grpc.ReferencesServicer):
             session.commit()
 
             # send the recipient of the reference an email
-            send_host_reference_email(reference, context, host_request, reference.reference_type, both_written=other_reference is not None)
+            send_host_reference_email(reference, reference.from_user_id, host_request, reference.reference_type, both_written=other_reference is not None)
 
             return reference_to_pb(reference, context)
 
--- a/app/backend/src/couchers/servicers/references.py
+++ b/app/backend/src/couchers/servicers/references.py
@@ -6,6 +6,7 @@
 
 * TODO: Get bugged about writing reference 1 day after, 1 week after, 2weeks-2days
 """
+from sqlalchemy.sql.expression import false
 import grpc
 from sqlalchemy.orm import aliased
 from sqlalchemy.sql import and_, func, literal, or_, union_all
@@ -221,7 +222,7 @@ class References(references_pb2_grpc.ReferencesServicer):
             session.commit()
 
             # send the recipient of the reference an email
-            send_host_reference_email(reference, reference.from_user_id, host_request, reference.reference_type, both_written=other_reference is not None)
+            send_host_reference_email(reference, reference.from_user, host_request, false, both_written=other_reference is not None)
 
             return reference_to_pb(reference, context)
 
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/references.py
```diff
@@ -221,7 +221,7 @@ def WriteHostRequestReference(self, request, context):
             session.commit()
 
             # send the recipient of the reference an email
-            send_host_reference_email(reference, both_written=other_reference is not None)
+            send_host_reference_email(reference, context, host_request, reference.reference_type, both_written=other_reference is not None)
```
aapeliv: The arguments are wrong given the method signature

```py
 def send_host_reference_email(reference, other_user, host_request, surfed, both_written):
```

app/backend/src/couchers/tasks.py
```diff
@@ -182,8 +182,8 @@ def send_friend_request_email(friend_relationship):
         },
     )
 
-
-def send_host_reference_email(reference, both_written):
+#called in references.py
```
aapeliv: No need for this comment, one can easily find this out by searching through the codebase

app/backend/templates/host_reference.md
```diff
@@ -15,7 +15,9 @@ This is what they wrote:
 
 Thanks for using Couchers to organize this interaction! We hope you had an enjoyable, fulfilling time.
 {% else %}
-Please go and write a reference for them too. It's a nice gesture and helps us build a community together! When you've both written a reference, both references will become visible. Otherwise {{ reference.from_user.name|couchers_escape }}'s reference will become visible 2 weeks after the end of your interaction, after which you cannot write a reference back.
+Please go and write a reference for them too.
+{{ urls.leave_reference_link | couchers_escape}}
+ It's a nice gesture and helps us build a community together! When you've both written a reference, both references will become visible. Otherwise {{ reference.from_user.name|couchers_escape }}'s reference will become visible 2 weeks after the end of your interaction, after which you cannot write a reference back.
```
aapeliv, line 20: Please format this link similar to other emails with a button on HTML, and link otherwise

aapeliv (changes requested):
Hi @zachary-thomas-115, thanks for taking this on! It's a good start.

Have a look at the comments, and please write a test to check the full expected URL is in the message and make sure that test passes locally.

app/backend/src/couchers/tasks.py
ghost: Removed

app/backend/src/couchers/servicers/references.py
ghost: This is fixed I think:
other_user is being passed in as reference.from_user.
surfed is changed to false, because this is just for the host I believe.

app/backend/src/couchers/tasks.py
```diff
@@ -182,8 +182,7 @@ def send_friend_request_email(friend_relationship):
         },
     )
 
-
-def send_host_reference_email(reference, both_written):
+def send_host_reference_email(reference, other_user, host_request, surfed, both_written):
```
kalvdans, line 185: Please document the types of the new arguments. I'm unsure about the fourth argument, an SQL expression is passed in, but it is used in the function body as a Python bool.

kalvdans (changes requested):

PR #1740, opened 2021-07-30
fix: #1713 - jail page keeps reloading

### LINKED ISSUE #1713: Jail page keeps reloading
Reported and recorded by Lukas: https://1drv.ms/v/s!Ajo4-4qFAquUgagRP3RS4uwakiBh5Q?e=BidB2W

niecnasowa:
closes #1713

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/AppRoutes.tsx
+++ b/app/frontend/src/AppRoutes.tsx
@@ -123,7 +123,7 @@ export default function AppRoutes() {
       <AppRoute isPrivate path={settingsRoute}>
         <Settings />
       </AppRoute>
-      <AppRoute variant="full-screen" isPrivate={false} path={jailRoute}>
+      <AppRoute isPrivate={false} path={jailRoute}>
         <Jail />
       </AppRoute>
       <AppRoute isPrivate={false} exact path={logoutRoute}>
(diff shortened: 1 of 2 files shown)


lucaslcode:
@aapeliv can you test this? Need a user to be put in jail

aapeliv:
> Need a user to be put in jail

I put everyone on next in jail.

aapeliv:
This does seem to fix the problem, but I think we should fix it in some other way than to just hide the whole rest of UI? It just looks very bare?

lucaslcode:
Yeah I guess you're right. 
@niecnasowa I thought if a better way - do you want to undo thus change, and instead, to the query in `features/useNotifications.ts`, you can disable it using the ["enabled" option](https://react-query.tanstack.com/guides/disabling-queries) if the user is jailed ~~(`useAuthStore().state.jailed`)~~ `useAuthContext().state.jailed`

darrenvong:
> (`useAuthStore().state.jailed`)

I think it's better to do it via `useAuthContext` instead, otherwise we'll end up with two auth stores

lucaslcode:
Sorry yes that's what I meant!

niecnasowa:
@lucaslcode Ok, I will change it based on your comments.

niecnasowa:
@aapeliv @darrenvong @lucaslcode Can you recheck it now?

lucaslcode (approved):
Great, thanks for your contribution!

lucaslcode:
You can merge this

PR #1741, opened 2021-07-31
Fix event threads

### LINKED ISSUE #1710: ListEvents should return events within the community
Not only those owned by the community

### LINKED ISSUE #1736: CreateEvent creates "invalid" discussion IDs
Reproduction steps:
1. Create a new event with `CreateEvent` from the `Events` service
2. Call `GetThread` from the `Threads` service, using the `thread_id` in the resulting `Event` object

**Expected result**
A thread (for a discussion) is returned

**Actual result**
A "discussion thread not found" is thrown from the backend

**Other notes**
My suspicion is this is using the thread's auto increment ID in the database rather than packing it with the correct format expected in `GetThread`.

aapeliv:
Closes #1736. Closes #1710.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/tests/test_fixtures.py
+++ b/app/backend/src/tests/test_fixtures.py
@@ -512,9 +512,6 @@ def requests_session(token):
 
 @contextmanager
 def threads_session(token):
-    """
-    Create a communities API for testing, uses the token for auth
-    """
     channel = fake_channel(token)
     threads_pb2_grpc.add_ThreadsServicer_to_server(Threads(), channel)
     yield threads_pb2_grpc.ThreadsStub(channel)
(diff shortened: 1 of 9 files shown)


app/backend/src/tests/test_fixtures.py
```diff
@@ -508,6 +510,16 @@ def requests_session(token):
     yield requests_pb2_grpc.RequestsStub(channel)
 
 
+@contextmanager
+def threads_session(token):
+    """
+    Create a communities API for testing, uses the token for auth
+    """
+    channel = fake_channel(token)
```
kalvdans, line 515: Change or remove copy-pasted docstring

kalvdans (dismissed):
Lgtm with or without my comment!

Issue #1737, opened 2021-07-30
Start signup on landing site


aapeliv:
As discussed in frontend meeting.

PR #1742, opened 2021-07-31
Add JSON transcoding to signup RPC

aapeliv:
This allows calling the signup rpc with something like

```
curl --data '{"basic":{"name":"tester","email":"testing@example.com"}}' http://localhost:8888/auth/signup
{
 "flowToken": "b04611f682299dc1222e5322a916944cf8ea9e9ad5471819c53db4684dba4fc3",
 "needBasic": false,
 "needAccount": true,
 "needFeedback": true,
 "needVerifyEmail": true,
 "needAcceptCommunityGuidelines": true
}
```

I'll use this to implement the first page of signup on the landing page.

Part of #1737


kalvdans:
Why not grpc?

app/backend/src/couchers/errors.py
```diff
@@ -118,7 +118,7 @@
 SIGNUP_FLOW_BASIC_FILLED = "Basic part already filled in."
 SIGNUP_FLOW_BASIC_NEEDED = "Basic part needed to sign up."
 SIGNUP_FLOW_EMAIL_STARTED_SIGNUP = "Please check your email for a link to continue signing up."
-SIGNUP_FLOW_EMAIL_TAKEN = "Email taken."
+SIGNUP_FLOW_EMAIL_TAKEN = "That email address is already associated with an account. Please log in instead!"
```
kalvdans, line 121: Move elaborate error messages to frontend.

kalvdans (approved):
Looks ok I guess, maybe someone more web-savvy can have a look. Let's switch completely from grpc to json someday.

aapeliv:
> Why not grpc?

The simple landing site doesn't speak gRPC, and it's just a one-off call; so easy to transcode to JSON for that.

> Let's switch completely from grpc to json someday.

I think the gRPC stuff is amazing from the backend side: having strict machine-readable definitions of the APIs, and having protobuf do all the serialization/deserialization and making sure everything is fine. I know it's a bit cumbersome at times and a slightly different style, but I think for that it's worth it.

For the users of the API: I do agree, it might be good in the future to transcode the whole API to JSON, and provide OpenAPI schemas. It's the way of the web and much nicer for consumers. For now though, our frontend team has said it's not much of a hassle and hasn't shown too much interest (I've brought it up several times!).

PR #1743, opened 2021-07-31
Add sentry to frontend

### LINKED ISSUE #1035: Investigate/set up Sentry on frontend

aapeliv:
I tried. I failed. Anyone want to take this up. Having issues mostly in the error boundary but I also think maybe I'm making stylistic mistakes elsewhere too.

Closes #1035

**Frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/frontend/src/components/Button/Button.tsx
+++ b/app/frontend/src/components/Button/Button.tsx
@@ -1,4 +1,5 @@
 import { Button as MuiButton, ButtonProps, useTheme } from "@material-ui/core";
+import * as Sentry from "@sentry/react";
 import classNames from "classnames";
 import React, { ElementType } from "react";
 import { useIsMounted, useSafeState } from "utils/hooks";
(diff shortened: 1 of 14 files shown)


darrenvong:
Ok, think I tidied up everything and added other recommended packages according to their docs so tests won't end up sending dummy errors up! Hopefully the app will build again. Not sure why errors aren't showing up in the Sentry dashboard still :/

app/frontend/src/components/Button/Button.tsx
```diff
@@ -55,6 +56,8 @@ export default function Button<D extends ElementType = "button", P = {}>({
     try {
       setWaiting(true);
       await onClick(event);
+    } catch (e) {
+      Sentry.captureException(e);
```
aapeliv, line 60: Is this supposed to be here? I thought it should capture exceptions automatically in the error boundary

darrenvong: This is needed here since even without the catch, I believe JS catches the error (equivalent of writing an empty catch block) and so it won't go to the error boundary. The error boundary only gets uncaught errors

darrenvong (approved):

PR #1744, opened 2021-07-31
Disable sentry excepthook integration

aapeliv:
It keeps on sending out handled exceptions otherwise

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/app.py
+++ b/app/backend/src/app.py
@@ -27,7 +27,6 @@ logging.getLogger("couchers.jobs.worker").setLevel(logging.INFO)
 
 if config.config["SENTRY_ENABLED"]:
     # Sends exception tracebacks to Sentry, a cloud service for collecting exceptions
-    integrations
     sentry_sdk.init(
         config.config["SENTRY_URL"],
         traces_sample_rate=0.0,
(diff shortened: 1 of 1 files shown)


app/backend/src/app.py
```diff
@@ -21,7 +27,23 @@
 
 if config.config["SENTRY_ENABLED"]:
     # Sends exception tracebacks to Sentry, a cloud service for collecting exceptions
-    sentry_sdk.init(config.config["SENTRY_URL"], traces_sample_rate=0.0, environment=config.config["COOKIE_DOMAIN"])
+    integrations
```
kalvdans: syntax error

kalvdans (commented):
What is the motivation for not getting the full sentry package?

aapeliv:
> What is the motivation for not getting the full sentry package?

By default it adds an excepthook that will vacuum up unhandled exceptions. I believe it's interfering with gRPC though, and is sending through exceptions that are handled by gRPC. This is polluting the sentry dashboard (and database) with a whole lot of user-errors, e.g. `UNAUTHENTICATED`, `NOT_FOUND`, etc.

app/backend/src/app.py
aapeliv: Thanks for picking that up!!

kalvdans (approved):
LGTM! Thanks for using trailing commas, it keeps the diff down when adding more parameters.

aapeliv:
> Thanks for using trailing commas, it keeps the diff down when adding more parameters.

It's done by black automatically!

PR #1749, opened 2021-08-01
Gitpodify Couchers

pawlean:
Love what you're building with Couchers! Thanks @darrenvong for telling me about this. I thought that I would raise this PR to help anyone wanting to contribute with Gitpod. :)

This PR allows you to run Couchers on [Gitpod](https://gitpod.io), allowing anyone who wants to start contributing to your project immediately by speeding up the onboarding process. 

All users have to do is either prefix the repo URL with _gitpod.io/#_ or [install the Gitpod extension](https://www.gitpod.io/docs/browser-extension) This will open Gitpod directly on their browser, install all the dependencies and start-up tasks and will be ready to start coding in a development environment immediately. 

**Backend checklist**
N/A :)

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,15 +1,18 @@
 # List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
 tasks:
-  - name: Compile the protocol buffers & launch the backend with docker-compose
+  - name: Compile the protocol buffers, launch the backend with docker-compose
+    before: cd app
     init: |
-      cd app
       docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
       docker-compose up --build
-  - name: Install and launch the frontend with yarn
-    init: |
-      cd app/frontend 
-      yarn install
-    command: yarn start
+
+    env:
+      REACT_APP_API_BASE_URL: $(gp url 8888)
+
+  - name: install and launch the frontend with yarn
+    before: cd app/frontend
+    init: yarn install
+    command: echo "wait for docker-compose to finish, then run yarn start"
 
 # List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
 ports:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,17 @@
+# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
+tasks:
+  - name: Compile the protocol buffers & launch the backend with docker-compose
+    init: |
+      cd app
+      docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+      docker-compose up --build
+  - name: Install and launch the frontend with yarn
+    init: |
+      cd app/frontend 
+      yarn install
+    command: yarn start
+
+# List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
+ports:
+  - port: 3000
+    onOpen: open-preview


aapeliv:
This looks interesting! How do I test this PR out?

pawlean:
> This looks interesting! How do I test this PR out?

Head over here: https://gitpod.io/#https://github.com/Couchers-org/couchers/pull/1749

aapeliv:
Ah, awesome. I knew there’d be some trick I could do to the URL to get to a pr/branch! Just couldn’t find it in gitpod docs

darrenvong:
Nice to see you here @pawlean! Thanks for sending in a PR, this does look like a cool way to lower the barrier for contributions 😄 intrigued to give it a spin later when I'm back on my laptop! 👀

aapeliv:
I tried this; two comments:

1. The frontend also requires the built protos to work: is that all fine the way it's done now? Is the filesystem fully shared (e.g. no dockerizing of each task, etc). Is there a way to do a "pre-task" step instead?
2. When I ran it, it didn't open the :3000 port, and I couldn't figure out how to access the frontend?

Looks pretty spiffy in general!

darrenvong:
> 2\. When I ran it, it didn't open the :3000 port, and I couldn't figure out how to access the frontend?

It spun up a browser for me by default and ~~it seems to work ok~~ - I think since the `generate_protos` script builds all the required protos for the project?

Edit: never mind, I think it's not quite fully working since the frontend is still calling the API with `localhost`, which I suspect doesn't exist in the Gitpod environment. Is there an easy way to work around that? 🤔 

~~There's some funky CORS issues based on how Gitpod maps the exposed port onto the URL parts instead 😅  not sure if that means we'd need to tweak the Envoy config as well to get this working~~ maybe not, I think this seems to be related to the VS Code environment in the browser

One other thing I haven't figured out immediately though - is there a way to open the browser again if you closed it by accident?

But once the minor teething issue is fixed, I think this is pretty cool and worthwhile to add some notes on the getting started docs for those who don't fancy downloading things onto their machine!

.gitpod.yml
```diff
@@ -0,0 +1,17 @@
+# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
+tasks:
+  - name: Compile the protocol buffers & launch the backend with docker-compose
+    init: |
+      cd app
+      docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+      docker-compose up --build
```
lucaslcode, line 7: Does this command already include a frontend docker container? Maybe it should specify `backend proxy postgres`

aapeliv:
The frontend is failing to connect to the backend, as it's trying to reach the backend at http://localhost:8888. I'm not sure how to fix this as the pod name will change dynamically? Maybe there's an env var with the pod name that we can inject into `REACT_APP_API_BASE_URL`?

> One other thing I haven't figured out immediately though - is there a way to open the browser again if you closed it by accident?

If the pod itself is deployed at an address like https://brown-tarsier-2sq3co1q.ws-eu13.gitpod.io/, it seems you can get to the port just by putting in https://{port}-brown-tarsier-2sq3co1q.ws-eu13.gitpod.io/, e.g. https://3000-brown-tarsier-2sq3co1q.ws-eu13.gitpod.io/

> But once the minor teething issue is fixed, I think this is pretty cool and worthwhile to add some notes on the getting started docs for those who don't fancy downloading things onto their machine!

Agree

aapeliv:
If you open the VS Code terminal and type in `gp port 3000` or `gp port 8888`, it'll tell you the URL to access that port

.gitpod.yml
```diff
@@ -0,0 +1,17 @@
...
+  - name: Install and launch the frontend with yarn
+    init: |
+      cd app/frontend 
+      yarn install
+    command: yarn start
```
aapeliv: I guess this needs to be changed to something like `REACT_APP_API_BASE_URL=$(gp port 8888) yarn start`

.gitpod.yml
darrenvong: It doesn't look like we do, I'm guessing it was removed since the prod app are a bunch of minified js and resource files on S3 so we don't actually need it running in a container?

.gitpod.yml
```diff
@@ -0,0 +1,17 @@
+# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
+tasks:
+  - name: Compile the protocol buffers & launch the backend with docker-compose
+    init: |
+      cd app
+      docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
```
darrenvong, line 6: I noticed that this (and commands before and after) runs in a separate terminal parallel to the "install and launch frontend step". Since the frontend also requires the proto files, it is possible that sometimes the frontend build fails because of the race condition.

Is there a way to make this a required task so that we run this command standalone first, then run the `docker-compose ...` and frontend commands in parallel?

pawlean:
Thank you all for your comments and feedback! I'll get back to y'all soon and make changes to the PR. 🎉 I'm glad that for now you're enjoying it and see the value in Gitpod for contributors :D

.gitpod.yml
pawlean: I realised that we have commands such as `gp-sync done` and `gp-sync-await` that allow this to happen. My colleague Jan explained this super well in a [community post.](https://community.gitpod.io/t/gitpod-yml-and-making-sure-init-completes-before-command/2818) However, because `docker-compose` is a command that is non-terminating, these commands, unfortunately, did not work. I'll speak with the Gitpod team next week and see if I can log this feature request. I'm sure that this is possible! 🙏🏼

Instead, I opted for `docker-compose` and `yarn install` to run in separate terminals but at the same time. The only "manual" thing developers would have to do is run a `yarn start` in the front-end terminal after the dependencies have been installed and the back-end is up and running in the other terminal.

pawlean: > One other thing I haven't figured out immediately though - is there a way to open the browser again if you closed it by accident?

CleanShot 2021-08-08 at 16 26 46

.gitpod.yml
pawlean: I've decided to leave out the automation for `yarn start` as I highlighted in [this comment](https://github.com/Couchers-org/couchers/pull/1749#discussion_r684790473).

So I guess you could run `REACT_APP_API_BASE_URL=$(gp port 8888) yarn start` from the terminal or change the file directly to use that API in `.env.development`

pawlean:
> Thank you all for your comments and feedback! I'll get back to y'all soon and make changes to the PR. 🎉 I'm glad that for now you're enjoying it and see the value in Gitpod for contributors :D

Sorry it took me so long to look at this again, folks! Honestly, great work - as I've been trying to configure Gitpod on your project, I'm learning so much about Couchers and just think it's brilliant. Good luck with it!

Happy to answer any questions on Gitpod! Completely optional, but if you'd like to try out Gitpod moving forward (and others reading this wants to contribute solely on the Cloud) Feel free to add this badge to your README / Documentation:

`[Open in Gitpod](https://gitpod.io/#https://github.com/Couchers-org/couchers)` 

[Open in Gitpod](https://gitpod.io/#https://github.com/Couchers-org/couchers)

It will allow ANYONE to get started anywhere. 💯

.gitpod.yml
aapeliv: I don't have that tab on the left!

darrenvong:
Another thing I didn't manage to figure out how to change... do you happen to know why the text in the integrated terminal is appearing all blurry like this @pawlean? Are there some settings I can tweak? Weirdly it only happens to the terminal, but it's all fine in the code editor itself. 🤔 
Integrated terminal in GitPod

.gitpod.yml
darrenvong: Do you have the sidebar with other tabs/icons at least? If you do you can right click to bring up the context menu and you might be able to make it appear if you click "Remote Explorer"?
Context menu shown when right clicking on the tab bar in VS Code

darrenvong: > The only "manual" thing developers would have to do is run a yarn start in the front-end terminal after the dependencies have been installed and the back-end is up and running in the other terminal.

I think that's fine in my opinion. At least the hard bit of setting up the environment is done so we can add some notes/instructions to let people know they need to do that if they go with the Gitpod route of developing.

aapeliv:
I've done a few fixes to this to automate it further: https://github.com/Couchers-org/couchers/compare/9422c3d0934dc9fe809dc0ef76aa4b0b9978cd9e...backend/feature/gitpod-test

https://gitpod.io/#https://github.com/Couchers-org/couchers/tree/backend/feature/gitpod-test

The only thing remaining is CORS issues with hitting the API from the frontend.

darrenvong:
> The only thing remaining is CORS issues with hitting the API from the frontend.

I'm guessing it's because GitPod exposes different ports as different subdomains, which means they are treated as different origins by the browser 😕 . Is there even a way around this without touching the Envoy config?

aapeliv:
Adding `suffix: .gitpod.io` in envoy.yaml seems to fix the cors issue.

Then it has issues with the cookie domain: the cookies in dev are by default set on `localhost`, but now need to be set on the right domain. There's no `gp` command to get the domain though, but something like `gp url 8888 | cut -c 9-` seems to work (but is a bit brittle)?

pawlean:
> Adding `suffix: .gitpod.io` in envoy.yaml seems to fix the cors issue.
> 
> Then it has issues with the cookie domain: the cookies in dev are by default set on `localhost`, but now need to be set on the right domain. There's no `gp` command to get the domain though, but something like `gp url 8888 | cut -c 9-` seems to work (but is a bit brittle)?

It might be worth checking out our local companion app: https://www.gitpod.io/blog/local-app It's currently in preview (https://www.gitpod.io/docs/develop/local-companion) but this may be a good way to get around this!

pawlean:
> Another thing I didn't manage to figure out how to change... do you happen to know why the text in the integrated terminal is appearing all blurry like this @pawlean? Are there some settings I can tweak? Weirdly it only happens to the terminal, but it's all fine in the code editor itself. 🤔
> Integrated terminal in GitPod

A fix is on the way for this, I believe! Thanks for raising 🧡

pawlean:
> I've done a few fixes to this to automate it further: [9422c3d...backend/feature/gitpod-test](https://github.com/Couchers-org/couchers/compare/9422c3d0934dc9fe809dc0ef76aa4b0b9978cd9e...backend/feature/gitpod-test)
> 
> https://gitpod.io/#https://github.com/Couchers-org/couchers/tree/backend/feature/gitpod-test
> 
> The only thing remaining is CORS issues with hitting the API from the frontend.

I just tested this, looks great!

darrenvong:
> It might be worth checking out our local companion app: https://www.gitpod.io/blog/local-app It's currently in preview (https://www.gitpod.io/docs/develop/local-companion) but this may be a good way to get around this!

This looks interesting! So in summary, is this how the local companion app works?
- We download the binary for our computer's OS, install the local companion app and have that running on our computer
- We spin up Gitpod as before by prefixing the repo URL with gitpod.io/# with the config we have here
- Instead of accessing the app through `{port}-xx.xx.gitpod.io`, because we have the local companion app running, `port` also forwards to `localhost` so we can also access the app via `localhost:{port}` as if we're running the app locally
- The cookie domain now matches up since it's set for `localhost`... profit? 💰

pawlean:
👀  I thought that this would be relevant here too https://github.com/gitpod-io/gitpod/issues/5216

darrenvong:
Just had another play around and gotten further with the Gitpod companion app... but now I'm stuck in jail with the "Unauthorised" error instead of the form that lets you accept TOS etc to get out of jail:
Screenshot 2021-09-05 at 14 58 03

Looks like `jailInfo` is failing with the error. Any ideas ~~what might be causing that~~ why that might be the case  @aapeliv or @lucaslcode?

lucaslcode:
Perhaps cors-related - cors might prevent the authorization header being set.

darrenvong:
Spot on, it was using the `gitpod.io` address for the API. I removed the `REACT_APP_API_BASE_URL` override in the command and it's all working nicely when used with the Gitpod companion app 🚀 

Gonna re-open this PR with the branch where Aapeli's added some stuff in, as I think this is a cool alternative to get started with the repo.

PR #1756, opened 2021-08-02
Events part 4: event discussion + unhiding join/leave event button

darrenvong:
Covers more of #541. Feature in question pretty much in title.

I'm leaving the button as the binary join/leave thing for the first version, as the API currently lists people who are "maybe" going as an attendee and the distinction isn't clear between "going" and "maybe" - i.e. should we include people who said "maybe" in attendees?

**Desktop**
Screenshot 2021-08-02 at 03 06 18

**Mobile**
Screenshot 2021-08-02 at 03 06 53

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/communities/events/EventPage.tsx
+++ b/app/frontend/src/features/communities/events/EventPage.tsx
@@ -226,24 +226,24 @@ export default function EventPage() {
                   </Typography>
                 )}
               </div>
-              {process.env.REACT_APP_IS_COMMUNITIES_PART2_ENABLED && (
-                <Button
-                  className={classes.attendanceButton}
-                  loading={isSetEventAttendanceLoading}
-                  onClick={() => setEventAttendance(event.attendanceState)}
-                  variant={
-                    event.attendanceState ===
-                    AttendanceState.ATTENDANCE_STATE_GOING
-                      ? "outlined"
-                      : "contained"
-                  }
-                >
-                  {event.attendanceState ===
+
+              <Button
+                className={classes.attendanceButton}
+                loading={isSetEventAttendanceLoading}
+                onClick={() => setEventAttendance(event.attendanceState)}
+                variant={
+                  event.attendanceState ===
                   AttendanceState.ATTENDANCE_STATE_GOING
-                    ? LEAVE_EVENT
-                    : JOIN_EVENT}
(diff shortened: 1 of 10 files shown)


app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -162,7 +167,9 @@ export default function EventPage() {
           eventKey(eventId),
           updatedEvent
         );
-        queryClient.invalidateQueries(eventKey(eventId));
+        queryClient.invalidateQueries(eventKey(eventId), {
+          refetchActive: false,
+        });
```
darrenvong, line 172: Can't remember from memory whether our other API mutations also return back the updated object we can use to update the cache (with "optimistic" update, in quotes since it's done in `onSuccess` rather than `onMutate`), but if it does, could be good to follow this to reduce API refetch calls from invalidating the query, since the returned result from the mutation should be "fresh enough" without needing to check API again straight away.

Vinalti:
> 
> I'm leaving the button as the binary join/leave thing for the first version, as the API currently lists people who are "maybe" going as an attendee and the distinction isn't clear between "going" and "maybe" - i.e. should we include people who said "maybe" in attendees?
> 

Maybe the best would be to show the « `Maybe` »  has « `Interested` ».  This way you can have a button « join event » and a button « interested » (or "receive updates", "follow", etc. ). Also, without listing the people we could just display the counts of interested people. Or listing them in a separate section.

darrenvong:
> > I'm leaving the button as the binary join/leave thing for the first version, as the API currently lists people who are "maybe" going as an attendee and the distinction isn't clear between "going" and "maybe" - i.e. should we include people who said "maybe" in attendees?
> 
> Maybe the best would be to show the « `Maybe` » has « `Interested` ». This way you can have a button « join event » and a button « interested » (or "receive updates", "follow", etc. ). Also, without listing the people we could just display the counts of interested people. Or listing them in a separate section.

That would be nice, and what you're suggesting is essentially how Facebook does it! But unfortunately the API doesn't give that distinction at the moment and so there's not enough information on the frontend to display things that way. I suspect we might want to change that going forward...

lucaslcode (approved):

PR #1757, opened 2021-08-02
add default prompts to edit profile

ItsiW:
**Frontend checklist**
- [ ] All tests pass


app/frontend/src/features/profile/edit/constants.ts
```diff
@@ -1,21 +1,36 @@
...
 <br>
+e.g. I think couch surfing should be free...
+<br>
+# My favorite travel story
+<br>
+e.g. One time I...
 `;
 
 export const DEFAULT_HOBBIES_HEADINGS = `# Art
+
```
aapeliv: Missing `<br>`

aapeliv (changes requested):
The "e.g." isn't that pleasing, but 🤷‍♂️

aapeliv:
Also I think it would be good to give better examples. I personally find it a bit lame when people just list bands/movies/etc in their free time section

ItsiW:
> The "e.g." isn't that pleasing, but man_shrugging

This is intentional. It needs to look a little bad so it prompts people to delete the line. 

Could you provide some examples of desired text? I've mostly only ever seen lists, I tend to not fill these sections at all.

aapeliv (approved):
LGTM

Issue #1760, opened 2021-08-02
Server Side Rendering


aapeliv:
Some things that we'd want to do, and why we care, in order of importance:

1. Pre-rendering a list of static pages (e.g. /about, /blog/..., etc; list known at build time), so we could do the landing site in react and burn the vue shitshow
2. Nice link previews for profiles (or TBA invitation links)
3. Incrementally adding rendering to a set of other classes of pages, e.g. events, profiles, etc. This would hugely help SEO with crawlers being able to read some real site content.
4. Rendering everything server side to make the experience faster for users

1 is pretty high priority, but seems it might be possible with CRA much easier: https://create-react-app.dev/docs/pre-rendering-into-static-html-files. We will want 2 at some point too, but it might be able possible to do some server-side tweakery that just replaces/adds meta tags/etc. It would be great to do 3 sometime too, especially once we start having more organic content. I'm not sure if we'll ever care about doing 4: it's probably expensive and quite complicated.

lucaslcode:
I am against *real* server side rendering for a site like this, it will make server responses slow and hosting expensive. 

However we can definitely do static site generation using react snapshot (or whatever the latest one is) and hydrate, which I'm assuming it does. 

I'd be a bit concerned about our time to interactive since we're such a chunky site, we'd need to test it out.

lucaslcode:
> do some server-side tweakery that just replaces/adds meta tags/etc

No need for that, we can use `react-helmet` (actually I was going to open an issue about that to fix our page titles, I must have forgotten)

lucaslcode:
I'm not sure off the top of my head how to manage this statically for pages which are different when logged in without having a "flash" of the non-logged in version

aapeliv:
> > do some server-side tweakery that just replaces/adds meta tags/etc
> 
> No need for that, we can use `react-helmet` (actually I was going to open an issue about that to fix our page titles, I must have forgotten)

That's sill on the client side javascript, right? Won't be picked up by non-JS clients.

> I'm not sure off the top of my head how to manage this statically for pages which are different when logged in without having a "flash" of the non-logged in version

We can just have the middleware/renderer/proxy never return any pre-rendered stuff if the `couchers-sesh` cookie is present.

lucaslcode:
> Won't be picked up by non-JS clients.

It will if we use `react-snap`!

aapeliv:
> I am against _real_ server side rendering for a site like this, it will make server responses slow and hosting expensive.

[...]

PR #1761, opened 2021-08-02
SSR stuff

aapeliv:
Been wasting way too much time playing with this... I just stumbled across [`ReactDOMServer.renderToString()`](https://reactjs.org/docs/react-dom-server.html#rendertostring) and it looked *so deceptively simple*. But of course, all the fuckery is in driving Create React App, babel, webpack, and the concoction they've taped together to make react somewhat pleasant to use for newbies like me.

I'm by no means suggesting we merge this, but it might be of interest and I wanted to share.

I originally typed a bunch of other things here, but thought I better move it into an issue (#1760) since it's better to keep the conversation about if/how we should approach SSR there.


aapeliv:
This complains a lot about `useLayoutEffect`:

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

darrenvong:
Yeah I believe certain hooks can't be run (like `useLayoutEffect`) in a server rendered environment, and so any components that uses this hook and needs to be whacked on the server side will need to be re-designed slightly for things to work.

lucaslcode:
Closing in favour of #1965 for now

PR #1764, opened 2021-08-03
Fix alphabetization

### LINKED ISSUE #1750: Alphabetization by region code instead of name
image

UAE and Spain out of place

ShmuelTreiger:
~~Weird errors locally, seeing if they persist on gitlab~~ No errors on Gitlab, tests pass

Closes #1750

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -303,7 +303,6 @@ class API(api_pb2_grpc.APIServicer):
 
             if request.HasField("regions_visited"):
                 session.execute(delete(RegionVisited).where(RegionVisited.user_id == context.user_id))
-                session.flush()
 
                 for region in request.regions_visited.value:
                     if not region_is_allowed(region):
@@ -317,7 +316,6 @@ class API(api_pb2_grpc.APIServicer):
 
             if request.HasField("regions_lived"):
                 session.execute(delete(RegionLived).where(RegionLived.user_id == context.user_id))
-                session.flush()
 
                 for region in request.regions_lived.value:
                     if not region_is_allowed(region):
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -5,7 +5,7 @@ from urllib.parse import urlencode
 import grpc
 from google.protobuf import empty_pb2
 from sqlalchemy.orm import aliased
-from sqlalchemy.sql import and_, func, intersect, or_, union
+from sqlalchemy.sql import and_, delete, func, intersect, or_, union
 
 from couchers import errors, urls
 from couchers.config import config
@@ -302,8 +302,7 @@ class API(api_pb2_grpc.APIServicer):
                     )
 
             if request.HasField("regions_visited"):
-                for region in user.regions_visited:
-                    session.delete(region)
+                session.execute(delete(RegionVisited).where(RegionVisited.user_id == context.user_id))
                 session.flush()
 
                 for region in request.regions_visited.value:
@@ -317,8 +316,7 @@ class API(api_pb2_grpc.APIServicer):
                     )
 
             if request.HasField("regions_lived"):
-                for region in user.regions_lived:
-                    session.delete(region)
+                session.execute(delete(RegionLived).where(RegionLived.user_id == context.user_id))
                 session.flush()
 
                 for region in request.regions_lived.value:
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/servicers/api.py
```diff
@@ -302,8 +302,7 @@ def UpdateProfile(self, request, context):
                     )
 
             if request.HasField("regions_visited"):
-                for region in user.regions_visited:
-                    session.delete(region)
+                session.execute(delete(RegionVisited).where(RegionVisited.user_id == context.user_id))
                 session.flush()
```
ShmuelTreiger: Not sure if flush is still needed

app/backend/src/couchers/servicers/api.py
```diff
@@ -317,8 +316,7 @@ def UpdateProfile(self, request, context):
                     )
 
             if request.HasField("regions_lived"):
-                for region in user.regions_lived:
-                    session.delete(region)
+                session.execute(delete(RegionLived).where(RegionLived.user_id == context.user_id))
                 session.flush()
```
ShmuelTreiger: Ditto

app/backend/src/couchers/servicers/api.py
```diff
@@ -302,9 +302,7 @@ def UpdateProfile(self, request, context):
                     )
 
             if request.HasField("regions_visited"):
-                for region in user.regions_visited:
-                    session.delete(region)
-                session.flush()
+                session.execute(delete(RegionVisited).where(RegionVisited.user_id == context.user_id))
```
aapeliv, line 305: Does this sync session?

app/backend/src/couchers/servicers/api.py
```diff
@@ -840,8 +836,8 @@ def user_model_to_pb(db_user, session, context):
             api_pb2.LanguageAbility(code=ability.language_code, fluency=fluency2api[ability.fluency])
             for ability in db_user.language_abilities
         ],
-        regions_visited=[region.region_code for region in db_user.regions_visited],
-        regions_lived=[region.region_code for region in db_user.regions_lived],
+        regions_visited=[region.code for region in db_user.regions_visited],
```
aapeliv, line 839: Isn't it `region_code` in models?

aapeliv (commented):
The deletes are a lot cleaner. Maybe keep the flush in just in case?

But I think the `region_code` thing might break tests?

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: I'm not sure I understand?

aapeliv: Don't worry, I didn't think that comment through.

Same question as with flush: does it work this way? :P

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: I changed the relationship. They're now relationships between `Region` and `User`, using `RegionsLived` and `RegionsVisited` as secondary tables (which, I think, is how it's supposed to be architecturally). So these `region.code`s are from the `Region` table instead of one of the other `Regions<...>` tables

ShmuelTreiger:
>Maybe keep the flush in just in case?

Deleted flush and added a test to make sure it works properly

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: Certainly seems to. All tests pass. And I added a check to make sure that the regions delete and then get re-added without violating unique constraints

app/backend/src/couchers/servicers/api.py
aapeliv: Ah okay. That causes an extra join; but no problem.

Also I'm not sure what we'll do once we start translating. We'll probably drop the names from the backend and just go by codes.

aapeliv (approved):
LGTM if you tested and works!

app/backend/src/couchers/servicers/api.py
ShmuelTreiger: >Ah okay. That causes an extra join; but no problem.

Yes, but I needed a join for alphabetization anyway (and this is the only way to do it, even with two joins)

>Also I'm not sure what we'll do once we start translating. We'll probably drop the names from the backend and just go by codes.

Was wondering that. Alas.

aapeliv: The thing is that the frontend can do this easily since we don't paginate these, so the frontend can sort a localized list (not just names, but sorting is also different in different locales!)

ShmuelTreiger: True. Easy to remove if/when they ever implement.

PR #1772, opened 2021-08-04
keep search dialog mounted

### LINKED ISSUE #1638: Search filter dialog doesn't keep filters when re-opening
Subject: No (intuitive) way to clear search filters?
Description:
I applied a filter on the map search (Last active: 1 week) and it worked as expected. But when looking to clear that filter afterwards and show all users, I couldn't find a way to do so. If that's not already in the plan, can a button of "Clear all search filters" (or, better, clear search filter for each search field) be added?

Results:

Backend version: prod-004b36cd
Frontend version: develop-9c207b16
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.124 Safari/537.36
Page: https://app.couchers.org/search?lastActive=7
User: natalia (198)

lucaslcode:
closes #1638 

I tried adding a test but it didn't fail...

### DIFF
--- a/app/frontend/src/features/search/FilterDialog.tsx
+++ b/app/frontend/src/features/search/FilterDialog.tsx
@@ -102,7 +102,7 @@ export default function FilterDialog({
       open={isOpen}
       onClose={onClose}
       aria-labelledby="filter-dialog-title"
-      keepMounted={false}
+      keepMounted={true}
     >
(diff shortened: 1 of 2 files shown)


darrenvong (commented):
There's a failing test on CI. I wonder if it's because the element technically doesn't get removed now since `keepMounted` is true?

darrenvong (approved):

PR #1774, opened 2021-08-06
Static rendering of auth pages

### LINKED ISSUE #1760: Server Side Rendering

lucaslcode:
Introduces `react-snap` to close #1760 (for now).

It requires refactoring all the `<Hidden>` components out of the pages we want to statically render, others the snapshot doesn't include all breakpoints.


aapeliv:
Is it not enough to use the css implementation for the <Hidden>s?

lucaslcode:
Yes you can use CSS but the Hiddens actually change the dom based on the
viewport size, so the snapshot is taken of the wrong size

aapeliv:
That's with the JS implementation, right? When I did this, it seemed material-ui has also a CSS implementation that uses media queries and so doesn't change the dom?

https://material-ui.com/components/hidden/#implementations

lucaslcode:
I've introduced changes to the dockerfile to install chromium for `react-snap`. However I suspect it will take 100 years to build now. @aapeliv Can we do some kind of docker caching? We should probably be caching `node_modules` too.

aapeliv:
> Can we do some kind of docker caching?

Yes, it'll cache stuff; but it only caches off the latest `develop` build because it's dumb.



> We should probably be caching `node_modules` too.

It's already doing that. For PRs that don't change yarn.lock it's cached.

lucaslcode:
I am going to remove all the `Hidden`s for the auth pages anyway - though it now works with the css implementation, it has no accessibility at all since everything is now present in the DOM.

lucaslcode:
Gonna redo this in parts

PR #1775, opened 2021-08-07
Events part 5: Create event page

darrenvong:
Includes a new page for creating a new event. More of #541. Almost there...

**Desktop**
Screenshot 2021-08-07 at 03 50 53

**Mobile**
Screenshot 2021-08-07 at 03 51 17
Screenshot 2021-08-07 at 03 51 29

**Frontend checklist**
- [x] All tests pass


app/frontend/src/components/Timepicker/Timepicker.tsx
```diff
@@ -0,0 +1,61 @@
...
+  defaultValue?: Dayjs;
+  id: string;
+  label: string;
+  name: string;
+  className?: string;
+  error?: boolean;
+  errorText?: string;
+}
+
+export default function Timepicker({
```
darrenvong: I'm a bit lost ~~as to why this doesn't work that well at all on keyboard. Tried changing the value back to using JavaScript `Date` instead and that's improved keyboard usability slightly,~~ never mind, that was actually unrelated. Following their example of adding a mask actually screwed it up more! But for some reason you still can't switch to "PM" without opening up the dialog :/ but in their demo it works...

darrenvong: Also the clock seems to only allow drag and drop rather than clicking on the numbers, which makes this pretty difficult/probably less worthwhile to unit test. The author of RTL seems to suggest that too (and recommends doing this in Cypress e2e tests, which we currently don't have yet) https://github.com/testing-library/user-event/issues/440#issuecomment-685010755

app/frontend/src/features/communities/events/EventsList.tsx
```diff
@@ -50,6 +56,13 @@ export default function EventsList({ community }: EventsListProps) {
   return (
     <>
       <SectionTitle icon={<CalendarIcon />}>{EVENTS_TITLE}</SectionTitle>
+      <Button
+        component={Link}
+        className={classes.createEventButton}
+        to={newEventRoute}
+      >
```
lucaslcode: Can this be the same styling (should probably be shared) as the "Create discussion" button on the discussions listing page

app/frontend/src/features/communities/events/CreateEventPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+      if (data.isOnline) {
+        createEventInput = {
+          isOnline: data.isOnline,
+          title: data.title,
+          content: data.content,
+          photoKey: data.eventImage,
+          startTime: finalStartDate,
+          endTime: finalEndDate,
+          // TODO: not hardcode this and allow user to specify community ID?
+          parentCommunityId: 1,
```
lucaslcode, line 54: Until we do this or decide on something, I think add a little subtitle to the checkbox "Virtual events will be added to the global community"

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+  return (
+    <>
+      <ImageInput
+        alt={EVENT_IMAGE_INPUT_ALT}
+        className={classes.imageInput}
+        control={control}
+        id="event-image-input"
+        name="eventImage"
+        type="rect"
+      />
```
lucaslcode: I think this needs a little pencil symbol somewhere to make it more obvious. But that can be another issue/pr.

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <PageTitle>{title}</PageTitle>
+      {error || errors.eventImage || errors.location ? (
+        <Alert severity="error">
+          {error?.message ||
+            errors.eventImage?.message ||
+            // @ts-expect-error
```
lucaslcode: add a reason - "react-hook-form typing incorrect for nested types"

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+          id="title"
+          inputRef={register({ required: TITLE_REQUIRED })}
+          name="title"
+          label={TITLE}
+          variant="standard"
+        />
+        <div className={classes.duoContainer}>
+          <Datepicker
+            control={control}
+            // @ts-expect-error
```
lucaslcode: as above, and the others

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+        <TextField
+          fullWidth
+          id="title"
+          inputRef={register({ required: TITLE_REQUIRED })}
+          name="title"
+          label={TITLE}
+          variant="standard"
+        />
+        <div className={classes.duoContainer}>
+          <Datepicker
```
lucaslcode: I think the default start and end date should be tomorrow (arguable), and the default time should be rounded to the nearest hour.

Also, when you change the date, the end date should also change to the same interval apart (ie. if the dates are 2021-01-01 and 2021-01-03 and the start date is changed to 2021-01-03, the end date should change to 2021-01-05).
Same for the end time.

Maybe we need to add a `onPostChange` callback to call `setValue` in?

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+              inputRef={register({ required: LINK_REQUIRED })}
+              label={VIRTUAL_EVENT_LINK}
+              variant="standard"
+            />
+          ) : (
+            <LocationAutocomplete
+              control={control}
+              fullWidth
+              label={LOCATION}
+              required={LOCATION_REQUIRED}
```
lucaslcode: This is currently stripping parts of the address for the profile editing. Probably needs a prop to disable this.

app/frontend/src/features/communities/events/CreateEventPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+          lng: data.location.location.lng,
+        };
+      }
+      return service.events.createEvent(createEventInput);
+    },
+    {
+      onMutate({ parentCommunityId }) {
+        return { parentCommunityId };
+      },
+      onSuccess(_, __, context) {
```
lucaslcode: redirect to the event on success

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+          fullWidth
+          id="title"
+          inputRef={register({ required: TITLE_REQUIRED })}
+          name="title"
+          label={TITLE}
+          variant="standard"
+        />
+        <div className={classes.duoContainer}>
+          <Datepicker
+            control={control}
```
lucaslcode: This probably needs a rules prop and we should verify that the event is in the future and the end is after the start. The backend gives the error but better catch it early.

app/frontend/src/features/communities/events/EventForm.tsx
```diff
@@ -0,0 +1,257 @@
...
+        <div className={classes.eventDetailsContainer}>
+          <Typography id="content-label" variant="h3" component="p">
+            {EVENT_DETAILS}
+          </Typography>
+          <MarkdownInput
+            control={control}
+            id="content"
+            name="content"
+            labelId="content-label"
+          />
```
lucaslcode: We should add a required prop to catch the error before the backend.

app/frontend/src/features/communities/events/CreateEventPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+          isOnline: data.isOnline,
+          title: data.title,
+          content: data.content,
+          photoKey: data.eventImage,
+          startTime: finalStartDate,
+          endTime: finalEndDate,
+          address: data.location.simplifiedName,
+          lat: data.location.location.lat,
+          lng: data.location.location.lng,
+        };
```
lucaslcode, line 69: missing parent community id?

lucaslcode (changes requested):
Creating a physical event didn't work for me on preview. I go the success message but couldn't find the event after - even after clearing the react-query cache. Is it because the parent community id is missing? Maybe that's also a bug in the backend then - accepting a parent community id of 0? @aapeliv 

I haven't looked thoroughly at the code yet, but it seems largely good, will continue.

app/frontend/src/features/communities/events/CreateEventPage.tsx
darrenvong: I think for offline events, the backend has some magic which figures out which community ID to assign the event to based on location, so it's not required here

app/frontend/src/features/communities/events/EventForm.tsx
darrenvong: Where would you put the pencil without it looking a bit out of place? The way it is in the design would require some refactoring of the image input/possibly re-working it altogether which I don't fancy doing in this PR :p

app/frontend/src/features/communities/events/EventForm.tsx
darrenvong: Hmm interesting, considering I barely changed the implementation of this apart from adding an optional required prop, so it should be the same as before if nothing gets passed in that prop like in all other places.

What locations were you typing in? I can't seem to reproduce it on the preview build

lucaslcode: It is the same as before, but before it strips the street number and name from the address (right?). I tried something like 432 12th Street New York

darrenvong: Oh I misread this and see what you mean now. The search results that come up here do only seem to show the stripped versions of the address, so yeah needs some tweaking...

app/frontend/src/features/communities/events/CreateEventPage.tsx
lucaslcode: Looking at the backend code, it's optional and if you don't include it, it automatically assigns a community as you said. But I think we definitely want to specify here - I thought something was broken because I was using the global community to create a physical event in New York, and then I couldn't see it. But I just found it in the NY community.

app/frontend/src/components/Timepicker/Timepicker.tsx
lucaslcode: This whole picker is a hot mess. I don't like the material time picker anyway - can we just use [`<input type="time">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time)? It's well supported, falls back to text input, and we can add format validation with the message "Must be a valid time in 24-hour format HH:MM" in case it does fall back.

app/frontend/src/features/communities/constants.ts
```diff
@@ -62,5 +63,9 @@ export const SEE_ALL_MODERATORS = "See all Community Builders";
 export const SEE_MORE_DISCUSSIONS_LABEL = "See more discussions";
 export const SEE_MORE_INFORMATION = "See more information";
 export const SEE_MORE_PLACES_LABEL = "See more places";
+export const SHOW_ALL_EVENTS = "Show all events";
 export const UNKNOWN_USER = "Unknown user";
+export const UPLOAD_IMAGE = "Upload image";
+export const UPLOAD_IMAGE_INSTRUCTIONS =
+  "Use the following markdown to display this image:";
```
lucaslcode: I think this can actually be removed

app/frontend/src/features/communities/events/EventPage.tsx
```diff
@@ -218,7 +230,7 @@ export default function EventPage() {
                     >
                       {VIRTUAL_EVENT}
                     </Typography>
-                    <Link href={event.onlineInformation.link}>
+                    <Link href={event.onlineInformation.link} variant="body1">
```
lucaslcode: Does this make a difference?

app/frontend/src/service/events.ts
```diff
@@ -86,3 +90,55 @@ export async function setEventAttendance({
...
+
+export type CreateEventInput = OnlineEventInput | OfflineEventInput;
+
+export async function createEvent(input: CreateEventInput) {
+  const req = new CreateEventReq();
+  req.setTitle(input.title);
+  req.setContent(input.content);
+  req.setStartTime(Timestamp.fromDate(input.startTime));
+  req.setEndTime(Timestamp.fromDate(input.endTime));
+  req.setTimezone("UTC");
```
lucaslcode: Why UTC? I think this should be empty/not set: `// the tzdata timezone identifier, empty string means tz of creating user`

lucaslcode (changes requested):

app/frontend/src/service/events.ts
darrenvong: I'm a little confused about this tbh, since `Timestamp.fromDate` always converts thing back to UTC so that's why I set it... but maybe it's fine to leave it out? 🤷  The frontend also doesn't use it at the moment, it converts the `startTime`/`endTime` from `Timestamp` back to a `Dayjs` object and figures out the correct local time to display there (currently with the user's browser timezone, but could be updated to use the user profile's timezone).

So this separate timezone field feels a bit redundant?

app/frontend/src/features/communities/events/CreateEventPage.tsx
darrenvong: How do we feel about the possibility of letting people create an event that is clearly not in the community though? E.g. they could theoretically create an offline event in Paris in the New York community.

Otherwise, it's better for the cache invalidation too as it currently has to invalidate all community events keys due to not knowing the specific community it's in... so no complaints to do it technically this way!

app/frontend/src/components/Timepicker/Timepicker.tsx
darrenvong: It would make things a lot simpler technically, but I feel like since we are using the Material date picker, having that side by side makes the UI looks quite inconsistent. Might it be worth using the combined time and date picker instead? Not sure if it suffers the same keyboard usability problem I'm having here with the time picker though.

app/frontend/src/features/communities/events/EventForm.tsx
darrenvong: I'll just write for the comment for this first `ts-expect-error`. The rest follows the same pattern so I don't think the comment needs to be repeated all over the place

app/frontend/src/features/communities/events/EventPage.tsx
darrenvong: I swear it did before, not anymore probably ~~because of the theme changes from the donation PR...~~ That's a lie, it's because I added this to the theme as a default prop to use if one isn't specified. I don't see why link text should be smaller than `body1` so I think that's ok.

Removed here!

darrenvong:
Lots of good comments here, I've only partially addressed some of them so this is kinda back to a draft PR state again since there will be broken stuff again. Will look to address them in the coming days...

app/frontend/src/features/communities/events/CreateEventPage.tsx
darrenvong: In addition to the above question... doing this is would be a bit clunky at the moment under `/event/new`, since we don't know the community ID the user comes from if they go to the URL directly...

Should we not have the route and do the `#new` thing like we did for discussions instead?

lucaslcode: I think creating events outside the community is either fine or should be checked in the backend.

I would just do `/event/new/:communityId?` and if one isn't provided, do the automatic way.

darrenvong: > if one isn't provided, do the automatic way.

As in, let the backend decide which community to put the event at as before?

darrenvong: I've done it with route state, not put it under community wrapper though but I think that's ok if the event is being created for the community. It's closer to the design and `/event/new` could be visited standalone by going to the URL directly, so I think it makes sense to not have the wrapper to make things simpler.

darrenvong:
Uhh I tried running the tests with `test-ci` locally with lots of workers and the tests that repeatedly fail seems to be ones we know about already, so hard to say whether anything I've written here is also too slow for GitLab...

darrenvong:
Otherwise, ready to be looked at again

darrenvong:
Assuming the tests will be green (finally, looks like the multiple datepicker clicks were too slow for CI lol), ready to be looked at again now @lucaslcode

lucaslcode (dismissed):
LGTM

Perhaps if using route state, it would be a good idea to test submission with route state and without? Up to you though

darrenvong:
> Perhaps if using route state, it would be a good idea to test submission with route state and without? Up to you though

Good shout, I got lazy because I couldn't get the `LocationAutocomplete` thing working... turned out it was the fake timers messing with it. Glad I looked into it more as the tests caught some bug that'd have slipped in 😅

app/frontend/src/features/communities/events/CreateEventPage.test.tsx
```diff
@@ -75,3 +72,96 @@ describe("Create event page", () => {
...
+    expect(createEventMock).toHaveBeenCalledWith({
+      isOnline: false,
+      lat: 2,
+      lng: 1,
+      address: "test city, test county, test country",
+      title: "Test event",
+      content: "sick social!",
+      photoKey: "",
+      startTime: expect.any(Date),
+      endTime: expect.any(Date),
```
darrenvong, line 135: msw and fakeTimers don't work well together - advancing the timer was causing all sort of weird React warnings 😢  so the check here sadly have to be more generic as I couldn't use fake timer here to lock the Date. At least the more specific check is done with the online event test above though, which is shared logic so I think this will do...

lucaslcode (approved):

PR #1779, opened 2021-08-07
Notifications

### LINKED ISSUE #557: Advanced push notifications
@henrivuorinen and I spent a few hours today and part of it was a discussion on notifications. Here's some preliminary ideas on database models and interfaces for it. I didn't want to open a PR since I imagine this will be worked on a lot later, so I'll just drop the code here.

```py3

class NotificationDeviceType(enum.Enum):
    email = 1
    web = 2
    android = 3
    ios = 4


class NotificationDevice(Base):
    __tablename__ = "notification_devices"

    id = Column(BigInteger, primary_key=True)
    device_type = Column(Enum(NotificationDeviceType), nullable=False)
    user_id = Column(ForeignKey("users.id"), nullable=False)

    # other registration info, bookkeeping, etc, e.g.
    # device_name = Column(String, nullable=False)
    muted = Column(Boolean, nullable=False, default=False)

    user = relationship("User", backref="discussion_subscriptions")


class NotificationType(enum.Enum):
    friend_request = 1
    group_chat_message = 2
    # ...


class Notification(Base):
    __tablename__ = "notifications"

    id = Column(BigInteger, primary_key=True)
    recipient_user_id = Column(ForeignKey("users.id"), nullable=False)

    notification_type = Column(Enum(NotificationType), nullable=False)
    # used to deduplicate notifications and e.g. send only one email from multiple messages
    notification_key = Column(String, nullable=True)

    message = Column(String, nullable=False)

    sent = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
    seen = Column(DateTime(timezone=True), nullable=True)

    user = relationship("User", backref="discussion_subscriptions")


class NotificationDelivery(Base):
    __tablename__ = "notification_deliveries"

    id = Column(BigInteger, primary_key=True)

    notification_id = Column(ForeignKey("notifications.id"), nullable=False)
    notification_device = Column(ForeignKey("notification_devices.id"), nullable=False)

    sent = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
    seen = Column(DateTime(timezone=True), nullable=True)
    cancelled = Column(DateTime(timezone=True), nullable=True)

def send_notification(user, message, notification_key=None):
    pass
```

Each notification should be sent to all devices a user has. Then once the notification has been marked as read/interacted with on one device, the others should be sent some kind of "cancel" signal. Email could work on a "delay": it would only send an email after x minutes and collate any other messages into the same one.

aapeliv:
Closes #557 

This creates a new notification system. The idea is to have push notifications (apps & web), email notifications, and email digests.

This is kind of rough, but I'd like to push it out and try it. It's behind a preview feature flag, so only those who opt in will get notifications. Then once it's out there we can do another pass at it, as long as this initial thing is roughly in the right direction.

**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -239,6 +241,8 @@ class User(Base):
     # e.g. cus_JjoXHttuZopv0t
     stripe_customer_id = Column(String, nullable=True)
 
+    new_notifications_enabled = Column(Boolean, nullable=False, server_default=text("false"))
```
lucaslcode, line 246: Why is this `text("false")`?

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    friend_request__accept = ("friend_request", "accept", [_dt.push, _dt.digest])
+
+    # host requests
+    host_request__create = ("host_request", "create", [_dt.email, _dt.push, _dt.digest])
+    host_request__accept = ("host_request", "accept", [_dt.email, _dt.push, _dt.digest])
+    host_request__reject = ("host_request", "reject", [_dt.push, _dt.digest])
+    host_request__confirm = ("host_request", "confirm", [_dt.email, _dt.push, _dt.digest])
+    host_request__cancel = ("host_request", "cancel", [_dt.push, _dt.digest])
+    host_request__message = ("host_request", "message", [_dt.push, _dt.digest])
+
```
lucaslcode, line 1823: Missing regular messages?

app/backend/src/couchers/servicers/account.py
```diff
@@ -123,6 +123,16 @@ def ChangePassword(self, request, context):
 
             send_password_changed_email(user)
 
+            notify(
```
lucaslcode, line 127: Should these still send emails as well as the new notification?

app/backend/templates/digest.md
```diff
@@ -0,0 +1,19 @@
...
+
+This is your Couchers.org digest of notifications you missed.
+
+{% for notification in notifications %}
+
+Title: {{ notification.title|couchers_escape }}
+
+Content: {{ notification.content|couchers_escape }}
+
+Debug: (topic, action) = ({{ notification.topic|couchers_escape }}, {{ notification.action|couchers_escape }})
```
lucaslcode: What's the plan for this?

app/proto/notifications.proto
```diff
@@ -0,0 +1,27 @@
...
+message GetNotificationSettingsRes {
+  bool new_notifications_enabled = 1;
+}
+
+message SetNotificationSettingsReq {
+  bool enable_new_notifications = 1;
+}
+
+message SetNotificationSettingsRes {
+  bool new_notifications_enabled = 1;
```
lucaslcode: Should this just be an empty response?

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,58 @@
...
+    GetNotificationSettingsRes.AsObject,
+    GrpcError
+  >(
+    notificationSettingsQueryKey,
+    service.notifications.getNotificationSettings
+  );
+
+  const toggleNewNotifications = async () => {
+    await service.notifications.setNotificationSettings(
+      !data?.newNotificationsEnabled
```
lucaslcode: Use a react-query mutation for this, so you have loading and error states. Then you can also make it typesafe by throwing if data isn't defined.

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,58 @@
...
+    );
+    queryClient.invalidateQueries(notificationSettingsQueryKey);
+  };
+
+  const { className } = props;
```
lucaslcode: We usually destructure directly in the function definition.

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,58 @@
...
+      {isLoading ? (
+        <CircularProgress />
+      ) : error ? (
+        <Alert severity="error">{error.message}</Alert>
+      ) : (
+        <Typography variant="body1">
+          The new notification system is currently{" "}
+          <b>{data.newNotificationsEnabled ? "enabled" : "disabled"}</b> for
+          your account.
+          <br />
```
lucaslcode: I imagine there's no margin here. Close the typography here instead (and maybe add a new one for the button?)

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <Button onClick={() => toggleNewNotifications()}>
```
lucaslcode: Add loading state once react-query mutation is used.

lucaslcode (changes requested):
Mostly looked at the frontend stufff

lucaslcode:
Also this shows the problem with getting the latest develop proto files on vercel...

app/backend/src/couchers/models.py
aapeliv: Postgres will take `false` as a string. I don't think `False` works, and this is the same way we do it elsewhere.

app/backend/src/couchers/models.py
aapeliv: Yep

app/backend/src/couchers/servicers/account.py
aapeliv: Yes, until the notifications are ironed out and enabled for everyone, we should still do emails.

app/backend/templates/digest.md
aapeliv: This debug thing can go, but we need to redo the emails at some point.

app/proto/notifications.proto
aapeliv: I suppose it could be, I'll change it

app/backend/src/couchers/models.py
lucaslcode: Just above, it uses a plain string without the Text function. But doesn't matter

kalvdans, line 246:
```suggestion
    # True if the user has opted in to get notifications. This column will be removed in the future when notifications are enabled for everyone.
    new_notifications_enabled = Column(Boolean, nullable=False, server_default=text("false"))
```

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+class NotificationDeliveryType(enum.Enum):
+    # send push notification to mobile/web
+    push = enum.auto()
+    # send individual email immediately
+    email = enum.auto()
+    # send in digest
+    digest = enum.auto()
+
+
+_dt = NotificationDeliveryType
```
kalvdans: Remove underscore. Use `__all__ = [<list of identifiers>]` to denote which top-level objets that are supposed to be public instead.

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    def __init__(self, topic, action, defaults):
+        self.topic = topic
+        self.action = action
+        self.defaults = defaults
+
+    def unpack(self):
+        return self.topic, self.action
+
+    # topic, action, default delivery types
+    friend_request__send = ("friend_request", "send", [_dt.email, _dt.push, _dt.digest])
```
kalvdans: Replace double underscore with single underscore for readability.

app/backend/src/couchers/models.py
kalvdans: How is false here different from having nothing in the notification_preferences table?

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
+
+    topic_action = Column(Enum(NotificationTopicAction), nullable=False)
+    delivery_type = Column(Enum(NotificationDeliveryType), nullable=False)
+    deliver = Column(Boolean, nullable=False)
+
+    user = relationship("User", foreign_keys="NotificationPreference.user_id")
+
+
+class Notification(Base):
```
kalvdans, line 1852:
```suggestion
# Table for accumulating notifications until it is time to send email digest
class Notification(Base):
```

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
+
+    # recipient user id
+    user_id = Column(ForeignKey("users.id"), nullable=False)
+
+    topic_action = Column(Enum(NotificationTopicAction), nullable=False)
+    key = Column(String, nullable=False)
+
+    avatar_key = Column(String, nullable=True)
+    icon = Column(String, nullable=True)
```
kalvdans: Describe if this is an url or base64-encoded inline image or?

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    title = Column(String, nullable=True)
```
kalvdans:
```suggestion
    title = Column(String, nullable=True)  # bold markup surrounded by double asterisks allowed, otherwise plain text
```

app/backend/src/couchers/models.py
```diff
@@ -1776,6 +1786,117 @@ def __repr__(self):
...
+    content = Column(String, nullable=True)
```
kalvdans: markdown without images?

app/backend/src/couchers/notifications/background.py
```diff
@@ -0,0 +1,199 @@
...
+                )
+                .where(subquery.c.key == None)
+            )
+        ).all()
+
+        for user, topic_action, key, notification_id, notification_delivery_id in email_notifications_to_send:
+            topic, action = topic_action.unpack()
+            logger.info(f"Sending notification id {notification_id} to {user.id} ({topic}/{action}/{key})")
+            notification_delivery = session.execute(
+                (select(NotificationDelivery).where(NotificationDelivery.id == notification_delivery_id))
```
kalvdans:
```suggestion
                select(NotificationDelivery).where(NotificationDelivery.id == notification_delivery_id)
```

app/backend/src/couchers/notifications/background.py
```diff
@@ -0,0 +1,199 @@
...
+            ).scalar_one()
+            send_notification_email(notification_delivery.notification)
+            assert notification_delivery.delivery_type == NotificationDeliveryType.email
+            assert not notification_delivery.delivered
+            assert notification_delivery.notification == notification_id
```
kalvdans: Move asserts before sending email, otherwise we risk spamming the same mails over and over.

app/backend/src/couchers/notifications/notify.py
```diff
@@ -0,0 +1,54 @@
...
+from couchers.notifications.utils import enum_from_topic_action
+from proto.internal import jobs_pb2
+
+
+def notify(
+    *,
+    user_id,
+    topic,
+    key,
+    action,
```
kalvdans, line 13: I think it better to let the caller pass in an NotificationTopicAction enum value instead of `topic` and `action`, to make it clear which choices are valid. It seems we get KeyError now if the combination of `topic` and `action` is not supported.

app/backend/proto/internal/jobs.proto
```diff
@@ -11,3 +11,8 @@ message SendEmailPayload {
   string plain = 5;
   string html = 6;
 }
+
+message HandleNotificationPayload {
+  // the database table pk
```
kalvdans: pk?

kalvdans (commented):
Looks generally good. Can be simplified if we skip storing already sent notifications.

app/backend/src/couchers/models.py
aapeliv: > How is false here different from having nothing in the notification_preferences table?

I've decided to make the notification preferences overrides over the default config. So that way there is some default that we can change over time, so for example if there is a new notification type then existing users will get notifications for it.

The other thing is that there should be some security notifications that the user shouldn't be able to turn off (e.g. password changed). Similarly I've been thinking something along the lines of; if a user has said they're hosting last minute, they have to have some type of notification enabled for new host requests.

Anyway, this is just a temporary thing until we phase out emails and switch everyone to thsi.

app/backend/src/couchers/models.py
aapeliv: I'll do them in a later PR

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: OK, it's 'cos I'm a grummy copy-paste code... :S

app/backend/proto/internal/jobs.proto
aapeliv: Clarified

app/backend/src/couchers/notifications/background.py
aapeliv: Good point, thanks

app/backend/src/couchers/models.py
aapeliv: Same as the title, clarified

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: OK

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: How do I do that?

aapeliv:
@lucaslcode: I tried doing the mutation, it looks ugly, I'm not sure if I should combine the `data` state better or if this is what you meant?

aapeliv:
@kalvdans: I fix most of your comments, except the topic-action stuff. The idea of having a topic was that we could then collapse notifications for a given (topic, key) pair. E.g. on messaging apps, mobile phone will collapse all notifications from a given chat. The topic allows doing this, while including different actions in the same stream (e.g. different group chat things).

What do you think?

aapeliv:
@kalvdans: ping 😃

app/web/features/auth/FeaturePreview.tsx
```diff
@@ -29,11 +29,12 @@ export default function FeaturePreview() {
...
         </Typography>
       </MuiAlert>
-      <Section
+      <NotificationSettings className={classes.section} />
+      {/* <Section
         className={classes.section}
         title={t("feature_preview.no_previews.title")}
         content={t("feature_preview.no_previews.explanation")}
-      />
+      /> */}
```
kalvdans: Remove instead of comment out?

kalvdans (dismissed):
> The idea of having a topic was that we could then collapse notifications for a given (topic, key) pair. E.g. on messaging apps, mobile phone will collapse all notifications from a given chat. The topic allows doing this, while including different actions in the same stream (e.g. different group chat things).
> 
> What do you think?

I think it is reasonable, and I'm happy with the patch now! LGTM with or without my small frontend complaint.

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+    service.notifications.getNotificationSettings
+  );
+
+  const mutation = useMutation<
+    SetNotificationSettingsRes.AsObject,
+    GrpcError,
+    NotificationSettingFormData
+  >(
+    ({ newNotificationsEnabled }) =>
+      service.notifications.setNotificationSettings(newNotificationsEnabled),
```
lucaslcode, line 42:
```suggestion
    async ({ newNotificationsEnabled }) => {
      if (newNotificationsEnabed === undefined) throw Error("Current setting not loaded yet.");
      return service.notifications.setNotificationSettings(newNotificationsEnabled);
    },
```
This kind of thing would be most "correct", as it would display the thrown error message in the error alert, but maybe it's a bit pedantic. See below for alternative. (If you went with this you'd have to make the type `newNotificationsEnabled?: boolean;`

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+        queryClient.invalidateQueries(notificationSettingsQueryKey);
+      },
+    }
+  );
+
+  const toggleNewNotifications = async () => {
+    mutation.mutate({
+      newNotificationsEnabled: !data?.newNotificationsEnabled,
+    });
+  };
```
lucaslcode:
```suggestion
  const toggleNewNotifications = async () => {
    if (!data) return;
    mutation.mutate({
      newNotificationsEnabled: !data.newNotificationsEnabled,
    });
  };
```

Alternatively this. It invisibly does nothing if data is undefined, but if that state is impossible (ie. this function is never called because the button isn't shown when data is undefined), maybe it doesn't matter

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+      ) : (
+        <>
+          <Typography variant="body1">
+            The new notification system is currently{" "}
+            <b>{data.newNotificationsEnabled ? "enabled" : "disabled"}</b> for
+            your account.
+          </Typography>
+          {mutation.error ? (
+            <Alert severity="error">{mutation.error.message}</Alert>
+          ) : (
```
lucaslcode: Here just do `{mutation.error && <Alert severity="error">{mutation.error.message}</Alert>}` since you still want the button to be shown if there is a mutation error, to try again. Actually, you can put it at the top under the title.

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+            <Typography variant="body1">
+              <Button
+                onClick={() => toggleNewNotifications()}
+                loading={mutation.isLoading}
+              >
+                {data.newNotificationsEnabled ? "Disable" : "Enable"} new
+                notification system.
```
lucaslcode: Translation keys? Do a key each for the whole sentence (ie. no need to interpolate)

lucaslcode (changes requested):

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: I was thinking that this is a temporary beta feature and won't need to be translated. (Once the new notification system is accepted, the language will change)

lucaslcode: ok

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: I did this, thanks!

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: Done

app/web/features/auth/notifications/NotificationSettings.tsx
```diff
@@ -48,15 +48,19 @@ export default function NotificationSettings({
...
   };
 
   return (
     <div className={className}>
       <Typography variant="h2">{t("notification_settings.title")}</Typography>
-      {(isLoading || !data) ? (
+      {mutation.error && (
+        <Alert severity="error">{mutation.error.message}</Alert>
+      )}
+      {isLoading || !data ? (
```
lucaslcode, line 64: If there is an error while loading, it won't be shown because data is `undefined`. Sorry for not picking up on it before. Maybe add the error at the top like the mutation error.

lucaslcode (changes requested):
Also need to merge in develop

app/web/features/auth/notifications/NotificationSettings.tsx
aapeliv: What do you mean? What should I do?

...sorry

lucaslcode: Just above this line add 
```
{error && <Alert severity="error">{error.message}</Alert>}
```
Then delete what is currently lines 65 and 66

aapeliv: Done, thanks!

aapeliv:
> Also need to merge in develop

Yeah, that'll be a pain, I'll sort it out when this is ready to go

lucaslcode (approved):

PR #1438, opened 2021-06-15
Backend/bugtool/1013 friend accept notification email

austinburnett:
**Backend checklist**
- [ ] Formatted my code by running `isort . && black .` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [ ] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Frontend checklist**
- [ ] Formatted my code with `yarn format && yarn lint --fix`
- [ ] There are no warnings from `yarn lint`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #1782, opened 2021-08-07
Format

aapeliv:
I think this was left over from https://github.com/Couchers-org/couchers/pull/1438.


kalvdans (approved):
We really need a ci job to check formatting :)

PR #1815, opened 2021-08-14
Backend fixes and docs updates

### LINKED ISSUE #1804: Dummy data fixes
As brought up in #1800 by @kalvdans.

* Dummy data tests should load the resources into the database that will be used by devs, not the testing ones
* Should silently ignore exceptions and should instead check something real e.g. whether `count(users)>0`

### LINKED ISSUE #1810: Update docs to use new git commands
* `git switch` instead of `git checkout`
* `git switch -c new-branch-name` instead of `git checkout -b new-branch-name`
* `git restore filename.md` instead of `git checkout filename.md`

### LINKED ISSUE #1814: Update docs to use gitlab docker registry not docker hub
Since this is what we test against and where the container gets built to

aapeliv:
Closing a bunch of random small issues.

Fixes #1804. Fixes #1810. Fixes #1814.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/readme.md
+++ b/app/readme.md
@@ -37,7 +37,7 @@ Install `docker-compose` from <https://docs.docker.com/compose/install/>.
 If you are on **macoS**, **Linux** or **WSL2**, run the following command:
 
 ```sh
-docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+docker run --rm -w /app -v $(pwd):/app registry.gitlab.com/couchers/grpc ./generate_protos.sh
 ```
 
 An alternative on **macOS** is to install the tools locally with `brew install grpc protoc-gen-grpc-web`, then run `./generate_protos.sh`.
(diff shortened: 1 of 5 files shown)


app/readme.md
```diff
@@ -37,7 +37,7 @@ Install `docker-compose` from <https://docs.docker.com/compose/install/>.
 If you are on **macoS**, **Linux** or **WSL2**, run the following command:
 
 ```sh
-docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+docker run --rm -w /app -v $(pwd):/app registry.gitlab.com/couchers/grpc ./generate_protos.sh
```
darrenvong, line 40: Will we need an account with access to our GitLab registry for things to work? Or is it public like the Docker registry we used before?

aapeliv: It's a public github registry: so no need to log in!

kalvdans (approved):
LGTM!

PR #1822, opened 2021-08-15
Content reporting/flagging button

### LINKED ISSUE #1592: Add form to report content next to bug reporting form
Currently to report anything other than a person's profile (via their profile), you must click on "report a problem", click on "report inappropriate behavior or content" and then compose an e-mail manually, from scratch. See screenshot below:
image

People are already hesitant and/or lazy to report things, so setting the bar that high to report things means that a lot will go unreported. To lessen the barrier, we should create a simple form that collects data to send to notify@couchers.org for review. 

So when a user clicks on "report inappropriate behavior or content" they will see a form similar to the one on the profile page. The form should contain 3 fields:

1. Person being reported (if applicable)
2. Reason
3. Description

When the form is sent to notify@couchers.org it should have the subject "Report sent by: [USERNAME]" and should also include the following metadata in the body of the e-mail:
1. Username of person reporting
2. E-mail of person reporting
3. The page where the report was sent from

The header above the field should say "Report inappropriate behavior or content"
The text above the field should say "Please provide us with detailed information about what you are reporting and why. This report will be kept private. Thank you for helping make the community a better place!"

Eventually we will want to make most content flag-able so that when the person clicks the flag this form will be displayed and our system can capture the specific content being reported (rather than rely on the user to provide a description of where the content was and what it is).

aapeliv:
Adds a flag button component. I'll explain my thinking behind this. The idea is that the button takes two props:

* A `contentRef`: a "dev/support readable" reference to the content being reported. Something like `profile/4321` or `community/1/discussion/34/comment/203`
* The username/id of the user who authored that content

This is then combined with the user's reason + description, as well as the page the user is on, and that's shipped over to the ticketing system for our support folks to take care of.

This would allow us to sprinkle these buttons around the place with minimal effort while still being able to backtrack and figure out what content was reported with relative ease.

Screen Shot 2021-08-15 at 9 24 21 pm
Screen Shot 2021-08-15 at 9 24 26 pm
Screen Shot 2021-08-15 at 9 24 38 pm

Closes #1592.

**Backend checklist**
- [x] All tests pass

**Frontend checklist**
- [ ] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -346,6 +347,13 @@ def display_last_active(self):
         """
         return last_active_coarsen(self.last_active)
 
+    @property
+    def user_link(self):
+        """
+        Link to this user's profile
+        """
+        return urls.user_link(self.username)
+
```
kalvdans: Move this helper function to email generation. Backend database has nothing to do with urls.

app/backend/src/couchers/models.py
```diff
@@ -867,26 +875,35 @@ def __repr__(self):
         return f"Message(id={self.id}, time={self.time}, text={self.text}, author={self.author}, conversation={self.conversation})"
 
 
-class Complaint(Base):
+class Report(Base):
```
kalvdans: I think `Complaint` was clearer, since the noun `Report` means more like a monthly summary to me.

app/backend/src/couchers/models.py
```diff
@@ -867,26 +875,35 @@ def __repr__(self):
...
 
+    # reason, e.g. spam, inappropriate, etc
     reason = Column(String, nullable=False)
+    # a short description
     description = Column(String, nullable=False)
 
-    author_user = relationship("User", foreign_keys="Complaint.author_user_id")
-    reported_user = relationship("User", foreign_keys="Complaint.reported_user_id")
+    # a reference to the content
+    content_ref = Column(String, nullable=False)
```
kalvdans: Please describe what it is, url without the host and slug part?

app/backend/src/couchers/models.py
```diff
@@ -867,26 +875,35 @@ def __repr__(self):
...
+    # the author of the content
+    author_user_id = Column(ForeignKey("users.id"), nullable=False)
+
+    user_agent = Column(String, nullable=False)
+    page = Column(String, nullable=False)
```
kalvdans, line 897: Please describe what user_agent and page represent.

app/backend/src/couchers/models.py
```diff
@@ -867,26 +875,35 @@ def __repr__(self):
...
+
+    reporting_user = relationship("User", foreign_keys="Report.reporting_user_id")
+    author_user = relationship("User", foreign_keys="Report.author_user_id")
```
kalvdans: Aren't the reporting user also the author?

app/proto/reporting.proto
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+message ReportReq {
+  // e.g. spam, inappropriate
+  string reason = 1;
+  // short description
+  string description = 2;
+  // a reference to the content being reported, this should be filled out by the UI automatically
+  string content_ref = 3;
```
kalvdans: Please describe the format of the reference, "url without host and slug parts, for example "user/1234"".

app/proto/reporting.proto
```diff
@@ -0,0 +1,30 @@
...
+  // the id or username of the author of this piece of content
+  string author_user = 4;
+
+  string user_agent = 5;
```
kalvdans, line 28: Describe what the frontend should pass as the user_agent.

app/proto/reporting.proto
```diff
@@ -0,0 +1,30 @@
...
+  // url that the user was on when they reported
+  string page = 6;
```
kalvdans, line 30: Rename to url to not confuse with what we call "pages" in the API. Also what should non-web clients send here?

kalvdans (changes requested):

lucaslcode (commented):
Frontend looks good to me

app/backend/src/couchers/models.py
aapeliv: Sure, I'll inject this in another way into the templates

app/backend/src/couchers/models.py
aapeliv: The purpose is for `reporting_user` to be the user that pressed the flag button to report the content to moderators/support.

`author_user` is the user who authored and created that object.

E.g. if there is a spam comment, the person who wrote the comment is the author, the person who clicked the button is the reporting user.

I don't like the confusion here either, but I'm having a hard time coming up with better names?

kalvdans: Aha! I parsed "author" as the author of the report. I can't think about any better name, a comment would be good to clarify it is the author of the offending content.

aapeliv: Will do!

AllhandsNoFeet:
@aapeliv One design-related request on this issue: Can we keep the "..." button since that will be used for other things later, move the flag below the other buttons, align them so they are all in a row (so it looks nicer), and fill the flag icon? See below:

image

jesseallhands:
Can the following changes be implemented before this is pulled? 
1. Can we keep the "..." button since that will be used for other things later (and keep the "report this user" option for those who are unfamiliar with the flag symbol).
2. Use the layout as proposed in screenshot below to keep things cleaner-looking (current design looks strange with the buttons mis-aligned and the flag next to the button). Keep in mind the "message" and "add friend" buttons need to fit somehow once we enable messaging to anyone (not just friends):

image

app/backend/src/couchers/models.py
aapeliv: I could make it `ContentReport`?

app/frontend/src/features/FlagButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+    },
+  },
+}));
+
+export interface ReportButtonProps {
+  contentRef: string;
+  authorUser: string | number;
+}
+
+export default function ReportButton({
```
darrenvong:
```suggestion
export default function FlagButton({
```
For consistency with the component name

app/frontend/src/features/FlagButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+    },
+  },
+  formControl: {
+    "& .MuiOutlinedInput-root": {
+      borderRadius: theme.shape.borderRadius * 3,
+    },
+  },
+}));
+
+export interface ReportButtonProps {
```
darrenvong:
```suggestion
export interface FlagButtonProps {
```

app/frontend/src/features/profile/view/Overview.tsx
```diff
@@ -85,7 +85,10 @@ export default function Overview({ setIsRequesting }: OverviewProps) {
           <>
             <Button onClick={() => setIsRequesting(true)}>{REQUEST}</Button>
             <FriendActions user={user} setMutationError={setMutationError} />
-            <ProfileActionsMenuButton />
+            <FlagButton
+              contentRef={"profile/" + user.userId}
```
```suggestion
              contentRef={`profile/${user.userId}`}
```

darrenvong (commented):
Couple of small nitpicks 😅

darrenvong (commented):
Frontend looks fine to me :P

app/proto/reporting.proto
aapeliv: It's consistent with bug reporting, so I'd prefer to keep as-is

app/backend/templates/content_report.md
```diff
@@ -0,0 +1,39 @@
+---
+subject: "Content Report #{{ report.id|couchers_escape }}"
+---
+
+{% from "macros.html" import button, link, support_email, email_link, newline %}
+
+Content Report #{{ report.id|couchers_escape }}.
+
+
+* Reason{{ newline(html)|couchers_safe }}
```
kalvdans, line 10: Trust markdown's ability to render html from text, instead of cluttering the file like this.

kalvdans (approved):
Backend looks good to me, template could be simplified.

PR #1836, opened 2021-08-17
Blog/bugfix/tott format

naspitha:
Found some more inconsistencies and extra spaces. Also, I change the date format to "Month 1st"on blog.md to match the format used in the _Updates from the development team_.


aapeliv (approved):
Sorry, I didn't realize you did a second PR. I got the emails but thought they were the same one! Thanks, I'll merge

PR #1853, opened 2021-08-22
Update local info tab heading

ItsiW:
The two uses of "General information" heading are confusing, so changing the one on the local info tab


aapeliv:
Tests need fixing

darrenvong (approved):

PR #1856, opened 2021-08-22
fix: make allowCsv=false the default for ProfileTagInput

### LINKED ISSUE #1818: Cannot select languages with comma in name
Subject: "Invalid language"
Description:
I try to save my profile after filling in the data about it.

Results:
It shows in red "invalid language" and doesn't let me proceed.

Backend version: develop-ec85e772
Frontend version: develop-ec85e772
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.131 Safari/537.36
Page: https://app.couchers.org/profile/edit
User: saox (3338)

thdk:
Disable csv functionality in the ProfileTagInput component so that we wont try to split inputs that contain a comma into two separate values.

Since the only place where this component was used is on the Edit Profile page, I have changed the default. This looks more future proof so we don't run into this bug again when we start using this component on other pages where the options have not been cleared from commas.

Fixes #1818

**Frontend checklist**
- [ ] All tests pass
I have some failed tests regarding to the event page and datepicker, something broke on develop branch?

### DIFF
--- a/app/frontend/src/features/profile/ProfileTagInput.tsx
+++ b/app/frontend/src/features/profile/ProfileTagInput.tsx
@@ -125,7 +125,7 @@ export default function ProfileTagInput({
   options,
   label,
   id,
-  allowCsv = true,
+  allowCsv = false,
   className,
 }: ProfileTagInputProps) {
   const classes = useStyles();
--- a/app/frontend/src/features/profile/ProfileTagInput.tsx
+++ b/app/frontend/src/features/profile/ProfileTagInput.tsx
@@ -125,7 +125,6 @@ export default function ProfileTagInput({
   options,
   label,
   id,
-  allowCsv = false,
   className,
 }: ProfileTagInputProps) {
   const classes = useStyles();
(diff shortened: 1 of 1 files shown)


darrenvong:
Given how highly specific this component is (it's a more custom component already built on top of more "primitive" abstractions), it's not really meant to be super re-usable elsewhere... what other use cases do we foresee with the CSV functionality?

Looking at the current usage of it, it seems like it was created as a "shortcut" so power users can copy/paste a bunch of values without having to click through all the language/region checkboxes, but personally I feel like this is a feature that has more downsides than benefits - it allows too much flexibility which means more room for users to trip up (and also more work for us to validate), as proven by the issues caused by this... and now we're turning it off anyway for these fields!

I think we should remove it altogether. Curious to what your or @lucaslcode thoughts are!

thdk:
Removing sounds good for me.

Note that I don't think entering CSV values actually worked. It just filters the list to zero results and pressing enter doesn't add them.

I thought there might have been an option to allow to add items not in the options array but couldn't find it immediately.

lucaslcode:
Yep take it out of you wan, it's not so complex we can't add it back later

darrenvong:
> Note that I don't think entering CSV values actually worked. It just filters the list to zero results and pressing enter doesn't add them.

I think it's because `freeSolo` is not set on the `Autocomplete`, so the users are restricted to the list of options anyway which means the CSV functionality here really is a bit useless. More reason to remove it then and simplify the code a bit.

darrenvong (approved):

PR #1862, opened 2021-08-23
Events part 6: edit event page

darrenvong:
This is probably the final piece for the MVP of #541. It looks more or less identical to the "create event page" since they share the same form UI. This has been intentional with how I put the form UI into its own component:

Screenshot 2021-08-24 at 01 11 36

Also added a new "edit event" button that's only visible to users with edit or community moderation permissions:
Screenshot 2021-08-24 at 01 11 04
Screenshot 2021-08-24 at 01 11 16

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/package.json
+++ b/app/frontend/package.json
@@ -33,11 +33,11 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "CI=true react-scripts build",
-    "test": "react-scripts test",
+    "test": "TZ=UTC react-scripts test",
     "eject": "react-scripts eject",
     "format": "prettier --write src/*",
     "lint": "eslint src/",
-    "test-ci": "tsc && CI=true react-scripts test --coverageReporters=\"text\" --coverageReporters=\"cobertura\" --coverageReporters=\"lcov\" --reporters=\"default\" --reporters=\"jest-junit\" --coverage",
+    "test-ci": "tsc && CI=true TZ=UTC react-scripts test --coverageReporters=\"text\" --coverageReporters=\"cobertura\" --coverageReporters=\"lcov\" --reporters=\"default\" --reporters=\"jest-junit\" --coverage",
     "jest": "jest --modulePaths=src",
     "storybook": "start-storybook -p 6006 -s public",
     "build-storybook": "NODE_PATH=. build-storybook -s public",
--- a/app/frontend/src/features/communities/events/CreateEventPage.test.tsx
+++ b/app/frontend/src/features/communities/events/CreateEventPage.test.tsx
@@ -23,6 +23,14 @@ const createEventMock = service.events.createEvent as jest.MockedFunction<
 >;
 
 describe("Create event page", () => {
+  beforeAll(() => {
+    server.listen();
+  });
+
+  afterAll(() => {
+    server.close();
+  });
+
   beforeEach(() => {
     createEventMock.mockResolvedValue(events[0]);
     jest.useFakeTimers("modern");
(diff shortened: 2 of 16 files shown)


app/frontend/package.json
```diff
@@ -33,11 +33,11 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "CI=true react-scripts build",
-    "test": "react-scripts test",
+    "test": "TZ=UTC react-scripts test",
```
darrenvong, line 36: Wish I found this out earlier for ensuring the timezone used is the same across our computer and CI. The timezone mock thing has been causing me pain for a while as it'd error with something about unknown Date constructor as soon as I need to interact with the date inputs/dayjs.

This seems like the only solution that works by changing the test timezone to UTC with fake timers to "fix" the current date so our test data would work/pass date validation etc

app/frontend/src/features/communities/events/EventForm.test.tsx
```diff
@@ -190,3 +206,33 @@ describe("Event form", () => {
...
+
+    userEvent.type(screen.getByLabelText(TITLE), "Test event");
+    userEvent.type(screen.getByLabelText(LOCATION), "tes{enter}");
+    userEvent.click(
+      await screen.findByText("test city, test county, test country")
+    );
+    userEvent.type(screen.getByLabelText(EVENT_DETAILS), "sick social!");
+
+    jest.useFakeTimers("modern");
+    jest.setSystemTime(new Date("2021-08-01 00:00"));
```
thdk, line 234: Move to beforeAll on line 211? Else adding a second test in this describe method will still have this system date. Or not yet have this system date if the second test was executed before this test that sets the system date.

app/frontend/src/features/communities/events/CreateEventPage.test.tsx
```diff
@@ -131,20 +110,26 @@ describe("Create event page (offline events)", () => {
...
-      startTime: expect.any(Date),
-      endTime: expect.any(Date),
+      startTime: new Date("2021-08-01 01:00"),
+      endTime: new Date("2021-08-01 02:00"),
     });
   });
 
   it("creates on offline event with route state correctly", async () => {
+    jest.useFakeTimers("modern");
+    jest.setSystemTime(new Date("2021-08-01 00:00"));
```
thdk: Can be removed? Same is set in beforeEach

app/frontend/src/features/communities/events/CreateEventPage.tsx
```diff
@@ -97,6 +105,21 @@ export default function CreateEventPage() {
...
+            type="submit"
+          >
+            {CREATE}
+          </Button>
+          <Typography className={classes.disclaimer} variant="body1">
+            {CREATE_EVENT_DISCLAIMER}
+          </Typography>
+        </>
+      )}
+    </EventForm>
```
lucaslcode, line 123: Not a big deal but to me this seems more opaque than a required `submitButton` prop.

lucaslcode (changes requested):
Changing an event from offline to online doesn't move it to the global community.

Also, there's a backend bug - changing something other than the date gives an error about not being allowed to have overlapping occurrences.

app/frontend/src/features/communities/events/EventForm.test.tsx
darrenvong: This is specifically placed here (after the location click from `LocationAutocomplete` on line 225-226) since it has the same problem as the other test - that the `msw` server doesn't respond when the fake timers are on. I probably could leave a comment here too.

As for the concern that this will break things if we add more tests here, I could move the `jest.useRealTimers()` to an `afterEach` instead so it will correctly reset the timer after every test

darrenvong:
> Changing an event from offline to online doesn't move it to the global community.

I think that's also a backend flaw. Currently `UpdateEventReq` doesn't allow a `parentCommunityId` to be specified, so the API will need to be updated to allow that too. Otherwise the current workaround would be an additional API call to `TransferEvent`

app/frontend/src/features/communities/events/CreateEventPage.tsx
darrenvong: I thought about using an explicitly named prop than using `children`, but since this is literally appearing at the bottom of the form, it fits naturally enough with composition for me in this case.

darrenvong:
Also moving the event from online to offline to a different community (going from what you said) is also technically impossible at the moment, as all events will originate from the global community in those cases and there's no way for the frontend to figure out which community a location belongs to... but I guess that logic's kinda broken anyway since we are allowing events to be created in locations that might not match with where the community say it is.

So maybe it doesn't make much sense to move an event to the global community because it's going from offline to online? Or are we allowing it anyway as a non-reversible thing?

lucaslcode (approved):

PR #1863, opened 2021-08-23
Backend/docs/vscode setup

### LINKED ISSUE #1838: How to debug backend in VS Code
I wasn't able to find great documentation on how to debug **just the backend** in VS Code since I have to use a Docker Compose file. I go the /app directory, open VS Code, aand I can't find any info on how to use an existing docker-compose file with VS Code. What development environment is recommended to be used here and if it involves VS Code can you tell me how to set it up?

faizjamil:
Closes #1838 

Had to add a folder and two files to automate setup when reopening the `app` directory in a container in VS Code.


app/.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,44 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.191.1/containers/docker-existing-docker-compose
+// If you want to run as a non-root user in the container, see .devcontainer/docker-compose.yml.
+{
+	"name": "Existing Docker Compose (Extend)",
```
aapeliv: Probably call it "Couchers"

aapeliv (commented):
It's not clear to me how this works and why to do this. Basically you're running the dev environment inside the backend container? How does that help one in debugging, since they're not really running the backend itself, just replacing the command with `sleep` and using it for dev? After setting it up, how do I use it?

Anyway; if this is right and works for you, I'm happy to merge it (after a bit of cleanup)! Could you please remove unrelated comments (it's fine to have explanations for config options when they are used, but I think there's a few unrelated things that are cluttering it). Or maybe that's a bad idea?

app/.devcontainer/devcontainer.json
faizjamil: Yeah I can do a bit of cleanup with the config files I added. My understanding of how being able to access the container the backend runs in is helpful is that it allows you to set breakpoints and run app.py when you want to debug something.

That being said I'll spend a few days making sure this is a good way to do that, I'll look through any documentation Microsoft has on working with Docker Compose files.

faizjamil: As for how to use this

1. Open the`/app` directory in vscode.
2. Make sure you have both the [`Docker`](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-docker) and [`Remote - Containers`](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extensions installed.
3. Open the Command Palate (CTRL+SHIFT+P)
4. Start typing "Reopen in Container"
5. Navigate to and hit ENTER on the `Remote-Containers: Reopen in Container` option. This will reopen the current directory open in VSCode into a container.

aapeliv:
Is this ready @faizjamil ?

faizjamil:
Not quite. While the commits I've made here work for debugging, I'm trying
to go through this process of attaching a debugger to a container. That's
requiring some troubleshooting and research.

faizjamil:
Ok, I just made some slight changes. This should be ready to go. While this may not be the optimal way to work with the backend in VS Code, it gets the job done. I'll update this configuration so that you can attach to the existing `backend` container at a later date, once I figure out how to automatically do the VS Code setup in that `backend` container.

aapeliv (approved):

PR #1871, opened 2021-08-26
fix: event card title overlapping with location

### LINKED ISSUE #1877: Formatting error when event title goes over one line
Subject: Formatting error when event title goes over one line
Description:
viewed event from overview page

Results:
Title of the event is greater than one line, and there's not enough spacing between it and the address

Backend version: develop-2fe0b565
Frontend version: develop-2fe0b565
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:91.0) Gecko/20100101 Firefox/91.0
Page: https://app.couchers.org/community/128/hamburg
User: itsi (2)

darrenvong:
Fixes #1877 

|| Before      | After |
|----------- | ----------- | ----------- |
| Web | Screenshot 2021-08-26 at 03 24 31      | Screenshot 2021-08-26 at 03 21 02 |
| Mobile | Screenshot 2021-08-26 at 03 24 54   | Screenshot 2021-08-26 at 03 21 29 |

For some weird reasons, on small mobile screens sometimes the event title goes onto 3 lines. I think the `LineEllipsis` thing is kinda flaky and doesn't really work when the width of the viewport changes (e.g. when device rotates). But I think this will do for now as a quick fix.

**Frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/frontend/src/features/communities/events/EventCard.tsx
+++ b/app/frontend/src/features/communities/events/EventCard.tsx
@@ -6,19 +6,19 @@ import {
   Theme,
   Typography,
 } from "@material-ui/core";
+import type { TypographyStyleOptions } from "@material-ui/core/styles/createTypography";
 import classNames from "classnames";
 import { AttendeesIcon, CalendarIcon } from "components/Icons";
 import { Event } from "proto/events_pb";
 import { useMemo } from "react";
-import LinesEllipsis from "react-lines-ellipsis";
 import { Link } from "react-router-dom";
 import { routeToEvent } from "routes";
 import { timestamp2Date } from "utils/date";
 import dayjs from "utils/dayjs";
 import makeStyles from "utils/makeStyles";
+import stripMarkdown from "utils/stripMarkdown";
 
 import { getAttendeesCount, ONLINE } from "../constants";
-import getContentSummary from "../getContentSummary";
 import { details, VIEW_DETAILS_FOR_LINK } from "./constants";
 import eventImagePlaceholder from "./eventImagePlaceholder.svg";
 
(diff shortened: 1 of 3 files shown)


app/frontend/src/features/communities/events/EventCard.tsx
```diff
@@ -48,9 +49,23 @@ const useStyles = makeStyles<Theme, { eventImageSrc: string }>((theme) => ({
   },
   title: {
     ...theme.typography.h2,
-    height: `calc(2 * calc(${theme.typography.h2.lineHeight} * ${theme.typography.h2.fontSize}))`,
+    height: `calc(3 * calc(${theme.typography.h2.lineHeight} * ${theme.typography.h2.fontSize}))`,
```
lucaslcode: This should still be 2, no?

lucaslcode (changes requested):

app/frontend/src/features/communities/events/EventCard.tsx
darrenvong: > For some weird reasons, on small mobile screens sometimes the event title goes onto 3 lines. I think the LineEllipsis thing is kinda flaky

That's why it's 3. See the before screenshot for mobile when it's 2

lucaslcode: It's not flaky, it just only updates on load, not resize. We could try to change that or just decide to not worry about viewport size changes, but I think it should be 2 here

darrenvong: Even with 3 * line height, I get this on refresh (initial load, no orientation changes) using the dev tools on iPhone 6/7/8 Plus screen size from the dev tool:
Screenshot 2021-08-29 at 23 11 46

Maybe we need to truncate by character counts as well, but not sure I'm a fan of that as that still doesn't seem to guarantee things are going to be on two lines. Or re-think the design slightly, putting a fixed height on dynamic content is always asking for trouble and a bad pattern imo.

lucaslcode (dismissed):
OK I think it's fine as a bandaid for now.

I noticed facebook manages to get accurate 2-line ellipses. Alternatively, they also do this:
image

Where the title is inside the image and expands upwards . We would still have to truncate on characters but seems more flexible?

darrenvong:
> Where the title is inside the image and expands upwards . We would still have to truncate on characters but seems more flexible?

All done based on character count and CSS? Yeah probably easier for lining things up too as I dropped this extreme example in next and realised what we got before didn't line up anyway when the location is also super long...
Screenshot 2021-08-29 at 23 22 50

darrenvong:
Had a quick spy on the dev tools to see how Facebook did what you said and looks like what they used is [mentioned in CSS tricks too](https://css-tricks.com/line-clampin/#weird-webkit-flexbox-way):

```css
.line-clamp {
  display: -webkit-box;
  -webkit-line-clamp: 2;
  -webkit-box-orient: vertical;  
}
```

Browser support seems pretty good for this (yes, weirdly Firefox also supports this with the `-webkit` prefix even though they're not using WebKit!), so I might go with this actually 🙌

lucaslcode (approved):

PR #1888, opened 2021-08-28
Fix event title overflow

lucaslcode:


darrenvong:
Lol I have a PR opened for this too in https://github.com/Couchers-org/couchers/pull/1871 with working tests ;)

PR #1907, opened 2021-09-01
Add release frontend source map step on CI

### LINKED ISSUE #1784: Upload frontend source maps to Sentry
It's nice to see some errors showing up in Sentry raised by the frontend! However, they aren't very useful at the moment as any caught errors point to the minified files rather than linked to the relevant point in our frontend source code.

The next step would be to update our build command so it uploads the source maps to Sentry on build, so that hopefully we will get nicer highlights of exactly where things error'ed in our code.

darrenvong:
Adds an additional step to CI to upload source maps for the frontend on merge to develop, so that it works more reliably with Sentry error reporting for the frontend. Closes #1784 

The mapping thing actually seems to work before already without this, probably because the [source maps can be accessed publicly](https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/uploading/hosting-publicly/), but from the Sentry docs:

> While making source maps available to Sentry from your servers is the most natural integration, it is not always advisable:
>
> * Sentry may not always be able to reach your servers.
> * If you do not specify versions in your asset URLs, there may be a version mismatch
> * The additional latency may mean that source mappings are not available for all errors.
>
> For these reasons, it is best practice to upload source maps to Sentry beforehand (see below).

It seems like it doesn't hurt to explicitly upload them, especially as this is only configured to do so on `develop` only.

https://gitlab.com/couchers/couchers/-/jobs/1554637146 shows that it works :D

**Frontend checklist**
N/A, no frontend code has changed 😎

### DIFF
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -512,6 +512,24 @@ release:frontend:
       changes:
       - app/**/*
 
+release:frontend-sourcemaps:
+  needs: ["build:frontend"]
+  stage: release
+  image: $FRONTEND_NEXT_TAG
+  inherit:
+    # no docker login
+    default: false
+  script:
+    - cd /app && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN yarn upload-sourcemap
+  rules:
+    - if: ($DO_CHECKS == "true") && ($CI_COMMIT_BRANCH == $RELEASE_BRANCH)
+      changes:
+      - app/**/*
+    - if: ($DO_CHECKS == "true") && ($CI_COMMIT_BRANCH != $RELEASE_BRANCH)
+      changes:
+      - app/proto/**/*
+      - app/frontend/**/*
+
 release:nginx-next:
   stage: release
   script:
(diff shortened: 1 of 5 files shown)


app/.gitlab-ci.yml
```diff
@@ -512,6 +512,20 @@ release:frontend:
...
+  needs: ["build:frontend"]
+  stage: release
+  image: $FRONTEND_NEXT_TAG
+  inherit:
+    # no docker login
+    default: false
+  script:
+    - cd /app && SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN yarn upload-sourcemap
+  rules:
+    - if: ($DO_CHECKS == "true") && ($CI_COMMIT_BRANCH == $RELEASE_BRANCH)
```
aapeliv: This should have the same checks as the others, just `- if: $CI_COMMIT_BRANCH == $RELEASE_BRANCH`

app/.gitlab-ci.yml
```diff
@@ -512,6 +512,20 @@ release:frontend:
       changes:
       - app/**/*
 
+release:frontend-sourcemaps:
+  needs: ["build:frontend"]
+  stage: release
+  image: $FRONTEND_NEXT_TAG
```
aapeliv: I think this should be `$FRONTEND_TAG`; the `next` one is for the dev frontend (which points to `dev-api...`)

aapeliv (approved):
LGTM

PR #1909, opened 2021-09-04
Initial setup for mobile app

nathanschepers:
Initial commit. I'll start following the commit message template once we actually do something.


app/mobile-app/src/storybook/stories/Button/Button.stories.js
```diff
@@ -0,0 +1,20 @@
+import { action } from '@storybook/addon-actions';
```
darrenvong: Appreciate that these are probably examples that come from the setup by default... but since these Storybook entries are all written in `.js`, will this setup work with TypeScript?

darrenvong:
Also, the `app/.DS_Store` file shouldn't be committed and be purged? Looks like it might have been committed after you added that pattern in gitignore

app/mobile-app/src/components/LandingScreen.test.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import renderer from 'react-test-renderer';
```
darrenvong: I think it might be worth setting up [Native testing library](https://callstack.github.io/react-native-testing-library/docs/getting-started) with this initial commit/setup PR. It seems to offer some much nicer API on top of `react-test-renderer`, and encourages similar testing ideas as React Testing Library that we're using for the web frontend, which I think is good to follow here too. What do you think?

app/mobile-app/src/storybook/stories/Button/Button.stories.js
nathanschepers: I haven't personally done this, but this tutorial is essentially what we went through, and does say

"You don't have to add any packages or configuration for writing your stories in TypeScript; just use the .stories.tsx extension."

When @tvcologan is back next weekend we'll make some progress on it and sort out any weirdness.

meanwhile maybe we get rid of the examples. :)

app/mobile-app/src/components/LandingScreen.test.tsx
nathanschepers: Good suggestion, I'll look at it.

lucaslcode:
Side note, can this be formatted with`prettier` with the same rules as the web app?

app/mobile-app/src/storybook/index.js
```diff
@@ -0,0 +1,25 @@
+// if you use expo remove this line
+import { AppRegistry } from 'react-native';
```
darrenvong: Should remove this since this setup example comment says so? 🤔

app/mobile-app/src/storybook/index.js
```diff
@@ -0,0 +1,25 @@
...
+  require('./stories');
+}, module);
+
+// Refer to https://github.com/storybookjs/storybook/tree/master/app/react-native#start-command-parameters
+// To find allowed options for getStorybookUI
+const StorybookUIRoot = getStorybookUI({});
+
+// If you are using React Native vanilla and after installation you don't see your app name here, write it manually.
+// If you use Expo you should remove this line.
+AppRegistry.registerComponent('%APP_NAME%', () => StorybookUIRoot);
```
darrenvong: Ditto of the above about removing this line

app/mobile-app/src/yarn-error.log
```diff
@@ -0,0 +1,17648 @@
+Arguments:
```
darrenvong: We might want to squash merge to purge this from the repo history so it doesn't get bloated

.gitignore
```diff
@@ -5,3 +5,6 @@
 
 # helps Lennart configure his vim 
 pyrightconfig.json
+
```
darrenvong, line 8: Can't comment directly on it so I'll write here - The `.DS_Store` stuff is still in the commit history here so can we delete them please? Also, note to remember to squash merge so we purge extra redundant file like this out of the commit history (since deleting it alone in subsequent commits won't be enough)

app/mobile/src/App.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <Text
+          style={{
+            height: 20,
+            width: 20,
+            display: "flex",
+            justifyContent: "center",
+            textAlign: "center",
+          }}
+        >
+          {isOpen ? "X" : "SB"}
```
darrenvong, line 32: Shouldn't we only show this in dev mode? Don't think we should expose Storybook publicly like this in the prod version of the app

app/mobile/src/App.tsx
```diff
@@ -0,0 +1,54 @@
...
+      <TouchableOpacity
+        onPress={() => setIsOpen(!isOpen)}
+        style={{
+          position: "absolute",
+          top: 30,
+          right: 10,
+          backgroundColor: "red",
+          zIndex: 99,
+          borderRadius: 25,
+        }}
```
darrenvong, line 21: We probably want to avoid from doing inline styles like this with ad-hoc values, as it's gonna introduce a tonnes of inconsistency with magic values all over the place, and prefer using values from a theme of defined tokens similar to how it's done on web.

Maybe one to address once we sort out UX stuff more formally in the app though? Since I don't think MUI we use on web can be used here since it's dependent on web primitives.

app/mobile/src/README.md
```diff
@@ -0,0 +1,24 @@
+# Mobile app development
+
+We are building the mobile app using react native & expo.
+
+To run locally for development purposes, you will need an Android or iOS phone
+with the Expo app installed. The app can also be run in an emulator but emulator
+setup is currently outside the scope of this document.
+
```
darrenvong, line 9: > The app can also be run in an emulator but emulator setup is currently outside the scope of this document.

Are there even some hacky instructions for this you might be able to share on Slack for the time being, until this process is ironed out more?

app/mobile/src/components/LandingScreen.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
+import { StyleSheet, Text, View, Image } from "react-native";
+import LandingImage from "../assets/landing-bg.png";
+import { Dimensions } from "react-native";
+import LargeButton from "./atoms/LargeButton";
+
+const windowWidth = Dimensions.get("window").width;
+const windowHeight = Dimensions.get("window").height;
```
darrenvong, line 8: Might want to move this inside the screen component (function) before the JSX view stuff, so it gets re-computed on screen orientation.

app/mobile/src/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "storybook": "start-storybook -p 7007",
+    "test": "jest",
+    "build-storybook": "build-storybook"
+  },
+  "dependencies": {
+    "@react-native-async-storage/async-storage": "~1.15.0",
+    "expo": "~42.0.1",
+    "expo-cli": "^4.11.0",
+    "expo-status-bar": "~1.0.4",
+    "prettier": "^2.4.0",
```
darrenvong: Move this to `devDependencies` as `prettier` is not needed for the final app to function

app/mobile/src/storybook/stories/Button/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { TouchableHighlight } from "react-native";
+
+export default function Button({ onPress, children }: {onPress: any, children: any}) {
```
darrenvong: Is this an example component from the Storybook setup? Didn't we say we'll delete them to avoid confusion?

It's using things like `propTypes` here which is unnecessary since we are using TypeScript which will check the types for us at compile time rather than at run time.

darrenvong (commented):
Haven't looked at the preview yet, but here are some passing comments from the code I've looked at so far

app/mobile/src/App.tsx
nathanschepers: The current 'deploy on tag' thing is only for internal use and I would like to keep storybook there for now in case someone wants to use it. We have created ticket #2015 to address this issue so that we can have storybook on the preview 'release channel', and turn it off for the production release channel. We'll configure via an env variable or something.

app/mobile/src/App.tsx
nathanschepers: As this is the X button for storybook I think it's ok for now? I don't want to overengineer this, and as we have ticket #2015 above this will actually disappear before we get to any kind of production-ready stage.

app/mobile/src/README.md
nathanschepers: added links to emulator/simulator to the readme.

app/mobile/src/storybook/stories/Button/index.tsx
nathanschepers: we will remove all this stuff.

app/mobile/src/components/LandingScreen.tsx
nathanschepers: to be resolved today by tomas by tyranically locking us to portrait mode.

.gitignore
nathanschepers: we promise to squash merge so as to avoid using the bfg.

app/mobile/src/components/LandingScreen.tsx
nathanschepers: on review, we are already locked to portrait in app.json. resolving.

.gitignore
darrenvong: You still need to make a commit to delete the file first, otherwise squash merge alone won't work as it won't have the delete commit to combine with to get rid of the file in the final squashed (single) commit

app/mobile/src/components/atoms/LargeButton.tsx
```diff
@@ -0,0 +1,50 @@
+import React from "react";
+import { StyleSheet, TouchableOpacity, Text } from "react-native";
+import { COLORS } from "../../constants/colorPalette";
+
+interface ButtonProps {
+  title: string;
+  onPress: any;
```
darrenvong:
```suggestion
  onPress: TouchableOpacityProps['onPress'];
```
Better to type the prop fully rather than using `any` if possible, otherwise it limits the type checking TypeScript gives us

app/mobile/src/storybook/stories/CenterView/index.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
```
darrenvong: Should probably write this in TypeScript too for consistency, otherwise might be confusing for new devs

app/mobile/src/storybook/stories/CenterView/index.js
```diff
@@ -0,0 +1,16 @@
...
+import PropTypes from "prop-types";
+import { View } from "react-native";
+import style from "./style";
+
+export default function CenterView({ children }) {
```
darrenvong:
```suggestion
export default function CenterView({ children }: { children: React.ReactNode }) {
```

app/mobile/src/storybook/stories/CenterView/index.js
```diff
@@ -0,0 +1,16 @@
...
+  return <View style={style.main}>{children}</View>;
+}
+
+CenterView.defaultProps = {
+  children: null,
+};
+
+CenterView.propTypes = {
+  children: PropTypes.node,
+};
```
darrenvong: This can be deleted since type checking is done in TypeScript

app/mobile/src/storybook/stories/LargeButton/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { TouchableHighlight } from "react-native";
+
+export default function Button({ onPress, children }: {onPress: any, children: any}) {
```
darrenvong: I think we can delete this too as it's not your `LargeButton`?

app/mobile/src/components/atoms/LargeButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  theme?: "primary" | "secondary" | "accent";
+  height?: number | string;
+  width?: number | string;
+  fontSize?: number;
+  color?: string;
+}
+
+export default function CustomButton({
```
darrenvong:
```suggestion
export default function LargeButton({
```
We try to make the component name match the file name for consistency and less confusion.

darrenvong (changes requested):
Sorry, final round of code quality comments that shouldn't be too hard to fix up hopefully. Otherwise this is a good starting point! @lucaslcode you might want to have a quick look too?

.gitignore
tvcologan: Right, my bad. I still had the file in the root folder.

darrenvong (dismissed):

darrenvong (approved):
Ok I've done a bit of fiddling (like removing the rogue unnecessary `package-lock.json` and solved the recent merge conflict) here to reduce more back and forth. LGTM for a starting point unless anyone spots anything else

lucaslcode:
Merging this :)

PR #1916, opened 2021-09-06
Gitpodify Couchers wrap up

darrenvong:
Taking over and wrapping up #1749. Added some instructions on using Gitpod in the frontend readme as another option for getting started!

Thanks @pawlean for introducing this to us in the first place :D

Open and test this at: https://gitpod.io/#https://github.com/Couchers-org/couchers/tree/backend/feature/gitpod-test

N/A on checklists, since no code changes to either frontend nor backend

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,15 +6,13 @@ tasks:
       docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
       docker-compose up --build
 
-    env:
-      REACT_APP_API_BASE_URL: $(gp url 8888)
-
   - name: install and launch the frontend with yarn
     before: cd app/frontend
     init: yarn install
-    command: echo "wait for docker-compose to finish, then run yarn start"
+    command: REACT_APP_API_BASE_URL=$(gp url 8888) yarn start
 
 # List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
 ports:
   - port: 3000
     onOpen: open-preview
+  - port: 8888
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -20,4 +20,6 @@ tasks:
 ports:
   - port: 3000
     onOpen: open-preview
+    visibility: public
   - port: 8888
+    visibility: public
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -13,7 +13,7 @@ tasks:
 
   - name: Instructions
     before: cd app/
-    init: until (curl --head -s localhost:8888 localhost:8888 > /dev/null); do echo "Waiting for backend & frontend to come up. This might take a few minutes!"; done
+    init: until (curl --head -s localhost:8888 localhost:8888 > /dev/null); do sleep 1; clear; echo "Waiting for backend & frontend to come up. This might take a few minutes!"; done
     command: echo "Success! You can now launch the frontend by going to" $(gp url 3000)
 
 # List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -6,10 +6,15 @@ tasks:
       docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
       docker-compose up --build
 
-  - name: install and launch the frontend with yarn
+  - name: Install and launch the frontend with yarn
     before: cd app/frontend
     init: yarn install
     command: REACT_APP_API_BASE_URL=$(gp url 8888) yarn start
+
+  - name: Instructions
+    before: cd app/
+    init: until (curl --head -s localhost:8888 localhost:8888 > /dev/null); do echo "Waiting for backend & frontend to come up. This might take a few minutes!"; done
+    command: echo "Success! You can now launch the frontend by going to" $(gp url 3000)
 
 # List the ports to expose. Learn more https://www.gitpod.io/docs/config-ports/
 ports:
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -4,14 +4,14 @@ tasks:
     before: cd app
     init: |
       docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
-      docker-compose up --build
+      COOKIE_DOMAIN=gitpod.io docker-compose up --build
 
   - name: Install and launch the frontend with yarn
     before: cd app/frontend
     init: |
       yarn install
       gp await-port 8888
-    command: BROWSER=none yarn start
+    command: REACT_APP_API_BASE_URL=$(gp url 8888) yarn start
 
   - name: Instructions
     before: cd app/
@@ -22,6 +22,6 @@ tasks:
 ports:
   - port: 3000
     visibility: public
-    onOpen: ignore
+    onOpen: open-preview
   - port: 8888
     visibility: public
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,15 +1,18 @@
 # List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
 tasks:
-  - name: Compile the protocol buffers & launch the backend with docker-compose
+  - name: Compile the protocol buffers, launch the backend with docker-compose
+    before: cd app
     init: |
-      cd app
       docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
       docker-compose up --build
-  - name: Install and launch the frontend with yarn
-    init: |
-      cd app/frontend 
-      yarn install
-    command: yarn start
+
+    env:
+      REACT_APP_API_BASE_URL: $(gp url 8888)
+
+  - name: install and launch the frontend with yarn
+    before: cd app/frontend
+    init: yarn install
+    command: echo "wait for docker-compose to finish, then run yarn start"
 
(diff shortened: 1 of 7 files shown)


aapeliv:
Is there really no way to do this without the local companion app? The whole point of gitpodify would be to be able to run it in their VM and have a completely standardized environment with no local setup needed.

(Also, that link is 404-ing.)

darrenvong:
Unless there is a way to workaround the fact cookies aren't being set properly when it comes from `8888-xxx.gitpod.io` to `3000-xxx.gitpod.io`... I don't think so. Maybe setting `COOKIE_DOMAIN` to `gitpod.io` in the backend environment variable? But that's part of the docker compose so I don't know how we can override that without changing that too, or make a Gitpod copy of docker-compose.

darrenvong:
> (Also, that link is 404-ing.)

The Gitpod badge link? It works for me (takes you to a login page, but should be fine once logged in with GitHub)

aapeliv:
> Maybe setting `COOKIE_DOMAIN` to `gitpod.io` in the backend environment variable?

That should be pretty straightforward, I think.

> > (Also, that link is 404-ing.)
> 
> The Gitpod badge link? It works for me (takes you to a login page, but should be fine once logged in with GitHub)

This link was failing: https://www.gitpod.io/blog/local-app. Either it was down very briefly (it's back up now), or they're doing some odd stuff with SSR and it defaults to a 404! I was getting that on all docs pages for a sec.

darrenvong:
> That should be pretty straightforward, I think.

Can it be done without creating another docker-compose and .env file though? Also I think the media stuff will still be broken, but that's broken locally anyway so maybe less of a big deal

aapeliv:
Yeah I think you can just add env vars before the docker-compose command.

aapeliv:
No, actually; that probably won't work. Sorry, I misunderstood something.

I think we can create an override docker-compose file though. Something like `docker-compose.gitpod.yml` which only contains the overrides; then you can do something like `docker-compose -f docker-compose.yml -f docker-compose.gitpod.yml ...` and it should override the stuff specified in the latter file: https://docs.docker.com/compose/extends/#multiple-compose-files

darrenvong:
That extra `docker-compose.gitpod.yml` override did the trick and fixed the cookies issue! 🚀  So now it's all working within the browser:

https://gitpod.io/#https://github.com/Couchers-org/couchers/tree/backend/feature/gitpod-test

I've also updated the README and removed the reference to the companion app, so now it is really all on in the browser!

readme.md
```diff
@@ -18,12 +20,12 @@ If you are a developer, please read [docs/contributing.md](docs/contributing.md)
...
-* Docker: [docs/docker.md](docs/docker.md) has some tips on how to work with docker and be less miserable
+- Getting started with dev: [app/readme.md](app/readme.md) outlines basic steps to get a dev environment set up, start here!
+- Architectural overview: [docs/architecture.md](docs/architecture.md) outlines the basic workings of the app, the different components, and the way they interact.
+- Developer Contributor Guide: [docs/contributing.md](docs/contributing.md) explains how we work with this codebase and helps you get up and running with development.
+- CI/CD goodies: [docs/cicd.md](docs/cicd.md) lists the different CI/CD functionality that we've got set up and how you can use this to simplify life
+- Database & migrations: [docs/database.md](docs/database.md) explains how to use the database and how we version the database with Alembic and update it when new tables or columns are added
+- Docker: [docs/docker.md](docs/docker.md) has some tips on how to work with docker and be less miserable
```
aapeliv: Please don't replace change the `*` with `-`.

app/frontend/README.md
```diff
@@ -75,4 +88,3 @@ Runs storybook, good for testing and developing components in isolation.
 ### `yarn build`
 
 Builds the app for production to the `build` folder.
-
```
aapeliv: Missing new line at EOF

app/docker-compose.yml
```diff
@@ -41,4 +41,3 @@ services:
       - 5000:5000
     depends_on:
       - backend
-
```
aapeliv, line 44: Missing new line at EOF

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
...
+    before: cd app
+    init: |
+      docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+      docker-compose -f docker-compose.yml -f docker-compose.gitpod.yml up --build
+
+  - name: Install and launch the frontend with yarn
+    before: cd app/frontend
+    init: |
+      yarn install
+      gp await-port 8888
```
aapeliv, line 13: Is this needed anymore? It should be able to `yarn start` before the backend is up?

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
...
+    command: REACT_APP_API_BASE_URL=$(gp url 8888) yarn start
+
+  - name: Instructions
+    before: cd app/
+    init: until (curl --head -s localhost:8888 localhost:8888 > /dev/null); do sleep 1; clear; echo "Waiting for backend & frontend to come up. This might take a few minutes!"; done
```
aapeliv: I think one of these should be `localhost:3000`?

aapeliv (dismissed):
Getting very nitpicky... sorry!

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
...
+    command: echo "Success! You can now launch the frontend by going to http://localhost:3000/"
```
aapeliv: Also; I think this should be `gp url 3000`, not localhost?

aapeliv:
Just tried running this. It's looking good!

A few more things that I noticed:

* Map isn't loading because the dev key isn't set to allow `.gitpod.io`: I'll see if I can fix this one up
* In the same way you set `COOKIE_DOMAIN`, you should set `BASE_URL` to `gp url 3000`, so that e.g. signup links point to the right place. This might be a tad more tricky :/

aapeliv:
> Map isn't loading because the dev key isn't set to allow `.gitpod.io`: I'll see if I can fix this one up

Tried to fix this

.gitpod.yml
darrenvong: Yeah, but the build is usually a lot slower than the frontend dev server so if the preview is up and people click around at that in between point, it'll error until the backend finish building, so I thought it might be better to wait for it to be up first before starting the frontend

aapeliv: Ahh, fair enough. It just adds a bit, I guess since it'll only actually build the frontend when you do `yarn start`?

aapeliv: But yeah; I don't mind leaving it in!

darrenvong:
> > Map isn't loading because the dev key isn't set to allow `.gitpod.io`: I'll see if I can fix this one up
> 
> Tried to fix this

Is that a setting in some couchers mapbox account? Any luck?

> In the same way you set COOKIE_DOMAIN, you should set BASE_URL to gp url 3000, so that e.g. signup links point to the right place. This might be a tad more tricky :/

Yeah I think that might be difficult without some command to substitute in the value into the file itself before spinning the Docker containers up, but then people might end up accidentally committing those changes so I'm not sure that's a good idea :/ lol maybe the companion app is still the easiest way due to how finicky our setup is / Gitpod's domains are too dynamic

.gitpod.yml
darrenvong: Ahh yes, I forgot to change it back when I was testing with the companion app, oops

aapeliv:
> > In the same way you set COOKIE_DOMAIN, you should set BASE_URL to gp url 3000, so that e.g. signup links point to the right place. This might be a tad more tricky :/
> 
> Yeah I think that might be difficult without some command to substitute in the value into the file itself before spinning the Docker containers up, but then people might end up accidentally committing those changes so I'm not sure that's a good idea :/ lol maybe the companion app is still the easiest way due to how finicky our setup is / Gitpod's domains are too dynamic

I think if you add to the gitpod override docker compose file like this

```
version: "3"
services:
  backend:
    environment:
      - BASE_URL
```

with no value, then I think you should be able to pass `BASE_URL` from the command line, by doing `BASE_URL=$(gp url 3000) docker-compose ...`.

https://docs.docker.com/compose/environment-variables/#pass-environment-variables-to-containers

aapeliv:
@darrenvong could you review please? (You made the pr so can't ask for review through gh)

.gitpod.yml
```diff
@@ -0,0 +1,27 @@
+# List the start up tasks. Learn more https://www.gitpod.io/docs/config-start-tasks/
+tasks:
+  - name: Compile the protocol buffers, launch the backend with docker-compose
+    before: cd app
+    init: |
+      docker run --rm -w /app -v $(pwd):/app couchers/grpc ./generate_protos.sh
+      BASE_URL=$(gp url 3000) MEDIA_SERVER_BASE_URL=$(gp url 5000) docker-compose -f docker-compose.yml -f docker-compose.gitpod.yml up --build
+
+  - name: Install and launch the frontend with yarn
+    before: cd app/frontend
```
darrenvong:
```suggestion
    before: cd app/web
```

And can we not rename `web` back to `frontend/blah` again... (unrelated rant)

darrenvong:
Just tried with my latest tweaks and this looks good to me! Can't approve it as I originally opened the PR 😅

lucaslcode (approved):
Works for me :)

PR #1922, opened 2021-09-06
Fix #1770

### LINKED ISSUE #1770: Contributor Form emails sending when fields are empty
from models.ContributorForm

```py3
    def should_notify(self):
        """
        If this evaluates to true, we send an email to the recruitment team.
        """
        return (
            (self.ideas != None)
            | (self.features != None)
            | (self.experience != None)
            | (self.contribute_ways != None)
            | (self.expertise != None)
        )
```

however emails are being sent when all should be None

aapeliv:
Closes #1770

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/migrations/versions/74e7f371be76_make_contribute_ways_a_non_null_array.py
+++ b/app/backend/src/couchers/migrations/versions/74e7f371be76_make_contribute_ways_a_non_null_array.py
@@ -1,0 +1,23 @@
+"""Make contribute_ways a non-null array
+
+Revision ID: 74e7f371be76
+Revises: 725813da484d
+Create Date: 2021-09-06 11:31:50.349578
+
+"""
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "74e7f371be76"
+down_revision = "725813da484d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/migrations/versions/74e7f371be76_make_contribute_ways_a_non_null_array.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "74e7f371be76"
+down_revision = "725813da484d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("UPDATE contributor_forms SET contribute_ways = '{}' WHERE contribute_ways IS NULL")
```
kalvdans, line 18: Where's the `'{}'` syntax documented? I can only see `ARRAY[]::string[]` to denote empty array described [here](https://www.postgresql.org/docs/13/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS)

kalvdans (approved):
Great simplification!

app/backend/src/couchers/migrations/versions/74e7f371be76_make_contribute_ways_a_non_null_array.py
aapeliv: Yeah doesn't seem like it's documented there! I just looked at the database and saw that's how the empty arrays are stored, and it seems to work :P

PR #1923, opened 2021-09-06
Remove 'needs' from sourcemap upload job

aapeliv:
This makes CI wait for all checks to complete before uploading the sourcemaps to sentry, so if some tests fail and the release isn't uploaded to the docker registry, the sourcemaps are also not uploaded

### DIFF
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -513,7 +513,6 @@ release:frontend:
       - app/**/*
 
 release:frontend-sourcemaps:
-  needs: ["build:frontend"]
   stage: release
   image: $FRONTEND_TAG
   inherit:


darrenvong (approved):

darrenvong:
Wait, if the release isn't uploaded, don't we want to avoid uploading source map?

darrenvong (commented):
Wait, if the release isn't uploaded, don't we want to avoid uploading source map?

aapeliv:
> Wait, if the release isn't uploaded, don't we want to avoid uploading source map?

Yes: if the checks fail then no release and no source map. Right now this will upload the source map right after the frontend build completes without waiting for all checks.

(The other thing is that we either want *all* releases or none, so we don't want to depend only on frontend checks, we want all checks to pass. That's achieved by having it be a fresh stage `release` without any `needs` on any of the jobs in that stage.)

aapeliv:
> Right now this will upload the source map right after the frontend build completes without waiting for all checks.

as in before this PR

darrenvong:
Ahh gotcha, sounds like the `needs` was skipping the default check that would have happened oops. LGTM then 👍

aapeliv:
> Ahh gotcha, sounds like the `needs` was skipping the default check that would have happened oops. LGTM then 👍

Ah yeah, I see. Yeah it's confusing.

By default each stage will wait for prior stages: but if you add the `needs`, it turns into a DAG and allows you to break that stage-by-stage order by only waiting for the jobs in `needs`...

PR #1936, opened 2021-09-08
WIP Attempt to use yarn workspaces

lucaslcode:
Having trouble with importing and entry points. CRA doesn't allow
importing uncompiled external code, so the couchers-core typescript has
to be compiled. However I can't get the right import statements working
ideally it would be `import { service } from "couchers-core/service";`

Also I have no idea if the ci stuff is now broken.


darrenvong:
Lol the diff is so large that it's crashing my browser trying to review this 🙃  looks like there are lots of generated code which shouldn't be committed that accounts for most of the diff?

I'm guessing you pulled out the `service` stuff into `couchers-core` and left the rest in `couchers-web`? Doesn't that mean the mobile app is also coupled to this workspace?

I'm kinda feeling a bit iffy about sharing the service client code between web and mobile, as it couples the release cycle between the two apps whenever there's an API update. But maybe that's more of a backend problem on designing more backward compatible APIs, rather than doing what we've been doing now where our API updates are so breaking that it f***s the whole app and it leads to giant PRs of both frontends needing to update.

Actually another thing... we should probably not include React as a dependency in the shared code if we are going down that route (and have it as a peer dependency maybe), otherwise we'll end up with multiple versions of React and upgrading React will become a job of needing to be done across web and mobile, which isn't always ideal, as React Native often lags behind on supporting the latest version of React. I've experienced this at work and this kind of coupling will cause pain further down the line.

lucaslcode:
Yeah sorry, it's because of the removed yarn.lock file and two new ones.

Is there a nice balanced solution between having to rewrite the service layer for react native + maintain two service layers, and having one inflexible one? Out of those two choice, the latter seems more ideal.

lucaslcode:
And yes, I was envisioning a couchers-mobile folder in the same workspace.

darrenvong:
I don't think there's a good in between, it's either:
- share the grpc client/service code and trying to make both apps work,
- maintain two separate versions of it

Re-reading all the ramble I wrote, I think what I said is definitely a backend concern rather than a frontend one as it's ultimately down to how the API changes. If it changes in a way such that it doesn't force us to update both apps in one big bang each time, I think it might actually be fine. So could give this a go once all the generated stuff is purged from the repo? :p

lucaslcode:
Closing this until mobile is in place. Don't delete branch plz

PR #1941, opened 2021-09-09
Info on how to login to the web dev environment

papeldeorigami:
Info on how to login to the development environment - helpful for newcomers like me

### DIFF
--- a/app/web/README.md
+++ b/app/web/README.md
@@ -60,6 +60,8 @@ In the project directory, you can run:
 Runs the app in the development mode.<br />
 Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 
+*hint*: You can find a set of users for logging in at the [dummy data loaded in the docker container](https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/data/dummy_users.json)
+
 The page will reload if you make edits.<br />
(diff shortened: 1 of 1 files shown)


app/web/README.md
```diff
@@ -60,6 +60,8 @@ In the project directory, you can run:
 Runs the app in the development mode.<br />
 Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
 
+*hint*: You can find a set of users for logging in at the [dummy data loaded in the docker container](https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/data/dummy_users.json)
+
```
darrenvong: Good idea! I think it's better to move this under the `Option 1: ...` heading instead, since these accounts won't be available if you target the preview API instead, which most people who only work on the frontend seem to do as they don't want to set up Docker.

darrenvong (changes requested):

app/web/README.md
papeldeorigami: Yeah, makes sense. Done - see next commit.

aapeliv:
You could also just put one of the user+pass combos there so people don't have to dig through the json

darrenvong (approved):

PR #1935, opened 2021-09-08
Feat/frontend/community photo

thdk:
This adds an image input component to the edit community page. Not sure if that's what was originally intended with this issue. Some refactoring might be needed to reuse the page form for other pages as well.  (like how the EditEventForm component is now reused)

**Web frontend checklist**
- [x] Formatted my code with `yarn format && yarn lint --fix`
- [x] There are no warnings from `yarn lint`
- [x] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [x] All tests pass
- [x] Clicked around my changes running locally and it works
- [x] Checked Desktop, Mobile and Tablet screen sizes
  -> styling is not ideal now but that is better done in separate issue

PR #1944, opened 2021-09-10
Refactor/web/community page

thdk:
continues on Couchers-org/couchers#1935 and prepares for Couchers-org/couchers#4180 

Remark: Just a suggestion

Problem with the CommunityBase component it doesn't allow the different pages to have different header components.
This refactoring groups some of the elements from CommunityBase into smaller components which can be used by pages if the page needs it.

There should be no change in functionality or UX with this PR. 

**Web frontend checklist**
- [x] All tests pass


app/web/src/features/communities/CommunityBase.tsx
```diff
@@ -96,58 +54,5 @@ export default function CommunityBase({
...
-              )}`
-            )
-          }
-          labels={communityTabBarLabels}
-        />
-      </TabContext>
-      {children({ community, communitySlug })}
-    </div>
-  );
+  return <div className={classes.root}>{children({ community })}</div>;
```
darrenvong, line 57: The reason this was using a render prop was because the header was assumed to be the same across the pages that use this component. But since you've now stripped everything out to allow more flexibility, the only thing left that's worth sharing is the stateful logic in here, and so there's not much point in keeping `CommunityBase`.

We may as well strip everything out here and turn it all into a custom hook that can be used everywhere else? Maybe move the additional logic into `useCommunity`? (It's more or less analogous to a `Context.Consumer` to `useContext` refactor)

darrenvong: Ahh actually scrap what I said, there are still some error handling logic here so maybe this component isn't completely useless 😅

app/web/src/features/communities/hooks.ts
```diff
@@ -36,21 +37,58 @@ import {
...
+  const { data: { slug } = {} } = queryResult;
+
+  const history = useHistory();
+
+  useEffect(() => {
+    // guarantee the most recent slug is used if the community was loaded from url params
+    // if no slug was provided in the url, then also redirect to page with slug in url
+    if (!id && slug && slug !== communitySlugFromUrl) {
+      communityId && history.push(routeToCommunity(communityId, slug));
+    }
```
darrenvong:
```suggestion
    if (!id && slug && slug !== communitySlugFromUrl && communityId) {
      history.push(routeToCommunity(communityId, slug));
    }
```

darrenvong:
Found a minor issue with this, when navigating to the individual discussion, the "overview" tab is being selected as opposed to the "discussion" tab as before.

Screenshot 2021-09-13 at 02 48 55

app/web/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -88,55 +90,64 @@ export default function DiscussionPage() {
...
+          <CommunityBase communityId={discussion.ownerCommunityId}>
+            {({ community }) => (
+              <>
+                {community.mainPage && (
+                  <PageHeader
+                    page={community.mainPage}
+                    className={classes.header}
+                  />
+                )}
+                <CommunityPageSubHeader community={community} />
```
darrenvong:
```suggestion
                <CommunityPageSubHeader community={community} defaultTab="discussions" />
```
I think that's why what I mentioned before is not updating to the correct tab when navigating to individual discussions

app/web/src/features/communities/EditCommunityInfoPage.tsx
```diff
@@ -109,6 +111,8 @@ export default function EditCommunityPage() {
       {({ community }) => {
         return community.mainPage?.canEdit ? (
           <>
+            <PageHeader page={community.mainPage} />
+            <CommunityPageSubHeader community={community} />
```
darrenvong:
```suggestion
            <CommunityPageSubHeader community={community} defaultTab="info" />
```
I actually forgot to do this in the original implementation!

darrenvong (changes requested):

app/web/src/features/communities/hooks.ts
thdk: yeah I see what you mean here but it makes that if statement even less readable :(
I will change the code so we can infer the correct type for both slug and communityid. Since both of the variables can never be undefined after the query has been successfully fetched.

app/web/src/features/communities/discussions/DiscussionPage.tsx
thdk: also noticed that the edit community page shows the overview tab as selected while that should remain 'local info' 
I decided not to fix that yet, not to broaden the scope of my commits.
Ill have a quick look now though :)

thdk: yeah that seems to do it for both discussions page and for edit community page

app/web/src/features/communities/PageHeader.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+import community from "test/fixtures/community.json";
+
+import PageHeader from "./PageHeader";
+
+export default {
+  component: PageHeader,
+  title: "Communities/PageHeader",
+} as Meta;
+
+const Template: Story<any> = (args) => <PageHeader {...args} />;
```
darrenvong:
```suggestion
const Template: Story<PageHeaderProps> = (args) => <PageHeader {...args} />;
```
Instead of using `any` since the `args` here are the same as the prop type? I realised that means you'll need to extract the type into `PageHeaderProps` rather than inlining it in the component but that shouldn't be a big job!

darrenvong (dismissed):

darrenvong:
Looks like there's some formatting issues on CI, but looks good once that's fixed

thdk:
> Looks like there's some formatting issues on CI, but looks good once that's fixed

tests also failed but they aren't failing locally
just fixed the formatting now, will look into tests if they fail again.

darrenvong (approved):

PR #1954, opened 2021-09-12
[#1915] Frontend - Trim name on login page

### LINKED ISSUE #1915: Name not validated on first page of login
Subject: Name not validated on first page of login
Description:
Type in a name starting with a space, e.g. ` Aapeli`.

Results:
You get a pretty cryptic error message, "Name not supported."

Backend version: develop-43c6f39a
Frontend version: develop-43c6f39a
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:91.0) Gecko/20100101 Firefox/91.0
User: aapeli (1)

goskan93:
Closes #1915

I'm just not sure if it's a good behavior to trim name without user's knowledge, but I don't have any other idea, since it's uncontrolled component

**Web frontend checklist**
- [ :heavy_check_mark: ] Formatted my code with `yarn format && yarn lint --fix`
- [ :heavy_check_mark: ] There are no warnings from `yarn lint`
- [:heavy_check_mark: ] There are no console warnings when running the app
- [ :white_circle: ] Added any new components to storybook - no component added
- [ :white_circle:  ] Added tests where relevant - no tests added
- [ :grey_exclamation: ] All tests pass - still don't know how to run tests on widnows :see_no_evil:
- [:heavy_check_mark:  ] Clicked around my changes running locally and it works
- [ :heavy_check_mark: ] Checked Desktop, Mobile and Tablet screen sizes

### DIFF
--- a/app/web/src/features/auth/signup/BasicForm.tsx
+++ b/app/web/src/features/auth/signup/BasicForm.tsx
@@ -23,6 +23,7 @@ import {
   NAME_EMPTY,
   NAME_LABEL,
   NAME_REQUIRED,
+  NAME_NOT_TRIMMED,
 } from "../constants";
 
 type SignupBasicInputs = {
@@ -42,9 +43,9 @@ export default function BasicForm() {
   const mutation = useMutation<void, GrpcError, SignupBasicInputs>(
     async (data) => {
       const sanitizedEmail = lowercaseAndTrimField(data.email);
-      const saniitzedName = data.name.trim();
+      const sanitizedName = data.name.trim();
       const state = await service.auth.startSignup(
-        saniitzedName,
+        sanitizedName,
         sanitizedEmail
       );
       return authActions.updateSignupState(state);
@@ -86,6 +87,11 @@ export default function BasicForm() {
                 value: nameValidationPattern,
               },
               required: NAME_REQUIRED,
+              validate: {
+                startsWithWhiteSpace: (value) => {
+                  return !value.startsWith(" ") || NAME_NOT_TRIMMED;
+                },
+              },
             });
           }}
           helperText={errors?.name?.message ?? " "}
--- a/app/web/src/features/auth/signup/BasicForm.tsx
+++ b/app/web/src/features/auth/signup/BasicForm.tsx
@@ -22,8 +22,8 @@ import {
   EMAIL_REQUIRED,
   NAME_EMPTY,
   NAME_LABEL,
-  NAME_REQUIRED,
   NAME_NOT_TRIMMED,
+  NAME_REQUIRED,
 } from "../constants";
 
 type SignupBasicInputs = {
--- a/app/web/src/features/auth/signup/BasicForm.tsx
+++ b/app/web/src/features/auth/signup/BasicForm.tsx
@@ -42,7 +42,11 @@ export default function BasicForm() {
   const mutation = useMutation<void, GrpcError, SignupBasicInputs>(
     async (data) => {
       const sanitizedEmail = lowercaseAndTrimField(data.email);
-      const state = await service.auth.startSignup(data.name, sanitizedEmail);
+      const saniitzedName = data.name.trim();
+      const state = await service.auth.startSignup(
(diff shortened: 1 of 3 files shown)


app/web/src/features/auth/signup/BasicForm.tsx
```diff
@@ -82,6 +87,11 @@ export default function BasicForm() {
                 value: nameValidationPattern,
               },
               required: NAME_REQUIRED,
+              validate: {
+                startsWithWhiteSpace: (value) => {
+                  return !value.startsWith(" ") || NAME_NOT_TRIMMED;
+                },
+              },
```
lucaslcode: If we trim the name (which I think is fine, especially since it is easily edited later), I don't think this is needed.

lucaslcode (changes requested):

app/web/src/features/auth/signup/BasicForm.tsx
goskan93: I thought it is better to be sure that no one gives white spaces in purpose (which would be odd, but still), but ok, can change

darrenvong:
One other note - would you be able to fix up the commit messages to something that reflects your changes rather than `v1` / `v2` please? They don't really mean much for future us looking back on how (and why, if the commit messages are used well!) changes are introduced otherwise.

goskan93:
> One other note - would you be able to fix up the commit messages to something that reflects your changes rather than `v1` / `v2` please? They don't really mean much for future us looking back on how (and why, if the commit messages are used well!) changes are introduced otherwise.

you're right, I changed commit messages

app/web/src/features/auth/signup/BasicForm.tsx
```diff
@@ -42,7 +42,11 @@ export default function BasicForm() {
   const mutation = useMutation<void, GrpcError, SignupBasicInputs>(
     async (data) => {
       const sanitizedEmail = lowercaseAndTrimField(data.email);
-      const state = await service.auth.startSignup(data.name, sanitizedEmail);
+      const sanitizedName = data.name.trim();
```
darrenvong, line 45:
```suggestion
      const sanitizedName = lowercaseAndTrimField(data.name);
```
Might be worth using the same function for sanitising the name as the one used for email? I believe it's also case sensitive.

darrenvong:
> * still don't know how to run tests on widnows 🙈

If you are using Powershell, I think you can try running the following instead to run the tests? 

```powershell
($env:TZ = "UTC") -and (react-scripts test)
```

I'm not on Windows to test this and it might not work well though, as I remember reading there's some issues setting timezones this way on Windows. 

Or it might be easier to set up the repo in WSL2 instead, which I believe Lucas uses and he's on Windows.

app/web/src/features/auth/signup/BasicForm.tsx
lucaslcode: We probably don't want to lowercase people's names lol

lucaslcode:
We should be using `cross-env`. In the web folder, run `yarn add -D cross-env` then change the test line in package.json to `"test": "cross-env TZ=UTC react-scripts test",`

lucaslcode (dismissed):
LGTM after running the tests with `cross-env` per my previous comment

app/web/src/features/auth/signup/BasicForm.tsx
darrenvong: Oh wait I thought this is the username oops, yeah probably fine to ignore this then. Doing this kinda stuff on the frontend low key bugs me as I think the backend should really handle it tbh... 🙄

darrenvong (approved):
I think the `cross-env` stuff is unrelated to this PR so it's good to go. I can fix this in an upcoming PR to fix a flaky test I wrote 🙈

darrenvong:
Going to merge this before it gets stale. In future @goskan93 feel free to merge the branch yourself when it's approved :)

goskan93:
aaa sorry didn't even noticed ;(

PR #1957, opened 2021-09-13
fix: crpytic error message with invalid username

### LINKED ISSUE #1950: Misleading error message when user not found
When trying to log in with a bogus username, the error message is cryptic.

Screen Shot 2021-09-11 at 3 28 08 pm

I've checked the inspector, the returned error is "Couldn't find that user.", but instead the UI displays this: "This shouldn't have happened. Please contact us and tell us what you were doing to help us prevent this error in the future."

darrenvong:
Closes #1950. Forgot to switch over to using the `isGrpcError` helper when I refactored on this particular component when I was doing the TS upgrade 🙃 

Only focused on adding regression tests to verify the fix.

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/features/auth/login/Login.test.tsx
+++ b/app/web/src/features/auth/login/Login.test.tsx
@@ -1,7 +1,6 @@
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { ERROR_INFO_FATAL } from "components/ErrorFallback/constants";
-import { LoginRes } from "proto/auth_pb";
 import { service } from "service";
 import wrapper from "test/hookWrapper";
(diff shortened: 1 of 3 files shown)


lucaslcode (approved):
There's [another `instanceof` here](https://github.com/Couchers-org/couchers/blob/090f45949ed0d48b1b2385024ba296479c924f7a/app/web/src/utils/hooks.ts#L106) but it looks okay, right? Edit: yes it's right.

darrenvong:
Yeah that one actually throws an `Error` instance, so technically the fatal error should never be shown. And if it does, the user should definitely report so we can look into why!

PR #1965, opened 2021-09-13
Implement react snap

lucaslcode:
This is the implementation of react-snap with route-splitting and also refactoring of the auth page (landing, login, signup) styles to eliminate the need for `<Hidden>`. You can just do `implementation="css"` but there was a reason I can't remember which made me decide to just refactor completely.

There is no merging of the landing pages and app in this PR.

**Web frontend checklist**
- [x] There are no warnings from `yarn lint` **there are some unused variables in tests unrelated to this PR**
- [x] There are no console warnings when running the app **there is a warning from the cookie banner I think, caused by wrong nesting of Link and MuiLink (I'm guessing)**
- [ ] All tests pass


darrenvong:
Can this be tested reliably on the preview (assuming Aapeli has fixed the routing issues on `next` / preview sites)?

Also can we even test this locally with the dev build (from `yarn start`)? Or is this only testable with the built app since `react-snap` is a post build thing?

darrenvong (changes requested):
Had a quick look around the preview site (assuming it now correctly loads the pages), and let's just say the entire app's styling is a bit f**ked across the board.

Main things I noticed:
- The button stylings are all off and doesn't look like what we had before
- The spacing between certain things seems to have been stripped so now things are mis-aligned and really close to each other
- The link styling might also be off, as the link now blends into the background colour of the cookie banner
- The arrow and the cross in the Autocomplete components seem to be mis-aligned
- The logo is broken - the couch is now black for some reason
- The map is broken completely

I think most of these seem to be because the `snap` styles are injected after and they don't match up with what we have in the components/theme for some reason...

lucaslcode:
The problem was the Mui classes get added again on hydrate which changes the priority. I changed it so the snapshot jss is purged on hydrate. Let's see if it works.

lucaslcode:
I just realized something else - any page we expect the user might land on for a first load, we shouldn't code split. Otherwise, they will get the snapshot, then a loading state while the chunk is loaded, then back to the real thing.

darrenvong:
> I just realized something else - any page we expect the user might land on for a first load, we shouldn't code split. Otherwise, they will get the snapshot, then a loading state while the chunk is loaded, then back to the real thing.

How does this differ when it's one of the private pages? Or is that not a problem because we still client-side render those pages (i.e. falls back to `index.html` as before)?

lucaslcode:
Hmm you're right - other pages will probably get a home page flash. I'm sure there's a way around it though with some devops.

darrenvong:
A lot better than before! The styling on the dashboard and message pages now oddly takes the full width of the screen if you are authenticated and refresh, and only "goes back" to the right styling when I click to the "profile" page and back (which doesn't seem to suffer from this problem 🤔 )

darrenvong:
A video of the described problem above:

https://user-images.githubusercontent.com/13669362/133346318-6226f083-7493-4779-bbcc-2bb27ed9f8c8.mov



Actually in this it shows another problem that might be hard to solve... if user is authenticated and refreshes for whatever reason, when they initially land on the page, we now see a flash of the landing page before it goes onto the current page. Probably because the initial snapshot has no way of knowing the user is authenticated until it's hydrated, does a check and re-renders the right thing

Edit: never mind, the second problem is probably what you described just there

darrenvong:
Just noticed another thing... the positioning of the tagline and the form looks quite inconsistent at the moment between the sign in and sign up page. I'm guessing that can't be intentional? 
Screenshot 2021-09-15 at 16 33 00
Screenshot 2021-09-15 at 16 33 15

lucaslcode:
> The styling on the dashboard and message pages now oddly takes the full width of the screen if you are authenticated and refresh, and only "goes back" to the right styling when I click to the "profile" page and back

Looks like this is because what is rendered is different if you're authenticated or not authenticated, so the `AppRoute` ends up with the wrong classname (fullscreen). There are workarounds, but it's better to just render the same thing, and since we're planning to not use react-snap for authenticated users, this won't be a problem.

I have moved `react-snap` from postbuild (automatic on `yarn build`) to `yarn snap`, so @aapeliv and create and serve the two different versions. Otherwise, there's no way to create a non-snapped version. However this means it won't be using react-snap now on CI.

darrenvong:
> `yarn snap`, so @aapeliv and create and serve the two different versions.

How would this work? The command is run manually, then the generated static HTML files are moved to the appropriate folders, so authenticated pages without a matching static HTML file would fall back to client-side rendering?

darrenvong:
The "sign up" form still floats near the top right, whereas the "log in" form is centred on the right half of the screen on desktop. They should be both centred as that's how it was before?

lucaslcode:
I was thinking
- `yarn build`. Save this artifact somewhere to serve for authenticated users
- `yarn snap`. Save this artifact somewhere else to serve for unauthenticated users.

It's necessary because once you run `react-snap` it seems you loose the original, non-snapped SPA.

lucaslcode:
> The "sign up" form still floats near the top right

This was a bit tricky (because of the fancy scrolling on the signup page), I had to use `minHeight: 100vh` so need to check it's okay on mobile with the bottom navigation.

darrenvong:
> I was thinking
> 
> * `yarn build`. Save this artifact somewhere to serve for authenticated users
> * `yarn snap`. Save this artifact somewhere else to serve for unauthenticated users.
>
> It's necessary because once you run `react-snap` it seems you loose the original, non-snapped SPA.

Is there some ops/nginx magic that could make this work? Otherwise, everyone would start from the snapped SPA right? As the user would initially be unauthenticated, then become authenticated which is where the hydrated version of the app gets a bit f**ked with JSS. Are we planning to force re-load/hard link the user (i.e. "normal" `<a>`s rather than React Router `Link`s) to the authenticated route so it's then loading the client-side SPA?

Also, it might be still nice if we can automate it on CI at some point? I guess not necessary for this initial version.

app/web/src/AppRoutes.tsx
```diff
@@ -68,6 +42,65 @@ import {
   userRoute,
 } from "./routes";
 
+const EditCommunityInfoPage = React.lazy(
```
darrenvong, line 43: We might have to do more ops magic here to ensure we don't flush the cache of old JS chunks now that we started code splitting, otherwise we might start [running into this `ChunkLoadError` issue described here](https://dimaip.github.io/2020/04/25/deploying-apps-with-code-splitting/)

app/web/src/features/auth/AuthPage.tsx
```diff
@@ -26,66 +19,87 @@ import { COUCHERS } from "../../constants";
...
+      lineHeight: "1.15",
+    },
   },
-  introduction: {
-    color: theme.palette.common.white,
-    textAlign: "left",
-    width: "72%",
+  underline: {
+    borderTop: `5px solid ${theme.palette.primary.main}`,
+    boxShadow: "0px 4px 4px rgba(0, 0, 0, 0.25)",
```
darrenvong, line 87: I wonder if it's worth adding stuff like this into the `shadow` part of our theme. Or maybe not since this styling is so distinct from the rest of our authenticated app?

app/web/src/features/auth/login/Login.tsx
```diff
@@ -59,62 +43,50 @@ export default function Login() {
...
+            <Typography
+              classes={{ root: authClasses.subtitle }}
+              variant="h2"
+              component="span"
+            >
+              {INTRODUCTION_SUBTITLE}
+              <Divider className={authClasses.underline}></Divider>
+            </Typography>
+          </div>
+          <div className={authClasses.formWrapper}>
```
darrenvong, line 75: Now this doesn't seem to be centre aligned on mobile, whereas the sign up form does centre align correctly:
Screenshot 2021-09-19 at 02 28 28

darrenvong (changes requested):
Should be the final round of stuff! Apart from the sign in form alignment, the other two are more like question/less blocking stuff

app/web/src/AppRoutes.tsx
lucaslcode: Oh damn

app/web/src/features/auth/AuthPage.tsx
lucaslcode: Yeah it's not a material shadow and not used anywhere else so let's leave it

darrenvong:
Oops think the latest stuff I merged in might have caused a merge conflict 😅

lucaslcode:
It still has the hydration issue anyway...

darrenvong:
> It still has the hydration issue anyway...

How do you see/test this now it's not building the snapped version by default? So the force re-render did nothing?

lucaslcode:
I put it back to snap by default. Yes forced render didn't seem to do anything.

darrenvong:
> I put it back to snap by default.

~~Does it work with the dev server locally with the snapped app? Or do you mean you still have to push to the preview site temporarily?~~ Ahh never mind, I see that you literally changed back

darrenvong:
It seems like the `variant="full-width"` prop value on the `ProfilePage` and `SearchPage` help to not render the wrong classes but I can't tell you why 🙃  otherwise the problem is still there... Wonder if it's related to https://github.com/facebook/react/issues/11064?

lucaslcode:
@darrenvong hang on, maybe this isn't a problem. The issue currently is this:

- authenticated user loads snapped version of home page and gets dodgy dashboard due to unauthenticated snap -> authenticated user mismatch

If we go with the idea to not serve the snapped version to authenticated users, we get the following scenarios:

- authenticated user loads home page - gets normal SPA so it's ok
- unauthenticated user loads home page, logs in and is authenticated - **this is actually okay** because they stay inside the SPA for the route changes so won't encounter any more snapshots

So there are no situations in which the bug will appear as long as we don't serve snapshots to authenticated users, right?

If that's the case (and my last commit doesn't fix the issue), then I guess this is good to go once I fix the merge conflicts and separate snap out of the build process again.

lucaslcode:
Also - just figured out `200.html` is the un-snapped version, so no need to separate.

app/web/package.json
```diff
@@ -34,6 +34,7 @@
   "scripts": {
     "start": "react-scripts start",
     "build": "cross-env CI=true react-scripts build",
+    "postbuild": "react-snap",
```
darrenvong, line 38: Do we still want to leave this as a postbuild step?

app/web/src/AppRoutes.tsx
```diff
@@ -208,23 +243,33 @@ export default function AppRoutes() {
...
-            <NewGuidePage />
-          </AppRoute>
-          <AppRoute isPrivate path={guideRoute}>
-            <PagePage pageType={PageType.PAGE_TYPE_GUIDE} />
-          </AppRoute>
-          <AppRoute isPrivate path={groupRoute}>
-            <GroupPage />
-          </AppRoute>
-        </Switch>
+        <Route
```
darrenvong, line 246: I was wondering for a while why the 404 page didn't show up in dev. How does this extra route helps? Was it just not matching the `AppRoute` within this extra `Switch`?

app/web/package.json
lucaslcode: Yes since 200.html is the original index.html

app/web/src/AppRoutes.tsx
lucaslcode: I'm not sure tbh. Seems like the last 404 rote was being ignored. Maybe `Switch` only renders the first thing that is valid, which was the nested `Switch`

darrenvong (dismissed):

darrenvong:
Is this blocked by ops changes you mentioned on serving the snapped pages/SPA depending on user's auth state?

aapeliv:
> Is this blocked by ops changes you mentioned on serving the snapped pages/SPA depending on user's auth state?

Yes, it was. I've now done that for the previews.

app/web/src/features/auth/login/Login.tsx
```diff
@@ -60,65 +44,52 @@ export default function Login() {
...
+              </Alert>
+            )}
+            <LoginForm />
+            <Typography>
               {NO_ACCOUNT_YET + " "}
-              <Link className={classes.signUpLink} to={signupRoute}>
+              <MuiLink to={signupRoute} component={Link}>
                 {SIGN_UP}
-              </Link>
+              </MuiLink>
```
darrenvong: I guess we can now use the `StyledLink` here instead?

app/web/src/features/auth/signup/Signup.tsx
```diff
@@ -178,66 +197,52 @@ export default function Signup() {
...
+            <Typography
+              classes={{ root: authClasses.subtitle }}
+              variant="h2"
+              component="span"
+            >
+              {INTRODUCTION_SUBTITLE}
+              <Divider className={authClasses.underline}></Divider>
+            </Typography>
+          </div>
+          <div style={{ width: "45%" }}></div>
```
darrenvong: I have to say this filler `div` seems a bit hacky. How come we need this here but not for the login page? It's the same layout as far as I can see...

lucaslcode: Yes, the layout is a bit complicated in order to scroll the tall signup form only but keep the scrollbar on the root page.
To do this, the page is in two parts: the header + left content container, which is `fixed`, and the form, which scrolls, and they have to be siblings. So this hacky div is to match the layout on the login page, where the left and right content are siblings in a flex.

darrenvong (approved):

PR #1966, opened 2021-09-14
Community browser

### LINKED ISSUE #1881: Communities navigation component
There are now too many communities to easily navigate via the list. It would be good if we could do something like nested drop-downs that show sub-communities

lucaslcode:
closes #1881
errors don't seem to be appearing for some reason
need to scroll to last column on select

**Web frontend checklist**
- [ ] All tests pass


app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,166 @@
...
+  },
+}));
+
+export default function CommunityBrowser() {
+  const classes = useStyles();
+  const [selected, setSelected] = useState<Community.AsObject[]>([]);
+
+  const globalCommunityQuery = useCommunity(1);
+
+  //react-query doesn't have useInfiniteQueries
```
thdk, line 45: What do yo mean with react-query does not have useInfinitQueries? :)  https://react-query.tanstack.com/guides/infinite-queries

lucaslcode: I mean there's no infinite equivalent to `useQueries` which takes an array of query parameters and returns an array of query results. This is sometimes necessary because of the rule of hooks means you can't have a dynamic number of hooks (but you can have a dynamic array in one hook).

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,166 @@
...
+          hasMore: query.hasNextPage,
+        },
+      ]);
+    } else {
+      setCachedQueryResults(cachedQueryResults.slice(0, level + 1));
+    }
+    setSelected([...selected.slice(0, level), community]);
+  };
+
+  return (
```
darrenvong, line 99: How the heck does this know to hide the nested community columns when you switch to a different community at the higher level - e.g. going from Oceania (?) to Europe and back to Oceania knows to only show two columns again? Doesn't look immediately obvious to me where "the magic" from just staring at the code

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,166 @@
...
+  );
+  const [cachedQueryResults, setCachedQueryResults] = useState<
+    {
+      data: Community.AsObject[];
+      hasMore?: boolean;
+    }[]
+  >([]);
+
+  const handleClick = (community: Community.AsObject, level: number) => {
+    if (level === selected.length) {
```
lucaslcode, line 72: @darrenvong Level here is the level that was clicked on, so this if block adds the last level to the cache if it was clicked, otherwise an earlier level was clicked so it slices the cache to that level (or actually, one lower)

darrenvong: Ohh so it's using the cached results to display previous levels and React Query's cache for the final level, so that you can fetch more communities at that final level if it has more than the default number of communities in one page?

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,166 @@
...
+            handleClick={(community) => handleClick(community, selected.length)}
+          />
+          {query.hasNextPage && (
+            <Button
+              onClick={() => query.fetchNextPage()}
+              loading={query.isFetchingNextPage}
+            >
+              {LOAD_MORE}
+            </Button>
+          )}
```
darrenvong: Is this meant to appear at the next column? I'd have expect the load more button to be in the same column as the last column and below all loaded communities so far, as it looks a bit odd to have it stretched the entire column

app/web/src/features/dashboard/CommunityBrowser.tsx
thdk: Sorry, I was indeed to quick with my comment :)

app/web/src/features/dashboard/CommunityBrowser.tsx
lucaslcode: it's wip 😉

lucaslcode:
This is done but I need more eyes on it for error displaying. When I try to fake an error by blocking my network in dev tools, it never enters the error state despite showing the error in the console. `console.log(query.status)` always gives `success`. No idea why.

darrenvong:
I can see this so I guess it's fine?
Screenshot 2021-09-17 at 02 20 40

Only worked when I disable network for that particular API call, otherwise it just gets stuck on the spinner for me

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,196 @@
...
+import Alert from "components/Alert";
+import Button from "components/Button";
+import CircularProgress from "components/CircularProgress";
+import {
+  useCommunity,
+  useListSubCommunities,
+} from "features/communities/hooks";
+import { LOAD_MORE } from "features/dashboard/constants";
+import { Community } from "proto/communities_pb";
+import { forwardRef, ReactNode, Ref, useEffect, useRef, useState } from "react";
```
darrenvong: Got an unused variable lint warning here (CI also failed)

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,196 @@
...
+        </ListItem>
+      ))}
+    </List>
+  );
+}
+
+const StyledLink = forwardRef<
+  HTMLAnchorElement,
+  { to: string; children: ReactNode }
+>((props, ref) => <MuiLink component={Link} {...props} ref={ref} />);
```
darrenvong: Might it be worth pulling this into `components`? Seems like something we could re-use wherever we need a React Router link?

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,196 @@
...
+  >([]);
+
+  const lastColumnRef = useRef<HTMLDivElement>(null);
+  useEffect(() => {
+    setTimeout(
+      () =>
+        lastColumnRef.current?.scrollIntoView({
+          behavior: "smooth",
+          block: "end",
+        }),
```
darrenvong: Do we always want to scroll to the bottom of a long list of communities whenever selection changes? Might be a bit annoying if you unselect a previous selected community (maybe by accident), then re-select only to find it scrolls you all the way down when the community you want is near the top half of the list.

app/web/src/features/dashboard/CommunityBrowser.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  it("has correct display logic", async () => {
+    render(<CommunityBrowser />, { wrapper: hookWrapper });
+
+    //initial display
+    expect(await screen.findByRole("button", { name: "11" })).toBeVisible();
+    expect(screen.getByRole("link", { name: "Global" })).toBeVisible();
+
+    //clicking down the tree
+    userEvent.click(screen.getByRole("button", { name: "11" }));
```
darrenvong: I wonder if it's worth storing the button found from `findByRole` and re-use it here (and repeat this comment for other instances below), since we mentioned before `*ByRole` queries can be quite slow? Seeing there are some thorough checks here, repeating the slow query might add up pretty quickly, especially as it doesn't need to be repeated (except when you're checking for the absence of it in the DOM).

But maybe it's not a big deal if CI is happy with it? 🤷  Would be interesting to see if it makes a big difference to the running time of the test.

darrenvong (changes requested):

app/web/src/features/dashboard/CommunityBrowser.tsx
lucaslcode: This is for sideways scrolling, not vertical

darrenvong: Don't you want `inline: "end"` instead? `block` scrolls things vertically according to MDN

app/web/src/features/dashboard/CommunityBrowser.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import { Meta, Story } from "@storybook/react";
```
darrenvong, line 1: This is broken when I checked :(

Screenshot 2021-09-19 at 01 21 40

Missing mock for `mockedService.communities.getCommunity`? Intriguing how this worked before without it 🤔

app/web/src/features/dashboard/CommunityBrowser.tsx
```diff
@@ -0,0 +1,186 @@
...
+          )}
+        </div>
+      ) : (
+        <Alert severity="error">{query?.error?.message || ""}</Alert>
+      )}
+    </div>
+  );
+}
+
+function BrowserColumn({
```
darrenvong, line 142: For parent communities that have no child communities (like "Europe, Africa and Middle East" in next):
Screenshot 2021-09-19 at 01 28 33

Might be worth adding an empty state message and a test for the message in that state? Otherwise, might look a bit confusing / as if it's broken even though it's working

darrenvong (changes requested):
Some final things I found, should be good after this!

darrenvong (approved):

PR #2000, opened 2021-09-21
Web meeting notes 2021/09/21

lucaslcode:


docs/meetings/web/2021-09-21.md
```diff
@@ -0,0 +1,36 @@
+# Web frontend meeting 21 September 2021
+
+**Present**: Aapeli, Jesse, Lucas, Natalia, Ricardo
+
+## Discussion
+
+- weblate and i18next is waiting for workspaces
```
darrenvong, line 7: Did we actually get a conclusion from the investigation/research since the issue is still open?

docs/meetings/web/2021-09-21.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Do we need lerna?
+  - Lerna is good for versioning and executing commands across packages
+  - Don't need it, at least to start with
+  - yarn workspaces should work but we need to transpile the code
```
darrenvong, line 12: Would tweaking the `generate_protos` script to compile stuff into the `core` workspace not work?

darrenvong (approved):

docs/meetings/web/2021-09-21.md
lucaslcode: I didn't implement anything, but a little research and recommendations from a few people indicated these are most likely the best options

docs/meetings/web/2021-09-21.md
lucaslcode: No - what if the code in `core` is changed but not the proto files?
But we can add a `yarn workspace core build` to `prerun`/`prebuild` etc (or maybe there an even more elegant way)

darrenvong: Isn't that what we're doing in CI at the moment anyway? The `proto` step is always run at the beginning of each pipeline run so I don't see how it's different, unless there's some quirks with workspace I'm overlooking here.

lucaslcode: True, I was thinking more just local development though. 
Also the proto script is a bash script and transpiling is node, so it seems more correct to me to put it with other node stuff. Maybe that's silly, though.

PR #2006, opened 2021-09-24
HTML Page Titles

### LINKED ISSUE #1762: All page titles are the same
We should use React Helmet to change the title based on the page.

alesdvorakcz:
closes #1762

Added a component HtmlMeta which can be used to set html page title, social share description meta, etc.
The component is used on majority existing pages.

**Web frontend checklist**
- [x] All tests pass


app/web/src/components/HtmlMeta.tsx
```diff
@@ -0,0 +1,56 @@
...
+  title?: string;
+  sharingTitle?: string;
+  description?: string;
+  noSuffix?: boolean;
+}
+
+export default function HtmlMeta(props: HtmlMetaProps) {
+  const title = props.title || HTML_META_TITLE;
+  const sharingTitle = props.sharingTitle || title;
+  const description = props.description || HTML_META_DESCRIPTION;
```
lucaslcode: Just stylistic thing, but in the rest of the repo we destructure the props in the function definition and use defaults, like this:
`export default function HtmlMeta({firstProp = DEFAULT_VALUE, secondProp = X}: HtmlMetaProps) {`

app/web/package.json
```diff
@@ -22,6 +22,7 @@
     "react-dom": "^17.0.2",
     "react-error-boundary": "^3.1.3",
     "react-gtm-module": "^2.0.11",
+    "react-helmet": "^6.1.0",
```
lucaslcode: This is no longer maintained. How about [`react-helmet-async`](https://github.com/staylor/react-helmet-async)? It's a few extra lines but probably a safer option (until the NYT decides to stop maintaining it 😓)

lucaslcode (changes requested):
So much better, this one was really annoying me

app/web/src/App.tsx
```diff
@@ -19,6 +20,7 @@ function App() {
             <AuthProvider>
               <CssBaseline />
               <EnvironmentBanner />
+              <HtmlMeta noSuffix />
```
lucaslcode: Does this definitely get overwritten by later ones? (I mean, it does seem to work)

app/web/src/features/communities/NewGuidePage.tsx
```diff
@@ -6,6 +7,7 @@ import NewGuideForm from "./NewGuideForm";
 export default function NewGuidePage() {
   return (
     <>
+      <HtmlMeta title={"Create a new guide"} />
```
lucaslcode, line 10: Just for the record, this page is basically a placeholder which is why it isn't using constants (yet)

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -174,8 +175,15 @@ export default function GroupChatView() {
 
   const handleBack = () => history.goBack();
 
+  const title = groupChat
+    ? groupChatTitleText(groupChat, groupChatMembersQuery, currentUserId)
+    : groupChatError
+    ? "Error"
```
lucaslcode: Can this be a constant? I think reusing "Chat" or "Messages" is fine - they'll see the error on the page

lucaslcode (changes requested):

app/web/src/App.tsx
alesdvorakcz: This one will be used as default for pages without more specific override.

For example:
- user goes to Login page which will set title to 'Login'
- you go to some other page which does not have its own HtmlMeta
- if there is no 'root' react helmet component, the title would not re render and stays 'Login'

app/web/src/features/messages/groupchats/GroupChatView.tsx
alesdvorakcz: A constant? Do you mean to set a title to "Chat" or "Messages"? Now there will be name of user you are chatting with.

lucaslcode: Sorry - I meant specifically for the error case

alesdvorakcz: okay I simplified it. Hope I get it right

app/web/package.json
alesdvorakcz: migrated to `react-helmet-async`

alesdvorakcz:
Is it possible to re run CI pipeline? It somehow made two runs simultaneously, one timeouted, second one passed
https://gitlab.com/couchers/couchers/-/pipelines/376925747

app/web/package.json
```diff
@@ -115,6 +116,7 @@
     "@types/react": "^17.0.22",
     "@types/react-dom": "^17.0.9",
     "@types/react-gtm-module": "^2.0.1",
+    "@types/react-helmet": "^6.1.2",
```
darrenvong: We can uninstall this package as `react-helmet-async` comes with its own types, and also this wouldn't match up with it anyway since the package name is different.

app/web/src/features/auth/jail/Jail.tsx
```diff
@@ -52,6 +53,7 @@ export default function Jail() {
   return (
     <>
       {!isJailed && <Redirect to="/" />}
+      <HtmlMeta title="More information required" />
       <PageTitle>More information required</PageTitle>
```
darrenvong: Can we move these string values into constants as well please?

app/web/src/features/communities/discussions/DiscussionPage.tsx
```diff
@@ -85,6 +86,7 @@ export default function DiscussionPage() {
 
   return (
     <>
+      <HtmlMeta title={discussion?.title} />
```
darrenvong, line 89: Looking at how `HtmlMeta` works, it's possible that the title will read "Couchers.org Beta | Couchers.org Beta" while `discussion.title` is undefined (i.e. the query is still loading) which I'm not sure we want that. We might want to set the `noSuffix` depending on the loading state instead to avoid this.

app/web/src/features/communities/events/EditEventPage.tsx
```diff
@@ -112,23 +113,26 @@ export default function EditEventPage() {
...
-        )}
-      </EventForm>
+      <>
+        <HtmlMeta title="Edit event" />
+        <EventForm
+          error={error}
+          event={event}
+          isMutationLoading={isLoading}
+          mutate={updateEvent}
+          title="Edit event"
```
darrenvong: Move "Edit event" to a constant please (my bad not doing it in the first place...)

app/web/src/features/communities/events/EventPage.tsx
```diff
@@ -196,6 +197,7 @@ export default function EventPage() {
     <NotFoundPage />
   ) : (
     <>
+      <HtmlMeta title={event?.title} />
```
darrenvong, line 200: Same comment as in `DiscussionPage` - set `noSuffix` depending on the loading state so we don't get "Couchers.org Beta" twice.

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -186,17 +192,7 @@ export default function GroupChatView() {
...
             <PageTitle className={classes.title}>
-              {groupChat ? (
-                groupChatTitleText(
-                  groupChat,
-                  groupChatMembersQuery,
-                  currentUserId
-                )
-              ) : groupChatError ? (
-                "Error"
-              ) : (
```
darrenvong, line 197: We should probably still show an error alert when `groupChatError` is not null (with whatever in `groupChatError.message` rather than this state being completely ignored/removed in the new code.

app/web/src/features/profile/view/UserPage.tsx
```diff
@@ -35,6 +36,7 @@ export default function UserPage() {
 
   return (
     <>
+      <HtmlMeta title={user?.name} />
```
darrenvong, line 39: Set `noSuffix` based on query loading state here

darrenvong (changes requested):
Mostly little comments, but this is a nice improvement towards better SEO 🎉

app/web/package.json
alesdvorakcz: removed

app/web/src/features/auth/jail/Jail.tsx
alesdvorakcz: done

app/web/src/features/communities/discussions/DiscussionPage.tsx
alesdvorakcz: I changed the behaviour. Now when title has default value, it will automatically omit the suffix. This should fix the problem with loading state with no extra code this component is used

app/web/src/features/communities/events/EditEventPage.tsx
alesdvorakcz: done

app/web/src/features/communities/events/EventPage.tsx
alesdvorakcz: fixed by updating the HtmlMeta component

app/web/src/features/profile/view/UserPage.tsx
alesdvorakcz: fixed by updating the HtmlMeta component

app/web/src/features/messages/groupchats/GroupChatView.tsx
alesdvorakcz: I added generic error message to alert when groupCharError is not null but groupChatError.message is not provided. The error message should not be in the page title imho

darrenvong: Ahh I didn't realise `groupChatError.message` was already being displayed in the error alert as mentioned. Good to have a default message fallback too rather than the empty string before 👏

darrenvong (commented):
Also looks good pending CI builds - I've kicked it off to re-run again as I suspect it might be a GitLab issue rather than the code

alesdvorakcz:
> Also looks good pending CI builds - I've kicked it off to re-run again as I suspect it might be a GitLab issue rather than the code

Yeah CI build is timing out. I cannot even repeat it to give it another try

darrenvong (approved):

lucaslcode (approved):

Issue #1611, opened 2021-07-11
Add a "Save" button and text to the top of the profile editing page


jesseallhands:
Users are overlooking the save button at the bottom of the profile editing page. This is causing confusion and frustration (people think it auto-saves but doesn't and they lose all the text they just typed up because they didn't click save!) The same happens with the profile photo.

**Solution:**
Keep the Save button at the bottom of the page, but also add a "Save" button at the top of the page next to the header "Edit Profile". To the right of the button (on desktop) or below the button (on mobile) there should also be a warning text:
"Click save after making any changes or uploading a photo!"

![image](https://user-images.githubusercontent.com/79384268/125209231-11004b80-e25d-11eb-8a99-72432e2938b5.png)
![Screenshot_20210711-154729_Firefox](https://user-images.githubusercontent.com/79384268/125209661-f8456500-e25f-11eb-9a89-57ece951692b.jpg)

kerrylaval:
@lucaslcode @Couchers-org/frontend this is a bit more complicated than copy and paste! If we really want the save button to be next to "Edit Profile" heading then there needs to be some rearranging done. `<EditProfilePage />` is currently rendering `<EditProfile />` where the save button and all the form functionality lives. If we want to put the save button in `<EditProfilePage />` we'll need to lift the state up. Unless it would just be easier to put the button as far to the top of the form as possible. What do you think?

darrenvong:
How about if we display an "all your progress will be lost" dialog instead before the user switches page? That seems like another common and well known pattern other sites use for warning the users they have unsaved changes before they go off the site/another page of the site.

I'm not a fan of having to lift state up for this - considering there is also another form for the page, it's a technical nightmare to have to switch between the save action depending on which form is active and I think it adds unnecessary technical complexity if the dialog solution will suffice.

jesseallhands:
>How about if we display an "all your progress will be lost" dialog instead before the user switches page?

I think this could work, but because of how long some of these profiles are, it's not a good UX to have the person scroll all the way to the bottom of a super long page just to click "save". Another option would be a floating "save" button, but I'm not sure how much trouble it would be to implement that. No matter which solution we go with, we should have the "Any changes would be lost" warning.

kerrylaval:
@AllhandsNoFeet not sure what you mean by a floating save button but I can start by adding the warning

jesseallhands:
@kerryja I mean something like what we have in the bottom right of the dev-preview website:

[...]

PR #2007, opened 2021-09-24
floating save button

alesdvorakcz:
part of #1611 

Save button on edit profile page sticks at the bottom of screen. It should be good enough until the edit page form will be reworked.

Desktop
image

Tablet
image

Phone
image

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/features/profile/edit/EditProfile.tsx
+++ b/app/web/src/features/profile/edit/EditProfile.tsx
@@ -80,6 +80,7 @@ const useStyles = makeStyles((theme) => ({
     left: 0,
     width: "100%",
     display: "flex",
+    zIndex: 105,
     backgroundColor: theme.palette.background.paper,
     borderTop: `1px solid ${theme.palette.divider}`,
     justifyContent: "center",
@@ -94,7 +95,7 @@ const useStyles = makeStyles((theme) => ({
     [theme.breakpoints.up("md")]: {
       margin: theme.spacing(0, 10),
     },
-    paddingBottom: theme.spacing(50)
+    paddingBottom: theme.spacing(50),
   },
   // .field is the free text fields
   field: {
--- a/app/web/src/features/profile/edit/EditProfile.tsx
+++ b/app/web/src/features/profile/edit/EditProfile.tsx
@@ -75,7 +75,13 @@ const useStyles = makeStyles((theme) => ({
     },
   },
   buttonContainer: {
+    position: "fixed",
+    bottom: 0,
+    left: 0,
+    width: "100%",
     display: "flex",
+    backgroundColor: theme.palette.background.paper,
+    borderTop: `1px solid ${theme.palette.divider}`,
     justifyContent: "center",
     paddingBottom: theme.spacing(1),
     paddingTop: theme.spacing(1),
@@ -88,6 +94,7 @@ const useStyles = makeStyles((theme) => ({
     [theme.breakpoints.up("md")]: {
       margin: theme.spacing(0, 10),
     },
+    paddingBottom: theme.spacing(50)
   },
   // .field is the free text fields
   field: {
--- a/app/web/src/features/profile/edit/EditProfile.tsx
+++ b/app/web/src/features/profile/edit/EditProfile.tsx
@@ -50,68 +50,12 @@ import { Controller, useForm } from "react-hook-form";
 import { useQueryClient } from "react-query";
 import { service, UpdateUserProfileData } from "service/index";
 import { useIsMounted, useSafeState } from "utils/hooks";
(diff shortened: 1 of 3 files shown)


app/web/src/features/profile/edit/EditProfile.tsx
```diff
@@ -88,6 +95,7 @@ const useStyles = makeStyles((theme) => ({
     [theme.breakpoints.up("md")]: {
       margin: theme.spacing(0, 10),
     },
+    paddingBottom: theme.spacing(50),
```
lucaslcode: I think 5 is enough :D

lucaslcode (changes requested):
Great! Can you also do this for the Home tab? (The style may need to be moved to the parent or something to avoid copy-pasting)

alesdvorakcz:
I am trying to find how to reuse styles in the project and I am without luck. Is there a preferred way how to solve it? To create a styles.ts and where it between features/profile/edit components?

alesdvorakcz:
Ok, I tried, I hope I am not inventing wheel again....

app/web/src/features/profile/edit/styles.ts
```diff
@@ -0,0 +1,81 @@
...
+  },
+  // Everything under the mapbox
+  bottomFormContainer: {
+    display: "flex",
+    flexDirection: "column",
+    justifyContent: "center",
+    [theme.breakpoints.up("md")]: {
+      margin: theme.spacing(0, 10),
+    },
+    paddingBottom: theme.spacing(5),
```
lucaslcode, line 42: Sorry I misled you, this needs to be a tiny bit more because it blocks slightly on the home tab.

app/web/src/features/profile/edit/styles.ts
```diff
@@ -0,0 +1,81 @@
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
```
lucaslcode, line 3: A lot of these classes are not shared. There's an argument for putting them here anyway, though (rather than being strict and only having shared classes here). I'm not really convinced either way, it hasn't come up that many times before. Maybe it's fine like this.

lucaslcode (changes requested):
Minor thing

app/web/src/features/profile/edit/styles.ts
```diff
@@ -0,0 +1,81 @@
...
+      width: "100%",
+    },
+  },
+  buttonContainer: {
+    position: "fixed",
+    bottom: 0,
+    left: 0,
+    width: "100%",
+    display: "flex",
+    zIndex: 105,
```
darrenvong, line 26: Can this be done without arbitrary `z-index` values? I'm a bit wary of the [z-index inflation problem](https://www.joshwcomeau.com/css/stacking-contexts/#debugging-stacking-context-issues) briefly referenced here (the blog post is very good at talking about how to deal with `z-index` in general!) and this leading to more hacky CSS for us in the future

lucaslcode: I'm not sure it's possible because `tui-editor` sets a z-index which necessitates more of them....

app/web/src/features/profile/edit/styles.ts
alesdvorakcz: I was thinking the same but I decided to not over-engineer in this specific case. There are not many styles and these two edit forms are isolated enough so it should not grow much...

app/web/src/features/profile/edit/styles.ts
alesdvorakcz: Exactly the tui-editor was the problem why I needed to use the z-index in this case. I don't think there is an easy fix for this

app/web/src/features/profile/edit/styles.ts
alesdvorakcz: The home tab needed bottom padding at different place. Now it should be okay

app/web/src/features/profile/edit/styles.ts
darrenvong: Ahh yeah that's annoying :/ let's leave it then

darrenvong (approved):
LGTM!

alesdvorakcz:
@lucaslcode Is there anything else I should change or is it ready to merge?

lucaslcode:
Yes please merge! One approval is fine if there's nothing that hasn't been addressed :)

alesdvorakcz:
But it says Merging is blocked
image

lucaslcode (approved):

lucaslcode:
Done - you can also just dismiss it next time if it's stale

PR #2026, opened 2021-09-28
Show events from user communities on dashboard

### LINKED ISSUE #1880: Add events from my communities to the dashboard.
Include a component on the dashboard that shows all events from communities the user is a member of, in chronological order. Placed underneath the communities section

darrenvong:
Probably my first time getting my hands dirty on the backend a little bit on this project, so hopefully I haven't done anything too terrible :p it was too inefficient and annoying me to have to send one request per community just to find out which events are in the communities the user is in, so I extended the `ListMyEvents` rpc to include events from the user's communities by default as well.

**Desktop**
Screenshot 2021-09-28 at 02 21 33

**Mobile**
Screenshot 2021-09-28 at 02 22 07

Closes #1880

**Backend checklist**
- [ ] All tests pass - I'm not sure why the migration tests are failing for me locally in Docker, but it's fine in GitLab??

**Web frontend checklist**
- [x] All tests pass


app/backend/src/couchers/servicers/events.py
```diff
@@ -734,6 +736,21 @@ def ListMyEvents(self, request, context):
...
+            if include_my_communities:
+                my_communities = (
+                    session.execute(
+                        select(Node.id)
+                        .join(Cluster, Cluster.parent_node_id == Node.id)
+                        .join(ClusterSubscription, ClusterSubscription.cluster_id == Cluster.id)
+                        .where(ClusterSubscription.user_id == context.user_id)
+                        .where(Cluster.is_official_cluster)
+                        .order_by(Node.id)
+                        .limit(100000)
```
darrenvong, line 748: I kinda copied this from `threads.py` where 100000 is an arbitrarily "large enough" (?) number/page size cap so it should be sufficient to include all communities the user is in one page, but not unbounded

kalvdans: Whenever we convert this into a subquery, the list of clusters will not be extracted from the database and this arbitrary limit will go away. So it is fine for now :)

app/backend/src/tests/test_events.py
```diff
@@ -1,15 +1,14 @@
...
 import grpc
 import pytest
 from google.protobuf import wrappers_pb2
 from psycopg2.extras import DateTimeTZRange
+from sqlalchemy.sql.expression import update
 
 from couchers import errors
 from couchers.db import session_scope
+from couchers.models import EventOccurrence
 from couchers.utils import Timestamp_from_datetime, now, to_aware_datetime
```
kalvdans, line 12: Are these header rearrangements caused by a new version of `isort`? We should be doing them all at once in a separate single PR then.

app/proto/events.proto
```diff
@@ -255,6 +255,8 @@ message ListMyEventsReq {
   bool subscribed = 2;
   bool attending = 3;
   bool organizing = 4;
+  // include events in the user's communities
```
kalvdans, line 258: Great with comments!

app/backend/src/tests/test_events.py
```diff
@@ -1741,9 +1744,12 @@ def new_event(hours_from_now, online=True):
         res = api.ListMyEvents(events_pb2.ListMyEventsReq(organizing=True))
         assert [event.event_id for event in res.events] == [e3, e4]
 
+        res = api.ListMyEvents(events_pb2.ListMyEventsReq(my_communities=True))
+        assert [event.event_id for event in res.events] == [e1, e2, e3, e4, e5]
```
kalvdans, line 1748: Great with tests!

kalvdans (dismissed):
Backend looks good! There's still some failing frontend CI tests though.

app/backend/src/tests/test_events.py
darrenvong: Since it's not pinned to a version in `requirements.txt`, I guess it's possible that I installed a newer version than the one you/Aapeli have on your machine as it's been the first time I really touched the backend. Maybe we should pin that to a version too? 🤷

darrenvong:
> There's still some failing frontend CI tests though.

It's because of this - https://gitlab.com/gitlab-com/gl-infra/production/-/issues/5590 😠  apparently the adjacent `web-next` job built and the tests from that passed though so I think we might be ok overall!

app/web/src/features/dashboard/MyEvents.tsx
```diff
@@ -0,0 +1,86 @@
...
+                  key={event.eventId}
+                  event={event}
+                  className={classes.placeEventCard}
+                />
+              ))}
+          </HorizontalScroller>
+          {hasNextPage && (
+            <StyledLink
+              className={classes.allUpcomingEventsLink}
+              to={eventsRoute}
```
lucaslcode: This doesn't quite make sense to me - if there's more than one page of events in communities the user has joined, show a link to all upcoming events?

I know it's in the design but since there is no "My Events" page at this stage, I think it's better to make it an ordinary load more button (perhaps inside the `HorizontalScroller`).

lucaslcode (changes requested):
One thing

app/web/src/features/dashboard/MyEvents.tsx
darrenvong: Do you mean have the "load more" button but the results will be horizontally scrollable on all screen sizes rather than going onto the next row below forever?

lucaslcode: From memory `HorizontalScroller` does nothing above mobile screen sizes, so load more would add more horizontally on mobile and below otherwise.

But any way that works really - the main thing is not linking to the all events page with a label that makes it seem like its a "your events" page.

app/backend/src/tests/test_events.py
kalvdans: It might be a PR that was merged which we did not run `isort` on. I've opened https://github.com/Couchers-org/couchers/pull/2030 to add a check to CI.

Fine by me to do this change in this PR.

app/web/src/features/dashboard/MyEvents.tsx
```diff
@@ -70,12 +69,9 @@ export default function MyEvents() {
...
           {hasNextPage && (
-            <StyledLink
-              className={classes.allUpcomingEventsLink}
-              to={eventsRoute}
-            >
-              {SHOW_ALL_UPCOMING_EVENTS}
-            </StyledLink>
+            <Button onClick={() => fetchNextPage()}>
+              {SEE_MORE_EVENTS_LABEL}
+            </Button>
```
lucaslcode, line 86: This is a bit weird on mobile (clicking the button makes more appear in the horizontal scroller).
How do you feel about an intersection observer div at the end of the horizontal scroller (only visible at mobile size) which triggers the load more (and displays a spinner?) Like in `InfiniteMessageLoader`.

lucaslcode (changes requested):

app/web/src/features/dashboard/MyEvents.test.tsx
```diff
@@ -89,4 +95,46 @@ describe("My events", () => {
...
+      render(<MyEvents />, { wrapper });
+      expect(await screen.findAllByRole("link")).toHaveLength(2);
+      expect(
+        screen.queryByRole("button", {
+          name: SEE_MORE_EVENTS_LABEL,
+        })
+      ).not.toBeInTheDocument();
+
+      // Simulates scrolling horizontally to the end
+      mockIsIntersecting(screen.getByRole("progressbar"), true);
```
lucaslcode, line 128: nice

app/web/src/features/dashboard/MyEvents.tsx
```diff
@@ -57,18 +64,23 @@ export default function MyEvents() {
               classes.cardContainer,
               classes.upcomingEventContainer
             )}
+            fetchNext={isBelowSm ? fetchNextPage : undefined}
+            hasMore={hasNextPage}
+            isFetching={isFetching}
```
lucaslcode, line 69: Wow, I totally forgot I put this functionality in :D

lucaslcode (approved):

app/web/src/features/dashboard/MyEvents.tsx
darrenvong: Indeed, made my job very easy 😉

PR #2034, opened 2021-09-30
Client build improvements

aapeliv:
Due to recent changes in the Let's Encrypt / ISRG certificate chain, the client was failing to validate the certs of the server.

This hardcodes the ISRG roots for now which fixes it for the time being.

**Backend checklist**
- [ ] All tests pass


kalvdans:
We had the same problem at work with Ubuntu 20.04. But updating `ca-certificates` package fixed the issue.

app/client/src/couchers/__init__.py
```diff
@@ -1 +1,2 @@
-from .client import get_client  # noqa
+from .client import get_client
+from .client import version as __version__  # noqa
```
kalvdans, line 2: Where is `__version__` read?

kalvdans (commented):
Remove `app/client/src/couchers/certs/__init__.py` as this folder is just for data.

aapeliv:
It’s very useful for debugging installation issues with the support team (to make sure they’ve the right version installed).

kalvdans:
> It’s very useful for debugging installation issues with the support team (to make sure they’ve the right version installed).

So this PR is just for debugging and not meant to be merged?

aapeliv:
Upstream issue: https://github.com/grpc/grpc/issues/27532

As discussed in the meeting, we'll wait for upstream to fix up their root store, then we'll merge this excluding the addition of extra roots (there's a few other useful things here).

aapeliv:
@kalvdans: do you mind having another look? I've cleaned this up and removed the cert CA pinning, so now this just cleans up the build a bit.

app/client/src/couchers/__init__.py
aapeliv: This allows you to do something like this:

```py3
import couchers
print(couchers.__version__)
```

It's useful so we can ask people using the package to tell us what version they're using. Sometimes people mess up their virtual environments and are actually in a different one, so it's good the package tells you the exact one it is.

kalvdans: Just use `couchers.client.version` then, having it in thunder makes me think it is used by Python internally.

aapeliv: It's pretty common to call it `__version__`, and it seems there's a PEP for it: [PEP 0396](https://www.python.org/dev/peps/pep-0396/)

kalvdans (approved):
LGTM with or without my comment about `__version__`

kalvdans:
> This pull request is being automatically deployed with Vercel ([learn more](https://vercel.link/github-learn-more)).

Dead link, didn't learn more.

PR #2035, opened 2021-09-30
Make location filter required for other filters

### LINKED ISSUE #1676: Search pagination causes users to disappear from map
Subject: Hosting status filters removes everyone
Description:
- Was looking at an area with a handful of "may host" people (Oslo, Norway).
- Turned on May host and Can host filters
- Everyone on the map disappeared
- Removed the filter, everyone reappeared
- Turned on just May host filter, everyone disappeared again
- Clicked the X to turn off filter, no one reappeared, the map is still empty on the zoom level I'm looking at

Results:
Everyone disappears from the area I'm looking at. Only those that do not match the filter should disappear.

Backend version: develop-fd12be58
Frontend version: develop-fd12be58
User Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:90.0) Gecko/20100101 Firefox/90.0
Page: https://app.couchers.org/search#loc=13.67/59.9119/10.78319
User: forteller (2611)

### LINKED ISSUE #1776: Only include cities in location search
Add an option in the geocode hook to only include cities/towns in results. 

This is because our location search currently does radius around a point, so using a country result just gives you an X km circle in the middle of the country.

### LINKED ISSUE #1971: Not all ppl who can host show when the filter is on
Subject: Not all ppl who can host show when the filter is on
Description:
.

Results:
.

Backend version: develop-c46dacca
Frontend version: develop-c46dacca
User Agent: Mozilla/5.0 (Linux; Android 11; SM-G781V) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.181 Mobile Safari/537.36 (Ecosia android@88.0.4324.181)
Page: https://app.couchers.org/search?hostingStatusOptions=2#loc=5.43/41.287/-74.138/16.7
User: self_directed_coucher (4251)

### LINKED ISSUE #2121: User location missing from search result cards
Multiple users have reported to the support team that they have received hosting requests from people who believe they are located in a different country than they are located in. This is because the current search results do not list the actual location of the  person (eg. city, country) and the search is based upon keyword (searching for "Rome" will show any user with "Rome" in their profile, eg. a user named "Romeo" or a person located in "Rome Corners, Wisconsin".)

To improve the user experience and avoid confusion we should include the user's location. This would save screen real-estate and would eliminate the need for a "show on map" button because when the user clicks on the location link, it would take them to the pin of the user on the map. Once the search functionality is improved to be based upon location, it will also help in situations where someone doesn't actually live in a city center but nearby (eg. someone searches "Chicago" and someone from "Naperville" shows up in the search results because they are nearby, but in reality Naperville is nearly an hour from downtown Chicago.)

**TLDR: Adding location to the search results cards would eliminate confusion which is currently happening, improve search UX and save screen real-estate.**

See mockup of what it would look like below:

image

### LINKED ISSUE #2158: Selecting logo on top nav bar should direct user to dashboard
Current logo on the nav bar isn't interactive. Saw a user on the community forums express this complaint.

lucaslcode:
Fix #1676 fix #1971 
I say fix but really it's a stop-gap solution until we do better search in the backend
This PR makes location required for filters. That means there won't be a pagination issue which happens when you apply a filter to the whole world.

I removed the text input on the results list because it was confusing for users (they would type locations in there) and to make the code more maintainable (don't need the URLSearchParams to be some weird kind of global state). But it looks pretty bad now, needs design input.

**Web frontend checklist**
- [ ] All tests pass


aapeliv:
I think the idea is OK for now :+1:

lucaslcode:
#2121

lucaslcode:
This is just waiting on tests now, but this can be reviewed @darrenvong 
I'm not super happy with the architecture of this, it seems pretty messy. Open to tips/suggestions.

app/web/src/components/UserSummary.tsx
```diff
@@ -136,11 +133,6 @@ export default function UserSummary({
                 {!user ? <Skeleton /> : user.city}
               </Typography>
             )}
-            {process.env.REACT_APP_IS_VERIFICATION_ENABLED && (
```
lucaslcode, line 139: This is just annoying since dev ends up looking different from production.

app/web/src/features/search/FilterDialog.tsx
```diff
@@ -77,88 +114,131 @@ export default function FilterDialog({
...
+      lastActive: lastActiveOptions.find(
+        (o) => o.value === searchFilters.active.lastActive
+      ),
+      hostingStatusOptions: searchFilters.active.hostingStatusOptions,
+      numGuests: searchFilters.active.numGuests,
+    }),
+    //we specifically want abnormal behaviour of not updating default values
+    //eslint-disable-next-line react-hooks/exhaustive-deps
+    [isOpen]
+  );
```
darrenvong: Been a little while so just to refresh my memory... is this to stop react-hook-form from complaining about changing from uncontrolled to controlled component or something like that?

Also, why switch from a ref to `useMemo`?

app/web/src/features/search/LocationAutocomplete/LocationAutocomplete.tsx
```diff
@@ -90,24 +96,32 @@ export default function LocationAutocomplete({
...
+            <Typography variant="body2">
+              {geocodeResult2String(option, showFullDisplayName)}
+              <br />
+              {MUST_BE_MORE_SPECIFIC}
+            </Typography>
+          </>
+        ) : (
+          geocodeResult2String(option, showFullDisplayName)
+        );
       }}
```
darrenvong: Is there a good reason to still show the result if selecting regions are disabled? Can we hide it by not rendering the result instead?

app/web/src/features/search/FilterDialog.tsx
```diff
@@ -66,9 +71,41 @@ export default function FilterDialog({
...
 }) {
   const classes = useStyles();
-  const { control, handleSubmit } = useForm({ mode: "onBlur" });
+  const { control, handleSubmit, register, setValue, getValues, errors } =
+    useForm<FilterDialogFormData>({
+      mode: "onBlur",
+    });
   const queryClient = useQueryClient();
-  const onSubmit = handleSubmit(() => {
+  const onSubmit = handleSubmit((data) => {
```
darrenvong, line 80: I'm guessing this is a refactor as we don't really need to change the filters on every change until the "apply" button is clicked on the modal?

app/web/src/features/search/SearchBox.tsx
```diff
@@ -20,69 +56,127 @@ export default function SearchBox({
...
-                  size="small"
-                >
-                  <FilterIcon />
-                </IconButton>
                 <IconButton
                   aria-label={CLEAR_SEARCH}
                   onClick={() => {
                     setValue("query", "");
-                    searchFilters.clear();
-                    onSubmit();
```
darrenvong, line 85: I think this might still be needed? The button is not clearing the keyword input search when this is being clicked at the moment

app/web/src/features/search/useSearchFilters.ts
```diff
@@ -44,7 +44,15 @@ export default function useSearchFilters(route: string) {
...
   }, []);
 
-  return { active, change, remove, apply, clear };
+  const any = !!(
+    active.hostingStatusOptions?.length ||
+    active.lastActive ||
+    active.location ||
+    active.numGuests ||
+    active.query
+  );
```
darrenvong, line 53: I'm not sure I completely follow what the purpose of this `any` value is doing - is it to control whether to show everyone on the map or only those specific to the filters?

app/web/src/features/search/SearchResultsList.tsx
```diff
@@ -14,14 +14,14 @@ import maplibregl, { LngLatBounds, Map as MaplibreMap } from "maplibre-gl";
...
-import React, { MutableRefObject, useRef } from "react";
+import { MutableRefObject, useRef } from "react";
 import { useInfiniteQuery } from "react-query";
 import { service } from "service";
 import hasAtLeastOnePage from "utils/hasAtLeastOnePage";
 
 const useStyles = makeStyles((theme) => ({
   mapResults: {
-    height: "15rem",
+    height: "17rem",
```
darrenvong, line 24: This seems a bit random. Why does it need more height?

app/web/src/features/search/SearchResult.tsx
```diff
@@ -129,6 +129,8 @@ export default function SearchResult({
               </Typography>
             </div>
           </div>
+          <div></div>
+          <Typography noWrap>{user.city}</Typography>
```
darrenvong: Could add some top padding instead of an empty div?

darrenvong (commented):
Mostly questions rather than any concrete suggestions to change anything major.

I don't really have any good alternatives in mind at the moment, but I agree it feels a bit messy. Most of the complication is probably to do with the fact we're using the modal and having to share state between that and the underlying page? Might be hard to simplify this too much without re-thinking what this bit of the UI should look/work.

darrenvong:
Also, the location search thing still seems a bit buggy - if I use the location search straight after I land on the map search page and search for "New York", select the eligible result, I'm still getting everyone back on the map and in the search result (see screenshot below, where Itsi is not in New York but he still shows up!):
Screenshot 2021-10-24 at 20 03 22

Somehow, if I open the filter dialog and click "apply", it works correctly. But the keyword search don't suffer this same problem 🤔 

Getting this warning in the console when I ran locally:
```
location is missing in the 'defaultValue' prop of either its Controller (https://react-hook-form.com/api#Controller) or useForm (https://react-hook-form.com/api#useForm) 
    at LocationAutocomplete (http://localhost:3000/static/js/20.chunk.js:42:3)
...
```
Wonder if that's related? 🤷

app/web/src/features/search/FilterDialog.tsx
lucaslcode: I wanted it to update automatically every time the dialog is opened, but not when the search filters change.

lucaslcode: Combined with the fact the dialog is unmounted when closed, yes, this makes the default values from the `TextField`'s point of view not change.

app/web/src/features/search/LocationAutocomplete/LocationAutocomplete.tsx
lucaslcode: Yes, but if you do that, you can type `Australia` and not get any results that are in Australia. This way, it's obvious that you need to type a city

lucaslcode: Actually, I'm going to change this so it fails validation instead.

app/web/src/features/search/SearchResultsList.tsx
lucaslcode: Added location text, and it wasn't quite right before on mobile anyway

app/web/src/features/search/SearchResult.tsx
lucaslcode: Whoops

app/web/src/features/search/useSearchFilters.ts
lucaslcode: It's for components to use to know where there are any filters which currently apply. I found I was using a check similar to this in a few places

lucaslcode:
Took soooo long to fix the default value warnings - turned out to be the `setValue()` in the `useEffect` that makes sure changes in the filter dialog are applied. Ran out of time for the tests for tonight.

darrenvong:
Had a quick click around/test again and the filters seem to be all working now! Will try and have a look at the code again soon when I get more time. I don't think it looks too terrible but maybe clicking the cross in the input field clearing the filters in the dialog too might not be too obvious? 🤷  Let's see how people find this I guess...

> turned out to be the setValue() in the useEffect that makes sure changes in the filter dialog are applied.

Do you mean the one in `SearchBox`?

lucaslcode:
> Do you mean the one in `SearchBox`?

[This one](https://github.com/Couchers-org/couchers/pull/2035/files#diff-ad756c7d869593a696b74e7893681901aba970302226cd0981fa8063a9b9f9a5R133)

Tests are done! I was getting failures in events tests about touched fields - the api calls were not being called.  Let's see if ci passes...

darrenvong:
~~Hmm interesting that this seems to have broken the form when you're trying to edit the event time, the submitted result is now somehow an hour behind... but you've practically changed none of the code (apart from the location's default value so I'm very confused 🤔 )~~ Never mind, seems to work again, strange!

I had a play around and for some reason the form validation is taking longer to run (test performance issues again? 😢 ) so increasing the timeout on the `waitFor` wrapping the assertion that the API is called seems to have fixed it

app/web/src/features/search/FilterDialog.tsx
```diff
@@ -77,88 +116,136 @@ export default function FilterDialog({
...
+        lastActiveOptions.find(
+          (o) => o.value === searchFilters.active.lastActive
+        ) ?? null,
+      hostingStatusOptions: searchFilters.active.hostingStatusOptions,
+      numGuests: searchFilters.active.numGuests ?? "",
+    }),
+    //we specifically want abnormal behaviour of not updating default values
+    //eslint-disable-next-line react-hooks/exhaustive-deps
+    [isOpen]
+  );
```
darrenvong: After reading this about `undefined` being an invalid value for `defaultValue`](https://github.com/react-hook-form/react-hook-form/discussions/2797) (e.g. in `location` before you changed to `"" as const`), I think this `useMemo` (or previously `useRef`) might be unnecessary and was really a workaround because we didn't use the library correctly? I had a play around locally by removing the `useMemo` and I didn't see the warning anymore.

app/web/src/features/search/SearchBox.tsx
```diff
@@ -20,84 +56,182 @@ export default function SearchBox({
...
+            name: searchFilters.active.location,
+            simplifiedName: searchFilters.active.location,
+            location: new LngLat(
+              searchFilters.active.lng,
+              searchFilters.active.lat
+            ),
+          }
+        : ("" as const),
+    keywords: searchFilters.active.query ?? "",
+  }).current;
```
darrenvong: Same comment as the `useMemo` in `FilterDialog`

app/web/src/features/search/SearchBox.test.tsx
```diff
@@ -31,34 +33,54 @@ const View = ({
...
+    userEvent.click(await screen.findByText("test city, test country"));
     await waitFor(() => {
-      expect(setActive).toBeCalledWith({ query: "test search" });
+      expect(setActive).toBeCalledWith({
+        location: "test city, test country",
+        lng: 1.0,
+        lat: 2.0,
+      });
     });
+    server.close();
```
darrenvong: I think it's better to put this in an `afterEach` so it closes the server even if this test fails, so it doesn't affect other tests.

app/web/src/features/search/SearchBox.test.tsx
```diff
@@ -68,31 +90,36 @@ describe("SearchBox", () => {
...
       expect(dialog).not.toBeVisible();
     });
 
-    expect(input).toHaveValue("test search");
-    await waitFor(() => {
-      expect(setActive).toBeCalledWith({ query: "test search" });
-    });
+    expect(input).toHaveValue("new search");
   });
 });
```
darrenvong, line 154: Missing tests to check:
- clearing the location from the input clears filter
- clearing the keyword search from input clears filter

darrenvong (changes requested):
Should be the last thing (plus the other suggestions I have on fixing the events tests that seem to be strangely failing now...)!

app/web/src/features/search/FilterDialog.test.tsx
```diff
@@ -158,13 +168,14 @@ describe("FilterDialog", () => {
     expect(numGuestsInput).toHaveValue(3);
 
     userEvent.clear(locationInput);
-    userEvent.clear(lastActiveInput);
+    userEvent.click(lastActiveInput);
+    userEvent.click(screen.getByText("Any"));
```
darrenvong, line 172: Is this intentional? I noticed that clicking the X in the location input field doesn't clear the filters set in the modal dialog now unless we open to clear it there too.

lucaslcode: Yes - seemed to be the only way to get rid of warnings that "" was not in the list of options. "Any" has a value of null which clears the last active filter. 

Ah yes I have to fix the clearing though - otherwise you can bypass the "filters need location" check

darrenvong: Oh yeah, also there's nothing stopping the user from URL hacking the filter right? Does it send a search request to the backend straight away?

lucaslcode: Yes, but if they encounter weird behaviour that way I think that's on them? They can also set arbitrary location text for a lat/long (not that it has any effect other than cosmetic)

darrenvong (approved):

PR #2036, opened 2021-09-30
Fix cannot edit ongoing event

### LINKED ISSUE #1924: Can't edit event
Subject: Can't edit event
Description:
Tried to edit an event and change the time

Results:
It says my event can't have overlapping times... 

Backend version: develop-c14158d6
Frontend version: develop-c14158d6
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/92.0.4515.159 Safari/537.36
Page: https://app.couchers.org/event/242/istanbul-online-3/edit
User: emily (135)

### LINKED ISSUE #2005: Cannot edit event while event is going on
I tried editing my event while it was going on (I was updating information on where our table is located on the lakeside terrace) and it wouldn't let me update it. I got an error "The event start time must be in the future" and it highlighted the start time.

You should be able to update the event up until the event is over (or perhaps until the end of the day since sometimes events go longer and you may need to provide an update). This is especially true for multi-day events.

The event this happened to is here:

https://app.couchers.org/event/285/happy-hour-social

darrenvong:
Fixes #2005. 

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/features/communities/events/EditEventPage.tsx
+++ b/app/web/src/features/communities/events/EditEventPage.tsx
@@ -62,24 +62,24 @@ export default function EditEventPage() {
       updateEventInput = {
         eventId,
         isOnline: data.isOnline,
-        title: data.touched.title ? data.title : undefined,
-        content: data.touched.content ? data.content : undefined,
-        photoKey: data.touched.eventImage ? data.eventImage : undefined,
+        title: data.dirtyFields.title ? data.title : undefined,
+        content: data.dirtyFields.content ? data.content : undefined,
+        photoKey: data.dirtyFields.eventImage ? data.eventImage : undefined,
         startTime:
-          data.touched.startTime || data.isStartDateTouched
+          data.dirtyFields.startTime || data.dirtyFields.startDate
             ? finalStartDate
             : undefined,
         endTime:
-          data.touched.endTime || data.isEndDateTouched
+          data.dirtyFields.endTime || data.dirtyFields.endDate
             ? finalEndDate
             : undefined,
       };
 
       if (data.isOnline) {
         updateEventInput = Object.assign(updateEventInput, {
-          link: data.touched.link ? data.link : undefined,
+          link: data.dirtyFields.link ? data.link : undefined,
         });
-      } else if (data.touched.location) {
+      } else if (data.dirtyFields.location) {
         updateEventInput = Object.assign(updateEventInput, {
           address: data.location.name,
           lat: data.location.location.lat,
--- a/app/web/src/features/communities/events/EditEventPage.tsx
+++ b/app/web/src/features/communities/events/EditEventPage.tsx
@@ -15,7 +15,10 @@ import type { UpdateEventInput } from "service/events";
 import dayjs, { TIME_FORMAT } from "utils/dayjs";
 
 import { EDIT_EVENT } from "../constants";
-import EventForm, { CreateEventData, useEventFormStyles } from "./EventForm";
+import EventForm, {
+  CreateEventVariables,
+  useEventFormStyles,
+} from "./EventForm";
 import { useEvent } from "./hooks";
 
 export default function EditEventPage() {
@@ -38,7 +41,7 @@ export default function EditEventPage() {
   } = useMutation<
     Event.AsObject,
     GrpcError,
-    CreateEventData,
+    CreateEventVariables,
     { parentCommunityId?: number }
   >(
     (data) => {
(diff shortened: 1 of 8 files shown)


app/web/src/features/communities/events/EventForm.tsx
```diff
@@ -127,8 +138,14 @@ export default function EventForm({
     register,
     setValue,
     watch,
+    formState: { touched },
   } = useForm<CreateEventData>();
 
+  // This is a bit of a dirty workaround, but I have no ideas why react-hook-form doesn't pick up when
+  // the datepicker field has been touched...
+  const isStartDateTouched = useRef(false);
+  const isEndDateTouched = useRef(false);
```
darrenvong: This is the part I'm not happy about doing, because changing the date using the datepicker doesn't seem to get detected by react-hook-form as a "touched" field and so without this, updating the event time becomes impossible unless you touch the startTime/endTime fields as well, as the conditional logic checks either the date or time has been touched to decide whether to send that in the `UpdateEventReq` to the backend.

app/web/src/features/communities/events/EventTimeChanger.tsx
```diff
@@ -139,6 +149,10 @@ export default function EventTimeChanger({
               value: timePattern,
             },
             validate: (time) => {
+              if (!touched.startTime) {
+                return true;
+              }
```
darrenvong: I'll comment on this but it's the same across the date and time fields, but when this is used in the edit event context, without this, the react-hook-form validation would still prevent the user from updating the event even if they haven't touched this. Not sure if there's a cleaner way to disable the validation temporaily other than like this.

app/web/src/features/communities/events/EventForm.tsx
```diff
@@ -145,7 +155,10 @@ export default function EventForm({
 
   const onSubmit = handleSubmit(
     (data) => {
-      mutate(data);
+      mutate({
+        ...data,
+        dirtyFields,
+      });
```
lucaslcode, line 161: why not send the dirty fields only?

app/web/src/features/communities/events/EventForm.tsx
```diff
@@ -180,10 +193,7 @@ export default function EventForm({
           fullWidth
           helperText={errors.title?.message || ""}
           id="title"
-          inputRef={(element: HTMLInputElement | null) => {
-            element?.focus();
```
lucaslcode, line 184: I guess this was giving initial focus to the title, why was it removed?

lucaslcode (changes requested):
Small bug - I think the end date isn't marked as dirty when it is automatically updated via the start date. To reproduce:

1. Set the start date a week or two in the future
2. Move the end date 6 days later than that.
3. Save the event
4. Move the start date to tomorrow.
5. Try to save the event, get "Event can't be longer than 7 days" error even though end date shows 6 days.

app/web/src/features/communities/events/EventForm.tsx
darrenvong: It was a workaround for the editor bug where it was stealing focus. Also I found it was focusing on the field on submit sometimes and I had to hit submit again for things to go through properly... Not quite sure why though

app/web/src/features/communities/events/EventForm.tsx
darrenvong: Initially I was thinking that'd cause problems since the fields in that object is dynamic, but actually it'd just return `undefined` if trying to access the field and it's not there, so I guess no good reasons really, could do that too.

darrenvong: Wait actually that's not true, pretty much all the fields in `dirtyFields` are used to determine what to send to the backend, so it's just easier to pass the whole object through rather than field by field.

lucaslcode: I mean `mutate({ ...dirtyfields })` - then the mutate function doesn't need knowledge of what is and isn't dirty

darrenvong: That wouldn't work since `dirtyFields` don't give the actual value of the field, it's only an object of `true`s for fields that's dirty (e.g. if `title` is dirty, `dirtyFields` = `{ title: true }`.

darrenvong:
@lucaslcode want to take another look?

app/web/src/features/communities/events/EditEventPage.tsx
```diff
@@ -56,30 +59,32 @@ export default function EditEventPage() {
...
-          eventId,
-          isOnline: data.isOnline,
-          title: data.title,
-          content: data.content,
-          photoKey: data.eventImage,
-          startTime: finalStartDate,
-          endTime: finalEndDate,
+        updateEventInput = Object.assign(updateEventInput, {
+          link: data.dirtyFields.link ? data.link : undefined,
+        });
```
lucaslcode, line 81: Out of curiosity, is there a reason for using `Object.assign` insead of `updateEventInput.link = ...`?

lucaslcode (approved):
Looks good, sorry for forgetting this one.

I'd like to check photo upload is working as expected but preview is down. If you want, just merge it and check on next, since seems like it should be fine.

app/web/src/features/communities/events/EditEventPage.tsx
darrenvong: I had some trouble with that because of the discriminate union type I used, so TS was complaining about `link` not being assignable to `UpdateOfflineEventInput`. So it's either doing that, having to type assert or loosen the type. I guess option 1 (what I'm doing here) or 2 would be fine in this case over option 3 for better type safety/checking.

PR #2037, opened 2021-10-02
Web/bugfix/links too large

### LINKED ISSUE #1914: Links too large
Screen Shot 2021-09-05 at 5 42 47 pm

wbruntra:
closes Couchers-org/couchers#1914

Changed profile and account settings alert links to buttons, eliminated/changed text.

Some explanation of the option to set a password seems useful, so I maintained that text in the alert box.

**Web frontend checklist**
- [x ] Formatted my code with `yarn format && yarn lint --fix`
- [x] All tests pass


app/web/src/features/dashboard/DashboardBanners.tsx
```diff
@@ -1,12 +1,12 @@
-import { Typography } from "@material-ui/core";
+import { Button as MuiButton, Typography } from "@material-ui/core";
```
lucaslcode: Prefer our `Button` component from `"components/Button"`. It has default variant of contained and you should be able to set colour to default.

app/web/src/features/dashboard/DashboardBanners.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { Alert as MuiAlert } from "@material-ui/lab/";
 import Alert from "components/Alert";
-import StyledLink from "components/StyledLink";
 import { Error as GrpcError } from "grpc-web";
 import { GetAccountInfoRes } from "proto/account_pb";
 import { accountInfoQueryKey } from "queryKeys";
 import React from "react";
 import { useQuery } from "react-query";
+import { Link as RouterLink } from "react-router-dom";
```
lucaslcode: `Link` is fine - usually we import the MUI link as `MuiLink` but since we added `StyledLink` recently we shouldn't need to do that either

lucaslcode (changes requested):
Thanks so much! Just some very minor style things

wbruntra:
PR updated to reflect reviewer comments

app/web/src/features/dashboard/DashboardBanners.tsx
```diff
@@ -51,20 +52,19 @@ export default function DashboardBanners() {
                 {UPLOAD_PHOTO}
               </Typography>
               <Typography variant="inherit" paragraph>
-                <StyledLink to={routeToEditProfile()}>
-                  {CLICK_HERE_TO_EDIT}
-                </StyledLink>
+                <Button color="default" role="link">
```
goskan93: sorry for cut in on, but as I know putting a tag inside button tag is not valid html (some sources: https://www.w3schools.com/tags/tag_button.asp
https://stackoverflow.com/questions/51755195/is-it-a-valid-markup-to-put-some-tags-inside-button-tag )
maybe its better just to do onClick
```suggestion
                <Button color="default" role="link" onClick=(() => { history.push(routeToEditProfile()) })>
```

wbruntra: Good catch, thank you.

wbruntra:
I made the suggested change, but if I understand correctly, per @goskan93 's comment, the `Button` component has a problem because it renders its children inside a `<div>` by default, which is not valid.

lucaslcode:
You can use the `react-router-dom` link as the button component: `<Button component={Link} to={routeToX()}>...`

app/web/src/features/dashboard/constants.ts
```diff
@@ -4,14 +4,13 @@ export const FILL_IN_WHO_I_AM =
...
+export const CLICK_HERE_TO_EDIT = "Edit your profile";
 export const DONT_YOU_HATE =
   "Don't you hate how other platforms have empty \"ghost\" profiles with no information. Please don't ghost us! 👻👻👻";
 
-export const PASSWORD_TEXT_1 = "You can set a password on";
-export const PASSWORD_TEXT_LINK = "the account settings page";
-export const PASSWORD_TEXT_2 =
-  "so you don't need check your email each time you log in.";
+export const PASSWORD_TEXT_1 =
+  "Tip: You can set a password so you don't need check your email each time you log in";
```
lucaslcode:
```suggestion
  "You can set a password so you don't need check your email each time you log in";
```
Just because it's not really a tip at the moment, this never goes away if you don't set a password :P

app/web/src/features/dashboard/DashboardBanners.tsx
```diff
@@ -51,20 +53,32 @@ export default function DashboardBanners() {
...
+                className={classes.button}
+                color="default"
+                role="link"
+                onClick={() => {
+                  history.push(settingsRoute);
+                }}
+              >
+                {PASSWORD_TEXT_LINK}
+              </Button>
+              <Typography variant="inherit">{PASSWORD_TEXT_1}</Typography>
```
lucaslcode: To me it would make more sense for the button to be after the text

lucaslcode (changes requested):

wbruntra:
By replacing the `component` these buttons (using `a` tags) are fine. I've created a new issue to make a note of the general problem with the `Button` component. https://github.com/Couchers-org/couchers/issues/4167

lucaslcode (approved):
Thanks for your contribution!

PR #2038, opened 2021-10-02
Web - #2014 - change close button to icon button 

### LINKED ISSUE #2014: Remove "close button" and add X to the top left (community nav)
In the current model, the close action button is a motivator to close the pop up window instead of interact with it. If possible, let's try to remove it and add an X component in the top left instead. This encourages the user to interact and it scales better as communities are added to the list. Example below: 

Screen Shot 2021-09-24 at 6 18 39 PM

User interaction can also be boosted by adding arrow indicators to convey the displayed options can be expanded into more sub-communities. As such: 

Screen Shot 2021-09-24 at 6 23 57 PM

goskan93:
Closes #2014

obraz

**Web frontend checklist**
- [ ] Added any new components to storybook - I don't know, should I?
- [ ] All tests pass -- ehh, I dont know xp (windows)

I was also thinking about putting communities' list in the center of the dialog, what do you think?

### DIFF
--- a/app/web/src/components/Dialog.tsx
+++ b/app/web/src/components/Dialog.tsx
@@ -77,12 +77,15 @@ export function DialogContentText(props: DialogContentTextProps) {
   );
 }
 
-interface DialogTitleProps extends MuiDialogTitleProps {
+interface DialogTitleProps extends Omit<MuiDialogTitleProps, "className"> {
   onClose?: () => void;
 }
 
-export function DialogTitle(props: Omit<DialogTitleProps, "className">) {
-  const { children, onClose, ...dialogTitleProps } = props;
+export function DialogTitle({
+  children,
+  onClose,
+  ...dialogTitleProps
+}: DialogTitleProps) {
   const classes = useStyles();
   return (
     <MuiDialogTitle {...dialogTitleProps} className={classes.title}>
--- a/app/web/src/components/Dialog.tsx
+++ b/app/web/src/components/Dialog.tsx
@@ -8,9 +8,11 @@ import {
   DialogContentTextProps,
   DialogProps,
   DialogTitle as MuiDialogTitle,
-  DialogTitleProps,
+  DialogTitleProps as MuiDialogTitleProps,
 } from "@material-ui/core";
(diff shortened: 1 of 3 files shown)


app/web/src/components/Dialog.tsx
```diff
@@ -69,7 +77,25 @@ export function DialogContentText(props: DialogContentTextProps) {
   );
 }
 
+interface DialogTitleProps extends MuiDialogTitleProps {
+  onClose?: () => void;
+}
+
 export function DialogTitle(props: Omit<DialogTitleProps, "className">) {
```
darrenvong: Two things:
- (minor) we usually prefer to destructure in the function parameter rather than another line of code (might want to Prettier format after if applying suggestion directly...)
- It's clearer to reference the type of `props` as `DialogTitleProps` rather than utility type - the required work should be done in the `DialogTitleProps` type itself

```suggestion
export function DialogTitle({ children, onClose, ...dialogTitleProps }: DialogTitleProps) {
```

app/web/src/components/Dialog.tsx
```diff
@@ -69,7 +77,25 @@ export function DialogContentText(props: DialogContentTextProps) {
   );
 }
 
+interface DialogTitleProps extends MuiDialogTitleProps {
```
darrenvong:
```suggestion
interface DialogTitleProps extends Omit<MuiDialogTitleProps, "className"> {
```
So you can use `DialogTitleProps` directly in the component

darrenvong (changes requested):
Mostly looking good!

> I was also thinking about putting communities' list in the center of the dialog, what do you think?

Do you mean it will then expand/move to the left from the centre as more levels are opened? Hmm not too sure about this until I see it in action

goskan93:
> 
> 
> Mostly looking good!
> 
> > I was also thinking about putting communities' list in the center of the dialog, what do you think?
> 
> Do you mean it will then expand/move to the left from the centre as more levels are opened? Hmm not too sure about this until I see it in action

ok, will leave it for now ;)

darrenvong (approved):

PR #2047, opened 2021-10-03
Open new window in browser in order to navigate to Forum and About websites

### LINKED ISSUE #2040: About and Forum  navigation buttons should open new tab with respective website
obraz

What I was told once, is that we should never let a user to leave our page when using :) and those two lead to another websites, so later it is hard to come back to dashboard.

goskan93:
closes #2040

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/components/Navigation/ExternalNavButton.tsx
+++ b/app/web/src/components/Navigation/ExternalNavButton.tsx
@@ -15,7 +15,12 @@ export default function ExternalNavButton({
 }: ExternalNavButtonProps) {
   const classes = useNavLinkStyles();
   return (
-    <a href={route} className={classes.link}>
+    <a
+      href={route}
+      className={classes.link}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
       <Typography variant={labelVariant} className={classes.label} noWrap>
         {label}
       </Typography>


darrenvong:
I'm not too sure about this being an improvement, as it's also known to be bad for accessibility because it breaks the back button as written extensively in [these](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) [articles](https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/). Also, the user loses the control to decide whether to open the link in a new tab or not.

In general, we should only open links in a new tab if it means the user will lose important state - e.g. they are in the midst of editing their profile/filling out some form. The dashboard doesn't really fall under that category hence my hesitancy on this!

goskan93:
> 
> 
> I'm not too sure about this being an improvement, as it's also known to be bad for accessibility because it breaks the back button as written extensively in [these](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) [articles](https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/). Also, the user loses the control to decide whether to open the link in a new tab or not.
> 
> In general, we should only open links in a new tab if it means the user will lose important state - e.g. they are in the midst of editing their profile/filling out some form. The dashboard doesn't really fall under that category hence my hesitancy on this!

What you say is true for internal links but not external one. It feels like a terrible thing to be kicked out from the website I was using because I clicked a button that I actually didnt know will take me out from main content. It doesn't seem like a good user experience (at least I didn't)
My source says something different https://uxmovement.com/navigation/why-external-links-should-open-in-new-tabs/

goskan93:
> 
> 
> > I'm not too sure about this being an improvement, as it's also known to be bad for accessibility because it breaks the back button as written extensively in [these](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) [articles](https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/). Also, the user loses the control to decide whether to open the link in a new tab or not.
> > In general, we should only open links in a new tab if it means the user will lose important state - e.g. they are in the midst of editing their profile/filling out some form. The dashboard doesn't really fall under that category hence my hesitancy on this!
> 
> What you say is true for internal links but not external one. It feels like a terrible thing to be kicked out from the website I was using because I clicked a button that I actually didnt know will take me out from main content. It doesn't seem like a good user experience (at least I didn't) My source says something different https://uxmovement.com/navigation/why-external-links-should-open-in-new-tabs/

Now I think maybe there is a small misunderstanding, because in the title of this PR I wrote to open new window, but obviously I meant new tab :)

darrenvong:
> > I'm not too sure about this being an improvement, as it's also known to be bad for accessibility because it breaks the back button as written extensively in [these](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) [articles](https://www.nngroup.com/articles/the-top-ten-web-design-mistakes-of-1999/). Also, the user loses the control to decide whether to open the link in a new tab or not.
> > In general, we should only open links in a new tab if it means the user will lose important state - e.g. they are in the midst of editing their profile/filling out some form. The dashboard doesn't really fall under that category hence my hesitancy on this!
> 
> What you say is true for internal links but not external one. It feels like a terrible thing to be kicked out from the website I was using because I clicked a button that I actually didnt know will take me out from main content. It doesn't seem like a good user experience (at least I didn't) My source says something different https://uxmovement.com/navigation/why-external-links-should-open-in-new-tabs/

I don't really want to get too deep into the semantics of things... but is it really an external link though? The forum and about page is technically our website so I'd still say it's internal to our site. This won't really be an issue once the landing site and the app merges, although I'll concede the design of the the forum looks different enough that it looks like an external site. I'm not clued up on if there's much control over the forum design though as we didn't write the software.

I largely disagree with most things mentioned on there (especially as those points at the second half of the blog are irrelevant to us). Clicking on the link only introduce one more page on the history stack, which is hardly "back button fatigue". Clicking links and being able to go back (with back button, or keyboard shortcuts for power users, screen reader users etc) is a well known behaviour and so I'd favour keeping that over the "fatigue" opinion.

The comment section of that blog post is more useful than the post itself:
> I’m incredibly surprised that you can suggest this course of action and not have written anything in your post about accessibility or screen readers? Why not just **leave the user in control** – after all, it is their experience, and they know best whether they like tabbed browsing or not, or whether it causes them a problem.

The bit I highlighted in bold is the main reason as mentioned before, regardless of if the link is an external or internal one. Opening things in a new tab when it's not absolutely necessary removes that control from the user which I'd say makes our site less accessible overall.

W3C also seem to agree with the comment about avoiding open links in new tabs/windows where possible as it's disorientating for certain users - https://www.w3.org/TR/WCAG20-TECHS/G200.html - and given they set standard for the web (including accessibility ones), I trust their recommendation is based on some extensive research.

lucaslcode:
I agree with Darren. I think there is some argument if the external links were from, say, a chat message, or in the edit profile screen, but these are intended as navigation links. Other points in favour of not opening a new tab:

- It is easy to open a normal link in a new tab, but difficult to open a new tab link in the same tab
- Back button usage is extremely common, but "breaking the back button" can be a big frustration
- Other websites sometimes open links in new tabs, but the context is different, and also we are very happy for people to leave the website unlike Facebook :P A better comparison might be Google, which doesn't open new tabs

It will be better once we merge the app and landing page - maybe we can revisit then.

goskan93:
Hmmm looks you are more right than me :) BUUUUT, I can't get out of the bad imporession that I had when I noticed it - I think we could still improve it.
When I look at the navigation bar of some website, I am assuming that the links in its navigation bar will lead me around the website that I am visiting, and not suprisingly goes to another one. If you have a button that goes to facebook, user immediately knows that this button will open another website, and in our navigation bar you cannot distinguish between external links and internal one. 
So if not open a new tab, then mark somehow (another style, icons, put them in other place, I dont know) those buttons that lead to other websites, so that user knows that those buttons are little bit different.

darrenvong:
It sounds like from what you're saying, the main source of confusion is the fact the forum/main site is on a "different" website, which is more of a problem with the current architecture of our sites than the link styling. I think it is better to address this by (eventually) merging the landing site (what looks like an "external" site) with the web app. Given this is quite a low priority issue, I think this is acceptable to wait for now.

Since our navigation bar is so packed already, I don't see a good way other than adding an icon next to the link, but it's debatable [whether icons add much clarity](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/) other than the really well known ones, so it might be adding noise in that case and slight technical complexity.

goskan93:
> It sounds like from what you're saying, the main source of confusion is the fact the forum/main site is on a "different" website, which is more of a problem with the current architecture of our sites than the link styling. I think it is better to address this by (eventually) merging the landing site (what looks like an "external" site) with the web app. Given this is quite a low priority issue, I think this is acceptable to wait for now.
> 
> Since our navigation bar is so packed already, I don't see a good way other than adding an icon next to the link, but it's debatable [whether icons add much clarity](https://designnotes.blog.gov.uk/2016/11/28/removing-the-external-link-icon-from-gov-uk/) other than the really well known ones, so it might be adding noise in that case and slight technical complexity.

Merging websites would be good solution too.

PR #2048, opened 2021-10-03
feature: use enter key to send chat

### LINKED ISSUE #2049: press enter to send chat
In online chats, users will generally expect to be able to use the `enter` key to send messages (e.g. WhatsApp, Facebook Messenger). The chat function on couchers should allow this behavior.

I have created a pull request here: https://github.com/Couchers-org/couchers/pull/2048

wbruntra:
I hadn't opened an issue for this before creating the PR, which would have given more opportunity for discussion. Take this as a suggestion and feel free to discuss/reject.

In online chats, users will generally expect to be able to use the `enter` key to send messages (e.g. WhatsApp, Facebook Messenger). The chat function on couchers should allow this behavior.

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
@@ -29,9 +29,16 @@ export default function GroupChatSendField({
 
   const { register, handleSubmit, reset } = useForm<MessageFormData>();
   const onSubmit = handleSubmit(async (data: MessageFormData) => {
-    handleSend(data.text);
+    handleSend(data.text.trimRight());
     reset();
   });
+
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === "Enter" && !event.shiftKey) {
+      event.preventDefault();
+      onSubmit();
+    }
+  };
 
   return (
     <form onSubmit={onSubmit} className={classes.container}>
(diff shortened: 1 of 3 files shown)


lucaslcode (commented):
Summary from speaking to design team, we want requests to not be "chat"-y but normal messages can be. Probably there will be a different UI for initial requests. We're also planning a refactor of messaging soon(tm) so host requests are integrated into messages rather than separate.

For now, if you want, we can enable this but not for the first message on host request chats.
Actually, looks like you've only done it for "GroupChats" (these should really just be called chats and will be renamed in the refactor) so it's probably fine as is

lucaslcode (dismissed):
After going through the flow, we actually already have a separate bit for the initial message, so I think this is fine for host messages too. Do you want to add it there (`HostRequestSendField` or whatever it is)?

I know copy-paste isn't ideal but it's fine here - one of these will be deleted with the refactor.

lucaslcode (changes requested):
See above

wbruntra:
> Actually, looks like you've only done it for "GroupChats" (these should really just be called chats and will be renamed in the refactor) so it's probably fine as is

Yeah, I went ahead and updated it on the host/surfer messages as well, per your second comment, but initially my thinking was to reflect the different expectations in terms of "chattiness" between a "chat" and the host/surfer interaction.

No strong opinion on whether it's better to use enter-to-send on both for consistency or treat host/surfer messages differently. And after all, you can always just send two messages :). For what it's worth, I agree that the best thing at some point in the future will be an interface that distinguishes chats from other types of messages.

lucaslcode (approved):

PR #2071, opened 2021-10-10
Add a footer

### LINKED ISSUE #1690: Footer

lucaslcode:
closes #1690 

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/AppRoute.tsx
+++ b/app/web/src/AppRoute.tsx
@@ -31,6 +31,8 @@ export const useAppRouteStyles = makeStyles((theme) => ({
     paddingRight: 0,
   },
   loader: {
+    //minimal-effort reduction of layout shifting
+    minHeight: "50vh",
     display: "flex",
     justifyContent: "center",
     alignItems: "center",
@@ -88,7 +90,17 @@ export default function AppRoute({
                       <Redirect to={jailRoute} />
                     ) : (
                       <>
-                        <ErrorBoundary>{children}</ErrorBoundary>
+                        <ErrorBoundary>
+                          <Suspense
+                            fallback={
+                              <div className={classes.loader}>
+                                <CircularProgress />
+                              </div>
+                            }
+                          >
+                            {children}
+                          </Suspense>
+                        </ErrorBoundary>
                       </>
                     )}
                   </Container>
--- a/app/web/src/AppRoute.tsx
+++ b/app/web/src/AppRoute.tsx
@@ -1,7 +1,8 @@
-import { Container } from "@material-ui/core";
+import { Container, useTheme } from "@material-ui/core";
 import classNames from "classnames";
 import CookieBanner from "components/CookieBanner";
 import ErrorBoundary from "components/ErrorBoundary";
+import Footer from "components/Footer";
 import { useEffect } from "react";
 import { Redirect, Route, RouteProps } from "react-router-dom";
 import makeStyles from "utils/makeStyles";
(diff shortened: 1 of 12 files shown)


lucaslcode:
image

Need to fix the mobile layout of the very bottom bit and also remove the top links.
What links do we still want at the top? @ItsiW @AllhandsNoFeet ?

jesseallhands:
"Help" needs to be at the top right. The top-left links should stay for now I think.

ItsiW:
Could we do a dropdown at the top right for profile, settings, help, log out?

Then we could get rid of everything top right and profile.

jesseallhands:
A drop-down menu (or slide-out menu) could work if we have a hamburger button. Maybe similar to what airbnb has? But I would argue we should keep profile over on the upper-left corner since it's something that gets updated quite often on a new platform.

image

app/web/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,198 @@
...
+    backgroundColor: theme.palette.background.paper,
+  },
+  upperContainer: {
+    width: "100%",
+    display: "grid",
+    rowGap: theme.spacing(1),
+    columnGap: theme.spacing(1),
+    gridTemplateColumns: "auto auto",
+    [theme.breakpoints.up("sm")]: {
+      gridTemplateColumns: "auto auto auto auto",
```
darrenvong:
```suggestion
      gridTemplateColumns: "repeat(4, auto)",
```
I think this would also work and slightly shorter to write

app/web/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,198 @@
...
+    },
+    [theme.breakpoints.up("md")]: {
+      gridTemplateColumns: "auto auto auto auto",
```
darrenvong: Same as above

app/web/src/AppRoute.tsx
```diff
@@ -58,86 +59,96 @@ export default function AppRoute({
...
-                    <ErrorBoundary>
-                      <Suspense
-                        fallback={
-                          <div className={classes.loader}>
-                            <CircularProgress />
-                          </div>
-                        }
-                      >
-                        {children}
-                      </Suspense>
```
darrenvong, line 96: I think we still need this `Suspense` here since we are now lazy loading most routes. It's crashing on preview at the moment with this error in the console:

```
react-dom.production.min.js:216 Error: A React component suspended while rendering, but no fallback UI was specified.

Add a <Suspense fallback=...> component higher in the tree to provide a loading indicator or placeholder to display.
```

app/web/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,198 @@
...
+  );
+}
+
+function FooterLink({ to, children }: { to: string; children: ReactNode }) {
+  if (to.startsWith("http")) {
+    return (
+      <MuiLink
+        href={to}
+        color="textSecondary"
+        style={{ display: "block" }}
```
darrenvong: This should be done with CSS and classes like in the rest of the component?

app/web/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,198 @@
...
+        target="_blank"
+        rel="noopener"
+      >
+        {children}
+      </MuiLink>
+    );
+  }
+  return (
+    <StyledLink to={to} color="textSecondary" style={{ display: "block" }}>
```
darrenvong: Same comment as above

darrenvong (changes requested):
Mostly looking good already. I think things are still behaving weirdly with the responsive styling when the "report a problem" button becomes too long on specific breakpoints and causes horizontal scrolling (i.e. things don't look as "centred" when that happens)

app/web/src/AppRoute.tsx
lucaslcode: Whoops, it was a bad merge. Git formatted the diff really strangely

app/web/src/components/Navigation/ExternalNavButton.tsx
```diff
@@ -15,7 +15,7 @@ export default function ExternalNavButton({
 }: ExternalNavButtonProps) {
   const classes = useNavLinkStyles();
   return (
-    <a href={route} className={classes.link}>
+    <a href={route} target="_blank" rel="noreferrer" className={classes.link}>
```
darrenvong:
```suggestion
    <a href={route} target="_blank" rel="noreferrer noopener" className={classes.link}>
```
Good security practice as mentioned here - https://web.dev/external-anchors-use-rel-noopener/

app/web/src/theme.ts
```diff
@@ -86,6 +86,10 @@ const themeOptions: ThemeOptions = {
       200: "#e9e9e9",
       600: "#767676",
     },
+    text: {
+      primary: "#313539",
+      secondary: "#767676",
+    },
```
darrenvong, line 92: Are these on Figma somewhere?

darrenvong (changes requested):

app/web/src/components/Navigation/ExternalNavButton.tsx
lucaslcode: That page says noopener *or* noreferrer. O actually want to keep referrer so external pages know if we are giving them traffic.

app/web/src/theme.ts
lucaslcode: I just clicked of a few instances of primary and secondary text to see, I don't know if they are all the same.

app/web/src/components/Navigation/ExternalNavButton.tsx
darrenvong: Lol I only skimmed and didn't see the _or_ - apparently `noreferrer` implies `noopener` but it's n[ot well supported by all browsers](https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable), so probably still best to include them both.

Apparently it doesn't affect SEO from all the top Google results, so I think it's fine to do this.

darrenvong: Or maybe only do `noopener` if we want to keep the referrer? What's the downside of not keeping that around apart from messing with other site's analytics (which doesn't sound like our problem to solve)?

lucaslcode: I guess it doesn't really matter, committed your suggestions

darrenvong (dismissed):

darrenvong (approved):

Issue #2083, opened 2021-10-13
Pre-messaging refactor stopgap bug fixes


lucaslcode:
- Allow sending of messages to non-friends
- Allow marking of past requests as seen

Any other frequent complaints worth fixing first?

aapeliv:
> Allow marking of past requests as seen

I believe this is a web thing. I looked in the backend and it doesn't seem like there's anything stopping you from marking past reqs as seen there.

PR #2090, opened 2021-10-14
Allow messaging anybody

aapeliv:
* Backend: removes the requirement to be friends to message a user
* Frontend: add button to message a user regardless of friendship status
* Frontend: add a button next to messages to report that message

Screenshot from 2021-10-14 12-58-20

Closes part of #2083

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [x] All tests pass


aapeliv:
Frontend: It would be great to get the chat id into the message so that it can be added to the `contentRef` for the flag button, but that requires passing it down a few levels into subcomponents, so not sure how to do that exactly

darrenvong:
Isn't the `messageId` globally unique at the moment since all messages are on the same table? Would that not be sufficient for the `contentRef`?

aapeliv:
Yes, it suffices for the time being; but it’d be easier for the support team to have both. So not a blocker.

darrenvong:
Should this be added to the `content_ref` docs as well?

darrenvong (dismissed):
Web frontend looks fine to me!

aapeliv:
> Should this be added to the `content_ref` docs as well?

Yes, thanks

kalvdans (approved):
Backend looks fine!

Missing a link to some motivation for the decision though.

aapeliv:
> Missing a link to some motivation for the decision though.

It's been discussed a lot on Slack

PR #2093, opened 2021-10-15
Set up react-i18next on web

darrenvong:
Sets up the react-i18next library for web, so it's ready for localisation. Since we are planning to do this incrementally, I've chosen one feature area (the donations page) to start off with, just so there is a reference on what the new code looks like as well as giving a JSON resource file for setting up Weblate once this is merged.

**Web frontend checklist**
- [x] All tests pass


app/web/src/i18n.ts
```diff
@@ -0,0 +1,42 @@
...
+            const hasDecimal = value % 1 !== 0;
+            return Intl.NumberFormat(lng, {
+              style: "currency",
+              currency: rest[0],
+              minimumFractionDigits: hasDecimal ? 2 : 0,
+            }).format(value);
+          }
+        }
+        return value;
+      },
```
darrenvong, line 37: It's a bit annoying I have to add this custom formatter in, as I couldn't figure out why the built-in `currency` formatter in the example - https://www.i18next.com/translation-function/formatting#currency - won't work properly (it just returns the number itself without the dollar sign 🙄 )

app/web/src/features/donations/Donations.tsx
```diff
@@ -158,11 +149,16 @@ export default function Donations() {
...
+          <Trans t={t} i18nKey="benefactor_contact">
+            If you wish to contribute over $1000, please contact us at
+            <Link
+              className={classes.link}
+              href={`mailto:${t("benefactor_email")}`}
+            >
+              {{ email: t("benefactor_email") }}
+            </Link>
+            for us to arrange a lower fee transfer.
+          </Trans>
```
darrenvong, line 161: This may look like we're hardcoding the message again and it's a bit confusing, but it's not and only used for figuring out what to swap in from this template string in the language JSON file: `"If you wish to contribute over $1000, please contact us at <1>{{email}}</1> for us to arrange a lower fee transfer."`

It's needed since we are mixing React/HTML components with the sentence, so the `Trans` component takes this "hardcoded template" and break into 3 children: the hardcoded text on the left, the `Link` and the hardcoded text on the right. It then figures out the actual text to swap/substitute in based on what's to the left/right of `<1>{{email}}</1>` (1 since `Link` is the second child element to the `Trans` component).

From looking at the examples, it seems like the best practice is to leave the English version of the text as template here and change the value in the JSON accordingly, so it gives us a rough idea of what to expect on screen.

app/web/src/features/donations/DonationsBox.tsx
```diff
@@ -259,32 +251,50 @@ export default function DonationsBoxMixed() {
...
-          {DONATIONSBOX_TITLE}
+          {t("donations_box.title")}
         </Typography>
         <Controller
           id="recurring"
           control={control}
           name="recurring"
           rules={{
-            required: DONATIONSBOX_REQUIRED,
+            required: t("donations_box.validation_message") as string,
```
lucaslcode: Why is the type assertion needed?

app/web/src/features/donations/Donations.tsx
```diff
@@ -128,15 +117,17 @@ const useStyles = makeStyles((theme) => ({
 export default function Donations() {
   const classes = useStyles();
 
+  const { t } = useTranslation("donations");
+
   return (
     <>
-      <HtmlMeta title={DONATE} />
+      <HtmlMeta title={t("donate")} />
```
lucaslcode, line 124: Do these keys autocomplete?

lucaslcode (changes requested):
Looks good. Two things:

- Do we need to namespace like that? I guess you were trying to let the file co-exist in the folder with the feature which is nice, but I don't think it's going to work nicely. Not all our stuff fits easily in the features folder (there are strings for components, for example), ~~and if we want "common" strings for common stuff share by most pages we'll have to do multiple namespaces and use `t("cancel", { ns: "common" })`. Is there a downside to having it all together other than the fact it's going to be one huge file? (Or can we split without namespaces?)~~
- Actually looks like you can do `t("common:cancel")` according to the docs? And we can always add more imports as necessary.
- Do you want to try doing the typescript setup too: https://react.i18next.com/latest/typescript

app/web/src/features/donations/DonationsBox.tsx
darrenvong: It was complaining that it's not returning a string, but I wonder if doing the TypeScript setup will fix this (that I missed on their page, the docs are annoyingly scattered between i18next and react-i18next), so will try that first!

darrenvong:
I think the namespacing will help prevent key collisions, which will get trickier to spot as the app continues to grow and become more complex. The other benefit of splitting up the translations into multiple files is code splitting - this seems to be the suggested approach by the docs to ensure we only load in the translation needed to keep the initial app bundle size small. It fits well with the feature folder structure already so I thought it makes sense to go this way and it's also nicer for maintenance than having everything in a giant file imo.

For shared stuff, I guess the keys can either be under `common` or namespaced by the component name? I don't think having to add multiple namespaces in the `useTranslation` hook is going to be that bad - it's not that different to an additional import to me at least.

app/web/src/features/donations/Donations.tsx
darrenvong: It does after adding augmenting the TypeScript types mentioned in the docs!

lucaslcode (approved):
Good once green

PR #2098, opened 2021-10-17
Enable Gitpod prebuilds in all branches

darrenvong:
I saw this in the Gitpod docs and it seems like turning this on means it will build the workspace whenever there's a new change pushed to a branch, as controlled by this update in the `.gitpod.yml` config. This means less waiting around for things to build as the workspace should hopefully be ready by the time someone opens it.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -25,3 +25,9 @@ ports:
     onOpen: open-preview
   - port: 8888
     visibility: public
+
+github:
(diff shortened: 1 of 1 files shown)


aapeliv (approved):
Good if it works and we don't have to pay for it!

darrenvong:
Let's see, the worst case is it doesn't work and it's the same as before so not much to lose!

PR #2102, opened 2021-10-17
Change reference reminders

aapeliv:
New schedule:

1. A reminder on the day they leave, at around 8 pm
2. A reminder about a week after, at 2 pm
3. A reminder 3 days and 14 hours before end time to write ref

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -242,11 +242,11 @@ def process_send_reference_reminders(payload):
     reference_reminder_schedule = [
         # (number, days before you end of write reference period, text for how long they have left to write the ref)
         # 8 pm ish on the last day of the stay
-        (1, timedelta(days=14) - timedelta(hours=20), "14 days"),
+        (1, timedelta(days=15) - timedelta(hours=20), "14 days"),
         # 2 pm ish a week after stay
-        (2, timedelta(days=7) - timedelta(hours=14), "7 days"),
+        (2, timedelta(days=8) - timedelta(hours=14), "7 days"),
         # 10 am ish 3 days before end of time to write ref
-        (3, timedelta(days=3) - timedelta(hours=10), "3 days"),
+        (3, timedelta(days=4) - timedelta(hours=10), "3 days"),
     ]
 
     with session_scope() as session:
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -241,8 +241,8 @@ def process_send_reference_reminders(payload):
     # Keep this in chronological order!
     reference_reminder_schedule = [
         # (number, days after stay, text for how long they have left to write the ref)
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -240,13 +240,13 @@ def process_send_reference_reminders(payload):
 
     # Keep this in chronological order!
     reference_reminder_schedule = [
-        # (number, days after stay, text for how long they have left to write the ref)
-        # 6 pm ish two days after stay
-        (1, timedelta(days=2) - timedelta(hours=6), "12 days"),
+        # (number, days before you end of write reference period, text for how long they have left to write the ref)
```
kalvdans: spelling error + that it is easier to state how long after the stay ended that the reminder should be sent. It is a bit awkward to have it relative the last time to write reference.

app/backend/src/couchers/models.py
```diff
@@ -938,6 +938,7 @@ class HostRequest(Base):
     # timezone aware start and end times of the request, can be compared to now()
     start_time = column_property(date_in_timezone(from_date, timezone))
     end_time = column_property(date_in_timezone(to_date, timezone) + text("interval '1 days'"))
+    start_time_to_write_reference = column_property(date_in_timezone(to_date, timezone))
```
kalvdans, line 941: Document new state. When is this state set?

kalvdans (changes requested):

app/backend/src/couchers/jobs/handlers.py
aapeliv: Will fix the spelling of course. Did you mean I should switch it back the other way?

I think it's better to have it relative to time left to write the reference... after all, it's a reminder to write a reference before time runs out.

app/backend/src/couchers/models.py
aapeliv: These are expressions that SQLAlchemy will execute in the db when pulling the row. They're not extra state, just some helper expressions. This is because the timezone transformations (that don't yet happen since currently this is in UTC) are easier to do in the db.

app/backend/src/couchers/jobs/handlers.py
kalvdans: Yes but the comments "10 am" etc is really hard to judge if they are correct when we don't know the reference expiration time. And they are easy to forget to update should we change the reference expiration time.

app/backend/src/couchers/models.py
kalvdans: True, sorry for the noise! Resolving.

app/backend/src/couchers/jobs/handlers.py
aapeliv: I tried to clarify

kalvdans: Good! Sorry to respond late.

kalvdans (approved):

PR #2117, opened 2021-10-19
Disable enter-to-send on host requests/replies

### LINKED ISSUE #2096: Enter key sends message unexpectedly
Subject: Message response to request to host
Description:
Hit return while writing a reply.

Results:
In replying to a message, I discovered that a carriage return *sends* the message to the recipient rather than simply moving down one line (paragraph formatting). Thus far, I have sent two single lines unexpectedly instead of a normally formatted paragraph. 

Backend version: develop-e33b9ff0
Frontend version: develop-e33b9ff0
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.0 Safari/605.1.15
Page: https://app.couchers.org/messages/request/2187
User: charlesfrenchquarter (2736)

wbruntra:
Intending to close https://github.com/Couchers-org/couchers/issues/2096

Partially reverting https://github.com/Couchers-org/couchers/pull/2048, now using Enter key only for the "Chat" feature, not for host messages/replies, which are intended to be longer and have paragraphs. 

May want to remove enter-to-send in Chat as well, for consistency or just because, though personally I think enter-to-send makes sense for chats.

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
@@ -34,7 +34,7 @@ export default function GroupChatSendField({
   });
 
   const handleKeyDown = (event: React.KeyboardEvent) => {
-    if (event.key === "Enter" && !event.shiftKey) {
+    if (event.key === "Enter" && event.ctrlKey) {
       event.preventDefault();
       onSubmit();
     }
--- a/app/web/src/features/messages/groupchats/GroupChatView.test.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatView.test.tsx
@@ -308,7 +308,7 @@ describe("GroupChatView", () => {
     expect(getGroupChatMessagesMock).toHaveBeenCalledTimes(2);
   });
 
-  it("sends the message successfully via enter key", async () => {
+  it("sends the message successfully via ctrl+enter combination", async () => {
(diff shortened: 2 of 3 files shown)


wbruntra:
Upon further consideration (thanks @aapeliv), this has been changed to use ctrl+enter to send messages, to make the change as unobtrusive as possible to casual users.

lucaslcode (dismissed):
Thanks!!

wbruntra:
Wait sorry, I forgot to update and re-run the tests when I changed the chat behavior.

aapeliv (approved):

darrenvong:
Hmm both Facebook and Slack uses Enter to send in a chat, so I'm not sure Ctrl+Enter is the norm and seems more obscure to me 🤔

aapeliv:
I think it should be the same everywhere for consistency

wbruntra:
> Hmm both Facebook and Slack uses Enter to send in a chat, so I'm not sure Ctrl+Enter is the norm and seems more obscure to me thinking

As I understand it the plan is ultimately to have more visual/design differentiation between Chat and Messages, at which point I'd agree that just `Enter` should be enough within a feature called "chat." But it seems like that kind of fine-tuning should be low-priority for now.

PR #2124, opened 2021-10-22
Automatically convert links within user messages

wbruntra:
Adding a component to render user messages containing links with `<a>` tags instead of plaintext.

This actually turns out to be surprisingly complicated, and would be even harder if you really wanted to be sure to cover all possible URLs while avoiding any false positives. So I decided to try for something "good enough" instead.

URL regex modified from: https://github.com/kevva/url-regex

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/src/components/Linkify.tsx
+++ b/app/web/src/components/Linkify.tsx
@@ -11,9 +11,6 @@ const useStyles = makeStyles((theme) => {
 });
 
 const urlRegex = () => {
-  /*
-
-  */
   const protocol = `(?:https?://)?`;
   const auth = "(?:\\S+(?::\\S*)?@)?";
   const host = "(?:(?:[a-z\\u00a1-\\uffff0-9][-_]*)*[a-z\\u00a1-\\uffff0-9]+)";
--- a/app/web/src/components/Linkify.tsx
+++ b/app/web/src/components/Linkify.tsx
@@ -1,0 +1,67 @@
+import classNames from "classnames";
+import React from "react";
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => {
+  return {
+    a: {
+      color: theme.palette.primary.main,
+    },
+  };
+});
+
+const urlRegex = () => {
+  /*
+
+  */
+  const protocol = `(?:https?://)?`;
+  const auth = "(?:\\S+(?::\\S*)?@)?";
+  const host = "(?:(?:[a-z\\u00a1-\\uffff0-9][-_]*)*[a-z\\u00a1-\\uffff0-9]+)";
+  const domain =
+    "(?:\\.(?:[a-z\\u00a1-\\uffff0-9]-*)*[a-z\\u00a1-\\uffff0-9]+)*";
+  const tld = `(?:\\.)[a-z]{2,}`;
+  const path = '(?:[/?#][^\\s"]*)?';
+  const regex = `(?:${protocol}${auth}(?:${host}${domain}${tld})${path})`;
+  const result = new RegExp(`(${regex})`, "ig");
+  return result;
+};
+
+interface LinkifyProps {
+  text: string;
+}
+
+function Linkify(props: LinkifyProps) {
+  const classes = useStyles();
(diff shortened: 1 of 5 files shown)


app/web/src/components/Linkify.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const regex = `(?:${protocol}${auth}(?:${host}${domain}${tld})${path})`;
+  const result = new RegExp(`(${regex})`, "ig");
+  return result;
+};
+
+interface LinkifyProps {
+  text: string;
+}
+
+function Linkify(props: LinkifyProps) {
```
lucaslcode: Nitpick - we usually dereference props

app/web/src/components/Linkify.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const classes = useStyles();
+  const nonCapturingRegex = urlRegex();
+  const parts = props.text.split(nonCapturingRegex);
+
+  const result = parts.map((part, i) => {
+    if (!part || !part.match) {
+      return undefined;
+    }
```
lucaslcode: I think returning `null` is preferred when not rendering anything (though this is indeed valid)

lucaslcode (changes requested):
Looks good, and thanks for tests!

Comments on events and discussions are actually Markdown, not plain text as I thought, but do you want to enable autolinks by adding `extendedAutolinks: true,` to both `components/Markdown.tsx` and `components/MarkdownEditor/MarkdownEditor.tsx` in the `ToastUIEditor` constructor? (That's the markdown library we're using).
Also, you can clone the repo and make a branch next time, and you'll get our automated checks/previews etc.

app/web/src/components/Linkify.tsx
wbruntra: I'm not the best on terminology so to be sure I understand what you're looking for, you mean to use `{ text }` instead of `props` and `props.text`?

wbruntra:
@lucaslcode I believe this addresses all of your specific comments. You referenced `components/MarkdownEditor/MarkdownEditor.tsx`, which does not exist, so I changed `MarkdownInput.tsx`. Let me know if that's not the file you intended.

> Also, you can clone the repo and make a branch next time, and you'll get our automated checks/previews etc.

Okay, will do

wbruntra:
I'll just note here that the extended links feature on Toast UI recognizes links a bit differently from my implementation. Specifically, they require the link start with either `www` or a protocol, which I didn't do, although that's definitely a good way of avoiding false positives. But honestly, you see a lot of false positives on e.g. twitter, which makes me think it isn't really that big of a deal. And being a frugal person, I am a fan of non-`www` subdomains. It works well enough.it doesn't have to be perfect.

app/web/src/components/Linkify.tsx
lucaslcode: Yep!

lucaslcode:
> it doesn't have to be perfect

I agree! If people complain about false-positives we can always change it later.

app/web/src/components/Linkify.tsx
```diff
@@ -0,0 +1,64 @@
...
+    if (part.match(nonCapturingRegex)) {
+      // Exclude email addresses
+      if (part.includes("@") && !part.includes("/")) {
+        return <React.Fragment key={i}>{part}</React.Fragment>;
+      }
+      const href = part.endsWith(".") ? part.slice(0, -1) : part;
+      const protocolPrefix = part.match(/https?:?\/\//) ? "" : "//";
+      return (
+        <a
```
lucaslcode: If you use `<Link>` from material ui (not from react-router-dom), it will get correct styling from the theme automatically and you won't need a className

lucaslcode (changes requested):
This is great, sorry for forgetting to mention this one thing last time. Then it's ready to merge.

app/web/src/components/Linkify.tsx
wbruntra: Ah, very good, thanks

lucaslcode (approved):
Great! Go ahead and merge it :)

PR #2135, opened 2021-10-26
Chore: Convert constants in web to JSON resources

darrenvong:
As in the title - converts all the constant file into the corresponding JSON. This doesn't include the work to actually use it in the UI. I think that's too much work to do in one go and better done gradually as we make changes to the app as we go.

Also since it only picks out constants with string values, there are some things missing in this (i.e. anything that uses a `getX` function to derive the string), but I thought that's ok since converting that into proper i18n JSON format can be part of the work of actually using the resources in the UI?

Another thing to note is that the variable naming is not great according to translation key naming best practices [as mentioned here](https://lokalise.com/blog/translation-keys-naming-and-organizing/#2_Give_translation_keys_meaningful_names) - i.e. the keys currently are too specific to the actual string value, so are a bit brittle to changes. Again, I think any renaming is best sorted when we actually swap over to use the resource in the UI.

**Web frontend checklist**
- [x] All tests pass


lucaslcode:
I think we should rename any keys that we ultimately want to before merging this. People are going to start translating already most likely, so we don't want to change the keys later

lucaslcode:
Actually, is there a way we can hide these files from weblate for now?

darrenvong:
I don't think renaming the keys will be too bad as long as they are not being used in the UI code yet (since the change will be scoped to one JSON vs multiple files that use the translation key once it's in use). I guess it's deciding whether we want to do an audit of the full codebase now and change everything in one go vs doing it bit by bit.

The translation files won't actually show up in Weblate until I set up a component for them for each file, so maybe we only set that up once the keys have been renamed?

lucaslcode (approved):

PR #2146, opened 2021-10-30
NextJS

lucaslcode:
RIP


lucaslcode:
Notes so far:
This isn't actually THAT many lines, git just didn't pick up some files as renames/moving location and the yarn.lock change is huge.

I think NextJS 12 is fine, I think it's working using the rust compiler (swc) for the app and babel for testing.

Things that I have been going through and refactoring:
- react-router Link doesn't exist, need to change to next's Link which a) takes `href` instead of `to` and b) doesn't render anything. See [docs](https://nextjs.org/docs/api-reference/next/link).
- react router switch and routes no longer exist. Routing is file-based in the `pages` folder. I moved `AppRoute` to `components` and am using it as a "layout" as suggested in the NextJS docs. See `pages/login.tsx`
- Also note from that page, routes in the `pages` folder are just a skeleton that has the relevant component from the feature folder. I think this is the best way for now as it keeps as much co-location as possible - may have to modify as we do more server rendering.
- There are lots of tests which use <Switch> and <Route>. I put in a `next-router-mock` library - haven't tested yet but seems it should work. You can set current route with `mockRouter.setCurrentUrl` and get current with `mockRouter.pathname`. Can't test if `router.back` is successful I think...
- `useHistory` -> `useRouter`. If using outside of event callbacks, you need to check `typeof window !== "undefined"` because the hook doesn't work on the server.
- No route state in nextjs. Will replace with query strings.
- `REACT_APP_ENV_VAR` -> `NEXT_PUBLIC_ENV_VAR`
- image imports are no longer urls, but objects with src, width, height and blurDataUrl

darrenvong:
> * Also note from that page, routes in the `pages` folder are just a skeleton that has the relevant component from the feature folder. I think this is the best way for now as it keeps as much co-location as possible - may have to modify as we do more server rendering.

Is the plan to convert the react-snap pages to server-side render, then leave the rest of the app client-side rendered as before and change to server rendering as needed for this first step? Otherwise it will be a lot of code changes...

lucaslcode:
Nothing will be server-rendered in this PR (but everything will be "snapped" ie. static export. Don't think you can do it halfway)

darrenvong:
> but everything will be "snapped" ie. static export.

How would that work for pages that require the user to be logged in first? Will it be static HTML with the loader, then React Query fetches data client-side once the app is hydrated?

lucaslcode:
Yes

lucaslcode:
@darrenvong this might be a good point to have a look and make some comments. Only login and dashboard are working, but I also moved across (and fixed type errors of) all the related dependencies (which are a lot!) and also components. I also managed to get all the tests for those files working locally, which took a lot of time replacing `MemoryRouter`!

app/web/utils/date.ts
```diff
@@ -19,16 +19,22 @@ const monthNames = [
...
-  day: "2-digit",
-});
+const dateFormatter = new Intl.DateTimeFormat(
+  typeof navigator !== "undefined" ? navigator.language : "en",
+  {
+    year: "numeric",
+    month: "short",
+    day: "2-digit",
+  }
+);
```
darrenvong: Do we want to always default to English on server renders? Would this not lead to flickers once hydrated on the client side when their actual preferred browser language is detected?

But maybe this won't be an issue/with proper usage of `react-i18next`? 🤷  Thought it's worth mentioning

app/web/tsconfig.json
```diff
@@ -1,27 +1,32 @@
 {
```
darrenvong, line 1: Do the changes here come out of the box with setting up Next (assuming they offer a similar CLI to Create React App)?

app/web/test/setupTests.ts
```diff
@@ -4,13 +4,29 @@
...
+import user from "./fixtures/defaultUser.json";
+
+jest.mock("service");
+jest.mock("next/dist/client/router", () => require("next-router-mock"));
+// Mock next/dynamic to skip the dynamic part
+// This works by extracting the require("path/to/component")
+// It needs to be in the form dynamic(() => import("components/MarkdownNoSSR"))
+// This is hacky. Really we need to just ditch any non-ssr components
+/// TODO: Get an SSR-friendly markdown editor
+jest.mock("next/dynamic", () => ({
```
darrenvong, line 23: What happens if you don't mock this? Is this a somewhat known hack used by people to set up tests? 😅

app/web/routes.ts
```diff
@@ -0,0 +1,143 @@
+/*
+The source of truth for URLs is
+//docs/urls.md
+Please make sure this file stays in sync with that file as well as
+//app/backend/src/couchers/urls.py
+*/
+// import {
+//   filtersToSearchQuery,
+//   SearchFilters,
+// } from "features/search/useSearchFilters";
```
darrenvong: Is there something wrong with search in SSR land as well? Seems unrelated to what you said that's working so far otherwise

app/web/.eslintrc.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": ["next/core-web-vitals", "prettier"],
```
darrenvong: I think this should extend `eslint-config-next` as well, as it looks like Next sets up a lot of good sane defaults there like checking for the rule of hooks and the a11y plugin enabled through that.

app/web/components/Alert.tsx
```diff
@@ -6,7 +6,7 @@ import classNames from "classnames";
 import React from "react";
 import makeStyles from "utils/makeStyles";
 
-import { grpcErrorStrings, ObscureGrpcErrorMessages } from "../constants";
+import { grpcErrorStrings, ObscureGrpcErrorMessages } from "../appConstants";
```
darrenvong:
```suggestion
import { grpcErrorStrings, ObscureGrpcErrorMessages } from "appConstants";
```
Wouldn't this also work now? I remember importing `constants` was giving me funny results in our current app with Create React App, as it seems to clash with a module named the same in Node?

app/web/components/AppRoute.tsx
```diff
@@ -0,0 +1,115 @@
+import { Container, useTheme } from "@material-ui/core";
+import classNames from "classnames";
+import CircularProgress from "components/CircularProgress";
+import CookieBanner from "components/CookieBanner";
+import ErrorBoundary from "components/ErrorBoundary";
+import Footer from "components/Footer";
+import { useAuthContext } from "features/auth/AuthProvider";
+import { useRouter } from "next/dist/client/router";
```
darrenvong:
```suggestion
import { useRouter } from "next/router";
```
Shouldn't this work too based on their example in the docs?

app/web/components/AppRoute.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  if (typeof window !== "undefined") {
+    if (!isAuthenticated && isPrivate) {
+      authActions.authError("Please log in.");
+      router.push({ pathname: loginRoute, query: { from: location.pathname } });
+    }
+    if (isAuthenticated && isJailed) {
+      router.push(jailRoute);
+    }
+  }
```
darrenvong: Do we not want to do this too in an `useEffect` still as in the original implementation? We might not need the `window` check since `useEffect` only runs on the client-side... or have you run into problems with this?

app/web/components/Button/Button.test.tsx
```diff
@@ -18,6 +18,8 @@ it("should try to log the error to Sentry if one is thrown when the button is cl
 
   userEvent.click(await screen.findByRole("button", { name: "Test button" }));
 
-  expect(testKit.reports()).toHaveLength(1);
-  expect(testKit.reports()[0]).toHaveProperty("error.message", "oops");
+  await waitFor(() => {
```
darrenvong, line 22: Interesting this needs an extra `waitFor` as I thought `userEvent.click` already waits for side-effect to flush first before following lines get executed

app/web/components/Button/Button.tsx
```diff
@@ -69,6 +72,7 @@ export default function Button<D extends ElementType = "button", P = {}>({
   return (
     <MuiButton
       {...otherProps}
+      ref={ref}
```
darrenvong, line 75: How come this now needs a `ref` passed to this?

app/web/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -1,9 +1,10 @@
 import { Typography } from "@material-ui/core";
 import Actions from "components/Actions";
 import Button from "components/Button";
+import ReportButton from "components/Navigation/ReportButton";
 import PageTitle from "components/PageTitle";
-import ReportButton from "features/ReportButton";
-import { Link, useHistory } from "react-router-dom";
+import { useRouter } from "next/dist/client/router";
```
darrenvong: Import from `next/router`? Same comment as some file above

app/web/components/ErrorFallback/ErrorFallback.tsx
```diff
@@ -42,8 +42,10 @@ export default function ErrorFallback({ isFatal }: { isFatal?: boolean }) {
 
       <Actions>
         {!isFatal && (
-          <Link to={{ pathname: baseRoute }}>
-            <Button variant="outlined">{GO_TO_HOMEPAGE}</Button>
+          <Link href={baseRoute} passHref>
+            <a>
+              <Button variant="outlined">{GO_TO_HOMEPAGE}</Button>
+            </a>
```
darrenvong: Hmm I guess the problem was already there before, but I think this is invalid HTML since `<a>` can't have more `<a>`s or `<button>`s nested.

I think we could set the `component` prop on `Button` to `a` instead? Maybe this will work without warnings now about invalid props ending up in the component since we aren't using `react-router` here anymore.

app/web/components/HtmlMeta.tsx
```diff
@@ -3,7 +3,7 @@ import {
   HTML_META_DEFAULT_TITLE,
   HTML_META_TITLE_SUFFIX,
 } from "features/constants";
-import { Helmet } from "react-helmet-async";
+import Head from "next/head";
```
darrenvong, line 6: Nice to see this come out of the box 🎉

app/web/components/MarkdownNoSSR.tsx
```diff
@@ -2,6 +2,7 @@ import "@toast-ui/editor/dist/toastui-editor-viewer.css";
 
 import ToastUIEditorViewer from "@toast-ui/editor/dist/toastui-editor-viewer";
 import classNames from "classnames";
+import dynamic from "next/dynamic";
```
darrenvong: Unused import?

app/web/components/Navigation/NavButton.tsx
```diff
@@ -0,0 +1,48 @@
+import { Typography, TypographyProps } from "@material-ui/core";
+import classNames from "classnames";
+import NotificationBadge from "components/NotificationBadge";
+import { useRouter } from "next/dist/client/router";
```
darrenvong: Same comment about importing from `next/router` as before

app/web/features/communities/EditCommunityInfoPage.tsx
```diff
@@ -106,14 +111,27 @@ export default function EditCommunityPage() {
...
+              )}
+            />
+          );
+        }
+        return (
           <>
             <HtmlMeta title={EDIT_LOCAL_INFO} />
-            <PageHeader page={community.mainPage} />
-            <CommunityPageSubHeader community={community} defaultTab="info" />
+            <PageHeader page={community.mainPage!} />
```
darrenvong: How come this only need a non-null assertion now?

app/web/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -66,36 +64,42 @@ export default function PlaceCard({
...
+                component="h3"
+                className={classes.title}
+              />
               <LinesEllipsis
-                text={contentPreview}
-                maxLine={isMdUp ? 6 : 2}
+                text={place.address}
+                maxLine={isMdUp ? 4 : 2}
                 component="p"
-                className={classes.preview}
```
darrenvong, line 94: Interesting this has the same CSS... duplicate and potentially can remove one of them in the `makeStyles(...)`?

app/web/features/communities/discussions/DiscussionsSection.tsx
```diff
@@ -47,18 +47,19 @@ export default function DiscussionsSection({
...
           "discussions"
         )}#${composingDiscussionHash}`}
       >
-        {NEW_POST_LABEL}
-      </Button>
+        <a>
+          <Button size="small" className={classes.createResourceButton}>
+            {NEW_POST_LABEL}
+          </Button>
+        </a>
```
darrenvong:
```suggestion
          <Button size="small" className={classes.createResourceButton} component="a">
            {NEW_POST_LABEL}
          </Button>
```
So the HTML is valid

app/web/features/communities/events/CreateEventPage.tsx
```diff
@@ -27,7 +28,17 @@ const useStyles = makeStyles((theme) => ({
 
 export default function CreateEventPage() {
   const classes = { ...useEventFormStyles(), ...useStyles() };
-  const history = useHistory<{ communityId?: number }>();
+  const router = useRouter();
+
+  const urlCommunityIdString =
+    typeof window !== "undefined"
+      ? stringOrFirstString(router.query["communityId"])
```
darrenvong: Are there situations where `router.query` is not consistently returning the same thing? Seems interesting we need to do the extraction ourselves...

darrenvong (changes requested):
Damn this is gonna be a big refactor 😅  not really sure how far in I am with the screwed up diffs (probably 40-50%) so will have to continue looking when I have more time.

What's the plan from this? Are we actually thinking of porting the whole app all in one go once the bumps and issues are ironed out?

How will this work for ops/CI as well? I guess instead of serving files from CloudFront in a S3 bucket, we'd be running this as a dockerised app instead and have nginx route to this... somehow? Curious as to what extra complexity it adds on that front as it looks simple for local development, but might not be the case for ops 😅

app/web/utils/date.ts
lucaslcode: I think it is indeed an issue, maybe something we can solve with middleware, but I thought better to not worry about it for this PR.

app/web/tsconfig.json
lucaslcode: Yes

app/web/features/communities/events/CreateEventPage.tsx
lucaslcode: Unlike react router, next router automatically handles `urlSearchQuery` as well as the path, so it will be an array is the url is `/example?communityId=3&communityId=3`

app/web/features/communities/discussions/DiscussionsSection.tsx
lucaslcode: Quite right. I note in this case I will need to add `passHref` to the `Link`, which is necessary when the child is not "a"

app/web/features/communities/CommunityPage/PlaceCard.tsx
lucaslcode: I think it's just the diff being weird?

app/web/features/communities/EditCommunityInfoPage.tsx
lucaslcode: I have no idea how it was working before :| mainPage in theory could always have been null...

lucaslcode: Funnily enough I don't get any lint errors if I take out the assertion. But `page` wants `page: Community.AsObject` and community.mainPage is `mainPage?: Community.AsObject`. Not sure why.

app/web/components/Navigation/NavButton.tsx
lucaslcode: Whoops this is from auto-importing

app/web/components/MarkdownNoSSR.tsx
lucaslcode: Hmm seems I'm not getting lint errors for that anymore. Will fix

lucaslcode: Maybe we should discuss what rules/plugins we want, seems like the next default has fewer than cra.

app/web/components/Button/Button.tsx
lucaslcode: Children of next `Link` need to forward refs

app/web/components/Button/Button.test.tsx
lucaslcode: Especially since this worked before, but I have gotten used to RTL and jest being mysterious/odd/annoying...

app/web/components/AppRoute.tsx
lucaslcode: Good idea

app/web/components/Alert.tsx
lucaslcode: Yes

app/web/routes.ts
lucaslcode: I think it needs a bit of extra refactoring because of the query params which I haven't done yet.

app/web/test/setupTests.ts
lucaslcode: If you don't mock it, it just renders nothing for the component. "Well-known" I'm not sure, but it came up in some googling. Really I want to get rid of it asap by removing anything that doesn't support ssr.

app/web/features/communities/events/CreateEventPage.tsx
darrenvong: Ohh so it combines both `3`s into `[3, 3]` if you repeat a URL param but as `3` literally if it only exists once?

app/web/components/AppRoute.tsx
```diff
@@ -0,0 +1,115 @@
...
+          [classes.fullWidthContainer]: variant === "full-width",
+          [classes.fullscreenContainer]: variant === "full-screen",
+          [classes.standardContainer]: variant === "standard",
+        })}
+        maxWidth={
+          variant === "full-screen" || variant === "full-width" ? false : "lg"
+        }
+      >
+        <ErrorBoundary>
+          {!isMounted && isPrivate ? (
```
darrenvong, line 92: Is this because for public routes, we'd be statically generating those pages and so it's safe to render `children` even on the server side?

app/web/features/communities/CommunityPage/PlaceCard.tsx
darrenvong: Is it? If you look at line 29-42 (unchanged so can't highlight it in PR comments...), the styles look exactly the same

app/web/features/communities/EditCommunityInfoPage.tsx
darrenvong: It's because you flipped the boolean check - before if `community.mainPage?.canEdit` is true, then `mainPage` must exist so the TypeScript type narrowing correctly inferred it can't be null.

The opposite check doesn't give that same guarantee - i.e. if `!community.mainPage?.canEdit` is false, it could be falsey because `mainPage` is undefined or `canEdit` is false, so we'd need to do another check on `mainPage` to see if it's really not null.

Couldn't we keep the old logic? Seems like it's doing the same thing as before?

app/web/features/auth/AuthProvider.tsx
```diff
@@ -1,6 +1,6 @@
 import { Error as GrpcError } from "grpc-web";
+import { useRouter } from "next/dist/client/router";
```
darrenvong:
```suggestion
import { useRouter } from "next/router";
```

app/web/features/auth/login/Login.tsx
```diff
@@ -1,14 +1,15 @@
 import { Divider, Typography } from "@material-ui/core";
+import { COUCHERS } from "appConstants";
 import classNames from "classnames";
 import StyledLink from "components/StyledLink";
+import { useRouter } from "next/dist/client/router";
```
darrenvong:
```suggestion
import { useRouter } from "next/router";
```

app/web/features/communities/CommunityBase.tsx
```diff
@@ -38,10 +38,9 @@ export default function CommunityBase({
     isLoading: isCommunityLoading,
     error: communityError,
     data: community,
-    queryCommunityId,
-  } = useCommunity(communityId);
+  } = useCommunity(communityId ?? 0);
```
darrenvong: Could we not still have the defaulting logic as part of the `useCommunity` hook as before? Too many files changed here in the GitHub diff for me to see whether the implementation of this has been changed 😅

app/web/features/communities/CommunityPage/CommunityPage.tsx
```diff
@@ -64,7 +63,7 @@ export const useCommunityPageStyles = makeStyles((theme) => ({
 export default function CommunityPage() {
   const classes = useCommunityPageStyles();
 
-  return (
+  return null; /*(
```
darrenvong: Why is this now returning nothing for the community page? 🤔

app/web/features/communities/events/EventCard.tsx
```diff
@@ -125,7 +124,7 @@ export interface EventCardProps {
 
 export default function EventCard({ event, className }: EventCardProps) {
   const classes = useStyles({
-    eventImageSrc: event.photoUrl || eventImagePlaceholder,
+    eventImageSrc: event.photoUrl || "/img/eventImagePlaceholder.svg",
```
darrenvong: Might be worth extracting the default value to an `eventImagePlaceholder` constant (rather than importing as before) since it's needed in a few places?

darrenvong (commented):
More comments/replies. Will have to continue looking when I get more time again

app/web/features/communities/CommunityPage/CommunityPage.tsx
lucaslcode: I'm not up to it yet :P

app/web/components/AppRoute.tsx
lucaslcode: yes

darrenvong:
How you feel about setting up another environment and push this under another subdomain (`beta.couchershq.org` or something) if we are planning to a full migration/refactor and eventually go across to NextJS with this? I think it'll be good to get a few of our regulars on Slack to do a bug bash first before fully replacing the client-side SPA we have right now, just so that the main bugs are mostly caught/ironed out, as I imagine it'll be quite hard to catch everything even at code review given how big of a job this is.

lucaslcode:
Good idea!

lucaslcode:
@darrenvong  this is kind of done. Some things aren't working in storybook and I need to figure out how to generate the protos on vercel but it seems to be working in general

darrenvong:
How do you see it in the Vercel preview? Or is that not fully working yet and only works locally?

darrenvong:
Also I'll try my best to review, but probably stick to more high level testing/poking around... I'm not sure it's ~~even reviewable from the web UI without getting a big lag with so many files changed/moved 🙈~~ actually seems to be alright once I give it a few mins to load, let's see

app/web/.eslintrc.json
```diff
@@ -0,0 +1,51 @@
...
+    "no-unused-vars": "off",
+    "unused-imports/no-unused-vars": [
+      "warn",
+      {
+        "argsIgnorePattern": "theme",
+        "varsIgnorePattern": "classes|useStyles"
+      }
+    ],
+    //good in theory, but ts isn't perfect and library types can be wrong
+    "@typescript-eslint/ban-ts-comment": "off",
```
darrenvong, line 36: Especially with the gRPC generated types lol 👍

app/web/components/CookieBanner.tsx
```diff
@@ -31,6 +32,9 @@ const useStyles = makeStyles((theme) => ({
 export default function CookieBanner() {
   const classes = useStyles();
   const [hasSeen, setHasSeen] = usePersistedState("hasSeenCookieBanner", false);
+  // since we are using localStorage, make sure don't render unless mounted
+  // or there will be hydration mismatches
+  const isMounted = useIsMounted().current;
```
darrenvong, line 37: With this change recently in React - https://github.com/facebook/react/pull/22114 - I thought this hook might be a bit hacky (and maybe still is, only time will tell how compatible this will be with the newer React 18 concurrent patterns) but interesting to see it become useful again in the context of SSR

app/web/components/Datepicker.test.tsx
```diff
@@ -60,16 +60,16 @@ describe("DatePicker", () => {
     const spy = jest.spyOn(global, "Date").mockImplementation(() => mockDate);
     let date: Dayjs | undefined;
     render(<Form setDate={(d) => (date = d)} />, { wrapper });
-    userEvent.click(screen.getByRole("button", { name: SUBMIT }));
+    userEvent.click(await screen.findByRole("button", { name: SUBMIT }));
 
     await waitFor(() => {
-      expect(date?.format().split("T")[0]).toEqual("2021-03-20");
+      expect(date?.format().split("T")[0]).toBe("2021-03-20");
```
darrenvong, line 66: I thought `toEqual` and `toBe` works the same when it's a primitive value like strings 🤔

app/web/components/Footer/Footer.tsx
```diff
@@ -40,25 +40,18 @@ import {
...
-  containerPadding({ maxWidth, paddingInline }: FooterProps) {
-    return {
-      maxWidth,
-      paddingInlineStart: paddingInline,
-      paddingInlineEnd: paddingInline,
-    };
+  containerPadding: {
+    maxWidth: theme.breakpoints.values.md,
+    paddingInlineStart: theme.spacing(4),
+    paddingInlineEnd: theme.spacing(4),
```
darrenvong, line 54: How do these change the look/layout of things with what's added in line 112 compared to before?

app/web/components/MarkdownNoSSR.tsx
```diff
@@ -82,20 +83,3 @@ export default function Markdown({
...
-    //loop through each header level, and add extra # as necessary
-    //also put a ~ in front so we know that line is done
-    convertedSource = convertedSource.replace(
-      new RegExp(`^${"#".repeat(i)}`, "gm"),
-      `~${"#".repeat(Math.min(i + topHeaderLevel - 1, 6))}`
-    );
-  }
-  //take out the ~ markers (line start only)
-  return convertedSource.replace(/^~#/gm, "#");
-}
```
darrenvong, line 101: Any reason not to leave this function here as it was before instead of moving it?

Slightly less change but maybe it doesn't matter in a big refactor like this anyway lol

app/web/features/auth/AuthProvider.tsx
```diff
@@ -20,26 +20,26 @@ function useAppContext<T>(context: Context<T | null>) {
 export default function AuthProvider({ children }: { children: ReactNode }) {
   const store = useAuthStore();
 
-  const push = useHistory().push;
+  const router = useRouter();
```
darrenvong: Do you happen to know if `router` is memoised/stable? Otherwise, wonder if it defeats the point of using the dependency array check in the `useEffect`.

darrenvong (commented):
Mostly questions, but looks good from what I've gone through so far. Will continue looking when I have more time

app/web/components/CookieBanner.tsx
lucaslcode: I personally don't like the fact it needs `.current`, I tripped up by omitting that and then obviously `isMounted` was always true.

app/web/components/Datepicker.test.tsx
lucaslcode: I can't remember if this was necessary or was just left behind from something else. Maybe the latter

app/web/components/Footer/Footer.tsx
lucaslcode: It doesn't spread out as much at large breakpoints, but I think it looks good and the previous way doesn't work with hydration. If we want it to look closer to before, we could add media queries here.

app/web/components/MarkdownNoSSR.tsx
lucaslcode: Yes - other files do `import { increaseMarkdownLevel } from ".......Markdown"` and it wasn't working with `next/dynamic`

app/web/features/auth/AuthProvider.tsx
lucaslcode: You're right it's not: https://github.com/vercel/next.js/issues/18127

lucaslcode: Actually,  isn't this right? Because if it's not stable, we don't want to accidentally call the "old" `router.push`?

lucaslcode:
"successfully deployed" but:
- I had to use protos from gitlab develop branch artifact - couldn't get docker working in vercel and could use branch names with / @aapeliv 
- it's trying to target production api
- it's getting cors errors

Will fix the last two later.

app/web/.storybook/i18n.ts
```diff
@@ -8,18 +8,7 @@ i18n
...
-            callback(null, resources);
-          })
-          .catch((error) => {
-            callback(error, null);
-          });
-        return;
-      }
-
-      import(`resources/locales/${language}.json`)
+      import(`public/locales/${language}/${namespace}.json`)
```
darrenvong: So I'm guessing this is still fine to be used for tests/Storybook stuff? Maybe worth moving the `i18next-resources-to-backend` dependency as a dev dependency then to reduce the final app bundle size a little

app/web/features/auth/AuthProvider.tsx
darrenvong: Yeah, but then we run into that infinite loop issue I think you faced in that other `useEffect` in some cases so seems like a bug to me? It's unclear what bad things would happen if we call the "old" `router.push`... I'd assume nothing since it doesn't have router state like react-router? 🤷

app/web/components/Datepicker.test.tsx
```diff
@@ -26,7 +26,7 @@ const Form = ({ setDate }: { setDate: (date: Dayjs) => void }) => {
   );
 };
 
-describe("DatePicker", () => {
+describe.skip("DatePicker", () => {
```
darrenvong, line 29: Is this meant to be skipped still? Does that mean the DatePicker doesn't work well when server rendered?

app/web/dl-protos.sh
```diff
@@ -0,0 +1,3 @@
+curl -L https://gitlab.com/couchers/couchers/-/jobs/artifacts/develop/download\?job\=protos -o /tmp/protos.zip
```
darrenvong, line 1: Is this only for the preview environment? Or will we need it for production too?

app/web/features/auth/signup/Signup.tsx
```diff
@@ -178,13 +179,15 @@ export default function Signup() {
           authActions.authError(
             isGrpcError(err) ? err.message : t("global:fatal_error_message")
           );
-          history.push(signupRoute);
+          router.push(signupRoute);
```
darrenvong, line 183: Ahh this was what I'm trying to refer to in `AuthProvider` - I'm guessing since we ignored `router` as a dependency here and it's still fine (?), maybe it doesn't matter if we call the "old" `router.push`?

app/web/features/communities/PagePage.tsx
```diff
@@ -5,37 +5,46 @@ import HtmlMeta from "components/HtmlMeta";
...
 import { Page, PageType } from "proto/pages_pb";
 import React, { useEffect, useState } from "react";
-import { useTranslation } from "react-i18next";
-import { useHistory, useParams } from "react-router-dom";
+import { routeToGuide, routeToPlace } from "routes";
 import { service } from "service";
 import isGrpcError from "utils/isGrpcError";
 
-export default function PagePage({ pageType }: { pageType: PageType }) {
+export default function PagePage({
```
darrenvong, line 16: Lol I hope we rename this "model" in some way so we don't have this `PagePage` confusion...

app/web/features/communities/discussions/DiscussionsListPage.tsx
```diff
@@ -45,7 +44,7 @@ export default function DiscussionsListPage({
     ...useDiscussionsListStyles(),
     ...useStyles(),
   };
-  const hash = useLocation().hash;
+  const hash = typeof window !== "undefined" ? window.location.hash : "";
```
darrenvong, line 47: Does it not matter this defaults to `""` since it will be behind a loading spinner when rendered server side? How does it deal with the change of receiving `#new` once it's at client side?

app/web/features/communities/events/CreateEventPage.tsx
darrenvong:
```suggestion
      ? stringOrFirstString(router.query.communityId)
```
Minor thing tbh

app/web/features/communities/events/EventPage.tsx
```diff
@@ -51,7 +51,9 @@ export const useEventPageStyles = makeStyles<Theme, { eventImageSrc: string }>(
       },
       width: "100%",
       objectFit: ({ eventImageSrc }) =>
-        eventImageSrc === eventImagePlaceholder ? "contain" : "cover",
+        eventImageSrc === "/img/eventImagePlaceholder.svg"
```
darrenvong:
```suggestion
        eventImageSrc === eventImagePlaceholderUrl
```
Could use that from `appConstants`?

app/web/features/communities/events/EventPage.tsx
```diff
@@ -185,12 +191,12 @@ export default function EventPage() {
...
     if (event?.slug && event.slug !== eventSlug) {
-      history.replace(routeToEvent(event.eventId, event.slug));
+      router.replace(routeToEvent(event.eventId, event.slug));
     }
-  }, [event, eventSlug, history]);
+  }, [event, eventSlug, router]);
 
   const classes = useEventPageStyles({
-    eventImageSrc: event?.photoUrl || eventImagePlaceholder,
+    eventImageSrc: event?.photoUrl || "/img/eventImagePlaceholder.svg",
```
darrenvong: Same comment as above about using the `eventImagePlaceholderUrl` constant

app/web/features/communities/events/EventPage.tsx
```diff
@@ -210,13 +216,13 @@ export default function EventPage() {
           <>
             <img
               className={classes.eventCoverPhoto}
-              src={event.photoUrl || eventImagePlaceholder}
+              src={event.photoUrl || "/img/eventImagePlaceholder.svg"}
```
darrenvong: And here

app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -60,7 +59,7 @@ const useStyles = makeStyles<Theme, { eventImageSrc: string }>((theme) => ({
   },
   image: {
     objectFit: ({ eventImageSrc }) =>
-      eventImageSrc === eventImagePlaceholder ? "contain" : "cover",
+      eventImageSrc === "/img/eventImagePlaceholder.svg" ? "contain" : "cover",
```
darrenvong: Use `eventImagePlaceholderUrl` constant here too?

app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -83,7 +82,7 @@ export interface LongEventCardProps {
 
 export default function LongEventCard({ event }: LongEventCardProps) {
   const classes = useStyles({
-    eventImageSrc: event.photoUrl || eventImagePlaceholder,
+    eventImageSrc: event.photoUrl || "/img/eventImagePlaceholder.svg",
```
darrenvong: And here as above

app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -100,40 +99,41 @@ export default function LongEventCard({ event }: LongEventCardProps) {
...
-          className={classes.image}
-          src={event.photoUrl || eventImagePlaceholder}
-        />
+          <Typography className={classes.content} variant="body1">
+            {truncatedContent}
+          </Typography>
+          <img
+            alt=""
+            className={classes.image}
+            src={event.photoUrl || "/img/eventImagePlaceholder.svg"}
```
darrenvong: And here

app/web/features/connections/friends/FriendTile.tsx
```diff
@@ -22,7 +22,7 @@ const useStyles = makeStyles((theme) => ({
     fontWeight: theme.typography.fontWeightBold,
     marginBottom: theme.spacing(2),
     marginLeft: theme.spacing(1),
-  },
+  } as CSSProperties,
```
darrenvong, line 25: Strange that this suddenly cause TS errors 🤔

app/web/pages/messages/[[...slug]].tsx
```diff
@@ -0,0 +1,62 @@
...
+  const router = useRouter();
+  const slugs =
+    typeof router.query.slug === "undefined"
+      ? ["chats"]
+      : typeof router.query.slug === "string"
+      ? [router.query.slug]
+      : router.query.slug;
+  let content: ReactNode;
+
+  const tab = messageTypeStrings.find((valid) => valid === slugs?.[0]);
```
Comment (darrenvong) [inline app/web/pages/messages/[[...slug]].tsx:39]:
```suggestion
  const tab = messageTypeStrings.find((valid) => valid === slugs[0]);
```

app/web/pages/messages/[[...slug]].tsx
```diff
@@ -0,0 +1,62 @@
...
+      content = <GroupChatsTab />;
+    } else {
+      content = <GroupChatView chatId={id} />;
+    }
+  } else if (slugs[0] === "request") {
+    const id = Number.parseInt(slugs?.[1]);
+    if (isNaN(id)) {
+      return <NotFoundPage />;
+    } else {
+      content = <HostRequestView hostRequestId={id} />;
```
Comment (darrenvong) [inline app/web/pages/messages/[[...slug]].tsx:53]: In the old switch logic, the `HostRequestView` doesn't have the `MessageHeader` rendered (probably why `header` is horribly repeated, I kinda like this a bit better lol). Do we now want it instead?

app/web/features/messages/groupchats/CreateGroupChat.tsx
```diff
@@ -136,31 +133,54 @@ export default function CreateGroupChat({ className }: { className?: string }) {
...
-                  />
-                );
-              }}
-            />
+            {createMessageToUserQuery.error && (
+              <Alert severity="error">{createMessageToUserQuery.error}</Alert>
+            )}
+            {
+              // need to mount the autocomplete with the correct default value
+              // of the "to" user, display a dummy loader until then
```
darrenvong, line 141: So this is effectively bailing out on the server side and let the hydrated app deal with it later in client side?

app/web/features/profile/view/leaveReference/ReferenceForm.tsx
```diff
@@ -85,36 +89,34 @@ export default function ReferenceForm() {
...
-    path: `${leaveReferenceBaseRoute}/:referenceType/:userId/:hostRequestId/:step?`,
-  });
-
-  const step =
-    friendMatch?.params.referenceType ===
-    referenceTypeRoute[ReferenceType.REFERENCE_TYPE_FRIEND]
-      ? friendMatch?.params.step
-      : hostingMatch?.params.step;
-
   return step === undefined ? (
```
darrenvong: Given the pattern I've seen your refactor so far, I think this is kinda gross, as it's mixing routing concerns within the "page" component itself which should really... be responsible for the page itself only.

Maybe make an issue about this as a low priority refactor? It's a big change here already so I think it's fine to leave it here and do separately if we decide to later.

app/web/features/profile/view/leaveReference/formSteps/submit/SubmitReference.tsx
```diff
@@ -92,25 +99,29 @@ export default function SubmitReference({
       },
       {
         onSuccess: () => {
-          history.push(routeToUser(user.username, "references"));
+          if (userQuery.data) {
+            router.push(routeToUser(userQuery.data.username, "references"));
+          } else {
+            router.push(routeToProfile("references"));
```
darrenvong, line 105: Would this have been stuck on the end of leave references page before? 🤔

app/web/features/search/FilterDialog.tsx
```diff
@@ -117,8 +117,16 @@ export default function FilterDialog({
   // This requirement for certain filters to have a location specified
   // should be removed when we show users according to bounding box
   // or have some other solution to the pagination issue #1676
-  const validateHasLocation = (data: any) => {
-    if (!data || data.length === 0 || data === "" || data.value === null)
+  const validateHasLocation = (
+    data: string | number | string[] | number[] | { value: null | unknown }
+  ) => {
+    if (!data) return true;
```
darrenvong, line 123: I guess this also covers `data === ""` from before since "" is falsey?

app/web/features/search/LocationAutocomplete/LocationAutocomplete.stories.tsx
```diff
@@ -29,8 +29,8 @@ const Template: Story<any> = (args) => {
         params={urlParams.current}
       />
       <p>
-        Pressing enter in the field shouldn't perform a submit action, but the
-        button should.
+        Pressing enter in the field shouldn&apos;t perform a submit action, but
```
darrenvong, line 32: What happens on the server side without converting the character to an escape sequence?

app/web/next-i18next.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  i18n: {
+    defaultNS: "global",
+    defaultLocale: "en",
+    locales: ["en", "fr"],
```
darrenvong: Not sure if we should have "fr" in here until we have a decent amount of the app translated

app/web/package.json
```diff
@@ -1,155 +1,96 @@
 {
-  "name": "couchers-react",
+  "name": "web",
   "version": "0.1.0",
   "private": true,
+  "scripts": {
+    "build-storybook": "build-storybook",
+    "build": "next build",
+    "dev": "next dev",
+    "start": "yarn dev",
```
darrenvong, line 9:
```suggestion
    "start": "next dev",
```
We could combine these to mimic `start` to be starting the development server like before with CRA?

app/web/package.json
```diff
@@ -1,155 +1,96 @@
...
     "@testing-library/react": "^12.1.2",
     "@testing-library/react-hooks": "^7.0.2",
-    "@testing-library/user-event": "^13.5.0",
+    "@testing-library/user-event": "^13.4.1",
     "@types/css-mediaquery": "^0.1.1",
     "@types/geojson": "^7946.0.8",
     "@types/google-protobuf": "^3.15.5",
     "@types/jest": "^26.0.24",
-    "@types/node": "^14.18.0",
+    "@types/node": "16.11.6",
```
darrenvong: Have we upgraded to Node 16 on Vercel?

Also, loosely related, do we not need a Dockerfile for this anymore?

darrenvong (commented):
Don't think there was anything major and so mostly questions here and there having gone through the boring stuff. But boring is good, as that means we kept most of the React code fairly intact, so nice job on that!

Only looked through the `pages` / routing stuff on an ad-hoc (as I mostly went from top to bottom to stop GitHub crapping out on this large diff), but I noticed now the routing logic has moved there instead of some of the `Switch` stuff we had with React Router. Might it be worth having unit test for that logic? Or is it gonna be a pain since it's bundled within Next specific components now?

app/web/features/auth/AuthProvider.tsx
lucaslcode: Added a workaround taken from that bug thread which allows to use the latest router.push but doesn't trigger useEffect. I also found the `t` dependency in `useAuthStore` was causing that useEffect to fire too often too, so fixed that.

app/web/.storybook/i18n.ts
lucaslcode: it is!

app/web/dl-protos.sh
lucaslcode: Yeah still figuring this bit out

app/web/features/auth/signup/Signup.tsx
lucaslcode: This case was for the `next-router-mock` router, not the real one. But see my other comment

lucaslcode:
Yeah, could use this to test pages: https://github.com/toomuchdesign/next-page-tester

app/web/components/Datepicker.test.tsx
lucaslcode: I could not for the life of me get this to work - as is currently, it hangs forever. I think it's to do with timezone mock and mocking the `Date` implementation, but I never got it working.

Maybe we should follow your advice and ditch this anyway for native pickers...

app/web/features/communities/discussions/DiscussionsListPage.tsx
lucaslcode: Yes, the server just renders the header, spinner, and footer; and initially, so does the client.

app/web/features/connections/friends/FriendTile.tsx
lucaslcode: It's because of fontWeight, but not sure why

app/web/pages/messages/[[...slug]].tsx
Comment (lucaslcode) [reply app/web/pages/messages/[[...slug]].tsx:39]: It might not have any elements (`/messages/`

app/web/pages/messages/[[...slug]].tsx
Comment (lucaslcode) [reply app/web/pages/messages/[[...slug]].tsx:53]: Does it look odd? It's easier to get around.

app/web/features/messages/groupchats/CreateGroupChat.tsx
lucaslcode: No (well, yes, that data isn't in the server render), it's just before the whole `createMessageTo` user was being passed in the router state, now we have to load from the id in the url. But you can't mount the `Control` and change the default value once the user data is loaded, so I'm actually starting with a dummy loading autocomplete (L174) and swapping for the proper one with a `Control` and `defaultValue` once the data is loaded.

app/web/features/profile/view/leaveReference/ReferenceForm.tsx
lucaslcode: Good idea, I think especially now there is a nice place to keep all routing logic.

app/web/features/profile/view/leaveReference/formSteps/submit/SubmitReference.tsx
lucaslcode: I think so yes.

app/web/features/search/LocationAutocomplete/LocationAutocomplete.stories.tsx
lucaslcode: Probably nothing but there was a new lint rule complaining. Maybe I should just disable it.

app/web/next-i18next.config.js
lucaslcode: That was mostly for me to test :P

app/web/package.json
lucaslcode: I kept both so people used to the old way can use start, but people who know Next uses dev can use that.

app/web/package.json
lucaslcode: No, not sure why I changed that

app/web/components/Datepicker.test.tsx
darrenvong: Ah yeah this `datetime` mock seems pretty buggy and I never got it working outside of these tests... well I guess it just doesn't work if they start failing here too.

app/web/pages/messages/[[...slug]].tsx
Comment (darrenvong) [reply app/web/pages/messages/[[...slug]].tsx:39]: Wouldn't `slugs` be an empty array in that case? I checked out the branch, changed it locally and TS seems to think it's always an array. Is that not true?

app/web/features/search/LocationAutocomplete/LocationAutocomplete.stories.tsx
darrenvong: I guess it doesn't matter that much as we should be moving these messages into JSON files too as part of the i18n work, so probably fine to leave it for now

app/web/features/messages/groupchats/CreateGroupChat.tsx
darrenvong: Could we not do the data loading of the user on server side too, since Next will have access to the id from the URL parameter? Or maybe too much change to this otherwise even if it works?

app/web/features/auth/useAuthStore.ts
```diff
@@ -161,14 +162,14 @@ export default function useAuthStore() {
...
             },
           });
-          setError(isGrpcError(e) ? e.message : t("fatal_error_message"));
+          setError(isGrpcError(e) ? e.message : fatalErrorMessage.current);
         }
         setLoading(false);
       },
     }),
-    //note: there should be no dependenices on the state, or
+    //note: there should be no dependenices on the state or t, or
```
darrenvong, line 170: That might explain why I run into funky issues where all of the tests failed because of this! Kinda disappointing frameworks don't think about the importance of keeping their function stable if possible, especially as this is implemented with hooks so you'd thought they expect people to use this in an `useEffect` in some places...

app/web/package.json
darrenvong: Ahh fair, guess it's not really a big deal, just curious why we repeat the command but the convention from the Next docs makes sense

darrenvong:
I had a quick play around and ran into an infinite refresh loop in jail where the page keeps flashing without letting me accept the terms/interact with the page with an existing account locally.

app/web/utils/date.ts
```diff
@@ -19,24 +19,30 @@ const monthNames = [
   "December",
 ];
 
-const dateTimeFormatter = new Intl.DateTimeFormat(navigator.language, {
-  month: "short",
-  year: "numeric",
-});
+const dateTimeFormatter = new Intl.DateTimeFormat(
+  typeof navigator !== "undefined" ? navigator.language : "en",
```
darrenvong: Since we now learnt that Next pass the locale through internationalised routing, might we be able to restructure this so it takes that into consideration? Maybe one for a follow-up PR if you don't want to add to this diff

app/web/components/CookieBanner.tsx
darrenvong: Could probably change the `useIsMounted` hook to return `.current` within the hook result, but then I think there'll be a few places that use the `useSafeState` hook will need updating. Based on the linked issue in the React repo, I think the `useSafeState` should probably be removed and swapped to a normal `useState` instead, as they acknowledged that warning is mostly a false positive if they have remove it.

app/web/features/auth/signup/Signup.tsx
darrenvong: Shouldn't we be using the workaround here then, rather than leaving it out of the dependency array?

app/web/pages/group/[id]/[slug].tsx
```diff
@@ -0,0 +1,35 @@
...
+  props: {
+    ...(await serverSideTranslations(
+      locale ?? "en",
+      ["global", "communities"],
+      nextI18nextConfig
+    )),
+  },
+});
+
+export default function GroupPage() {
```
Comment (darrenvong) [inline app/web/pages/group/[id]/[slug].tsx]: We should probably hide this behind the `is_communities_part2_enabled` feature flag as before and return 404 as these pages aren't really ready yet? Especially now this is gonna be server rendered, not sure we want the search engine crawlers to be indexing broken pages 😅

app/web/pages/guide/[id]/[slug].tsx
```diff
@@ -0,0 +1,42 @@
...
+  props: {
+    ...(await serverSideTranslations(
+      locale ?? "en",
+      ["global", "communities"],
+      nextI18nextConfig
+    )),
+  },
+});
+
+export default function PagePage() {
```
Comment (darrenvong) [inline app/web/pages/guide/[id]/[slug].tsx]: Same comment as in `GroupPage`

app/web/features/profile/view/leaveReference/LeaveReferencePage.tsx
```diff
@@ -76,15 +78,15 @@ export default function LeaveReferencePage() {
           user.friends === User.FriendshipStatus.FRIENDS) ||
         (hostRequestId &&
           availableRefrences.availableWriteReferencesList.find(
```
darrenvong:
```suggestion
          availableReferences.availableWriteReferencesList.find(
```
Ever since I installed the spell checker plugin, this typo has been annoying me :p so I think we should fix it here

app/web/features/messages/MessagesHeader.tsx
```diff
@@ -0,0 +1,85 @@
...
+  //all: "All",
+  chats: <MessagesNotification />,
+  hosting: <HostRequestsReceivedNotification />,
+  surfing: <HostRequestsSentNotification />,
+  //meet: "Meet",
+  //archived: "Archived",
+};
+
+export default function MessagesHeader({
+  tab,
```
darrenvong:
```suggestion
  tab = "chats",
```
So it defaults to the first tab when the given route is `/messages` like before (also similar to what you've done for the communities page tabs)

app/web/pages/place/[id]/[slug].tsx
```diff
@@ -0,0 +1,41 @@
...
+export const getStaticProps: GetStaticProps = async ({ locale }) => ({
+  props: {
+    ...(await serverSideTranslations(
+      locale ?? "en",
+      ["global", "communities"],
+      nextI18NextConfig
+    )),
+  },
+});
+export default function PagePage() {
```
Comment (darrenvong) [inline app/web/pages/place/[id]/[slug].tsx]: Hide behind feature flag and return 404 here too?

app/web/pages/profile/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+    ...(await serverSideTranslations(
+      locale ?? "en",
+      ["global", "profile"],
+      nextI18nextConfig
+    )),
+  },
+});
+
+export default function ProfilePage() {
+  return <ProfilePageComponent tab="about" />;
```
darrenvong:
```suggestion
  return <ProfilePageComponent />;
```
Wonder if it's better to do this and leave the defaulting to "about" on `ProfilePage` instead, since that's how it's done in the community page. Don't think I have a strong preference either way, as long we do it consistently across all tabbed pages.

app/web/pages/user/[[...slug]].tsx
```diff
@@ -0,0 +1,42 @@
...
+  if (isNaN(parsedId)) return <NotFoundPage />;
+  //first element of params is the username
+  const username = stringOrFirstString(router.query.params);
+  if (!username) return <NotFoundPage />;
+  const tab = router.query.params?.[1];
+  const parsedTab = tab ? userTabs.find((valid) => tab === valid) : null;
+  //null = not found, undefined = blank
+  if (!parsedTab && parsedTab !== undefined) return <NotFoundPage />;
+
+  return <UserPageComponent username={username} tab={parsedTab ?? "about"} />;
```
Comment (darrenvong) [inline app/web/pages/user/[[...slug]].tsx]: Same default comment as in `ProfilePage`

app/web/service/client.ts
```diff
@@ -88,7 +91,7 @@ const client = {
   threads: new ThreadsPromiseClient(URL, null, opts),
 };
 
-if (process.env.REACT_APP_COUCHERS_ENV !== "prod") {
+/*if (process.env.NEXT_PUBLIC_COUCHERS_ENV !== "prod") {
```
darrenvong: Does the dev tool break with SSR?

app/web/test/hookWrapper.tsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+  });
+  const wrapper = ({ children }: { children?: React.ReactNode }) => (
+<Suspense fallback="loading..."><ThemeProvider theme={theme}>
+      <QueryClientProvider client={client}>
+        <AuthProvider>{children}</AuthProvider>
+      </QueryClientProvider>
+    </ThemeProvider>
+    </Suspense>
+  );
```
darrenvong: Might need to format this to pass Prettier on CI eventually

app/web/test/utils.tsx
```diff
@@ -31,6 +31,7 @@ export function wait(milliSeconds: number) {
   return new Promise<void>((resolve) => setTimeout(resolve, milliSeconds));
 }
 
+//eslint-disable-next-line
```
darrenvong: What was this complaining about?

darrenvong (changes requested):
Probably the final round of code comments/questions! Will try and have a general poke around all the routes next

app/web/package.json
```diff
@@ -1,155 +1,99 @@
...
+    "typescript": "~4.5.4"
   },
-  "reactSnap": {
-    "puppeteerArgs": [
-      "--no-sandbox",
-      "--disable-setuid-sandbox",
-      "--disable-dev-shm-usage"
-    ]
+  "resolutions": {
+    "next-i18next": "git+https://github.com/Couchers-org/next-i18next.git#39a50ec9"
```
darrenvong: Why is `next-i18next` pinned to a specific commit version?

app/web/pages/_app.tsx
```diff
@@ -0,0 +1,63 @@
...
+      Sentry.init({
+        dsn: "https://5594adb1a53e41bfbb9f2cc5c91e2dbd@o782870.ingest.sentry.io/5887585",
+        environment: process.env.NEXT_PUBLIC_COUCHERS_ENV,
+        release: process.env.NEXT_PUBLIC_VERSION,
+      });
+      TagManager.initialize({ gtmId: "GTM-PXP3896" });
+    }
+  }, []);
+
+  return (
```
darrenvong, line 47: No need for `Suspense` anymore for components that are loading?

app/web/pages/_document.tsx
```diff
@@ -0,0 +1,67 @@
+import { ServerStyleSheets } from "@material-ui/core/styles";
+import Document, { Head, Html, Main, NextScript } from "next/document";
+import { Children } from "react";
+import { theme } from "theme";
+
+export default class MyDocument extends Document {
+  render() {
+    return (
+      <Html lang="en">
```
darrenvong: We probably don't want this to always be `en` here - it should change depend on what locale is loaded?

app/web/pages/group/[id]/[slug].tsx
Comment (darrenvong) [inline app/web/pages/group/[id]/[slug].tsx:25]: We should probably hide this behind the `is_communities_part2_enabled` feature flag as before and return 404 as these pages aren't really ready yet? Especially now this is gonna be server rendered, not sure we want the search engine crawlers to be indexing broken pages 😅

app/web/pages/guide/[id]/[slug].tsx
Comment (darrenvong) [inline app/web/pages/guide/[id]/[slug].tsx:26]: Same comment as in `GroupPage`

app/web/features/profile/view/leaveReference/LeaveReferencePage.tsx
```diff
@@ -36,24 +35,27 @@ const useStyles = makeStyles((theme) => ({
...
-  }>();
 
   const {
     data: user,
     isLoading: isUserLoading,
     error: userError,
-  } = useUser(+userId);
+  } = useUser(userId);
   const {
     data: availableRefrences,
```
darrenvong:
```suggestion
    data: availableReferences,
```
Ever since I installed the spell checker plugin, this typo has been annoying me 😛 so I think we should fix it here

app/web/features/profile/view/leaveReference/LeaveReferencePage.tsx
darrenvong:
```suggestion
          availableReferences.availableWriteReferencesList.find(
```
And on line 73 (can't highlight because of no change)

app/web/features/messages/MessagesHeader.tsx
darrenvong, line 62:
```suggestion
  tab = "chats",
```
So it defaults to the first tab when the given route is `/messages` like before (also similar to what you've done for the communities page tabs)

app/web/pages/place/[id]/[slug].tsx
Comment (darrenvong) [inline app/web/pages/place/[id]/[slug].tsx:25]: Hide this behind the community feature flag too and return 404

app/web/pages/profile/index.tsx
darrenvong:
```suggestion
  return <ProfilePageComponent />;
```
I wonder if it's better to do this and leave the defaulting to "about" on `ProfilePage` instead, since that's how it's done in the community page. Don't think I have a strong preference either way, as long we do it consistently across all tabbed pages.

app/web/pages/user/[[...slug]].tsx
Comment (darrenvong) [inline app/web/pages/user/[[...slug]].tsx]: Same default comment as in `ProfilePage` on doing it on the page component instead?

app/web/service/client.ts
darrenvong: Does the dev tool break with SSR?

app/web/test/hookWrapper.tsx
darrenvong: Might need to format this to pass Prettier on CI eventually

app/web/test/utils.tsx
darrenvong, line 34: What was this complaining about?

darrenvong (changes requested):

app/web/package.json
lucaslcode: we are waiting for PR to be merged upstream

app/web/features/messages/MessagesHeader.tsx
lucaslcode: The reason it is like this is because when viewing a host request, the url is `/messages/requests/x` so we don't know if it's a surfing or hosting one. If I make this change, it will show "chats".

Really we should change it to `/messages/hosting/X` and not care about hosting/surfing when displaying the request, but let's do that in another PR.

app/web/test/utils.tsx
lucaslcode: the use of `any` in the next line

lucaslcode: use of `any`

darrenvong:
Found some more things from some testing:
- the banner background on the donation page doesn't seem to stretch the whole width of the page anymore like it was before:
Screenshot 2021-12-29 195551

- Terms of Service is incorrectly wrapped with a header even for unauthenticated users. Clicking any of the links correctly redirects back to `/login` though so that's good, but should still fix since we shouldn't be showing things people have no access to.
Screenshot 2021-12-29 200144

- User page is broken - gets taken to 404

app/web/pages/user/[[...slug]].tsx
```diff
@@ -0,0 +1,44 @@
...
+  },
+});
+
+export default function UserPage() {
+  const router = useRouter();
+
+  const parsedId = Number.parseInt(stringOrFirstString(router.query.id) ?? "");
+  if (isNaN(parsedId)) return <NotFoundPage />;
+  //first element of params is the username
+  const username = stringOrFirstString(router.query.params);
```
Comment (darrenvong) [inline app/web/pages/user/[[...slug]].tsx]:
```suggestion
  const username = stringOrFirstString(router.query.slug);
```
I think this is why it's breaking for all user pages

lucaslcode:
Freshly rebased, this just needs CI/devops work now @aapeliv

app/.gitlab-ci.yml
```diff
@@ -275,9 +298,9 @@ test:media:
       - app/media/**/*
 
 test:web:
-  needs: ["build:web"]
+  needs: ["build:web-dev"]
   stage: test
-  image: $WEB_TAG
+  image: $WEB_DEV_TAG
```
aapeliv, line 301: Why did you switch the web tests to run in the dev container and not the prod one?

app/.gitlab-ci.yml
```diff
@@ -130,15 +132,59 @@ build:media:
...
+    paths:
+      - artifacts/web/
+
+build:web-next:
+  needs: ["protos"]
+  stage: build
+  script:
+    - docker pull $WEB_NEXT_RELEASE_TAG || true
+    # this is a bit hacky, but I don't see any other good way to get the preview env vars into the container
+    - cp app/web/.env.preview app/web/.env.production
```
aapeliv: You can do this in the dockerfile. This is how the `nginx` containers picks production or preview. I think it's cleaner to do it there.

Short version:

1. In dockerfile

```dockerfile
# 'production' or 'preview'
ARG environment=production

COPY .env.${environment} .env
```

2.
Add docker build arg: `--build-arg environment=preview`

app/web/README.md
```diff
@@ -1,6 +1,8 @@
-# couchers-react
+# couchers-next
```
aapeliv:
```suggestion
# Couchers web frontend
```

aapeliv (commented):
I reviewed the non-TS bits, looks very good overall, see small comments

app/.gitlab-ci.yml
lucaslcode: The production container doesn't have any source code or development dependencies

aapeliv: Ohh true, good point!

aapeliv (dismissed):
LGTM for the bits I reviewed!

darrenvong (changes requested):
I think the Prettier command is broken since we don't have a `src` folder anymore.

Also some tests are failing for me locally because `fetch is not defined`? 🤔 
Edit: Tests are hanging for me locally unless I disable watch mode and use `--runInBand`, pretty strange...

darrenvong:
> User page is broken - gets taken to 404

This is still broken for me... locally I'm getting "no user id available to get current user" in the console

app/web/pages/user/[[...slug]].tsx
```diff
@@ -0,0 +1,44 @@
...
+      nextI18nextConfig
+    )),
+  },
+});
+
+export default function UserPage() {
+  const router = useRouter();
+
+  const parsedId = Number.parseInt(stringOrFirstString(router.query.id) ?? "");
+  if (isNaN(parsedId)) return <NotFoundPage />;
```
Comment (darrenvong) [inline app/web/pages/user/[[...slug]].tsx]:
```suggestion
```
Looks like this is the problematic part since we don't have `id` explicitly defined in the URL pattern matcher in the filename. And if we want to support numeric ID for getting to user profiles like `/user/2`, that'd be covered by `slug` too. So we should probably remove this

Comment (lucaslcode) [reply app/web/pages/user/[[...slug]].tsx]: Whoops, that was bad copy pasting

lucaslcode:
Turns out CRA creates a fetch mock for you by itself.
What is weird is that the test CI didn't have any failing tests, but still had exit code 1???

Anyway, I locally now have 0 warnings/console errors when running tests. 🏋🏻

app/web/pages/leave-reference/[[...slug]].tsx
```diff
@@ -0,0 +1,48 @@
...
+      nextI18nextConfig
+    )),
+  },
+});
+export default function LeaveReferencePage() {
+  const router = useRouter();
+
+  // leave-reference/:type/:userId/:hostRequestId?
+  const slug = router.query.slug;
+  if (!slug?.[0] || slug?.[1]) return <NotFoundPage />;
```
Comment (darrenvong) [inline app/web/pages/leave-reference/[[...slug]].tsx]:
```suggestion
  if (!slug?.[0] || !slug?.[1]) return <NotFoundPage />;
```
Otherwise this was sending valid `userId` to the not found page

Comment (darrenvong) [reply app/web/pages/leave-reference/[[...slug]].tsx]: But beyond that, this whole reference leaving thing is broken after I changed the above :/ And most of it is because of the routing mess... we're probably not making the logic here any easier by having `step` as `slug[3]` if it's a friend reference but `slug[4]` if it's a host/surfer reference.

I think the URL should reference the request ID if this is available tbh so it can be unified as `/leave-reference/[reference-id]/[step]`, where `step` should always exist rather than weirdly being optional/undefined when it's the first step (and so more edge case to deal with, convoluting the logic more...). Maybe a future thing to change unless fixing that is not enough? Haven't delved deeper beyond that.

app/web/pages/messages/[[...slug]].tsx
```diff
@@ -0,0 +1,67 @@
...
+    }
+  } else if (slugs[0] === "hosting") content = <RequestsTab type="hosting" />;
+  else if (slugs[0] === "surfing") content = <RequestsTab type="surfing" />;
+  else if (slugs[0] === "surfing") content = <RequestsTab type="surfing" />;
```
Comment (darrenvong) [inline app/web/pages/messages/[[...slug]].tsx]:
```suggestion
```
Duplicate code from the above

app/web/pages/messages/[[...slug]].tsx
Comment (darrenvong) [reply app/web/pages/messages/[[...slug]].tsx:53]: Had another look at this and I wonder if that was why we hid the tab bar in some places, as I am getting this error/warning in the console now:
Screenshot 2022-01-05 001640

Problem seems to be that when we click requests from either the surfing or hosting tab, it takes the user to `/messages/request/[request-id]` and so we have no way of knowing whether "request" should activate the "hosting" or "surfing" tab...

app/web/features/messages/groupchats/CreateGroupChat.tsx
```diff
@@ -136,31 +133,54 @@ export default function CreateGroupChat({ className }: { className?: string }) {
                 className={classes.field}
               />
             )}
```
darrenvong, line 135: I find that this create chat dialog is now always starting with an error message even when I've just opened it! Probably caused by when `to` doesn't exist in the URL param...
Screenshot 2022-01-05 002324

darrenvong (changes requested):
Found a few more issues as I didn't test those before when the user thing was broken😅

darrenvong:
> What is weird is that the test CI didn't have any failing tests, but still had exit code 1???

I saw that, very odd! But now CI seems to be green again?

Also, how come the Vercel build is failing to deploy again? 🤔

lucaslcode:
I am thinking about just ignoring that warning about the tabs. I think the behaviour is correct! What do you think?

darrenvong:
> I think the behaviour is correct!

Yeah, otherwise I think a random white space will get highlighted if we add the empty string to the list of possible tabs, which would probably look more weird. I'm not sure anything bad would happen apart from no tabs get highlighted, which seems fine to me.

Only thing I'm less sure of is having the "Messages" title displayed even for host requests, as that might be a bit confusing as those aren't just messages? But maybe it's not worth nitpicking over in this refactor if we are planning to re-do requests separately at some point anyway

app/web/features/profile/view/leaveReference/LeaveReferencePage.tsx
```diff
@@ -68,23 +72,27 @@ export default function LeaveReferencePage() {
...
               <Hidden smDown>
                 <UserOverview />
               </Hidden>
               <div className={classes.form}>
-                <ReferenceForm />
+                <ReferenceForm
+                  referenceType={referenceType}
+                  userId={userId}
+                  step={step}
+                />
```
darrenvong:
```suggestion
                <ReferenceForm
                  hostRequestId={hostRequestId}
                  referenceType={referenceType}
                  userId={userId}
                  step={step}
                />
```
Otherwise host/surfer references are still broken

darrenvong (changes requested):
Should be the last thing hopefully!

darrenvong:
Have you come across this "unhandled runtime error" whenever you're using the Toast UI editor? Initially I thought this was caused by React when seeing a bunch of similar errors on Sentry, but now the stacktrace here quite clearly points to the editor, so that seems to be the actual issue...
Screenshot 2022-01-06 215759

darrenvong:
Another thing - Next seems to put things in a `.next` folder rather than `build`, so not sure if the existing Sentry integration would work anymore. Looking at [their setup guide](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/), it looks quite different.

It might be worth running through [their config wizard](https://docs.sentry.io/platforms/javascript/guides/nextjs/#configure) that supposedly set up a few things automatically? I think we'll miss some routing errors otherwise with how we set up at the moment.

app/web/sentry.js
```diff
@@ -17,8 +17,8 @@ async function createReleaseAndUpload() {
     await cli.releases.new(release);
     console.log("Uploading source maps");
     await cli.releases.uploadSourceMaps(release, {
-      include: ["build/static/js"],
-      urlPrefix: "~/static/js",
+      include: [".next/static/chunks"],
+      urlPrefix: "~/static/chunks",
```
darrenvong:
```suggestion
      urlPrefix: "~/_next/static/chunks",
```
Based on what I can see from inspecting the dev tool from the preview URL/site

darrenvong: Or on second thought, I wonder if we should just leave this script out altogether in CI and let Sentry fetch them automatically as mentioned here for [publicly hosted sourcemaps](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/hosting-publicly/) 🤔

lucaslcode: Agreed, I just checked and the source files do have the `# sourceMapping=...` comment

darrenvong (dismissed):
Let's ship it and see how it goes 🚀

PR #2161, opened 2021-11-03
Web meeting 2021-11-03

lucaslcode:


docs/meetings/web/2021-11-03.md
```diff
@@ -0,0 +1,19 @@
+# Web meeting Nov 03 2021
+
+**Present:** Aapeli, Chad, Lucas, Ricardo, William
+
+## Discussion
+
+- Translations: all (?) strings now exist in json files, but app is still using the constants. Need to move across
+
+- NextJS in progress
+  - May need to downgrade to v11 for a while if I can't get tests working on 12
```
darrenvong, line 10: What's different in v11 compared to v12?

docs/meetings/web/2021-11-03.md
```diff
@@ -0,0 +1,19 @@
+# Web meeting Nov 03 2021
+
+**Present:** Aapeli, Chad, Lucas, Ricardo, William
+
+## Discussion
+
+- Translations: all (?) strings now exist in json files, but app is still using the constants. Need to move across
```
darrenvong, line 7: I'm thinking the incremental move to use the JSON files is a nice idea, but it also means we can't set up the translation components until we do the swap/check the naming of the translation key etc as discussed in my other PR, which means people who want to help us translate stuff won't really be able to help. Should we just bite the bullet and do it in large batches/all in one go? 🥲

darrenvong (approved):
Non blocking questions I should have had answers to if I didn't miss the meeting in the first place... 🙈

docs/meetings/web/2021-11-03.md
lucaslcode: 12 uses a new rust transpiler and I'm having trouble getting the tests to run (I think there is a way to use Babel though so I'll try that first).

docs/meetings/web/2021-11-03.md
lucaslcode: Yeah maybe, want to do half each?

darrenvong: Yeah sounds like a plan!

PR #2197, opened 2021-11-22
i18n part 2: Swap constants in the auth feature folder to use i18n JSON resources

darrenvong:
As in title, as well as:
- Tweaked the i18n setup, so it only loads the "global" namespace by default. This is where generically shared strings like "Continue", "Submit" etc are contained
- Restructure of the translation key for the auth feature area from the literal constant file conversion to something more resilient to changes where possible. There are some messages I can't think of a better name for, so those I have left more or less the literal value as the key (but maybe still moved it within an object named after the page/form for better organisation)

Still WIP to fix the billion of tests I have broken because of the constant removals since they are used in tests too...

**Web frontend checklist**
- [x] All tests pass


app/web/src/features/auth/signup/AccountForm.tsx
```diff
@@ -276,7 +258,21 @@ export default function AccountForm() {
...
+                  [HostingStatus.HOSTING_STATUS_MAYBE]: t(
+                    "auth:account_form.hosting_status.maybe"
+                  ),
+                  [HostingStatus.HOSTING_STATUS_CANT_HOST]: t(
+                    "auth:account_form.hosting_status.cant_host"
+                  ),
+                  [HostingStatus.HOSTING_STATUS_UNSPECIFIED]:
+                    t("global:ask_me"),
+                  [HostingStatus.HOSTING_STATUS_UNKNOWN]: t("global:ask_me"),
+                }}
```
lucaslcode, line 275: I was wondering about this - will we definitely have to move all label maps like this into components?

app/web/src/features/auth/useAuthStore.ts
```diff
@@ -72,7 +71,7 @@ export default function useAuthStore() {
               action: "logout",
             },
           });
-          setError(isGrpcError(e) ? e.message : ERROR_INFO_FATAL);
+          setError(isGrpcError(e) ? e.message : "");
```
lucaslcode: What's going on here?

app/web/src/i18n.ts
```diff
@@ -16,7 +18,7 @@ i18n
           });
         return;
       }
-      import(`./features/locales/${language}.json`)
+      import(`./resources/locales/${language}.json`)
```
lucaslcode: Why resources? Seems a bit odd to have only global strings there and others in the respective feature folder.

app/web/src/features/auth/signup/AccountForm.tsx
```diff
@@ -318,8 +318,9 @@ export default function AccountForm() {
           )}
         />
         <Typography variant="body1">
-          {SIGN_UP_TOS_TEXT}
-          <TOSLink />.
+          <Trans i18nKey="auth:account_form.tos_prompt">
+            To continue, please read and accept the <TOSLink />.
```
lucaslcode, line 322: Is the test here just ignored?

lucaslcode (commented):
Looks good so far, few questions

app/web/src/features/auth/useAuthStore.ts
darrenvong: I'm hoping I can revert this... it was using the `t` function from `useTranslation` before, but for some strange reasons, a lot of tests are failing when it's **used here only** and not at the rest of the UI (e.g. in community components where I haven't got around to swapping out the constants), since it just gets stuck at the Suspense fallback state for some reason. Not really sure why  🤷

app/web/src/i18n.ts
darrenvong: I started adding the constants in some components (under the `components` folder, e.g. the error boundary's constants) into the global namespace, so it doesn't belong to `features` exclusively anymore. The `resources` folder feels "global-ish" to me based on what we got in there so far.

app/web/src/features/auth/signup/AccountForm.tsx
darrenvong: I think it's the most reliable that way, since it integrates with Suspense and ensures i18next is fully initialised before rendering this. Otherwise, we could use the `t` function on the `i18n` object directly, but then it might be possible that i18next hasn't finished initialising if we do it that way - e.g. if the user loads this page straight away.

I ran into this problem in tests because of said issue, but in tests at least I managed to work around it by doing a `waitFor` on `i18n.isInitialized` until it's true. Not sure how we'd work around that if this object is still statically listed in a constant file though when `t` is not ready to be used.

app/web/src/features/auth/signup/AccountForm.tsx
darrenvong: It shouldn't be ignored as this is a template string for i18next to swap to the right component due to how weird this `Trans` component works, so not sure what you mean by this being ignored?

lucaslcode: Because there's English text here, but also under the auth:account_form.tos_prompt key in the json, so I thought maybe the text here doesn't get used, just the `<TosLink/>`. Of course I wouldn't change it - just wondering.

darrenvong: Yeah you're right, the English text doesn't get displayed directly here, so it's not used in that respect, but it is used for determining the swap position of `<TOSLink />` in the JSON (or any React element if there are more) as the "template string". In this case, it'll read `<1></1>` in the JSON since it is the second children nested under `<Trans>` in this "template". 1 because it counts from 0.

Took me a few reading of this to understand how it works - https://react.i18next.com/latest/trans-component#how-to-get-the-correct-translation-string 😅

darrenvong:
Moved a few things around, the main change being the tests now uses a separate `i18n.init` with all of the resources loaded in rather than doing so lazily via dynamic import like in the main app. This means I don't have to convert all of the `render`s to be async to do

```typescript
  await waitFor(() => {
    expect(i18n.isInitialized).toBe(true);
  });
```
in a custom `render` that wraps React Testing Library's `render`, which was a bit awkward as it meant I had to update every test to add `await` in front of every `render` in every test suite... reverting that was not fun either 😢 

Otherwise, for some strange reasons I couldn't use the `useTranslation` hook in auth store without most tests being stuck in the `Suspense` fallback loading state. It's super weird and I still don't know why. I think it's ok though since in the tests, we are more interested in the translations being applied in the UI properly rather than testing specifically the Suspense behaviour, which should be tested by react-i18next (I hope) and so shouldn't be our concern to test that.

app/web/src/i18n/resources.ts
```diff
@@ -0,0 +1,23 @@
...
+  communities,
+  connections,
+  dashboard,
+  donations,
+  messages,
+  profile,
+  search,
+  global,
+} as const;
+
```
darrenvong, line 22: Since this is needed for the TypeScript types (in `src/i18n/react-i18next-types.d.ts` and the test setup of `i18n`), I thought it makes sense to put it in this separate file so we only have to update this in one place, then the types and the test i18n instance will update automatically.

lucaslcode (approved):

PR #2217, opened 2021-12-06
Add dropdown menu

### LINKED ISSUE #2183: Add profile + dropdown menu on navigation bar
Add a profile avatar + drop-down menu navigation bar. 

For example (for features live on the beta)  - 

Screen Shot 2021-11-14 at 8 15 34 PM

This would obviously change accordingly as more features are developed.

vrouilhac:
closes #2183 

I added the dropdown menu and the user avatar, if you have any suggestions of improvements please comment a list and i'll do the adjustements.

**Web frontend checklist**
- [ ] All tests pass

@lucaslcode

### DIFF
--- a/app/web/src/components/Menu.tsx
+++ b/app/web/src/components/Menu.tsx
@@ -4,7 +4,6 @@ import {
   MenuItemProps as MuiMenuItemProps,
   MenuProps,
 } from "@material-ui/core";
-import { grey } from "@material-ui/core/colors";
 import classNames from "classnames";
 import React from "react";
 import makeStyles from "utils/makeStyles";
@@ -20,7 +19,7 @@ const useStyles = makeStyles((theme) => ({
     alignItems: "center",
   },
   itemDivider: {
-    borderBottom: `1px solid ${grey[400]}`,
+    borderBottom: `1px solid ${theme.palette.divider}`,
   },
 }));
 
--- a/app/web/src/components/Menu.tsx
+++ b/app/web/src/components/Menu.tsx
@@ -1,15 +1,26 @@
 import {
   Menu as MuiMenu,
   MenuItem as MuiMenuItem,
-  MenuItemProps,
+  MenuItemProps as MuiMenuItemProps,
   MenuProps,
 } from "@material-ui/core";
+import { grey } from "@material-ui/core/colors";
+import classNames from "classnames";
 import React from "react";
 import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
   item: {
     paddingInline: theme.spacing(2),
+  },
+  itemMessage: {
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+  itemDivider: {
+    borderBottom: `1px solid ${grey[400]}`,
   },
 }));
 
@@ -20,16 +31,21 @@ export default function Menu(props: Omit<MenuProps, "className">) {
 
 //forwarding ref is necessary because Menu
 //injects refs into MenuItems
+interface MenuItemProps extends Omit<MuiMenuItemProps, "className"> {
+  hasNotification?: boolean;
+  hasBottomDivider?: boolean;
+}
+
 export const MenuItem = React.forwardRef(
-  (
-    props: Omit<MenuItemProps, "className">,
-    ref: React.ForwardedRef<HTMLLIElement>
-  ) => {
+  (props: MenuItemProps, ref: React.ForwardedRef<HTMLLIElement>) => {
     const classes = useStyles();
     return (
       <MuiMenuItem
         {...props}
-        className={classes.item}
+        className={classNames(classes.item, {
+          [classes.itemMessage]: props.hasNotification,
+          [classes.itemDivider]: props.hasBottomDivider,
+        })}
         button={true}
         ref={ref}
       />
--- a/app/web/src/components/Navigation/Navigation.tsx
+++ b/app/web/src/components/Navigation/Navigation.tsx
@@ -49,18 +49,10 @@ import {
 } from "../../constants";
 import NavButton from "./NavButton";
 
-const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
+const menu = () => [
   {
     name: DASHBOARD,
     route: "/",
-  },
-  {
-    name: MESSAGES,
-    route: messagesRoute,
-    notificationCount:
-      (data?.unseenMessageCount ?? 0) +
-      (data?.unseenReceivedHostRequestCount ?? 0) +
-      (data?.unseenSentHostRequestCount ?? 0),
   },
   {
     name: MAP_SEARCH,
@@ -226,13 +218,13 @@ export default function Navigation() {
   const drawerItems = (
     <div>
       <List>
-        {menu(data).map(({ name, route, notificationCount }) => (
+        {menu().map(({ name, route }) => (
           <ListItem button key={name}>
             <NavButton
               route={route}
               label={name}
               labelVariant="h2"
-              notificationCount={notificationCount}
+              notificationCount={undefined}
             />
           </ListItem>
         ))}
@@ -381,12 +373,12 @@ export default function Navigation() {
           <CouchersLogo />
           <Hidden smDown>
             <div className={classes.flex}>
-              {menu(data).map(({ name, route, notificationCount }) => (
+              {menu().map(({ name, route }) => (
                 <NavButton
                   route={route}
                   label={name}
                   key={`${name}-nav-button`}
-                  notificationCount={notificationCount}
+                  notificationCount={undefined}
                 />
               ))}
             </div>
--- a/app/web/src/components/Navigation/Navigation.tsx
+++ b/app/web/src/components/Navigation/Navigation.tsx
@@ -49,14 +49,26 @@ import {
 } from "../../constants";
 import NavButton from "./NavButton";
 
-const menu = () => [
+const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
   {
     name: DASHBOARD,
     route: "/",
   },
   {
+    name: MESSAGES,
+    route: messagesRoute,
+    notificationCount:
+      (data?.unseenMessageCount ?? 0) +
+      (data?.unseenReceivedHostRequestCount ?? 0) +
+      (data?.unseenSentHostRequestCount ?? 0),
+  },
+  {
     name: MAP_SEARCH,
     route: searchRoute,
+  },
+  {
+    name: PROFILE,
+    route: routeToProfile(),
   },
   {
     name: EVENTS,
@@ -218,13 +230,13 @@ export default function Navigation() {
   const drawerItems = (
     <div>
       <List>
-        {menu().map(({ name, route }) => (
+        {menu(data).map(({ name, route, notificationCount }) => (
           <ListItem button key={name}>
             <NavButton
               route={route}
               label={name}
               labelVariant="h2"
-              notificationCount={undefined}
+              notificationCount={notificationCount}
             />
           </ListItem>
         ))}
@@ -373,12 +385,12 @@ export default function Navigation() {
           <CouchersLogo />
           <Hidden smDown>
             <div className={classes.flex}>
-              {menu().map(({ name, route }) => (
+              {menu(data).map(({ name, route, notificationCount }) => (
                 <NavButton
                   route={route}
                   label={name}
                   key={`${name}-nav-button`}
-                  notificationCount={undefined}
+                  notificationCount={notificationCount}
                 />
               ))}
             </div>
--- a/app/web/src/components/Navigation/Navigation.tsx
+++ b/app/web/src/components/Navigation/Navigation.tsx
@@ -187,12 +187,6 @@ const useStyles = makeStyles((theme) => ({
   menuItem: {
     width: "100%",
   },
-  menuItemMessage: {
-    display: "flex",
-    flexDirection: "row",
-    justifyContent: "space-between",
-    alignItems: "center",
-  },
   notificationCount: {
     color: grey[500],
     fontWeight: "bold",
@@ -205,7 +199,7 @@ const useStyles = makeStyles((theme) => ({
     borderRadius: 999,
     backgroundColor: grey[200],
     padding: theme.spacing(1),
-    transition: "300ms ease-in-out",
+    transition: `${theme.transitions.duration.short}ms ${theme.transitions.easing.easeInOut}`,
     "&:hover": {
       opacity: 0.8,
       backgroundColor: grey[300],
@@ -216,11 +210,10 @@ const useStyles = makeStyles((theme) => ({
     width: "2rem",
   },
   badge: {
-    right: "-4px",
-    top: "4px",
-  },
-  menuItemDivider: {
-    borderBottom: `1px solid ${grey[400]}`,
+    "& .MuiBadge-badge": {
+      right: "-4px",
+      top: "4px",
+    },
   },
 }));
 
@@ -261,65 +254,63 @@ export default function Navigation() {
     </div>
   );
 
-  const menuItems = (
-    <div>
-      {menuDropDown(data).map(
-        ({
-          name,
-          notificationCount,
-          route,
-          target,
-          externalLink,
-          hasBottomDivider,
-        }) => {
-          const hasNotification =
-            notificationCount !== undefined && notificationCount > -1;
+  const menuItems = menuDropDown(data).map(
+    ({
+      name,
+      notificationCount,
+      route,
+      target,
+      externalLink,
+      hasBottomDivider,
+    }) => {
+      const hasNotification =
+        notificationCount !== undefined && notificationCount > 0;
 
-          const linkContent = (
-            <MenuItem
-              classes={{
-                root: classNames(classes.menuItem, {
-                  [classes.menuItemMessage]: hasNotification,
-                  [classes.menuItemDivider]: hasBottomDivider,
-                }),
-              }}
-            >
-              {hasNotification ? (
-                <Badge
-                  color="primary"
-                  variant="dot"
-                  classes={{ badge: classes.badge }}
-                >
-                  <Typography noWrap>{name}</Typography>
-                </Badge>
-              ) : (
-                <Typography noWrap>{name}</Typography>
-              )}
-
-              {hasNotification ? (
-                <Typography
-                  noWrap
-                  variant="subtitle2"
-                  className={classes.notificationCount}
-                >
-                  {`${notificationCount} unseen`}
-                </Typography>
-              ) : null}
-            </MenuItem>
-          );
-
-          return externalLink ? (
-            <a href={route} key={name} target={target}>
-              {linkContent}
-            </a>
+      const linkContent = (
+        <MenuItem
+          hasNotification={hasNotification}
+          hasBottomDivider={hasBottomDivider}
+        >
+          {hasNotification ? (
+            <Badge color="primary" variant="dot" className={classes.badge}>
+              <Typography noWrap>{name}</Typography>
+            </Badge>
           ) : (
-            <Link to={route} target={target} key={name}>
-              {linkContent}
-            </Link>
-          );
-        }
-      )}
-    </div>
+            <Typography noWrap>{name}</Typography>
+          )}
+
+          {hasNotification ? (
+            <Typography
+              noWrap
+              variant="subtitle2"
+              className={classes.notificationCount}
+            >
+              {`${notificationCount} unseen`}
+            </Typography>
+          ) : null}
+        </MenuItem>
+      );
+
+      return externalLink ? (
+        <a
+          href={route}
+          key={name}
+          target={target}
+          onClick={() => setMenuOpen(false)}
+        >
+          {linkContent}
+        </a>
+      ) : (
+        <Link
+          to={route}
+          target={target}
+          key={name}
+          onClick={() => setMenuOpen(false)}
+        >
+          {linkContent}
+        </Link>
+      );
+    }
   );
 
   const handleDrawerOpen = () => {
@@ -333,6 +324,7 @@ export default function Navigation() {
   if (!authenticated) {
     return null;
   }
+
   return (
     <AppBar
       position="sticky"
--- a/app/web/src/components/Navigation/Navigation.tsx
+++ b/app/web/src/components/Navigation/Navigation.tsx
@@ -1,22 +1,29 @@
 import {
   AppBar,
+  Avatar,
+  Badge,
   Drawer,
   Hidden,
   IconButton,
   List,
   ListItem,
   Toolbar,
+  Typography,
 } from "@material-ui/core";
+import { grey } from "@material-ui/core/colors";
 import classNames from "classnames";
 import { CloseIcon, MenuIcon } from "components/Icons";
+import Menu, { MenuItem } from "components/Menu";
 import ExternalNavButton from "components/Navigation/ExternalNavButton";
 import { useAuthContext } from "features/auth/AuthProvider";
 import useAuthStyles from "features/auth/useAuthStyles";
 import ReportButton from "features/ReportButton";
 import useNotifications from "features/useNotifications";
+import useCurrentUser from "features/userQueries/useCurrentUser";
 import React from "react";
 import CouchersLogo from "resources/CouchersLogo";
 import {
+  donationsRoute,
   eventsRoute,
   handbookURL,
   logoutRoute,
@@ -25,6 +32,7 @@ import {
   searchRoute,
 } from "routes";
 import makeStyles from "utils/makeStyles";
+import { acronym } from "utils/names";
 
 import {
   COUCHERS,
@@ -62,6 +70,38 @@ const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
   {
     name: EVENTS,
     route: eventsRoute,
+  },
+];
+
+const menuDropDown = (data: ReturnType<typeof useNotifications>["data"]) => [
+  {
+    name: "Profile",
+    title: true,
+  },
+  {
+    name: MESSAGES,
+    route: messagesRoute,
+    notificationCount:
+      (data?.unseenMessageCount ?? 0) +
+      (data?.unseenReceivedHostRequestCount ?? 0) +
+      (data?.unseenSentHostRequestCount ?? 0),
+  },
+  {
+    name: "Account",
+    route: routeToProfile(),
+  },
+  {
+    name: HELP,
+    target: "_blank",
+    route: handbookURL,
+  },
+  {
+    name: "Donate",
+    route: donationsRoute,
+  },
+  {
+    name: "Sign out",
+    route: logoutRoute,
   },
 ];
 
@@ -118,7 +158,6 @@ const useStyles = makeStyles((theme) => ({
   bug: {
     alignItems: "center",
     display: "flex",
-    justifyContent: "flex-end",
     [theme.breakpoints.down("sm")]: {
       paddingRight: theme.spacing(2),
     },
@@ -127,6 +166,57 @@ const useStyles = makeStyles((theme) => ({
     alignSelf: "center",
     fontWeight: "bold",
   },
+  menu: {
+    boxShadow: theme.shadows[1],
+    minWidth: "12rem",
+  },
+  menuPopover: {
+    transform: "translateY(2rem)",
+  },
+  menuTitle: {
+    fontWeight: "bold",
+    position: "relative",
+    pointerEvents: "none",
+    "&:after": {
+      content: "''",
+      position: "absolute",
+      left: 16,
+      bottom: 0,
+      borderBottom: "1px solid black",
+      width: 32,
+    },
+  },
+  menuItem: {
+    width: "100%",
+  },
+  menuItemMessage: {
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+  notificationCount: {
+    color: grey[500],
+    fontWeight: "bold",
+  },
+  menuBtn: {
+    display: "flex",
+    flexDirection: "row",
+    alignItems: "center",
+    border: `1px solid ${grey[300]}`,
+    borderRadius: "35%",
+    backgroundColor: grey[200],
+    padding: 4,
+    cursor: "pointer",
+  },
+  avatar: {
+    height: 30,
+    width: 30,
+  },
+  badge: {
+    right: "-4px",
+    top: "4px",
+  },
 }));
 
 export default function Navigation() {
@@ -134,7 +224,10 @@ export default function Navigation() {
   const classes = useStyles();
   const authenticated = useAuthContext().authState.authenticated;
   const [open, setOpen] = React.useState(false);
+  const [menuOpen, setMenuOpen] = React.useState<boolean>(false);
+  const menuRef = React.useRef<HTMLDivElement>(null);
   const { data } = useNotifications();
+  const { data: user } = useCurrentUser();
 
   const drawerItems = (
     <div>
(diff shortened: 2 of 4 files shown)


vrouilhac:
PS: Here is a screenshot of how it's looking, if you know a component that would make the button that wrap the avatar and burger icon looks better, i'd be pleased to know it and make the changes.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
-            />
-            <NavButton route={logoutRoute} label={LOG_OUT} />
-          </Hidden>
-          <ReportButton />
-        </div>
+
+        <Hidden smDown>
+          <div ref={menuRef} style={{ display: "flex", flexDirection: "row" }}>
+            <div
+              style={{ marginRight: 16, display: "flex", alignItems: "center" }}
```
darrenvong: We should style with class names in the `useStyle` like everywhere else for consistency.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
+            <div
+              className={classes.menuBtn}
+              onClick={menuOpen ? handleMenuClose : handleMenuOpen}
+            >
+              <MenuIcon />
+              <Avatar src={user?.avatarUrl} className={classes.avatar}>
+                {user?.avatarUrl && user.avatarUrl?.length > 0
+                  ? null
+                  : acronym(user?.name)}
+              </Avatar>
```
darrenvong: Looking at how this is used in other places, this logic is unnecessary since it's dealt with by MUI already if the user doesn't have an image if you use our Avatar component. You should be able to do this instead:

```suggestion
              <Avatar user={user} className={classes.avatar} />
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -1,22 +1,29 @@
 import {
   AppBar,
+  Avatar,
```
darrenvong: Don't import from `@material-ui/core` for the `Avatar` since we have a component built on top of that we use across the app. Instead, import from `components/Avatar` instead

app/web/src/components/Navigation/Navigation.tsx
darrenvong: Also, we don't need this extra `div` - instead, you can make the outer `div` on line 398 to have `alignItems: center` and apply the margin on the button instead.

Another thing - use the spacing in our theme - e.g. `theme.spacing(2)` when you move this to the `useStyles` instead of a literal pixel value, again for more consistency.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
+        <Hidden smDown>
+          <div ref={menuRef} style={{ display: "flex", flexDirection: "row" }}>
+            <div
+              style={{ marginRight: 16, display: "flex", alignItems: "center" }}
+            >
+              <ReportButton />
+            </div>
+            <div
+              className={classes.menuBtn}
+              onClick={menuOpen ? handleMenuClose : handleMenuOpen}
```
darrenvong:
```suggestion
              onClick={() => { setMenuOpen(!menuOpen) }}
```
Could inline and simplify to this rather than create two separate event handlers, since this is only toggling the boolean and so the abstraction seems unnecessary.

app/web/src/utils/names.ts
```diff
@@ -1,3 +1,7 @@
 export function firstName(fullName?: string) {
   return fullName?.split(" ")[0] ?? "";
 }
+
+export function acronym(name?: string) {
+  return name?.trim()?.slice(0, 1).toUpperCase();
+}
```
darrenvong: We can remove this function if using the `Avatar` component we have already as mentioned in the `Navigation` file.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    border: `1px solid ${grey[300]}`,
+    borderRadius: "35%",
+    backgroundColor: grey[200],
+    padding: 4,
+    cursor: "pointer",
+  },
+  avatar: {
+    height: 30,
+    width: 30,
+  },
```
darrenvong: Use the `rem` equivalent for this so it scales in and out better when the user change default font size

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
+
+        <Hidden smDown>
+          <div ref={menuRef} style={{ display: "flex", flexDirection: "row" }}>
+            <div
+              style={{ marginRight: 16, display: "flex", alignItems: "center" }}
+            >
+              <ReportButton />
+            </div>
+            <div
+              className={classes.menuBtn}
```
darrenvong: This should be implemented with the `button` element as using a `div` for buttons with an `onClick` is bad for accessibility (e.g. the menu can't be opened if you are a keyboard user). An `IconButton` from MUI might be a good choice here to use instead.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
...
+            notificationCount !== undefined && notificationCount > 0;
+
+          return (
+            <MenuItem key={name}>
+              <a
+                href={route}
+                target={target || "_self"}
+                className={`${classes.menuItem} ${
+                  hasNotification && classes.menuItemMessage
+                }`}
```
darrenvong:
```suggestion
                className={classNames(
                  classes.menuItem,
                  { [classes.menuItemMessage]: hasNotification }
                )}`}
```
We use the `classNames` library do this kinda of class joining more easily/clearly. Otherwise this would also wrongly interpolate "false" into the class name list if `hasNotification` is false

darrenvong (changes requested):
Looks like a good start! Feel free to add anything else/wrap this up if I don't reply quickly @lucaslcode 🙈

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -65,6 +73,38 @@ const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
...
+    name: MESSAGES,
+    route: messagesRoute,
+    notificationCount:
+      (data?.unseenMessageCount ?? 0) +
+      (data?.unseenReceivedHostRequestCount ?? 0) +
+      (data?.unseenSentHostRequestCount ?? 0),
+  },
+  {
+    name: "Account",
+    route: routeToProfile(),
```
lucaslcode: This should go to `/account-settings` using the appropriate constant.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -65,6 +73,38 @@ const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
   },
 ];
 
+const menuDropDown = (data: ReturnType<typeof useNotifications>["data"]) => [
+  {
+    name: "Profile",
+    title: true,
```
lucaslcode: Make this one link to profile instead of "account" linking to profile. (I know it's a 'heading', but people will still try to click it).

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -65,6 +73,38 @@ const menu = (data: ReturnType<typeof useNotifications>["data"]) => [
...
+    notificationCount:
+      (data?.unseenMessageCount ?? 0) +
+      (data?.unseenReceivedHostRequestCount ?? 0) +
+      (data?.unseenSentHostRequestCount ?? 0),
+  },
+  {
+    name: "Account",
+    route: routeToProfile(),
+  },
+  {
```
lucaslcode, line 98: In between Account and Help, it would be good to have another separator (to separate the "profile"-related items from the "general" items.) Maybe I would add a `hasBottomDivider` boolean to "Profile" and "Account"

app/web/src/components/Navigation/Navigation.tsx
lucaslcode: Try [`ButtonBase`](https://v4.mui.com/components/buttons/#customized-buttons) like in that example

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    fontWeight: "bold",
+  },
+  menuBtn: {
+    display: "flex",
+    flexDirection: "row",
+    alignItems: "center",
+    border: `1px solid ${grey[300]}`,
+    borderRadius: "35%",
+    backgroundColor: grey[200],
+    padding: 4,
```
lucaslcode: The button will look even better with twice as much padding and [a large px border radius](https://stackoverflow.com/questions/29966499/border-radius-in-percentage-and-pixels-px-or-em). Try this:
```suggestion
    borderRadius: 999,
    backgroundColor: grey[200],
    padding: theme.spacing(1), //0.5rem, 8px
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    cursor: "pointer",
```
lucaslcode: If you ever have to add the cursor property, it's a good indication that there is an accessibility problem (as Darren mentioned)

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
     </div>
   );
 
+  const menuItems = (
+    <div>
```
lucaslcode: A fragment should be fine I think: `<>.....</>` (I know you were copying the other one, maybe change that one too if it works)

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
...
+              <MenuItem classes={{ root: classes.menuTitle }}>{name}</MenuItem>
+            );
+
+          const nameElement = <Typography noWrap>{name}</Typography>;
+          const hasNotification =
+            notificationCount !== undefined && notificationCount > 0;
+
+          return (
+            <MenuItem key={name}>
+              <a
```
lucaslcode: Use the `Link` component from react-router for internal links

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
     </div>
   );
 
+  const menuItems = (
+    <div>
+      {menuDropDown(data).map(
+        ({ name, title, notificationCount, route, target }) => {
+          if (title)
+            return (
+              <MenuItem classes={{ root: classes.menuTitle }}>{name}</MenuItem>
```
lucaslcode: use `className={classes.menuTitle}

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
     alignSelf: "center",
     fontWeight: "bold",
   },
+  menu: {
+    boxShadow: theme.shadows[1],
+    minWidth: "12rem",
+  },
+  menuPopover: {
+    transform: "translateY(2rem)",
```
lucaslcode:
```suggestion
    transform: "translateY(1rem)",
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+  notificationCount: {
+    color: grey[500],
+    fontWeight: "bold",
+  },
+  menuBtn: {
```
lucaslcode, line 191: Make sure this has :hover styles too

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    top: "4px",
+  },
 }));
 
 export default function Navigation() {
   const authClasses = useAuthStyles();
   const classes = useStyles();
   const authenticated = useAuthContext().authState.authenticated;
   const [open, setOpen] = React.useState(false);
+  const [menuOpen, setMenuOpen] = React.useState<boolean>(false);
```
lucaslcode:
```suggestion
  const [menuOpen, setMenuOpen] = useState(false);
```
Importing useState directly is a style thing to match the rest of the repo, no need for the typing since it's inferred.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
...
+
+          const nameElement = <Typography noWrap>{name}</Typography>;
+          const hasNotification =
+            notificationCount !== undefined && notificationCount > 0;
+
+          return (
+            <MenuItem key={name}>
+              <a
+                href={route}
+                target={target || "_self"}
```
lucaslcode:
```suggestion
                target={target}
```
target={undefined} is fine - that's the default

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
...
+                  hasNotification && classes.menuItemMessage
+                }`}
+              >
+                {hasNotification ? (
+                  <Badge
+                    color="primary"
+                    variant="dot"
+                    classes={{ badge: classes.badge }}
+                  >
+                    {nameElement}
```
lucaslcode: Just inline `nameElement` - it's only one line, you're only using it twice, and it makes the code easier to read since you don't have to go back up to see what it actually is.

app/web/src/components/Navigation/Navigation.tsx
lucaslcode: Also, it should be outside `MenuItem`, not inside, otherwise nothing will happen if I click on the padding of `MenuItem`

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +256,57 @@ export default function Navigation() {
...
+                  </Badge>
+                ) : (
+                  nameElement
+                )}
+
+                {hasNotification ? (
```
lucaslcode: This looks great!

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
+                  ? null
+                  : acronym(user?.name)}
+              </Avatar>
+            </div>
+          </div>
+
+          <Menu
+            open={menuOpen}
+            anchorEl={menuRef.current}
+            elevation={0}
```
lucaslcode: Why?

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +393,47 @@ export default function Navigation() {
...
+              paper: classes.menu,
+            }}
+            anchorOrigin={{
+              vertical: "bottom",
+              horizontal: "right",
+            }}
+            getContentAnchorEl={null}
+            PopoverClasses={{
+              root: classes.menuPopover,
+            }}
```
lucaslcode, line 438: Instead of doing this, do`<Menu className={classes....`. Before there was no reason expect to match the rest of the codebase, but there is actually a very good reason now - we will have to do this when we migrate to MUI v5

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+  menu: {
+    boxShadow: theme.shadows[1],
+    minWidth: "12rem",
+  },
+  menuPopover: {
+    transform: "translateY(2rem)",
+  },
+  menuTitle: {
+    fontWeight: "bold",
+    position: "relative",
```
lucaslcode:
```suggestion
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +166,68 @@ const useStyles = makeStyles((theme) => ({
...
+    pointerEvents: "none",
```
lucaslcode: Remember to remove this when making profile a link

lucaslcode (changes requested):
This looks great!

Some small pointers - those lovely frontend details which are the last 2% of the work but can be a bit of a pain :D

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I added a Button, should i create a custom Button to which i add the icon or should i leave it as <Button ...> ?

lucaslcode: Not sure what you mean.

Darren is talking about `<button>`, the native html element. We also have a button component we use, `<Button>`, which has some extra stuff added to the MUI component `<Button>`. However I'm not sure it will work - if it doesn't, you could use MUI's `<ButtonBase>` component, which is more customizable.

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Great tip, thanks ! @darrenvong

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I also changed string names for the one already declared in contants :)

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I guess the difference between both us experience plays a role here ^^ look much better yes, thanks

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I'll remember that thanks !

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I thought that too but i think it's for a good reason, here's the error thrown by material-ui

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: It throws me an error as in Menu.tsx there is a Omit<..., "className">

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: ~~I always do that but i read some code and i'm pretty sure i see React.useState somewhere, should i change that too ?~~ 

Edit: my bad, i saw them in stories xD

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: I removed that property, but i'm not 100% sure i understand how elevation works o_O

Edit: Okay it just clicked ! ^^'

vrouilhac:
I fixed most of the feedback you gave @lucaslcode @darrenvong but there's still two things that goes wrong and i'm not sure how to solve those.

- First i can't replace classes => className as this throws me an error ( Navigation.tsx, lines 426, 280 )
- I get an error about refs but not sure i understand it ( see picture below ) ( Navigation.tsx, line 405 )

app/web/src/components/Navigation/Navigation.tsx
darrenvong: That's a nice CSS trick right there!

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +261,67 @@ export default function Navigation() {
...
+        ({
+          name,
+          notificationCount,
+          route,
+          target,
+          externalLink,
+          hasBottomDivider,
+        }) => {
+          const hasNotification =
+            notificationCount !== undefined && notificationCount > -1;
```
darrenvong:
```suggestion
            notificationCount !== undefined && notificationCount > 0;
```
I think what you had before was correct, as now the notification dot is showing all the time with "0 unseen"

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +400,44 @@ export default function Navigation() {
...
+            anchorEl={menuRef.current}
+            onClose={() => setMenuOpen(false)}
+            classes={{
+              paper: classes.menu,
+            }}
+            anchorOrigin={{
+              vertical: "bottom",
+              horizontal: "right",
+            }}
+            getContentAnchorEl={null}
```
darrenvong, line 435: Why is this prop necessary?

Also wondering if this is breaking arrow key navigation between the menu items, as only the whole menu can be focused by keyboard at the moment. It would be nice and more accessible too to be able to go between the links.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +172,67 @@ const useStyles = makeStyles((theme) => ({
...
+  },
+  menuBtn: {
+    display: "flex",
+    flexDirection: "row",
+    alignItems: "center",
+    border: `1px solid ${grey[300]}`,
+    borderRadius: 999,
+    backgroundColor: grey[200],
+    padding: theme.spacing(1),
+    transition: "300ms ease-in-out",
```
darrenvong:
```suggestion
    transition: `${theme.transitions.duration.short}ms ${theme.transitions.easing.easeInOut}`,
```
Only just noticed this, but we should use the value from the theme directly for more consistency here too

darrenvong:
> * I get an error about refs but not sure i understand it ( see picture below ) ( Navigation.tsx, line 405 )

I think that's a known "bug" from Material UI using an old/deprecated React API that's outside of our control. Should be fixed when we eventually move to v5 of the library.

> First i can't replace classes => className as this throws me an error ( Navigation.tsx, lines 426, 280 )

Yeah I think you're right that this doesn't work with our custom `MenuItem` unless we remove the `Omit<MenuItemProps, 'classname'>` on the component's props typing. @lucaslcode what do you think about that? I think you have more experience touching that component than I do to comment more :)

lucaslcode:
I will have another look later, but I believe I omitted  `classNames` to avoid customization so we wouldn't have different looking menus in different spots.

In this case, we do want to be able to have a 'title' item and a divider, so add them as props onto our `MenuItem` to be used whenever we want :) (To be clear, the classNames stuff will the be moved to the component, and in Navigation is will just use the props.)

app/web/src/components/Navigation/Navigation.tsx
lucaslcode: How about an array then?

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -395,30 +402,27 @@ export default function Navigation() {
...
-              <ReportButton />
-            </div>
-            <div
+          <div className={classes.menuContainer} ref={menuRef}>
+            <ReportButton />
+            <Button
               className={classes.menuBtn}
-              onClick={menuOpen ? handleMenuClose : handleMenuOpen}
+              onClick={() =>
+                setMenuOpen((prevMenuOpen: boolean) => !prevMenuOpen)
```
lucaslcode, line 410: the boolean can be inferred (not a bigdeal)

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -259,49 +264,59 @@ export default function Navigation() {
...
+                  [classes.menuItemMessage]: hasNotification,
+                  [classes.menuItemDivider]: hasBottomDivider,
+                }),
+              }}
+            >
+              {hasNotification ? (
+                <Badge
+                  color="primary"
+                  variant="dot"
+                  classes={{ badge: classes.badge }}
```
lucaslcode: use className and target the badge with `"& .MuiBadge-badge": {...}` or whatever it is

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -259,49 +264,59 @@ export default function Navigation() {
...
-                  </Badge>
-                ) : (
-                  nameElement
-                )}
+          const linkContent = (
+            <MenuItem
+              classes={{
+                root: classNames(classes.menuItem, {
+                  [classes.menuItemMessage]: hasNotification,
+                  [classes.menuItemDivider]: hasBottomDivider,
```
lucaslcode: move these to our `MenuItem` component as I mentioned in my comment

lucaslcode (changes requested):
Looking good!

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Yes, seems like the array is working

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Should i create an interface that extends Material UI's "MenuItemProps" type ?

lucaslcode: I would import the original as `MuiMenuItemProps`, make the interface `MenuItemProps` and extend the existing `Omit` type so you don't need that in the arguments.

vrouilhac: Okay, ^^ sorry if i don't understood what you meant but is that what you expected ?  :)

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Yes i was doing some test but i now removed it and make sure it'll be pushed as "0" :)

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Thanks, i didn't know there was transition values in theme. (not learn all the possibilities yet ;) )

app/web/src/components/Navigation/Navigation.tsx
darrenvong: Using this as an array is quite important here - I just had a play around and it's because of the outer div here, it's preventing us from going between the items in the menu when you are using the keyboard arrows, since the `div` somehow became the only focusable thing...

app/web/src/components/Navigation/Navigation.tsx
darrenvong: Ahhh never mind, looks like you get MUI warning without this while trying to set `anchorOrigin`

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +261,67 @@ export default function Navigation() {
...
+          );
+
+          return externalLink ? (
+            <a href={route} key={name} target={target}>
+              {linkContent}
+            </a>
+          ) : (
+            <Link to={route} target={target} key={name}>
+              {linkContent}
+            </Link>
```
darrenvong: I think the menu should be closed as soon as the user clicks on a link on it, just like how it works here on GitHub and on most other websites. Might have to add an onClick here to call `setMenuOpen(false)`

vrouilhac:
@lucaslcode @darrenvong i applied every changes you requested, i hope you'll find it good ! Tell me if anything i can change to perfect it and successfully close this issue ^^

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +392,44 @@ export default function Navigation() {
...
+              <Avatar
+                user={user}
+                className={classes.avatar}
+                isProfileLink={false}
+              />
+            </Button>
+          </div>
+
+          <Menu
+            open={menuOpen}
```
darrenvong:
```suggestion
          <Menu
            id="navigation-menu"
            open={menuOpen}
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +392,44 @@ export default function Navigation() {
...
-              label={HELP}
-              labelVariant="h3"
-            />
-            <NavButton route={logoutRoute} label={LOG_OUT} />
-          </Hidden>
-          <ReportButton />
-        </div>
+
+        <Hidden smDown>
+          <div className={classes.menuContainer} ref={menuRef}>
```
darrenvong:
```suggestion
          <div className={classes.menuContainer}}>
```

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -241,17 +392,44 @@ export default function Navigation() {
...
+            <ReportButton />
+            <Button
+              className={classes.menuBtn}
+              onClick={() =>
+                setMenuOpen((prevMenuOpen: boolean) => !prevMenuOpen)
+              }
+            >
```
darrenvong:
```suggestion
            <Button
              aria-controls="navigation-menu"
              aria-haspopup="true"
              className={classes.menuBtn}
              onClick={() =>
                setMenuOpen((prevMenuOpen: boolean) => !prevMenuOpen)
              }
              ref={menuRef}
            >
```
As shown in the [MUI Menu Docs](https://v4.mui.com/components/menus/#menus). It's good practice for accessibility reasons again. Because of that, the `menuRef` should be referencing the button that controls whether the menu is opened or not, rather than the generic `div` container that has both this button and the report button.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -127,14 +172,60 @@ const useStyles = makeStyles((theme) => ({
...
 }));
 
 export default function Navigation() {
   const authClasses = useAuthStyles();
   const classes = useStyles();
   const authenticated = useAuthContext().authState.authenticated;
-  const [open, setOpen] = React.useState(false);
+  const [open, setOpen] = useState(false);
+  const [menuOpen, setMenuOpen] = useState(false);
+  const menuRef = React.useRef<HTMLDivElement>(null);
```
darrenvong:
```suggestion
  const menuRef = React.useRef<HTMLButtonElement>(null);
```
Because of suggested change further down the file

darrenvong (dismissed):
Last thing just to make sure we are following the doc's best usage practice for accessibility stuff with aria props!

Otherwise, I think [there is a bug](https://github.com/mui-org/material-ui/issues/29371) around links in menu not being focusable in MUI when using the keyboard... so let's leave that as a separate issue after my suggestion

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +254,65 @@ export default function Navigation() {
...
+      ) : (
+        <Link
+          to={route}
+          target={target}
+          key={name}
+          onClick={() => setMenuOpen(false)}
+        >
+          {linkContent}
+        </Link>
+      );
```
darrenvong: Actually... one more thing 😅  just noticed this `<a>` is wrapping the `MenuItem` which generates invalid HTML... it should be the other way around as `Menu` is rendering as an `ul` under the hood.

i.e. this logic should be within `<MenuItem>`:
```jsx
<MenuItem>
  {
externalLink ? (
        <a
          href={route}
          key={name}
          target={target}
          onClick={() => setMenuOpen(false)}
        >
          {linkContent}
        </a>
      ) : (
        <Link
          to={route}
          target={target}
          key={name}
          onClick={() => setMenuOpen(false)}
        >
          {linkContent}
        </Link>
      );
  }
</MenuItem>
```

etc (excuse the bad formatting, you'll probably fix it easily with Prettier when you save with the editor locally)

lucaslcode: This might be my bad 😅

darrenvong: Why's that? Did you suggest this initially haha?

lucaslcode: I think so

app/web/src/components/Menu.tsx
```diff
@@ -1,16 +1,27 @@
...
     paddingInline: theme.spacing(2),
   },
+  itemMessage: {
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+    alignItems: "center",
+  },
+  itemDivider: {
+    borderBottom: `1px solid ${grey[400]}`,
```
lucaslcode: use `theme.palette.divider`

lucaslcode (changes requested):
Tiny comment and what darren said, then good to go I think!

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Thanks for this, i made some research because i struggled finding is MUI's doc about aria-controls for menu but found a link from their docs that redirect to w3.org that made me understand the aria-controls https://www.w3.org/TR/wai-aria-practices/#menubutton ;)

vrouilhac:
@darrenvong @lucaslcode I applied requested changes, and i also removed profile and messages from the header as those are presents inside the menu. If they should remain, i can revert those last changes

lucaslcode:
@Vincent-Projects Yes leave them there I think, people will be confused. It's okay to have twice for now.

app/web/src/components/Navigation/Navigation.tsx
```diff
@@ -163,6 +254,69 @@ export default function Navigation() {
...
+        <MenuItem
+          hasNotification={hasNotification}
+          hasBottomDivider={hasBottomDivider}
+        >
+          {externalLink ? (
+            <a
+              href={route}
+              key={name}
+              target={target}
+              onClick={() => setMenuOpen(false)}
```
lucaslcode, line 301: Okay REALLY last thing I promise, can you make this and the `Link` below have 100% width with a `className`. Right now they only work if you click the text, not the space on the right (and not the padding, but I think that's no big deal).

lucaslcode (changes requested):

app/web/src/components/Navigation/Navigation.tsx
vrouilhac: Of course :smile:

vrouilhac: @lucaslcode in order to do that i need to remove padding on li which is on line 13 inside components/Menu.tsx, should i add another props and add a styles inside classNames of that component ? Or should i remove that paddingInline ? ( the latter may impact other part of the app. Because `<a>` and `<Link>` are actually 100% of their parent :monocle_face:

darrenvong: I think what Lucas said should "just work" based on playing around the styles interactively in the browser. The `<a>`/`<Link>`s aren't 100% width since they don't have a width explicitly set at the moment

lucaslcode: I don't think they are?
image

vrouilhac: @lucaslcode my bad, i added the 100% locally and i meant about that green padding ^^ i'm pushing the changes

lucaslcode (approved):
🎉

PR #2224, opened 2021-12-10
Make it possible to write references if the host request wasn't confirmed by surfer

aapeliv:
This is a thing a lot of users complain about, including me :P

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -958,7 +958,7 @@ class HostRequest(Base):
     @hybrid_property
     def can_write_reference(self):
         return (
-            (self.status == HostRequestStatus.confirmed | self.status == HostRequestStatus.accepted)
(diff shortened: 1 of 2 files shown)


lucaslcode (commented):
Simple test case would be good

app/backend/src/couchers/models.py
```diff
@@ -958,15 +958,15 @@ class HostRequest(Base):
     @hybrid_property
     def can_write_reference(self):
         return (
-            (self.status == HostRequestStatus.confirmed)
+            (self.status == HostRequestStatus.confirmed | self.status == HostRequestStatus.accepted)
```
kalvdans: Read up on the priority of bitwise operations versus comparators. (You need parentheses here or switch to `or`)

aapeliv: Yes... I should've known that if I'd thought a bit harder...

kalvdans (approved):

Issue #1576, opened 2021-07-05
Epic: Internationalization (i18n), lokalization (l10n) and related tools


openscript:
As we are an international community languages are important to help couchers to access and use the platform. Not only translating the content, but also showing the time, dates or other localizable content will help for a good user experience. This is an epic which helps to organize the requirements to i18n and l10n.

- [x] Couchers-org/couchers#1577 
- [ ] Define process to create a new locale
- [ ] Set up i18n/l10n platform
- [ ] Determine users preferred locale
- [ ] Select locale and save selection
- [ ] Save locale selection in backend

lucaslcode:
Hi @openscript, how is this going?

openscript:
Hi Lucas
Recently I didn't have much time to work on this, but I will get on it this week.

lucaslcode:
closed in favour of Couchers-org/couchers#4152

PR #2226, opened 2021-12-11
feat: i18n for events@communities

papeldeorigami:
Translate the events basic text.

This is part of #1576 epic implemented with i18next.
Complex strings that depend on interpolation or logic are intentionally left out of this PR. A future PR should address that.

Screenshot of the affected screens: 
create-edit-comment

**Web frontend checklist**
- [x] Added any new components to storybook (storybook checked)
- [x] Added tests where relevant (tests fixed)
- [x] All tests pass


darrenvong (commented):
Looks good from a quick skim so far! Nice to see most of the "static" constants removed apart from the helper functions, which is expected, as they are more complex and will require more thoughts than doing the "naive" switch over as you describe it haha (this is not taking into consideration renaming anything)

app/web/src/features/dashboard/MyEvents.tsx
```diff
@@ -10,12 +10,11 @@ import Button from "components/Button";
 import HorizontalScroller from "components/HorizontalScroller";
```
darrenvong: (Comment not specific to this line, more generally for this file)

The CI pipeline mentions a failing test because `SEE_MORE_EVENTS_LABEL` no longer exists, so I think you'll need to update the tests for this file too (i.e. `MyEvents.test.tsx`) to use `t` from `test/utils` like you have done everywhere else

app/web/src/features/communities/events/constants.ts
```diff
@@ -1,53 +1,8 @@
...
-  "The event end time must be after the start time.";
-export const EVENT_DETAILS = "Event details";
-export const EVENT_DETAILS_REQUIRED = "Please fill in the event details field.";
-export const EVENT_DISCUSSION = "Event discussion";
-export const EVENT_IMAGE_INPUT_ALT = "Event image input";
-export const EVENT_LINK = "Event link";
-export const EVENTS_EMPTY_STATE = "No events at the moment.";
-export const EVENTS_LABEL = "Events";
-export const EVENTS_TITLE = "Events";
 export const getExtraAvatarCountText = (count: number) => `+${count}`;
```
darrenvong: If you are up for it, feel free to take a look at the [Formatting section of the i18next](https://www.i18next.com/translation-function/formatting) and see how we might convert this to use that so we can get rid of the functions in this file altogether.

app/web/src/features/communities/events/constants.ts
```diff
@@ -1,53 +1,8 @@
 import { EventsType } from "queryKeys";
 
-export const ATTENDEES = "Attendees";
 export const details = ({ colon = false }: { colon?: boolean } = {}) =>
   `Details${colon ? ":" : ""}`;
```
darrenvong: Looking at the usage of this, I suspect we'll probably want to duplicate this so that we have `Details` as its own standalone constant, and have `Details: {eventDetails}` as its own interpolation, since some languages might have two different words for this depending on the context, and so we probably want that flexibility as opposed to coupling ourselves against the English implementation by over DRYing this. What do you think?

(Feel free to leave this if you are not up for removing these functions as part of this)

lucaslcode:
I have converted this to a draft since it's WIP

app/web/src/features/dashboard/MyEvents.tsx
codehashi-dev: Hi, Darren! Sorry for the delay to answer. I don't know why `yarn test` wasn't able to catch this error, but `yarn test-ci` did the job. I fixed the issue (I hadn't change this file because it was in a another feature folder) and I will submit the code soon again for review. Thank you very much!

app/web/src/features/communities/events/constants.ts
codehashi-dev: Thanks for the hint! I am planning to use interpolation as soon as I finish the basic constants.

app/web/src/features/communities/events/constants.ts
codehashi-dev: Yeah, I'm definitely postponing this change until I finish the easy ones. But I agree that this can be solved with interpolation.

app/web/src/features/dashboard/MyEvents.tsx
papeldeorigami: Sorry, I was logged in to my corporate account when I added my last comments here, haha

papeldeorigami: Fixed. Thanks a lot for the heads up. I ran `yarn lint` but it didn't catch the lint error. Only `yarn test-ci` did the job. Maybe I'll investigate this or add this info to the README later.

app/web/src/features/communities/events/constants.ts
papeldeorigami: ok! I'll address this later

app/web/src/features/communities/events/constants.ts
papeldeorigami: Agree, I'll address this in another PR

papeldeorigami:
@lucaslcode thanks for setting it to draft mode. I finished the requested changes and updated the code to work with next.js. It should be ready for merge now.

darrenvong:
Overtaken by #2527 @papeldeorigami ?

papeldeorigami:
Right @darrenvong ! I will delete this branch altogether.

PR #2227, opened 2021-12-11
Disable postgres statement logging in local dev

aapeliv:
This makes life a bit easier. I don't think it's adding much.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/readme.md
+++ b/app/readme.md
@@ -64,19 +64,6 @@ docker-compose up --build
 
 This will build the backend, database, and proxy, and start them up.
 
-<!--
-If successful you will see this message being repeated every few seconds which indicates that the backend is working.
-```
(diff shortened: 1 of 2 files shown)


app/readme.md
```diff
@@ -64,6 +64,7 @@ docker-compose up --build
 
 This will build the backend, database, and proxy, and start them up.
 
+<!--
```
kalvdans: Remove text instead of comment it out. It is kept in the git history.

kalvdans (dismissed):
Great, I think too that debugging should be enabled when the need arises. (unless you are hunting non-reprocible bugs in production)

kalvdans (approved):

PR #2238, opened 2021-12-16
Update rules for contributor form notify emails

ItsiW:


app/backend/src/couchers/models.py
```diff
@@ -532,7 +532,15 @@ def should_notify(self):
...
         """
-        return (self.experience != None) | (self.contribute_ways != []) | (self.expertise != None)
+        return (
+            # these can't be all empty
+            (self.experience != None)
+            | (self.contribute_ways != [])
+            | (self.expertise != None)
+        ) & ~(
+            # can't have these empty if no expertise is given
+            set(self.contribute_ways).issubset(set(["community", "blog", "other"]) & self.expertise == None)
```
aapeliv: Does this work?

ItsiW: fixed

aapeliv (approved):
OK

PR #2241, opened 2021-12-18
Adjust margins in account settings page

tsuruby:
This PR adds proper margins to the components on the account settings page. It also adds a new component `Section.tsx` in settings page. Section component contains a title and a content which are passed as props.:
```
type SectionProps = {
  title: string;
  content: string | React.ReactNode;
  className?: string;
};
```

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [X] All tests pass

### DIFF
--- a/app/web/src/components/PageTitle.tsx
+++ b/app/web/src/components/PageTitle.tsx
@@ -8,8 +8,19 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function PageTitle(props: TypographyProps) {
+type PageTitleProps = TypographyProps & {
+  className?: string;
+};
+
+export default function PageTitle({ className, ...props }: PageTitleProps) {
   const classes = useStyles();
 
-  return <Typography {...props} classes={{ h1: classes.root }} variant="h1" />;
+  return (
+    <Typography
+      {...props}
+      className={className}
+      classes={{ h1: classes.root }}
+      variant="h1"
+    />
+  );
 }
--- a/app/web/src/components/PageTitle.tsx
+++ b/app/web/src/components/PageTitle.tsx
@@ -8,19 +8,8 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-type PageTitleProps = TypographyProps & {
-  className?: string;
-};
-
-export default function PageTitle({ className, ...props }: PageTitleProps) {
+export default function PageTitle(props: TypographyProps) {
   const classes = useStyles();
 
-  return (
-    <Typography
-      {...props}
-      className={className}
-      classes={{ h1: classes.root }}
-      variant="h1"
-    />
-  );
+  return <Typography {...props} className={classes.root} variant="h1" />;
 }
--- a/app/web/src/features/auth/Settings.tsx
+++ b/app/web/src/features/auth/Settings.tsx
@@ -1,14 +1,15 @@
-import { Typography } from "@material-ui/core";
 import Alert from "components/Alert";
 import CircularProgress from "components/CircularProgress";
 import HtmlMeta from "components/HtmlMeta";
 import PageTitle from "components/PageTitle";
 import ChangeEmail from "features/auth/email/ChangeEmail";
 import { ChangePassword } from "features/auth/password";
+import Section from "features/auth/section/Section";
 import Timezone from "features/auth/timezone/Timezone";
 import Username from "features/auth/username/Username";
 import { GetAccountInfoRes } from "proto/account_pb";
 
+import makeStyles from "../../utils/makeStyles";
 import {
   ACCOUNT_SETTINGS,
   CHANGE_BIRTHDATE,
@@ -18,6 +19,12 @@ import {
 } from "./constants";
 import useAccountInfo from "./useAccountInfo";
 
+const usePageTitleStyles = makeStyles((theme) => ({
+  root: {
+    paddingBottom: 0,
+  },
+}));
+
 export default function Settings() {
   const {
     data: accountInfo,
@@ -25,14 +32,16 @@ export default function Settings() {
     isLoading: isAccountInfoLoading,
   } = useAccountInfo();
 
+  const pageTitleClasses = usePageTitleStyles();
+
   return (
     <>
       <HtmlMeta title={ACCOUNT_SETTINGS} />
-      <PageTitle>{ACCOUNT_SETTINGS}</PageTitle>
-      <Typography variant="h2">{CHANGE_GENDER}</Typography>
-      <Typography variant="body1">{CHANGE_GENDER_CONTACT}</Typography>
-      <Typography variant="h2">{CHANGE_BIRTHDATE}</Typography>
-      <Typography variant="body1">{CHANGE_BIRTHDATE_CONTACT}</Typography>
+      <PageTitle className={pageTitleClasses.root}>
+        {ACCOUNT_SETTINGS}
+      </PageTitle>
+      <Section title={CHANGE_GENDER} content={CHANGE_GENDER_CONTACT} />
+      <Section title={CHANGE_BIRTHDATE} content={CHANGE_BIRTHDATE_CONTACT} />
       {isAccountInfoLoading ? (
         <CircularProgress />
       ) : accountInfoError ? (
--- a/app/web/src/features/auth/Settings.tsx
+++ b/app/web/src/features/auth/Settings.tsx
@@ -8,8 +8,8 @@ import Section from "features/auth/section/Section";
 import Timezone from "features/auth/timezone/Timezone";
 import Username from "features/auth/username/Username";
 import { GetAccountInfoRes } from "proto/account_pb";
+import makeStyles from "utils/makeStyles";
 
-import makeStyles from "../../utils/makeStyles";
 import {
   ACCOUNT_SETTINGS,
   CHANGE_BIRTHDATE,
@@ -19,9 +19,12 @@ import {
 } from "./constants";
 import useAccountInfo from "./useAccountInfo";
 
-const usePageTitleStyles = makeStyles((theme) => ({
-  root: {
-    paddingBottom: 0,
+const useStyles = makeStyles((theme) => ({
+  section: {
+    margin: theme.spacing(4, 0),
+    "&:first-of-type": {
+      marginTop: theme.spacing(2),
+    },
   },
 }));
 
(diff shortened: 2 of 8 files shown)


app/web/src/components/PageTitle.tsx
```diff
@@ -8,8 +8,19 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default function PageTitle(props: TypographyProps) {
+type PageTitleProps = TypographyProps & {
+  className?: string;
+};
+
+export default function PageTitle({ className, ...props }: PageTitleProps) {
```
lucaslcode: `className?` should already be in `TypographyProps` so you don't actually need the new type.

app/web/src/components/PageTitle.tsx
```diff
@@ -8,8 +8,19 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
 
-  return <Typography {...props} classes={{ h1: classes.root }} variant="h1" />;
+  return (
+    <Typography
+      {...props}
+      className={className}
+      classes={{ h1: classes.root }}
```
lucaslcode: I know this isn't your code, but we're moving away from using `classes` since we eventually want to move to MUIv5 and Emotion, which doesn't support it (it's not necessary). Instead can you do `className={classNames(className, classes.root)}` (using the `classNames` package which we already use to combine classes etc.)

app/web/src/features/auth/Settings.tsx
```diff
@@ -18,21 +19,29 @@ import {
 } from "./constants";
 import useAccountInfo from "./useAccountInfo";
 
+const usePageTitleStyles = makeStyles((theme) => ({
```
lucaslcode: We usually just call it `useStyles` if it's the styles only for this file/component (and then change `root` to `pageTitle`) - but not a big deal

app/web/src/features/auth/email/ChangeEmail.tsx
```diff
@@ -11,16 +11,24 @@ import { useMutation } from "react-query";
 import { service } from "service";
 import { lowercaseAndTrimField } from "utils/validation";
 
+import makeStyles from "../../../utils/makeStyles";
```
lucaslcode: use absolute imports for things outside this feature (see the line above this one)

app/web/src/features/auth/email/ChangeEmail.tsx
```diff
@@ -11,16 +11,24 @@ import { useMutation } from "react-query";
...
 import useChangeDetailsFormStyles from "../useChangeDetailsFormStyles";
 
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
```
lucaslcode: Avoid root level margins because it affects things outside this component and is more difficult to maintain/reason about. That probably means adapting these components to accept `className` prop and using that in the parent. Slightly more work now, but much easier in the long run!

app/web/src/features/auth/password/ChangePassword.tsx
```diff
@@ -12,6 +12,14 @@ import { useTranslation } from "react-i18next";
 import { useMutation, useQueryClient } from "react-query";
 import { service } from "service";
 
+import makeStyles from "../../../utils/makeStyles";
```
lucaslcode: absolute

app/web/src/features/auth/password/ChangePassword.tsx
```diff
@@ -12,6 +12,14 @@ import { useTranslation } from "react-i18next";
...
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
```
lucaslcode: as above

app/web/src/features/auth/section/Section.tsx
```diff
@@ -0,0 +1,29 @@
+import { Typography } from "@material-ui/core";
+import classNames from "classnames";
+import React from "react";
+
+import makeStyles from "../../../utils/makeStyles";
```
lucaslcode: absolute

app/web/src/features/auth/section/Section.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
+  },
+}));
+
+type SectionProps = {
+  title: string;
+  content: string | React.ReactNode;
```
lucaslcode: `ReactNode` can already be a string and import `ReactNode` directly instead of `React`
```suggestion
  content: ReactNode;
```

app/web/src/features/auth/section/Section.tsx
```diff
@@ -0,0 +1,29 @@
+import { Typography } from "@material-ui/core";
+import classNames from "classnames";
+import React from "react";
+
+import makeStyles from "../../../utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
```
lucaslcode: Same thing about the margin.

app/web/src/features/auth/timezone/Timezone.tsx
```diff
@@ -2,23 +2,34 @@ import { Typography } from "@material-ui/core";
 import { GetAccountInfoRes } from "proto/account_pb";
 import dayjs from "utils/dayjs";
 
+import makeStyles from "../../../utils/makeStyles";
```
lucaslcode: Absolute import

app/web/src/features/auth/timezone/Timezone.tsx
```diff
@@ -2,23 +2,34 @@ import { Typography } from "@material-ui/core";
...
 import {
   TIMEZONE,
   TIMEZONE_HELPER,
   YOUR_LOCAL_TIME_IS,
   YOUR_TIMEZONE,
 } from "../constants";
 
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
```
lucaslcode: no root margin

app/web/src/features/auth/username/Username.tsx
```diff
@@ -1,20 +1,33 @@
...
 import { GetAccountInfoRes } from "proto/account_pb";
+import React from "react";
 import { useTranslation } from "react-i18next";
 
+import makeStyles from "../../../utils/makeStyles";
 import { USERNAME_HELPER, YOUR_USERNAME_IS } from "../constants";
 
+const useStyles = makeStyles((theme) => ({
+  root: {
+    margin: theme.spacing(4, 0),
```
lucaslcode: no root margin

app/web/src/features/auth/username/Username.tsx
```diff
@@ -1,20 +1,33 @@
...
+  const classes = useStyles();
+
   return (
-    <>
+    <div className={classes.root}>
       <Typography variant="h2">
         {t("account_form.username.field_label")}
       </Typography>
-      <Typography variant="body1" paragraph>
+      <Typography className={classes.username} variant="body1" paragraph>
```
lucaslcode: I think `paragraph` is adding the margin, so maybe you can just get rid of that instead of adding a new class?

lucaslcode (changes requested):
Thanks so much for working on this! Just some comments about conventions in our codebase.

tsuruby:
Thanks a lot for the very valuable comments 🙏. I applied your requests :)

app/web/src/components/PageTitle.tsx
tsuruby: Changed it in last commit

app/web/src/components/PageTitle.tsx
tsuruby: I removed it in last push. In this case I didn't need to merge two classes because extra styling of the pageTitle was not needed anymore.

app/web/src/features/auth/Settings.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/email/ChangeEmail.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/email/ChangeEmail.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/section/Section.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/timezone/Timezone.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/timezone/Timezone.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/username/Username.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/username/Username.tsx
tsuruby: Changed it in last push

app/web/src/features/auth/section/Section.tsx
tsuruby: Changed in last push

app/web/src/features/auth/Settings.tsx
```diff
@@ -18,31 +19,60 @@ import {
...
         <Alert severity="error">{accountInfoError.message}</Alert>
       ) : (
         <>
-          <Username {...(accountInfo as GetAccountInfoRes.AsObject)} />
-          <Timezone {...(accountInfo as GetAccountInfoRes.AsObject)} />
-          <ChangeEmail {...(accountInfo as GetAccountInfoRes.AsObject)} />
-          <ChangePassword {...(accountInfo as GetAccountInfoRes.AsObject)} />
+          <Username
+            className={classes.section}
+            {...(accountInfo as GetAccountInfoRes.AsObject)}
```
lucaslcode, line 62: For future reference, we avoid spreading props since it's a bit opaque, and the type casting shows that typescript isn't being used very well here - but that wasn't your code and I don't want to be too annoying for your first contribution :)

lucaslcode (approved):
Thanks so much! 🎉

Issue #2187, opened 2021-11-18
Message screen: selecting name on the message header directs viewer to user profile 


dommers-ux:
<img width="593" alt="Screen Shot 2021-11-17 at 8 15 03 PM" src="https://user-images.githubusercontent.com/83376913/142339017-20eb6763-7494-459d-8eab-b489aebb0127.png">

Discussed in: https://community.couchers.org/t/unable-to-click-on-profile-in-messages-please-fix-this/1605

aapeliv:
On purpose, #1908

PR #2249, opened 2021-12-22
(done) Added link to user profile in chat username

vrouilhac:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/src/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatView.tsx
@@ -19,7 +19,10 @@ import MessageList from "features/messages/messagelist/MessageList";
 import useMarkLastSeen, {
   MarkLastSeenVariables,
 } from "features/messages/useMarkLastSeen";
-import { groupChatTitleText } from "features/messages/utils";
+import {
+  groupChatTitleText,
+  groupChatUsernameText,
+} from "features/messages/utils";
 import useUsers from "features/userQueries/useUsers";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { Error as GrpcError } from "grpc-web";
@@ -36,7 +39,7 @@ import {
   useQuery,
   useQueryClient,
 } from "react-query";
-import { useHistory, useParams } from "react-router-dom";
+import { Link, useHistory, useParams } from "react-router-dom";
 import { service } from "service";
 
 import { ERROR_UNKNOWN } from "../constants";
@@ -63,6 +66,7 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
     height: `calc(100vh - ${theme.shape.navPaddingXs})`,
   },
   title: {
+    fontSize: "1.2rem",
     flexGrow: 1,
     marginInlineEnd: theme.spacing(2),
     marginInlineStart: theme.spacing(2),
@@ -179,6 +183,10 @@ export default function GroupChatView() {
   const title = groupChat
     ? groupChatTitleText(groupChat, groupChatMembersQuery, currentUserId)
     : undefined;
+  const groupChatUsername = groupChatUsernameText(
+    groupChatMembersQuery,
+    currentUserId
+  );
 
   return (
     <div>
@@ -192,9 +200,15 @@ export default function GroupChatView() {
               <BackIcon />
             </HeaderButton>
 
-            <PageTitle className={classes.title}>
-              {title || <Skeleton width={100} />}
-            </PageTitle>
+            {groupChatUsername ? (
+              <Link to={`/user/${groupChatUsername}`} className={classes.title}>
+                {title || <Skeleton width={100} />}
+              </Link>
+            ) : (
+              <PageTitle className={classes.title}>
+                {title || <Skeleton width={100} />}
+              </PageTitle>
+            )}
 
             {!groupChat?.isDm && (
               <>
--- a/app/web/src/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatView.tsx
@@ -19,10 +19,7 @@ import MessageList from "features/messages/messagelist/MessageList";
 import useMarkLastSeen, {
   MarkLastSeenVariables,
 } from "features/messages/useMarkLastSeen";
-import {
-  formatGroupChatUsernames,
-  groupChatTitleText,
-} from "features/messages/utils";
+import { getDmUsername, groupChatTitleText } from "features/messages/utils";
 import useUsers from "features/userQueries/useUsers";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { Error as GrpcError } from "grpc-web";
@@ -52,6 +49,7 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
     paddingBottom: theme.spacing(2),
   },
   header: {
+    paddingTop: theme.spacing(1),
     alignItems: "center",
     display: "flex",
     flexGrow: 0,
@@ -186,10 +184,8 @@ export default function GroupChatView() {
   const title = groupChat
     ? groupChatTitleText(groupChat, groupChatMembersQuery, currentUserId)
     : undefined;
-  const groupChatUsername = formatGroupChatUsernames(
-    groupChatMembersQuery,
-    currentUserId
-  );
+
+  const groupChatUsername = getDmUsername(groupChatMembersQuery, currentUserId);
 
   return (
     <div>
--- a/app/web/src/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatView.tsx
@@ -185,8 +185,6 @@ export default function GroupChatView() {
     ? groupChatTitleText(groupChat, groupChatMembersQuery, currentUserId)
     : undefined;
 
-  const groupChatUsername = getDmUsername(groupChatMembersQuery, currentUserId);
-
   return (
     <div>
       <HtmlMeta title={title} />
@@ -199,8 +197,13 @@ export default function GroupChatView() {
               <BackIcon />
             </HeaderButton>
 
-            {groupChatUsername && groupChat?.isDm ? (
-              <Link to={`/user/${groupChatUsername}`}>
+            {groupChat?.isDm ? (
+              <Link
+                to={`/user/${getDmUsername(
+                  groupChatMembersQuery,
+                  currentUserId
+                )}`}
+              >
                 <PageTitle className={classes.title}>
                   {title || <Skeleton width={100} />}
                 </PageTitle>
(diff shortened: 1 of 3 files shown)


vrouilhac:
I added a link to the user profile using already existing blocks of code, also i fixed a small error on Navigation.tsx ( my last PR #2217  ), since your very last feedback i made a change and forgot to move `key={name}` to the right place inside `.map` function. This is fixed now.

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -192,9 +200,15 @@ export default function GroupChatView() {
               <BackIcon />
             </HeaderButton>
 
-            <PageTitle className={classes.title}>
-              {title || <Skeleton width={100} />}
-            </PageTitle>
+            {groupChatUsername ? (
```
lucaslcode: I think you also want to add a check for `groupChat?.isDm` here, in case it's a group chat (which can have a title), but people have left so there's only 2 users left.

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -192,9 +200,15 @@ export default function GroupChatView() {
...
+              <Link to={`/user/${groupChatUsername}`} className={classes.title}>
```
lucaslcode: Instead of explicitly setting the size in the className, you can have `PageTitle` inside this link. That way if we decide to change `PageTitle` later, it will still be consistent between the link and non-link version.

app/web/src/features/messages/utils.ts
```diff
@@ -58,3 +58,18 @@ export function groupChatTitleText(
         .map((user) => firstName(user?.name))
         .join(", ");
 }
+
+export function groupChatUsernameText(
```
lucaslcode: We don't need JSDoc comments as often when using typescript, but it's not clear what this function does just from the name and type. By making this change, hovering the mouse over the function name in VSCode (or any good IDE) will have the written description so you don't have to switch files and read the funciton carefully to figure out what's going on.
```suggestion
/** Returns the other user's username, or null if there are more than 2 users. */
export function groupChatUsernameText(
```

lucaslcode: Maybe also worth renaming this function to `groupChatOtherUsername`

app/web/src/features/messages/groupchats/GroupChatView.tsx
lucaslcode: Indeed, it does look quite different between the link and no link versions at the moment

lucaslcode (changes requested):
Thanks for doing this!

I also noticed, that at some point the layout on the messages title bar broke. It's meant to gave the buttons on either side, and there's no margins! Do you want to add this to `classes.header`:

```
"& > * + *": { marginInlineStart: theme.spacing(2) },
```

And give `width: "100%"` to `classes.title`. That should give the right effect.

app/web/src/features/messages/utils.ts
```diff
@@ -58,3 +58,18 @@ export function groupChatTitleText(
...
+  groupChatMembersQuery: ReturnType<typeof useUsers>,
+  currentUserId: number
+) {
+  const users = Array.from(groupChatMembersQuery.data?.values() ?? []);
+  if (users.length === 2) {
+    const username = users.find(
+      (user) => user?.userId !== currentUserId
+    )?.username;
+    return username ? username : null;
```
lucaslcode:
```suggestion
    return username ?? null;
```

app/web/src/features/messages/utils.ts
darrenvong: Or maybe something like `formatGroupChatUsernames`? Not too strongly opinionated on this personally, but I find often when you have to write a comment to explain a function (especially when it's not some unavoidable obscure code that has to be done for perf reasons or something), it's a bit of a code smell

darrenvong: Actually, I'm confused why we need this at all. Doesn't  `groupChatTitleText` already format the names correctly? Surely we could bundle the logic in there and instead of running the same check/processing twice?

vrouilhac: groupChatTitleText format the name Good but it does not takes the username used for the url /user/{username} ? Or maybe i can use the same function and add a parameters to the function like `returnUsername` as boolean ? @darrenvong

app/web/src/features/messages/groupchats/GroupChatView.tsx
vrouilhac: Thanks ^^ i wasn't aware of that case

vrouilhac:
I applied the changes, and wonder if i should add a spacing top as i see that the button is a bit hidden.
 


Suggestions for the header margin:
`paddingTop: theme.spacing(1)` ?

Here is the result ( not commited yet )

app/web/src/features/messages/utils.ts
```diff
@@ -58,3 +58,19 @@ export function groupChatTitleText(
         .map((user) => firstName(user?.name))
         .join(", ");
 }
+
+/** Returns the other user's username, or null if there are more than 2 users. */
+export function formatGroupChatUsernames(
```
darrenvong: > groupChatTitleText format the name Good but it does not takes the username used for the url /user/{username} ? Or maybe i can use the same function and add a parameters to the function like returnUsername as boolean ?

Oops first of all, I didn't read the comment above the function/code properly until now, sorry! 🙈  But it looks like this is only intended for "group chats" with exactly one other user (i.e. it's a DM), so maybe it's better to name the function `formatUserInDm` or something similar to those words.

~~And I see what you mean about the problem of not being able to get the username back out. Instead of separating the formatting of the text and the JSX separately, I think maybe it's worth combining them into the `groupChatTitleText` helper function above.~~

~~So I mean something like the following (rough code to give you an idea, so might not work exactly!):~~

(Never mind, ignore this 🙃)
```tsx
export function groupChatTitleText(
  groupChat: GroupChat.AsObject,
  groupChatMembersQuery: {
    data?: Map<number | undefined, User.AsObject | undefined>;
  },
  currentUserId: number
) {
  if (groupChat.title) {
    return <PageTitle>{groupChat.title}</PageTitle>;
  }

  if (groupChatMembersQuery.isLoading) {
    return <PageTitle>Chat</PageTitle>;
  }

  const otherGroupChatMembers = Array.from(
    groupChatMembersQuery.data?.values() ?? []
  ).filter((user) => user?.userId !== currentUserId);

  // DM
  if (otherGroupChatMembers.length === 1) {
    return (
      <Link>
        <PageTitle>{firstName(user?.name)}</PageTitle>
      </Link>
    );
  }

  // group chat
  return (
    <PageTitle>
      {otherGroupChatMembers.map((user) => firstName(user?.name)).join(", ")}
    </PageTitle>
  );
}
```

darrenvong:
Also, does this close #2187? Thought I mention it so it's linked and we don't have loose issues lying around once this is resolved/merged :)

lucaslcode (changes requested):
All looks good - maybe just look at Darren's recommendation which is a better idea than what I said

app/web/src/features/messages/utils.ts
vrouilhac: I understand and made those changes, in order to achieve this i had to replace the file `utils.ts` to `utils.tsx`, and i changes parts of the code that uses that function.

There is one remaining problem tho, line 190 on file `GroupChatView.tsx` we make use of title ( which is the result of the above function ) as title for HtmlMeta, function which now returns JSX, do you have any idea how i could solve this in a clean way ? 

Also this causes the GroupChatListItem a red warning as it makes nesting of two `a` element as follow 
`<a><a> </a></a>`. ^^'

One idea i have is keeping the code as you suggested above, and adding a fourth parameters `withoutJSX` as a boolean to either get back the JSX or the string, default to false ?

vrouilhac: PS: Sorry for delay the end of the year was busy ^^'

darrenvong: Ahh I didn't see the `title` being used for the HTML metadata as well, what a pain in the arse!

In that case, I think what you got there already is probably "cleaner" than what I suggested, as I think mixing the helper to sometimes return JSX, and at other times return text only is more confusing, since the function is doing too much imo at that point.

app/web/src/features/messages/utils.ts
```diff
@@ -59,7 +59,8 @@ export function groupChatTitleText(
         .join(", ");
 }
 
-export function groupChatUsernameText(
+/** Returns the other user's username, or null if there are more than 2 users. */
+export function formatGroupChatUsernames(
```
darrenvong:
```suggestion
export function getDmUsername(
```
I'd suggest renaming the function to this then and remove the need of the comment

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -200,9 +203,11 @@ export default function GroupChatView() {
               <BackIcon />
             </HeaderButton>
 
-            {groupChatUsername ? (
-              <Link to={`/user/${groupChatUsername}`} className={classes.title}>
-                {title || <Skeleton width={100} />}
+            {groupChatUsername && groupChat?.isDm ? (
+              <Link to={`/user/${groupChatUsername}`}>
```
darrenvong:
```suggestion
              <Link to={`/user/${getDmUsername(groupChatMembersQuery, currentUserId)}`}>
```
I think maybe inline the function call and remove 189-191 might make this slightly clearer too, as we don't really need to call the helper for a group chat

app/web/src/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -183,7 +186,7 @@ export default function GroupChatView() {
   const title = groupChat
     ? groupChatTitleText(groupChat, groupChatMembersQuery, currentUserId)
     : undefined;
-  const groupChatUsername = groupChatUsernameText(
+  const groupChatUsername = formatGroupChatUsernames(
     groupChatMembersQuery,
     currentUserId
```
darrenvong:
```suggestion
```

app/web/src/features/messages/groupchats/GroupChatView.tsx
darrenvong: P.S. This might even warrant its own component, so the `getDmUsername` stuff and the `groupChatMembersQuery` stuff can be kept closer altogether rather than spread across this large component and be used like `<GroupChatTitle />`! But that might be too perfectionist (since you didn't even write most of the code here in the first place!) and so happy to leave it as is so this can be merged soon. Thought I'll still mention that as something for you to think about in the future :)

vrouilhac: Yes i'll keep that in mind feel free to tag me once this has it's own issue open ;)

darrenvong (approved):
Looks good 😄

lucaslcode (approved):

lucaslcode:
Merged this as I'm rebasing another pr

PR #2273, opened 2021-12-27
fix: format issues that's failing Prettier on CI + headers with a back button breaking globally across the app

darrenvong:
This is failing the build on `develop` otherwise...

Edit: Actually never mind, the Vercel stuff is probably failing it too, but good to fix it anyway!

Edit2: Lol I guess this has now turned into a bit of random hotfix of things I came across that were broken recently because I was testing out the react query dependabot change. How did I get into this rabbit hole...

**Web frontend checklist**
- [x] Formatted my code with `yarn format && yarn lint --fix` !!!
- [ ] All tests pass

### DIFF
--- a/app/web/src/features/profile/ProfileTagInput.tsx
+++ b/app/web/src/features/profile/ProfileTagInput.tsx
@@ -196,8 +196,8 @@ export default function ProfileTagInput({
       >
         <div className={classes.header}>
           Press 'Enter' to add <br></br>
-          <Link href="mailto:support@couchers.org">Email us</Link> to
-          add a language
+          <Link href="mailto:support@couchers.org">Email us</Link> to add a
+          language
         </div>
         <Autocomplete
           open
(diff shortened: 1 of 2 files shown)


app/web/src/features/profile/ProfileTagInput.tsx
```diff
@@ -195,9 +195,9 @@ export default function ProfileTagInput({
         className={classes.popper}
       >
         <div className={classes.header}>
-          Press 'Enter' to add <br></br>
-          <Link href="mailto:support@couchers.org">Email us</Link> to
-          add a language
+          <Typography>Press 'Enter' to add</Typography>
+          <Link href="mailto:support@couchers.org">Email us</Link> to add a
+          language
```
darrenvong, line 200: Removed the `<br>`s here and wrapped the "Press 'Enter' ..." in `Typography` instead since:
- they should really be two sentences, so the `<br />` isn't use semantically and the spacing is better dealt with by CSS in this case
- it makes internationalising this slightly easier later on - otherwise, we'd need to bundle the whole message into one long string, which I don't think is right for reason above.

lucaslcode (approved):

lucaslcode:
My bad

lucaslcode:
How was the back button broken?

darrenvong:
It was visually broken rather than the button actually broken! All the margins were gone and the back and setting buttons were squashed together

PR #2283, opened 2021-12-31
Add meeting 37 notes

aapeliv:


lucaslcode (approved):
I think you could just omit this...

PR #2288, opened 2021-12-31
Feature preview page and navigation cleanup

aapeliv:
* Adds a "feature preview" page for feature flags/experimental stuff
* Cleans up navigation stuff, so that all menus are defined in a consistent way

I want to use the feature previews:

* Try out the SMS verification stuff before it's rolled out to everyone
* Settings page for new notification system before it replaces the old email-only stuff

**Web frontend checklist**
- [ ] All tests pass


lucaslcode:
Can this wait until next is merged?

aapeliv:
> Can this wait until next is merged?

Sure

app/web/.prettierignore
```diff
@@ -2,3 +2,9 @@ proto/
 node_modules
 **/locales/*.json
 .next/
+coverage/
+junit.xml
+dl-protos.sh
+*.Dockerfile
+*.log
+yarn.lock
```
darrenvong, line 10:
```suggestion
```
We should keep the lock file committed like before!

app/web/features/auth/FeaturePreview.tsx
```diff
@@ -0,0 +1,42 @@
...
+import Section from "features/auth/section/Section";
+import { useTranslation } from "react-i18next";
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  section: {
+    margin: theme.spacing(4, 0),
+    "&:first-of-type": {
+      marginTop: theme.spacing(2),
+    },
```
darrenvong:
```suggestion
```
Seems like this is redundant to me

darrenvong (changes requested):
Only some small things, rest looks good to me! Nice to see this start adding to the translation JSON too 🚀

app/web/.prettierignore
aapeliv: These are all committed, but this just stops prettier from looking at them. It seems when it doesn't find a formatter like for this extension, it'll return an error code

app/web/features/auth/FeaturePreview.tsx
aapeliv: Fixed

app/web/.prettierignore
darrenvong: Ahh yeah my bad! Misread that file as `gitignore`

darrenvong (approved):

PR #2302, opened 2022-01-04
Send report to support team if user leaves bad reference

### LINKED ISSUE #2079: Send report to support team if user leaves a bad reference
i.e. either didn't tick "was appropriate" or rating < 0.5 or whatever you get if you slide to "negative"

aapeliv:
Closes #2079

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -534,10 +534,7 @@ def should_notify(self):
 
         We currently send if expertise is listed, or if they list a way to help outside of a set list
         """
-        return (
-            (self.expertise != None)
-            | (not set(self.contribute_ways).issubset(set(["community", "blog", "other"])))
-        )
+        return (self.expertise != None) | (not set(self.contribute_ways).issubset(set(["community", "blog", "other"])))
```
kalvdans, line 537: `not` can not be overloaded so you can't generate an sql expression from it... i.e it has never worked as a hybrid property. But since you just joined lines it should be equally broken as before :)

app/backend/src/couchers/email/__init__.py
```diff
@@ -56,6 +59,8 @@ def _render_email(template_file, template_args={}):
 
     The body is run through twice, once for the plaintext, once for HTML, and then the HTML version is run through
     render_html above to turn it into HTML.
+
+    If is_system_notification is set, it'll be rendered on a separate system template.
```
kalvdans: Don't use a separate template, it just complicates things.

app/backend/src/couchers/models.py
```diff
@@ -1047,6 +1044,13 @@ class Reference(Base):
         ),
     )
 
+    @hybrid_property
+    def should_report(self):
+        """
+        If this evaluates to true, we send a report to the moderation team.
+        """
+        return (self.rating <= 0.4) | (self.was_appropriate != True)
+
```
kalvdans: Make a regular function instead of a hybrid property. We will never SELECT on this anyway.

kalvdans (changes requested):

app/backend/src/couchers/email/__init__.py
aapeliv: What should I do instead?

app/backend/src/couchers/models.py
aapeliv: OK, I'll just make these `@property`s since this is just complicating things

app/backend/src/couchers/models.py
aapeliv: Thanks, good idea!

app/backend/src/couchers/email/__init__.py
kalvdans: Use `plain/html_base_template`. Or will it break?

aapeliv: No, it's fine... But it just annoys me when the emails come in to the ticket system, etc with that annoying and inaccurate footer.

kalvdans: Which footer, you mean:

> You're receiving this email because you signed up for Couchers.org. If you have any issues, email us at <support@couchers.org>.

aapeliv: Yeah, that one

kalvdans: Live with it :) Rather than having to maintain two highly untested plain/html template files.

I was going to suggest moving the footer to the markdown templates, but I think we broke it out to be able to do a fancy HTML footer.

aapeliv: Okay... 😢

app/backend/src/couchers/models.py
```diff
@@ -1047,6 +1044,13 @@ class Reference(Base):
         ),
     )
 
+    @property
+    def should_report(self):
+        """
+        If this evaluates to true, we send a report to the moderation team.
+        """
```
kalvdans, line 1051: The docstring of a function should not document what the caller is doing. I would have inlined this in the caller instead. But it's minor.

app/backend/src/couchers/models.py
```diff
@@ -1047,6 +1044,13 @@ class Reference(Base):
...
+        return (self.rating <= 0.4) | (self.was_appropriate != True)
```
kalvdans, line 1052:
```suggestion
        return self.rating <= 0.4 or not self.was_appropriate
```
code golfed shorter variant that does the same thing

kalvdans (approved):
Looks good to me, with or without fixing two minor complaints.

PR #2303, opened 2022-01-04
Web/feature/multi line text field

tsuruby:
This PR transforms the TextInputField to a single line input field which grows to multiple lines for long messages. 
It affects input fields under the following route:
`/messages/chats/:id`

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ X] There are no warnings from `yarn lint`
- [ X] All tests pass
- [X ] Clicked around my changes running locally and it works
- [X ] Checked Desktop, Mobile and Tablet screen sizes

### DIFF
--- a/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
@@ -51,6 +51,7 @@ export default function GroupChatSendField({
         multiline
         fullWidth
         onKeyDown={handleKeyDown}
+        size="small"
       />
 
       <Button
--- a/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
+++ b/app/web/src/features/messages/groupchats/GroupChatSendField.tsx
@@ -48,8 +48,6 @@ export default function GroupChatSendField({
         name="text"
         defaultValue={""}
         inputRef={register}
-        minRows={4}
-        maxRows={4}
         multiline
         fullWidth
         onKeyDown={handleKeyDown}
--- a/app/web/src/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/src/features/messages/requests/HostRequestSendField.tsx
@@ -245,9 +245,8 @@ export default function HostRequestSendField({
           inputRef={register}
           multiline
           name="text"
-          minRows={4}
-          maxRows={6}
           onKeyDown={handleKeyDown}
+          size="small"
         />
         <FieldButton
           callback={onSubmit}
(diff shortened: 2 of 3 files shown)


tsuruby:
@lucaslcode I am not sure if /messages/chats/:id is the only place where this input behavior is needed. Could you confirm it?

lucaslcode:
There's also /messages/request/:id

tsuruby:
I applied the new input field behavior on the /messages/request/:id page. However, there is that 'Cancel' button that seems a bit misplaced. Is there a design somewhere where I can check how to position that cancel button?

Screen Shot 2022-01-05 at 6 35 55 PM

lucaslcode:
The cancel is referring to the host request - it would start off as an Accept and Reject button.

This page was redesigned in the mockups and we haven't made the changes in the app yet. In the updated design, the request response buttons are under the last message: 
image

That might be a bit of a challenge with the way the message pages work at the moment (they were some of the first pages we did, so the code is a bit less.... mature).

So the moral is, I think for this PR just give it some margin so it's not right next to the text box :) Everything else looks good!

tsuruby:
I added some margin and I guess this PR is ready to be reviewed @lucaslcode :)

lucaslcode (changes requested):
Looks good! One thing:

For closed requests, the "This request is closed..." message now overflows on mobile as it's quite long. I think we need to move it out of the placeholder and either add it on to the end of the message list, or above the send field (see what you think is better).

Relevant files/lines would be: `HostRequestView.tsx:215` is the list of messages where you could add a "fake" message if the request is closed, and `HostRequestSendField.tsx:119` has the `isRequestClosed` logic which could be moved into a `utils.ts` file in the same folder.

Also, when typing a really long message, the list doesn't stay scrolled to the bottom. But I suspect fixing that is quite complicated (we already have some somewhat complex logic to scroll to the bottom (or not) when sending a message I believe) so let's not worry about that for now.

tsuruby:
Hi @lucaslcode for some reason I can't login in develop anymore. 
I am using Option 2: Target the preview api and backend. 
When I use the URL, like 
http://localhost:3000/login?token=x2I3WKq1...
The login page just gets refreshed but I don't sign in. 
Have you faced this issue before?

lucaslcode:
That's strange. Can you log in at next.couchershq.org?
Also try clearing all your cookies.
We recently merged the NextJS refactor, you may want to rebase your PR (that shouldn't affect login, though). 
If that doesn't work, send me an email.

PR #2307, opened 2022-01-06
Add private text to references

aapeliv:
So you can tip off the mods if there was something bad that happened. Most of that is frontend work.

Missing migraitons for now.

**Backend checklist**
- [x] All tests pass


kalvdans (approved):
Lgtm! (with migrations)

PR #2309, opened 2022-01-07
Add optional prefix to emails and making sender name configurable

### LINKED ISSUE #1711: Add prefix to email subject in dev/next
Get it from an environment variable, so the next emails have a prefix like `[NEXT] Finish signing up for Couchers.org`

aapeliv:
This has been confusing a lot of people on the team.

E.g. @AllhandsNoFeet  getting emails about references on staging (next) and confusing them for prod emails:

Screen Shot 2022-01-07 at 8 03 34 pm

This PR allows setting the displayed sender name and adding a prefix to the email subject, so it's easier to discern prod vs staging emails.

Closes #1711.

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/email/__init__.py
```diff
@@ -95,11 +95,15 @@ def queue_email(sender_name, sender_email, recipient, subject, plain, html):
 
 def enqueue_email_from_template(recipient, template_file, template_args={}):
     frontmatter, plain, html = _render_email(template_file, template_args)
+    if not config["NOTIFICATION_EMAIL_PREFIX"]:
+        subject = frontmatter["subject"]
+    else:
+        subject = config["NOTIFICATION_EMAIL_PREFIX"] + " " + frontmatter["subject"]
```
kalvdans:
```suggestion
    subject = config["NOTIFICATION_EMAIL_PREFIX"] + " " + frontmatter["subject"]
```

and include the space in NOTIFICATION_EMAIL_PREFIX.

kalvdans (changes requested):
Everything looks good except that we a source line that is only run on production ;)

app/backend/src/couchers/email/__init__.py
aapeliv: I just thought it'd be easier this way since trailing spaces can be a bit finicky with env vars? But I can change it

kalvdans: Ah true, but not the end of the world if the space is omitted, it is only for testing.

aapeliv: Good point, I'll change it

aapeliv: Done!

kalvdans (approved):
I can't find the description of docker-compose env file syntax but it looks correct to me! (In shell, you'd have to use single quotes to avoid glob expansion)

PR #2328, opened 2022-01-09
fix: fallback to English if empty string for translation key found

darrenvong:
Add the `returnEmptyString: false` setting back that was in the old config, so that translation keys with an empty string will be treated as missing and falls back to English accordingly. I found this whilst I was poking around locally and enabled another language when I was reviewing #2326 

Technically not an issue yet as we haven't enabled translations other than English at the moment, but best to fix it before it comes up as a surprise.

**Web frontend checklist**
N/A as it's only an i18next config change

### DIFF
--- a/app/web/next-i18next.config.js
+++ b/app/web/next-i18next.config.js
@@ -18,6 +18,7 @@ module.exports = {
     "profile",
     "search",
   ],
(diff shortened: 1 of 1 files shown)


aapeliv (approved):
lgtm

PR #2329, opened 2022-01-09
feat: isolate i18n deps with proxy/adapters

papeldeorigami:
Avoid importing i18n deps like next-i18next and react-18next directly.

**Notice:** After merging this PR, we will still have to update the files that import useTranslation to use the new constants and import the i18n proxy instead. That can be done any time soon because this PR doesn't break compatibility with existing code.

These are the ideas in this PR:
1) instead of importing next-i18next or react-i18next everywhere, I believe it is beneficial to import only the files defined in our project. This way, we could switch between different i18n libraries in the future.
2) Define a generic "translationStaticProps" high order function that can be called in every NextJS page reducing duplicated code as well as isolating the code from i18n dependencies.
3) Replace namespace strings like "communities" or "events" for constants. Usage of plain strings has already proven to be error prone while we were implementing the events translation.


app/web/i18n/index.ts
```diff
@@ -0,0 +1,7 @@
+import {
+  Trans as nextTrans,
+  useTranslation as nextUseTranslation,
+} from "next-i18next";
+
+export const Trans = nextTrans;
+export const useTranslation = nextUseTranslation;
```
darrenvong, line 7:
```suggestion
export { Trans, useTranslation } from "next-i18next";
```
I think you can also do this to simplify the code a bit!

app/web/i18n/namespaces.ts
```diff
@@ -0,0 +1,21 @@
...
+  AUTH,
+  COMMUNITIES,
+  CONNECTIONS,
+  DASHBOARD,
+  DONATIONS,
+  GLOBAL,
+  MESSAGES,
+  PROFILE,
+  SEARCH,
+};
```
darrenvong:
```suggestion
export const NAMESPACES = [
  AUTH,
  COMMUNITIES,
  CONNECTIONS,
  DASHBOARD,
  DONATIONS,
  GLOBAL,
  MESSAGES,
  PROFILE,
  SEARCH,
];
```
In `next-18next.config.js`, it's expecting all namespaces in an array. Maybe we can do it here and re-use it in that config too so this becomes the source of truth?

darrenvong: Or keep this as an object and have this be shared with `i18n/resources.ts`, and generate the namespace array from this in the config file, so they always stay in sync.

A bit of background of when I was made `i18n/resources.ts`: it was done as a convenience to provide custom typing for the `react-i18next-types.d.ts` file, so we get better autocomplete and thus better developer experience. I believe the `i18n/resources.ts` file pattern also helps to avoid incurring the penalty of having to import all of the English translation JSONs in the app, when not all of the JSONs are needed by the app at a particular point in time, as it's only used for TS typing and so should get tree shaked away in the final JS bundle.

darrenvong (commented):
Some minor code comments, but otherwise I like the look of this! I think there's a good chance the `getStaticProps` (or `getServerSideProps` when we end up doing more rendering server-side in some cases TBD) will become more complex as we start to do more with Next.js, so this will help keep the translation specific logic co-located in one spot as well

app/web/i18n/index.ts
papeldeorigami: Thanks! Actually I started with that version but this simplified code doesn't work well with VS Code auto-import feature. So, when trying to import useTranslation, it wasn't showing our i18n.ts file as an option, at least in my editor. I prefer to leave this one as it is.

app/web/i18n/namespaces.ts
papeldeorigami: Nice. I think I prefer to have both the array and the object versions exported from namespaces. I'll update the PR, please let me know if that is ok.

app/web/i18n/index.ts
darrenvong: I see, it is still coming up but annoyingly not as the first option unless something is already being imported from this file. I'd prefer the simpler syntax... but I guess it's not a big deal breaker for now since the file is relatively simple. Maybe one to review again when we add more complex exports here.

app/web/i18n/namespaces.js
```diff
@@ -0,0 +1,44 @@
...
+  AUTH,
+  COMMUNITIES,
+  CONNECTIONS,
+  DASHBOARD,
+  DONATIONS,
+  GLOBAL,
+  MESSAGES,
+  PROFILE,
+  SEARCH,
+};
```
darrenvong: In fact, do we even need all of the namespaces in an object anywhere in our code at the moment? Since they are also individually exported again below, I don't think we need this?

papeldeorigami:
@lucaslcode, the code is ready for merge, in my opinion. We will proceed with changing other files progressively while translations are being done.

app/web/i18n/namespaces.js
papeldeorigami: Oh... you are right. Maybe YAGNI, but it seems to me that having it exported in both formats (object and array) but defined just once wasn't a waste of time and might well be used in some places like we just tried to do in the resources file. It is also more like a pattern I like to follow. Maybe we could remove the lists from the exports section at all, but I see it like different "endpoints" for accessing namespaces (individual, array and object)

darrenvong: The list is being used in the `next-i18next` config though, whereas the object usage I guess is TBD. Should we keeping code if it's YAGNI? :p

Probably one more small thing on naming... should the keys be all uppercase too? i.e. `NAMESPACES.AUTH` instead of `NAMESPACES.auth`? I guess I haven't seen many object properties in uppercase unless it's some "private" value.

darrenvong (changes requested):
Wait, I've focused on the code so far and haven't actually tested until now. On the preview, it shows that the keys are not being interpolated correctly:
Screenshot 2022-01-13 011942

app/web/i18n/server-side-translations.ts
```diff
@@ -0,0 +1,17 @@
...
+  locale: string | undefined,
+  namespaces: Array<string>
+) =>
+  await serverSideTranslations(locale ?? "en", namespaces, nextI18NextConfig);
+
+export const translationStaticProps =
+  (namespaces: Array<string>): GetStaticProps =>
+  async ({ locale }) => ({
+    props: {
+      ...serverSideTranslationProps(locale, namespaces),
```
darrenvong:
```suggestion
      ...(await serverSideTranslationProps(locale, namespaces)),
```
I think it might be because this is not being awaited?

lucaslcode (commented):
Once this is working and Darren has given approval, I'm happy to merge this

app/web/i18n/server-side-translations.ts
papeldeorigami: Oh, I am really sorry about that. I'll test more carefully and update the PR.

app/web/i18n/namespaces.ts
papeldeorigami: Ok, I removed the object and kept just the array for the next-i18next config.

app/web/i18n/namespaces.js
papeldeorigami: Done

app/web/i18n/server-side-translations.ts
papeldeorigami: Yes, that was it. It is fixed now

darrenvong (approved):
Let's get it in 🚀

aapeliv:
@darrenvong you tagged this with release notes, but I don't think it's user-facing? I can just add a line about translations?

PR #2308, opened 2022-01-06
Add a link to write a reference in email notifications

aapeliv:
Continues from #2128.

Closes #1671

**Backend checklist**
- [ ] Formatted my code by running `autoflake -r -i --remove-all-unused-imports src && isort . && black .` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [ ] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Web frontend checklist**
- [ ] Formatted my code with `yarn format && yarn lint --fix`
- [ ] There are no warnings from `yarn lint`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #2345, opened 2022-01-11
Fix user id in reference email

### LINKED ISSUE #2343: You’ve received a reference email has wrong  link
Subject: You’ve received a reference email has wrong  link
Description:
I think it includes the wrong user’s ID in the url

Results:

Backend version: develop-55e3437e
Frontend version: 
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 14_8_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Mobile/15E148 Safari/604.1
Page: https://next.couchershq.org/leave-reference/hosted/1/90
User: aapeli (1)

aapeliv:
It should be the user id of the user you're going to write the reference about

Closes #2343

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/tasks.py
+++ b/app/backend/src/couchers/tasks.py
@@ -196,7 +196,7 @@ def send_host_reference_email(reference, both_written):
         template_args={
             "reference": reference,
             "leave_reference_link": urls.leave_reference_link(
-                "surfed" if surfed else "hosted", reference.host_request.host.id, reference.host_request.conversation_id
+                "surfed" if surfed else "hosted", reference.from_user_id, reference.host_request.conversation_id
             ),
             # if this reference was written by the surfer, then the recipient hosted
             "surfed": surfed,


kalvdans (approved):
Sorry for a sloppy review in https://github.com/Couchers-org/couchers/pull/2308 !

aapeliv:
> Sorry for a sloppy review in #2308 !

I don't think anyone would've caught it! It's so hard to see what the right thing is.

Maybe we should use keyword args for these url functions since otherwise there's no way to guess what they're doing?

kalvdans:
> Maybe we should use keyword args for these url functions since otherwise there's no way to guess what they're doing?

Nah, it is ok as it is now. [edit] Oh, I see you already did a [PR](https://github.com/Couchers-org/couchers/pull/2346) :) will check it

PR #2346, opened 2022-01-11
Make all functions in urls.py accept only keyword arguments

aapeliv:
This helps a little bit avoid bugs like #2343 by making it clearer what's going on at the call site.

Follow up from #2345.

**Backend checklist**
- [ ] All tests pass


kalvdans (approved):
LGTM! Forcing keyword is good habit.

PR #2361, opened 2022-01-15
Fix frontend version setting

lucaslcode:


app/web/prod.Dockerfile
```diff
@@ -42,7 +43,6 @@ EXPOSE 3000
 ENV PORT 3000
 
 ARG version
```
aapeliv: Need to also move this line, I think

aapeliv (approved):
lgtm

PR #2364, opened 2022-01-15
Start enforcing passwords on log in and sign up

aapeliv:
Start of the end of magic login links.

* Enforce password on signup
* Jail users if they don't have a password, so they're forced to set one the next time they log in
* Jail bit in web frontend
* Password field on account form in web frontend

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/web/features/auth/jail/PasswordSection.tsx
```diff
@@ -0,0 +1,116 @@
+import { TextField, Typography, useMediaQuery } from "@material-ui/core";
```
darrenvong: There's a `TextField` component in our `components` folder. If you use that then won't need the `<br />`s to force the input onto a new line.

app/web/features/auth/jail/PasswordSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+      <form className={classes.form} onSubmit={onSubmit}>
+        <TextField
+          id="password"
+          inputRef={register()}
+          label={t("auth:change_password_form.new_password")}
+          name="password"
+          type="password"
+          fullWidth={!isMdOrWider}
+        />
+        <br />
```
darrenvong:
```suggestion
```

app/web/features/auth/jail/PasswordSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+              value === getValues("password") ||
+              t("auth:change_password_form.password_mismatch_error"),
+          })}
+          label={t("auth:change_password_form.confirm_password")}
+          name="passwordConfirmation"
+          fullWidth={!isMdOrWider}
+          type="password"
+          helperText={errors.passwordConfirmation?.message}
+        />
+        <br />
```
darrenvong:
```suggestion
```

app/web/features/auth/jail/PasswordSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  const {
+    error: changePasswordError,
+    isLoading: isChangePasswordLoading,
+    isSuccess: isChangePasswordSuccess,
+    mutate: changePassword,
+    variables: setPasswordVariables,
+  } = useMutation<Empty, GrpcError, SetPasswordVariables>(
+    ({ password }) => service.jail.setPassword(password),
```
darrenvong: Seeing type errors on this in the build - surprised the dev build doesn't complain about this.

```suggestion
  } = useMutation<void, GrpcError, SetPasswordVariables>(
    async ({ password }) => { await service.jail.setPassword(password); },
```
This should fix the type error

app/web/features/auth/jail/PasswordSection.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { theme } from "theme";
+import useChangeDetailsFormStyles from "features/auth/useChangeDetailsFormStyles";
+
+interface SetPasswordVariables {
+  password?: string;
+}
+
+interface SetPasswordFormData extends SetPasswordVariables {
+  passwordConfirmation?: string;
+}
```
darrenvong:
```suggestion
interface SetPasswordVariables {
  password: string;
}

interface SetPasswordFormData extends SetPasswordVariables {
  passwordConfirmation: string;
}
```
I think they should both be required since we aren't letting the user sending through empty passwords in the validations. You've probably copied from the other form and I think it was only optional because we were allowing people to clear their password before :p

That'd fix the other type error on `service.jail.setPassword(...)` complaining about `string | undefined` is not assignable to `string`

app/proto/jail.proto
```diff
@@ -61,3 +66,8 @@ message SetLocationReq {
 message AcceptCommunityGuidelinesReq {
   bool accept = 1;
 }
+
+message SetPasswordReq {
+  // the frontend should ask for the password twice and whatnot
+  string new_password = 1 [ (sensitive) = true ];
```
darrenvong, line 72: What does grpc do to sensitive fields?

(Obviously unrelated to the frontend review, mainly curious about this)

darrenvong (changes requested):
Only some minor things, but frontend mostly looks fine

app/web/features/auth/jail/PasswordSection.tsx
aapeliv: Thanks!

app/proto/jail.proto
aapeliv: It's used in the backend to clear those fields before they're stored in to a debugging table:

https://github.com/Couchers-org/couchers/blob/bfdfb49536d1eb2ee135a37186443f33a3fcc871/app/backend/src/couchers/interceptors.py#L185-L195

app/backend/src/tests/test_auth.py
```diff
@@ -47,7 +47,7 @@ def test_UsernameValid(db):
         assert not auth_api.UsernameValid(auth_pb2.UsernameValidReq(username="")).valid
 
 
-def test_signup_incremental(db):
+def test_signup_incremental(db, fast_passwords):
```
kalvdans: No need to include `fast_passwords` fixture here since it is in the autouse above.

app/backend/src/couchers/servicers/auth.py
```diff
@@ -208,11 +208,8 @@ def SignupFlow(self, request, context):
                     if not self._username_available(request.account.username):
                         context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.USERNAME_NOT_AVAILABLE)
 
-                    if request.account.password:
-                        abort_on_invalid_password(request.account.password, context)
-                        hashed_password = hash_password(request.account.password)
-                    else:
-                        hashed_password = None
+                    abort_on_invalid_password(request.account.password, context)
+                    hashed_password = hash_password(request.account.password)
```
kalvdans, line 212: Remove the `// optional` markup from `SignupAccount::password` in `auth.proto` since it is now required.

kalvdans (commented):
Backend looks fine, two minor comments

app/backend/src/tests/test_auth.py
aapeliv: Good point, thanks

PR #2368, opened 2022-01-15
Build protos in vercel

aapeliv:
This works, just need to change the vercel script to use this script instead of `dl-protos.sh` (see the second last worked vercel build)


lucaslcode (approved):
Amazing! Just have to remember to change this as well when we change `generate_protos.sh`

darrenvong:
> Just have to remember to change this as well when we change `generate_protos.sh`

And when we update the generator / grpc-web's library version?

PR #2373, opened 2022-01-16
Release notes 16th January

aapeliv:
plus stuff from `site` branch


ItsiW (approved):

jesseallhands (approved):
Should we include a note about the updates to the donation page/stripe now enabling tax deductible donations? I know the upcoming blog post will go into more detail about this, but still seems like a significant update to mention.

PR #2390, opened 2022-01-17
Update envoy to v1.21.0

aapeliv:


aapeliv:
Works in local testing and doesn't complain

kalvdans (approved):

PR #2392, opened 2022-01-17
Web/feature/multi line text field

tsuruby:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [X ] There are no warnings from `yarn lint`
- [ ] All tests pass
- [ X] Clicked around my changes running locally and it works
- [X ] Checked Desktop, Mobile and Tablet screen sizes

### DIFF
--- a/app/web/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/features/messages/requests/HostRequestSendField.tsx
@@ -255,9 +255,8 @@ export default function HostRequestSendField({
           inputRef={register}
           multiline
           name="text"
-          minRows={4}
-          maxRows={6}
           onKeyDown={handleKeyDown}
+          size="small"
         />
         <FieldButton
           callback={onSubmit}
--- a/app/web/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/features/messages/requests/HostRequestSendField.tsx
@@ -243,10 +243,10 @@ export default function HostRequestSendField({
       </div>
       <div className={classes.container}>
         <TextField
-          defaultValue={""}
+          defaultValue={isRequestClosed ? REQUEST_CLOSED_MESSAGE : ""}
           disabled={isRequestClosed}
           fullWidth
-          label={isRequestClosed ? REQUEST_CLOSED_MESSAGE : "Message"}
+          label={!isRequestClosed ? "Message" : ""}
           id="host-request-message"
           InputLabelProps={{
             className: isRequestClosed ? classes.requestClosedLabel : undefined,
--- a/app/web/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/features/messages/requests/HostRequestSendField.tsx
@@ -256,6 +256,7 @@ export default function HostRequestSendField({
           multiline
           name="text"
           onKeyDown={handleKeyDown}
+          maxRows={6}
           size="small"
         />
         <FieldButton
--- a/app/web/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/features/messages/requests/HostRequestSendField.tsx
@@ -246,6 +246,7 @@ export default function HostRequestSendField({
           defaultValue={isRequestClosed ? REQUEST_CLOSED_MESSAGE : ""}
           disabled={isRequestClosed}
           fullWidth
+          aria-label={isRequestClosed ? REQUEST_CLOSED_MESSAGE : "Message"}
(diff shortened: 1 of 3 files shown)


tsuruby:
@lucaslcode regarding the overflowing label on mobile in case of a closed request: 
Instead of using it as a label I chose to show it as a defaultValue:

`defaultValue={isRequestClosed ? REQUEST_CLOSED_MESSAGE : ""}`

And only if the request is not closed we show the label:

`label={!isRequestClosed ? "Message" : ""}`

Wdyt?

darrenvong:
This is maybe a bit of an edge case... but the main point is, with this change, now the textbox grows forever and ever and can lead to situation like this, where the chat itself is no longer visible because the message box takes up the whole space:
Screenshot 2022-01-17 230454

I think we probably still need `maxRows` set to be some upper bound to cap the box's growth after a certain number of lines?

tsuruby:
Good point @darrenvong ! I added the maxRow prop. Thank you for mentioning it. However it would be best if we could shift the content so we don't need this limitation.

darrenvong:
What do you mean by shifting the content? I think I don't completely follow 😅

Maybe the height of the input can be more generous, but I think there'll still need to be a minimum height such that the chat content can't get under, even if the scrollbar auto scrolls to the right position, otherwise it may be quite frustrating to view messages over such a short viewport/box. Of course if you have a better idea or maybe I misunderstood you, I'd be interested to hear your suggestion.

app/web/features/messages/requests/HostRequestSendField.tsx
```diff
@@ -243,10 +243,10 @@ export default function HostRequestSendField({
       </div>
       <div className={classes.container}>
         <TextField
-          defaultValue={""}
+          defaultValue={isRequestClosed ? REQUEST_CLOSED_MESSAGE : ""}
           disabled={isRequestClosed}
           fullWidth
-          label={isRequestClosed ? REQUEST_CLOSED_MESSAGE : "Message"}
+          label={!isRequestClosed ? "Message" : ""}
```
darrenvong, line 250: I think this would cause accessibility issues for non-visual users, as ideally they should be able to find the field as disabled with the "Message" label when the request is closed, which would not be possible if we render an empty string in that case.

The best way to solve this without changing the visible UI might be via an `aria-label` with the same label value.

tsuruby: I added an aria-label.

tsuruby:
> What do you mean by shifting the content? I think I don't completely follow 😅
> 
> Maybe the height of the input can be more generous, but I think there'll still need to be a minimum height such that the chat content can't get under, even if the scrollbar auto scrolls to the right position, otherwise it may be quite frustrating to view messages over such a short viewport/box. Of course if you have a better idea or maybe I misunderstood you, I'd be interested to hear your suggestion.

I don't understand why there should be a minimum height? I thought the idea is to start with a one-row input field and let it grow with its content.

darrenvong:
Oh I see you meant the minimum height for the chat input box - yeah, I'm not sure why it couldn't start with one line (its absolute minimum without the text being cut off) and grow from there

app/web/features/messages/requests/HostRequestSendField.tsx
darrenvong: Sorry if I wasn't being clear - it's better to stick with the same "Message" label since the input box hasn't changed its purpose, but disabled when the request is closed.

tsuruby:
Could you take a screenshot maybe so I can see what exactly the issue is 😅 That would be great 💪 :)

app/web/features/messages/requests/HostRequestSendField.tsx
tsuruby: Ah I see, got it 😅 I updated it, so the aria-label is always "Message"

darrenvong (approved):

darrenvong:
That was my imprecise way to say "I agree with you" (that the input should grow from one line), so I don't have any issues screenshot to show you!

tsuruby:
Ah 😅 sorry, my bad 🙈 I guess I was too focused on issues :D

tsuruby:
The last time @lucaslcode merged my PR, so I am not sure about how to proceed now. Am I good to merge it?

aapeliv:
> The last time @lucaslcode merged my PR, so I am not sure about how to proceed now. Am I good to merge it?

Yes, please merge!

PR #2394, opened 2022-01-18
Global Translation Keys

lucaslcode:
There is really a lot stuffed in here, like most of communities and profiles...

**Web frontend checklist**
- [x] All tests pass


darrenvong:
Are you thinking of moving the constants in the `components` folder into the global namespace too? I just found things like "terms" that's only used by the footer should really go with it, but instead it's currently in the auth namespace...

app/web/components/constants.ts
```diff
@@ -1,10 +1,5 @@
 export const CANCEL_UPLOAD = "Cancel upload";
 export const CONFIRM_UPLOAD = "Confirm upload";
```
darrenvong, line 2: ~~Where are these being used at the moment?~~

Actually, can see these being used in the image input. Shouldn't we move this to the global namespace too under the `components` object?

darrenvong (commented):
Mostly looks good from a quick glance, will have a quick test too when I get some time!

What do you think about moving the footer constants into the global namespace too? 

There's also the contributor form which I realised I didn't do as it was in the `components` folder, but it's a bit of a weird one to put in the global namespace. Maybe more suitable for the profile or auth namespace?

app/web/components/constants.ts
```diff
@@ -1,10 +1,5 @@
 export const CANCEL_UPLOAD = "Cancel upload";
```
darrenvong, line 1: What about the converting the constants in this file? Or are you planning to do this separately?

app/web/features/profile/locales/en.json
```diff
@@ -65,5 +65,12 @@
     "message": "Message",
     "default_about_me_headings": "# Current mission\n<br>\ne.g. I want to meet people and learn...\n<br>\n# Why I use Couchers.org\n<br>\ne.g. I think couch surfing should be free...\n<br>\n# My favorite travel story\n<br>\ne.g. One time I...\n",
     "default_hobbies_headings": "# Art\n<br>\ne.g. Painting, ...\n<br>\n# Books\n<br>\neg. The Cat in the Hat, ...\n<br>\n# Movies\n<br>\ne.g. David Attenborough documentaries, ...\n<br>\n# Music\n<br>\ne.g. The Wiggles, ...\n<br>\n",
-    "default_about_home_headings": "# What I can share with guests\n<br>\ne.g. I can share...\n<br>\n"
+    "default_about_home_headings": "# What I can share with guests\n<br>\ne.g. I can share...\n<br>\n",
```
darrenvong, line 68: I wonder if it's better to break these up in separate strings and have a function or something in a private namespace that piece them back together to make these more translator friendly.

app/web/test/utils.tsx
```diff
@@ -59,7 +61,7 @@ export function createMatchMedia(width: number) {
 }
 
 export function t(
-  translationKey: string,
+  translationKey: Parameters<TFunction>[0],
```
darrenvong: Does this help much with autocomplete anyway?

app/web/features/communities/events/EventPage.tsx
```diff
@@ -130,7 +129,7 @@ function getEventTimeString(
   const start = dayjs(timestamp2Date(startTime));
   const end = dayjs(timestamp2Date(endTime));
 
-  return `${start.format("LLLL")} ${TO} ${end.format(
+  return `${start.format("LLLL")} to ${end.format(
     end.isSame(start, "day") ? "LT" : "LLLL"
   )}`;
 }
```
darrenvong, line 135: I guess technically unrelated to this namespace, but might be worth getting @papeldeorigami who's worked on the `communities` namespace to follow up on converting this one as this should probably an interpolated string.

app/web/test/utils.tsx
lucaslcode: No but just managed to fix it!

app/web/components/constants.ts
lucaslcode: Maybe separately

app/web/features/communities/events/EventPage.tsx
lucaslcode: yes, I didn't want to do it in this PR

app/web/features/profile/locales/en.json
lucaslcode: I think they'll manage with some guidance. Don't want to over-complicated at the moment

darrenvong (commented):
Vercel build says there's some build error reading from `next.config.js`?

app/web/test/utils.tsx
```diff
@@ -59,7 +61,9 @@ export function createMatchMedia(width: number) {
 }
 
 export function t(
-  translationKey: string,
+  translationKey: Parameters<
+    TFunction<(keyof typeof resources)[], undefined>
+  >[0],
```
darrenvong, line 66: Love a bit of TypeScript 🧙 there when it works. Completely overlooked that `i18next` expose these types where most of the magic have been done by them!

app/web/components/Navigation/ReportButton.tsx
```diff
@@ -145,77 +131,77 @@ export default function ReportButton({
...
+                {t("cancel")}
               </Button>
             </DialogActions>
           </>
         ) : type === "content" ? (
           <>
             <DialogContent>
               <Typography variant="body1" paragraph>
-                {REPORT_CONTENT_MESSAGE}
+                {t("cancel")}
```
darrenvong:
```suggestion
                {t("report.content.message")}
```
Since it's completely broken at the moment with the wrong text

app/web/features/ContributePage.tsx
```diff
@@ -1,15 +1,17 @@
...
 import HtmlMeta from "components/HtmlMeta";
 import PageTitle from "components/PageTitle";
 import { CONTRIBUTE_TITLE } from "features/dashboard/constants";
+import { useTranslation } from "i18n";
+import { GLOBAL } from "i18n/namespaces";
 
 export default function ContributePage() {
+  const { t } = useTranslation(GLOBAL);
   return (
     <>
```
darrenvong, line 13: Unrelated to this file really but the closest where I can comment, but looks like `pages/contribute.tsx` is missing the `getStaticProps` stuff needed to load the translations properly in some places like the nav bar (shown in screenshot) and the footer:
Screenshot 2022-03-07 at 22 51 09

app/web/features/NotFoundPage.tsx
```diff
@@ -25,21 +21,24 @@ const useStyles = makeStyles((theme) => ({
...
           className={classes.graphic}
         />
-        <Typography>{NOT_FOUND}</Typography>
+        <Typography>{t("not_found_text_1")}</Typography>
         <Typography>
-          {DO_YOU_WANT}
-          <StyledLink href={baseRoute}>{GO_HOME}</StyledLink>
+          <Trans t={t} i18nKey="not_found_text_2">
+            Do you just want to
+            <StyledLink href={baseRoute}>go home?</StyledLink>
```
darrenvong: When I'm logged in, this takes me to the unauthenticated home page instead of the dashboard:
Screenshot 2022-03-07 at 22 54 27

Isn't this meant to be fixed by the middleware re-write? 🤔

darrenvong (changes requested):
Found a few more issues from a general bug bash around the areas changed, but looks good otherwise, should be the last one!

app/web/features/NotFoundPage.tsx
darrenvong: To be fair, think I'm going to set up a separate issue for this as strictly not related to the translation stuff

darrenvong: Interesting that this is actually working as expected in prod. I wonder what's different here - outdated config?

lucaslcode: Not sure why it works, but middleware doesn't apply to internal links I think. I added a check to make it link to dashboard for logged-in users

app/web/components/Navigation/ReportButton.tsx
lucaslcode: Whoops

app/web/features/NotFoundPage.tsx
```diff
@@ -23,6 +25,11 @@ const useStyles = makeStyles((theme) => ({
 export default function NotFoundPage() {
   const { t } = useTranslation(GLOBAL);
   const classes = useStyles();
+  const {
+    authState: { authenticated },
+  } = useAuthContext();
+  const [isMounted, setIsMounted] = useState(false);
+  useEffect(() => setIsMounted(true), []);
```
darrenvong, line 32: There's an `useIsMounted` hook in `utils/hooks` we can use instead?

darrenvong, line 32: There's an `useIsMounted` hook in `utils/hooks` we can use instead?

darrenvong (changes requested):

app/web/features/NotFoundPage.tsx
lucaslcode: It doesn't work in this case since it uses a ref and not state, so I think the component wasn't re-rendering

darrenvong: Does it need to re-render though if at the logic bit, it's already reading the up-to-date value from `.current` for `mounted`? And I believe the `authState` stuff is already in state so that should be fine?

lucaslcode: It always showed the `baseRoute` when I tried using the existing hook. I assumed it was because it would render and `isMounted.current` would be false, it would then update to true but not render again because it's a ref.

darrenvong: Ahh fair enough, would be interesting to see if we encounter this enough and whether it's worth updating the original hook in the future

darrenvong (approved):

PR #2397, opened 2022-01-18
Phone number validation

aapeliv:
Web frontend for phone number validation

@kalvdans: fixed some things in the backend on non-happy path.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

# Screenshots:

Screen Shot 2022-01-18 at 7 06 36 pm
Screen Shot 2022-01-18 at 7 06 48 pm
Screen Shot 2022-01-18 at 7 06 57 pm
Screen Shot 2022-01-18 at 7 07 02 pm

### DIFF
--- a/app/backend/src/couchers/phone/sms.py
+++ b/app/backend/src/couchers/phone/sms.py
@@ -33,8 +33,6 @@ def send_sms(number, message):
 
     if not config["ENABLE_SMS"]:
         logger.info(f"SMS not enabled, need to send to {number}: {message}")
-        if config["DEV"]:
-            return "success"
         return "SMS not enabled."
 
     sns = boto3.client("sns")
--- a/app/backend/src/couchers/phone/sms.py
+++ b/app/backend/src/couchers/phone/sms.py
@@ -1,3 +1,5 @@
+import logging
+
 import boto3
 import luhn
 
@@ -5,6 +7,8 @@ from couchers import crypto
 from couchers.config import config
 from couchers.db import session_scope
 from couchers.models import SMS
+
+logger = logging.getLogger(__name__)
 
 
 def generate_random_code():
@@ -25,11 +29,13 @@ def send_sms(number, message):
     success, "unsupported operator" on unsupported operator, and any other
     string for any other error."""
 
-    assert len(message) <= 140, "message too long"
+    assert len(message) <= 140, "Message too long"
 
     if not config["ENABLE_SMS"]:
         logger.info(f"SMS not enabled, need to send to {number}: {message}")
-        return
+        if config["DEV"]:
+            return "success"
+        return "SMS not enabled."
 
     sns = boto3.client("sns")
     sender_id = config["SMS_SENDER_ID"]
--- a/app/web/features/auth/FeaturePreview.tsx
+++ b/app/web/features/auth/FeaturePreview.tsx
@@ -49,7 +49,7 @@ export default function FeaturePreview() {
         <>
           <ChangePhone
             className={classes.section}
-            {...(accountInfo as GetAccountInfoRes.AsObject)}
+            accountInfo={accountInfo!}
           />
         </>
       )}
--- a/app/web/features/auth/FeaturePreview.tsx
+++ b/app/web/features/auth/FeaturePreview.tsx
@@ -1,11 +1,15 @@
-import { Typography } from "@material-ui/core";
-import { Alert as MuiAlert } from "@material-ui/lab/";
+import { CircularProgress, Typography } from "@material-ui/core";
+import { Alert, Alert as MuiAlert } from "@material-ui/lab/";
 import HtmlMeta from "components/HtmlMeta";
 import PageTitle from "components/PageTitle";
 import NotificationSettings from "features/auth/notifications/NotificationSettings";
-import { useTranslation } from "i18n";
 import { AUTH } from "i18n/namespaces";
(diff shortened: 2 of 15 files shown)


app/backend/src/couchers/phone/sms.py
```diff
@@ -25,11 +29,13 @@ def send_sms(number, message):
...
 
-    assert len(message) <= 140, "message too long"
+    assert len(message) <= 140, "Message too long"
 
     if not config["ENABLE_SMS"]:
         logger.info(f"SMS not enabled, need to send to {number}: {message}")
-        return
+        if config["DEV"]:
+            return "success"
+        return "SMS not enabled."
```
kalvdans: You're testing one config variable too many here. Why do you want to return success when no sms was sent?

aapeliv: So that it's possible to test it by clicking around locally. Otherwise all the api calls fail.

kalvdans: ok, fair enough. The testconfig autofixture sets DEV so I'm not sure why your added test works... will look into it.

kalvdans: aha, test didn't work :) Will have another look after you test run before commit :D

app/web/features/auth/FeaturePreview.tsx
```diff
@@ -30,6 +40,18 @@ export default function FeaturePreview() {
...
       <NotificationSettings className={classes.section} />
+      {isAccountInfoLoading ? (
+        <CircularProgress />
+      ) : accountInfoError ? (
+        <Alert severity="error">{accountInfoError.message}</Alert>
+      ) : (
+        <>
+          <ChangePhone
+            className={classes.section}
+            {...(accountInfo as GetAccountInfoRes.AsObject)}
```
darrenvong:
```suggestion
            accountInfo={accountInfo}
```
It's more preferred to do pass props explicitly than spreading, as it's quite opaque what props the component accepts otherwise, as shown by the need to type assert...

app/web/features/auth/phone/ChangePhone.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+interface VerifyPhoneFormData {
+  code: string;
+}
+
+type ChangePhoneProps = GetAccountInfoRes.AsObject & {
+  className?: string;
+};
+
+export default function ChangePhone(props: ChangePhoneProps) {
```
darrenvong: Might be worth breaking the various forms into its own component... it'd make testing the various UI states easier too for each form.

app/web/features/auth/phone/ChangePhone.tsx
```diff
@@ -0,0 +1,258 @@
...
+  phone: string;
+}
+
+interface VerifyPhoneFormData {
+  code: string;
+}
+
+type ChangePhoneProps = GetAccountInfoRes.AsObject & {
+  className?: string;
+};
```
darrenvong:
```suggestion
interface ChangePhoneProps {
  accountInfo: GetAccountInfoRes.AsObject;
  className?: string;
};
```
To promote the prop being passed into the component explicitly

darrenvong (changes requested):
Would be good to see some tests for the `ChangePhone` component too 😛

darrenvong:
Also, looks like there's some compile error from the existing frontend tests because of the proto contract change in the `accountInfo` query

darrenvong:
I got the code but not the UI to enter the verification code. And now I'm just stuck with this error:
Screenshot 2022-02-09 at 01 39 54

aapeliv:
> I got the code but not the UI to enter the verification code. And now I'm just stuck with this error:
> Screenshot 2022-02-09 at 01 39 54

That's because the API change; I made some small changes here to the verification output, so it won't ever get to the next phase, I don't think.

darrenvong:
Also a minor thing, but would be nice if the input matches the styling of the rest of the app eventually. But I get this page is for incubating experimental features so maybe it's fine to leave it for now and for it to look a bit rough :p

aapeliv:
> Also a minor thing, but would be nice if the input matches the styling of the rest of the app eventually. But I get this page is for incubating experimental features so maybe it's fine to leave it for now and for it to look a bit rough :p

Yeah, I know. I'm sorry. I just ran out of steam on this, I lack the persistence on frontend stuff...

darrenvong:
So if the API's changed... is there a way to advance to the "pending" state without being stuck? Is that being reflected in the new API response in some way?

aapeliv:
> So if the API's changed... is there a way to advance to the "pending" state without being stuck? Is that being reflected in the new API response in some way?

No, if you try too many times then you're stuck having to wait until you can try again. The rate limits are kind of draconian.

darrenvong:
What's the rate limit on this thing? I still got the same error trying now even though it's been over 24 hours ago.

Also the rate limit is not the issue since I got the verification code on my phone (not sure that's still valid by now though??), but the UI to enter the verification code wasn't showing up for me.

Or I should say, it's still not showing up for me, but the backend clearly knows I've entered a number before if it's showing that error...

aapeliv:
Yes, because you're testing against `next` which doesn't follow the right state machine.

I can separate out the backend stuff and deploy so you can test.

The rate limit is long... months to try again. Maybe we should have a shorter rate limit though.

PR #2401, opened 2022-01-20
WIP: auth translation key tidy

darrenvong:
Relates to Couchers-org/couchers#4152 for the "auth" feature area. This PR does another sweep to refactor things based on things we recently agreed on (e.g. only importing from `i18n` rather than the module directly), as well as aiming to move some straggler components over to using the JSON file that were missed the first time (looks like there were still quite a few 😢 )

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/features/auth/Settings.tsx
+++ b/app/web/features/auth/Settings.tsx
@@ -7,16 +7,9 @@ import { ChangePassword } from "features/auth/password";
 import Section from "features/auth/section/Section";
 import Timezone from "features/auth/timezone/Timezone";
 import Username from "features/auth/username/Username";
-import { GetAccountInfoRes } from "proto/account_pb";
+import { useTranslation } from "i18n";
 import makeStyles from "utils/makeStyles";
 
-import {
-  ACCOUNT_SETTINGS,
-  CHANGE_BIRTHDATE,
-  CHANGE_BIRTHDATE_CONTACT,
-  CHANGE_GENDER,
-  CHANGE_GENDER_CONTACT,
-} from "./constants";
 import useAccountInfo from "./useAccountInfo";
 
 const useStyles = makeStyles((theme) => ({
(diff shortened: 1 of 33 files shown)


app/web/features/auth/Settings.tsx
```diff
@@ -39,42 +33,43 @@ export default function Settings() {
...
-      <HtmlMeta title={ACCOUNT_SETTINGS} />
-      <PageTitle>{ACCOUNT_SETTINGS}</PageTitle>
+      <HtmlMeta title={t("account_settings_page.title")} />
+      <PageTitle>{t("account_settings_page.title")}</PageTitle>
       <Section
         className={classes.section}
-        title={CHANGE_GENDER}
-        content={CHANGE_GENDER_CONTACT}
+        title={t("account_settings_page.gender_section.title")}
+        content={t("account_settings_page.gender_section.explanation")}
```
tsuruby, line 42: It is probably not scope of this ticket but I saw that for example in `features/donations` we use "body" instead of "explanation".  And for "jail" it is "description".

app/web/features/auth/Settings.tsx
```diff
@@ -29,6 +22,7 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 export default function Settings() {
+  const { t } = useTranslation("auth");
```
tsuruby:
```suggestion
  import { AUTH } from "i18n/namespaces";
  const { t } = useTranslation(AUTH);
```

darrenvong: Ahh nice catch, thanks! 👍

app/web/features/auth/Settings.tsx
darrenvong: That's a good point, maybe we should have some convention on standardising how we name these things... but that can be tricky as it really depends on the context of the component.

For example with this, I'm trying to name all of the sections consistently with "title", "description" and "explanation"... except the gender and birthdate sections don't have a description. I guess the "explanation" could change to something else, so maybe could name it `description2` or something like that.

darrenvong: But I'll be looking at the donations keys next, so I'll change it there if we come to some agreement on what's best :)

tsuruby (approved):
lgtm

PR #2429, opened 2022-01-24
Authentication strings reworked

comradekingu:
This will help translations greatly.
Generally shorter, more consistent, and made to produce fewer errors in translation.

Particularly "gender identity" is not something Couchers is privy to or have any business recording. The terms of service "must be accurate" even undermine its departure from biology.

Edit: TL; DR The only reasonable claim to recording sex is for safety.
The difference in strength between the two is an immutable characteristic.
There is an admission of both in that only admins can change it.
Identity is not sex, but still it is being recorded.
This puts Couchers.org in a perilous position to explain why it collects
this info about citizens in EU-GDPR covered jurisdictions.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/auth/locales/en.json
+++ b/app/web/features/auth/locales/en.json
@@ -53,7 +53,7 @@
       "field_label": "Username",
       "required_error": "Enter your username",
       "username_taken_error": "This username is already in use.",
-      "validation_error": "Usernames can only have numbers or _, starting with a letter."
+      "validation_error": "Usernames can only have numbers or _, and have to start with a letter."
     },
     "password": {
       "field_label": "Password",
--- a/app/web/features/auth/locales/en.json
+++ b/app/web/features/auth/locales/en.json
@@ -1,43 +1,43 @@
 {
   "change_email_form": {
-    "title": "Change Email",
-    "current_email_message": "Your email address is currently <1>{{email}}</1>.",
-    "success_message": "Your email change has been received. Check your new email to complete the change.",
+    "title": "Change E-mail",
+    "current_email_message": "Your e-mail address is currently <1>{{email}}</1>.",
+    "success_message": "Check your old e-mail account to complete the change to the new one.",
     "current_password": "Current password",
-    "new_email": "New email"
+    "new_email": "New e-mail address"
   },
   "change_email_confirmation": {
-    "change_in_progress": "Email change in progress...",
-    "success_message": "Your email has been changed successfully!",
-    "requires_confirmation_from_new_email": "Thank you for confirming your old email address. To complete the change, you still need to confirm your new email address.",
-    "requires_confirmation_from_old_email": "Thank you for confirming your new email address. To complete the change, you still need to confirm your old email address.",
-    "invalid_confirmation_state": "Invalid email confirmation state",
-    "error_message": "Error changing email: {{message}}"
+    "change_in_progress": "Changing e-mail address…",
+    "success_message": "Your e-mail address has been changed.",
+    "requires_confirmation_from_new_email": "Thank you for confirming your old e-mail address. To complete the change, you still need to confirm your new e-mail address.",
+    "requires_confirmation_from_old_email": "Thank you for confirming your new e-mail address. To complete the change, you still need to confirm your old e-mail address.",
+    "invalid_confirmation_state": "Invalid e-mail confirmation state",
+    "error_message": "Could not change e-mail address: {{message}}"
   },
   "change_password_form": {
     "title": "Change password",
     "confirm_password": "Confirm password",
-    "password_changed_success": "Your new password has been set and a notification was sent to your email.",
+    "password_changed_success": "You can now use your new password. (A notification was sent to your e-mail address.)",
     "old_password": "Old password",
     "new_password": "New password",
-    "password_mismatch_error": "This does not match the new password you typed above."
+    "password_mismatch_error": "Mismatching passwords."
   },
   "jail_set_password_form": {
     "title": "Set a password",
-    "explanation": "You have not previously set a password for your account. Please create one now to continue."
+    "explanation": "You have not set a password for your account. Please create one to continue."
   },
   "reset_password_form": {
-    "enter_email": "Enter your username/email",
-    "success_message": "Check your email for a reset password link!"
+    "enter_email": "Enter your username/e-mail address",
+    "success_message": "Check your e-mail for a link to reset your password."
   },
   "login_page": {
     "title": "Log in",
-    "header": "Welcome back!",
+    "header": "Welcome back.",
     "no_account_prompt": "No account yet? <2>Sign up</2>",
     "form": {
-      "username_field_label": "Email/Username",
+      "username_field_label": "Username/e-mail address",
       "password_field_label": "Password",
-      "no_password_login_prompt": "Check your email for a link to log in! :)",
+      "no_password_login_prompt": "Check your e-mail for a link to log in! :)",
       "remember_me": "Remember me",
       "forgot_password": "Forgot password?"
     }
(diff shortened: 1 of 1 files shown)


app/web/features/auth/locales/en.json
```diff
@@ -1,96 +1,95 @@
...
   },
   "account_form": {
     "header": "Your account details",
     "username": {
       "field_label": "Username",
       "required_error": "Enter your username",
-      "username_taken_error": "This username is taken.",
-      "validation_error": "Username can only have numbers or _, starting with a letter."
+      "username_taken_error": "This username is already in use.",
+      "validation_error": "Usernames can only have numbers or _, starting with a letter."
```
comradekingu:
```suggestion
      "validation_error": "Usernames can only have numbers or _, and have to start with a letter."
```

aapeliv (changes requested):
Hi @comradekingu, I appreciate you taking the time to look through these. Here's some comments:

* Do not change the language around gender identity and sex. It is good and as as intended, non-binary isn't there by accident! I understand you may take issue with this and have your own beliefs about gender and sex; but we want to build Couchers into an inclusive community and allow people to express themselves fully, and we're not changing this.
* Let's keep email without a hyphen; with the hyphen it sounds too formal/archaic
* Please don't change the meaning of phrases, e.g. it's "Free forever" very purposefully and means a completely different thing to "Libre software."
* Please don't change/remove the "friendly touch". E.g. when we say "You're all done!", we do it on purpose, changing it to "All done." removes that.

comradekingu:
@aapeliv 1. The matter as it relates to admins having to change it is due to a concern for safety.
There is no legitimate interest for collecting and storing data outside of this scope, so it is not lawful.
The beliefs of anyone is irrelevant to this fact, because it doesn't amount to immutable characteristics.
You end up in a scenario having to explain "have to be accurate" as completely indistinguishable from made up.
The "we" that want to do something somehow doesn't include me.
What does "inclusive" mean when you say it?

2. The sound is the same, and this strictly more correct variant causes fewer errors in translation since English has more leniency.
3. "Purposefully" what? "Free forever" could either mean perpetually gratis, or free as in freedom (free software), in the sense of copyleft here.
Couchers is [specifically neither](https://github.com/Couchers-org/couchers/blob/develop/license.md).
>**Permission is hereby granted**, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, **and/or sell copies of the Software**

and
>"**to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so**"

I don't know why the Expat license was chosen, nor why the ambiguous language made it in.
It is strange to extol "free forever" merely as a promise, contrary to the actual licensing.
4. Ending sentences in exclamation marks on at least 13 occasions is not a friendly gesture. Particularly not in commands directed at the user. The resulting effect in translation is worse yet.

aapeliv:
> 1.

This is not something that's up for discussion, it has been decided by the product teams a long time ago, and I'm not going to justify it here.

> 2. The sound is the same, and this strictly more correct variant causes fewer errors in translation since English has more leniency.

People use "email" in casual contexts these days, "e-mail" is much more formal and archaic.

> 3. "Purposefully" what? "Free forever" could either mean perpetually gratis, or free as in freedom (free software), in the sense of copyleft here.

If you don't know what it means, then you should not pick an interpretation and change it. Anyway, in this context it mostly means perpetually gratis, but we also believe in libre software.

> 4. Ending sentences in exclamation marks on at least 13 occasions is not a friendly gesture. Particularly not in commands directed at the user. The resulting effect in translation is worse yet.

OK, could you then remove the extra excalamtion marks while leaving the tone unchanged, please?

comradekingu:
@aapeliv 
I don't know what decisions product teams might have made. Anything is up for discussion even if you aren't arguing the point.
At-least insofar as 
>The new alternative to Couchsurfing™
Free forever. **Community‑led.** Non‑profit. Modern.

goes. I am forwarding what I think is a legal issue, beyond the reach of marketing.
How much time has passed since a decision was made is to that end irrelevant.
Also serving a EEC market without prominently displaying EU-GDPR info is a huge risk.

"E-mail" isn't much more formal, and far from archaic. Both are actively used.
In English it is at best (in favour of email) something that amounts to a preference, whereas in other languages an active error is avoided. Since the latter is the bigger issue, it mitigates the issue. It is hard enough to translate all the resources.
Also, the lines between an e-mail, and "e-mail address" are blurred. This sequence is in use:
"New email"
"Email change in progress..."
"Your email has been changed successfully!"

This relies completely on visual clues and context. To make matters worse, users of screen-readers will hear "dot-dot-dot".
The statement of "successfully" means either it could happen in an unsuccessful fashion, or that there is something extraordinary (!) about the event. etc.

I suggest instead to focus on things like "Unhandled signup flow state." which is perhaps more technical than it has to be.
The rest of the string-base is in a similarly sorry state, so it is possible to return to other issues once the immediate ones are fixed.

I do know what "free" means, as described. This is what I have to go by:
>it's "Free forever" very purposefully and means a completely different thing to "Libre software."

It is impossible to discern your meaning of "gratis", (as opposed to libre) there.

>Anyway, in this context it mostly means perpetually gratis, but we also believe in libre software.

No, that context is lost. For a product that is "free software", the duality is best avoided.
Having to pick in languages other than English introduces an active error in translation.
This is because (often) a choice limited to one or the other has to be made, and there is thus no possibility of leaving it
up to interpretation.

As for the "forever", (as mentioned) the license is specifically neither forever libre, nor forever gratis.
Yes, the claim is made right below "The new alternative to Couchsurfing™", 
bilde

I think it misleads the audience to restrict its meaning to be that of the brand name or this trademark for what is otherwise exactly the same.
Intrinsically, the alternative to Couchsurfing is not branding or trademarks, and it seems the ™ is there only to distinguish
"Couchsurfing™". So we are left with the brand itself (couchers.org), and that is neither new or alternative in any interesting capacity.

TL;DR Things need to have meaning in a conveyed sense, and ideally be meaningful beyond that.
It can't just sound good and be mutually exclusive.

aapeliv:
You are not the only member of the Couchers.org and couch surfing community: the project can be community-led without things having been done in the way that you personally believe they should be done.

Re emails: I agree, it's probably better to use "email address" rather than "email" alone. Yes, "email" vs "e-mail" is a question of preference.

Re libre/gratis: yes, I'm aware that the context isn't fully there in the short byline, we believe most people will get the right idea. Our beliefs about making the software free to use forever, as well as producing it in public and as libre software are explained further elsewhere.

I am aware of what the MIT license does and does not grant. We chose a permissive license on purpose: it gives more developers and organizations the ability to reuse the code we produce, not limiting it to a single-use or e.g. GPL-only open-source projects.

Sustainability of the project itself is built in other ways, not by picking a license.

Could you please cooperate and change the gender/sex language back to how it was?

comradekingu:
@aapeliv Me not being the only member of the community (or the translation team) is not in dispute, and it was never contested.
You are citing prior decision of "product teams" for decisions relating to the platform, which is a marketing-driven and radical approach. As limited to the platform, it runs afoul of:
bilde

To re-iterate the problem of "email". If this is translated without the hyphen, it is most often an active error.
This is far more likely when it is omitted, as opposed to carrying an extra hyphen over into a word that isn't supposed to have one. All the strings on https://translate.couchershq.org/projects/couchers/ have no added context.

The context of gratis/libre, or your particular intended split isn't just not fully there, it isn't there at all.
Believing it will forever be free to use forever either in gratis or libre form is not the same as the claim that it will be.
What it actually is happens to be different.

If GPL is a single-use, you prevent it with Expat, because it is perpetual.
Every libre license is available to exactly as many people and organisations, you are just preventing the ones that don't like your own advertisement. Since you picked a libre license, it can't be non-commercial.
The sustainability of couchers.org is beside the point, this relates to claims made.

Why would I change the language back to what in my view is a legal error threatening the platform, and an opened topic of contention within the community?
Happy to cooperate. Not interesting in adhering to behind the scenes decisions with no public transparency.

aapeliv:
Okay! We appreciate your effort and good intensions here, however, unfortunately we can't accept these changes.

comradekingu:
@aapeliv Who is "we"?

ItsiW:
The action to close this PR is fully supported by the support team, and is in line with the broad and long discussions that have occurred with our members in conversations and forums

comradekingu:
@ItsiW I am merely questioning who the "we" making actual decisions is. One can't be privy as to "why", I think it answers all questions as to why not. That level of honesty helps in my view, and I am not in a position to change it.
I just wanted to know how much truth there were to claims made.

I have read those discussions, and participated in some. Didn't see a "support team" or the deciding "product teams" chime in.
The forum is and will be an arena for something anyone can pretend is in line with closing down suggestions without discussion.
The argument as moved over to the forum is the first thing users see, which will probably continue, but my contributions won't.
Even not registering this data would do, but "we" opted to double-down on disregarding a practice of not allowing people to express their "identity" freely, over what has to be a concern for safety in even recording it in the first place.
Pretend that is not a legal concern or rooted in biology all you like. The end result is a counterproductive hypocrisy at your peril.

PR #2434, opened 2022-01-24
WIP: web/refactor/profile translation keys

tsuruby:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/profile/view/About.tsx
+++ b/app/web/features/profile/view/About.tsx
@@ -1,15 +1,9 @@
 import { Typography } from "@material-ui/core";
 import Divider from "components/Divider";
 import Markdown from "components/Markdown";
-import {
-  ADDITIONAL,
-  HOBBIES,
-  LIVED_IN,
-  NONE,
-  OVERVIEW,
-  TRAVELED_TO,
-  WHO,
-} from "features/constants";
+import { NONE } from "features/constants";
+import { useTranslation } from "i18n";
+import { GLOBAL,PROFILE } from "i18n/namespaces";
 import { User } from "proto/api_pb";
 import makeStyles from "utils/makeStyles";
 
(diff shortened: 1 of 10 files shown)


app/web/features/profile/view/About.tsx
```diff
@@ -34,34 +28,35 @@ const useStyles = makeStyles((theme) => ({
 export default function About({ user }: AboutProps) {
   const classes = useStyles();
```
darrenvong, line 29: I'd suggest move everything in the `profile` namespace instead - it doesn't belong in global since it's specific to this feature area.

app/web/features/profile/view/Home.tsx
```diff
@@ -59,121 +36,122 @@ interface HomeProps {
 
 export default function Home({ user }: HomeProps) {
   const classes = useStyles();
+  const { t } = useTranslation([GLOBAL]);
```
darrenvong, line 40: Move all these strings/constants to profile namespace - same reason as the other file.

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -208,18 +201,18 @@ export default function NewHostRequest({
...
             fullWidth
-            placeholder={REQUEST_DESCRIPTION}
+            placeholder={t("global:request_description")}
           />
           <CardActions className={classes.send}>
-            <Button onClick={() => setIsRequesting(false)}>{CANCEL}</Button>
+            <Button onClick={() => setIsRequesting(false)}>{t("global:cancel")}</Button>
             <Button type="submit" onClick={onSubmit}>
-              {SEND}
+              {t("global:send")}
```
darrenvong, line 215: With these two, we can probably keep in the global namespace since they are quite generic text, which is a lot more reusable throughout the app than the more specific constants/strings above.

app/web/features/profile/view/ReferencesView.tsx
```diff
@@ -65,14 +67,14 @@ export default function ReferencesView({
                 loading={isFetchingNextPage}
                 onClick={() => fetchNextPage()}
               >
-                {SEE_MORE_REFERENCES}
+                {t("profile:see_more_references")}
```
darrenvong, line 70: I'd probably consider renaming this since this is named quite literal to the button text as is right now. (e.g. what if we re-word this to "Show more references"?) This is a bad example of how we should name things from our older code/more naive self... 😅

app/web/features/profile/view/UserCard.tsx
```diff
@@ -39,13 +40,20 @@ export default function UserCard({
...
+  }
   return (
     <Card className={classes.detailsCard} id={REQUEST_ID}>
       <UserTabContext tab={tab}>
         <TabBar
           setValue={onTabChange}
-          labels={SECTION_LABELS}
-          ariaLabel={SECTION_LABELS_A11Y_TEXT}
+          labels={sectionLabels}
+          ariaLabel={t("global:section_labels_a11y_text")}
```
darrenvong, line 56: Move these to `profile` namespace

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -64,13 +61,13 @@ export default function UserOverview({ children }: PropsWithChildren<unknown>) {
...
-            description={COMMUNITY_STANDING_DESCRIPTION}
+            label={t("global:community_standing")}
+            description={t("global:community_standing_description")}
           />
           <BarWithHelp
             value={user.verification || 0}
-            label={VERIFICATION_SCORE}
-            description={VERIFICATION_SCORE_DESCRIPTION}
+            label={t("global:verification_score")}
+            description={t("global:verification_score_description")}
```
darrenvong, line 70: Move to `profile` namespace

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -81,6 +73,7 @@ export default function NewHostRequest({
   const isPostBetaEnabled = process.env.NEXT_PUBLIC_IS_POST_BETA_ENABLED;
   const [numVisitors, setNumVisitors] = useState(1);
   const user = useProfileUser();
+  const { t } = useTranslation([GLOBAL]);
```
darrenvong, line 77: Even though this component is in the `profile` folder... it's debatable whether it should be here since it's more about sending host requests. Maybe these messages/constants should be in `messages` instead

darrenvong (changes requested):
I think it's unfortunate that the "global" namespace in the old constant file became a dumping ground for strings that should have been located in the "profile" feature folder/namespace instead. We probably shouldn't copy things 1:1 as there's good opportunity to really clear out that constant file.

Another thing worth noting - even after you swap over things to the `profile` namespace, there are still quite a lot of strings/constants here! I think it will also be good to group those into objects based on the component's they are in within the JSON, similar to what's done in the `auth` namespace.

Sorry if I'm jumping on this early as you might be planning to do that next!

aapeliv:
@dieterpankratz how're you doing on this? It'd probably be good to get this merged in so the components don't have time to change and cause merge issues!

PR #2466, opened 2022-01-31
error message on DonationsBox

### LINKED ISSUE #2432: Typing in a fraction in custom donation box
Subject: Typing in a fraction in custom donation box
Description:
Should give meaningful error or succeed

Results:
Says "Assertion failed" (probably from protobuf serialization since it expects an int and gets a float)

Backend version: develop-ea44d455
Frontend version: develop-ea44d455
User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:97.0) Gecko/20100101 Firefox/97.0
Page: https://app.couchers.org/donate
User: aapeli (1)

Zenpanic:
better error message when donating decimals or negative numbers

closes #2432 
when entering a negative number or a decimal number on the donation page, and submitting the form, the "Assertion error" message is replaced by "Whole numbers only, please!".

**Web frontend checklist**
- [X] Formatted my code with `yarn format && yarn lint --fix`
- [X] All tests pass

### DIFF
--- a/app/web/features/donations/DonationsBox.tsx
+++ b/app/web/features/donations/DonationsBox.tsx
@@ -219,11 +219,11 @@ export default function DonationsBox() {
   } = useForm<DonationFormData>();
 
   const customAmountInput = useRef<HTMLInputElement>(null);
-  
+
   const checkForValidAmount: boolean = (amount: number) => {
-      if (!Number.isInteger(amount)) return false;
-      if (amount < 1) return false;
-      return true;
+    if (!Number.isInteger(amount)) return false;
+    if (amount < 1) return false;
+    return true;
   };
 
   const {
@@ -232,9 +232,9 @@ export default function DonationsBox() {
     mutate: initiateDonation,
   } = useMutation<void, RpcError, DonationFormData>(
     async ({ amount, recurring }) => {
-        if  (!checkForValidAmount(amount)) {
-          throw Error(t("donations_box.amount_validation_error"));
-      };
+      if (!checkForValidAmount(amount)) {
+        throw Error(t("donations_box.amount_validation_error"));
+      }
       const stripe = (await stripePromise)!;
       const session_id = await service.donations.initiateDonation(
         amount,
--- a/app/web/features/donations/DonationsBox.tsx
+++ b/app/web/features/donations/DonationsBox.tsx
@@ -219,6 +219,12 @@ export default function DonationsBox() {
   } = useForm<DonationFormData>();
 
   const customAmountInput = useRef<HTMLInputElement>(null);
+  
+  const checkForValidAmount: boolean = (amount: number) => {
+      if (!Number.isInteger(amount)) return false;
+      if (amount < 1) return false;
+      return true;
+  };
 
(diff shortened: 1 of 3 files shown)


app/web/features/donations/DonationsBox.tsx
```diff
@@ -275,7 +275,7 @@ export default function DonationsBox() {
   return (
     <>
       <form onSubmit={onSubmit} className={classes.donationsBox}>
-        {error && <Alert severity="error">{error.message}</Alert>}
+        {error && <Alert severity="error">{error.message !== "Assertion failed" ? error.message : "Whole numbers only, please!"}</Alert>}
```
darrenvong: I don't think we should rely on the "Assertion failed" message to always mean the user has typed in a non positive integer, as this would end up showing the wrong message if the backend API gets modified/extended with more assertions in the future.

Since we can read the `amount` on form submission, it's more reliable to intercept there as a custom validation rule to check that the number is non-negative and a whole number instead with React hook form.

darrenvong: Also, can you please move this to `/app/web/features/donations/locale/en.json`, name the translation key `amount_validation_error` within the `donations_box` object and read the message here from there with the `t` function? We are trying to move away from hardcoded messages so they can be translated to multiple languages.

darrenvong (changes requested):

app/web/features/donations/DonationsBox.tsx
Zenpanic: Thanks for the feedback, @darrenvong I'm on it.

darrenvong (changes requested):
Can you run `yarn format` and `yarn lint` on the code as well to make sure the code is formatted properly and there aren't any linting issues? I can see the code formatting looks a bit misaligned at least 😅  

And thanks for the French translation too 🇫🇷

Zenpanic:
It should be alright now :)

app/web/features/donations/DonationsBox.tsx
```diff
@@ -220,12 +220,21 @@ export default function DonationsBox() {
 
   const customAmountInput = useRef<HTMLInputElement>(null);
 
+  const checkForValidAmount: boolean = (amount: number) => {
```
darrenvong:
```suggestion
  const checkForValidAmount = (amount: number) => {
```
This type error is failing the build

darrenvong (changes requested):
Sorry, hopefully the last thing!

Forgot to mention but in future, you can run `yarn typecheck` to catch this same problem too, since the [development build doesn't check for type checking issues](https://github.com/vercel/next.js/issues/14997). I will look into adding that into the checklist or our get started doc... 😅

app/web/features/donations/DonationsBox.tsx
Zenpanic: Thanks for your patience and the advice! I'll keep that in mind for the next issue. I've fixed the code and tried it, hopefuly it's all good now!

darrenvong (approved):

PR #2471, opened 2022-02-01
Implement conversation muting

aapeliv:
**Backend checklist**
- [ ] All tests pass


kalvdans (dismissed):
LGTM! Although a bit non-symmetrical to not also have a dummy timestamp in the past for the unmuted case.

aapeliv:
> LGTM! Although a bit non-symmetrical to not also have a dummy timestamp in the past for the unmuted case.

That's a good point, it'll probably simplify if done that way...

aapeliv:
By the way @kalvdans, are you able to join the backend meeting in ~2h? Details at https://couchers.org/contributors/backend-meeting.ics

aapeliv:
> LGTM! Although a bit non-symmetrical to not also have a dummy timestamp in the past for the unmuted case.

Fixed!

app/proto/conversations.proto
```diff
@@ -133,6 +138,12 @@ message Message {
   }
 }
 
+message MuteInfo {
+  bool muted = 1;
+  // muted_until is not present if muted forever
+  google.protobuf.Timestamp muted_until = 2;
```
kalvdans, line 145: Perhaps it is easier to pass a duration here. "mute for the next 8 hours". Then we don't need the error code for muting in the past, and timezones are less of a trouble.

kalvdans (dismissed):
LGTM with anther remark

app/proto/conversations.proto
aapeliv: Good point!

aapeliv: Fixed

kalvdans: Hmm, we didn't get rid of the error code anyway since Duration is signed :( But still better than before IMHO!

kalvdans (approved):

app/proto/conversations.proto
aapeliv: Yes, and it's better for the frontend

PR #2472, opened 2022-02-01
Implement new notifications for chats

aapeliv:
Implements the new notification system for conversations. The actual notifications are generated in the background in case of a huge group chat so that the system is still responsive.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -24,6 +24,7 @@ from couchers.models import (
     User,
 )
 from couchers.notifications.background import handle_email_digests, handle_email_notifications, handle_notification
+from couchers.notifications.notify import notify
 from couchers.servicers.blocking import are_blocked
 from couchers.sql import couchers_select as select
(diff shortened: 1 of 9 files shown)


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -57,6 +58,54 @@ def process_purge_login_tokens(payload):
...
+
+        notify_args = {
+            "topic": "chat",
+            "key": str(message.conversation_id),
+            "action": "message",
+            "icon": "message",
+            "title": f"{message.author.name} sent a message in {group_chat.title}.",
+            "content": message.text,
+            "link": urls.chat_link(chat_id=message.conversation_id),
+        }
```
kalvdans: Avoid the intermediary dict and pass the parameters in the call directly.

aapeliv: Fixed

kalvdans (approved):
LGTM! Thanks for fixing my complaint

aapeliv:
Merged in 528d311c69db2f11898d6d337ac6717beced8737

PR #2473, opened 2022-02-01
Add meeting 40 notes

aapeliv:


kalvdans (approved):
thanks for keeping notes!

PR #2478, opened 2022-02-02
fix: message input box getting cut off the screen

### LINKED ISSUE #2366: Reply box/button below fold on messagin pages
Subject: Missplaced reply text frame in chat in 4K display
Description:
just went to reply a chat and didn't see how to do that until I scrolled and found way down there (after a huge white space) the reply input frame

Results:
The huge white space should not be there, and the reply frame should have been right under the original text message received.

Backend version: develop-c3b46fed
Frontend version: develop-c3b46fed
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.71 Safari/537.36
Page: https://app.couchers.org/messages/chats/4789
User: antu (185)

darrenvong:
Closes #2366

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/messages/messagelist/InfiniteMessageLoader.tsx
+++ b/app/web/features/messages/messagelist/InfiniteMessageLoader.tsx
@@ -74,7 +74,7 @@ export default function InfiniteMessageLoader({
     const messageEl = document.getElementById(
       messageElementId(prevTopMessageId.current ?? 0)
     );
-    messageEl?.scrollIntoView();
+    messageEl?.scrollIntoView({ block: "center" });
   }, [isFetchingNextPage]);
 
   // Scroll to bottom on load
--- a/app/web/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/features/messages/groupchats/GroupChatView.tsx
@@ -60,13 +60,10 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
     },
   },
   pageWrapper: {
-    [theme.breakpoints.up("sm")]: {
-      height: `calc(100vh - ${theme.shape.navPaddingSmUp})`,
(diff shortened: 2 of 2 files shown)


app/web/features/messages/messagelist/InfiniteMessageLoader.tsx
```diff
@@ -74,7 +74,7 @@ export default function InfiniteMessageLoader({
     const messageEl = document.getElementById(
       messageElementId(prevTopMessageId.current ?? 0)
     );
-    messageEl?.scrollIntoView();
+    messageEl?.scrollIntoView({ block: "center" });
```
darrenvong, line 77: I think this is "good enough" of a fix. Without this tweak, when the user scroll up, there's a weird scroll jerk where the whole page gets scrolled. Scrolling into view to align to the center of the screen at least seems to have reduce the sudden jerky movement.

It's probably because this is not taking up as much height on the page now. I think a more robust fix to make the jerk even less noticeable would probably require re-working the complex scroll logic we currently have which I didn't fancy touching given the size of the other change 😅

lucaslcode:
I'm sorry I missed this!
In my last PR I just got rid of the messages header so it was the same as before. I thought keeping it ends up with not enough real estate for the conversation itself on mobile. What do you think?

darrenvong:
> In my last PR I just got rid of the messages header so it was the same as before

Has this been merged? Does that mean the header only gets shown on `/messages`?

darrenvong:
Oh yeah I can see it's merged. Yeah I think I prefer not having the headers there for the same reason you said... so maybe we can close this

PR #2509, opened 2022-02-06
Create hook to auto-scale font size to fit into the parent container

### LINKED ISSUE #1730: Long display name on profile overflows container
image

image

For long display names, can we scale down the font size to fit within the container bounds?

urbullet:
Created a hook that takes two refs (parent container & text container). And then checks if the text container is overlapping the parent, then some action is required (resizing). And it will resize the font until 6px, as it will be impossible to read after that. Instead, it informs the text container to break anywhere when there is a text overflow, while adding hyphens for more clarity of the word break. 

I have tested it with the longest known name (according to google, it is "Hubert Blaine Wolfeschlegelsteinhausenbergerdorff Sr."). And even Mr. Hubert can now see his name displayed properly on our profile page.

I have not added any tests for it, as TDD and css can get really messy; add hooks & refs into the blunder and you have a very hard test. 

Closes #1730. 

**Web frontend checklist**
- [ ] All tests pass


darrenvong (dismissed):
This seems a bit overkill to me. Can we not design this such that it flows over to multiple lines with only CSS?

I also have accessibility concerns over this solution - the fact this will keep shrink the user name to 6px just to fit a design is really bad for users with poor eyesight. Realistically, anything below 10px (i.e. the size of our current "caption text" in our theme) is just too small to be easily read, and undermines the fact the user name is meant to be an important text!

Also setting it against absolute pixel value will break things for user who set their default font size to something larger in the browser.

lucaslcode:
This is a really cool hook, but I think `word-break: break-word` (god css wrapping is a mess) might do the trick?

urbullet:
A simple word-break does the trick yes, I believe it has to be anywhere
though not break-word.

I think i misunderstood the requirements as I took it too literal maybe,
because the ticket mentioned being able to scale down font size dynamically
when it overflows.

In that case I'll do the required changes and do another review request.

lucaslcode:
just make sure to check caniuse.com - from memory the most "correct" option isn't supported on mobile ios or something (typical 🤣 )

urbullet:
I've removed the created hook, and replaced it with simple CSS properties `word-break: 'break-word'; overflow-wrap: 'break-word';`. 

This should do the trick, it is supported across all browsers.

lucaslcode (approved):
Thanks so much! Feel free to merge this

PR #2522, opened 2022-02-07
Update to pytest 7

aapeliv:
pytest-pythonpath is no longer required with the new built-in option


kalvdans (approved):
Confirmed that pythonpath is still relative to the `pytest.ini` file. https://pytest.org/en/latest/reference/reference.html#confval-pythonpath

LGTM! I'll do the same update at work :D

PR #2526, opened 2022-02-07
doc: Instructions to update the docker grpc image

papeldeorigami:
Add information on how to update the docker image that create the local protocol buffers.

### DIFF
--- a/app/readme.md
+++ b/app/readme.md
@@ -54,7 +54,7 @@ If you are on **Windows** (without WSL2), run the following command:
 docker run --rm -w /app -v %cd%:/app registry.gitlab.com/couchers/grpc sh -c "cat generate_protos.sh | dos2unix | sh"
 ```
 
-### How to update the protocol buffers
+### How to update the protocol buffers compiler
 
 If you start to get errors after some update in the grpc dependencies, you may need to pull the latest docker image with this command:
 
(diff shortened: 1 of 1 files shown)


aapeliv (dismissed):
Ah, thank you!

app/readme.md
```diff
@@ -54,6 +54,17 @@ If you are on **Windows** (without WSL2), run the following command:
 docker run --rm -w /app -v %cd%:/app registry.gitlab.com/couchers/grpc sh -c "cat generate_protos.sh | dos2unix | sh"
 ```
 
+### How to update the protocol buffers
```
darrenvong:
```suggestion
### How to update the protocol buffers compiler
```
Very minor nitpick as the above steps already update the protocol buffers, but errors because of an updated compiler.

darrenvong: In fact, I'll just commit the suggestion to save more round trips of this being blocked from merging :p

darrenvong (dismissed):

aapeliv (dismissed):
Actually, I think you can get around this by using the `--pull always` flag on the `docker run` command; like this:

```sh
docker run --pull always --rm -w /app -v $(pwd):/app registry.gitlab.com/couchers/grpc ./generate_protos.sh
```

lucaslcode (dismissed):

app/readme.md
darrenvong:
```suggestion
### How to update the protocol buffers compiler
```

darrenvong (approved):

PR #2527, opened 2022-02-07
fix: translate keys for communities/events

papeldeorigami:
Solve Couchers-org/couchers#4152 for the scope of communities/events.

**Web frontend checklist**
- [x] All tests pass


papeldeorigami:
Notice: there is one remaining constant called "details" that is left behind:

_app/web/features/communities/events/constants.ts_
```js
export const details = ({ colon = false }: { colon?: boolean } = {}) =>
  `Details${colon ? ":" : ""}`;
```
As the word Details is not translated, I'll leave this change for a future PR.

darrenvong:
> As the word Details is not translated, I'll leave this change for a future PR.

I wonder if it's worth just swapping in the literal string as translation key for places that use this function and just get rid of it here altogether. This kind of "DRYness" doesn't really work well with translation as we found from encountering similar issues in other parts of the app already.

Separately, do you want to go through checking the naming some of the keys here as well? It's not as bad as the old code where we named things literally to the English message everywhere 🙈 , but maybe there are some places worth checking

lucaslcode (approved):
I renamed a few which were the worst offenders.

In general though, I'm thinking we shouldn't go crazy with naming and structuring the translations. I'd like to speed the process up for everyone and it can get really tedious. I'm wary of perfect being the enemy of good in this case.

Of course, any new key can be done cleanly and maybe also tidying up the old ones when we are working with them. Also, any broken-up sentences or functions should definitely be changed.
What do you think @darrenvong?

lucaslcode:
I also note this PR also is showing the translation keys instead of the strings on vercel for event pages, but not locally. I wonder if it's some strangeness related to the fact the event pages are marked "fallback" and the translations aren't being loaded properly when running as a cloud function?
Hopefully this doesn't happen for us as we use the normal next.js server for next.couchershq.org and couchers.org.

darrenvong:
Yeah it's a bit of a balance, and it doesn't look like it's too bad in this part of the codebase from what you renamed.

It's mostly more effort for translators than for us I guess when we find later we're renaming a bunch of things as we fix things, but maybe it's not that bad as there seem to be more of them working on the translation side of things than us at the code at the moment.

PR #2528, opened 2022-02-07
Send reasons for account deletion to support

aapeliv:


app/backend/templates/account_deletion_report.md
```diff
@@ -0,0 +1,22 @@
+---
+subject: "Account deletion reason #{{ reason.id|couchers_escape }}"
+---
+
+{% from "macros.html" import button, link, support_email, email_link, newline %}
+
+Someone deleted their account and wrote a reason.
+
+
+* Reason{{ newline(html)|couchers_safe }}
```
kalvdans, line 10: Post-merge comment: Remove all `{{ newline(html)|couchers_safe }}`, they make the template unreadable. Reports should do fine with markdown:s html generator.

aapeliv: Sorry for bypassing review...

We're not actually using a markdown generator anymore because it caused a security bug and I thought it safer to just do it by hand.

kalvdans: Oh... then we can switch to an easier syntax ;)

aapeliv: Yes, that'd be good! I would be keen to change it; the current templates are a bit of a pain. But it works for now... We can also have a look when we redo the emails soon-ish when the notifications start being more complete.

PR #2532, opened 2022-02-08
set node version for web to >=14

### LINKED ISSUE #2531: error when running web project with old node version
When I tried to run the project for the first time with node 12 I got the following error

```
 1 of 1 unhandled error
Server Error
SyntaxError: Unexpected token '?'

This error happened while generating the page. Any console logs will be displayed in the terminal window.
Call Stack
Loader.moduleStrategy
internal/modules/esm/translators.js (140:18)
```

I solved it by running with node 16, but I think it would be nicer for the next person if we update the `package.json` with the node version required for this project.

If it sounds good I can open a PR for this, I think I just need repo permission to push my branch.

wginsberg:
closes #2531 

This change is to make it more obvious for new devs that node 14 is required for the project to run without errors. This is the output you will get now:

```
$ yarn install
yarn install v1.12.3
[1/5] 🔍  Validating package.json...
error web@0.1.0: The engine "node" is incompatible with this module. Expected version ">=14". Got "12.22.3"
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/package.json
+++ b/app/web/package.json
@@ -2,6 +2,9 @@
   "name": "web",
   "version": "0.1.0",
   "private": true,
+  "engines": {
+    "node": ">=14"
+  },
   "scripts": {
     "build-storybook": "build-storybook",
     "build": "next build",


lucaslcode (changes requested):
Good idea! I think it has to be 14 though - vercel doesn't support 16 yet

lucaslcode (approved):

PR #2538, opened 2022-02-10
chore: Move `pt` translations to `pt-BR`

darrenvong:
As per discussions in the #dev channel, as it has been agreed by the translation teams that they will translate Portuguese (Brazil) in parallel with Portuguese (Portugal).

I'm doing this in a code PR as it's less tedious to rename the JSON file for Weblate to pick this up than to manually move over the translation one by one by hand!

**Web frontend checklist**
N/A - only translation JSON reshuffles


app/web/features/communities/events/constants.ts
```diff
@@ -1 +0,0 @@
-export const getExtraAvatarCountText = (count: number) => `+${count}`;
```
darrenvong, line 1: This is technically unrelated, but it's not being used anymore since Ricardo's PR it seems, so thought I'll clear up here too since it's only another line changed

aapeliv (approved):

PR #2542, opened 2022-02-10
Include translation files in serverless contexts

### LINKED ISSUE #2541: error_fallback.title
Subject: error_fallback.title
Description:
got an error when refreshing on a user's page after closing browser, reopening browser, and refreshing.

Results:
image

Refreshing again resulted in the expected message:
image

Backend version: develop-cdf50a3d
Frontend version: develop-cdf50a3d
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:97.0) Gecko/20100101 Firefox/97.0
Page: https://couchers.org/user/delayshyr/about
User: jesse (57)

lucaslcode:
Fixes #2541

### DIFF
--- a/app/web/next-i18next.config.js
+++ b/app/web/next-i18next.config.js
@@ -14,9 +14,13 @@ module.exports = {
   returnEmptyString: false,
   serializeConfig: false,
   localePath: (locale, namespace) => {
+    const path = require("path");
     if (namespace === "global") {
-      return `resources/locales/${locale}.json`;
+      return path.resolve(process.cwd(), `resources/locales/${locale}.json`);
     }
-    return `features/${namespace}/locales/${locale}.json`;
+    return path.resolve(
+      process.cwd(),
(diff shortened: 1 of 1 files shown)


lucaslcode:
Well that was easier than I thought. Seems to be working on vercel (go to the event page), hopefully will work on our servers too.

lucaslcode:
For context, the problem was that the pages which are generated on the fly like profiles and events are created using a serverless function, and the translation resources didn't exist in the serverless function context. Adding the `path.resolve` makes it get picked up by next's [file tracing](https://nextjs.org/docs/advanced-features/output-file-tracing) (I think??)

darrenvong (approved):
Nice! How did you know it was caused by this? It's not that obvious they mention this from the file tracing docs 😅 

Also is that serverless function true even in our prod environment where we're doing everything in Docker?

aapeliv:
Merging this as prod is currently broken with the mute translations...

PR #2545, opened 2022-02-10
filter duplicate map search results

wginsberg:
part of Couchers-org/couchers#4132

**Web frontend checklist**
- [x] All tests pass

## before
Screen Shot 2022-02-10 at 11 22 55 AM

## after
Screen Shot 2022-02-10 at 11 22 48 AM

### DIFF
--- a/app/web/utils/nominatim.ts
+++ b/app/web/utils/nominatim.ts
@@ -42,3 +42,16 @@ export const simplifyPlaceDisplayName = (place: NominatimPlace) => {
   }
   return addressParts.join(", ");
 };
+
+export const filterDuplicatePlaces = (places: NominatimPlace[] = []) => {
+  const deduplicatedPlaces = places.reduce((previousRecord, currentPlace) => {
+    const importance = currentPlace.importance ?? 0;
+    const displayName = simplifyPlaceDisplayName(currentPlace);
+
+    return previousRecord[displayName]?.importance >= importance
+      ? previousRecord
+      : { ...previousRecord, [displayName]: currentPlace };
+  }, {} as Record<string, NominatimPlace>);
+
+  return Object.values(deduplicatedPlaces)
+};
--- a/app/web/utils/hooks.ts
+++ b/app/web/utils/hooks.ts
@@ -9,7 +9,11 @@ import {
   useRef,
   useState,
 } from "react";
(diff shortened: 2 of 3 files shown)


app/web/utils/nominatim.ts
```diff
@@ -42,3 +42,21 @@ export const simplifyPlaceDisplayName = (place: NominatimPlace) => {
   }
   return addressParts.join(", ");
 };
+
+export const filterDuplicatePlaces = (places: NominatimPlace[] = []) => {
+  const placeImportance = places.reduce((places, place) => {
```
lucaslcode: This naming is a tiny bit confusing:
- `places` is shadowed
- `placeImportance` sounds like a `Record<string, number>`. Maybe it should be `deduplicatedPlaces`

lucaslcode (changes requested):
I could be wrong, but [looking here](https://nominatim.openstreetmap.org/ui/search.html?q=burlington&addressdetails=1), all the dupes are because of one result for the city, and another for the administrative boundary?
Maybe we should eliminate boundaries instead?

wginsberg:
@lucaslcode 

In other cases there is no "city" result at all, just two "administrative" results. Working with addresses is fun, eh!

https://nominatim.openstreetmap.org/ui/search.html?q=toronto

app/web/utils/nominatim.ts
```diff
@@ -42,3 +42,21 @@ export const simplifyPlaceDisplayName = (place: NominatimPlace) => {
...
+
+    return previousRecord[displayName]?.importance >= importance
+      ? previousRecord
+      : { ...previousRecord, [displayName]: currentPlace };
+  }, {} as Record<string, NominatimPlace>);
+
+  const filtered = places.filter((place) => {
+    const displayName = simplifyPlaceDisplayName(place);
+    return deduplicatedPlaces[displayName].place_id === place.place_id;
+  });
```
darrenvong: Could this bit not be done as part of the `reduce` too? Seems unnecessary to loop through `places` again when it's already doing some work there.

wginsberg: I was thinking this was needed to preserve ordering but actually `Object.values(deduplicatedPlaces)` does that too. will replace it with that

lucaslcode (approved):
Thanks so much, and for the great tests!

PR #2546, opened 2022-02-10
Fix migration scripts and add index on background jobs

aapeliv:
When doing migrations before; it would always pick up the `logging.api_calls` table as having changed since it wasn't looking the the `logging` schema. I also added the indexes automatically added by postgis. This means that when you now run the migration generation script it actually generates the right-ish migration without you having to remove a bunch of cruft!

**Backend checklist**
- [ ] All tests pass


kalvdans (approved):
I haven't tried but less manual editing is better!

Issue #1718, opened 2021-07-26
Add Skip on Contributor Form on Sign Up 


dommers-ux:
We should add a **skip** option on the contributor / feedback form on the sign up flow. 

It currently looks like this -
<img width="529" alt="Screen Shot 2021-07-25 at 6 11 58 PM" src="https://user-images.githubusercontent.com/83376913/126918064-c3f644cb-4c07-48b3-9afe-9da5636545e5.png">

It is important for users to finish the sign up process as fast as possible. The more steps required --> the higher likelihood of quitting during registration. Maybe something like this - 
<img width="368" alt="Screen Shot 2021-07-25 at 6 14 38 PM" src="https://user-images.githubusercontent.com/83376913/126918163-de360561-dae2-415d-833c-488d3c24c288.png">

confused-cabbage:
Hey! I'd be willing to take this issue on!

alceil:
Can I take up this issue? @dommers-ux

lucaslcode:
Sure! @alceil

lucaslcode:
Maybe reduce some of the other text as suitable if you add that

alceil:
I have linked  a pr. Can you review it? @lucaslcode

Tay0108:
Should this issue be closed?

PR #2552, opened 2022-02-13
Added Skip on Contributor Form on Sign Up

alceil:
#1718 
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [x] All tests pass


app/web/features/auth/signup/Signup.tsx
```diff
@@ -126,6 +126,9 @@ function CurrentForm() {
         <Typography variant="h1" gutterBottom>
           {t("auth:feedback_form.header")}
         </Typography>
+        <Typography variant="h1" gutterBottom>
+        {t("auth:feedback_form.skiptitle")}
```
darrenvong: This doesn't actually link to anything to let you skip the step at the moment. You'll need to add a `<Link>` element that more or less clone the behaviour of clicking the submit button, and do this with [the `Trans` component](https://react.i18next.com/latest/trans-component) so the link is interpolated correctly from the JSON resource.

app/web/features/auth/locales/en.json
```diff
@@ -105,7 +105,8 @@
     }
   },
   "feedback_form": {
-    "header": "Have your say"
+    "header": "Have your say",
+    "skiptitle":"Not interested? Click <a>here</a> to skip this step."
```
darrenvong:
```suggestion
    "skip_title":"Not interested? Click <1>here</1> to skip this step."
```
Please ~~space~~ separate words with the underscores like in the surrounding examples.

Also, you might have to do something like this rather than putting an `<a>` here. The `<1>`s allow the bits to be replaced with the correct element in the code.

darrenvong (changes requested):

app/web/features/auth/locales/en.json
alceil: got it

app/web/features/auth/signup/Signup.tsx
alceil: I just went through the codebase of what submit button does and it does state changing and it doesn't not do route changing!.So what should I do? Shall I route it to login page instead? @darrenvong

alceil:
I have done the suggested changes @darrenvong but I have doubt regarding the routing when we click here. The functionality of submit button is it takes data from the feedback form and then it changes state. State change happens instead of routing to another page. So how should I proceed I need some pointers

darrenvong:
Yeah you've stumbled across some rather clunky parts of our code... sorry about that! Perhaps this isn't such a good first issue after all...

To replicate the submit button behaviour, you'd have to move the skip text JSX into the same level as the `FeedbackForm` component so you have access to the submission/state mutation logic like you have noted.

But the following is the clunky part: the submit logic expects a "form" object, so you'll have to do something similar to what's in `ContributorForm`, except you'd submit an empty form object (literally do a `form.toObject()` created by `new ContributorFormPb()`) since we don't actually care about populating the form here. In fact, if you do that now by clicking the submit button without filling in the form, it's essentially doing the skipping behaviour except in a non-obvious way.
 
I hope all of that kinda makes sense when you look at what's been done in `ContributorForm` - it's a bit hard to explain this one briefly! But even with my suggested solution (or more like a workaround), It's a bit hacky but I don't think there's a better way forward unless we re-work the flow token logic slightly, which seems way out of scope of what you're trying to do here. Unless you can think of something better @lucaslcode?

app/web/features/auth/signup/Signup.tsx
```diff
@@ -126,6 +126,14 @@ function CurrentForm() {
...
           {t("auth:feedback_form.header")}
         </Typography>
+        <Trans i18nKey="auth:feedback_form.skiptitle">
+        Not interested? Click 
+            <StyledLink href={loginRoute}>
+            here
+            </StyledLink>
+            to skip this step.
+          </Trans>
+        </Typography>
```
darrenvong: I think you've missed an opening `<Typography>` here - the build choked on some syntax error.

Also, I think it's better to wrap the "skip this step" as the link and remove "click here" in the text as that's quite poor for accessibility to have as the link label for low vision users. Not really your fault as I know you were following the designs, but probably the better thing to do when you get around changing things

lucaslcode:
Hi @alceil, do you plan to continue working on this?

alceil:
Ya but can you give me pointers on how to approach it since it ain't beginner friendly level and has some complications. I dont wanna mess up the code base @lucaslcode

lucaslcode:
Hmm maybe you can put the new text in `ContributorForm.tsx` instead? That way you can just have a Button with onClick = submit and a text variant. (So basically nearly the same as the real submit button, which allows skipping anyway)

alceil:
Ya thats a good work around I can try that

alceil:
I have added the workaround as you said. Could you review it? @lucaslcode

PR #2561, opened 2022-02-13
Add languages to preview

lucaslcode:


darrenvong (approved):
Nice, is it to ease testing for translators too?

PR #2579, opened 2022-02-14
Make hosting status non-nullable

### LINKED ISSUE #2529: Hosting status is nullable
We should make it non-null and make it `cant_host` for those for whom it's now null

aapeliv:
Closes #2529


kalvdans (dismissed):
LGTM! Great to restrict the entropy in the database.

PR #2581, opened 2022-02-14
Use column_property for faster tz lookup

aapeliv:
The underlying SQL turns from

```sql
SELECT timezone_areas.id AS timezone_areas_id, timezone_areas.tzid AS timezone_areas_tzid, ST_AsEWKB(timezone_areas.geom) AS timezone_areas_geom 
FROM timezone_areas 
WHERE ST_Contains(timezone_areas.geom, ST_GeomFromEWKT('0101000020E61000007C5B6C9B927D52C06066C2B869674440'));
```

(Note it returns the full geometry).

To this:

```sql
SELECT (SELECT timezone_areas.tzid
FROM timezone_areas 
WHERE ST_Contains(timezone_areas.geom, users.geom) 
 LIMIT 1) AS anon_1
FROM users
WHERE users.id = 1
```

~~Better yet would be to split the tz areas at say, squares of 10x10 degrees so the R-tree becomes much simpler and faster to use.~~

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/models.py
+++ b/app/backend/src/couchers/models.py
@@ -23,7 +23,9 @@ from sqlalchemy.dialects.postgresql import TSTZRANGE, ExcludeConstraint
 from sqlalchemy.ext.associationproxy import association_proxy
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import backref, column_property, declarative_base, relationship
-from sqlalchemy.sql import func, text
+from sqlalchemy.sql import func
+from sqlalchemy.sql import select as sa_select
+from sqlalchemy.sql import text
 
 from couchers.config import config
 from couchers.constants import (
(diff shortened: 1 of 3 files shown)


aapeliv:
For context: these lookups were in general ~5-10 ms, under load they can go up to 60 ms. This brings them down to the sub-100 microsecond range, around a 100-1000x speedup.

app/backend/src/couchers/models.py
```diff
@@ -90,6 +92,16 @@ class TimezoneArea(Base):
     tzid = Column(String)
     geom = Column(Geometry(geometry_type="MULTIPOLYGON", srid=4326), nullable=False)
 
+    __table_args__ = (
+        # Verified phone numbers should be unique
```
kalvdans: update comment

kalvdans (dismissed):
Postgres wizardry approved. How did you perform your benchmark?

app/backend/src/couchers/models.py
aapeliv: Oh right, thanks

aapeliv:
> Postgres wizardry approved. How did you perform your benchmark?

It's not very rigorous: just looking at `EXPLAIN ANALYZE` output, some of the timings, but mostly the query plans. I've got the prod postgres logging all queries > 10 ms so this has been popping up a fair bit.

PR #2594, opened 2022-02-20
fix: change node requirement to say 14.x

darrenvong:
So it matches what's on Vercel for them to apply the correct Node version automatically.

### DIFF
--- a/app/web/dev.Dockerfile
+++ b/app/web/dev.Dockerfile
@@ -1,7 +1,7 @@
 # https://nextjs.org/docs/deployment#docker-image
 
 # Install dependencies only when needed
-FROM node:alpine
+FROM node:14.19.0-alpine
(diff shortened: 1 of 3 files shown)


darrenvong:
More info: https://vercel.com/docs/runtimes#official-runtimes/node-js/node-js-version

In particular,

> Defining the node property inside `engines` of a `package.json` file will override the selection made in the Project Settings and print a Build Step warning if the version does not match.

lucaslcode (dismissed):

darrenvong:
Screwed up the Docker build stuff with that change because the image wasn't pinned to a specific Node version 😅  Updated the Dockerfile to specify the latest version of Node 14.

The error: `error web@0.1.0: The engine "node" is incompatible with this module. Expected version "14.x". Got "17.5.0"`

darrenvong:
@lucaslcode Do you mind re-approving this ^^

lucaslcode (approved):

PR #2610, opened 2022-02-21
Add Dummy Superuser API Key

### LINKED ISSUE #1731: Create an API key for testing in local dev

grellyd:
This PR introduces a dummy data API key for the dummy default Aapeli Admin user. The API key will be printed in the `backend` container logs as an email. On every restart of the server, an API key is checked for, and if one is found, it will be printed into the logs. 

In order to use the API Key, the dummy default Aapeli admin user was elevated to SuperUser. 

No tests were added as this is just operating on the dummy data. 

The PR can be tested via:
 * Start up the backend
 * Find the API key email (`You recently requested an API key for Couchers.org. We've issued you with the following API key:`)
 * Leveraging [grpcurl](https://github.com/fullstorydev/grpcurl), give it a try:
 ```bash
 grpcurl -H "Authorization: Bearer $BEARER_TOKEN" -plaintext -import-path /home/grellyd/projects/couchers/app/proto -proto /home/grellyd/projects/couchers/app/proto/api.proto -d '{"user": "1"}' localhost:1751 org.couchers.api.core.API/GetUser 
 ```
 * You should get a user object as a result, not an authentication error
 
 
 Resolves #1731 

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/admin.py
+++ b/app/backend/src/couchers/servicers/admin.py
@@ -99,21 +99,9 @@ class Admin(admin_pb2_grpc.AdminServicer):
             user = session.execute(select(User).where_username_or_email_or_id(request.user)).scalar_one_or_none()
             if not user:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
-            token, expiry = create_session(
-                context, session, user, long_lived=True, is_api_key=True, duration=timedelta(days=365)
-            )
-            send_api_key_email(session, user, token, expiry)
 
-            notify(
-                user_id=user.id,
-                topic="api_key",
-                key="",
-                action="create",
-                icon="wrench",
-                title=f"An admin created an API key for you, please check your email",
-                link=urls.account_settings_link(),
-            )
-
+            create_api_key()
+        
             return _user_to_details(user)
 
     def CreateCommunity(self, request, context):
@@ -130,3 +118,20 @@ class Admin(admin_pb2_grpc.AdminServicer):
             )
 
             return community_to_pb(node, context)
+
+def create_api_key(context, session, user):
+    token, expiry = create_session(
+        context, session, user, long_lived=True, is_api_key=True, duration=timedelta(days=365)
+    )
+    send_api_key_email(session, user, token, expiry)
+
+    notify(
+        user_id=user.id,
+        topic="api_key",
+        key="",
+        action="create",
+        icon="wrench",
+        title=f"An admin created an API key for you, please check your email",
+        link=urls.account_settings_link(),
+    )
+
--- a/app/backend/src/couchers/servicers/auth.py
+++ b/app/backend/src/couchers/servicers/auth.py
@@ -98,23 +98,6 @@ def create_session(context, session, user, long_lived, is_api_key=False, duratio
     return token, user_session.expiry
 
 
-def create_api_key(context, session, user):
-    token, expiry = create_session(
-        context, session, user, long_lived=True, is_api_key=True, duration=timedelta(days=365)
-    )
-    send_api_key_email(session, user, token, expiry)
-
-    notify(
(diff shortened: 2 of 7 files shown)


app/backend/src/dummy_data.py
```diff
@@ -365,6 +369,41 @@ def add_dummy_communities():
...
+                .where(UserSession.is_api_key)
+                .where(UserSession.is_valid)
+                .where(UserSession.user_id == 1)
+            ).scalar_one()
+            > 0
+        ):
+            logger.info("API Keys not empty, not adding dummy API key.")
+            token = session.execute(
+                select(UserSession.token)
+                .select_from(UserSession)
```
aapeliv, line 394: This is redundant: SQLAlchemy figures out that it's from `UserSession` since it's the only table being queried!

app/backend/src/couchers/servicers/auth.py
```diff
@@ -97,6 +98,23 @@ def create_session(context, session, user, long_lived, is_api_key=False, duratio
...
+        user_id=user.id,
+        topic="api_key",
+        key="",
+        action="create",
+        icon="wrench",
+        title=f"An admin created an API key for you, please check your email",
+        link=urls.account_settings_link(),
+    )
+
+
```
kalvdans: Please move this helper function to `admin.py` and make the `CreateApiKey` method call this function. This is to avoid duplicated code that does approximately the same thing.

kalvdans (changes requested):
Looks great, I just have one comment.

(I have not test-run it)

app/backend/src/couchers/servicers/auth.py
grellyd: Did I not already do that? Oops 😬 

Will do!

 - [x] Move `create_api_key` into `admin.py`
 - [x] Use `create_api_key` in `CreateApiKey`

grellyd:
Getting a single test failure, where in `test_dummy_data` the email is [failing to send](https://gitlab.com/couchers/couchers/-/jobs/2119194625#L130).

What advice do you folks have here?

aapeliv:
> Getting a single test failure, where in `test_dummy_data` the email is [failing to send](https://gitlab.com/couchers/couchers/-/jobs/2119194625#L130).
> 
> What advice do you folks have here?

It seems that file isn't injecting the test config. I think you can fix it by adding the following snippet at the top (with appropriate imports):

```py3
@pytest.fixture(autouse=True)
def _(testconfig):
    pass
```

(It's in most of the other `test_*` files)

kalvdans:
> Getting a single test failure, where in `test_dummy_data` the email is [failing to send](https://gitlab.com/couchers/couchers/-/jobs/2119194625#L130).
> 
> What advice do you folks have here?

Avoid the "send email" step when creating dummy data, by calling `create_session(..., is_api_key=True)` instead of `create_api_key(...)`.

grellyd:
@aapeliv I'm afraid I am going to need a bit of help here.

I tried your change to no effect. I also confirmed that the call we are making in the API creation func uses the `is_api_key=true` setting as Christan noted. 

Do you have any insights into the underlying issue?

app/backend/src/couchers/servicers/admin.py
```diff
@@ -99,21 +99,9 @@ def CreateApiKey(self, request, context):
...
-                user_id=user.id,
-                topic="api_key",
-                key="",
-                action="create",
-                icon="wrench",
-                title=f"An admin created an API key for you, please check your email",
-                link=urls.account_settings_link(),
-            )
 
+            create_api_key()
```
aapeliv, line 103: This call needs some parameters, I think that's the main issue

aapeliv (commented):
Sorry for the delay, been out in the field

aapeliv:
@grellyd: how're you doing on this? Need anything to keep moving?

PR #2612, opened 2022-02-22
Refactor: Removing weird translation keys that shouldn't appear in Weblate for donations

darrenvong:
As in title, after feedback from the translation team that there were certain keys that's causing confusion. Removing them as they don't really need to be translated...

**Web frontend checklist**
- [x] All tests pass


app/web/features/donations/locales/pt_BR.json
```diff
@@ -1,5 +1,5 @@
 {
-  "donations_value": "{{val, currency}}",
+  "": "{{val, currency}}",
```
lucaslcode:
```suggestion
```

app/web/features/donations/locales/ja.json
```diff
@@ -22,7 +22,5 @@
     "donations_text": "",
     "donations_text2": "",
     "benefactor_contact": "",
-    "benefactor_email": "benefactor@couchers.org",
-    "donations_recurrence_explainer": "",
-    "donations_value": "{{val, currency}}"
+    "donations_recurrence_explainer": ""
```
lucaslcode: Incidently, these should probably not be here, right? Otherwise it won't fall back

lucaslcode (changes requested):

app/web/features/donations/locales/ja.json
lucaslcode: (I mean the blank strings)

darrenvong: It would fallback fine since we have `returnEmptyString` configured to false in `next-i18next` for situation like these, as Weblate seems to randomly output untranslated messages as empty strings sometimes...

app/web/features/donations/locales/pt_BR.json
darrenvong: Oops

lucaslcode (approved):

PR #2630, opened 2022-02-26
Bump pool size to 50 for db connections

aapeliv:
We are starting to get enough simultaneous connections that the db connection pool gets exhausted; this bumps it up a bit.


kalvdans (commented):
What is our grpc connection thread limit? We should make sure that each thread gets a database connection.

lucaslcode:
Can we merge this? I suspect a number of issues are related to this

aapeliv:
> What is our grpc connection thread limit? We should make sure that each thread gets a database connection.

Done!

PR #2634, opened 2022-02-27
web: fix isRegion on location search (fix #2631)

### LINKED ISSUE #2631: Please choose a more specific place
https://github.com/Couchers-org/couchers/blob/f4b91b03df20e8d1cfded48ecf4f36dcc81d156a/app/web/features/search/FilterDialog.tsx#L163

It looks like regions are disabled when searching on the map. So assuming you want to go to Paris you'll get the error `Please choose a more specific place`. So I'm not sure what I'm supposed to type.
image

I've actually found a workaround:
1. Use the left search box instead of the filter dialog, the search by region works but the results are only displayed in the search result list but not on the map
2. Search for Paris, the results will appear in the list
3. Click on filter, the location is automatically input and for some reason the its `isRegion` field is `undefined` so it passes the checks
4. Add your filters such as the hosting status
5. Apply and you have the map results located to Paris

verdie-g:
**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/utils/hooks.ts
+++ b/app/web/utils/hooks.ts
@@ -15,6 +15,20 @@ import {
   simplifyPlaceDisplayName,
 } from "utils/nominatim";
 
+// Locations having one of these keys are considered non-regions.
+// https://nominatim.org/release-docs/latest/api/Output/#addressdetails
+const nonRegionKeys = [
+  "municipality",
+  "city",
+  "town",
+  "village",
+  "city_district",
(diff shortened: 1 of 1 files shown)


lucaslcode (commented):
What happens with this if the user types something specific enough but that isn't a city? Like "Euston station"

lucaslcode (approved):
Thanks so much for your contribution!

PR #2669, opened 2022-03-07
web: persist chat message drafts in localstorage

### LINKED ISSUE #2342: Saving Message draft
Subject: Saving Message draft
Description:
Message gone after navigating away from then back to Messages page.

Results:
Started a message and before sending, clicked on the recipients profile picture to navigate to their profile, then went back to messages and draft was gone. Would like message draft to be saved even when visiting different tabs on the website.

Backend version: develop-7291678f
Frontend version: develop-7291678f
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36
Page: https://app.couchers.org/messages/chats/1792
User: aedoleza (3033)

wginsberg:
closes #2342

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/features/auth/useAuthStore.test.ts
+++ b/app/web/features/auth/useAuthStore.test.ts
@@ -40,21 +40,21 @@ describe("usePersistedState hook", () => {
   it("saves then loads a value from sessionStorage", () => {
     const value = { test: "session test" };
     const { result } = renderHook(() =>
-      usePersistedState("key", { test: "" }, sessionStorage)
+      usePersistedState("key", { test: "" }, 'sessionStorage')
     );
     expect(result.current[0]).toStrictEqual({ test: "" });
     act(() => result.current[1](value));
     expect(result.current[0]).toStrictEqual(value);
     expect(sessionStorage.getItem("key")).toBe(JSON.stringify(value));
     const { result: result2 } = renderHook(() =>
-      usePersistedState("key", { test: "" }, sessionStorage)
+      usePersistedState("key", { test: "" }, 'sessionStorage')
     );
     expect(result2.current[0]).toStrictEqual(value);
   });
 
   it("clears a value", () => {
     const { result } = renderHook(() =>
-      usePersistedState("key", { test: "" }, sessionStorage)
+      usePersistedState("key", { test: "" }, 'sessionStorage')
     );
     expect(result.current[0]).toStrictEqual({ test: "" });
     act(() => result.current[2]());
--- a/app/web/features/auth/useAuthStore.test.ts
+++ b/app/web/features/auth/useAuthStore.test.ts
@@ -8,7 +8,6 @@ import wrapper from "../../test/hookWrapper";
 import { addDefaultUser } from "../../test/utils";
 import useAuthStore, {
   usePersistedState,
-  useSessionStorage,
 } from "./useAuthStore";
 
 const getUserMock = service.user.getUser as jest.Mock;
@@ -37,30 +36,26 @@ describe("usePersistedState hook", () => {
     );
     expect(result2.current[0]).toStrictEqual(value);
   });
-});
 
-describe("useSessionStorage hook", () => {
-  it("uses a default value", () => {
-    const defaultValue = "Test string";
-    const { result } = renderHook(() => useSessionStorage("key", defaultValue));
-    expect(result.current[0]).toBe(defaultValue);
-  });
-
-  it("saves then loads a value", () => {
-    const value = { test: "Test string" };
-    const { result } = renderHook(() => useSessionStorage("key", { test: "" }));
+  it("saves then loads a value from sessionStorage", () => {
+    const value = { test: "session test" };
+    const { result } = renderHook(() =>
+      usePersistedState("key", { test: "" }, sessionStorage)
+    );
     expect(result.current[0]).toStrictEqual({ test: "" });
     act(() => result.current[1](value));
     expect(result.current[0]).toStrictEqual(value);
     expect(sessionStorage.getItem("key")).toBe(JSON.stringify(value));
     const { result: result2 } = renderHook(() =>
-      useSessionStorage("key", { test: "" })
+      usePersistedState("key", { test: "" }, sessionStorage)
     );
     expect(result2.current[0]).toStrictEqual(value);
   });
 
   it("clears a value", () => {
-    const { result } = renderHook(() => useSessionStorage("key", { test: "" }));
+    const { result } = renderHook(() =>
+      usePersistedState("key", { test: "" }, sessionStorage)
+    );
     expect(result.current[0]).toStrictEqual({ test: "" });
     act(() => result.current[2]());
     expect(result.current[0]).toStrictEqual(undefined);
--- a/app/web/features/auth/useAuthStore.test.ts
+++ b/app/web/features/auth/useAuthStore.test.ts
@@ -6,7 +6,10 @@ import { service } from "service";
 
 import wrapper from "../../test/hookWrapper";
 import { addDefaultUser } from "../../test/utils";
-import useAuthStore, { usePersistedState } from "./useAuthStore";
+import useAuthStore, {
+  usePersistedState,
+  useSessionStorage,
+} from "./useAuthStore";
 
 const getUserMock = service.user.getUser as jest.Mock;
 const getCurrentUserMock = service.user.getCurrentUser as jest.Mock;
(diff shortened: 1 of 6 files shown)


app/web/features/messages/groupchats/GroupChatSendField.tsx
```diff
@@ -14,18 +15,26 @@ interface MessageFormData {
...
 
   const { mutate: handleSend, isLoading } = sendMutation;
 
   const { register, handleSubmit, reset } = useForm<MessageFormData>();
+  const [persistedMessage, setPersistedMessage, clearPersistedMessage] =
+    usePersistedState(`messages.${currentUserId}.${chatId}`, "");
+
   const onSubmit = handleSubmit(async (data: MessageFormData) => {
     handleSend(data.text.trimRight());
+    clearPersistedMessage();
```
darrenvong, line 41: I think it might be worth doing this when the user logs out as well, given our current privacy model. @lucaslcode thoughts?

darrenvong (changes requested):
It'd be good add a test for this to check reading a persisted message works, if exists in the storage

app/web/features/messages/groupchats/GroupChatSendField.tsx
wginsberg: good point.
Also I thinking now this should be in `sessionStorage` instead of `localStorage` so that it'll be cleared when the user closes the page

app/web/features/auth/useAuthStore.ts
```diff
@@ -29,10 +29,34 @@ export function usePersistedState<T>(
     },
     [key]
   );
+  return [_state, setState];
+}
+
+export function useSessionStorage<T>(
```
darrenvong: Considering the implementation is more or less the same as `usePersistedState`... I feel like it might be worth adding a third optional argument for the hook instead so that
- we can swap between `window.localStorage`/`window.sessionStorage` as needed via this argument, defaulting to `localStorage` (because really, this seems to be the first use case where we needed to switch to `sessionStorage`)
- minimise duplication of the same code and,
- we don't break existing places that use the `usePersistedState` hook

app/web/features/auth/useAuthStore.test.ts
```diff
@@ -36,6 +39,35 @@ describe("usePersistedState hook", () => {
...
+  });
+
+  it("clears a value", () => {
+    const { result } = renderHook(() => useSessionStorage("key", { test: "" }));
+    expect(result.current[0]).toStrictEqual({ test: "" });
+    act(() => result.current[2]());
+    expect(result.current[0]).toStrictEqual(undefined);
+    expect(sessionStorage.getItem("key")).toBe(null);
+  });
+});
```
darrenvong: I know you've wrote these based on the existing tests... but I think it'd be clearer for readability if we destructure `result.current` further and call the named destructured function - i.e.

```typescript
// instead of this (what the heck is `result.current[1]`? bit hard to tell without having to
// jump back into the implementation...
act(() => result.current[1](value));

// do...
const [state, setState, clearState] = result.current;
act(() => setState(value))
```
etc

(The classic of reading back old code when we were "more naive" a year ago and be like "why did we allow this crap" 🙈 )

darrenvong (changes requested):
And what about a UI test to check the text field is populated with the expected default value when something is available in the session storage?

But thanks a lot for asking the tests to check the log out cleared behaviour :D

lucaslcode (commented):
This is great stuff!! Just as a side note, next time you can clone the repo and you'll have the automatic CI and testing etc. Additionally, this is most needed in profile editing - but that can be another PR.

app/web/features/auth/useAuthStore.ts
wginsberg: Yeah I hear you. Personally I tend to ignore duplicating code until it's repeated `>=3` times.

Would it be cool if I just add a `TODO` comment for this? I'm not sure how to handle the ssr part of this hook with a dynamic storage object.

app/web/features/auth/useAuthStore.test.ts
wginsberg: I think it actually makes some sense to do it this way since destructuring `current` would make the `state` value stale 🤔

app/web/features/auth/useAuthStore.ts
darrenvong: > how to handle the ssr part of this hook with a dynamic storage object.

Wouldn't it be the same as what's being checked in `saved` - i.e. it becomes something like:

```typescript
// storage is either `window.sessionStorage` or `window.localStorage`
const saved =
    typeof window !== "undefined" ? storage.getItem(key) : null;
```

I'd generally agree with your `>=3` heuristic too to avoid premature abstraction, but I think in this case, since the `localStorage` and `sessionStorage` API is exactly the same (they are both of type `Storage`), I see little risk of having to undo this even if we merge them together.

app/web/features/auth/useAuthStore.test.ts
darrenvong: Very good point, and having to destructure each time might be too verbose. Hmm yeah maybe let's leave it like this then 👍

app/web/features/auth/useAuthStore.ts
wginsberg: It'd be nice if we could just add `storage = window.localStorage` as an argument to this function but it's going to throw an error during ssr.

Were you thinking to do it with a flag? Something like this?

```typescript
export function usePersistedState<T>(
  key: string,
  defaultValue: T,
  isLocalStorage = true,
): [T | undefined, (value: T) => void, () => void] {
  //in ssr, window doesn't exist, just use default
  const saved = typeof window !== "undefined"
    ? isLocalStorage
      ? window.localStorage.getItem(key)
      : window.sessionStorage.getItem(key)
    : null

  const [_state, _setState] = useState<T | undefined>(
    saved !== null ? JSON.parse(saved) : defaultValue
  );
  const setState = useCallback(
    (value: T) => {
      if (value === undefined) {
        console.warn(`${key} can't be stored as undefined, casting to null.`);
      }
      const v = value === undefined ? null : value;
      const storage = isLocalStorage ? window.localStorage : window.sessionStorage
      storage.setItem(key, JSON.stringify(v));
      _setState(value);
    },
    [key, isLocalStorage]
  );
  const clearState = useCallback(() => {
    const storage = isLocalStorage ? window.localStorage : window.sessionStorage
    storage.removeItem(key);
    _setState(undefined);
  }, [key, isLocalStorage]);
  return [_state, setState, clearState];
}

export function useSessionStorage<T>(
  key: string,
  defaultValue: T,
): [T | undefined, (value: T) => void, () => void] {
  return usePersistedState(key, defaultValue, false)
}
```

darrenvong: Or what about doing `storage = window?.localStorage` as the last argument? Even JS alone seems to allow optional chaining on default value assignment, so that should avoid the dreaded "can't access property `localStorage` of undefined" error in the server environment.

wginsberg: Huh, I guess I don't really understand serverside rendering with react ... I thought this wouldn't work but it does!

wginsberg: actually I think I spoke too soon ... I tried to add this change in https://github.com/Couchers-org/couchers/pull/2669/commits/0c4e8922a3f239973f3bd1605b6b29042a42eeb8 and now there is this error when running the app
```
error - features/auth/useAuthStore.ts (14:12) @ usePersistedState
ReferenceError: window is not defined
  12 |   key: string,
  13 |   defaultValue: T,
> 14 |   storage = window?.localStorage
     |            ^
  15 | ):
  ```

lucaslcode: `window?...` never works for some reason - that's why `typeof window === "undefined" ? ...` is sprinkled everywhere when the optional chaining would be simpler

wginsberg: I think it doesn't work because `?.` doesn't let you test something that hasn't been declared

I made another attempt at doing this refactor in https://github.com/Couchers-org/couchers/pull/2669/commits/2ea0a364698272294cf552fdbb531df8315bc804. I stole some "inspiration" from here https://dev.to/devlargs/nextjs-hook-for-accessing-local-or-session-storage-variables-3n0

Also I see there's a way to bypass ssr with next, but not sure if that's ideal or not https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr

darrenvong (approved):
This is really great work, thanks for your patience and doing the extra research to get around the troubles of having to make this work in SSR environment!

app/web/features/auth/useAuthStore.ts
darrenvong: I wonder if we can do `globalThis?.window` based on this MDN page about [`globalThis` being the standard way to access the global context/this object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis)

But what you've done is fine based off that blog post, so that was just me mentioning that out of curiosity!

I think the no SSR solution is probably best avoided in case we do authentication properly on the server side in future for SEO reasons, since we are technically still deferring 99% of the work client-side having only just ported over to Next.js

PR #2671, opened 2022-03-08
fix: message input not visible in mobile screens

darrenvong:
On mobile screens, the chat input is not visible until you scroll so this PR aims to fix it.

The fix itself is a bit hacky, since the height of the address bar could be subject to change between OS versions. But honestly, I can't think of a more simpler way without either:
- Slapping the 100% height back everywhere again (since the use of `100vh` is the thing that caused issues - on mobile, `vh` doesn't seem to take into account the address bar as taking up space...), so that we can use `100%` in place of `100vh`
- Read the real view height in JS and set the height of the container accordingly (seems a bit overkill to me)

| Before     | After |
| ----------- | ----------- |
| Screenshot_20220308-025307      | Screenshot_20220308-030628       |

**Web frontend checklist**
- [x] All tests pass
- [x] Clicked around my changes running locally and it works (although admittedly, this is one where the simulated tool fails since it doesn't have the address bar covering up some pixels)

### DIFF
--- a/app/web/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/features/messages/groupchats/GroupChatView.tsx
@@ -69,7 +69,8 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
     alignItems: "stretch",
     display: "flex",
     flexDirection: "column",
-    height: `calc(100vh - ${theme.shape.navPaddingXs})`,
+    // 56px = address bar height on mobile - https://dev.to/peiche/100vh-behavior-on-chrome-2hm8
+    height: `calc(100vh - ${theme.shape.navPaddingXs} - 56px)`,
   },
   title: {
     flexGrow: 1,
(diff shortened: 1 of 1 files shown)


lucaslcode (changes requested):
This needs browser sniffing as per that article, or there's wasted space on other mobile browsers

darrenvong:
> This needs browser sniffing as per that article, or there's wasted space on other mobile browsers

I thought that too initially, but after testing on mobile Firefox too:
Screenshot_20220312-224752

It seems like the problem is not only on Chromium browsers, so I thought it's safe to apply it across the board for narrow screens. Might need some testing on iOS Safari (I don't have an iPhone) but I think the thing mentioned in the article will include them too since they are also WebKit based.

darrenvong:
And it only gets applied up to 600px screen width, so I think it's fairly safe that tablets will be large enough such that they won't see the problem as I don't know many tablets with narrower screens than that

lucaslcode (approved):

PR #2672, opened 2022-03-09
Messages translation keys

darrenvong:
This is probably the most heavily hard-coded and thus painful one yet... not to mention some pretty hairy/gross time interval logic I've yet to deal with...

**Web frontend checklist**
- [x] All tests pass


app/web/features/messages/messagelist/ControlMessageView.tsx
```diff
@@ -41,8 +42,7 @@ export default function ControlMessageView({
   );
   const { ref } = useOnVisibleEffect(onVisible);
 
-  const isCurrentUser = author?.userId === currentUserId;
-  const authorName = isCurrentUser ? "you" : firstName(author?.name);
```
darrenvong, line 45: I decided to get rid of this "you" thing to simplify the i18n logic so even though I created the chat, it will say "Darren created the chat" rather than "You created the chat". Otherwise I'd have to duplicate all types of control messages just for the current user as a separate variant, since we can't reliably assume we can just swap out name for "you" as in English. Not worth the pain and complexity to keep this tiny thing imo.

app/web/features/messages/locales/en.json
```diff
@@ -1,35 +1,108 @@
...
+  "load_more": "Load more",
+  "mark_all_read": "Mark all as read",
+  "messages": "Messages",
+  "no_messages": "No messages",
+  "request_closed_message": "This host request is closed or in the past. To continue chatting, send the other person a normal message.",
+  "title": "Title",
+  "write_reference": "Write a Reference",
+  "close_request_dialog_title": "Are you done messaging?",
+  "close_request_dialog_host": "Please make sure you are done chatting before you    reject their request.",
+  "close_request_dialog_surfer": "Please make sure you are done chatting before you    cancel your request.",
```
lucaslcode:
```suggestion
  "close_request_dialog_host": "Please make sure you are done chatting before you reject their request.",
  "close_request_dialog_surfer": "Please make sure you are done chatting before you cancel your request.",
```

app/web/features/messages/utils.ts
```diff
@@ -17,6 +18,53 @@ export function messageTargetId(message: Message.AsObject) {
...
+      user,
+      status: "accepted",
+    });
+  } else {
+    throw Error(t("control_message.unknown_message_text"));
+  }
+}
+
+/**
+ * @deprecated Use controlMessage instead
```
lucaslcode: Is it still used?

lucaslcode (changes requested):

app/web/features/messages/utils.ts
darrenvong: It's in host requests for now, but might not need to do this if I finish converting the keys there in this PR later on as I can just delete the old function

app/web/utils/timeAgo.ts
```diff
@@ -19,6 +21,9 @@ export interface FuzzySpec {
   text: string;
 }
 
+/**
+ * @deprecated Use `timeAgoI18n` instead.
```
darrenvong, line 25: Had to do this since `profile` feature area still uses this function and I don't fancy dealing with this for now as part of this...

darrenvong:
@lucaslcode This was painful but I think finally it's ready for another look now 😵

lucaslcode (approved):
Amazing work

darrenvong:
Dropping Couchers-org/couchers#4152 here so it links back to the main issue

PR #2721, opened 2022-03-23
add profile save confirmation

lucaslcode:
Notes:
- This causes a "thrown error" dialog to appear in dev, but this is necessary to prevent navigation (and doesn't appear in prod)
- Tried to test this but couldn't get it working

**Web frontend checklist**
- [ ] All tests pass


darrenvong (commented):
Might it be worth making this into a custom hook? This seems like it could be useful in a lot of places in our app (basically most of community features like comments, updating local page etc?)

lucaslcode:
Good idea

darrenvong (approved):
Looks like there's some formatting errors and type errors in the tests unrelated to these changes that got slipped into develop... or is it just this branch is outdated with it?

PR #2726, opened 2022-03-26
Response rate API call

aapeliv:
Adds a response rate API call

**Backend checklist**
- [x] All tests pass


app/proto/requests.proto
```diff
@@ -134,3 +138,24 @@ message MarkLastSeenHostRequestReq {
...
+  RESPONSE_RATE_MOST = 4;
+  RESPONSE_RATE_ALMOST_ALL = 5;
+}
+
+message GetResponseRateRes {
+  ResponseRate response_rate = 1;
+  // only set for rate >= some
+  google.protobuf.Duration response_time_p33 = 2;
+  // only set for rate >= most
+  google.protobuf.Duration response_time_p66 = 3;
```
kalvdans: Document what p33 and p66 means. Are they percentiles of the actually responded to messages, or are the yet-to-be-responded to messages added to the distribution at +infinity?

app/backend/src/couchers/servicers/requests.py
```diff
@@ -584,3 +592,75 @@ def MarkLastSeenHostRequest(self, request, context):
...
+                select(
+                    User.id,
+                    func.count().label("n"),
+                    func.count(s.c.time) / (1.0 * func.greatest(func.count(t.c.time), 1)).label("response_rate"),
+                    percentile_disc(0.33)
+                    .within_group(func.coalesce(s.c.time - t.c.time, timedelta(days=1000)))
+                    .label("response_time_p33"),
+                    percentile_disc(0.66)
+                    .within_group(func.coalesce(s.c.time - t.c.time, timedelta(days=1000)))
+                    .label("response_time_p66"),
```
kalvdans: Remove the labels, they are not used anywhere, as far as I can see.

app/proto/requests.proto
```diff
@@ -134,3 +138,24 @@ message MarkLastSeenHostRequestReq {
   int64 host_request_id = 1;
   int64 last_seen_message_id = 2;
 }
+
+message GetResponseRateReq {
+  int64 user_id = 1;
+}
+
+enum ResponseRate {
+  RESPONSE_RATE_UNSPECIFIED = 0;
```
kalvdans, line 147: Remove this enum value which is never returned. Or do protobuf always recommend a dummy 0 value?

app/backend/src/couchers/servicers/requests.py
```diff
@@ -584,3 +592,75 @@ def MarkLastSeenHostRequest(self, request, context):
...
+                .where(Message.author_id == request.user_id)
+                .group_by(Message.conversation_id)
+                .subquery()
+            )
+
+            res = session.execute(
+                select(
+                    User.id,
+                    func.count().label("n"),
+                    func.count(s.c.time) / (1.0 * func.greatest(func.count(t.c.time), 1)).label("response_rate"),
```
kalvdans: what does the `1.0 *` do?

kalvdans (commented):
"There are three kinds of lies" etc :)

app/backend/src/couchers/servicers/requests.py
aapeliv: It's to cast it to a double so that you get double division and not int division

app/backend/src/couchers/servicers/requests.py
aapeliv: Ah, I see. Though I do think they're good for maintainability? I can of course convert them to comments too

app/proto/requests.proto
aapeliv: It's recommended so you can detect a missing/unset value

app/backend/src/couchers/servicers/requests.py
kalvdans: Aha, and a cast is more verbose :) :+1:

app/backend/src/couchers/servicers/requests.py
kalvdans: There's no automated checking that they correspond to the variable names when you unpack the result. So better use comments so that we don't get a false sense of some automated checks.

[EDIT] If they add debugging aid when looking at raw sql debug messages, you can keep them :)

aapeliv:
@kalvdans: I added a lot more documentation for the returned data, hope that's clearer now!

kalvdans:
Thank you. It would be more appropriate to assume unreplied messages follow the same distribution as the replied messages. Otherwise if a person gets bombarded with requests, its p33 and p66 percentiles suddenly goes up, even if the person afterwards replies to them in the usual time.

aapeliv:
> Thank you. It would be more appropriate to assume unreplied messages follow the same distribution as the replied messages. Otherwise if a person gets bombarded with requests, its p33 and p66 percentiles suddenly goes up, even if the person afterwards replies to them in the usual time.

I've thought about this. Firstly, hopefully response times are in the order of a couple of days, so the time during which you have a worse response rate than your usual is limited. Secondly, I do think it's a good "rate limiter"/regulator: if you get a tonne of requests, and your response rate goes down for two days, then fewer people will send you a request and that's presumably a good thing? We'd prefer no individual is bombarded with all requests anyway.

aapeliv:
@kalvdans happy for me to merge?

app/backend/src/couchers/servicers/requests.py
```diff
@@ -584,3 +593,81 @@ def MarkLastSeenHostRequest(self, request, context):
...
+                    User.id,
+                    # number of requests received
+                    func.count().label("n"),
+                    # percentage of requests responded to
+                    (func.count(s.c.time) / func.cast(func.greatest(func.count(t.c.time), 1.0), Float)).label(
+                        "response_rate"
+                    ),
+                    # the 33rd percentile response time
+                    percentile_disc(0.33)
+                    .within_group(func.coalesce(s.c.time - t.c.time, timedelta(days=1000)))
```
kalvdans, line 624: The coalesce function returns the first of its arguments that is not null, but how can the result of a subtraction be null? I don't understand, to me it sounds like we are excluding messages that haven't been responded to.

aapeliv: It's an outer join, so if there is no response to a message (no corresponding row in the `s` subquery) then `s.c.time - t.c.time` will be null, because with an outer join it will just replace missing columns with nulls.

kalvdans: Thanks for explaining SQL to me :)

kalvdans (approved):
Looks good to me! Great with tests.

PR #2730, opened 2022-03-27
use nvm to set node version to 14 inside gitpod

### LINKED ISSUE #2675: incompatible node version when running gitpod
Seems like this issue is fixed for local setup according to #2594 but still happens for gitpod

lpmi-13:
closes https://github.com/Couchers-org/couchers/issues/2675

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -9,6 +9,8 @@ tasks:
   - name: Install and launch the frontend with yarn
     before: cd app/web
     init: |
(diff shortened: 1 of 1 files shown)


lpmi-13:
Unable to request a reviewer via the UI, so tagging @aapeliv here.

darrenvong (approved):

darrenvong:
Thanks for the fix on this too!

PR #2756, opened 2022-04-03
Add reply rate and time to user overview

### LINKED ISSUE #2729: Web response rate display on profile
I've just finished a response rate+time backend call (`GetResponseRate` in `requests.proto`).

My thinking for the frontends would be to display this under "last seen", with text like this:

* `RESPONSE_RATE_INSUFFICIENT_DATA`: "Not enough data"
* `RESPONSE_RATE_LOW`: "Very low response rate"
* `RESPONSE_RATE_SOME`: "Responds to some messages, sometimes in {p33}"
* `RESPONSE_RATE_MOST`: "Responds to most messages, generally in {p33} to {p66}"
* `RESPONSE_RATE_ALMOST_ALL`: "Responds to almost all messages, generally in {p33} to {p66}"

e.g. "Responds to almost all messages, generally in 29 min to 5.5 h"

I'm not sure if this wording is good, etc.

lucaslcode:
**Web frontend checklist**
- [ ] All tests pass

closes #2729


app/web/features/profile/view/UserTextAndLabel.tsx
```diff
@@ -41,6 +46,60 @@ export const LabelsReferencesLastActive = ({ user }: Props) => (
...
+      : query.data?.responseRate === ResponseRate.RESPONSE_RATE_MOST
+      ? t("response_rate_text_most")
+      : query.data?.responseRate === ResponseRate.RESPONSE_RATE_ALMOST_ALL
+      ? t("response_rate_text_almost_all")
+      : undefined;
+  const timeText =
+    query.data?.responseRate === ResponseRate.RESPONSE_RATE_SOME
+      ? t("response_time_text_some", {
+          p33: dayjs
+            .duration(query.data.responseTimeP33!.seconds, "second")
```
darrenvong: According to the comments in the proto:
https://github.com/Couchers-org/couchers/blob/23b406054919a57b7d39eae12d8ecf943edbf863/app/proto/requests.proto#L165-L172

Shouldn't we be checking for `responseTimeP33` and `responseTimeP66` rather than assuming it's always non-null with `!` if it's sometimes not set by the backend?

app/web/features/profile/view/UserTextAndLabel.tsx
```diff
@@ -41,6 +46,60 @@ export const LabelsReferencesLastActive = ({ user }: Props) => (
...
+            .duration(query.data.responseTimeP66!.seconds, "second")
+            .humanize(),
+        })
+      : undefined;
+
+  return (
+    <>
+      <LabelAndText label={t("response_rate_label")} text={rateText ?? ""} />
+      {timeText && (
+        <LabelAndText label={t("response_time_label")} text={timeText} />
```
darrenvong, line 93: What about if `timeText` is undefined? Maybe in the ternary chains above, the last `else`s can fallback to the empty string instead so we don't need to use `??` here (same comment with `rateText`)?

app/web/features/profile/view/UserTextAndLabel.tsx
```diff
@@ -41,6 +46,60 @@ export const LabelsReferencesLastActive = ({ user }: Props) => (
   </>
 );
```
darrenvong, line 45: lol this file has become a bit of a component dump over time as no components in here are even called `UserTextAndLabel` anymore... 🙈

app/web/features/profile/view/UserTextAndLabel.tsx
lucaslcode: As is still follows the comments, as we assume it's not null in the cases it says it will be set. I agree it's not very nice though, I'll have a think

aapeliv: Yeah, I thought about this too. It's not in general good to create APIs that can represent impossible states.

A better option would be to have response rate be a `oneof` where each of the different choices is its own message so that if response rate is `some`, p66 isn't even there.

I can change it if you want?

lucaslcode: Sounds good

lucaslcode: You can do it on this branch if you want

aapeliv: OK, I'll do it in a bit

app/web/features/profile/view/UserTextAndLabel.tsx
lucaslcode: `timeText` is undefined when the rate text is "not enough data" or "rarely responds" or is loading.
`rateText` is only undefined when the user is loading. Maybe `"..."` is a better string? Is there a succinct way to make this better/more clearly typed?

app/web/features/profile/view/UserTextAndLabel.tsx
aapeliv: @lucaslcode: done

app/web/features/profile/view/UserTextAndLabel.tsx
darrenvong: Yeah, can't really think of anything more succinct than what you got already since it's impossible to know the value without checking through all the possible enum values.

Maybe for `timeText` /response time, when the user "rarely responds" or "not enough data", we should either have a fallback to something like "unknown" (or a better message) or hide that label altogether? Otherwise, I think it's a bit confusing to have "Response time: " with no value shown next to it.

app/web/features/profile/view/UserTextAndLabel.tsx
darrenvong: Ahh that's fair, overlooked that you were only checking if the response rate is "some" or higher, but the new API response is probably clearer to understand without the impossible states 👍

app/web/features/profile/view/UserTextAndLabel.tsx
darrenvong: Only thing I can think of to reduce the verbosity slightly is do the check once with the tweaked API response and set both text in the checks? I think if statements/switches are ok as long they are used after the hooks.

Could also contain them in two separate components. That doesn't necessarily reduce the verbosity but at least the logic is grouped together with the UI code that needs it straight after

app/web/features/profile/view/UserTextAndLabel.tsx
lucaslcode: I did some renaming include the name of this file to `userLabels.tsx`

app/web/features/profile/view/userLabels.tsx
```diff
@@ -41,7 +46,59 @@ export const LabelsReferencesLastActive = ({ user }: Props) => (
...
+    rateText = t("response_rate_text_almost_all");
+    timeText = t("response_time_text_almost_all", {
+      p33: dayjs
+        .duration(query.data.almostAll.responseTimeP33!.seconds, "second")
+        .humanize(),
+      p66: dayjs
+        .duration(query.data.almostAll.responseTimeP66!.seconds, "second")
+        .humanize(),
+    });
+  }
```
lucaslcode, line 87: @darrenvong how's this

app/web/features/profile/locales/en.json
```diff
@@ -72,5 +72,15 @@
     "additional_information_section_title": "Additional information",
     "travel_section_title": "Travelled to",
     "lived_section_title": "Lived in",
-    "regions_empty_state": "None"
+    "regions_empty_state": "None",
+    "response_rate_label": "Reponse rate",
```
darrenvong:
```suggestion
    "response_rate_label": "Response rate",
```
Typo :p

app/web/features/profile/view/userLabels.tsx
darrenvong: Yeah I think this will do.

Any idea why response rate is not showing up with a value though in the Vercel build? 
Screenshot 2022-04-08 at 21 49 28

darrenvong (changes requested):

app/web/features/profile/view/userLabels.tsx
lucaslcode: Yeah because aapeli changed the backend in this branch and it isn't merged, so the preview api is still using the old version

darrenvong (commented):
GitHub now says there's a merge conflict, but otherwise looks good!

darrenvong (approved):

PR #2775, opened 2022-04-06
Add meeting notes 2022/04/06

lucaslcode:


aapeliv (approved):
Looks good. Moving the meeting time would be nice so I could join too!

PR #2783, opened 2022-04-10
chore: update jest config to Next's SWC compiler

darrenvong:
According to the [Next docs compiler history](https://nextjs.org/docs/advanced-features/compiler), since 12.1.x, Jest tests can now be run with their Rust SWC compiler rather than needing babel, so I updated our config to use that instead of the old babel transforms.

Fingers crossed this might speed up test runs in our CI too!

**Web frontend checklist**
- [x] All tests pass


lucaslcode (approved):
Nice!

PR #2795, opened 2022-04-13
fix: disable request button when user is not hosting

### LINKED ISSUE #2713: "request" button still appears even if the person has "can't host" selected
We should not be displaying the "request" button if the person has selected "can't host". Right now the button appears like normal despite the "can't host" status and people are complaining to the support team because they are getting requests even though they indicated they cannot host:

image

Tay0108:
Closes https://github.com/Couchers-org/couchers/issues/2713

**Web frontend checklist**
- [x] All tests pass


lucaslcode (approved):
Thanks!!

PR #2835, opened 2022-04-26
Fix storybook dev mode in local (issue with dependencies + CaseSensitivePathsPlugin)

guillermodlpa:
Hi there!

I found that in my Mac, the command `yarn storybook` failed with errors related to case-sensitive imports inside `@storybook` and `@material-ui/core` dependencies. At first, I tried updating these dependencies, but that didn't work (I left the upgrades in this branch anyway).

The webpack plugin `CaseSensitivePathsPlugin` is responsible for enforcing case-sensitivity. I found a way to exclude it from the webpack config for Storybook, and now Storybook works as expected for me.

Would be great to confirm if others can execute `yarn storybook` without issues, particularly on a Mac with case-insensitive APFS, to confirm it wasn't just me and this works for everybody.

### Screenshot of the error

Screen Shot 2022-04-26 at 10 15 02
Screen Shot 2022-04-26 at 10 15 10
Screen Shot 2022-04-26 at 10 15 16
 of the errors

**Web frontend checklist**
- [ ] All tests pass


lucaslcode (approved):
Works for me! And on CI

lucaslcode:
Curious that CI says it's pending when it's actually successful

darrenvong (approved):
Tested on a Mac and works for me now too! This has been annoying me for some time so thanks for having a crack at fixing it, I didn't even think of disabling the plugin as an option 😄 

Nothing immediate seems broken so it's good that you get to bump the packages too (I was hesitant to do more updates with dependabot since this breakage)

guillermodlpa:
Thanks @lucaslcode and @darrenvong !

PR #2842, opened 2022-04-27
Update web setup instructions

guillermodlpa:
1. Include guidance to solve a problem when 3rd party cookies are blocked (that drove me a bit crazy to find out 😅)
2. Favor the option of using `yarn cross-env` to point to the dev API, instead of changing the env var files. It's easier and no need to worry about committing changes.

Demo:

https://user-images.githubusercontent.com/1557348/165460479-cda8ef34-2dd4-4150-9caf-f53b3a9b40e1.mov


darrenvong (approved):
Nice additions! I guess I never notice this problem since I use option 1 and run everything locally 😅  so thanks for spotting!

guillermodlpa:
Thanks for the comment and the approval, @darrenvong !

Issue #2811, opened 2022-04-21
Dashboard V2


lucaslcode:
![Dashboard (1)](https://user-images.githubusercontent.com/32044095/164457613-29fe7ce8-2241-48e1-a9e3-9edcdb5d75b2.png)

This can be split up into several parts:

- [ ] Image header with search bar
- [ ] Profile card updated design (but need to differentiate between the dashboard card and people's profiles' card)
- [ ] References section component
- [ ] Events card slight redesign (Probably no share button for now?)
- [ ] Communities section redesign

PR #2843, opened 2022-04-27
Dashboard v2 layout #2811

guillermodlpa:
Issue: Couchers-org/couchers#4123

**Changes**

Dashboard v2 design, specifically:

1. The new overall layout with profile summary on the left, and the new order of the sections in the right.
2. Remove the 3 expandables at the bottom of the current dashboard.
3. New event card design.
4. New design of list of communities.

**Technical details:**

* I reused the [UserOverview component](app/web/features/profile/view/UserOverview.tsx) from the profile, so I had to do a refactor to make it more reusable.
    * Added a [few unit tests](app/web/features/profile/view/UserOverview.test.tsx) for it.
* It seems frequent that we need to customize the margins around [Divider](app/web/components/Divider.tsx), so to make it more easy and robust, I added a `spacing` prop to it.
* Removed a few custom typography styles in the process in favor of our theme styles.
* Tweaked [Avatar](app/web/components/Avatar/Avatar.tsx) to allow it to render with no user (just for convenience when developing).

**Left to do later:**

1. Share button in profile summary
5. Share button in event cards
7. Pending reference shortcut
8. Button styles in general
9. Cover block with the search bar

**Web frontend checklist**
- [x] All tests pass

**Other checks**
- [x] Tested blocking request URLs to confirm error handling
- [x] Tested with slow connection to confirm loading states

**Views affected to test:**
- [x] Dashboard: /dashboard
- [x] User profile, e.g., /user/gulielmus
- [x] Profile: /profile
- [x] Events: /events
- [x] Write a reference page (because the profile summary is also rendered there), e,g., /leave-reference/hosted/104/96

**Browser compatibility**
- [x] Tested in Mac Chrome
- [x] Tested in Mac Safari 15
- [x] Tested in iOS Safari

---

**Before**

before

---

**After**

after

---

<details>
  <summary>📸 More screenshots</summary>

iPhone SE

https://user-images.githubusercontent.com/1557348/165470919-2f2401a3-f87c-476e-9628-92e8d28984db.mov

MacOS Safari
dashboard-v2_mac13-safari

MacOS Chrome (scrolled)
dashboard-v2-scroll_mac13-chrome

Profile page - MacOS Chrome 
profile_mac13-chrome

Write a reference - MacOS Chrome
writereference_mac13-chrome

</details>


app/web/components/Avatar/Avatar.tsx
```diff
@@ -82,7 +82,7 @@ export default function Avatar({
                 alt={user.name}
                 src={user.avatarUrl}
               >
-                {user.name.split(/\s+/).map((name) => name[0])}
+                {(user.name || "").split(/\s+/).map((name) => name[0])}
```
darrenvong: `user.name` should always exist if `user` is defined so the changes here seem unnecessary

app/web/components/HorizontalScroller.tsx
```diff
@@ -56,13 +56,13 @@ export default function HorizontalScroller({
     >
       {children}
       {fetchNext && hasMore && (
-        <div>
+        <Box display="flex" alignItems="center">
```
darrenvong: I wish I remember the exact reasons (maybe @lucaslcode can help me out with that too...) but I believe we decided to not use `Box` a while back since:
- It's an additional API to learn, whilst CSS is the standard
- It adds extra junk layers in the React DOM tree for each `Box` used, which makes debugging using the dev tool more challenging. Given it proxies everything to the pure CSS equivalent, the "convenience" of using it is not worth it to me
- MUI v5 (when we eventually migrate over...) doesn't seem to support these props, so writing in CSS will make this slightly less painful to port over

So I'd say stick with using `className` with a `div` and write the CSS please

app/web/features/communities/events/EventCard.tsx
```diff
@@ -155,41 +111,41 @@ export default function EventCard({ event, className }: EventCardProps) {
...
-            </ul>
-            <div className={classes.otherInfoSection}>
-              <Typography variant="h4">
-                {t("communities:details_subheading")}
-              </Typography>
-              <Typography className={classes.content} variant="body1">
+
+            <Divider spacing={1} />
+
+            <Box>
```
darrenvong:
```suggestion
            <div>
```
Based on reasons mentioned in previous comment, especially when we don't need it for additional styling rules other than grouping child elements for flex/grid etc

app/web/features/dashboard/CommunitiesSection.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <>
+      <Typography variant="h2" gutterBottom>
+        {t("dashboard:your_communities_heading")}
+      </Typography>
+      <Typography variant="body1" paragraph>
+        <Trans i18nKey="dashboard:your_communities_helper_text">
+          {`You have been added to all communities based on your location. Feel free to `}
+          <MuiLink
+            component="button"
+            style={{ verticalAlign: "baseline" }}
```
darrenvong: Do this via `className` and add it to the `useStyles` for consistency on how it's done throughout the code

app/web/features/dashboard/CommunitiesSection.tsx
```diff
@@ -0,0 +1,70 @@
...
+            onClick={() => {
+              setIsCommunitiesDialogOpen(true);
+            }}
+          >
+            {/* @todo: revisit this UI. A button that opens a popup shouldn't look like a link */}
```
darrenvong, line 40: Good point. This is bugging me quite a lot across our UI - we don't seem to have consistency between what we put in a link and button at all even between the designs!

I am semi-ok with links looking like a button as that feels more like implementation details (performing an action by taking the user to a different location happens to be easier), but a button that looks like a link definitely feels wrong to me. Maybe a conversation to be had with design...

app/web/features/dashboard/CommunitiesSection.tsx
```diff
@@ -0,0 +1,70 @@
...
+  createCommunityText: {
+    paddingBlockStart: theme.spacing(2),
+  },
+}));
+
+export default function CommunitiesSection() {
+  const { t } = useTranslation([GLOBAL, DASHBOARD]);
+  const classes = useStyles();
+  const [isCommunitiesDialogOpen, setIsCommunitiesDialogOpen] =
+    useState<boolean>(false);
```
darrenvong:
```suggestion
    useState(false);
```
Type inference is enough here and we prefer to lean on that as much as possible to write less code

app/web/features/dashboard/Dashboard.tsx
```diff
@@ -1,178 +1,77 @@
...
 export default function Dashboard() {
   const { t } = useTranslation([GLOBAL, DASHBOARD]);
   const classes = useStyles();
-  const [isCommunitiesDialogOpen, setIsCommunitiesDialogOpen] = useState(false);
 
   return (
-    <>
-      <HtmlMeta title={t("global:nav.dashboard")} />
-      <PageTitle>{t("dashboard:welcome")}</PageTitle>
+    <Container maxWidth="md">
```
darrenvong: Why is this only taking a `maxWidth` of `md` rather than `lg` as defined by the container at the route level? Is it to roughly match how it looks on the profile page?

The profile page should probably be doing what you're doing here rather than specifying the max width with the magic value of `"61.5rem"`... (but that's probably out of scope from what you're doing so this is more of a side comment of things I noticed from reviewing this!)

app/web/features/dashboard/DashboardUserProfileSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+export default function DashboardUserProfileSummary() {
+  const { data: user, error, isLoading } = useCurrentUser();
+  const desktopMode = useMediaQuery((theme: Theme) =>
+    theme.breakpoints.up("sm")
+  );
+  return (
+    <>
+      {error && <Alert severity="error">{error}</Alert>}
+      {isLoading ? (
+        <Box display="flex" justifyContent="center">
```
darrenvong: Use a `div` over `Box` please

app/web/features/dashboard/MyEvents.test.tsx
```diff
@@ -92,8 +92,8 @@ describe("My events", () => {
       expect(screen.getAllByRole("link")).toHaveLength(3);
       expect(listMyEventsMock).toHaveBeenCalledTimes(2);
       expect(listMyEventsMock.mock.calls).toEqual([
-        [{ pageSize: 3 }],
-        [{ pageToken: "2", pageSize: 3 }],
+        [{ pageSize: expect.any(Number) }],
+        [{ pageToken: "2", pageSize: expect.any(Number) }],
```
darrenvong:
```suggestion
        [{ pageSize: 2 }],
        [{ pageToken: "2", pageSize: 2 }],
```
I think it would be good to keep the exact `pageSize` assertion as before rather than say "any number". Otherwise if we change this in the future, we wouldn't notice something has changed and whether that change is intentional

app/web/features/dashboard/MyEvents.tsx
```diff
@@ -46,16 +59,21 @@ export default function MyEvents() {
     useInfiniteQuery<ListMyEventsRes.AsObject, RpcError>({
       queryKey: myEventsKey,
       queryFn: ({ pageParam }) =>
-        service.events.listMyEvents({ pageToken: pageParam, pageSize: 3 }),
+        service.events.listMyEvents({
+          pageToken: pageParam,
+          pageSize: PAGE_SIZE,
+        }),
       getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
     });
```
darrenvong, line 72: Use `div`s and CSS over `Box` please

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -43,25 +48,79 @@ const useStyles = makeStyles((theme) => ({
...
+  actions?: React.ReactNode;
+  showHostAndMeetAvailability?: boolean;
+}>) {
   const classes = useStyles();
   const user = useProfileUser();
 
   return (
     <Card className={classes.card}>
-      <Avatar user={user} grow />
+      <Box maxWidth="75%" mx="auto">
```
darrenvong: Use a `div` with CSS over `Box` please

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -43,25 +48,79 @@ const useStyles = makeStyles((theme) => ({
...
-export default function UserOverview({ children }: PropsWithChildren<unknown>) {
+// @todo: move this into /components and decouple it from features/profile because it's used
+//        from the dashboard as well
+export default function UserOverview({
+  actions,
+  showHostAndMeetAvailability = true,
+}: PropsWithChildren<{
+  actions?: React.ReactNode;
+  showHostAndMeetAvailability?: boolean;
+}>) {
```
darrenvong:
```suggestion
}: {
  actions?: React.ReactNode;
  showHostAndMeetAvailability?: boolean;
}) {
```
Since this component is no longer using `children` to render stuff? Or better still, pull that into `UserOverviewProps` so it's easier to read in autocompletes when hovering over the component's usage elsewhere

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -43,25 +48,79 @@ const useStyles = makeStyles((theme) => ({
...
   return (
     <Card className={classes.card}>
-      <Avatar user={user} grow />
+      <Box maxWidth="75%" mx="auto">
+        <Avatar user={user} grow />
+      </Box>
+
       <div className={classes.wrapper}>
-        <Typography variant="h1" className={classes.intro}>
+        <Typography variant="h1" className={classes.intro} align="center">
```
darrenvong:
```suggestion
        <Typography variant="h1" className={classes.intro}>
```
Probably nitpick, but I'd add it into the `intro` CSS so everything is in one place rather than having this random styling here

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -43,25 +48,79 @@ const useStyles = makeStyles((theme) => ({
...
+    },
+  },
 }));
 
-export default function UserOverview({ children }: PropsWithChildren<unknown>) {
+// @todo: move this into /components and decouple it from features/profile because it's used
+//        from the dashboard as well
+export default function UserOverview({
+  actions,
+  showHostAndMeetAvailability = true,
```
darrenvong: I think it's more conventional to default to `false` for optional props like these, rather than the other way around. I'd find it surprising behaviour if `<UserOverview showHostAndMeetAvailability />` results in the same as when the prop is not passed through.

app/web/features/profile/view/UserOverview.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  describe("when user is loaded and provided via context", () => {
+    it("should display the user name", () => {
+      const { getByText } = render(
+        <ProfileUserProvider user={defaultUser}>
+          <UserOverview showHostAndMeetAvailability={false} />
+        </ProfileUserProvider>,
+        { wrapper }
+      );
+      expect(getByText(defaultUser.name)).toBeInTheDocument();
```
darrenvong:
```suggestion
      expect(screen.getByText(defaultUser.name)).toBeInTheDocument();
```
Prefer using [screen](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) as recommended by the RTL author rather than destructuring the result of `render` to avoid having to update the destructure if we ever change these tests. Same comment for everywhere else throughout these tests!

app/web/features/profile/view/UserOverview.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <UserOverview showHostAndMeetAvailability />
+        </ProfileUserProvider>
+      );
+      expect(getByText(expectedLabelHosting)).toBeInTheDocument();
+      expect(getByText(expectedLabelMeeting)).toBeInTheDocument();
+      rerender(
+        <ProfileUserProvider user={defaultUser}>
+          <UserOverview />
+        </ProfileUserProvider>
+      );
```
darrenvong: I think my comment of defaulting to `false` instead in the component code has another additional benefit, you only need to test the case of being explicitly passed in rather than the missing case too.

app/web/features/profile/view/UserOverview.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+      const { queryByText, getByText, rerender } = render(
+        <ProfileUserProvider user={defaultUser}>
+          <UserOverview showHostAndMeetAvailability={false} />
+        </ProfileUserProvider>,
+        { wrapper }
+      );
+      expect(queryByText(expectedLabelHosting)).not.toBeInTheDocument();
+      expect(queryByText(expectedLabelMeeting)).not.toBeInTheDocument();
+      rerender(
```
darrenvong: I'd call `render` again in another test case rather than using `rerender` altogether, since it doesn't render with all of the right providers that you passed in from `wrapper` in the original `render`. It's also a bit of an escape hatch and generally should be avoided, since it doesn't encourage testing from how user would interact with the UI code.

app/web/features/profile/view/UserOverview.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+describe("UserOverview", () => {
+  beforeEach(() => {
+    addDefaultUser();
+  });
+
+  describe("when user is missing", () => {
+    it("doesn't fail", () => {
+      render(<UserOverview />, { wrapper });
+    });
+  });
```
darrenvong: This is a bit of a false test as this component has a dependency on `ProfileUserProvider` existing up the tree, so I don't think this test should exist to give the false impression that the component can be used without a provided user...

darrenvong (changes requested):
Apart from the `Box` -> `div` (makes up most of the comments 😅 ) and the odd comments here and there, this is really great work! Thanks for the thorough testing over different browsers with all of the screenshots as well

app/web/components/HorizontalScroller.tsx
lucaslcode: I think the main reason was to avoid splitting styles between two places. So yeah, we switched over to preferring `makeStyles` and `<div className={classes.x}>`

app/web/features/profile/view/UserOverview.tsx
lucaslcode: (If you wanted to avoid adding the new prop to a bunch of places, change it to `hideHostAndMeetAvailability`)

app/web/components/HorizontalScroller.tsx
```diff
@@ -13,7 +13,7 @@ const useStyles = makeStyles((theme) => ({
     display: "inline-flex",
     flexDirection: "row",
     height: "100%",
-    width: "100%",
+    width: "100vw",
```
lucaslcode, line 16: Why was this changed? This could cause horizontal overflow issues if used in a container with a margin, though I do believe we always use it with 0 margin atm

app/web/features/dashboard/CommunitiesList.tsx
```diff
@@ -49,21 +50,21 @@ export default function CommunitiesList({ all = false }: { all?: boolean }) {
             {communities.data.pages
               .flatMap((page) => page.communitiesList)
               .map((community) => (
-                <Link
+                <NextLink
```
lucaslcode: Is the `<StyledLink>` component suitable for what you want to do here?

app/web/features/dashboard/DashboardUserProfileSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <CircularProgress />
+        </Box>
+      ) : user ? (
+        desktopMode ? (
```
lucaslcode, line 54: This could cause problems later if we want to do SSR. But I imagine we won't do that for the dashboard, so probably fine here.

lucaslcode (dismissed):
Just a few comments/questions to add

app/web/features/dashboard/DashboardUserProfileSummary.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <ProfileUserProvider user={user}>
+            <UserOverview
+              actions={<DashboardUserProfileSummaryActions />}
+              showHostAndMeetAvailability
+            />
+          </ProfileUserProvider>
+        ) : (
+          <MinimalUserProfileCard user={user} />
+        )
+      ) : undefined}
```
darrenvong:
```suggestion
      ) : null}
```
More conventional in React to mean we want to render nothing rather than `undefined`. Also I believe it's invalid to render `undefined` until we upgrade to React 18 🤔

app/web/components/Avatar/Avatar.tsx
guillermodlpa: sounds good. I'll undo this. It was only to avoid errors as I was work-in-progress. It shouldn't happen in a real env

app/web/components/HorizontalScroller.tsx
guillermodlpa: Sounds good! I'll update all my code here and in future contributions to using class names

app/web/features/dashboard/CommunitiesSection.tsx
guillermodlpa: Ah yes, sounds good. Sometimes I feel `useStyles` is so much boilerplate and I shortcut to this approach when it's a small stylistic change. I'll stop doing this and use `className` / `useStyles` always moving forward!

app/web/features/dashboard/CommunitiesSection.tsx
guillermodlpa: Sounds good, I'll do that moving forward here.

app/web/features/dashboard/Dashboard.tsx
guillermodlpa: Exactly, I was trying to match more or less the same width as the profile page, but without hardcoding any value.

I hadn't realized that we had the layout defined in `components/AppRoute.tsx` `appGetLayout`. That's great! The `Container` I added is totally unnecessary, I'm gonna remove it and let the `maxWidth` come from there, which is `lg`. It looks wider, but still good, and more consistent with other views I imagine. 

Thanks!

app/web/features/dashboard/MyEvents.test.tsx
guillermodlpa: Sounds good! I'll make a variable in the test file, something like `cardPerRow` and use it here, so the number 2 has a meaning beyond being a hardcoded page size. Thanks!

app/web/features/profile/view/UserOverview.tsx
guillermodlpa: Ah good point! On it

app/web/features/profile/view/UserOverview.tsx
guillermodlpa: Sounds good! Makes sense

app/web/features/profile/view/UserOverview.tsx
guillermodlpa: I see. This is a tricky opinionated topic haha. 

I tend to dislike the `hide...` settings defaulting to false due to seeing DB field names using this pattern, which made the logic harder to read and maintain than it could have been.

For now, I lean towards making it not default to anything, and I'll make it required, so we're explicitly defining if it should show or not. The component is only used in a handful of places, so passing the prop isn't actually much overhead in terms of quantity of code

app/web/features/profile/view/UserOverview.test.tsx
guillermodlpa: Oh good to know, I wasn't aware of this recommendation! I'll update the tests. Thank you!

app/web/features/profile/view/UserOverview.test.tsx
guillermodlpa: totally

app/web/features/profile/view/UserOverview.test.tsx
guillermodlpa: that makes sense. I'll update it. Thanks

app/web/features/profile/view/UserOverview.test.tsx
guillermodlpa: you're right. I'll remove it

app/web/features/profile/view/UserOverview.tsx
lucaslcode: Sounds good to me!

app/web/components/HorizontalScroller.tsx
guillermodlpa: Yeah, that's something we need to be careful with, and maybe I need to test more across browsers and everywhere this is used.

With `100%`, when using `HorizontalScroller` in the dashboard or any other view that has the app layout (because it has paddings on the side), this is what it looks like:


https://user-images.githubusercontent.com/1557348/166420702-306921ea-fe1c-4ac9-97ba-98193b20f1cc.mov


The tricky thing of this use case is that the horizontal scroller class name is also combined with class names passed. Here, `classes.upcomingEventContainer` includes a `100vw` rule that is being overwritten. 

Screen Shot 2022-05-03 at 09 54 39

Also, `classes.cardContainer` seems to be setting the scroller to be centered and bypassing the paddings. It seems that these 3 sets of styles (`classes.cardContainer`, `classes.upcomingEventContainer`, and horizontal scroller's `classes.root`) are conflicting a bit with each other.

Screen Shot 2022-05-03 at 09 58 01

Maybe for now I can leave it with `100%`, despite the visual bug, and we tackle later refactoring the horizontal scroller so it's more self-contained and it works without needing to pass any styles from outside it? How does that sound?

app/web/features/dashboard/CommunitiesList.tsx
guillermodlpa: Oh yes! Thanks!

app/web/features/dashboard/DashboardUserProfileSummary.tsx
guillermodlpa: ah good catch, I hadn't considered that. When that time comes, we could simply not mount this part of the dashboard. But let me know if you prefer me to do a different implementation

app/web/features/dashboard/DashboardUserProfileSummary.tsx
guillermodlpa: Cool, will do

app/web/components/HorizontalScroller.tsx
lucaslcode: Thanks for the explanation! Just had a look.

I think your change is indeed the best/easiest solution, this component is always supposed to be full-width on mobile.

Also seems to me like we can get rid of `classes.upcomingEventContainer` completely since the only line is overriden or has no effect at any breakpoint anyway as far as I can tell.

app/web/features/profile/view/UserOverview.tsx
darrenvong: Yeah sounds good, at least making it required means we won't have the confusion I described earlier! 👍

app/web/features/dashboard/MinimalUserProfileCard.tsx
```diff
@@ -1,41 +1,53 @@
...
+import makeStyles from "utils/makeStyles";
+
+const useStyles = makeStyles((theme) => ({
+  container: {
+    display: "flex",
+    flexDirection: "row",
+    paddingTop: theme.spacing(1),
+    paddingBottom: theme.spacing(1),
+    paddingLeft: theme.spacing(2),
+    paddingRight: theme.spacing(2),
```
darrenvong:
```suggestion
    padding: theme.spacing(1, 2),
```
This is pretty minor, but I guess this shorthand would emulate the `px` and `py` props `Box` was trying to achieve with slightly less code.

app/web/features/profile/view/UserOverview.test.tsx
```diff
@@ -17,61 +17,58 @@ describe("UserOverview", () => {
...
+          <ProfileUserProvider user={defaultUser}>
+            <UserOverview showHostAndMeetAvailability />
+          </ProfileUserProvider>,
+          { wrapper }
+        );
+        expect(screen.getByText(expectedLabelHosting)).toBeInTheDocument();
+        expect(screen.getByText(expectedLabelMeeting)).toBeInTheDocument();
+      });
+
+      it("should npt display the hosting and meeting status when showHostAndMeetAvailability is false", () => {
```
darrenvong:
```suggestion
      it("should not display the hosting and meeting status when showHostAndMeetAvailability is false", () => {
```
Small typo

darrenvong:
> Left to do later:
>
> Number of references in profile summary

Isn't that already in the `UserOverview` component?

darrenvong (changes requested):
Only one minor typo left really, with the other comment being a more subjective way of doing the same thing so I'll leave it for you to decide if you want to follow through with that.

app/web/features/dashboard/MinimalUserProfileCard.tsx
guillermodlpa: ah great, thanks!

guillermodlpa:
> Isn't that already in the `UserOverview` component?

@darrenvong, you're right, the number is there already. I'm removing that from the list of to do's

darrenvong (approved):

PR #2844, opened 2022-04-27
fix message input overflow on android

wginsberg:
This has been bugging me on mobile. Hopefully this change makes it a little bit better.

| before | after |
|---|---|
|165563982-ff304dbc-534c-4b46-9163-47a4e672aad1-1| IMAGE 2022-04-28 11:45:49|

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/features/messages/groupchats/GroupChatSendField.tsx
+++ b/app/web/features/messages/groupchats/GroupChatSendField.tsx
@@ -66,6 +66,7 @@ export default function GroupChatSendField({
         onChange={(event) => setPersistedMessage(event.target.value)}
         maxRows={4}
         size="small"
+        className={classes.textField}
       />
 
       <Button
--- a/app/web/features/messages/requests/HostRequestSendField.tsx
+++ b/app/web/features/messages/requests/HostRequestSendField.tsx
@@ -254,6 +254,7 @@ export default function HostRequestSendField({
           onKeyDown={handleKeyDown}
           maxRows={6}
           size="small"
+          className={classes.textField}
         />
         <FieldButton
           callback={onSubmit}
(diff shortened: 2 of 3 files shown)


darrenvong:
I think it's better if the input and send button container (if there is one 🤔 ) push the footer down? Looks a bit more usable with the fix but still a bit strange to have the input overflowing/rendered on top of the footer imo...

goskan93:
sob., 30 kwi 2022, 01:42 użytkownik Darren Vong ***@***.***>
napisał:

darrenvong:
Did you mean to reply via email @goskan93? What are you trying to tell us? 😄

wginsberg:
> I think it's better if the input and send button container (if there is one 🤔 ) push the footer down? Looks a bit more usable with the fix but still a bit strange to have the input overflowing/rendered on top of the footer imo...

definitely, this is a bandaid fix because I couldn't figure out the css to fix the layout :P

the screenshot doesn't totally do this bug justice but when you're actually typing a message it is really really bad

lucaslcode:
What browser/OS and what page does this happen on?

wginsberg:
Screenshots in the description are from Android Chrome on a Pixel 3. The overflow was happening when I was responding to a hosting request and typing a multiline message.

I was able to partially reproduce it for other chat pages too - this is using Chrome dev tools to simulate a Nexus 5 with the soft keyboard up
Screen Shot 2022-05-02 at 10 16 01 AM

lucaslcode:
Looks like this is caused by this in `GroupChatView.tsx`:
```
pageWrapper: {
  ...
  height: `calc(100vh - ${theme.shape.navPaddingXs} - 56px)`,
```

So when the soft keyboard is up, the height is tiny so the middle flex item overflows and appears on top of the footer flex item.

An alternative is to set `overflowY` on that flex item, but then you're less likely to see what you're typing as you type it. I think this your solution is good considering how much of an edge case this is.

lucaslcode (approved):
Added an explanatory comment for future us

PR #2849, opened 2022-04-29
fix footer to bottom of screen

### LINKED ISSUE #2115: Footer not fixed to bottom of screen
The footer isn't fixed to the bottom of the screen. See example in the screenshot below:

image

This appears to be the case on all pages where the footer appears including the following pages:

https://app.couchers.org/ (dashboard)
https://app.couchers.org/contribute
https://app.couchers.org/donate

wginsberg:
closes #2115 

**Web frontend checklist**
- [x] All tests pass

Screen Shot 2022-04-29 at 1 54 41 PM

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -24,6 +24,7 @@ export const useAppRouteStyles = makeStyles((theme) => ({
     paddingLeft: theme.spacing(2),
     paddingRight: theme.spacing(2),
     paddingBottom: theme.spacing(2),
+    flex: 1,
   },
   fullWidthContainer: {
     margin: "0 auto",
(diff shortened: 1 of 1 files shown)


app/web/components/AppRoute.tsx
```diff
@@ -48,6 +50,14 @@ export const useAppRouteStyles = makeStyles((theme) => ({
       )})`,
     },
   },
+  "@global body": {
+    height: "100%",
```
darrenvong, line 54: I think we had this a while back (setting 100% height on html and body) but do you happen to remember why we removed it @lucaslcode?

The only thing I remember is the weird background problem where the background colour of body ends at 100% height even when the inner stuff is taller than that, but this solution doesn't appear to suffer from that so maybe this has always been the way to go 🤔

lucaslcode: Can't remember!

lucaslcode (changes requested):
On mobile, I get some kind of padding on the right (you can see backgrounds don't cover the whole horizontal width).

It seems to be caused by the card style in `LandingPage.tsx`:
```
  card: {
    margin: theme.spacing(0, -4),
    padding: theme.spacing(4),
  },
```

I remember this was a trick to get the card to be full-width on mobile. I guess giving html and body height makes this not work anymore. Happy with any other solution that looks okay!

lucaslcode (approved):
Apologies, looks like that didn't originate in this PR

PR #2858, opened 2022-05-01
Add Breton to languages json

ItsiW:
**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/resources/languages.json
+++ b/app/backend/resources/languages.json
@@ -192,6 +192,10 @@
     "name": "Braj"
   },
   {
+    "code": "bre",
+    "name": "Breton"
+  },
+  {
     "code": "bug",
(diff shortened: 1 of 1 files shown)


ItsiW:
@darrenvong, Aapeli mentioned we'd need to do a migration as well. Could you help with this? Also is this easy enough to do to warrant making updates to languages one at a time, or should we wait to get a few language requests to do them in bulk?

lucaslcode:
This commit shows what the migration should roughly look like: https://github.com/foormea/couchers/commit/8c9f3da8a2963a574ed3cd9f178e6f86ca4780de

But I think let's wait until we have a few?

ItsiW:
Sure thing, @AllhandsNoFeet will let us know once there are a few requests.

PR #2885, opened 2022-05-07
Remove duplicated menu items on mobile

ItsiW:
**Web frontend checklist**
- [ ] All tests pass

Matching mobile menu to desktop


lucaslcode (dismissed):
Maybe Forum should also be removed from the right dropdown, since it's in the main navigation on desktop and mobile?

ItsiW:
Yeah I was considering removing it from the left menu but that makes more sense.

lucaslcode (approved):

PR #2904, opened 2022-05-10
Rename button label "Filters" to "Search" on mobile

### LINKED ISSUE #2809: Rename "Filters" to "Search" on mobile

guillermodlpa:
Closes #2809

On desktop, since there's a search input visible, it makes sense for the button to say "Filters". But on mobile, since the search box isn't visible, it's caused confusion. "Search" is expected to be more clear for users.

**Technical details:**

* I noticed the button component for both mobile and desktop was the same in the code. I separated them because they feel so different, visually, that I thought a bit of duplication was preferable to differentiate them and make future changes be easier. I'm not strongly opinionated and I'm happy to undo this if you prefer how it was before!
* The translations file exists, but it isn't used yet. I made the same label changes there too, to keep it consistent until the constants are replaced.

**Demo:**

https://user-images.githubusercontent.com/1557348/167717878-34bc49cb-9680-44a3-a355-1c8e101995a5.mov

**Web frontend checklist**
- [n/a] Added any new components to storybook
- [n/a] Added tests where relevant
- [x] All tests pass


app/web/features/search/SearchBox.tsx
```diff
@@ -118,25 +119,29 @@ export default function SearchBox({
...
+    />
   );
+
   if (isSmDown) {
-    return filterDialogButton;
+    return (
+      <>
+        <Button
+          onClick={() => setIsFiltersOpen(true)}
+          className={classNames(className)}
```
darrenvong:
```suggestion
          className={className}
```
No need to use the `classNames` function if `className` is the only list of classes that will be applied here?

app/web/features/search/SearchBox.tsx
```diff
@@ -118,25 +119,29 @@ export default function SearchBox({
...
+          variant="contained"
+          size="small"
+        >
+          {FILTER_DIALOG_TITLE_MOBILE}
+        </Button>
```
darrenvong: Do we want the button to stretch almost the full width of the map, rather than sticking next to the +/- buttons as before?

darrenvong (commented):
I think what you've done makes sense to me (separating the button, unless @lucaslcode has stronger opinion on this since he implemented most of this 😄 ).

Thanks for keeping the translations in sync while we are still in the process of swapping over to using them completely!

lucaslcode:
I agree, looks good after Darren's comments

app/web/features/search/SearchBox.tsx
guillermodlpa: of course, thanks!

app/web/features/search/SearchBox.tsx
guillermodlpa: Hmm. You mean like moving the +/- buttons a bit below, so the "Search" button has full width? I hadn't heard this idea before

darrenvong: Sorry for the slow response! I meant literally as the styling before where the button is not stretched and placed to the left of the +/- buttons.

To rephrase, the question perhaps should have been "Is stretching the button (your change) a desirable behaviour?"

guillermodlpa: Ah I see. I didn't change that, I think 🤔 The button is already stretched in the live site:

Screen Shot 2022-05-15 at 15 26 22

Is there a difference, apart from the button label, that I'm not seeing?

lucaslcode: It used to not be stretched! 😆 
I think not stretched looks a little nicer

guillermodlpa: I see! No problem, I can make it not stretched then, and placed on the left of the +/- buttons. Will do in this same branch soon

guillermodlpa: Here's what it looks like aligned next to the zoom buttons without stretching:

Screen Shot 2022-05-16 at 09 28 12

Let me know!

darrenvong (approved):

PR #2918, opened 2022-05-14
Dashboard Hero Section

guillermodlpa:
Part of Couchers-org/couchers#4123

This pull request adds the Hero section on the top of the dashboard, with a prominent search input and a beautiful image on the background.

**Technical Notes:**

Refactors:

* I reused the [LocationAutocomplete component](app/web/components/LocationAutocomplete/LocationAutocomplete.tsx), so I moved it from `/features/search` to `/components`. I added some extra props as well to use a different variant for the dashboard.
* For the same reason, I had to reuse the type `SearchFilters` and the utility functions to parse it to URL format, so I moved those to [/utils/SearchFilters](app/web/utils/SearchFilters.ts)
* Additionally, I renamed `useSearchFilters` to [useRouteWithSearchFilters](app/web/features/search/useRouteWithSearchFilters.ts) to clarify that what the hook is doing is reading/writing in the URL query a search filters state.

Approaches:

* The [hero image](app/web/features/dashboard/Hero/mesut-kaya-eOcyhe5-9sQ-unsplash.jpeg) is rendered with `next/image`, so it appears blurred at first when loading, and we can take advantage of Next's image handling. Let me know if there's another way at Couchers that I should do it with.
* The hero section doesn't have much logic, so I didn't write tests for it. Let me know if any unit test that is worth it comes to mind.
* I made some changes in the `useEffect` of [useRouteWithSearchFilters](app/web/features/search/useRouteWithSearchFilters.ts), because before it caused an infinite redirect when the effect was triggered before `router.isReeady` was true and when `active` was passed as the object to trigger the effect by.

**Screenshots**

Screen Shot 2022-05-14 at 12 32 28

Screen Shot 2022-05-14 at 12 34 33
Screen Shot 2022-05-14 at 12 36 43

**Web frontend checklist**

- [x] All tests pass


app/web/features/dashboard/Hero/HeroButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+import Button from "components/Button";
+import { DASHBOARD } from "i18n/namespaces";
+import Link from "next/link";
+import { useTranslation } from "next-i18next";
+import { useMemo } from "react";
+import { searchRoute } from "routes";
+import makeStyles from "utils/makeStyles";
+
+// because this component is over an image background and has a special button, we adjust the theme
+const getHeroButtonTheme = (theme: Theme) =>
```
lucaslcode: Any particular reason to use a theme and not `className`? We can target inner elements if necessary with the relevant mui classes

app/web/features/search/useRouteWithSearchFilters.ts
```diff
@@ -0,0 +1,64 @@
...
+  const router = useRouter();
+  const [active, setActive] = useState<SearchFilters>(() =>
+    parsedQueryToSearchFilters(router.query)
+  );
+  const pending = useRef<SearchFilters>(active);
+
+  const expectedRouteWithQuery = `${route}?${parseSearchFiltersToQuery(
+    active
+  )}`;
+  const routerQueryIsReady = router.isReady;
```
lucaslcode, line 21: Didn't know about this when switching from cra to next! Did you encounter any bugs?

app/web/utils/SearchFilters.ts
```diff
@@ -1,61 +1,15 @@
-import { useRouter } from "next/router";
 import { ParsedUrlQuery } from "querystring";
```
lucaslcode: this file (and test file) should be uncapitalized

app/web/features/dashboard/Hero/HeroButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+  createTheme({
+    ...theme,
+    palette: {
+      ...theme.palette,
+      primary: {
+        main: "#ffffff",
```
lucaslcode: `theme.palette.common.white` (I think)

lucaslcode (changes requested):
Appreciate this PR a lot, especially the effort needed to sort through the old `useSearchFilters` code and add tests!

Only real question I have is about the usage of themes for the custom styling

app/web/features/dashboard/Hero/HeroButton.tsx
guillermodlpa: cool, makes sense. Doing now

app/web/utils/SearchFilters.ts
guillermodlpa: gotcha, cool. I wasn't sure since the main export was `SearchFilters` type, but there are other things too, so it makes sense lower case

app/web/features/search/useRouteWithSearchFilters.ts
guillermodlpa: yeah, when navigating with React Router to `earch?location=...`, `useEffect` below was being invoked when the router wasn't ready, and so `router.query` was empty. This didn't happen when opening the URL directly in a new tab or something (for example, opening this link: https://couchers.org/search?location=Barcelona%2C+Catalonia%2C+Spain&lat=41.3828939&lng=2.1774322)

It's tricky. React Router added this flag not so long ago I think. I came across issues from not checking it in another project recently, I think it's related to Next.js SSR, but I haven't dug deeper yet

app/web/features/dashboard/Hero/HeroButton.tsx
guillermodlpa: Yeah, I've opted to set a custom theme. My reasoning is that changing the color of the button feels like an override of an internal behavior of the library (applying the primary color on the primary button backgrounds), and we don't really know what the library does or will do in the future. Like maybe MUI in future releases will make the hover color be a darkened version of the button color, like Charka UI does. So either you know the internals of the library and override all the necessary CSS rules, or you pass a custom theme and don't worry about how the library sets or manipulates the theme colors when they are used on buttons or other components.

Your comment is a good prompt though, because reviewing this, it feels a bit strange to add a custom theme just for 1 button, when the entire hero block has also different styling needs.

So I've given it a pass in [this commit](https://github.com/Couchers-org/couchers/pull/2918/commits/2e3a42da41a4a4d36d79243d15b025b8a8ee7d3f), making a single new theme for the Hero block contents that are over the background image, and applying this to all the components over it. 

If we change the background image, and these colors should change, we can just change this theme.

Let me know your thoughts! :)

app/web/features/dashboard/Hero/HeroSearch.tsx
```diff
@@ -0,0 +1,60 @@
...
+export default function HeroSearch() {
+  const { t } = useTranslation(DASHBOARD);
+  const classes = useStyles();
+  const router = useRouter();
+  const searchInputId = "hero-search-input"; // @todo: replace with React 18's useId
+
+  const { control, errors } = useForm({ mode: "onChange" });
+
+  return (
+    <div className={classes.searchBoxContainer}>
```
darrenvong: Since you're using React hook form here, it might be more semantically correct to use a `form` element here?

app/web/features/dashboard/Hero/HeroSearch.tsx
```diff
@@ -0,0 +1,60 @@
...
+    borderRadius: theme.shape.borderRadius,
+    backgroundColor: theme.palette.background.paper,
+  },
+}));
+
+export default function HeroSearch() {
+  const { t } = useTranslation(DASHBOARD);
+  const classes = useStyles();
+  const router = useRouter();
+  const searchInputId = "hero-search-input"; // @todo: replace with React 18's useId
```
darrenvong: Does this ID need to be random?

app/web/features/dashboard/Hero/HeroLinks.tsx
```diff
@@ -0,0 +1,97 @@
...
+  tabAppearance: {
+    position: "relative",
+    paddingBottom: theme.spacing(1),
+    "&::after": {
+      content: '""',
+      position: "absolute",
+      bottom: 0,
+      left: "50%",
+      transform: "translateX(-50%)",
+      width: 40,
```
darrenvong, line 31: Probably a minor thing but just a thought - wouldn't this look a bit weird when the link text is too long, since this decorative line for indicating it's active would remain the same width/length? Might it be worth making this a percentage width instead? 🤔

app/web/utils/searchFilters.test.ts
```diff
@@ -0,0 +1,38 @@
+import { parse } from "querystring";
+
+import {
+  parsedQueryToSearchFilters,
+  parseSearchFiltersToQuery,
+} from "./searchFilters";
+
+describe("parsedQueryToSearchFilters", () => {
+  it("should return a SearchFilters object parsing the URL parameters", () => {
+    const parsedQuery = parse(
```
darrenvong, line 30: I guess since you're already at it... 😅  it'd be nice to exhaust other code paths in these tests too, like `radius` and `hostingStatusOptions` etc since it's logic we wrote compared to the opposite conversion where we can rely on `URLSearchParams` (a browser API)

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
```diff
@@ -0,0 +1,38 @@
...
+      },
+    },
+  });
+
+/**
+ * Hook that returns a modified theme that adapts text to be over the hero image background
+ */
+export default function useHeroBackgroundTheme(): Theme {
+  const theme = useTheme();
+  const imageOverlayTheme = useMemo(() => getTheme(theme), [theme]);
```
darrenvong, line 36: Is this actually expensive to re-create each time? Just wondering if the memo optimisation is necessary

app/web/features/dashboard/Hero/HeroButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+    "-webkit-text-fill-color": "transparent",
+  },
+  buttonContainer: {
+    display: "flex",
+    justifyContent: "center",
+    margin: theme.spacing(6, 0),
+  },
+}));
+
+export default function HeroButon() {
```
darrenvong:
```suggestion
export default function HeroButton() {
```
Typo

app/web/features/dashboard/Hero/HeroButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { searchRoute } from "routes";
+import makeStyles from "utils/makeStyles";
+
+import useHeroBackgroundTheme from "./useHeroBackgroundTheme";
+
+const useStyles = makeStyles((theme) => ({
+  textGradient: {
+    background: `-webkit-linear-gradient(0deg, ${theme.palette.primary.main}, ${theme.palette.secondary.main})`,
+    "-webkit-background-clip": "text",
+    "-webkit-text-fill-color": "transparent",
```
darrenvong, line 15: Interesting that this has pretty good cross-browser support, and yet only available with the `webkit` prefix 🤔

darrenvong (changes requested):
Got a couple more questions on this, but otherwise this is generally looking great! 💅

app/web/features/dashboard/Hero/HeroSearch.tsx
guillermodlpa: good suggestion, thanks!

app/web/features/dashboard/Hero/HeroSearch.tsx
guillermodlpa: not really, but I don't think it should be hardcoded to a fixed value either. I'll remove the `todo` comment though as using `useId` isn't really necessary here because we don't expect to render this component more than once in a page. If that ever changes, the dev can make the change (once we're in v18)

app/web/features/dashboard/Hero/HeroLinks.tsx
guillermodlpa: Interesting! It can make sense yeah.

I think I prefer it to be the same width, so when hovering the other links, the bars that show are the same width. I personally prefer that, but I don't have strong feelings at all. Let me know if you'd prefer to change it to a percentage


https://user-images.githubusercontent.com/1557348/168542063-a35dc5f4-65af-4d34-a451-da1bf1d00b81.mov

app/web/utils/searchFilters.test.ts
guillermodlpa: ah sounds good, easy enough to add some tests to account for them! I'll do it and push a commit in a moment

app/web/features/dashboard/Hero/HeroButton.tsx
guillermodlpa: whoopsie, thanks!

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
guillermodlpa: I haven't measured performance, but `createTheme` internally does a few deep merges and so I generally call it outside the component render (for the main theme) or memoizing the result. https://github.com/mui/material-ui/blob/v4.12.4/packages/material-ui/src/styles/createTheme.js

app/web/features/dashboard/Hero/HeroLinks.tsx
darrenvong: Yeah actually that looks ok to me too, as it's always placed to the centre of the words at least with the `left: 50%`. I'm happy to leave it as is 👍

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
darrenvong: I see. The `theme` is actually exported in `theme.ts` (at the root of the `web` project). Might it make more sense to import that directly and create the override theme directly from that outside of the component/React lifecycle?

app/web/features/dashboard/Hero/HeroLinks.tsx
guillermodlpa: sounds good!

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
guillermodlpa: Yes, that's possible as well. It would be more performant, although it likely is too small to measure any difference. I prefer to leave it inside the React lifecycle so the `imageOverlayTheme` will get redone if `theme` got changed. I prefer to expect the theme to come from context than to be an import from this component that is somewhere inside the node tree, not wrapping everything at the top. For example, if we added later dark mode, we'd have to change a property on the `theme` being passed to the provider at the top of the tree. Let me know though if you prefer to make the theme outside, I can change it.

app/web/features/dashboard/Hero/HeroButton.tsx
lucaslcode: I don't feel strongly about it, so feel free to leave it as is, but to me we are indeed overriding MUI behaviour so the className targeting/ is appropriate (although I take your point about MUI using the theme colours in non-obvious ways).

Using `<ThemeProvider>` this way doesn't feel very theme-like - themes are a style that apply to the whole page or large sections, having a theme for a single button or something isn't what a theme is for. It would make sense if it was applied to the whole Hero section, but then obviously all the text would be white etc.

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
lucaslcode: I think it's important to do it from the context this way in case there is a `ThemeProvider` further down the tree

app/web/features/dashboard/Hero/HeroButton.tsx
lucaslcode: Actually, maybe not all the text be white? Is it possible to only have one `ThemeProvider` for the whole hero component?

darrenvong:
Looks like there are some type errors with the `SearchFilter` tests 🤔

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
darrenvong: Yeah good point about keeping it as a hook so React will correctly sync against `theme` changes! I did think about this but also since we don't support having multiple themes currently, we'd have to change how we pass the `theme` to the `ThemeProvider` later on anyway and I thought we could have just updated it to the hook then.

I suppose it probably doesn't make a big difference either way, so happy to leave it as is, as it's probably the more correct way to express that we want the theme to be reactive to changes.

app/web/features/dashboard/Hero/HeroButton.tsx
guillermodlpa: Right, I thought about that too originally. It's possible, yes. But then, I'd make another `ThemeProvider` inside for the search box, to set different theme styles for all the components within it. Either way works.

Screen Shot 2022-05-18 at 11 30 43

About nesting themes, if it helps I can point to the docs where it's referenced as a standard pattern for this type of scenario: https://mui.com/material-ui/customization/theming/#nesting-the-theme

> You can nest multiple theme providers. This can be really useful when dealing with different areas of your application that have distinct appearance from each other.

app/web/features/dashboard/Hero/useHeroBackgroundTheme.ts
guillermodlpa: sounds good, leaving it as-is then 👍 thank you!

guillermodlpa:
> Looks like there are some type errors with the `SearchFilter` tests 🤔

Whoops. Thanks, @darrenvong. I couldn't see what it was at first. Just pushed a fix!

lucaslcode (approved):

darrenvong (approved):

PR #2919, opened 2022-05-14
Add link to donations page on the "support" word in the dashboard

guillermodlpa:
As requested via Slack: https://couchersorg.slack.com/archives/C01FAQY1H8X/p1652460973764429

I made the decision to make these 2 links open in new tabs, since they feel a bit outside the application. But I'm 2nd guessing myself. Any thoughts on just handling them as next/link components?

Screen Shot 2022-05-14 at 12 53 51

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/features/dashboard/Dashboard.tsx
+++ b/app/web/features/dashboard/Dashboard.tsx
@@ -10,7 +10,7 @@ import PageTitle from "components/PageTitle";
 import DashboardBanners from "features/dashboard/DashboardBanners";
 import { Trans, useTranslation } from "i18n";
 import { DASHBOARD, GLOBAL } from "i18n/namespaces";
-import { blogRoute } from "routes";
+import { blogRoute, donationsRoute } from "routes";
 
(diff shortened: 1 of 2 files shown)


darrenvong:
> I made the decision to make these 2 links open in new tabs, since they feel a bit outside the application. But I'm 2nd guessing myself. Any thoughts on just handling them as next/link components?

I'd say technically since the whole site lives on the same domain (now that we merged the whole site under `couchers.org`, all of the pages we link to are part of the same application, so it should be an "internal link" and open in the same window/tab.

It's also the expected default behaviour and the most accessible option (especially for screenreader users), as users can choose to open in a new tab still if they wish. Since the dashboard doesn't contain any work in progress stuff where navigating away would result in some loss of work, I'd be more inclined to keep it opened in the same window/tab as opposed to a new one.

guillermodlpa:
Makes sense @darrenvong. Thanks for offering the detailed response. I'll update these links and follow that approach moving forward!

darrenvong (approved):

PR #2944, opened 2022-05-19
redirect on reference form page when steps are missing

### LINKED ISSUE #28: Reference settings not saved
Subject: Reference settings not saved
Description:
I was writing a reference. I completed the first two steps about safety and community standing. While writing a text in the third step, I closed the tab by accident. I re-opened the tab.

Results:
When I re-opened the tab, I could continue the third step. However, on reviewing the reference, I found that the first two steps were not saved and defaulted to a negative reference.

Either the first two steps should have been saved, or the reference should have started from step 1 again, or unideally the first two steps should have defaulted to a neutral reference.

Backend version: develop-d7782a98
Frontend version: develop-d7782a98
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36
Page: https://couchers.org/
User: stijn (4966)

wginsberg:
closes Couchers-org/web-frontend#28 

Just redirect to the first page in the reference form if any steps are missing (e.g. because the user refreshed the page).

We could still add an alert and/or persist the form data in session storage, but this change covers the case where all data is lost and user ends up on a page they shouldn't be at yet.

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/features/profile/view/leaveReference/ReferenceForm.tsx
+++ b/app/web/features/profile/view/leaveReference/ReferenceForm.tsx
@@ -88,11 +88,13 @@ export default function ReferenceForm({
 
   const isSkippedStep =
     referenceData.wasAppropriate === "" && step !== "appropriate";
+  const redirectTo =
+    referenceType === "friend"
+      ? `${leaveReferenceBaseRoute}/${referenceType}/${userId}`
+      : `${leaveReferenceBaseRoute}/${referenceType}/${userId}/${hostRequestId}`;
 
   return isSkippedStep ? (
-    <Redirect
-      to={`${leaveReferenceBaseRoute}/${referenceType}/${userId}/${hostRequestId}`}
-    />
+    <Redirect to={redirectTo} />
   ) : step === "appropriate" ? (
     <Appropriate
       referenceData={referenceData}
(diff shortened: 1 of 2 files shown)


app/web/features/profile/view/leaveReference/ReferenceForm.tsx
```diff
@@ -85,7 +86,14 @@ export default function ReferenceForm({
     }));
   };
 
-  return step === "appropriate" ? (
+  const isSkippedStep =
+    referenceData.wasAppropriate === "" && step !== "appropriate";
+
+  return isSkippedStep ? (
+    <Redirect
+      to={`${leaveReferenceBaseRoute}/${referenceType}/${userId}/${hostRequestId}`}
```
darrenvong: This needs an additional check on whether `hostRequestId` is defined is not for the final redirect URL. Otherwise, for friend references, I got `undefined` at the end of the URL (i.e. something like `/leave-reference/friend/4/undefined`)

Might be worth an additional test to check for this case too once fixed

darrenvong (changes requested):
Looks good except for the case when it's a friend reference

wginsberg:
Updated to work correctly for the friend reference form.

I realized that this branch has a warning in the tests. Not sure what to do about this since as far as I can tell there is no redirect to the same page.
```
  console.warn
    Prevented redirect to same page

       6 |   useEffect(() => {
       7 |     if (router.asPath === to) {
    >  8 |       console.warn("Prevented redirect to same page");
         |               ^
       9 |       return;
      10 |     }
      11 |     router.push(to);
      ```

wginsberg:
https://github.com/Couchers-org/web-frontend/pull/137

PR #3291, opened 2022-10-19
Implement a CPU + SQL query profiler

aapeliv:
**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/profiler.py
+++ b/app/backend/src/couchers/profiler.py
@@ -1,0 +1,56 @@
+import contextvars
+import io
+import logging
+import pstats
+from cProfile import Profile
+from time import perf_counter_ns
+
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/profiler.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class CouchersProfiler:
+    def __init__(self, do_profile=False):
+        self._do_profile = do_profile
+        do_profile_ctx.set(do_profile)
+        sql_queries.set([])
+
+    def __enter__(self):
+        self.report = None
```
shuttle1987, line 33: Is there some reason `self.report` is first set here and not in `__init__`?

shuttle1987 (dismissed):
Looks good

PR #3362, opened 2022-12-11
Email event cluster's subscribed users when an event is created

schradert:
Here is an incomplete attempt at sending emails as a background job when an event is created. An email would be sent to any user that is subscribed to the cluster the event was created in. I include some filters to prevent spamming, such as excluding events in the four largest clusters, only allowing events created by users with completed profiles to send emails, and excluding subscribers too far away when the number notified would exceed the maximum.

Currently, the background job test is failing when during the first event/cluster query in the job handler can't find a matching event. Even without the association to a cluster in the query, no event matches the payload ID. Could use some help here @ItsiW @aapeliv, I could make a codepairing work too for running through it and debugging. I suspect I'm just calling the event testing functions like `create_event` and `create_community` wrong, or in the wrong context. Similar tests are passing. Perhaps I should rewrite the test from basics if these aren't adaptable.

**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -130,6 +137,61 @@ def process_generate_message_notifications(payload):
...
+        "Europe, Africa, and Middle East": 10,
+    }
+    # don't send emails to users more than max_radius away from an event if there are more than max_users subscribed
+    max_users = 1000
+    max_radius = 20000
+
+    with session_scope() as session:
+        event, cluster = session.execute(
+            select(Event, Cluster)
+            .join(ClusterEventAssociation, Event.id == ClusterEventAssociation.event_id)
```
aapeliv: `ClusterEventAssociation` lets you have the same event related to many clusters (though that's not implemented right now), so this query should accommodate for that (or at least handle the case with a meaningful exception).

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -130,6 +137,61 @@ def process_generate_message_notifications(payload):
...
+    """
+    logger.info(f"Sending out emails for event_id = {payload.event_id}")
+
+    # don't send emails to global and regional events
+    excluded_cluster_parent_node_ids = {
+        "Global Community": 1,
+        "Americas": 4,
+        "Asia-Pacific": 7,
+        "Europe, Africa, and Middle East": 10,
+    }
```
aapeliv: These shouldn't be hardcoded here, we can add a db flag or something

aapeliv (commented):
I still worry about misuse of this, we have to be pretty careful. But we do need to get somethign implemented. One option would be to only send out notifications if the organizer is a community builder, that way there is some vetting.

Another thing: are we sure we want to send out the notification when the event is created, or should there just be a button on the event saying "send email" when the organizer has finished setting it up and is ready for the info to go out?

aapeliv:
@schradert could you rebase this onto `develop`, please, that way I could look at the failing test and see what's up with the event ids? (sorry, I looked at code before looking at what you wrote :) )

schradert:
> @schradert could you rebase this onto `develop`, please, that way I could look at the failing test and see what's up with the event ids? (sorry, I looked at code before looking at what you wrote :) )

Just rebased

schradert:
> Another thing: are we sure we want to send out the notification when the event is created, or should there just be a button on the event saying "send email" when the organizer has finished setting it up and is ready for the info to go out?

I like this idea, but it requires frontend work. Should we wait until that is implemented on the frontend? Will the frontend repo be merged into this one so that we can have the PR address this if we prefer that approach with no intermediate solution?

schradert:
> I still worry about misuse of this, we have to be pretty careful. But we do need to get something implemented. One option would be to only send out notifications if the organizer is a community builder, that way there is some vetting.

Okay I can work on this right now. I was told that there are many events not made by community builders. Is the idea we want to encourage them to become community builders in order to get this kind of event exposure?

app/backend/src/couchers/jobs/handlers.py
schradert: I think since it's not implemented like this already and this is only a stop-gap solution for event notifications before we create a proper notifications framework (with frontend controls users can tweak their notification settings with), I think it's probably sufficient to leave the `.one()` filter, no? If there are multiple matches because what you're suggesting is implemented before notifications, in my mind, we want to at least send it to a cluster instead of throwing an exception. Maybe I could loop through each cluster, apply all of this logic to each, and then only grab unique users so that they aren't notified twice? Not sure how to apply the `max_users` constraint in this situation either.

app/backend/src/couchers/jobs/handlers.py
schradert: I think the region IDs can definitely be removed since the logic around maximum notified users kinda handles these same constraints. The global ID makes sense for preventing people from being notified about virtual events, which I understand are always global. How can we include a db flag for that still?

app/backend/src/couchers/models.py
```diff
@@ -1850,8 +1850,8 @@ class BackgroundJobType(enum.Enum):
     generate_message_notifications = enum.auto()
     # payload: google.protobuf.Empty
     update_recommendation_scores = enum.auto()
-    # payload: google.protobuf.Empty
-    refresh_materialized_views = enum.auto()
```
aapeliv: Is this a merge issue?

app/docker-compose.yml
```diff
@@ -39,6 +39,6 @@ services:
     volumes:
       - "./data/media/uploads/:/uploads/"
     ports:
-      - 5000:5000
+      - 5001:5000
```
aapeliv: Please don't commit this :)

app/backend/templates/event_created.md
```diff
@@ -0,0 +1,11 @@
+---
+subject: "New couchers event in {{ event.owner_cluster.name|couchers_escape }}!"
+---
+
+{% from "macros.html" import button, link %}
+
+{{ user }}
+Stuff: {{ event.organizers, event.creator_user, event.owner_cluster.name, event.title }}
```
aapeliv: Needs to be fleshed out

aapeliv (commented):
This is looking pretty good. A few things:

* There really needs to be some way to unsubscribe from these; I suggest you make a simple boolean in the user model. I can probably try hacking this together.

* I had a look at the DB, here are the biggest communities:

```
 cluster_id |              name               | count 
------------+---------------------------------+-------
          2 | Global Community                | 29076
         11 | Europe, Africa, and Middle East | 18065
          5 | Americas                        |  8419
         43 | United States of America        |  5873
          8 | Asia-Pacific                    |  2906
        126 | Germany                         |  2422
         46 | Turkey                          |  2100
        570 | United Kingdom                  |  1673
        497 | France                          |  1400
        374 | Italy                           |  1327
         95 | Canada                          |  1225
         49 | Istanbul                        |  1003
        397 | Spain                           |   996
       1019 | Poland                          |   887
        734 | Netherlands                     |   797
         55 | Russia                          |   765
         14 | Australia                       |   670
         17 | New York City area              |   557
        153 | Berlin                          |   548
        180 | Brasil                          |   526
        888 | Belgium                         |   503
        491 | Portugal                        |   465
        500 | Paris                           |   461
        189 | Sweden                          |   397
```

I would say the best thing would be to send notifications only for events in leaf clusters; but the 1k user limit is OK too if it's too much effort.

* The email needs to be fleshed out and polished, and we need to double check users can't inject random links into it, etc!

app/docker-compose.yml
schradert: AirPlay uses 5000 so I had to change it for myself. I guess it makes more sense for me to just turn AirPlay off than to commit this and confuse people.

aapeliv:
Let me know by tagging here when you're ready for another review!

schradert:
> Let me know by tagging here when you're ready for another review!

@ItsiW @aapeliv I think the PR is ready for review now finally. Sorry it took so long! Not sure what's going on with the tests.

schradert:
@aapeliv I followed the instructions at `docs/database.md` to add a migration to the database for the field `User.send_event_notifications`. It seems the new migration file has a lot more than just that change, and when I try to spin up the docker network, it fails with `cannot drop table state_lookup because extension postgis_geocoder requires it`. The migration file does clearly include an `op.drop_table("state_lookup")` but I don't know why that's showing up. I don't know how to clean up the migration file properly. My thought would be to just include the line relevant to your unsubscribe feature and delete everything else but I don't know if that's a good idea.

schradert:
@aapeliv also I'm thinking we probably want the unsubscribe feature to be more granular. I imagine people want to unsubscribe for a more specific reason than wanting to never receive an email for a new event. My thought is it could be just for the event in question. Maybe in the future it could have a cluster-level control too. In this case though, my idea would be to also add this boolean flag you introduced to the `EventSubscription` and/or `ClusterSubscription` models.

aapeliv:
> @aapeliv I followed the instructions at `docs/database.md` to add a migration to the database for the field `User.send_event_notifications`. It seems the new migration file has a lot more than just that change, and when I try to spin up the docker network, it fails with `cannot drop table state_lookup because extension postgis_geocoder requires it`. The migration file does clearly include an `op.drop_table("state_lookup")` but I don't know why that's showing up. I don't know how to clean up the migration file properly. My thought would be to just include the line relevant to your unsubscribe feature and delete everything else but I don't know if that's a good idea.

It seems it picked up some postgis system tables. It has some functionality built in for US geo things, called "tiger" and another one, I think.

Yeah just try deleting all but the stuff you added. There should be a test to pick up differences too.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -131,6 +137,43 @@ def process_generate_message_notifications(payload):
...
+                logger.info(f"The Global Community is too big for email notifications.")
+            elif cluster.is_leaf or creator in cluster.admins:
+                subscribers.update(cluster.members)
+            else:
+                max_radius = 20000
+                nearby_subscribers = session.execute(
+                    select(User)
+                    .join(ClusterSubscription, ClusterSubscription.user_id == User.id)
+                    .where(ClusterSubscription.cluster_id == cluster.id)
+                    .where(func.ST_Contains(func.ST_Buffer(event.occurrences[0].geom, max_radius), User.geom))
```
aapeliv, line 165: I think `func.ST_DWithin(User.geom, event.occurrences[0].geom, max_radius / 111111)` is preferred, according to the web. These geography fields use WGS84 coordinates, so distances are measured in degrees; a degree is ~111km at the equator.

app/backend/templates/event_created.md
```diff
@@ -0,0 +1,17 @@
...
+---
+
+{% from "macros.html" import button, link %}
+
+Hi {{ user.name | couchers_escape }}!
+
+A new event was created in your area!
+
+{% set clusters = event.clusters %}
+{% set is_plural = event.clusters | length > 1 %}
```
aapeliv, line 12: Nice templating :D

aapeliv (commented):
Looks very good, just some minor comments.

aapeliv:
@schradert See if you can fix the tests, if not let me know and I can have a look

schradert:
@aapeliv should we close this now that you've reworked in elsewhere?

aapeliv:
> @aapeliv should we close this now that you've reworked in elsewhere?

Yes, let's do that!

PR #3907, opened 2023-12-30
Change issue links to web frontend

Roshanjossey:
As frontend is moved to another repo, updating links to issues to reflect that.

We could also consider removing these links to keep the contributing guide just about this repo

An argument against removing the links to frontend changes would be there would be contributors who might miss the first line in the Readme.


aapeliv (approved):

bakeiro:
@aapeliv, @Roshanjossey merged :)

PR #3993, opened 2024-03-05
return only users with a complete profile

### LINKED ISSUE #3978: [map] create a filter to return only users with a complete profile
When I search for a destination it is usually full of profiles that are empty, I as a user would like to be able to see only the profiles that have a photo/description.

I would like to make a toggle in the filter modal that makes the backend return only the profiles that have a complete profile.

### Where
image

### Reference
image

### Task
Make the backend to return only complete profiles when the parameter `only_complete_profiles` was sent by the frontned

Krishnag09:
closes #3978 - create a filter to return only users with a complete profile

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -1,6 +1,7 @@
 """
 See //docs/search.md for overview.
 """
+
 import grpc
 from sqlalchemy.sql import func, or_
 
@@ -293,9 +294,9 @@ def _search_clusters(
     return [
         search_pb2.Result(
             rank=rank,
-            community=community_to_pb(cluster.official_cluster_for_node, context)
-            if cluster.is_official_cluster
-            else None,
+            community=(
+                community_to_pb(cluster.official_cluster_for_node, context) if cluster.is_official_cluster else None
+            ),
             group=group_to_pb(cluster, context) if not cluster.is_official_cluster else None,
             snippet=snippet,
         )
@@ -461,7 +462,6 @@ class Search(search_pb2_grpc.SearchServicer):
             if request.only_with_references:
                 statement = statement.join(Reference, Reference.to_user_id == User.id)
 
-
             # TODO:
             # google.protobuf.StringValue language = 11;
             # bool friends_only = 13;
@@ -486,7 +486,7 @@ class Search(search_pb2_grpc.SearchServicer):
                     )
                     for user in users[:page_size]
                 ],
-                next_page_token=encrypt_page_token(str(users[-1].recommendation_score))
-                if len(users) > page_size
-                else None,
+                next_page_token=(
+                    encrypt_page_token(str(users[-1].recommendation_score)) if len(users) > page_size else None
+                ),
             )
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -458,6 +458,8 @@ class Search(search_pb2_grpc.SearchServicer):
 
             if request.only_with_references:
                 statement = statement.join(Reference, Reference.to_user_id == User.id)
+            if request.profile_completed:
+                statement = statement.where(User.has_completed_profile == True)
 
             # TODO:
             # google.protobuf.StringValue language = 11;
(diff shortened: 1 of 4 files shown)


bakeiro:
oh, that was simple, thanks a lot!!

bakeiro:
I'd say that I do the frontend part, and once we test works fine, we merge it ok?

bakeiro (dismissed):

app/proto/search.proto
```diff
@@ -133,6 +133,7 @@ message UserSearchReq {
   google.protobuf.BoolValue drinks_at_home = 25;
   google.protobuf.BoolValue parking = 26;
   google.protobuf.BoolValue camping_ok = 27;
+  google.protobuf.BoolValue profile_completed = 31;
```
aapeliv: I think this should be a `bool` like `query_name_only`. The `google.protobuf.BoolValue` allows you to communicate three states: true/false/not specified. In this case I doubt anyone wants to explicitly show profiles that are not complete, so it would suffice to have an "only show complete profiles" boolean.

app/backend/src/couchers/servicers/search.py
```diff
@@ -458,6 +458,8 @@ def UserSearch(self, request, context):
 
             if request.only_with_references:
                 statement = statement.join(Reference, Reference.to_user_id == User.id)
+            if request.profile_completed:
+                statement = statement.where(User.has_completed_profile == True)
```
aapeliv: I think this won't work since `profile_completed` is a `google.protobuf.BoolValue`. You'd have to do the following:

```py
            if request.HasField("profile_completed"):
                statement = statement.where(User.has_completed_profile == request.profile_completed.value)
```

aapeliv (changes requested):
Hi @Krishnag09: thank you for working on this and writing a PR! This is a great new feature and sorely needed.

I've added some comments; I think as the PR is now, it won't work due to the `bool/BoolValue` mixup. I don't mind whichever way you fix it (I think the two comments are complementary).

Finally, if you would be able to, it'd be great to have a test for this.

app/proto/search.proto
Krishnag09: but isn't that also applicable to parking, `camping ok`- nobody wants to see the not `camping ok `users. Hence i placed this under here.

aapeliv: I suppose it is. I don't think anyone ever wants to find people particularly without parking in the same way I doubt anyone wants to find people with empty profiles.

I'm fully happy for it to be whichever way you prefer. If you keep it as a `BoolValue`, then I think the python implementation needs the `HasField` bit in the other comment!

Krishnag09: it made sense to have it as such `bool` b let me know if it looks ok now.
also may be we should move a bunch of these fields to the same then.

aapeliv:
Tests now running at https://gitlab.com/couchers/couchers/-/pipelines/1209647621

Krishnag09:
> Tests now running at https://gitlab.com/couchers/couchers/-/pipelines/1209647621

seems i have broken them, sorry i have not been able to run them locally due to env issue..

aapeliv:
Could you change to pushing to the couchers github repo? I think you have access for that now?

app/backend/src/tests/test_search.py
```diff
@@ -73,3 +73,29 @@ def test_regression_search_in_area(db):
...
+    """
+    Make sure the completed profile flag returns only completed user profile
+    """
+
+    user_complete_profile, token6 = generate_user(about_me="this about me is not empty", avatar_key="test_avatar_key_1")
+
+    user_incomplete_profile, token7 = generate_user(about_me=None, avatar_key="test_avatar_key_2")
+
+    with search_session(token7) as api:
+        res = api.UserSearch(
```
aapeliv: It's complaining about formatting here. These `res = ` lines should be one-liners.

app/backend/src/tests/test_fixtures.py
```diff
@@ -220,6 +220,7 @@ def generate_user(*, delete_user=False, **kwargs):
         user_opts = {
             "username": username,
             "email": f"{username}@dev.couchers.org",
+            "avatar_key": "test",
```
aapeliv: `avatar_key` is a foreign key to an element of the images table (from memory). I think you can insert something bogus in there.

aapeliv:
@Krishnag09 this is a dupe of the other PR, right? If so, could you please close this?

PR #4001, opened 2024-03-12
Update readme.md

bakeiro:
documentation change, in order to improve clarity (for the dev experience)

### DIFF
--- a/app/readme.md
+++ b/app/readme.md
@@ -11,9 +11,6 @@ To **run the app locally**, you need to do **four things**:
 1. Get the code, navigate to the `app` folder and install `docker` and `docker-compose`
 2. Compile the protocol buffers
 3. Launch the backend with `docker-compose`
-4. Install and launch the web frontend with `yarn`
-
-Are you only developing on the web frontend? If you don't want to install docker, you can follow the alterative instructions in `web/readme.md`.
 
 ### Note for Windows users
 
@@ -67,17 +64,6 @@ docker-compose up --build
 ```
 
 This will build the backend, database, and proxy, and start them up.
-
-## 4. Install and launch the web frontend with `yarn`
-
-Navigate to the `app/web` folder, and run the following commands to start the web frontend:
-
-```sh
-yarn install
-yarn start
-```
-
-This will take a moment, after which you can go to <http://localhost:3000/>, and you should see the app there. The command might also open the page for you automatically.
 
 ## FAQ
 
--- a/app/readme.md
+++ b/app/readme.md
@@ -46,7 +46,7 @@ An alternative on **macOS** is to install the tools locally with `brew install g
 If you are on **Windows** (without WSL2), run the following command:
 
 ```sh
-docker run -it --pull always --rm -w /app -v ${PWD}/app:/app registry.gitlab.com/couchers/grpc sh -c "apt-get update && apt-get install dos2unix && cat generate_protos.sh | dos2unix | sh"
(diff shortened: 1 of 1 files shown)


app/readme.md
```diff
@@ -31,8 +28,6 @@ Open a command line or terminal, and navigate to the `app` folder (where this `r
 
 Install `docker` from <https://docs.docker.com/engine/install/>.
 
-Install `docker-compose` from <https://docs.docker.com/compose/install/>.
```
bakeiro, line 34: merged in docker since July 2023

app/readme.md
```diff
@@ -51,7 +46,7 @@ An alternative on **macOS** is to install the tools locally with `brew install g
 If you are on **Windows** (without WSL2), run the following command:
 
 ```sh
-docker run --pull always --rm -w /app -v %cd%:/app registry.gitlab.com/couchers/grpc sh -c "cat generate_protos.sh | dos2unix | sh"
+docker run --pull always --rm -w /app -v ${PWD}/app:/app registry.gitlab.com/couchers/grpc sh -c "apt-get update && apt-get install dos2unix && cat generate_protos.sh | dos2unix | sh"
```
bakeiro, line 49: solved command on windows

aapeliv: If `dos2unix` is not in the container, then I think we should add it there instead of installing it here

bakeiro: sure, thing is, I don't have access to it 🙃

app/readme.md
```diff
@@ -11,9 +11,6 @@ To **run the app locally**, you need to do **four things**:
 1. Get the code, navigate to the `app` folder and install `docker` and `docker-compose`
 2. Compile the protocol buffers
 3. Launch the backend with `docker-compose`
-4. Install and launch the web frontend with `yarn`
```
nciemniak, line 14: I would keep this step, but instead include a link to the front-end repo, e.g.:
Install and launch the [web frontend](https://github.com/Couchers-org/web-frontend)

app/readme.md
```diff
@@ -11,9 +11,6 @@ To **run the app locally**, you need to do **four things**:
...
-
-Are you only developing on the web frontend? If you don't want to install docker, you can follow the alterative instructions in `web/readme.md`.
```
nciemniak, line 16: I would keep this, but update it to say something like:
Are you only developing on the web frontend? You can run the frontend by itself, just follow instructions outlined in "add link here"

nciemniak (approved):
Looks good apart from the points I highlighted above 👍

app/readme.md
```diff
@@ -11,9 +11,6 @@ To **run the app locally**, you need to do **four things**:
 1. Get the code, navigate to the `app` folder and install `docker` and `docker-compose`
```
nciemniak, line 11: Let's update here as well to instruct them to only install `docker`

app/readme.md
bakeiro: right!

app/readme.md
aapeliv: It's here: https://github.com/Couchers-org/grpc

aapeliv: I've installed it there now, so no need to install here!

bakeiro: okey, thank you!!

aapeliv:
@bakeiro how is this going? Would you like to rebase it and re-request review?

bakeiro:
@aapeliv I think you can forget this PR... most of the changes were intended for updating the docs because the repositories were split, now that are merged again doesn't make sense, maybe the windows command but I will have a look once I finish the current task I'm on

aapeliv:
Okay, thanks! Sorry the work didn't get merged in.

PR #4002, opened 2024-03-12
filter out incomplete profiles

### LINKED ISSUE #3978: [map] create a filter to return only users with a complete profile

Krishnag09:
closes #3978 - create a filter to return only users with a complete profile

- [x] All tests pass

### DIFF
--- a/app/backend/src/tests/test_search.py
+++ b/app/backend/src/tests/test_search.py
@@ -1,12 +1,11 @@
 import pytest
-
-from couchers.utils import create_coordinate
-from couchers.db import session_scope
-from couchers.crypto import random_hex
-from proto import search_pb2
-from couchers.models import Upload, User
 from google.protobuf import wrappers_pb2
 
+from couchers.crypto import random_hex
+from couchers.db import session_scope
+from couchers.models import Upload
+from couchers.utils import create_coordinate
+from proto import search_pb2
 from tests.test_communities import testing_communities  # noqa
 from tests.test_fixtures import db, generate_user, search_session, testconfig  # noqa
 
--- a/app/backend/src/tests/test_search.py
+++ b/app/backend/src/tests/test_search.py
@@ -86,8 +86,8 @@ def test_user_filter_complete_profile(db):
 
     with search_session(token7) as api:
         res = api.UserSearch(search_pb2.UserSearchReq(profile_completed=True))
-    assert [result.user.user_id for result in res.results] == [user_complete_profile]
+        assert [result.user.user_id for result in res.results] == [user_complete_profile]
 
     with search_session(token7) as api:
         res = api.UserSearch(search_pb2.UserSearchReq(profile_completed=False))
-    assert [result.user.user_id for result in res.results] == [user_incomplete_profile]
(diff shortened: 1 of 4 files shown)


app/backend/src/tests/test_search.py
```diff
@@ -73,3 +77,48 @@ def test_regression_search_in_area(db):
...
+        )
+        session.commit()
+
+    user_complete_profile, token6 = generate_user(about_me="this profile is complete", avatar_key=key)
+
+    user_incomplete_profile, token7 = generate_user(about_me="", avatar_key=key2)
+
+    with search_session(token6) as api:
+        req = search_pb2.UserSearchReq()
+        req.profile_completed = True
```
aapeliv: I think you need to do something like this: `wrappers_pb2.BoolValue(value=True)`, instead of just `True`

Krishnag09: @aapeliv -its finally ready for review. i had to change the assert to accommodate the extra search user result from the uploader user.

app/backend/src/tests/test_search.py
```diff
@@ -51,11 +55,11 @@ def test_regression_search_in_area(db):
...
-    # outside
-    user1, token1 = generate_user(geom=create_coordinate(1, 0), geom_radius=100)
-    # outside
-    user2, token2 = generate_user(geom=create_coordinate(0, 1), geom_radius=100)
-    # inside
+    # # outside
+    # user1, token1 = generate_user(geom=create_coordinate(1, 0), geom_radius=100)
+    # # outside
+    # user2, token2 = generate_user(geom=create_coordinate(0, 1), geom_radius=100)
+    # # inside
```
aapeliv: Why are these commented out?

app/backend/src/tests/test_search.py
```diff
@@ -73,3 +77,50 @@ def test_regression_search_in_area(db):
...
+        key = random_hex(32)
+        filename = random_hex(32) + ".jpg"
+        session.add(
+            Upload(
+                key=key,
+                filename=filename,
+                creator_user_id=uploader_user.id,
+            )
+        )
+        session.commit()
```
aapeliv: This commit is not needed when exiting the context manager, it'll commit automatically

aapeliv (dismissed):
Looks good! Thank you!

aapeliv:
@Krishnag09 did you have a chance to look at the comments?

Krishnag09:
> @Krishnag09 did you have a chance to look at the comments?

yes just pushed changes.

aapeliv:
Great, will have a look

aapeliv (approved):
Looks great to me! Thanks a lot!

aapeliv:
Also please squash these when you merge!

aapeliv:
It’s live now on prod!

PR #4030, opened 2024-04-07
Simplify background jobs

aapeliv:
This simplifies the background jobs system: now to create a job you just need to add it to `handlers.py`. The old system was way complicated, you had to mess with the database, etc; for really no great reason (other than making sure typos don't make jobs not run, but that ought to be caught in tests).

**Backend checklist**
- [x] All tests pass


aapeliv:
@Krishnag09 could you please review when you get a chance?

Krishnag09 (approved):
looks good. but if you have someone to give a third set of eyes.

PR #4044, opened 2024-04-11
Reworked notifications, notify for events

aapeliv:
**Backend checklist**
- [x] All tests pass


aapeliv:
This only applies to those with new notifications enabled, so it should be fine to merge.

aapeliv:
These are the defaults:

Screenshot from 2024-05-03 21-22-17

aapeliv:
Critical parts left:

* test for event notifs
* adjust digest frequency/disable it

PR #4050, opened 2024-04-12
Add path to download descriptors.pb from envoy proxy

aapeliv:
This allows one to go download the protos in use by envoy by going to https://api.couchers.org/descriptors.pb, which allows me to build a tool that will get the latest protos from the API it's hitting.

**Backend checklist**
- [ ] Formatted my code by running `autoflake -r -i --remove-all-unused-imports src && isort . && black .` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [ ] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

PR #4052, opened 2024-04-13
Implement Status and GetDescriptors methods

aapeliv:
This implements a status endpoint that checks that the database is working. I'll use it to set up status monitoring given the recent issues.

Also added the descriptors.pb thing from #4050 (this time in the backend). This allows one to go download the protos in use by envoy by going to https://api.couchers.org/descriptors.pb, which allows me to build a tool that will get the latest protos from the API it's hitting.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/descriptor_pool.py
+++ b/app/backend/src/couchers/descriptor_pool.py
@@ -5,6 +5,12 @@ from google.protobuf import descriptor_pb2, descriptor_pool
 
 
 @functools.lru_cache
+def get_descriptors_pb():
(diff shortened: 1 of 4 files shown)


developerHobday:
A bit late but LGTM!
Is this being called by a regular health check?  Some ides are wiring it to a slack channel when the health check fails, and having a health page.

aapeliv:
Thanks @developerHobday! Yes, wiring it up to slack would be awesome.

For now I've set it up on my personal infra so it texts me when it's down, but setting it up so it alerts everyone and having that reviewed would be even better!

PR #4053, opened 2024-04-13
Fix up donations

aapeliv:
Stripe changed their API: they no longer create a "payment intent" when you initiate a one time donation. This means that one time and recurring donations are now roughly the same, and so I've merged the tables and combined the logic. We also don't really do anything with these at the moment (except send a thank you email with an invoice).

**Backend checklist**
- [x] All tests pass


aapeliv:
@developerHobday: could you have a look at this please? I think the non-test code is ready. I just need to do a painstaking update of the tests!

aapeliv:
Fully ready for review now @developerHobday !

developerHobday:
It looks really great, I have just one question - is the stripe_subscription_id for RecurringDonation not being used?  Because it seems like we are losing this column.
Also, could you link to the change in the Stripe documentation?

schradert:
> is the stripe_subscription_id for RecurringDonation not being used? Because it seems like we are losing this column.

`stripe_subscription_id` [appears unused](https://github.com/search?q=repo%3ACouchers-org%2Fcouchers%20stripe_subscription_id&type=code)

schradert:
The comment on [Line 90](https://github.com/Couchers-org/couchers/pull/4053/files#diff-f90f763c00367d0eb38414943dde09e3303c7c846fa2bcb9dba5c5092c3eec6dR90) is outdated if there's no more `payment_intent` in `event["data"]`

schradert:
> Also, could you link to the change in the Stripe documentation?

Yeah some kind of changelog link would be helpful to contextualize, but I also don't know where to look to confirm the fields on this event object are what we expect them to be. I tried to review changelogs and reference docs for the stripe web api and python client but didn't find any obvious mention of this.

aapeliv:
@developerHobday: I will add it to the commit as well, but it's here: https://docs.stripe.com/upgrades#2022-08-01: "A PaymentIntent is no longer created during Checkout Session creation in payment mode. Instead, a PaymentIntent will be created when the Session is confirmed."



> The comment on [Line 90](https://github.com/Couchers-org/couchers/pull/4053/files#diff-f90f763c00367d0eb38414943dde09e3303c7c846fa2bcb9dba5c5092c3eec6dR90) is outdated if there's no more `payment_intent` in `event["data"]`

@schradert: Thanks, I'll fix that up!

aapeliv:
Honestly it's kind of annoying how much stripe is changing stuff. I thought their main point was to abstract the payments so that one doesn't have to worry about the idiosyncrasies of different types of payments!

aapeliv:
It ended up breaking things anyway.

PR #4057, opened 2024-04-14
Use a simpler query in enforce_community_memberships_for_user

aapeliv:
For whatever reason the old code was creating a ton of objects, which was significantly slowing down signup, hopefully this works faster.

**Backend checklist**
- [ ] All tests pass


aapeliv:
Could you do a quick review of this please @schradert?

PR #4068, opened 2024-04-15
Fixes donations for latest Stripe API version.

aapeliv:
Stripe changed their API: they no longer create a "payment intent" when you initiate a one time donation. This means that one time and recurring donations are now roughly the same, and so I've merged the tables and combined the logic. We also don't really do anything with these at the moment (except send a thank you email with an invoice).

In particular, they no longer create a payment intent when you create a checkout session, as per this changelog:

https://docs.stripe.com/upgrades#2022-08-01

**Backend checklist**
- [ ] All tests pass


aapeliv:
The migration didn't work on the last one, so here we are again.

PR #4076, opened 2024-04-18
Tweaks to auth flows

### LINKED ISSUE #4058: Password reset is way too complicated
As we push for some new action, people will be resetting their passwords en-masse. We should fix this to make the experience smoother.

### LINKED ISSUE #4059: Increase session expiry from 90 days

aapeliv:
* Sends a user a reminder to finish signing up if they try to log in with a username/email that hasn't completed signup
* Adds a new password reset endpoint that actually resets password, instead of just clearing it and making the user do it on the next screen, closes #4058
* Extends session expiry, closes #4059 
* Passes the real IP address through nginx to the backend

**Backend checklist**
- [x] All tests pass


aapeliv:
@developerHobday @Krishnag09 @schradert  : could one of you please review when you have a few minutes?

developerHobday (approved):
Looks good to me

PR #4247, opened 2024-04-23
Implement badges in the frontend

aapeliv:
Yet another dodgy frontend PR.

Screenshot from 2024-04-23 00-07-34

**Web frontend checklist**
- [ ] All tests pass


app/web/features/profile/view/Badges.tsx
```diff
@@ -0,0 +1,46 @@
...
+  badge: {
+    marginInlineStart: theme.spacing(1),
+    marginBottom: theme.spacing(1),
+  },
+}));
+
+export const Badges = ({ user }: Props) => {
+  const classes = useStyles();
+  const { badges } = useBadges();
+  console.log(user);
```
bakeiro: console.log trace 👀

app/web/service/resources.ts
```diff
@@ -20,3 +20,8 @@ export async function getLanguages() {
   const languages = await client.resources.getLanguages(new Empty());
   return languages.toObject();
 }
+
+export async function getBadges() {
```
bakeiro, line 24: this was already implemented by the backend??

bakeiro (commented):
.

app/web/features/profile/view/Badges.tsx
aapeliv: Oop, gotta fix that. I was having issues with the protos not updating somehow and with the docker containers, so I resorted to stone age debugging :P

app/web/service/resources.ts
aapeliv: No, that's from #4246 that this one builds on!

aapeliv:
The tests fail on this one because of proto stuff. It's very easy to fix, I'll do it in a bit.

app/web/features/profile/view/Badges.tsx
```diff
@@ -0,0 +1,45 @@
...
+    marginInlineStart: theme.spacing(1),
+    marginBottom: theme.spacing(1),
+  },
+}));
+
+export const Badges = ({ user }: Props) => {
+  const classes = useStyles();
+  const { badges } = useBadges();
+
+  if (badges === undefined) {
```
darrenvong: This should probably also check for the `badgesList` being empty so it renders nothing (the fragment, `<></>`, or better with `null`)? As at the moment, it would return an empty div if user has no badges (mostly "ok" but may mess up styling sometimes unexpected further down the line)

darrenvong:
Unrelated but damn, it's nice to be able to build the protos from source again rather than relying on that zip!

aapeliv:
> Unrelated but damn, it's nice to be able to build the protos from source again rather than relying on that zip!

Yes, but I think it's kind of a good idea though for new devs. I'll update the dockerfile I think to download the zip for develop from the build server in case there's no locally generated protos. I think that's a good compromise!

app/web/features/profile/view/Badges.tsx
aapeliv: Is the PR otherwise roughly good?

darrenvong: Yeah the rest looks fine to me!

app/web/service/resources.ts
bakeiro: gocha!

PR #4249, opened 2024-04-23
Add superuser and token/api key creation to dummy data

aapeliv:
Cut down version of #2610

**Backend checklist**
- [ ] All tests pass


aapeliv:
Closed #1731

PR #4251, opened 2024-04-23
Implement remember me and default to true

### LINKED ISSUE #4106: Show both login and password at login form
It's very-very annoying when I cannot use my auto-fill in browser because "password" field appears only after one click to "Continue". Thus it causes few extra clicks every time.

image

Please show both textboxes at the same time. If there is any kind of SSO support, then better *hide* password textbox when one enters SSO-related email.

In combination with non-working "Remember me" (#210) it creates a lot of inconvenience

### LINKED ISSUE #4107: "Remember me" doesn't work (or too short?)
It seems like "Remember me" doesn't really work as expected. Even if I use it, next time I see login form again. Not sure, maybe there is some expiration time for cookies which is too short (I login to Couchers somewhere between weekly-monthly)

Please check it and make sure it actually works and remembers the user for some meaningful amount of time (I'd suggest one year)

### LINKED ISSUE #4108: Enable "Remember me" by default
I think it would be beneficial for all users to have "remember me" checkbox set to "Enabled" all the time, this should be the most used functionality.

Currently it requires extra click during every login which is inconvenient

### LINKED ISSUE #4175: Remember me switch is just for decoration
IMO we should either:

- get rid of it completely; or

- Connect it properly, turn it on by default, change the backend behaviour so that when it is not selected, the session expires after the session

aapeliv:
Closes #4107, closes #4108, closes #4175, closes #4106

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/auth/login/LoginForm.tsx
+++ b/app/web/features/auth/login/LoginForm.tsx
@@ -10,7 +10,7 @@ import { useTranslation } from "i18n";
 import { AUTH, GLOBAL } from "i18n/namespaces";
 import { LoginRes } from "proto/auth_pb";
 import { useState } from "react";
-import { useForm } from "react-hook-form";
+import { Controller, useForm } from "react-hook-form";
 import { resetPasswordRoute } from "routes";
 import { service } from "service";
 import isGrpcError from "utils/isGrpcError";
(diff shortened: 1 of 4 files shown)


aapeliv:
This cannot be merged before the password reset flow is fixed in the frontend! There is currently no way to reset password if it's like this since there's no way to do magic link login!

app/web/features/auth/login/LoginForm.tsx
```diff
@@ -87,63 +74,59 @@ export default function LoginForm() {
...
-            control={<Switch size="small" />}
-            label={t("auth:login_page.form.remember_me")}
+          <Controller
+            control={control}
+            name="rememberDevice"
+            defaultValue={true}
+            render={({ onChange, value }) => (
+              <FormControlLabel
+                className={classes.rememberSwitch}
+                value="rememberDevice"
```
darrenvong:
```suggestion
```
Would this still work if it's left out? It's a bit confusing to have a value of `rememberDevice`when it's supposed to be the field's name

aapeliv: Let's see!

aapeliv:
This PR is back to being all about "remember me", other stuff moved to #4254

PR #4254, opened 2024-04-24
Frontend auth tweaks

### LINKED ISSUE #4106: Show both login and password at login form

aapeliv:
Closes #4106 

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


aapeliv:
Folded into #4427

PR #4259, opened 2024-04-24
Docs site

aapeliv:
Build with `sphinx-build -M html . _build`


aapeliv:
Canning this for now, based on discussions:

1. It's good for dev docs; but
2. Bad for support/help page: it just looks like a wall of text right now
3. Also "docs." is unintuitive for non-devs

PR #4272, opened 2024-04-29
Add timeout to fix map zoom animation; Remove redundant animation calls

### LINKED ISSUE #4155: On search page change "Keyword" to "Keyword or name"
To make it more clear, the radio button on the search page should be labeled "Keyword or name" instead of just "Keyword" since some users are still confused on how to find users by name.

spreeni:
Addresses part of #4113: Fixes the bug in the location search where the panning animation gets interrupted.
- Added Timeout to complete the animation
- Realized there were multiple redundant calls to the pan animation for each search, so removed the redundant ones

Steps to recreate:
1. search for a location (panning animation should work)
2. don't clear the search box
3. search for another location (panning animation gets interrupted)

or

1. search for a location via the Filters pop-up (panning animation gets interrupted)

Closes https://github.com/Couchers-org/couchers/issues/4155

**Web frontend checklist**
- [ ] ~Added any new components to storybook~
- [ ] ~Added tests where relevant~
- [x] All tests pass (locally 3 tests fail due to timezone issues, but in Gitlab they pass)


bakeiro:
oh, good job! I was about doing a refactor of the map part because code was a bit confusing... thanks for your change!

bakeiro:
if the behavior was tested, code looks good to me :)

bakeiro (dismissed):
LGTM (please test it in desktop and mobile before merging)

spreeni:
@bakeiro Thanks for the review! Actually forgot that I can test different screen sizes easily with Chrome developer tools (not a frontend dev). I noticed that the field label was a bit long on mobile so I shortened it a bit.

image

Also merged the current state of develop into it.

Please have a quick look again, then I would go ahead and merge.

app/web/features/search/SearchPage.tsx
```diff
@@ -83,15 +83,9 @@ export default function SearchPage() {
       showResults.current = searchFilters.any;
       setTimeout(() => {
         map.current?.resize();
-        updateMapBoundingBox(searchFilters.active.bbox);
```
bakeiro, line 86: here was deleted... sure it don't affect the behavior?

bakeiro:
I don't see any animation or anything noticeable in my local set up (with your branch's code) but the code looks good and I feel it more stable (maybe was deleting that redundant call?)

bakeiro (approved):

spreeni:
I noticed that with the real API and real data in production there is still an interrupted animation, once the search results for a location seem to come back from the API. I will try to fix this race condition again.

bakeiro:
> I noticed that with the real API and real data in production there is still an interrupted animation, once the search results for a location seem to come back from the API. I will try to fix this race condition again.

@spreeni 
that's what I mean, I didn't notice any particular change on my environment... only an apparent improvement in the stability...

spreeni:
@bakeiro I could reproduce the bug locally if I increase the backend latency artificially. I have not found an easy way to fix this in a clean way as my React knowledge is limited.

A hacky solution could be to remove the calls to `updateMapBoundingBox`, then the app waits for the results first and then executes the animation. This does introduce a small initial stutter, depending on the wait time for the backend search response, but fixes the interrupted animation.

The interrupt happens in SearchResultsList->map.current?.stop(); - but unfortunately that stop seems to be necessary to draw the user pins onto the map. Ideally, the app would wait for the end of the animation and then draw the user pins.

bakeiro:
I see...

IMHO we could invest efforts to create a good event page or a community page (or fixing/improving the map search, which has a lot of bugs) instead of putting of efforts on these animations... anyway please talk to me on Slack so maybe we can talk about this and design a good solution, or at least clarify this issue :)

PR #4321, opened 2024-05-16
AddAdmin/RemoveAdmin calls in Communities service

### LINKED ISSUE #4295: Add API calls to add/remove community builders to/from communities
Community builder means an `admin` on `Cluster` for the one that is `is_official_cluster` for a `Node`... `ListAdmins` should give a better idea.

The call should include the user (name/email/id, similar to rest of admin), and the community id.

schradert:
Currently imports some functions from `admin.proto` and `admin.py` servicer because they were originally implemented there and they are expected to return user details rather than a different response payload

Closes #4295 

**Backend checklist**
- [x] All tests pass


schradert:
Question for @aapeliv: how can we make sure that only admins of an ancestor node (or even just world admin) can use these endpoints?

schradert:
Currently having annoying issues with local podman I need to debug tomorrow in order to run the database properly and confirm tests work on my side

app/proto/communities.proto
```diff
@@ -30,6 +30,14 @@ service Communities {
...
   }
 
+  rpc AddAdmin(AddAdminReq) returns (AddAdminRes) {
+    // Makes a user a community builder
+  }
+
+  rpc RemoveAdmin(RemoveAdminReq) returns (RemoveAdminRes) {
+    // Removes user as community builder
+  }
+
```
aapeliv, line 49: I think maybe better to retun `google.protobuf.Empty`?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -180,6 +180,43 @@ def ListAdmins(self, request, context):
                 next_page_token=str(admins[-1].id) if len(admins) > page_size else None,
             )
 
+    def AddAdmin(self, request, context):
+        with session_scope() as session:
+            user = session.execute(select(User).where_username_or_email_or_id(request.user)).scalar_one_or_none()
+            if not user:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
```
aapeliv: I think better to pass in the `user_id` and do something like

```py
session.execute(select(User).where_users_visible(context).where(User.id == request.user_id)).scalar_one_or_none()
```

(The `where_[...]_email` function is only for admins, otherwise emails are private).

app/backend/src/couchers/servicers/communities.py
```diff
@@ -180,6 +180,43 @@ def ListAdmins(self, request, context):
                 next_page_token=str(admins[-1].id) if len(admins) > page_size else None,
             )
 
+    def AddAdmin(self, request, context):
```
aapeliv, line 183: I think you should do the permission check using `can_moderate_node(session, context.user_id, node.id)` (where `node.id == request.community_id`, but you probably need to grab the node first.

aapeliv (changes requested):
Looks good, please see the comments!

schradert:
@aapeliv why do you think this error is showing up in the tests:

```
    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "cluster_subscriptions" violates foreign key constraint "fk_cluster_subscriptions_cluster_id_clusters"
E       DETAIL:  Key (cluster_id)=(4) is not present in table "clusters".
E       
E       [SQL: INSERT INTO cluster_subscriptions (user_id, cluster_id, role) VALUES (%(user_id)s, %(cluster_id)s, %(role)s) RETURNING cluster_subscriptions.id]
E       [parameters: {'user_id': 3, 'cluster_id': 4, 'role': 'member'}]
E       (Background on this error at: https://sqlalche.me/e/20/gkpj)
```

It was able to get ID 4 in the first place on the line `c1_id = get_community_id(session, "Country 1")`, so I'm confused why it doesn't exist now...

app/backend/src/couchers/servicers/communities.py
```diff
@@ -180,6 +180,59 @@ def ListAdmins(self, request, context):
...
+                context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.NODE_MODERATE_PERMISSION_DENIED)
+
+            user = session.execute(
+                select(User).where_users_visible(context).where(User.id == request.user_id)
+            ).scalar_one_or_none()
+            if not user:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
+
+            subscription = user.cluster_subscriptions.where(
+                ClusterSubscription.cluster_id == request.community_id
```
aapeliv: Cluster id and node id are different. I think here you need `node.official_cluster.id`

aapeliv:
> @aapeliv why do you think this error is showing up in the tests:

See that comment I just added. If that wasn't the issue then please quote the line number/function where this is raised.

app/backend/src/couchers/servicers/communities.py
schradert: Oh I see, I wasn't paying close enough attention to the fact that `community_id` is a node id. I should've noticed that when just adding the `node` logic I found in the other functions.

app/backend/src/couchers/errors.py
```diff
@@ -96,6 +96,7 @@
 NO_PASSWORD = "That account has no password."
 NO_PENDING_VERIFICATION = "No phone number to verify."
 NO_RECIPIENTS = "No recipients."
+NODE_MODERATE_PERMISSION_DENIED = "You're not allowed to moderate that node"
```
aapeliv: Maybe replace node with community in user facing error?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -194,9 +194,9 @@ def AddAdmin(self, request, context):
             if not user:
                 context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
 
-            subscription = user.cluster_subscriptions.where(
-                ClusterSubscription.cluster_id == node.official_cluster.id
-            ).one_or_none()
+            subscription = next(
+                (sub for sub in user.cluster_subscriptions if sub.cluster_id == node.official_cluster.id), None
+            )
```
aapeliv: Sorry. I think better to replace with something like
```
            subscription = session.execute(
                select(ClusterSubscription)
                .where(ClusterSubscription.user_id == user.id)
                .where(ClusterSubscription.cluster_id == node.official_cluster.id)
            ).scalar_one_or_none()
```

app/proto/communities.proto
```diff
@@ -151,6 +168,16 @@ message ListAdminsRes {
   string next_page_token = 2;
 }
 
+message AddAdminReq {
+  uint32 user_id = 1;
```
aapeliv: `user_id` should be `int64`

aapeliv (changes requested):
Just the three small comments. Sorry about the `.where` issue with `cluster_subscriptions`, not sure why that didn't work!

app/proto/communities.proto
schradert: Sorry, I see that in UserDetails now

aapeliv (dismissed):
Looks good to me now, so when whatever checks are failing are sorted it’s good to go!

app/backend/src/tests/test_communities.py
```diff
@@ -563,6 +563,72 @@ def test_ListAdmins(testing_communities):
...
+
+        with communities_session(token8) as api:
+            with pytest.raises(grpc.RpcError) as err:
+                api.AddAdmin(communities_pb2.AddAdminReq(community_id=node_id, user_id=user2_id))
+            assert err.value.code() == grpc.StatusCode.FAILED_PRECONDITION
+            assert err.value.details() == errors.NODE_MODERATE_PERMISSION_DENIED
+
+        with communities_session(token4) as api:
+            res = api.ListAdmins(communities_pb2.ListAdminsReq(community_id=node_id))
+            assert res.admin_user_ids == [user4_id, user5_id]
```
schradert, line 618: The test is only failing here in CI because it expects user 2 to be an admin already, but within the context of `test_RemoveAdmin` this doesn't make any sense. When this test is run in isolation, it succeeds because the preceding attempt to add user 2 AS user 8 fails as expected. Could it be failing because of incomplete cleanup from `test_AddAdmin` where the final list of admin ids does include user 2?

aapeliv: Yes, I think maybe the communities don’t clean up the whole thing but reuse between tests.

Maybe you can make/use another community to make sure they are separate?

schradert: I just added a line of cleanup to both

aapeliv: So I don't remember the specifics of why the communities tests do one set up and one cleanup. Not sure if it was very slow or something. Sorry about that.

aapeliv (approved):

PR #4326, opened 2024-05-19
AddAdminNote API for user moderation

### LINKED ISSUE #4296: Add moderator notes for users
The admins/moderators need to keep track of why people have been banned, or sometimes other private notes.

Proposed idea:

* Add a text field to `models.User`, something like `admin_note`
* Add `admin_note` field to `org.couchers.admin.Admin.BanUser`
* Add an API call `org.couchers.admin.Admin.AddAdminNote`
* Return the admin note in `org.couchers.admin.Admin.UserDetails`

On adding a note, it should include a timestamp and the admin's id+username, and append to the note. Alternatively add a new table of `AdminNotes` or something and store it there instead of hacking around with concatenated strings. (Though I think concatenating strings is quite enough.)

schradert:
Closes #4296 

This implementation treats a new field on User (admin_note) as an append-only log that prefixes plaintext with a timestamp and the id and username of the superadmin that logged it.

This feature is intended to detail any relevant notes for moderation, like the reason behind banning users. Because of this, we tie it to the BanUser endpoint as well.

**Backend checklist**
- [x] All tests pass


schradert:
@aapeliv I had the same issue I encountered implementing the event notifications feature last year where the generated migration included hundreds of lines unrelated to the minor change I created. I deleted those lines from the migration so that it is as simple and relevant as possible, but just wanted to flag that again.

aapeliv (approved):
This looks great!!

jesseallhands:
Thanks for doing this!!! 🙏 It's a big help for the safety and trust team!

PR #4327, opened 2024-05-19
feat: 4079 password reset

### LINKED ISSUE #4079: Update frontend for new password reset flow
In Couchers-org/couchers#4076, I introduce a new API call for resetting the password in a better way. The frontend should be updated to make use of this new functionality.

## Background

* Initially we had email tokens instead of passwords: you could sign up without ever setting a password, and every time you logged in, it would just send you a token by email
* We have a "jail" system, where a user gets locked out unless they fix up something in their profile/account. We've used this e.g. for when we update ToS or similar where users have to accept before they can continue using it.

The current password reset works like this:

1. You click "reset password", web calls `ResetPassword`
2. Backend sends the right email a link to reset password
3. User clicks on link, takes you to the web app which calls `CompletePasswordReset` with **nothing but the token**
4. The backend clears the users' password
5. The user goes to log in again, this time it falls back to password-less login.
6. We email the user a login link
7. User clicks on link and logs in
8. User gets jailed
9. User has to reset password through `Jail/SetPassword`.
10. User finally gets in

Now with the new system we reduce it to 3 steps:

1. You click "reset password", web calls `ResetPassword`
2. Backend sends the right email a link to reset password
3. User clicks on link, takes you to the web app which calls `CompletePasswordResetV2` with the **reset token and a new password**
4. Backend sets the user's password, **and** logs the user in

bakeiro:
Closes #4079 

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [x] All tests pass


app/backend/src/couchers/urls.py
```diff
@@ -36,7 +36,7 @@ def account_settings_link():
 
 
 def password_reset_link(*, password_reset_token):
-    return f"{config['BASE_URL']}/complete-password-reset?token={password_reset_token}"
+    return f"{config['BASE_URL']}/set-new-password?token={password_reset_token}"
```
bakeiro: before we reset the password and then put a new one later on... now we put a new one directly, that's why I think the name change gives it more clarity, besides the change is different from how it was before.

open to comments

app/web/features/auth/deletion/ConfirmDeleteAccount.tsx
```diff
@@ -28,7 +28,7 @@ export default function ConfirmDeleteAccount() {
   >(
     async ({ token }) => {
       if (token === undefined) {
-        throw Error(t("auth:delete_account.missing_token"));
+        throw Error(t("auth:delete_account.request.missing_token"));
```
bakeiro: just some missing token that made the CI complain

app/web/features/auth/jail/Jail.tsx
```diff
@@ -80,12 +78,9 @@ export default function Jail() {
...
-      {jailInfo?.hasNotSetPassword && (
-        <PasswordSection
-          updateJailed={updateJailed}
-          className={classes.section}
-        />
-      )}
     </>
   );
 }
+{
```
bakeiro: We are not using the jail anymore to change the password, since now the use can change the password without even being logged in...

So this page it´s outside the jail system, now this gives me the doubt whether it´s possible that a user end up in a jail system with the attribute `hasNotSetPassword` by accident, I guess this scenario isn´t possible? (just wanted to double check)

app/web/features/auth/password/SetNewPassword.tsx
```diff
@@ -0,0 +1,155 @@
...
+    if (newPassword !== newPasswordCheck) {
+      alert(t("change_password_form.password_mismatch_error"));
+      return;
+    }
+
+    mutate(newPassword);
+  });
+
+  // TODO: needed?
+  if (authState.authenticated) {
```
bakeiro: don´t let the user change it´s password if he is logged in to avoid weird scenarios/edge cases?

app/backend/src/couchers/urls.py
aapeliv: "Password reset" is the usual term used in this context, so we should stick to that.

app/web/features/auth/jail/Jail.tsx
aapeliv: We shouldn't remove this yet. It's possible for the user to end up here by: not having a password now, then trying to log in. The user would get a login link and end up in jail in the next step.

app/web/features/auth/locales/en.json
```diff
@@ -16,11 +16,13 @@
   },
   "change_password_form": {
     "title": "Change password",
+    "subtitle": "Please change password",
```
aapeliv: "Use this form to reset your login password."

app/web/features/auth/locales/en.json
```diff
@@ -16,11 +16,13 @@
...
     "confirm_password": "Confirm password",
-    "password_changed_success": "Your new password has been set and a notification was sent to your email.",
-    "old_password": "Old password",
```
aapeliv: These are used in `ChangePassword.tsx`? (When you change your password while logged in.)

aapeliv (changes requested):

app/web/features/auth/locales/en.json
```diff
@@ -16,11 +16,13 @@
...
     "new_password": "New password",
-    "password_mismatch_error": "The passwords don't match."
+    "password_mismatch_error": "The passwords don't match.",
+    "token_error": "The're was a error processing the token from the email, please report the issue.",
```
aapeliv: "There was a error processing the password reset token, please try opening the link from the email again."

app/web/features/auth/locales/en.json
```diff
@@ -16,11 +16,13 @@
...
+    "reset_password_success": "Your password has been changed successfully!.",
```
aapeliv: Has an extra dot

app/web/features/auth/jail/Jail.tsx
bakeiro: I guess it´s only for the people who already reset its password before this change?

app/web/features/auth/locales/en.json
bakeiro: implemented ✅

app/web/features/auth/locales/en.json
bakeiro: changed ✅

app/web/features/auth/locales/en.json
bakeiro: updated ✅

app/web/features/auth/jail/Jail.tsx
bakeiro: the idea to delete this it's to delete dead code (to avoid code which isn't used)

also, if they don't have a password just clicking in `Forgot password` would solved it (if I'm not wrong), so in that scenario, they never reach this part?

app/web/features/auth/locales/en.json
bakeiro: ups, reverted!

bakeiro (commented):
comments

app/backend/src/couchers/urls.py
bakeiro: the thing is, for the sake of clarity I would set a new name, because there is already a password reset page which is only for putting the email, (image attached)

a different name could help to differentiate 

image

bakeiro: also we avoid using old references to the previous code

bakeiro: still if isn't enough I can revert it to the previous name

.gitignore
```diff
@@ -1,11 +1,14 @@
...
 
 # helps Lennart configure his vim
 pyrightconfig.json
 
-.DS_Store
-._.DS_Store
 /private/
+
+# TODO: temp
+app/web-frontend/proto/
```
bakeiro: I listen your opinions about this

aapeliv: I don't think there should be anything in the folder `app/web-frontend`?

app/web/features/auth/jail/Jail.tsx
aapeliv: It can still leave people in limbo though (they try to log in and get a login link, then it's not obvious why they are jailed and there's no feedback). Let's wait until we clean up the backend to remove password optionality?

app/web/features/auth/password/CompleteResetPassword.tsx
```diff
@@ -1,58 +1,151 @@
...
+
+    mutate(newPassword);
+  });
+
+  // TODO: needed?
+  if (authState.authenticated) {
+    return (
+      <Container className={formClass.standardContainer}>
+        <Alert severity="error">
+          Can't changed the password if you are logged in
```
aapeliv: This should be a translation string, right? Also add "You can change your password in account settings, or log out to reset your password." or similar

bakeiro: true!, I forgot to change it....

bakeiro: ah I remember I put a comment because I was unsure if I should keep this code snippet or not, I think github deleted the comment when I updated the file

bakeiro: image

bakeiro: note for me:

tell the user to log out
don't allow to change the password?

.gitignore
bakeiro: yap, tons of files 😅 (for real)

bakeiro: image

aapeliv: You can delete the whole folder.

While the repos were separate, the suggested way to work on both was to clone web-frontend into that folder. But now that’s not a thing anymore, so you can delete that whole folder.

app/web/features/auth/locales/en.json
```diff
@@ -16,12 +16,17 @@
...
+    "subtitle": "Use this form to reset your login password.",
     "confirm_password": "Confirm password",
     "password_changed_success": "Your new password has been set and a notification was sent to your email.",
-    "old_password": "Old password",
     "new_password": "New password",
-    "password_mismatch_error": "The passwords don't match."
+    "old_password": "Old password",
+    "password_mismatch_error": "The passwords don't match.",
+    "token_error": "There was a error processing the password reset token, please try opening the link from the email again.",
+    "reset_password_success": "Your password has been changed successfully!",
```
aapeliv, line 27: Could we reuse the string from below as they’re identical?

app/web/features/auth/locales/en.json
```diff
@@ -223,6 +225,7 @@
       "error_message": "Failed to recover account: {{message}}",
       "button_text": "Recover your account",
       "success": "Your account has been recovered, you can log back in. We're excited to have you back!"
-    }
+    },
+    "missing_token": "Error: missing token"
   }
 }
```
aapeliv: Actually, could you look through the diff in this file and try to remove extra changes please?

app/web/features/auth/password/CompleteResetPassword.tsx
bakeiro: done ✅

bakeiro:
@darrenvong any review? :)

app/web/features/auth/password/CompleteResetPassword.tsx
```diff
@@ -1,58 +1,152 @@
...
+
+    mutate(newPassword);
+  });
+
+  if (authState.authenticated) {
+    return (
+      <Container className={formClass.standardContainer}>
+        <Alert severity="error">
+          Can't changed the password if you are logged in
+        </Alert>
```
aapeliv, line 81: Please change this to a translation string, "You cannot reset your password if you are logged in. You can change your password in settings or log out to reset it."

app/web/features/auth/password/CompleteResetPassword.stories.tsx
```diff
@@ -1,31 +1,22 @@
-import { Meta, Story } from "@storybook/react";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
-import React from "react";
+import { Meta, Story } from "@storybook/react";
 import { mockedService } from "stories/serviceMocks";
+import React from "react";
 
-import CompleteResetPassword from "./CompleteResetPassword";
+import SetNewPassword from "./CompleteResetPassword";
```
aapeliv:
```suggestion
import CompleteResetPassword from "./CompleteResetPassword";
```

app/web/features/auth/password/CompleteResetPassword.stories.tsx
```diff
@@ -1,31 +1,22 @@
...
 
 export default {
-  component: CompleteResetPassword,
-  title: "Me/Auth/CompleteResetPasswordPage",
+  component: SetNewPassword,
+  title: "Me/Auth/SetNewPasswordPage",
```
aapeliv: Should also be `CompleterResetPassword`

aapeliv (changes requested):

app/web/features/auth/password/CompleteResetPassword.tsx
bakeiro: ou, sure! don't know how I uploaded the code like this... 🤦

app/web/features/auth/password/CompleteResetPassword.stories.tsx
bakeiro: done! thanks for the feedback!

app/web/features/auth/password/CompleteResetPassword.stories.tsx
bakeiro: other components have the `Page` suffix at the end... I guess that `CompleteResetPasswordPage` it's ok? (since the component name it's `CompleteResetPassword`

app/web/features/auth/password/CompleteResetPassword.tsx
bakeiro: done! ✅

app/web/features/auth/password/CompleteResetPassword.tsx
```diff
@@ -1,58 +1,152 @@
...
+import { useTranslation } from "i18n";
+import Alert from "components/Alert";
 import { loginRoute } from "routes";
+import { RpcError } from "grpc-web";
 import { service } from "service";
-import stringOrFirstString from "utils/stringOrFirstString";
+
+const useStyles = makeStyles((theme) => ({
+  form: {
+    "& > * + *": {
```
pktwum96, line 22: Selects all elements that are children of all elements that are a direct descendant of root? This selector is a bit hard to read. Also some of the components being selected are from MUI which we do not have direct access to (or control of ) their HTML structure and thus selecting it this way may be problematic

app/web/features/auth/password/CompleteResetPassword.tsx
```diff
@@ -1,58 +1,152 @@
...
+    string
+  >((newPassword) =>
+    service.account.CompletePasswordResetV2(resetToken as string, newPassword)
   );
 
-  useEffect(() => {
-    if (resetToken) {
-      completePasswordReset(resetToken);
+  const onSubmit = handleSubmit(({ newPassword, newPasswordCheck }) => {
+    if (newPassword !== newPasswordCheck) {
```
pktwum96, line 68: Should this check be done on Submit? Usually the comparison is done after the second password is typed. Is this a decision we have decided to make to only compare passwords onSubmit?

app/web/features/auth/password/CompleteResetPassword.tsx
```diff
@@ -1,58 +1,152 @@
...
+      <Typography variant="body1" gutterBottom>
+        {t("auth:change_password_form.subtitle")}
+      </Typography>
+
+      <form className={formClass.form} onSubmit={onSubmit}>
+        <TextField
+          className={formClass.textField}
+          id="newPassword"
+          inputRef={register({ required: true })}
+          label={t("auth:change_password_form.new_password")}
```
pktwum96, line 126: Do we do any checks on validity of password eg length, strength etc

darrenvong (commented):
> @darrenvong any review? :)

Haven't properly tested, but nothing too suspicious from a cursory look of the code - Aapeli probably have caught most things I may have mentioned! 😄 Nice to see the tests being updated as well

aapeliv (approved):
LGTM

pktwum96:
Left a few comments on this one. Looks like they were missed

aapeliv:
> Left a few comments on this one. Looks like they were missed

I can't see them on the timeline on this PR. Where did you comment them?

pktwum96:
> > Left a few comments on this one. Looks like they were missed
> 
> I can't see them on the timeline on this PR. Where did you comment them?

Its in here for some reason
Hidden items

aapeliv:
I only see some other activity behind those hidden items:

Screenshot from 2024-06-09 21-28-51
Screenshot from 2024-06-09 21-28-45

Are you sure you submitted the review and comments?

Could you please try re-posting the comments otherwise. Maybe something odd is going on with github!

bakeiro:
image

@aapeliv @pktwum96 I think this is the issue

aapeliv:
Ahh, I see. I think the “Submitting your review” section here has a good explanation on how to submit it: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#submitting-your-review

pktwum96:
Massive mess up on my end. Apologies to everyone. I am more used to Gitlab 😆

PR #4382, opened 2024-05-27
Add UI for Strong Verification Confirmation

### LINKED ISSUE #4348: Strong Verification indicators on profile page
The backend now returns SV info:

https://github.com/Couchers-org/couchers/blob/e7a6ee7882602db6b5f150a5061eb102589f04dd/app/proto/api.proto#L274-L276

This should be displayed on the profile page. If verification is VERIFIED, it should be a green tick next to age/gender, and if MISMATCH, then red cross. Maybe a tooltip.

Also we should have a text somewhere that says "This user has verified their identity using a biometric passport." if `has_strong_verification` is true. It would also be cute to give their picture a halo or something to put them apart from other users and encourage SV.

pktwum96:
closes #4348

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [x] All tests pass


pktwum96:
Hi @aapeliv ,

I have a few questions about the implementation of this feature:

1. Although we have separate age and gender verification, in the UI, both are currently displayed in a single field. 
age

Would you prefer we:
a. Separate them into two fields, each with its own inline tick mark for confirmation? This would also require separating them in the [locales and translations](https://github.com/Couchers-org/couchers/blob/develop/app\web\features\profile\locales\en.json).
b. Add an inline tick mark for each verification? This would involve changing the type for the [LabelAndText ](https://github.com/Couchers-org/couchers/blob/develop/app/web/components/LabelAndText.tsx) component to accept not just text but also React nodes for this purpose. 
c. Displaying a single verified icon if both age and gender are verified (this is the current implementation I have gone with).

2. The ticket mentions, "and if MISMATCH, then red cross." Are we supposed to display the red cross if genderVerificationStatus and birthdateVerificationStatus return false? Would this mean that if a user has not attempted to verify this information, those keys would return undefined? I am concerned about displaying a red cross by default on a user's profile unless they have attempted verification and there was a mismatch.

3. The ticket suggests, "It would also be cute to give their picture a halo or something." Should this "halo or something" appear only on their profile picture when viewed on their personal page, or should it be displayed everywhere their profile picture appears?

aapeliv:
> a. Separate them into two fields, each with its own inline tick mark for confirmation? This would also require separating them in the [locales and translations](https://github.com/Couchers-org/couchers/blob/develop/app%5Cweb%5Cfeatures%5Cprofile%5Clocales%5Cen.json).

Yes, I think it'd be better to it as "39 (tick) / Man (tick)"



> 2\. The ticket mentions, "and if MISMATCH, then red cross." Are we supposed to display the red cross if genderVerificationStatus and birthdateVerificationStatus return false? Would this mean that if a user has not attempted to verify this information, those keys would return undefined?

Yes, I was thinking of displaying a cross if the API returns a mismatch. For those who haven't verified, the API will return UNVERIFIED. The API will only return mismatch if the user tried to verify their passport and the data did not match.

So in short; I think the icon should be:

1. If `VERIFIED`: then a tick with tooltip "This user has verified their [age/gender] using a biometric passport."
2. If `MISMATCH`: then a cross with tooltip "This user's [age/gender] does not match their passport."
3. If `UNVERIFIED`: then a question mark (or similar?) with tooltip: "This user has not yet verified their [age/gender] using a biometric passport."




> 3\. The ticket suggests, "It would also be cute to give their picture a halo or something." Should this "halo or something" appear only on their profile picture when viewed on their personal page, or should it be displayed everywhere their profile picture appears?

I was thinking everywhere. This should use the `has_strong_verification` boolean from the API.

aapeliv:
I really like the little shield next to the name, that's great!

pktwum96:
Updated as per feedback. This is what it looks like now with pronouns as well. 

Verification with pronouns

If we are happy with this implementation I will add in the tests and work on the halo for the strong verification.

aapeliv:
This looks great to me! The blue tick is a bit hard to see, maybe there like a filled bubble with a tick or similar?

pktwum96:
@aapeliv PR ready for  review. I am adding the halo into a separate MR as I anticipate it will likely raise more conversation around its design

aapeliv:
This looks great to me!

@bakeiro: would you be able to review this when you get a chance please?

app/web/features/profile/locales/en.json
```diff
@@ -1,195 +1,200 @@
 {
```
aapeliv: Could you please reformat this file to 4 spaces per indent?

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -92,6 +98,17 @@ export default function UserOverview({
       <div className={classes.wrapper}>
         <Typography variant="h1" className={classes.intro}>
           {user.name}
+          {user.hasStrongVerification ? (
```
bakeiro, line 101: has this function been implemented in the serveside?

app/web/features/profile/view/About.test.tsx
```diff
@@ -36,6 +36,57 @@ describe("About (user)", () => {
...
+
+    const checkIcon = screen.queryByTestId("check-circle-icon");
+    expect(checkIcon).toBeNull();
+    const errorIcon = screen.queryByTestId("error-icon");
+    expect(errorIcon).toBeNull();
+  });
+  it("should display age and gender without verification icons when unverified", async () => {
+    renderAbout({
+      ...defaultUser,
+      genderVerificationStatus: 1,
```
bakeiro, line 54: can we use strings as values instead of plain integers? (just to make it more readable :) )

bakeiro: genderVerificationStatus: "unverified" | "verified" | "otherstate"

same for birthdateVerificationStatus :)

bakeiro: same for birthdateVerificationStatus :)

bakeiro:
image
not an UI/UX guy, but, makes sense to put only a blue check icon (like instagram's or twitter's verified badge) when that info it's verified? (only as a suggestion :) )

bakeiro:
everything looks good! thanks for the job, if this was properly properly and works without issues then it's ready to go for me

I wrote some small comments in the PR, they are optional to implement, (but IMHO better if they are ;) )

bakeiro (dismissed):
LGTM

app/web/features/profile/view/UserOverview.tsx
pktwum96: Yes it has.

app/web/features/profile/view/About.test.tsx
pktwum96: This is an api change but i agree with this as well. It was a bit confusing for me to understand what the numbers meant until i went through the typings. I see we use this number system for  most of the other keys as well. 

@aapeliv  to opine

aapeliv: They are enums. They should be exported in the generated protobuf code so you can use those instead of numbers in here, but I guess they map to ints in the end.

bakeiro (approved):
forgot to approve, LGTM :)

aapeliv:
This is live now on the staging instance at https://next.couchershq.org!

PR #4391, opened 2024-05-28
Remove sentry URL

schradert:
Don't want this in git :) We'll rotate the keys later


app/backend.dev.env
```diff
@@ -58,4 +58,4 @@ LISTMONK_API_KEY=...
 LISTMONK_LIST_UUID=...
 
 SENTRY_ENABLED=1
```
aapeliv: Please also set this to 0 to disable sentry fully

aapeliv (approved):

PR #4410, opened 2024-06-04
Re-enable vercel

aapeliv:


app/web/ci-build-protos.sh
```diff
@@ -0,0 +1,26 @@
+# download deps
```
bakeiro, line 1: so, just some doubts...

this gets the protos right? then it unzips it so they are ready to use, this is the only thing (apart of upgrading the node version) that was needed to enable vercel? (just out of curiosity)

another doubts, this file is called `ci-build-protos`, shouldn't this file being called somewhere? (like I don't see any reference to this file, I guess this is part of the Vercel magic?)

bakeiro: in case that both things are a yes, then I'll approve this PR right away

aapeliv: It gets the tools to build the protos, then builds them. It doesn't download the protos themselves, they are in `//app/proto`.

`ci-build-protos.sh` is set up in vercel config.

Part of the reason why there was so little to do was that this was already set up before and was just disabled!

PR #4428, opened 2024-06-09
Switch to ruff / Add pre-commit hook

spreeni:
I switched from black/isort/autoflake to ruff, which is much faster, combines these tools and also is a strong linter. Additionally, I set up a `pre-commit` hook that can optionally be used to enforce QA-checks before each commit. I personally find it very satisfying to work with this, as it often corrects/spots smaller errors.

I exclude stricter linting rules that might limit our development style, but applied multiple other linting rules which I think are sensible.

ruff linting modules that I normally use but omitted for now are pydocstyle ("D") and mccabe complexity ("C90"), as some functionality would need to be refactored to reduce complexity in our case and enforcing doc strings everywhere might also be quite restrictive.

**Backend checklist**
- [x] All tests pass


app/.gitlab-ci.yml
```diff
@@ -266,9 +266,9 @@ test:backend-format:
     default: false
   script:
     - cd app/backend
-    - autoflake --exclude src/proto -r -i --remove-all-unused-imports --check src
-    - isort --check --diff .
-    - black --check --diff .
+    - ruff check .
+    - ruff check --diff .
+    - ruff format --diff .
```
aapeliv, line 271: Is there a way to combine these three into one command?

app/backend/pyproject.toml
```diff
@@ -1,12 +1,29 @@
...
+    "E711",   # Comparison to `None` should be `cond is None`
+    "E712",   # Avoid equality comparisons to `True`; use `if approved:` for truth checks
+    "F811",   # Redefinition of unused variable
+    "F841",   # Local variable is assigned to but never used
+    "UP015",  # Unnecessary open mode parameters
+]
+
+[tool.ruff.lint.isort]
+known-first-party = ["couchers", "proto", "tests", "dummy_data"]
+extra-standard-library = ["zoneinfo"]
```
aapeliv, line 27: Hmm, this should be in 3.12?

app/backend/.pre-commit-config.yaml
```diff
@@ -0,0 +1,34 @@
+default_language_version:
```
aapeliv, line 1: I don't like pre-commit hooks personally, but do not mind it being in the repo for others if it's not automatic. Is this optional?

app/backend/src/couchers/migrations/versions/bf12729fa8eb_sms_verification_constraints.py
```diff
@@ -24,4 +24,4 @@ def upgrade():
 
 
 def downgrade():
-    raise hell
+    raise hell  # noqa: F821
```
aapeliv: I mean we can change this to a real exception, just `raise Exception("hell")`

app/backend/src/couchers/servicers/account.py
```diff
@@ -77,11 +77,11 @@ def _check_password(user, field_name, request, context):
...
-    out = dict(
-        birthdate_verification_status=api_pb2.BIRTHDATE_VERIFICATION_STATUS_UNVERIFIED,
-        gender_verification_status=api_pb2.GENDER_VERIFICATION_STATUS_UNVERIFIED,
-        has_strong_verification=False,
-    )
+    out = {
+        "birthdate_verification_status": api_pb2.BIRTHDATE_VERIFICATION_STATUS_UNVERIFIED,
+        "gender_verification_status": api_pb2.GENDER_VERIFICATION_STATUS_UNVERIFIED,
+        "has_strong_verification": False,
+    }
```
aapeliv: Could you noqa this one please if you don't have a strong opinion otherwise?

app/backend/src/couchers/servicers/search.py
```diff
@@ -42,10 +42,13 @@ def _join_with_space(coalesces):
     return out
 
 
-def _build_tsv(A, B=[], C=[], D=[]):
+def _build_tsv(A, B=None, C=None, D=None):
     """
     Given lists for A, B, C, and D, builds a tsvector from them.
     """
+    B = B if B else []
```
aapeliv: Why not `B=B or []` here too?

app/backend/src/tests/test_fixtures.py
```diff
@@ -919,7 +919,7 @@ def __getattr__(self, attr):
             try:
                 return self.kwargs[attr]
             except KeyError:
-                raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{attr}'")
+                raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{attr}'") from None
```
aapeliv, line 922: I'm not sure what this expression does?

app/backend/src/tests/test_interceptors.py
```diff
@@ -110,7 +110,7 @@ def test_logging_interceptor_all_ignored():
         message = random_hex()
 
         def TestRpc(request, context):
-            context.abort(status_code, message)
+            context.abort(status_code, message)  # noqa: B023
```
aapeliv, line 113: Online I see: 'B023 	[function-uses-loop-variable](https://docs.astral.sh/ruff/rules/function-uses-loop-variable/) 	Function definition does not bind loop variable {name}"

This seems to be the incorrect noqa?

app/backend/src/tests/test_interceptors.py
```diff
@@ -223,7 +223,7 @@ def TestRpc(request, context):
         request_type=auth_pb2.SignupAccount,
         response_type=auth_pb2.AuthReq,
     ) as call_rpc:
-        with pytest.raises(Exception):
+        with pytest.raises(Exception, match="Some error message"):
```
aapeliv, line 226: What does this line do?

app/backend/pyproject.toml
```diff
@@ -1,12 +1,29 @@
...
+]
+ignore = [
+    "B007",   # Loop control variable not used within loop body
+    "E501",   # Line too long
+    "E711",   # Comparison to `None` should be `cond is None`
+    "E712",   # Avoid equality comparisons to `True`; use `if approved:` for truth checks
+    "F811",   # Redefinition of unused variable
+    "F841",   # Local variable is assigned to but never used
+    "UP015",  # Unnecessary open mode parameters
+]
```
aapeliv: Is this still an accurate list? These linters rules seem useful, are we failing a lot of them?

aapeliv (commented):
Neat! Definitely caught a few subtle bugs/issues here, thanks for introducing the linter! I even learned some more python syntax I wasn't even aware of!

Personally I don't mind the empty f-strings, and I doubt they really affect performance much. But I guess good to get rid of either way.

.github/pull_request_template.md
```diff
@@ -10,7 +10,7 @@ Checklists - you can remove one that is not applicable (ie. remove backend check
 If you need help with any of these, please ask :)
 --->
 **Backend checklist**
-- [ ] Formatted my code by running `autoflake --exclude src/proto -r -i --remove-all-unused-imports src && isort . && black .` in `app/backend`
+- [ ] Formatted my code by running `ruff check --select I --fix . && ruff check . && ruff format .` in `app/backend`
```
aapeliv, line 13: Also this seems different from the other commands run elsewhere

app/backend/src/couchers/migrations/versions/bf12729fa8eb_sms_verification_constraints.py
aapeliv: Or even something nicer :)

app/backend/pyproject.toml
spreeni: Yes, it is. But I have noticed that the ruff isort module is not always up-to-date with these, so you sometimes need to declare a standard-library like this.

app/backend/.pre-commit-config.yaml
spreeni: Yes, this only is applied if you run `pre-commit install` at some times, otherwise it won't do anything. I was wondering if I put it into the root of the repository as it will run for the whole repository in general, but then again it is fairly specific to python and the backend code. What do you think?

app/backend/src/tests/test_interceptors.py
spreeni: It checks that the exception message matches the given regex. This is to prevent the linting rule B017 https://docs.astral.sh/ruff/rules/assert-raises-exception/

app/backend/src/tests/test_fixtures.py
spreeni: To have a traceback of exceptions you can write `raise Exception from other_exc`. B904 states that you should be explicit if you don't want that https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

app/.gitlab-ci.yml
spreeni: It does not seem to be available yet: https://github.com/astral-sh/ruff/issues/8232

I feel this is fine, although `ruff check .` and `ruff check --diff .` will have overlap. The first will show all lint rules being broken while the second will show the proposed diffs for all lint violations that can be fixed automatically. But let me know if you have a differing opinion on this.

.github/pull_request_template.md
spreeni: Yes, here I decided that we will always want to do import sorting automatically, the rest is similar. Just that it does not propose the diff - do you think that would be helpful? I find it a bit too verbose for my taste. In a CI pipeline I feel it is better to be more verbose.

app/backend/src/couchers/migrations/versions/bf12729fa8eb_sms_verification_constraints.py
spreeni: I guess you're right - did not want to break Christian's joke :) Changed it now

app/backend/src/couchers/servicers/search.py
spreeni: Oops, missed one from what I had previously.

app/backend/src/couchers/servicers/account.py
spreeni: Yeah, tbh I have also been on the fence with this rule as I feel it sacrifices readibility for minuscule performance gain. I just ignore C408 now. 👍

app/backend/pyproject.toml
spreeni: So after looking at the code these are checks that I excluded for the following reasons, correct me if I am wrong:
- B007: Sometimes we might want to have the variable unpacking to make clear what is being iterated over, no? As in `for topic, name, items in group:`, even if we just use topic/items?
- E501: Would need quite a bit broken up or marked as noqa, as it also considers comments and long strings
- E711: We use None to compare to sqlalchemy results - Correct me if I'm wrong, but I think these are not actually `None`, but just falsy?
- E712: Same as above, just for `True`
- F811: This currently applies to all fixture usages. We could fix this by not importing fixtures but having them in a `conftest.py` file
- F841: This I assumed often refers to things that are in a variable that might be valuable later on, so I did not want to remove them all
- UP015: Personal opinion, but I like to be excplicit about the open-mode

For each rule, you can comment out the ignore-line and check which lines would be affected to see for yourself as well. I think for `F841` especially, you will be able to judge better what to exclude and what to leave.

app/backend/src/tests/test_interceptors.py
spreeni: No, this is the correct noqa - it complains because we define a class with loop iterables that will mutate in the next iteration. Here, it does not matter as everything happens within that iteration and then the class is defined again, so I added the noqa tag.

app/.gitlab-ci.yml
aapeliv: Yes that sounds good. It would be good if it showed all the errors then fails the build if there are any.

app/backend/.pre-commit-config.yaml
aapeliv: OK sounds good. Would you like to add some docs on how to enable ti?

app/backend/src/tests/test_fixtures.py
aapeliv: Thanks!

app/backend/src/tests/test_interceptors.py
aapeliv: Okay, makes sense!

.github/pull_request_template.md
aapeliv: No this seems good!

app/.gitlab-ci.yml
spreeni: Yes, that's exactly how it works 👍

app/backend/.pre-commit-config.yaml
spreeni: I added a paragraph to the contributing Readme, have a look if it is self-explanatory :)

spreeni:
These are current files that are over 350kb - I guess most of these are fine, but maybe good to have an eye on it.

```
app/web/yarn.lock (658 KB) exceeds 350 KB.
docs/full_er_diagram.png (1585 KB) exceeds 350 KB.
app/web/features/dashboard/Hero/mesut-kaya-eOcyhe5-9sQ-unsplash.jpeg (3825 KB) exceeds 350 KB.
app/web/public/img/blog/20210407_lucas.jpg (484 KB) exceeds 350 KB.
app/web/public/img/blog/20210825_talkingtown.png (8739 KB) exceeds 350 KB.
app/backend/src/data/osm/80500.geojson (2582 KB) exceeds 350 KB.
app/web/public/img/blog/20210501_virtual_events.jpg (551 KB) exceeds 350 KB.
app/mobile/src/yarn.lock (747 KB) exceeds 350 KB.
app/backend/src/data/osm/62149.geojson (3206 KB) exceeds 350 KB.
app/backend/src/data/osm/61320.geojson (490 KB) exceeds 350 KB.
docs/current_er_diagram.png (1352 KB) exceeds 350 KB.
app/web/public/img/blog/20210604_mindy.jpg (439 KB) exceeds 350 KB.
app/backend/src/data/osm/54224.geojson (1003 KB) exceeds 350 KB.
app/backend/src/data/osm/148838.geojson (11182 KB) exceeds 350 KB.
```

spreeni:
At some point, we could also integrate `clang-format` and `eslint` into the hook. Then a new person could just code away and would always have fully linted code when it came to the CI/CD pipeline.

This formatted the files, but always fails with an error "No such file or directory", so I left it out for now.

```
-   repo: https://github.com/pre-commit/mirrors-clang-format
    rev: 'v18.1.6'
    hooks:
      - id: clang-format
        args: ['--style=file:app/proto/.clang-format', '-i', '*.proto']
        files: ^.*.proto
```

aapeliv (dismissed):
I think you addressed all my questions. Looks good to me now!

spreeni:
@aapeliv I merged develop into the branch, ready to be merged once you approve it

aapeliv (approved):
Thanks, please merge!

aapeliv:
I really like this... it's so friggin fast xD

app/backend/pyproject.toml
HansBambel: To `B007`: You can put an underscore in front of the variable that is not required, but named. E.g: `for topic, _name, items in group:`.

spreeni: @HansBambel Just saw that you commented, great to see you here 😄

PR #4504, opened 2024-06-30
Fix issue 4335

### LINKED ISSUE #4335: Admin.UnbanUser
We currently have org.couchers.admin.Admin.BanUser but if an error is made, there is no way to undo this action other than manually updating the database.

We also need org.couchers.admin.Admin.UnbanUser to reverse a ban in the unlikely event of an error.

lovesqualor:
I think this closes #4335? Not sure if there are more changes I need to make. 

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/web/.env.development
```diff
@@ -1,9 +1,12 @@
-NEXT_PUBLIC_COUCHERS_ENV=preview
```
aapeliv: You shouldn't commit these changes

app/backend/src/couchers/servicers/admin.py
```diff
@@ -159,6 +159,14 @@ def BanUser(self, request, context):
...
             user.is_banned = True
         return self.AddAdminNote(request, context)
+    
+    def UnbanUser(self, request, context):
+        with session_scope() as session:
+            user = session.execute(select(User).where_username_or_email_or_id(request.user)).scalar_one_or_none()
+            if not user:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
+            user.is_banned = False
+        return self.AddAdminNote(request, context)
```
aapeliv, line 169: The formatter is complaining, you should run
```
ruff check --select I --fix . && ruff check . && ruff format .
```

aapeliv (changes requested):
Looking good, see the few small notes.

aapeliv:
> I think this closes #1234?

That's the wrong issue it seems

lovesqualor:
OK i fixed them

aapeliv (changes requested):
Some tests are still failing. Please re-run the formatter.

PR #4545, opened 2024-07-13
Create an anonymous, randomized, public map layer

aapeliv:


nabramow (approved):
Tested the regular map still works okay! Will test this endpoint when I make the new map component.

We discussed and will leave the profile visibility out of the frontend for now until we discuss the details.

PR #4560, opened 2024-07-15
WIP: Frontend auth check

aapeliv:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/Dockerfile
+++ b/app/web/Dockerfile
@@ -16,9 +16,12 @@ RUN yarn install --frozen-lockfile
 
 COPY . .
 
-ENV PROTOS_SHA=6b89068325adbee2e413a9553afe13d2e8d5174a7e1703e8ed7d48c7f55fbdcf
+# ENV PROTOS_SHA=6b89068325adbee2e413a9553afe13d2e8d5174a7e1703e8ed7d48c7f55fbdcf
+# ENV PROTOS_URL=https://protos--by-sha.preview.couchershq.org/$PROTOS_SHA/ts.tar.gz
 
-RUN wget https://protos--by-sha.preview.couchershq.org/$PROTOS_SHA/ts.tar.gz
+ENV PROTOS_URL=https://develop--protos.preview.couchershq.org/ts.tar.gz
+
+RUN wget $PROTOS_URL
 RUN tar xf ts.tar.gz && rm -f typescript_protos.tar.gz
 
 RUN cp .env.$environment env && \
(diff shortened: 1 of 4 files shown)


aapeliv:
@bakeiro

app/web/Dockerfile
```diff
@@ -16,9 +16,12 @@ RUN yarn install --frozen-lockfile
...
 COPY . .
 
-ENV PROTOS_SHA=63b57d8c12b291f8f5c1fd740879530e7da7865797f783dafcda7b3432c2efba
+# ENV PROTOS_SHA=b084f43d1becbe18a256819907a9beeb2008966d47eb9f7786aac49c6c381706
+# ENV PROTOS_URL=https://protos--by-sha.preview.couchershq.org/$PROTOS_SHA/ts.tar.gz
 
-RUN wget https://protos--by-sha.preview.couchershq.org/$PROTOS_SHA/ts.tar.gz
+ENV PROTOS_URL=https://develop--protos.preview.couchershq.org/ts.tar.gz
+
+RUN wget $PROTOS_URL
```
bakeiro, line 24: why this? (just to understand it)

app/web/features/auth/useAuthStore.ts
```diff
@@ -161,12 +161,50 @@ export default function useAuthStore() {
...
+        setLoading(false);
+      },
     }),
     //note: there should be no dependenices on the state or t, or
     //some useEffects will break. Eg. the token login in Login.tsx
     [setAuthenticated, setJailed, setUserId, setFlowState, queryClient]
   );
 
+  useEffect(() => {
+    if (!authenticated && typeof window !== "undefined") {
```
bakeiro, line 203: when & how often could the variables `authActions` & `authenticated` change? this means, are we calling `checkAuthStatus` periodically?

app/web/Dockerfile
aapeliv: It's not really related to this PR: but it's a bit cumbersome every time people want to change the protos to have to change this hash. I am not sure if I want this though: I like it when they point to a particular version.

app/web/features/auth/useAuthStore.ts
aapeliv: I don't actually know how that works. I thought this meant it rusn every time `useAuthStore` is loaded?

app/web/Dockerfile
bakeiro: ✅

app/web/features/auth/useAuthStore.ts
bakeiro: like, this code runs every time the content of either `authenticated` or `authActions` change, then this code it's executed (and the first time this component is rendered).  

if this pretends to be executed periodically I don't know (by looking at the code) if this is achieved, you can use Useffect without a second parameter, so it's executed every time is redered

bakeiro: if works it works, anyhow I can double check the behavior in my local set up

aapeliv:
Superseded by https://github.com/Couchers-org/couchers/pull/4693

PR #4566, opened 2024-07-17
Implement user search by viewport

### LINKED ISSUE #4513: UserSearch should have ability to filter by viewport/bounds
Part of #4129

Frontend part: #4081

spreeni:
closes #4513

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -462,7 +462,7 @@ class Search(search_pb2_grpc.SearchServicer):
                         (1000 + request.search_in_area.radius) / 111111,
                     )
                 )
-            if request.HasField("search_in_viewport"):
+            if request.HasField("search_in_rectangle"):
                 # TODO: Don't allow huge viewports
                 statement = statement.where(
                     func.ST_Within(
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -447,6 +447,7 @@ class Search(search_pb2_grpc.SearchServicer):
                 statement = statement.where(User.camping_ok == request.camping_ok.value)
 
             if request.HasField("search_in_area"):
+                # TODO: Don't allow huge radii
                 # EPSG4326 measures distance in decimal degress
                 # we want to check whether two circles overlap, so check if the distance between their centers is less
                 # than the sum of their radii, divided by 111111 m ~= 1 degree (at the equator)
@@ -462,15 +463,16 @@ class Search(search_pb2_grpc.SearchServicer):
                     )
                 )
             if request.HasField("search_in_viewport"):
+                # TODO: Don't allow huge viewports
                 statement = statement.where(
                     func.ST_Within(
                         User.geom,
                         func.ST_MakeEnvelope(
-                            xmin=request.search_in_viewport.lng_min,
-                            ymin=request.search_in_viewport.lat_min,
-                            xmax=request.search_in_viewport.lng_max,
-                            ymax=request.search_in_viewport.lat_max,
-                            srid=4326,
+                            request.search_in_viewport.lng_min,
+                            request.search_in_viewport.lat_min,
+                            request.search_in_viewport.lng_max,
+                            request.search_in_viewport.lat_max,
+                            4326,
                         ),
                     )
                 )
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/search.py
```diff
@@ -461,6 +461,19 @@ def UserSearch(self, request, context):
                         (1000 + request.search_in_area.radius) / 111111,
                     )
                 )
+            if request.HasField("search_in_viewport"):
```
bakeiro: Hi!
thanks for the work :)

Does this implement an area size limit? for example, if the frontend sends (by a glitch for example) the viewport of the world, which could potentially cause an expensive server load, is the backend ready to reject this request and send an error back to the frontend?

bakeiro: I will implement this check in the frontend, but maybe it´s worth implementing it in the backend as well

app/proto/search.proto
```diff
@@ -104,6 +111,7 @@ message UserSearchReq {
     Area search_in_area = 28;
     // search inside the area defined by the community
     int64 search_in_community_id = 29;
+    RectArea search_in_viewport = 33;
```
aapeliv: Maybe standardize to either viewport or rect?

app/backend/src/tests/test_search.py
```diff
@@ -77,6 +77,40 @@ def test_regression_search_in_area(db):
         assert [result.user.user_id for result in res.results] == [user3.id, user4.id]
 
 
+def test_regression_search_in_viewport(db):
```
aapeliv: Shouldn’t have regression in the name

aapeliv (commented):
That was quick! It’s looking very good but I have a few tiny nits.

app/backend/src/couchers/servicers/search.py
aapeliv: It should be fine to send the backend anything, I think. Not doing it for any viewport in the frontend is a more rate-limiting thing.

app/backend/src/couchers/servicers/search.py
```diff
@@ -461,6 +461,19 @@ def UserSearch(self, request, context):
...
+                statement = statement.where(
+                    func.ST_Within(
+                        User.geom,
+                        func.ST_MakeEnvelope(
+                            xmin=request.search_in_viewport.lng_min,
```
aapeliv: I think you need to pass them in as positional arguments. I don't think postgres does keyword args!

app/backend/src/couchers/servicers/search.py
bakeiro: true... the thing is that I remember some conversations when we talked about the polygons search, to prevent the user to search in big areas..." 🤔

bakeiro: I think that was the reason (maybe I'm wrong) why the backend only searchs 50km around an specific point

aapeliv: The reason for not searching in big areas (countries) with the current system is that we currently only search around a point: if you put in "Australia" the pin ends up in the outback and no one is there. It needs to be a smaller locality so that the 50km means something!

app/backend/src/couchers/servicers/search.py
spreeni: Oh yeah, realized that too 😅 The PR was/is not really ready

app/proto/search.proto
spreeni: Named it `search_in_rectangle` now

app/backend/src/couchers/servicers/search.py
spreeni: So you don't think we should limit it for any searches? Then I'd remove my TODOs :D

aapeliv: I think it's fine for now: we can think about limits later. Right now it should not be a problem.

app/proto/search.proto
aapeliv: Looks like there's a few places left with `search_in_viewport`

spreeni: Oops, guess I wasn't really thorough... Changed the last occurrences

aapeliv (approved):
LGTM!

PR #4569, opened 2024-07-18
Remove raw uses of useAuthStore

aapeliv:


aapeliv:
@bakeiro: I think this is some cleaning up from #4560, the stuff that I think should not be contentious, so I can keep the other stuff there. Could you have a look please?

bakeiro:
code looks good, just wondering why this change? I guess you prefer to use `useAuthContext` so you can create a function there every time `useAuthContext` is called?

aapeliv:
> code looks good, just wondering why this change? I guess you prefer to use `useAuthContext` so you can create a function there every time `useAuthContext` is called?

Yes that's right. I think `useAuthStore` should basically be private, from my reading of the code

Issue #4467, opened 2024-06-16
Filtering of events


aapeliv:
There's been a request to add filtering functionality on the events page. Requested location filter or sort by distance.

aapeliv:
I am making this "important" because I think it's been specifically requested by the Manchester community and it'd be great if we could act on their feedback!

Krishnag09:
ok, i will take this up this week.

spreeni:
Closing as this was implemented with [ !4590](https://github.com/Couchers-org/couchers/pull/4590)

Issue #3989, opened 2024-03-01
[events] create the API endpoints for the new events page


bakeiro:
### Description.

I as a frontend developer:  
I am going to develop a brand new event page, for this I need several endpoints in order to display all the information. Some endpoints could exist, others would need to be created.

Endpoints needed:

- Events of the user's communities.
- Events in a specific location.
- All online events
- Events created by the current user
- Events joined by the current user
- Events created or participated in by the current user.

### Pending tasks:
Check which ones exist and which ones need to be created, then proceed to create the necessary endpoints (task can be split into smaller ones if needed).

Image as reference
![image](https://github.com/Couchers-org/couchers/assets/16777161/8af9497c-c6d1-4efc-bcda-7844b02da031)

Krishnag09:
i can work with you on this. @bakeiro once @aapeliv adds me to the repo as contributor.  And yes we should break this into smaller tasks starting with event creation and event search.

aapeliv:
Some of these exist already:



>  * Events of the user's communities.
> 
`ListMyEventsReq` with `my_communities`

>  * Events created by the current user
> 

`ListMyEventsReq` with `organizing`

>  * Events joined by the current user
> 
`ListMyEventsReq` with `attending`/`subscribed`

>  * Events created or participated in by the current user.
>
`ListMyEventsReq` with `attending`

bakeiro:
@aapeliv do we have an endpoint for the events in a certain location?

spreeni:
@bakeiro There currently is not an endpoint for this. Would you want this for the "Near you" tab? I mean it would be fairly easy to show events in a certain radius around a location, but this would not be very flexible against city size for example.

What exists are events on the community page, so I guess you could map a location to a community and then take these events. But communities are of course not exhaustive.

bakeiro:
about taking events from communities and showing them as if they were from the location you are looking for, it seems confusing to me... I don't think the user understands that, besides, in theory it is possible to do events from one city (barcelona) but in another community (madrid).

bakeiro:
I think the best is simply by location, for example, the user has a location, and it is searched in 50KM around (or 75km I do not know, I guess you can use a large number to avoid not showing events in cities with large areas?)

bakeiro:
also, I was thinking about replacing 'near you' by 'location' so user can specify the location they want.

spreeni:
[...]

PR #4590, opened 2024-07-23
Event search by location

spreeni:
Add event search similar to user search which, among other filters, allows to filter by area.

Addresses #3989, #4467

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -5,12 +5,26 @@ See //docs/search.md for overview.
 from datetime import timedelta
 
 import grpc
-from sqlalchemy.sql import func, or_
+from sqlalchemy.sql import and_, func, or_
 
 from couchers import errors
 from couchers.crypto import decrypt_page_token, encrypt_page_token
 from couchers.db import session_scope
-from couchers.models import Cluster, Event, EventOccurrence, Node, Page, PageType, PageVersion, Reference, User
+from couchers.models import (
+    Cluster,
+    ClusterSubscription,
+    Event,
+    EventOccurrence,
+    EventOccurrenceAttendee,
+    EventOrganizer,
+    EventSubscription,
+    Node,
+    Page,
+    PageType,
+    PageVersion,
+    Reference,
+    User,
+)
 from couchers.servicers.api import (
     hostingstatus2sql,
     meetupstatus2sql,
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/servicers/search.py
```diff
@@ -514,3 +523,91 @@ def UserSearch(self, request, context):
...
+                    )
+                )
+            if request.HasField("search_in_community_id"):
+                # could do a join here as well, but this is just simpler
+                node = session.execute(
+                    select(Node).where(Node.id == request.search_in_community_id)
+                ).scalar_one_or_none()
+                if not node:
+                    context.abort(grpc.StatusCode.NOT_FOUND, errors.COMMUNITY_NOT_FOUND)
+                statement = statement.where(func.ST_Contains(node.geom, EventOccurrence.geom))
```
aapeliv, line 646: Is this also what user search within community means? Events also have a parent community, that's another sense of being in the community. I think you should do an `or_` between the two?

app/backend/src/couchers/servicers/search.py
```diff
@@ -514,3 +523,91 @@ def UserSearch(self, request, context):
...
+
+            if request.HasField("after"):
+                statement = statement.where(EventOccurrence.start_time > to_aware_datetime(request.after))
+            if request.HasField("before"):
+                statement = statement.where(EventOccurrence.end_time < to_aware_datetime(request.before))
+
+            page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
+            # the page token is a unix timestamp of where we left off
+            page_token = dt_from_millis(int(request.page_token)) if request.page_token else now()
```
aapeliv, line 655: Sorry to be annoying but could you add a test to double check pagination works correctly both forwards and backwards please?

aapeliv (commented):
Looks good, just a few very minor comments!

app/backend/src/couchers/servicers/search.py
spreeni: @aapeliv Yes, I basically copied this from UserSearch. To list events associated with a community, there is the endpoint `ListEvents` in `communities.proto`. Tbh, I feel it might be more concise to have only one endpoint to list events with a lot of filter options.

The difference between both options right now is, that this filter would work also for parent communities. But on second thought, it probably is more efficient to do this without a geometry-check and with basic relational logic. What do you think?

aapeliv: The query itself should be fast enough either way. I am more concerned just about making the concepts and logic coherent.

app/backend/src/couchers/servicers/search.py
spreeni: Added tests for it now 👍

app/backend/src/couchers/servicers/search.py
spreeni: Ok, then I would maybe just leave it as is, as it has the same behaviour as the user search as an area search. And as you can list events associated with a certain community in ListEvents - or do you have another preference? I'm all ears and don't have a very strong opinion

aapeliv (dismissed):
LGTM!

spreeni:
@aapeliv Added the following filters, have a look again. But should be fairly complete now :)
```
  oneof online_status {
    bool only_online = 12;
    bool only_offline = 13;
  }

  // all false => everything
  bool subscribed = 14;
  bool attending = 15;
  bool organizing = 16;
  // include events in the user's communities
  bool my_communities = 17;
```

aapeliv (approved):
LGTM! Sorry for the delay :)

Issue #4349, opened 2024-05-21
Strong Verification


aapeliv:
* ~I think it does not currently handle expiring passports, or there is some other issue with doubles for users.~
* ~Needs update for getting initiation state, not just "waiting for user"~
* ~FAQ~
* ~Privacy policy~
* Frontend part: #4413 
* Technical blog post: #4354 
* Instructional video
* Onboarding email

nabramow:
@aapeliv what's up with this one? Does it go back to ideas?

Krishnag09:
@nabramow - we and @bakeiro have separate tickets for this that are done.

nabramow:
Closing as this is now happening in #4413 and #4354

PR #4594, opened 2024-07-25
Add field and check donation for phone verified 

Krishnag09:
The PR adds a new profile field to track donation, to act as a paywall for phone verification

closes the issue - https://github.com/Couchers-org/couchers/issues/4349

--->

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -608,6 +610,21 @@ class StrongVerificationCallbackEvent(Base):
     iris_status = Column(String, nullable=False)
 
 
+class PhoneVerificationCallbackEvent(Base):
```
aapeliv: I think this is unused?

app/backend/src/couchers/servicers/donations.py
```diff
@@ -112,6 +112,10 @@ def Webhook(self, request, context):
                 # amount comes in cents
                 amount = int(data_object["amount"]) // 100
                 receipt_url = data_object["receipt_url"]
+
+                # may be check for amount to enable phone verify
+                session.add(User(has_donated=True))
```
aapeliv: This syntax creates a new user with just this flag set. The SQLalchemy ORM loads the user and will automatically update it so you just need to do `user.has_donated = True`

app/proto/api.proto
```diff
@@ -265,6 +265,7 @@ message User {
   google.protobuf.BoolValue parking = 119;
   ParkingDetails parking_details = 120; // CommonMark without images
   google.protobuf.BoolValue camping_ok = 121;
+  google.protobuf.BoolValue has_donated = 123;
```
aapeliv: This probably shouldn't be returned to other users.

aapeliv (commented):
I think there are two things that still need to be implemented:

1. It would be good to return the `has_donated` flag in GetAccountInfo (account.proto/account.py)
2. We need to somehow populate the `has_donated` flag when we set it up so that those who have already donated get it ticked off. I can help make this happen in the migration.

app/backend/src/tests/test_fixtures.py
```diff
@@ -303,6 +303,11 @@ def invocation_metadata(self):
 
         session.commit()
 
+        if has_donated:
+            user.has_donated = True
+
+        assert user.has_donated == has_donated
+
```
aapeliv: This is superfluous

app/media/src/media/server.py
```diff
@@ -9,13 +9,12 @@
 import pyvips
 import sentry_sdk
 from flask import Flask, abort, request, send_file
+from media.crypto import verify_hash_signature
```
aapeliv: Please undo formatting in `app/media`

aapeliv (commented):
1. Database migrations (update the `has_donated` flag based on existing data in the database) -- I can help with this: it's going to be complicated

This needs a database query to check who has already donated based on existing tables, so we set `has_donated=True` for those users.

2. Add a env var flag or constant in `constants.py` as to whether to enforce this or not. So we can update the frontend first.

app/backend/src/tests/test_fixtures.py
aapeliv: The `kwargs` should take care of this, no need to handle `has_donated` separately

aapeliv:
@Krishnag09 : could you please rebase onto `develop` or merge `develop` into this branch, please?

Are you OK if I push a commit to fix up the migrations onto this branch? Other than that and minor formatting stuff (I can fix that up too), I think this is looking good to go!

Krishnag09:
> 1. Database migrations (update the `has_donated` flag based on existing data in the database) -- I can help with this: it's going to be complicated
> 
> This needs a database query to check who has already donated based on existing tables, so we set `has_donated=True` for those users.
> 
> 2. Add a env var flag or constant in `constants.py` as to whether to enforce this or not. So we can update the frontend first.

do you still want me to do these?

aapeliv:
> do you still want me to do these?

It's all done now in the commit I pushed. I think it just needs an update to the migration (to say down revision is `3ac39dcf3b5a`).

But let's wait a little bit before merging so I can do the frontend part first!

PR #4678, opened 2024-08-28
Change ports to not conflict with mac airplay

nabramow:
There's a [known issue on some macs with port 5000 being used for Airplay](https://archive.md/YhxlX). This was causing an error when I tried to spin the docker containers up.

I confirmed this was the problem by turning the Airplay Receiver off, running `docker-compose up` and it works. Turn it back on and it fails again. The setting on mac is here:

Screenshot 2024-08-27 at 11 11 50 PM

**TO TEST:**
- `docker-compose down`
- `docker-compose up --build`
- Go to `http://localhost:3000/`
- If it runs, nothing is broken! The error happened when running `docker-compose up` so if we get past that all is good. :-)

**Backend checklist**
- [N/A ] Added tests for any new code or added a regression test if fixing a bug
- [ x ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


aapeliv:
There's a couple other places you'll need to update the port, I think all these hits that are not hidden should be updated:

Screenshot from 2024-08-28 09-59-10

nabramow:
> There's a couple other places you'll need to update the port, I think all these hits that are not hidden should be updated:
> 
> Screenshot from 2024-08-28 09-59-10

Ooo good catch! Question. For the test ones is there a reason they're not referencing the .env var and use a manual reference?

I guess because it's not set to switch the vars out for the different environments?

aapeliv:
> Ooo good catch! Question. For the test ones is there a reason they're not referencing the .env var and use a manual reference?

The test env vars are set in `test_fixtures.py`. Writing out the full URLs in the test is clearer and easier to check IMO, compared to constructing the URL using the env vars. But I guess this comes down very much to preference!

aapeliv:
I mean you could leave the tests as-is, but I think it's better to be consistent with the ports, that way when people read it it's easier to connect different things.

nabramow:
> I mean you could leave the tests as-is, but I think it's better to be consistent with the ports, that way when people read it it's easier to connect different things.

No worries I changed them out. I think for now not a big deal. I was just used to it the other way lol.

app/media/Dockerfile
```diff
@@ -19,4 +19,4 @@ ENV MEDIA_UPLOAD_LOCATION="/uploads"
 ARG version
 ENV VERSION=$version
 
-CMD cd src && uwsgi --http 0.0.0.0:5000 --wsgi-file media/server.py --callable app
+CMD cd src && uwsgi --http 0.0.0.0:5001 --wsgi-file media/server.py --callable app
```
nabramow, line 22: I was a liiiiitle unsure about this one so just flagging it. Since it says media/server and I changed that env var I think I wanna change this too?

aapeliv: Yep! This tells the media server to actually run on that port

nabramow:
@aapeliv You happy to approve this? I think I got all the ports now!

aapeliv (approved):
Yep, think you got them all :D

LGTM!

PR #4682, opened 2024-08-29
Create notifications settings preference page in React

### LINKED ISSUE #4416: Settings page for notifications
See: https://next-console.couchershq.org/notifications

This should be brought into the web frontend.

nabramow:
Closes #4416 

Right now users edit their notification email and push preferences via this separate console page [here](https://next-console.couchershq.org/notifications). The task is to convert this to React and work it into the frontend web app.

We discussed in last standup the route should be `/account-settings/notifications` so it's there. To access it you can go to account settings and click the notification settings button.

In standup we also decided to change the order of the Account Settings page to put the not-editable ones up top, that's done here too.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ x ] Formatted my code with `make format`
- [ x ] There are no warnings from `make lint`
- [ x ] There are no console warnings when running the app - **Well the ones that don't involve upgrading Material UI at least**
- [ x ] Added any new components to storybook
- [ x ] All tests pass


app/backend/src/couchers/urls.py
```diff
@@ -39,6 +39,10 @@ def account_settings_link():
     return f"{config['BASE_URL']}/account-settings"
 
 
+def notification_settings_link():
+    return f"{config['BASE_URL']}/account-settings/notification-settings"
+
+
```
aapeliv, line 45: The file `app/backend/src/couchers/notifications/background.py`, on line 61, adds a link to "manage notification settings" to the email footer. Could you please update `urls.account_settings_link()` to `urls.notification_settings_link()`!

https://github.com/Couchers-org/couchers/blob/96dc7b746a1f1dacf8e51026ae3993fbb5d42c4a/app/backend/src/couchers/notifications/background.py#L61-L63

docs/urls.md
```diff
@@ -74,3 +74,6 @@ URL:  https://couchers.org/password-reset
 
 Title: Complete password reset / set new password page
 URL: https://couchers.org/complete-password-reset?token=abc123
+
+Title: Edit notification settings
+URL:  https://couchers.org/profile/notifications
```
aapeliv: This is now different, right?

aapeliv (commented):
Backend side is OK, just a minor comment!

app/backend/src/couchers/urls.py
nabramow: Done!

docs/urls.md
nabramow: Ooo good catch, thanks! Done!

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    ? "account_settings"
+                    : topic.topic;
+
+                if (!acc[key]) {
+                  acc[key] = [];
+                }
+
+                if (subTopic.userEditable) {
+                  acc[key].push({ ...subTopic, topic: topic.topic });
+                }
```
nabramow: FYI I decided to make the not userEditable ones not visible. If they can't change them, doesn't seem to make sense to show them?

I notice we show them on the console but I dunno.

Let me know if you prefer I make the switch disabled for them.

app/web/features/auth/locales/es.json
```diff
@@ -159,7 +159,97 @@
...
+          "verify": "Tu número de teléfono ha sido verificado"
+        },
+        "birthdate": {
+          "change": "Tu fecha de nacimiento ha sido cambiada"
+        },
+        "gender": {
+          "change": "El género mostrado en tu perfil ha sido cambiado"
+        }
+      }
+    }
```
nabramow, line 252: @bakeiro I 100% did these translations with ChatGPT so if anything looks weird let me know.

app/web/features/auth/notifications/NotificationSettingsListItem.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+  const [isCollapseOpen, setIsCollapseOpen] = useState<boolean>(false);
+
+  const handleCollapseClick = () => {
+    setIsCollapseOpen(!isCollapseOpen);
+  };
+
+  const renderItems = () =>
+    items
+      .filter((item) => item.userEditable)
```
nabramow, line 76: This is so if a group has no userEditable items (i.e. account security), it won't show the group at all (i.e. an empty collapse).

app/web/features/communities/events/InviteCommunityDialog.tsx
```diff
@@ -1,21 +1,21 @@
+import { DialogProps, Link as MuiLink } from "@material-ui/core";
+import Alert from "components/Alert";
+import Button from "components/Button";
```
nabramow: I didn't change anything here, the linting reordered the imports I think.

app/web/package.json
```diff
@@ -14,6 +14,7 @@
     "format:check": "prettier --check '**/(*.ts|.tsx)'",
     "jest": "jest --modulePaths=.",
     "lint": "next lint",
+    "lint:fix": "next lint --fix .",
```
nabramow, line 17: It was annoying me that there was only one to see the incorrect linting but not autofix it. There's an autofix for the backend so added this. Let me know if that's bad.

aapeliv (commented):
Backend looks good to me!

aapeliv:
I had a poke around at https://couchers-git-web-featuresettings-notifications-couchers-org.vercel.app/account-settings/notifications

The next/staging server it targets is in Australia (so that the devs in NA get a feel for what latency does to the app :P).

Right now it seems there's no loading state? If I click on a checkbox it doesn't do anything until (presumably the API call is executed and) the state is updated. It makes it feel non-responsive in a way. Would there be a way to add loading state somehow to indicate the toggle was clicked but it's saving?

Also I think I discovered a bug: if you toggle another checkbox while the first one is loading, it doesn't seem to consistently toggle both.

Here's a video (though you can't see when I click so not sure how useful it is):


https://github.com/user-attachments/assets/b4403dc5-d03a-4aa5-9dc3-c225820e1a5d

Ooo thanks for catching those. I was thinking how I wanted to handle that loading state but got distracted. 

After reading through some Material-UI github issues I found a good way to add a loading state since they don't have it built in.

Also fixed the bug with the two switches interfering with each other by giving separate loading state and disabling it from changing again while loading.

app/web/components/CustomColorSwitch.tsx
```diff
@@ -0,0 +1,84 @@
...
+    checked: {},
+    disabled: {},
+    track: {},
+  }))(Switch);
+
+  const Icon = () => (
+    <div
+      className={`${classes.circle} ${
+        checked && !isLoading ? classes.active : ""
+      }`}
```
aapeliv: I think there's a helper for this from `import classNames from "classnames";`.

I think this would be `classNames(classes.circle, {classes.active: checked && !isLoading})`

app/web/components/CustomColorSwitch.tsx
```diff
@@ -0,0 +1,84 @@
...
+        "& + $track": {
+          backgroundColor: checked ? color : theme.palette.grey[200],
+          opacity: 0.4,
+        },
+      },
+    },
+    checked: {},
+    disabled: {},
+    track: {},
+  }))(Switch);
```
aapeliv: I am wondering about this: I've not seen it elsewhere in the codebase (using `withStyles`). I feel we have some other custom-styled Mui components as well. I'm not sure how they are done.

Another thing is that should we just apply this override globally in `theme.ts`? Is there a case where we don't want this custom styled switch?

Anyway, this is not so important, feel free to ignore :)

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
aapeliv: I think that's good!

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+          });
+
+          return acc;
+        },
+        {}
+      );
+
+      setGroups(computedGroups);
+    } catch (error) {
+      console.error("Error fetching notification settings data:", error);
```
aapeliv: Should you re-raise it here?

app/web/features/auth/notifications/NotificationSettingsSubListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+  return (
+    <>
+      {mutationError && (
+        <Alert severity="error">
+          {mutationError || t("global:error.unknown")}
+        </Alert>
+      )}
+      <Typography className={classes.descriptionText}>
+        {t(
+          //@ts-ignore - I spent hours on this type with no luck
```
aapeliv, line 122: :(

aapeliv (commented):
This looks good to me! I left a few comments that I picked up.

app/web/components/CustomColorSwitch.tsx
nabramow: Okay this is done!

app/web/components/CustomColorSwitch.tsx
nabramow: I've changed it to using our `makeStyles` and switched to the custom-styled Mui components.

As for the global override, should I make a separate ticket for that? I can make sure all the existing switches use the loading state?

I originally did this as the existing switch was orange and it could only be made the primary, secondary or default colors, none of which we wanted based on the design.

I think better in a separate ticket then if I mess something up or miss one easier to notice and review.

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
nabramow: What specifically are you thinking I should re-raise?

app/web/components/CustomColorSwitch.tsx
aapeliv: Sounds good!

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
aapeliv: Right now you are catching and suppressing any error that occurs in this block, right? If I understand correctly you are just ignoring the error and the user will see no groups?

When I do backend stuff we try to not suppress errors: only catch and handle the ones we can recover from, and let others bubble up to a global exception handler.

I guess on the frontend there's an error boundary. Of course it will look bad for the user if they get that, but it won't silently fail.

I'm not sure if this is right for the frontend, hence the qeustion :)

aapeliv (dismissed):
Looks good to me! Feel free to merge when CI passes :)

app/web/features/auth/notifications/EditNotificationSettingsPage.tsx
nabramow: Ah yes you're right I should generally def do that. On further look of this specific one though, I removed the try/catch. This is because the api call error handling is already being handled in the hook in this case, so the only thing this would be catching is if my parsing of the groups went wrong, which I would already know at this point.

aapeliv (approved):
LGTM

app/web/features/auth/locales/es.json
bakeiro: ahahha ok!!

PR #4693, opened 2024-08-31
Auth check

aapeliv:
It is currently possible for the react frontend to lose the "persistent" state it keeps in local storage: it seems sometimes when we upgrade the app this happens, and it otherwise too seems flaky.

The website cookies on the other hand are relatively solid. This means that occasionally the frontend thinks it's logged out when in reality the cookie is still there for a valid session. This means that users sometimes need to log in unnecessarily: this (according to metrics) seems to represent about 15-20% of logins. Logging in is a pain and so this metric does not capture the number of people who gave up because they were logged out.

This PR introduces an auth check that will ping the backend (using a new API call called `AuthCheck`) which will return back information about the current auth state, and will re-login the user if their state is out of sync. (In case the frontend thinks it's logged in but the cookie/session is expired, we catch that and log the user out in the frontend.)

[The cookie is not accessible to JS for security reasons, it's only available on the transport layer, so the frontend needs to query the backend about cookie status.]

This continues from https://github.com/Couchers-org/couchers/pull/4560.

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/auth/useAuthStore.ts
+++ b/app/web/features/auth/useAuthStore.ts
@@ -4,7 +4,7 @@ import { GLOBAL } from "i18n/namespaces";
 import Sentry from "platform/sentry";
 import { clearStorage, usePersistedState } from "platform/usePersistedState";
 import { AuthRes, SignupFlowRes } from "proto/auth_pb";
-import { useMemo, useRef, useState } from "react";
+import { useEffect,useMemo, useRef, useState } from "react";
 import { useQueryClient } from "react-query";
 import { service } from "service";
 import isGrpcError from "service/utils/isGrpcError";
@@ -15,6 +15,10 @@ export default function useAuthStore() {
     false
   );
   const [jailed, setJailed] = usePersistedState("auth.jailed", false);
+  const [checkedAuthStatus, setCheckedAuthStatus] = usePersistedState(
+    "auth.checkedAuthStatus",
+    false
+  );
   const [userId, setUserId] = usePersistedState<number | null>(
     "auth.userId",
     null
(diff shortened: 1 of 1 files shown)


nabramow:
Just a big picture question while I wrap my head around this. Was the main issue that the `usePersistedState` hook was not persisting the state?

Reading the PR description I was wondering if this is a bandaid instead of fixing the core issue, but I'm still new to this all.

Do you know how I would replicate this issue? Usually I like to replicate the bug on the main branch, then use your branch to make sure it fixed it.

aapeliv:
> Just a big picture question while I wrap my head around this. Was the main issue that the `usePersistedState` hook was not persisting the state?
> 
> Reading the PR description I was wondering if this is a bandaid instead of fixing the core issue, but I'm still new to this all.

Yes, I guess this is a bandaid. The `usePersistedState` hook is *mostly* working fine, but it seems that sometimes it loses state (possibly around when the frontend is upgraded or similar?). So this is indeed a bandaid around that.

That's a good point. We could look into another way of syncing the cookie with the frontend better as an alternative.

> Do you know how I would replicate this issue? Usually I like to replicate the bug on the main branch, then use your branch to make sure it fixed it.

You can emulate it by: logging in, then going to dev tools and clearing local storage (but not cookies). The cookie will still be there, but the frontend will not know.

app/web/features/auth/useAuthStore.ts
```diff
@@ -131,12 +135,53 @@ export default function useAuthStore() {
...
+            const auth = res.authRes!;
+            setUserId(auth.userId);
+            Sentry.setUser({ id: auth.userId.toString() });
+
+            //this must come after setting the userId, because calling setQueryData
+            //will also cause that query to be background fetched, and it needs
+            //userId to be set.
+            setJailed(auth.jailed);
+            setAuthenticated(true);
+          }
```
nabramow, line 154: Would you want to logout here if `res.loggedIn` equals false? In an else clause for example?

app/web/features/auth/useAuthStore.ts
```diff
@@ -131,12 +135,53 @@ export default function useAuthStore() {
...
+        } catch (e) {
+          Sentry.captureException(e, {
+            tags: {
+              component: "auth/useAuthStore",
+              action: "checkAuthStatus",
+            },
+          });
+          setError(isGrpcError(e) ? e.message : fatalErrorMessage.current);
+        }
+      },
```
nabramow, line 164: Same here would you want to logout in the catch?

app/web/features/auth/useAuthStore.ts
```diff
@@ -131,12 +135,53 @@ export default function useAuthStore() {
         }
         setLoading(false);
       },
+      async checkAuthStatus() {
+        try {
```
nabramow, line 139: I wonder if adding a check `if (authenticated || loading) return;` would help here instead of the useEffect? That's kinda what you're trying to do yeah?

aapeliv: Let me comment here instead of the other two.

`checkAuthStatus` is right now only run if `isAuthenticated` is true, but I've left that condition in the `useEffect`. This is because of the comment at the bottom "there should be no dependenices on the state". I originally had it there, but then read this...

Also, regarding logging out: there is a well functioning mechanism that will log the user out in the frontend if the backend returns an unauthenticated response via the API.

aapeliv:
> I was wondering if this is a bandaid instead of fixing the core issue

This comment made me think some more about this. I think you are right that this is a bit of a poor bandaid. Let me try another technique (based on cookies) to synchronize the backend/frontend better, and if that doesn't work we can come back to this PR!

nabramow:
Okay so one thing I noticed might be an issue. In the `usePersistedState` hook the `storage` key in the function arguments defaults to `localStorage`, but the `clearStorage` function clears the `sessionStorage`.

I'm wondering it that might also be causing some unexpected behavior. Maybe it's not finding the logged in State as the wires are crossed there?

Could also be a race condition if the issue arises from rapid state changes (e.g., during login/logout), then debouncing the state updates to localStorage to prevent race conditions could help.

bakeiro:
Question: are we sure we lose the persistent data every time there is a deployment? or might be another reason?

bakeiro:
Question 2, can we log when a persistent state is deleted? (maybe with Sentry?)

I don't understand how the local storage gets deleted by itself, if this happens every time a new version is deployed I guess that Next.js detects this and deletes the storage after

aapeliv:
Taken over by #4751

PR #4694, opened 2024-08-31
Native link

aapeliv:
For now we are building the native app by largely embedding WebViews that load from the frontend web server.

This PR introduces `nativeLink`, which tells us if we're in a webview and allows us to accommodate the embedding in the frontend. For example this hides navigation menus (will be implemented natively) and some pretty stuff during the signup flow.

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/global.d.ts
+++ b/app/web/global.d.ts
@@ -1,0 +1,6 @@
+interface Window {
+  ReactNativeWebView?: {
+    injectedObjectJson: () => any;
+    postMessage: (message: string) => void;
+  };
+}
--- a/app/web/global.d.ts
+++ b/app/web/global.d.ts
@@ -1,6 +1,6 @@
 interface Window {
   ReactNativeWebView?: {
-    injectedObjectJson: () => any;
+    injectedObjectJson: () => string;
     postMessage: (message: string) => void;
   };
 }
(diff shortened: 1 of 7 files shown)


app/web/global.d.ts
```diff
@@ -0,0 +1,6 @@
+interface Window {
+  ReactNativeWebView?: {
+    injectedObjectJson: () => any;
```
nabramow: Any chance there's something else here you could use besides any?

aapeliv: I believe it’s an arbitrary JSON object. Is there a more restricted type for that?

nabramow:
Okay just ran the branch and clicked around. All seems good there!

app/web/global.d.ts
nabramow: You could do `Record<string, any>` if you want to represent an object where the keys are strings, and the values can be of any type. That would at least make sure it's an object!

nabramow (dismissed):
Tested and seems good to me! Just the one typescript type thing but that's not really blocking.

app/web/global.d.ts
aapeliv: OK, good idea! Thanks.

aapeliv: Well I realized it's actually just a string. Not sure why they call it JSON (maybe on the sending end it needs to be JSON).

nabramow: Ah it's probably a stringified object then. That's not uncommon when sending data via an api or something.

PR #4701, opened 2024-09-02
Add global message

aapeliv:
This allows us to put up a message on the site without having to rebuild the whole frontend.

This helps with outages so we can broadcast a message to all users on the platform with very short notice.

**Web frontend checklist**
- [ ] All tests pass


app/web/components/Navigation/Navigation.tsx
```diff
@@ -565,6 +565,7 @@ export default function Navigation() {
           </div>
         </Hidden>
       </Toolbar>
+      <GlobalMessage />
```
nabramow, line 568: I think it would be more efficient to fetch the data here in this component, then use it to conditionally render `<GlobalMessage/>` and pass what you need in the props.

Otherwise you're gonna do an api call every time the `Navigation` loads.

Or honestly since it's a global message might want to do it even higher up in a parent component.

nabramow:
I am seeing the global message always when running locally. I assume you want it to be invisible by default?

Screenshot 2024-09-03 at 12 15 36 PM

aapeliv:
> I am seeing the global message always when running locally. I assume you want it to be invisible by default?

This is the expected behavior when the message is set: that it's always visible. If there is no message (the "message" value is empty in the `.json`) then it will not show. But for now I've set this success message so we can see it when testing :).

I could make it dismissable, but then would have to make it show up again if the message changes.

nabramow:
> > I am seeing the global message always when running locally. I assume you want it to be invisible by default?
> 
> This is the expected behavior when the message is set: that it's always visible. If there is no message (the "message" value is empty in the `.json`) then it will not show. But for now I've set this success message so we can see it when testing :).
> 
> I could make it dismissable, but then would have to make it show up again if the message changes.

Do you mean to have it show on prod though? Looks like the prod one also has a message:

https://gm.couchershq.org/prod.json

aapeliv:
> > > I am seeing the global message always when running locally. I assume you want it to be invisible by default?
> > 
> > 
> > This is the expected behavior when the message is set: that it's always visible. If there is no message (the "message" value is empty in the `.json`) then it will not show. But for now I've set this success message so we can see it when testing :).
> > I could make it dismissable, but then would have to make it show up again if the message changes.
> 
> Do you mean to have it show on prod though? Looks like the prod one also has a message:
> 
> https://gm.couchershq.org/prod.json

Yes for now. I'll take it down before deploying! I think vercel uses a mix of prod env vars so it seems to be loading the prod one.

aapeliv:
@nabramow I made it dismissable and made it recheck for a message at most every 5 minutes! Let me know what you think :)

app/web/components/Navigation/Navigation.tsx
aapeliv: I've added fetch rate limiting into the component.

app/web/components/GlobalMessage.tsx
```diff
@@ -1,13 +1,32 @@
...
+import React, { useEffect } from "react";
 
 interface GlobalMessageData {
+  // the unix timestamp (as a string with milliseconds) when the message was issued
+  epoch: string;
   severity: "success" | "info" | "warning" | "error";
   message: string;
 }
 
+const TIME_BETWEEN_CHECKS_MS = 300_000; // 5 min
```
nabramow, line 12: Should it have the underscore?

aapeliv: It's a thousands separator to make it more readable: https://github.com/tc39/proposal-numeric-separator

app/web/components/GlobalMessage.tsx
```diff
@@ -0,0 +1,64 @@
...
+  // last time we queried it
+  const [lastCheck, setLastCheck] = usePersistedState<number | null>(
+    "globalmessage.lastcheck",
+    null
+  );
+  // the epoch value of the last message we dismissed
+  const [dismissedEpoch, setDismissedEpoch] = usePersistedState<string | null>(
+    "globalmessage.dismissed",
+    null
+  );
```
nabramow, line 29: This is more a consideration for the future, but I feel like you're using `usePersistedState` here, which stores something in localStorage/sessionStorage, like a reducer, when it might make more sense to use `useReducer` for something like this (i.e. when you wanna keep track of state above the level of a specific component).

Similarly React Context could be a good candidate.

app/web/components/GlobalMessage.tsx
```diff
@@ -0,0 +1,64 @@
...
+        console.error("Error fetching global message:", error);
+      }
+    };
+    if (
+      !lastCheck ||
+      new Date().getTime() - lastCheck > TIME_BETWEEN_CHECKS_MS
+    ) {
+      fetchData();
+    }
+  }, [setData, setLastCheck, lastCheck]);
```
nabramow, line 52: I know you're using `usePersistedState`, but since `lastCheck` is in the dependencies I worry that there's a chance this could end up in an infinite checking loop where it keeps checking and resetting the lastCheck variable?

Since you have the `data.epoch` could the if clause just be `if (data.epoch / Date.now() > TIME_BETWEEN_CHECKS_MS` or am I misunderstanding?

app/web/components/GlobalMessage.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const dismiss = () => {
+    if (!data) return;
+    setDismissedEpoch(data!.epoch);
+  };
+
+  return data && data.epoch && data.epoch != dismissedEpoch ? (
```
nabramow, line 59: Just a nit but maybe make `data && data.epoch && data.epoch != dismissedEpoch` a separate named variable so its more readable?

app/web/components/GlobalMessage.tsx
```diff
@@ -0,0 +1,64 @@
...
+    <MuiAlert severity={data.severity} onClose={dismiss}>
+      <span dangerouslySetInnerHTML={{ __html: data.message }} />
```
nabramow, line 61: Does this need to be html? Could it just be a string?

aapeliv: We sometimes need to add a link

nabramow (approved):
Approving after confirming on video this likely won't end up with an infinite useEffect re-render ;)

PR #4743, opened 2024-09-08
fix results map async results load

bakeiro:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/search/MapWrapper.tsx
+++ b/app/web/features/search/MapWrapper.tsx
@@ -56,8 +56,8 @@ const useStyles = makeStyles((theme) => ({
 
 interface mapWrapperProps {
   selectedResult:
-  | Pick<User.AsObject, "username" | "userId" | "lng" | "lat">
-  | undefined;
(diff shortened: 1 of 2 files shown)


app/web/features/search/MapWrapper.tsx
```diff
@@ -56,8 +56,8 @@ const useStyles = makeStyles((theme) => ({
 
 interface mapWrapperProps {
   selectedResult:
-  | Pick<User.AsObject, "username" | "userId" | "lng" | "lat">
-  | undefined;
+    | Pick<User.AsObject, "username" | "userId" | "lng" | "lat">
+    | undefined;
```
nabramow, line 60: Will the `selectedResult` be possibly totally undefined or just an empty object? Or are some of the object keys undefined?

aapeliv (approved):
LGTM

app/web/features/search/MapWrapper.tsx
bakeiro: hmm yap it's undefined by default, until a user selects a result :)

Issue #4565, opened 2024-07-16
Add frontend pagination to map results


bakeiro:
**Current situation:**
User needs to scroll down until the end of the list of users to load the new users page (which is absolutely not user friendly)

**Desired situation**
be like airBnb :D, let's use pagination [image attached]
![image](https://github.com/user-attachments/assets/310714df-f1a0-4e7b-a1fe-990135646608)

bakeiro:
how we implement this on mobile?

bakeiro:
On mobile
![image](https://github.com/user-attachments/assets/d2a5d579-637d-4a26-a0d8-e3a1a18ae477)

nabramow:
@aapeliv Do we have pagination on the backend that this ticket would need for map?

If not this should go to the backlog and we should make a backend ticket.

bakeiro:
do first this ticket https://github.com/Couchers-org/couchers/issues/5323

PR #4762, opened 2024-09-09
Fix linting errors

nabramow:
Looks like a branch got merged with some linting errors. Just quickly fixing these because I want to be able to check the "no linting errors" box in my PR ;).

**Web frontend checklist**
- [ x ] All tests pass


aapeliv:
Is @bakeiro working on these?

bakeiro:
oh, I see... are were these errors  or warnings? (out of curiosity)

bakeiro:
ah I though I executed that in my PR... seems that I mess up with `git stash` and the merge conflicts...

app/web/features/search/SearchBox.tsx
```diff
@@ -55,42 +56,47 @@ export default function SearchBox({
   const { t } = useTranslation([GLOBAL, SEARCH]);
   const classes = useStyles();
 
-  const { control, setValue, errors, register, handleSubmit } = useForm({
+  const { control, errors } = useForm({
     mode: "onChange",
   });
 
-  function handleOnChangeAutocomplete(event: "" | GeocodeResult) {
+  const handleOnChangeAutocomplete = (event: GeocodeResult) => {
```
bakeiro, line 63: why const rather normal function? (just to understand)

app/web/features/search/SearchBox.tsx
```diff
@@ -55,42 +56,47 @@ export default function SearchBox({
...
 
-  function handleOnChangeKeyword(event: React.ChangeEvent<HTMLInputElement>) {
+  const handleOnChangeKeyword = (
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
     setQueryName(event.target.value);
     setLocationResult({
       ...locationResult,
-      location: undefined,
+      location: new LngLat(0, 0),
```
bakeiro, line 75: is the behavior here the same?

app/web/features/search/SearchBox.tsx
```diff
@@ -55,42 +56,47 @@ export default function SearchBox({
...
+    event: React.ChangeEvent<HTMLInputElement>,
+    value: "location" | "keyword"
+  ) => {
+    setSearchType(value);
     setLocationResult({
       ...locationResult,
       name: "",
       simplifiedName: "",
-      location: undefined,
+      location: new LngLat(0, 0),
```
bakeiro, line 88: why not undefined?

app/web/features/search/SearchPage.tsx
```diff
@@ -104,12 +107,8 @@ export default function SearchPage({
         completeProfileFilter,
       ],
       ({ pageParam }) => {
-        const lastActiveComparation = parseInt(
-          lastActiveFilter as unknown as string
-        );
-        const hostingStatusFilterComparation = parseInt(
-          hostingStatusFilter as unknown as string
-        );
+        const lastActiveComparation = lastActiveFilter;
```
bakeiro: same here, I remember I did this because when I was debugging the backend didn't receive the same values... is this checked?

app/web/features/search/SearchResultsList.test.tsx
```diff
@@ -16,25 +19,28 @@ const getLanguagesMock = service.resources.getLanguages as MockedService<
   typeof service.resources.getLanguages
 >;
 
-function testResults() {
-  return {
+function mockTestResults() {
+  const mockResults: InfiniteData<UserSearchRes.AsObject> | undefined = {
```
bakeiro, line 23: 👍

app/web/features/search/SearchPage.tsx
bakeiro: ok I checked and backend receive the same values... all in order!

bakeiro: exactly, this creates a bug, there is a reason why I did this way, with your change if you use the lastActive filter (for example want to search lastActive "last day" and then want to unset this filter creates a bug)

image

bakeiro: that's why I created another issue (https://github.com/Couchers-org/couchers/issues/4565) and wanted to fix it there plus implement pagination

bakeiro: so please restore :)

app/web/features/search/FilterDialog.stories.tsx
```diff
@@ -12,7 +11,7 @@ export default {
 
 const Template: Story<any> = () => {
   const [isOpen, setIsOpen] = useState(false);
-  const searchFilters = useRouteWithSearchFilters("");
+
```
bakeiro, line 14: 👍

app/web/features/search/FilterDialog.tsx
```diff
@@ -86,8 +86,8 @@ interface FilterDialogProps {
   isOpen: boolean;
   onClose(): void;
   queryName: undefined | string;
-  setQueryName: Dispatch<SetStateAction<undefined | string>>;
-  setLocationResult: any;
+  setQueryName: Dispatch<SetStateAction<string>>;
+  setLocationResult: Dispatch<SetStateAction<GeocodeResult>>;
```
bakeiro, line 90: thanks, I always use `any` when I try to set the type multiple times and fail (and after I forget to change it 😅)

app/web/features/search/SearchResultsList.tsx
```diff
@@ -161,10 +162,10 @@ export default function SearchResultsList({
...
                   onSelect={() => {
                     setSelectedResult({
-                      username: result.user?.username as string,
-                      userId: result.user?.userId as number,
-                      lng: result.user?.lng as number,
-                      lat: result.user?.lat as number,
+                      username: result.user!.username,
+                      userId: result.user!.userId,
+                      lng: result.user!.lng,
+                      lat: result.user!.lat,
```
bakeiro: 🔝

app/web/features/search/SearchBox.tsx
nabramow: It's not a huge deal either way, const seems a bit more modern to me.

Technically, using const creates a function expression which has a few advantages:

1. block-scoped behavior - meaning the function is only available within the block it's defined in, which gives better control over where the function can be accessed.
2. prevents re-declaration - using const guarantees the function can't be accidentally re-declared or reassigned within the same block, which adds a layer of safety.

Both are valid though!

app/web/features/search/SearchPage.tsx
nabramow: Gotcha sorry about that will leave it to you then, I put it back with a ts-ignore and to-do comment.

app/web/features/search/FilterDialog.tsx
nabramow: Haha I gotcha, TS can be a pain sometimes.

I think some of the linting PRs I just opened will hopefully make it easier not to forget. I was also losing track of mine and only saw after Gitlab ran.

app/web/features/search/SearchBox.tsx
nabramow: You're right, I changed this back to undefined and updated the type to allow it. Sorry about that!

I got a little trigger happy last night...

Rather than making just the location undefined though, I'm returning the locationResult as undefined. That's what you had in the type originally anyway.

I did a lot of testing on the map and all seems to be working okay.

app/web/features/search/SearchBox.tsx
nabramow: Same as comment above, change whole object to undefined initially now.

app/web/features/search/FilterDialog.tsx
```diff
@@ -255,9 +253,7 @@ export default function FilterDialog({
                 id="num-guests-filter"
                 value={numberOfGuestFilter}
                 onChange={(e) => {
-                  setNumberOfGuestFilter(
-                    e.target.value as unknown as undefined
-                  );
+                  setNumberOfGuestFilter(e.target.value);
```
bakeiro: 😮 no ts complains?

app/web/features/search/MapWrapper.tsx
```diff
@@ -284,7 +287,7 @@ export default function MapWrapper({
       </div>
       <Map
         grow
-        initialCenter={locationResult.Location}
+        initialCenter={locationResult?.location}
```
bakeiro: good find

app/web/features/search/SearchBox.tsx
```diff
@@ -55,42 +55,39 @@ export default function SearchBox({
...
-      location: undefined,
-    });
-    setQueryName(undefined);
-  }
+  const handleOnChangeRadioButton = (
+    event: React.ChangeEvent<HTMLInputElement>,
+    value: "location" | "keyword"
+  ) => {
+    setSearchType(value);
+    setLocationResult(undefined);
```
bakeiro: I don't think this is what I wanted, I think I wanted to delete all the values of location but not bbox

app/web/features/search/SearchBox.tsx
```diff
@@ -114,10 +111,7 @@ export default function SearchBox({
                   )}
                   onClick={() => {
                     setQueryName("");
-                    setLocationResult({
-                      ...locationResult,
-                      location: undefined,
-                    });
+                    setLocationResult(undefined);
```
bakeiro: same here

app/web/features/search/FilterDialog.tsx
nabramow: It's because I set the type in the useState variable it references so now it matches :-)

app/web/features/search/SearchBox.tsx
nabramow: What behavior would that look like in the app? 

Since bbox represents a location, when would you want a bbox but not a location?

nabramow: Also on that note, do you know what triggers this particular function? I tried every form field in the modal, but nothing made this function run. I'm wondering if maybe it's old?

nabramow:
Hey @bakeiro when you get a chance could you see if the map works as you expect with the undefined locationResult vs. just location undefined? It looks okay to me but maybe you see something I don't.

The type for `GeocodeResult` doesn't allow the `location` key to be undefined, so whoever wrote that originally didn't intend for it to be undefined. Especially since it's value is in instance of the `LngLat` class, which only accepts non-null numbers, looks like it's not supposed to be undefined.

We could change the type but I guess I just want to understand what you're trying to achieve with the `location: undefined` first.

app/web/features/search/FilterDialog.stories.tsx
```diff
@@ -25,7 +24,7 @@ const Template: Story<any> = () => {
         setHostingStatusFilter={() => {}}
         completeProfileFilter={true}
         setCompleteProfileFilter={() => {}}
-        numberOfGuestFilter={undefined}
+        numberOfGuestFilter={""}
```
bakeiro: why not undefined? I want to be able to have undefined, since it's the default value of the dropdown, plus it's what is sent to the server

app/web/features/search/MapWrapper.tsx
```diff
@@ -68,8 +67,8 @@ interface mapWrapperProps {
     | Pick<User.AsObject, "username" | "userId" | "lng" | "lat">
     | undefined;
   isLoading: boolean;
-  locationResult: any;
-  setLocationResult: Dispatch<SetStateAction<any>>;
+  locationResult: GeocodeResult | undefined;
+  setLocationResult: Dispatch<SetStateAction<GeocodeResult | undefined>>;
```
bakeiro: 👍

app/web/features/search/MapWrapper.tsx
```diff
@@ -78,8 +77,8 @@ interface mapWrapperProps {
     >
   >;
   map: MutableRefObject<MaplibreMap | undefined>;
-  setWasSearchPerformed: any;
-  wasSearchPerformed: any;
+  setWasSearchPerformed: Dispatch<SetStateAction<boolean>>;
+  wasSearchPerformed: boolean;
```
bakeiro: 🔝

app/web/features/search/MapWrapper.tsx
```diff
@@ -123,7 +122,7 @@ export default function MapWrapper({
       const [lng, lat] = geom.coordinates;
       setSelectedResult({ username, userId, lng, lat });
     },
-    []
+    [setSelectedResult]
```
bakeiro: 👍

app/web/features/search/FilterDialog.tsx
```diff
@@ -85,17 +85,17 @@ interface FilterModalFormData
 interface FilterDialogProps {
   isOpen: boolean;
   onClose(): void;
-  queryName: undefined | string;
-  setQueryName: Dispatch<SetStateAction<undefined | string>>;
-  setLocationResult: any;
+  queryName: string;
```
bakeiro, line 88: I get confused here, when there is no query I want to send undefined to the server... should be allowed to have undefined type then?

app/web/features/search/SearchBox.test.tsx
```diff
@@ -13,12 +14,12 @@ const View = ({
...
   const [searchType, setSearchType] = useState(searchTypeParam);
   const [queryNameProp, setQueryNameProp] = useState("");
-  const [locationResultProp, setLocationResultProp] = useState({});
+  const [locationResultProp, setLocationResultProp] = useState<GeocodeResult>();
 
   return (
     <SearchBox
       searchType={searchType}
-      setSearchType={setSearchType as any}
+      setSearchType={setSearchType}
```
bakeiro, line 29: thanks for this (and sorry for my anys!!)

app/web/features/search/SearchBox.tsx
```diff
@@ -36,12 +36,12 @@ const useStyles = makeStyles((theme) => ({
...
-  locationResult: any;
-  setLocationResult: Dispatch<SetStateAction<any>>;
-  setQueryName: Dispatch<SetStateAction<any>>;
-  queryName: undefined | string;
+  searchType: "location" | "keyword";
+  setSearchType: Dispatch<"location" | "keyword">;
+  locationResult: GeocodeResult | undefined;
+  setLocationResult: Dispatch<SetStateAction<GeocodeResult | undefined>>;
+  setQueryName: Dispatch<SetStateAction<string>>;
+  queryName: string;
```
bakeiro, line 45: awesome!

app/web/features/search/SearchBox.tsx
bakeiro: 📝📝📝 good info... will keep it!!

app/web/features/search/SearchBox.tsx
bakeiro: it's the radio button on the form, available on desktop, basically let you choose between option a "location" and b "keyword" basically when user clicks keyword since the input of the location isn't defined I want to delete just the location, but keep the other values

bakeiro: as I suspected, this gets broken with that change 😅 (🤷‍♂️) 
image

bakeiro: don't worry, I make it work plus keepting the new types of this PR, I take care of this

app/web/features/search/SearchPage.test.tsx
```diff
@@ -10,6 +10,7 @@ import SearchPage from "./SearchPage";
 jest.mock("features/userQueries/useCurrentUser");
 jest.mock("maplibre-gl", () => {
   return {
+    ...jest.requireActual("maplibre-gl"),
```
bakeiro: 🙌 🙌

app/web/features/search/SearchBox.tsx
```diff
@@ -55,42 +55,39 @@ export default function SearchBox({
...
+  const handleOnChangeKeyword = (
+    event: React.ChangeEvent<HTMLInputElement>
+  ) => {
     setQueryName(event.target.value);
-    setLocationResult({
-      ...locationResult,
-      location: undefined,
-    });
-  }
+    setLocationResult(undefined);
```
bakeiro: same here, can't set a location to undefined (sorry for the hassle, you know, need to maintain the page behavior 😅 )

bakeiro: don't worry, I will update it to make like before but with the new types

app/web/features/search/SearchPage.tsx
```diff
@@ -73,22 +74,29 @@ export default function SearchPage({
...
+  const [locationResult, setLocationResult] = useState<
+    GeocodeResult | undefined
+  >({
     bbox: bbox,
     isRegion: false,
-    location: { lng: undefined, lat: undefined },
+    location: new LngLat(
+      (parseFloat(bbox[0].toString()) + parseFloat(bbox[2].toString())) / 2,
+      (parseFloat(bbox[1].toString()) + parseFloat(bbox[3].toString())) / 2
+    ), // turns the coordinates into latitude and longitude - means the map will center even if only bbox is provided
```
bakeiro: this is cool, what happens is that in the `mapWrapper` component it has some hardcoded coordinates which are the initial value (independently of the original state value)

that value is used by default, unless there is another value (when a user searches from the dashboard for example) then after the first load the bbox values are loaded, so it has no effect on the values that are loaded by default. (I feel bad typing this comments but it's how the map is wired 😅)

bakeiro: image

app/web/features/search/FilterDialog.stories.tsx
nabramow: I guess my confusion is whoever originally wrote the GeocodeResult type did not allow ‘location’ to be undefined. That’s why you’re having issues with TS, as it’s not allowed to be undefined according to the type.

The question is, did the person who wrote the original type do that for a reason? I assumed so, but maybe not.

In any case I’d say if you decide it should definitely be allowed to be undefined, then update the location key of the Geocode result type to allow undefined. That will make your life easier everywhere else.

app/web/features/search/SearchBox.tsx
nabramow: Sure, feel free to take over this PR. Probably easier than copy pasting and merge conflicts!

app/web/features/search/FilterDialog.tsx
nabramow: Okay looking at this more I think you are right. My original thinking was using "" could help ensure a consistent type for the queryName field. Generally text fields represent being empty by an empty string.

But, when I looked in app/web/service/search.ts, which is processing this search, on line 35 it's checking if `query !== undefined` to decide whether to include it.

Maybe a quirk of GRPC or React Query, I'm just used to it being set up a bit differently.

Feel free to add undefined sorry about that! Still learning this codebase.

app/web/features/search/SearchPage.tsx
nabramow: No worries maps are complicated.

app/web/features/search/FilterDialog.tsx
bakeiro: True, I could use an empty string and then if empty string then send an undefined to the server... that's also right :) [I'm my case I'm more used to undefined = something not set, and "" as empty string (but actual value)

app/web/features/search/FilterDialog.stories.tsx
bakeiro: first sorry for late reply
second, I think we are confusing `numberOfGuestFilter` with Geocoderesult, I'm with you that never should be undefined, I was talking about `numberOfGuestFilter` :)

app/web/features/search/FilterDialog.tsx
bakeiro: actually you are right... it's more stable having 1 type, that 2, makes code more "robust" in my opinion... keeping your change!

bakeiro (commented):
.

app/web/features/search/SearchResultsList.tsx
```diff
@@ -119,21 +119,21 @@ export default function SearchResultsList({
...
 
   let wasResultFound = false;
 
   // @TODO - Can make this one variable by using a find function, resultsList.find, etc.
-  resultsList?.map((value: any) => {
+  resultsList?.map((value) => {
     if (value.user?.userId === selectedResult?.userId) {
       wasResultFound = true;
     }
   });
```
nabramow: I believe you can turn this into one variable like:

```javascript
const wasResultFound = resultsList?.find((value) => value.user?.userId === selectedResult?.userId) !== undefined;
```

It's a bit more performant as `find` will stop as soon as the first value is true:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

app/web/features/search/SearchResultsList.tsx
```diff
@@ -119,21 +119,21 @@ export default function SearchResultsList({
...
 
   if (!wasResultFound && selectedUserData.data) {
-    resultsList = [{ user: selectedUserData.data }];
+    resultsList = [{ user: selectedUserData.data, rank: 0, snippet: "" }];
```
nabramow, line 132: Do we have any filters right now that take `rank` or `snippet` into account? I was also wondering this and if it could sort someone wrong due to that.

app/web/service/search.ts
```diff
@@ -32,7 +32,7 @@ export async function userSearch(
   const req = new UserSearchReq();
   req.setPageToken(pageToken);
 
-  if (query !== "") {
+  if (query !== undefined && query !== "") {
```
nabramow: Looks like you can just do:

```javascript
if (query) {}
```

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

Screenshot 2024-09-17 at 9 29 10 AM

Screenshot 2024-09-17 at 9 29 15 AM

app/web/features/search/SearchResultsList.tsx
```diff
@@ -118,19 +119,21 @@ export default function SearchResultsList({
...
 
   let wasResultFound = false;
-  resultsList?.map((value: any) => {
+
+  // @TODO - Can make this one variable by using a find function, resultsList.find, etc.
+  resultsList?.map((value) => {
     if (value.user?.userId === selectedResult?.userId) {
       wasResultFound = true;
     }
   });
```
nabramow: I believe you can turn this into one variable like:

```javascript
const wasResultFound = resultsList?.find((value) => value.user?.userId === selectedResult?.userId) !== undefined;
```

It's a bit more performant as find will stop as soon as the first value is true:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

app/web/service/search.ts
bakeiro: cool!!!

bakeiro: updated ✅

bakeiro: I've spend to much time learning these things on PHP that I forgot to learn it about JS

app/web/features/search/SearchResultsList.tsx
bakeiro: I think should be fine since this only happens when the result isn't on the list, so we are displaying only one result (basically the first time map is loaded all the users are shown, but due pagination we show only a subset in the list (only on desktop)

this shows the user when the user clicks on it on the map, so list would be only 1 result, so I don't think the hardcoded rank will be a problem :)

app/web/features/search/SearchResultsList.tsx
bakeiro: I guess that instead of `!== undefined` should be `false`?? (if I'm understanding properly)

nabramow: .find will return undefined if nothing passes the given test. Since you want it to return a boolean I added the !== undefined. So if it found something then it’s not equal to undefined it will return true so wasResultFound = true.

If it does return undefined then undefined !== undefined will return false so wasResultFound = false.

Does that make sense? Not sure I explained well.

bakeiro: sure, absolutely clear, will update it now!

bakeiro: done ✅

bakeiro (dismissed):
should be fine now :)

nabramow:
@bakeiro Looks like the prettier format is failing. If you fix that and gitlab is passing happy for you merge!

bakeiro (approved):

bakeiro:
pending testing, do not merge yet

PR #4763, opened 2024-09-09
Change switches to new custom one

nabramow:
This is just some tidying. Now that we have a custom switch that allows custom color and loading state, I switched out our other two switches to the new one. To be honest they don't really need loading state but at least it's consistent. I also made the custom switch more flexible for reuse.

**TO TEST**:

- Go to events page, make two events, one canceled and one not. Make sure the switch still filters.
- Open incognito window (or log out), go to login page. Add a console log in the `LoginForm` handleSubmit function. Make sure the value is changing when clicking the switch and hitting submit.
- Make sure notification settings switches still work.

**Web frontend checklist**
- [ N/ A] Added tests where relevant
- [ ] All tests pass


bakeiro:
that's so cool :D

aapeliv (approved):
LGTM

aapeliv:
I think you can also just call it `Switch`: I think we have that in some other places, e.g. with `<Alert/>`

PR #4764, opened 2024-09-09
Add VSCode recommended extensions for frontend and format on save

### LINKED ISSUE #4785: Improvements to linting and dev experience
- Require passing linting in Gitlab action (#4766)
- Add eslint and prettier to VSCode recommended extensions so it pops up. This improves DevEx by highlighting linting errors before pushing to the branch. (#4764)
- Format and lint on save. When you save in VSCode (command + s) it will automatically format and autofix linting (where possible). This means you don't always need to run the linting script for basic stuff. (#4764)

nabramow:
Right now you can only see the linting errors if you run a script, which makes it easy to forget and merge with the errors.

If you install the eslint extension, it will highlight the the linting errors in the code directly, making them more obvious.

We have eslint already as a dev dependency, but adding it to the recommended extensions will remind people developing on the frontend to install the eslint plugin to VSCode. I added prettier too.

This will provide a better developer experience so it's easier to see the linting errors as you write code, rather than having to specifically run a script.

To do this I had to remove ".vscode" from ".gitignore". I assume this was copy pasta originally as it's not usually gitignored.

Looks like this, a popup when you open VSCode  couchers repo and listed under "Recommended Extensions" in the extensions tab:

Screenshot 2024-09-08 at 10 55 13 PM


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.formatOnSave": true
+}
```
nabramow, line 3: This will automatically format code based on the prettier standards whenever you save in VSCode. Seems easier than always having to manually run the script.

darrenvong (approved):
LGTM!

bakeiro:
🔝

PR #4765, opened 2024-09-09
Add linting to web gitlab checks

nabramow:
We have prettier as a gitlab check on the branch for frontend but not linting, means it's possible to merge a branch with linting errors.

Here's my attempt to add it, feel free to yell at me if there's a reason I was unaware of that it's not there.


nabramow:
Closing as I have another with the same code open, was just tryin to understand if the branch names effecting which gitlab tests ran.

PR #4766, opened 2024-09-09
Add gitlab check for frontend linting

### LINKED ISSUE #4785: Improvements to linting and dev experience

nabramow:
We have prettier as a gitlab check on the branch for frontend but not linting, means it's possible to merge a branch with linting errors.

Here's my attempt to add it, feel free to yell at me if there's a reason I was unaware of that it's not there.

Screenshot 2024-09-08 at 10 22 51 PM

### DIFF
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -530,6 +530,9 @@ wait:before-release:
     - job: test:web-prettier
       artifacts: false
       optional: true
+    - job: test:web-linting
+      artifacts: false
+      optional: true
     - job: test:storybook
       artifacts: false
       optional: true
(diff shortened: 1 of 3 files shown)


nabramow:
I see my new gitlab test here so I think that means it worked?

https://gitlab.com/couchers/couchers/-/pipelines/1445183095

aapeliv (approved):
Looking good!

PR #4767, opened 2024-09-09
Add more metrics based on suggestions by Yannic

### LINKED ISSUE #4740: More metrics
from Yannic:
- male/female ratio over time or just in general
- ~Active users within last 3 months / 6 months / year~
- ~percentage of "can host"/"can't host"/"maybe"~
- distribution pf hosting request stati
- ~percentage of users that have a full profile~
- ~percentage of users that have at least one reference~
- ~percentage of users that have written at least one message~
- ~percentage of users that have sent a host request~
- percentage of users that have joined an event

aapeliv:
Closes #4740

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/metrics.py
+++ b/app/backend/src/couchers/metrics.py
@@ -6,7 +6,7 @@ from prometheus_client.registry import CollectorRegistry
 from sqlalchemy.sql import func
 
 from couchers.db import session_scope
-from couchers.models import User
+from couchers.models import EventOccurrenceAttendee, HostingStatus, HostRequest, Message, Reference, User
 from couchers.sql import couchers_select as select
 from couchers.utils import now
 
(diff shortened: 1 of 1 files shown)


app/backend/src/couchers/metrics.py
```diff
@@ -39,35 +39,143 @@ def observe_in_servicer_duration_histogram(method, user_id, status_code, excepti
...
+    (
+        select(func.count()).select_from(
+            select(User.id)
+            .where(User.is_visible)
+            .join(HostRequest, HostRequest.surfer_user_id == User.id)
+            .group_by(User.id)
+            .subquery()
+        )
+    ),
+)
```
spreeni, line 150: I guess for both it would be interesting to have the opposite side to see if we have issues with superhosts (which I reckon we do)

app/backend/src/couchers/metrics.py
```diff
@@ -39,35 +39,143 @@ def observe_in_servicer_duration_histogram(method, user_id, status_code, excepti
...
 
-def _get_users():
-    with session_scope() as session:
-        return session.execute(select(func.count()).select_from(User).where(User.is_visible)).scalar_one()
+woman_gauge = _make_gauge_from_query(
+    "couchers_users_woman",
+    "Total number of users with gender 'Woman'",
+    select(func.count()).select_from(User).where(User.is_visible).where(User.gender == "Woman"),
+)
```
spreeni, line 92: Here an additional one for the rate between genders, or a normalized growth index could be nice

spreeni (commented):
Looks good! When are these changes deployed each time?

I think it would be good to have one dynamic dashboard with real-time developments and then one more static one with overall statistics. Right now some of the graphs are a bit overwhelming.

app/backend/src/couchers/metrics.py
aapeliv: You can compute ratios etc in grafana!

PR #4779, opened 2024-09-10
Web push notifications

nabramow:
This PR listens for the push event from the backend and displays the notifications for different actions.

Right now this will only work in Chrome, Firefox or other browsers that support the web push API (not safari, but mostly everything else).

Screenshot 2024-09-10 at 9 54 15 PM

Screenshot 2024-09-10 at 9 54 23 PM

Screenshot 2024-09-11 at 9 16 42 PM

**HOW TO TEST:**

- You need to run both front and backend locally (so push notifications are sent from the same environment they are received)
- Generate protos locally if needed
- Don't use Safari, Chrome works well
- Go to Feature Preview and allow notifications there.
- I decided to test messaging push notifications, you can do something else. To test messaging notifications you need two couchers accounts. Log in to one in the regular browser, the other in an incognito window.
- Message yourself from the incognito window to the other (push notifications don't work in incognito mode)
- Check non-incognito window, push notification should show.
- Reloading the page also triggers a push notification, if you are lazy.

**Web frontend checklist**
- [ x ] All tests pass


app/backend/src/couchers/interceptors.py
```diff
@@ -399,7 +399,7 @@ def sanitizing_function(req, context):
                     logger.info("Probably an unknown error! Sanitizing...")
                     context.abort(grpc.StatusCode.INTERNAL, errors.UNKNOWN_ERROR)
                 else:
-                    logger.warning(f"RPC error: {code}")
+                    logger.warning(f"RPC error: {code} {handler_call_details.method}")
```
nabramow: Is it okay if I leave this @aapeliv ? It was helpful.

aapeliv: Yes but please change to

```logger.warning(f"RPC error: {code} in method {handler_call_details.method}")```

File: app/web/pages/_app.tsx [hunk 2]
```diff
@@ -43,15 +49,82 @@ function MyApp({ Component, pageProps }: AppWithLayoutProps) {
...
+          registration.scope
+        );
+
+        const existingPushSubscription =
+          await registration.pushManager.getSubscription();
+        const p256dhKey = existingPushSubscription?.getKey("p256dh");
+
+        if (existingPushSubscription && p256dhKey) {
+          const publicKey = arrayBufferToBase64(p256dhKey);
+          if (publicKey !== process.env.NEXT_PUBLIC_VAPID_PUBLIC_KEY) {
```
aapeliv: Should use `getVapidPublicKey()`?

app/web/pages/_app.tsx
```diff
@@ -43,15 +49,82 @@ function MyApp({ Component, pageProps }: AppWithLayoutProps) {
...
+        }
+
+        const { vapidPublicKey } = await getVapidPublicKey();
+        const subscription: PushSubscription =
+          await registration.pushManager.subscribe({
+            userVisibleOnly: true,
+            applicationServerKey: vapidPublicKey,
+          });
+
+        await registerPushNotification(subscription);
```
aapeliv: I think we should rename this function, right now it makes no sense. How about `regsiterPushNotificationSubscription`?

aapeliv:
So this will right now show the notifications *on the page*, right?

There is also [this method](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification) that will actually show a native notification like this:

IMG_4124

It should also work [cross browser](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification#browser_compatibility) with some (major) caveats e.g. on iOS.

I think this needs to be done in the service worker itself (since it'll be woken up and run in the background when the notification itself gets received).

app/web/pages/_app.tsx
nabramow: Oops missed that one. Fixed!

nabramow:
> So this will right now show the notifications _on the page_, right?
> 
> There is also [this method](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification) that will actually show a native notification like this:
> 
> IMG_4124
> 
> It should also work [cross browser](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/showNotification#browser_compatibility) with some (major) caveats e.g. on iOS.
> 
> I think this needs to be done in the service worker itself (since it'll be woken up and run in the background when the notification itself gets received).

This is fixed now!

app/web/utils/arrayBufferToBase64.ts
```diff
@@ -0,0 +1,9 @@
+export const arrayBufferToBase64 = (buffer: ArrayBuffer) => {
```
aapeliv, line 1: This is no longer needed, right?

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+export default function PushNotificationPermission({
+  className,
+}: {
+  className: string;
+}) {
+  const { t } = useTranslation(AUTH);
+  const classes = useStyles();
+
+  const [permission, setPermission] = useState(Notification.permission);
```
aapeliv: Won't this only set the initial state to whatever `Notification.permission` is when this component is mounted(?) ?

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,161 @@
...
+          });
+
+        await registerPushNotificationSubscription(subscription);
+      } else {
+        setErrorMessage(
+          t("notification_settings.push_notifications.error_unsupported")
+        );
+      }
+    } catch (error) {
+      setErrorMessage(
```
aapeliv: Should we send it over to sentry as well or something?

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,161 @@
...
+  };
+
+  const turnPushNotificationsOn = async () => {
+    if (permission === "default") {
+      const result = await Notification.requestPermission();
+      setPermission(result);
+
+      if (result === "granted") {
+        await onPermissionGranted();
+      }
```
aapeliv: What if I already granted permission (so the browser can show notifications), but I turned it off and want to re-turn on?

Then I think you should have `permission` equal to `granted` above and this function takes the error path instead?

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,161 @@
...
+      "/service-worker.js",
+      { scope: "/" }
+    );
+    const existingPushSubscription =
+      await registration.pushManager.getSubscription();
+
+    if (existingPushSubscription) {
+      await existingPushSubscription.unsubscribe();
+    }
+    setPermission("default");
```
aapeliv: Ohhh. I see. But now this is not actually what the `Notification.permission` setting would be?

If you re-sub you don't need to request again. But I guess `Notification.requestPermission` doesn't ask for perms again if it already has it, so that itself doesn't matter.

app/web/utils/arrayBufferToBase64.ts
aapeliv: OK scratch that I saw it is still used.

aapeliv (commented):
Just the one minor thing about the `permission` value not being actually synced to `Notification.permission`.

app/web/utils/arrayBufferToBase64.ts
nabramow: Ah yeah sorry I meant to leave a comment for you. I took it out but it didn't work then so I put it back.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: So the value of Notification.permission is stored in the browser's internal state. But you're right, if I deny it right now the switch won't enable it again. 

This is because the `Notification.requestPermission()` method only works when the permission status is "default" (i.e., when the user hasn't already made a choice). If the permission status is "denied", the browser will not ask for permission again automatically (unfortunately for us). Once the user denies the permission, it is stored at the browser level, and the browser does not allow the web app to prompt the user again.


As a work-around, I just added a check that will determine what browser and OS the user is using and show an error message specifically with instructions for how to enable notifications again for that browser.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Done!

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: I'm not totally following this.

I think maybe you're saying there should be a has permission but turned off push notifications state, but right now there's just the `Notification.permission` state?

I mean, I guess we could do that but we'd need to either store the user settings in `localStorage`, on the user object in the DB or something like that.

Do you mean you don't want turning the switch off to equal permission denied (if they granted it first then turned it off after)?

nabramow: Hmm so when a user unsubscribes from push notifications, they need to manually reset the permission through the browser settings to bring it back to default. There's no programmatic way to automatically set the permission back to default after unsubscribing unfortunately.

So only way around with would be to persist some kinda "push turned off" value if they ever enabled them.

I guess this would be either storing some user settings in that user's row in the database, or `localStorage` or something like that.

Do you know offhand how y'all persist user info like that? Is that with `usePersistedState`?

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Actually I think I am setting it to "default" but really it would be "denied".

But yes it won't ask again, I'll update the `setPermission` value.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Ugh okay after reading about this more I realized unsubscribing the user from push notifications with `.unsubscribe()` won't update `Notification.permission`.

Basically granting permission is a one way street and disabling it or reactivating again has to be done via the browser.

So I think the way to go is just have the switch to enable them, and give instructions for how to disable or reenable in the browser for the other scenarios. Annoying.

aapeliv: I mean once we have permission, then we don't need to worry about that anymore: we'll just subscribe/unsubscribe the user, right?

app/web/features/auth/notifications/PushNotificationPermission.tsx
aapeliv: Ohhhh, I see what you are saying. I mean in order to know if the user is currently subscribed to push notifs on this browser, I suppose we just have to check if they have a subscription on the service worker, right?

app/backend/src/couchers/interceptors.py
```diff
@@ -399,7 +399,7 @@ def sanitizing_function(req, context):
                     logger.info("Probably an unknown error! Sanitizing...")
                     context.abort(grpc.StatusCode.INTERNAL, errors.UNKNOWN_ERROR)
                 else:
-                    logger.warning(f"RPC error: {code}")
+                    logger.warning(f"RPC error: {code} in method {handler_call_details.method}")
```
bakeiro, line 402: 🔝

app/backend/src/couchers/servicers/notifications.py
```diff
@@ -102,25 +103,25 @@ def GetVapidPublicKey(self, request, context):
...
+                endpoint=data["endpoint"],
+                p256dh_key=decode_key(data["keys"]["p256dh"]),
+                auth_key=decode_key(data["keys"]["auth"]),
                 full_subscription_info=request.full_subscription_json,
                 user_agent=request.user_agent,
             )
             session.add(subscription)
             session.flush()
-            sub_id = subscription.id
             push_to_subscription(
```
bakeiro, line 122: how this works exactly? it's an http sent to the frontend and the frontend it's listening? (through the service worker?)

or is this some websocket connection? (just to understand what's going on)

app/backend/src/couchers/servicers/notifications.py
```diff
@@ -102,25 +103,25 @@ def GetVapidPublicKey(self, request, context):
 
         return notifications_pb2.GetVapidPublicKeyRes(vapid_public_key=get_vapid_public_key())
 
-    def RegisterPushNotification(self, request, context):
+    def RegisterPushNotificationSubscription(self, request, context):
```
bakeiro: how this `RegisterPushNotificationSubscription` function works exactly? it sends an http sent to the frontend and the frontend it's listening? (through the service worker?)

or is this some websocket connection? (just to understand what's going on)

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+export default function PushNotificationPermission({
+  className,
+}: {
+  className: string;
+}) {
+  const { t } = useTranslation(AUTH);
+  const classes = useStyles();
+
+  const [permission, setPermission] = useState(Notification.permission);
```
bakeiro, line 47: another question...

about this browser stored variable (in this case `Notification.permission`) does this trigger re-render when the user changes it? (just to know...)

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,236 @@
...
+        // Subscribe to push notifications using the PushManager API
+        const subscription: PushSubscription =
+          await registration.pushManager.subscribe({
+            userVisibleOnly: true,
+            applicationServerKey: vapidPublicKey,
+          });
+
+        await registerPushNotificationSubscription(subscription);
+      } else {
+        setErrorMessage(
```
bakeiro: register the used browser here using Sentry for metrics? (in case it makes sense)

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,236 @@
...
+          "/service-worker.js",
+          { scope: "/" }
+        );
+
+        const existingPushSubscription =
+          await registration.pushManager.getSubscription();
+        const p256dhKey = existingPushSubscription?.getKey("p256dh");
+        const { vapidPublicKey } = await getVapidPublicKey();
+
+        if (existingPushSubscription && p256dhKey) {
```
bakeiro: edge case here (ignore if you feel isn't important, maybe isn't worth it)

what if `existingPushSubscription` is true and `p256dhKey` is false (is that possible?)
if so, then `registration.pushManager.subscribe` could be re-subscribing? (in case it's a bad thing)

bakeiro:
question, does this works on Android??

bakeiro:
this PR it's so cool Nicole
apart of the `VAPID` and `p256dh` which I don't understand so I trust you there, the rest seems OK to me :)

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,236 @@
+import { Typography } from "@material-ui/core";
```
bakeiro: makes sense to create two components?

- one would be the frontside component (used in the feature preview)
- other one for the internal logic, which would be rendered (I guess) in every page, like wrapping the app, or rendered in the footer (or wherever) and contains the service-worker stuff

nabramow: Why would we want the logic on every page? You might be thinking of toast notification, which would run within the web app (and thus makes sense wrapping the app with a NotificationContext or something). Since these are system notifications, they are handled in the `service-worker.js` file totally outside of our web app.

I think might be an over-optimization at this point since we don't know where we will move it yet, or what we will use to trigger the permission request. IMO probably makes sense we move the buttons to the new notification settings page?

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: This is more a security check. I left a comment a couple lines below this explaining, but basically it's checking if we're looking at the right push notification subscription. If we have an existing push notification subscription, we want to get its key and compare it with the one we have form the backend. If the keys don't match we want to unsubscribe because we're looking at the wrong push notification subscription.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: This wouldn't be our error though, just if the user's browser doesn't support push notifications and/or service workers. Would we usually log that kinda thing with sentry?

I don't understand what we use sentry for yet haha.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Okay gonna resolve this since you are right and we talked about it on a zoom call. I'm gonna refactor this to base itself on the subscription being null or not rather than the permission.

app/backend/src/couchers/servicers/notifications.py
nabramow: This is exposing the backend grpc function so I can tell the backend that push notifications are registered once they subscribe via the buttons on the frontend.

It will insert a row into the `push_notification_subscriptions` table, I believe. Aapeli did all the backend though so I'm a little iffy. He just renamed this function and pushed it to my branch.

It's not a web socket because that would only send push notifications while the app was open. The cool thing about using a service worker is it can tell the user they got a new message even if they are not on our website at that moment, so it reminds them to go back to the website and check their message. It's kinda like we're running in circles around our app yelling things at it lol.

nabramow:
> question, does this works on Android??

It's a good question but I own zero Android devices lol. I think once the PR is totally ready we should find someone with Android to test it. Theoretically it should work!

aapeliv:
> > question, does this works on Android??
> 
> It's a good question but I own zero Android devices lol. I think once the PR is totally ready we should find someone with Android to test it. Theoretically it should work!

Yes, web push works: Jesse tested it via console!

aapeliv:
@nabramow do you want me to push a quick commit to fix the protos/backend stuff?

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Okay this is done now!

nabramow:
Okaaaaay, this is finally ready to go this time!

bakeiro (dismissed):
👍

app/backend/src/couchers/servicers/notifications.py
bakeiro: oh wow, that's cooler than what I though! I was expecting notifications only when the couchers tab was open, that's cool, good to know!! (now I understand why this instead of plain toast messages lol)

app/web/features/auth/notifications/PushNotificationPermission.tsx
bakeiro: yeah makes sense...

about moving it to the notification settings page isn't up to me, so I can't decide but makes sense though, for me makes sense to keep it in the feature preview page for now and move it there once we are ready

app/web/features/auth/FeaturePreview.tsx
```diff
@@ -45,6 +46,7 @@ export default function FeaturePreview() {
         <Alert severity="error">{accountInfoError.message}</Alert>
       ) : (
         <>
+          <PushNotificationPermission className={classes.section} />
```
aapeliv: Should this be called `PushNotificationSettings` or something? It's not just the permission anymore.

app/web/features/auth/locales/en.json
```diff
@@ -280,6 +280,30 @@
...
+      "title": "Push Notifications",
+      "description": "You can enable push notifications for Couchers.org on your device. We are working on expanding this feature.",
+      "allow_push": "Click the 'Allow' button in your browser popup to enable push notifications.",
+      "enabled_message": "You currently have push notifications <1>enabled</1>.",
+      "disabled_message": "You currently have push notifications <1>disabled</1>.",
+      "error_generic": "Error during subscription to push notifications",
+      "error_unsupported": "Push notifications or Service Workers are not supported in this browser.",
+      "permission_denied": {
+        "platform_settings_description": "Still having issues after enabling notifications in your browser? You may need to adjust your platform settings.",
+        "instructions": {
```
aapeliv, line 294: We will also need to add instructions for iOS (which requires pinning the app to the home screen): I think it returns "denied" without popping up the modal at all in general.

But I think we can leave that for later, I'd like to get this merged in soon :)

app/web/features/auth/notifications/PushNotificationDenied.tsx
```diff
@@ -0,0 +1,74 @@
...
+  },
+}));
+
+const PushNotificationDenied = () => {
+  const { t } = useTranslation(AUTH);
+  const userAgent = navigator.userAgent.toLowerCase();
+  const classes = useStyles();
+
+  const getBrowserInstructions = () => {
+    if (userAgent.includes("chrome")) {
```
aapeliv, line 29: I don't think this actually tells you the truth... Everything pretends for instance to be compatible with firefox. Again, can be sorted out later.

app/web/features/auth/notifications/PushNotificationDenied.tsx
```diff
@@ -0,0 +1,74 @@
...
+      );
+    } else if (userAgent.includes("linux")) {
+      return t(
+        "notification_settings.push_notifications.permission_denied.instructions.linux.gnome"
+      );
+    }
+
+    return t(
+      "notification_settings.push_notifications.permission_denied.instructions.generic",
+      "Please check your OS settings for notification control."
```
aapeliv: What does the second argument do?

app/web/features/auth/notifications/notificationUtils.ts
```diff
@@ -0,0 +1,16 @@
+const getCurrentSubscription = async () => {
+  let registration = await navigator.serviceWorker.getRegistration();
```
aapeliv, line 2: Do you need to check if `serviceWorker` is in `navigator` here?

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,193 @@
...
+    const checkPushEnabled = async () => {
+      const existingPushSubscription = await getCurrentSubscription();
+      setIsPushEnabled(
+        Notification.permission === "granted" &&
+          existingPushSubscription !== null
+      );
+    };
+
+    checkPushEnabled();
+    setIsLoading(false);
```
aapeliv: This will set loading to false before the async function above completes, right? So this should be moved to the end of `checkPushEnabled`

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,193 @@
...
+}: {
+  className: string;
+}) {
+  const { t } = useTranslation(AUTH);
+  const classes = useStyles();
+
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [isPushEnabled, setIsPushEnabled] = useState<boolean>(false);
+  const [isLoading, setIsLoading] = useState<boolean>(true);
+  const [shouldPromptAllow, setShouldPromptAllow] = useState<boolean>(false);
```
aapeliv: Could you add a comment saying "whether to show the user instructions to click 'Allow' in their browser"

app/web/features/auth/notifications/PushNotificationPermission.tsx
```diff
@@ -0,0 +1,193 @@
...
+    if (Notification.permission !== "denied") {
+      setIsLoading(true);
+      setShouldPromptAllow(true);
+      const result = await Notification.requestPermission();
+      setShouldPromptAllow(false);
+
+      if (result === "granted") {
+        await onPermissionGranted();
+        setIsPushEnabled(true);
+      }
```
aapeliv: Should this have an else with `setIsPushEnabled(false)`?

aapeliv (dismissed):
I added a few small comments, mostly future todos, so ignore most of them :)

Feel free to merge, we can clean up the rest later!

app/web/features/auth/FeaturePreview.tsx
nabramow: Changed the name now!

app/web/features/auth/locales/en.json
nabramow: I added a quick `isMobile` check and a generic message for now. Can go back and add better ones later.

app/web/features/auth/notifications/PushNotificationDenied.tsx
nabramow: I was using Chrome and it was working for me, but it's true I haven't tried all browsers. Can make a ticket to thoroughly test them for different ones.

app/web/features/auth/notifications/PushNotificationDenied.tsx
nabramow: Oops removed.

app/web/features/auth/notifications/notificationUtils.ts
nabramow: Ah yeah that can't hurt. Just added.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Ah true because I'm not awaiting it, good catch.

app/web/features/auth/notifications/PushNotificationPermission.tsx
nabramow: Yeah makes sense.

app/web/features/auth/notifications/PushNotificationDenied.tsx
aapeliv: Yeah we can probably use some lib, but yeah let's add an issue and do it later!

aapeliv (dismissed):
LGTM

aapeliv (approved):

PR #4793, opened 2024-09-11
Show all users map

bakeiro:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/search/MapWrapper.tsx
+++ b/app/web/features/search/MapWrapper.tsx
@@ -149,7 +149,6 @@ export default function MapWrapper({
     map.current?.stop();
     map.current?.easeTo({
       center: [user.lng, user.lat],
-      // zoom: 14
     });
   }, []);
 
--- a/app/web/features/search/MapWrapper.tsx
+++ b/app/web/features/search/MapWrapper.tsx
@@ -145,7 +145,7 @@ export default function MapWrapper({
   /**
    * Moves map to selected user's location
    */
-  const flyToUser = useCallback((user: Pick<User.AsObject, "lng" | "lat">, zoom = 0) => {
+  const flyToUser = useCallback((user: Pick<User.AsObject, "lng" | "lat">) => {
     map.current?.stop();
     map.current?.easeTo({
       center: [user.lng, user.lat],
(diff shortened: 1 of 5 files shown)


app/web/features/search/MapWrapper.tsx
```diff
@@ -77,6 +78,8 @@ interface mapWrapperProps {
     >
   >;
   map: MutableRefObject<MaplibreMap | undefined>;
+  setWasSearchPerformed: any;
+  wasSearchPerformed: any;
```
bakeiro, line 82: I know I know `any` is bad

app/web/features/search/SearchPage.tsx
```diff
@@ -84,14 +85,14 @@ export default function SearchPage({
   const [lastActiveFilter, setLastActiveFilter] = useState(0);
   const [hostingStatusFilter, setHostingStatusFilter] = useState(0);
   const [numberOfGuestFilter, setNumberOfGuestFilter] = useState(undefined);
-  const [completeProfileFilter, setCompleteProfileFilter] = useState(true);
+  const [completeProfileFilter, setCompleteProfileFilter] = useState(false);
```
bakeiro, line 88: if this is enabled by default to many profiles disapear, since it's a good 'selling point' show a lot of users in our platform maybe we want to disable this by default?

app/web/features/search/SearchResultsList.tsx
```diff
@@ -147,34 +162,33 @@ export default function SearchResultsList({
             breakpoint="sm"
             className={classes.scroller}
             isFetching={isLoading}
-            fetchNext={fetchNextPage}
+            // fetchNext={fetchNextPage} // TODO: disabled for now (until pagination)
```
bakeiro, line 168: let me disable this (which honestly I doubt that any user hardy uses it) and map is fixed :D

bakeiro: also how buggy was the map in production all this time, since how it right now after this update, this (IMHO) seems the like feature least important to keep (since it's totally non user friendly feature)

aapeliv:
`app/backend/src/couchers/servicers/search.py#L53`

aapeliv (approved):
LGTM

PR #4806, opened 2024-09-15
Add recommended python extension and inoffensive settings for vscode/vscodium

aapeliv:


.vscode/extensions.json
```diff
@@ -1,3 +1,7 @@
 {
-  "recommendations": ["dbaeumer.vscode-eslint", "esbenp.prettier-vscode"]
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "esbenp.prettier-vscode",
+    "charliermarsh.ruff"
```
nabramow, line 5: Do you use VSCode for Python?

aapeliv: Ah yes, maybe I should add that? I kind of thought it was installed by default.

.vscode/settings.json
```diff
@@ -8,5 +8,12 @@
...
     "typescript",
     "typescriptreact"
-  ]
+  ],
+  "files.insertFinalNewline": true,
+  "[python]": {
+    "editor.tabSize": 4,
+    "editor.rulers": [
+      120
+    ],
```
nabramow, line 17: I think a better place for this might be the linting rules? Since this will only enforce with VSCode, you could have the situation where people have different settings in different IDEs, better have it be repo-wide.

aapeliv: It is enforced by the linter, but just makes it nicer when in VScode, in particular the ruler

.vscode/extensions.json
nabramow: Ah just I know people tend to use other IDEs for Python so wanted to make sure you knew this would only work for VSCode :).

.vscode/settings.json
nabramow: Or rather whatever Prettier is for python, haha. It's more formatting than linting!

.vscode/extensions.json
nabramow: Wait add what?

aapeliv: Ah I thought you meant the default Python VScode extension, I think it has a name like "Python for VScode", or similar.

Anyway, `ruff` is the linter/formatter we use.

.vscode/settings.json
aapeliv: Yes, that's `ruff`, and it's enforced by CI as well

aapeliv:
@nabramow I've had another look at this and I think it's good to merge. I think I addressed your comments, but let me know if not! Maybe easier to discuss on Slack if you'd like to discuss.

nabramow (approved):

PR #4822, opened 2024-09-16
Implement login (session) management in frontend

aapeliv:
Screenshot from 2024-09-16 00-08-58

**Web frontend checklist**
- [ ] All tests pass


nabramow:
Not to poop on this party, but does it make more sense to just automatically log out any existing logins as part of the login process instead? So when you login, it logs you out of any other existing logins for your account.

Is there a use case for having multiple active logins?

Seems more secure that way, for example if someone used a public computer or internet cafe, etc. (do people still use internet cafes?).

aapeliv:
> Not to poop on this party, but does it make more sense to just automatically log out any existing logins as part of the login process instead? So when you login, it logs you out of any other existing logins for your account.
> 
> Is there a use case for having multiple active logins?

I'm constantly logged in on: phone, laptop, desktop!

nabramow:
> > Not to poop on this party, but does it make more sense to just automatically log out any existing logins as part of the login process instead? So when you login, it logs you out of any other existing logins for your account.
> > Is there a use case for having multiple active logins?
> 
> I'm constantly logged in on: phone, laptop, desktop!

Ah gotcha that makes sense.

app/web/features/auth/logins/LoginsPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const sessions = data?.pages.flatMap((page) => page.activeSessionsList);
+
+  const {
+    error: logoutAllError,
+    isLoading: logoutAllIsLoading,
+    mutate: logoutAll,
+  } = useMutation<void, RpcError>(async () => {
+    await service.account.logOutOtherSessions(true);
+    refetch();
```
nabramow: Instead of using `refetch` to force a refetch and passing it in the props, better put the value that should trigger the refetch in the `useInfiniteQuery` dependencies array on line 50 above

I think probably it's not fetching as you have the syntax wrong on line 50 above. It should be `queryKey: activeLoginsKey`.

app/web/features/auth/logins/LoginsPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+    isFetchingNextPage,
+    refetch,
+  } = useInfiniteQuery<ListActiveSessionsRes.AsObject, RpcError>(
+    activeLoginsKey,
+    ({ pageParam }) => service.account.listActiveSessions(pageParam),
+    {
+      getNextPageParam: (lastPage) =>
+        lastPage.nextPageToken ? lastPage.nextPageToken : undefined,
+    }
+  );
```
nabramow:
```suggestion
  } = useInfiniteQuery<ListActiveSessionsRes.AsObject, RpcError>({
    queryKey: activeLoginsKey,
    queryFn: ({ pageParam }) => service.account.listActiveSessions(pageParam),
    getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
  });
```

nabramow: That should hopefully fix it so you don't need `refetch`.

app/web/features/auth/logins/LoginsPage.tsx
```diff
@@ -0,0 +1,112 @@
...
+        {t("auth:active_logins.description")}
+      </Typography>
+      {error && <Alert severity="error">{error.message}</Alert>}
+      {logoutAllError && (
+        <Alert severity="error">{logoutAllError?.message}</Alert>
+      )}
+      {isLoading ? (
+        <CircularProgress />
+      ) : (
+        sessions!.map((session) => (
```
nabramow: The bang operator `!` isn't actually true here since `sessions` could be undefined I believe. On line 58, better add `|| []` to the end of the `sessions` variable, that will set it to an empty array if the value is undefined. Then remove the `!` here. That way you're actually accounting for undefined.

app/web/features/auth/logins/LoginsPage.tsx
nabramow: Just following up on this as I think you tagged me again @aapeliv ? This comment was the big one here, to add the queryKey instead of refetch.

aapeliv: I've changed it to not use refetch, but I'm invalidating the queries wherever that changes. I'm not sure if I misunderstood what you meant, if there was another way!

nabramow:
@aapeliv FYI I pushed the button style changes, apologies I thought you meant you wanted to refactor the Button component itself!

Besides that merged develop branch as the backend was failing and I wanted to test it!

LGTM!

nabramow (approved):

Issue #4799, opened 2024-09-14
Filtering options


CouchersBot:
Subject: Filtering options
Description:
Not a bug, just a filtering option suggestion.

Results:
The new map search is better! But the filtering options (any, can host, may host, can't host) for searching for a host should allow people to look at "can host" and "may host" at the same time, since they're essentially the same thing: "might host if your message is nice". I understand the difference could reflect a level of hosting enthusiasm, and that's fine, let them choose from 1-10 if you wanna get wild, but when searching for a host it would be nice to see all the possibilities without toggling back and forth.

Design-wise, you could eliminate the "any" option and change the remaining three (can, may, can't) to checkboxes/toggles, all enabled by default (any).

Love the improvement overall, keep it up. Good times.

Backend version: develop-033299bf
Frontend version: develop-033299bf
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36
Page: https://couchers.org/search#loc=5.52/51.058/8.995
User: kenflannery (8864)

nabramow:
@bakeiro this looks like it's done, is that right?

bakeiro:
yes! @nabramow closing now

PR #4853, opened 2024-09-19
Filter option

bakeiro:
before:
image

now:
image

closes:
https://github.com/Couchers-org/couchers/issues/4799

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [x] All tests pass


app/web/components/Select.tsx
```diff
@@ -31,10 +35,12 @@ export default function Select<T extends Record<string | number, string>>({
 }: Omit<SelectProps, "children"> & {
   id: string;
   options: Extract<keyof T, string | number>[];
-  value?: T extends undefined ? string | number : keyof T;
+  // value?: T extends undefined ? string | number : keyof T;
```
bakeiro, line 37: @nabramow feedback here pls :) (basically this was overwriting the type making it extra strict, I though removing it was better

app/web/components/Select.tsx
```diff
@@ -44,7 +50,7 @@ export default function Select<T extends Record<string | number, string>>({
     >
       <InputLabel htmlFor={id}>{label}</InputLabel>
       <MuiSelect
-        native
+        native={native}
```
bakeiro, line 54: user can now choose

app/web/features/search/FilterDialog.tsx
```diff
@@ -89,9 +116,9 @@ interface FilterDialogProps {
   setQueryName: Dispatch<SetStateAction<undefined | string>>;
   setLocationResult: any;
   lastActiveFilter: number;
-  setLastActiveFilter: Dispatch<SetStateAction<number>>;
-  hostingStatusFilter: number;
-  setHostingStatusFilter: Dispatch<SetStateAction<number>>;
+  setLastActiveFilter: Dispatch<SetStateAction<number[]>>;
+  hostingStatusFilter: Array<keyof typeof HostingStatusValues>;
```
bakeiro, line 119: typing is cool 🤓

app/web/features/search/FilterDialog.tsx
```diff
@@ -213,15 +240,26 @@ export default function FilterDialog({
               />
 
               <Select
-                id="can_host_status_filter"
+                id="can_host_status_filter_1"
+                variant="outlined"
+                label={"Hosting status"}
```
bakeiro: upsi, just realized this hardcoded label

app/web/components/Select.tsx
nabramow: What about `value?: string | number`.

Take a look at the type for the `Select` on line 24 you can see it says `Record<string | number, string>` which gives a clue.

Then I searched `<Select` in VSCode and checked the values of all the Selects being used and they all seem to be a string or number.

Obviously if we ever want to use something besides a string or number we'll need to update it but I think it's good to start strict and expand as we need it.

nabramow: Ooo this is a good question let me think about it. I think we want to make the Select itself more flexible since it's reusable but then the values in the components that use the Select more specific. I'll get back to you tomorrow on this about to board a flight. We should give it a type I just gotta think about which.

nabramow: Okay I got a minute already haha. So if you want to refactor this, I'd do something like this:

https://mariusschulz.com/blog/passing-generics-to-jsx-elements-in-typescript

Keep in mind if you change the value you also need to update the type of line 24 of the `Select`.

I checked the `Select` values and we are always using a `string | number` pretty much, you get a clue of that also on line 24.

I'd say make an `OptionValue` type like in the example above and use that. I think you can do it almost exactly as in that guide ^^.

It's going to cause errors for other types that are any or vague in other components like `FilterDialog`, but a lot of those should be fixed once the linting PR is merged.

app/web/features/search/FilterDialog.tsx
nabramow: Ah yeah probably want a translation there instead. Looks like there are already some if they apply here if you search "Hosting status" in VSCode.

app/web/features/search/FilterDialog.tsx
```diff
@@ -198,7 +225,7 @@ export default function FilterDialog({
                 id="last_active_filter"
                 className={classes.marginBottom}
                 value={lastActiveFilter}
-                onChange={(e) => setLastActiveFilter(e.target.value as number)}
+                onChange={(e) => setLastActiveFilter(parseInt(e.target.value as string))}
```
bakeiro: this solved this issue :) [actually was easy to spot, not sure how I didn't realise before that this was the bug]
image

app/web/features/search/SearchPage.tsx
```diff
@@ -110,25 +111,14 @@ export default function SearchPage({
       completeProfileFilter,
     ],
     ({ pageParam }) => {
-      // @ts-ignore @TODO David fixing these in a separate PR
```
bakeiro, line 113: finally removed :)

app/web/features/search/SearchPage.tsx
```diff
@@ -110,23 +112,17 @@ export default function SearchPage({
...
-      const lastActiveComparation = parseInt(lastActiveFilter);
-      // @ts-ignore @TODO David fixing these in a separate PR
-      const hostingStatusFilterComparation = parseInt(hostingStatusFilter);
 
+      const numberOfGuestsTemp = numberOfGuestFilter ? parseInt(numberOfGuestFilter) : 0;
```
bakeiro, line 117: why this:
- I want to display an empty input, if I use the type number (which must be 0 at least) display always shows 0 at the beginning
     - [which isn't an issue but if the user types a number would look like '01' instead of '1'
- making it a string input looks empty, when user deletes the last digit looks empty again (and not a '0') so I think it's better this way even if in the code internally is handled as a string

app/web/features/search/FilterDialog.tsx
bakeiro: solved ✅

app/web/components/Select.tsx
nabramow: Is this for the phone? What's "native"?

app/web/components/Select.tsx
nabramow: If you're not gonna update this in this PR btw, I'd put it back.

app/web/features/search/FilterDialog.tsx
```diff
@@ -212,15 +240,39 @@ export default function FilterDialog({
...
-                }}
+                id="can_host_status_filter_1"
+                variant="outlined"
                 label={t("search:form.host_filters.hosting_status_field_label")}
+                multiple={true}
+                menuItems={true}
+                value={hostingStatusFilter}
+                onChange={(e) =>
+                  setHostingStatusFilter(e.target.value as number[])
+                }
```
nabramow, line 251: I don't think this is right. When they select an item, the value will be a single number. You're saying it's an array of numbers, but it's not.

Since it's a multiselect, I assume what you want to do is compare the `hostingStatusFilter`, if the value is there already, remove it, if it's not, add it.

Something like:

```javascript
  const handleHostingStatusFilterChange = (
    e: React.ChangeEvent<{ value: number }>
  ) => {
    if (hostingStatusFilter.includes(e.target.value)) {
      setHostingStatusFilter(
        hostingStatusFilter.filter((value) => value !== e.target.value)
      );
    } else {
      setHostingStatusFilter([...hostingStatusFilter, e.target.value]);
    }
  };

```

nabramow:
Okay apologies I didn't have a chance to really dive into this until now and understand what was happening fully.

I think you have the right idea here but there's a lot changing at once and I think it's making your life more difficult.

What I would do is break it into small pieces:

1. Start a new PR with just one of these filters, for example JUST 'hostingStatusFilter' and get that working with the types and the multi-select (from my understanding the main thing in this ticket is add multiple selection).
2. Don't remove the value type in`Select`. It seems you're removing it just to fix Typescript, but since this `Select` is used elsewhere, that increases the chance of a bug. Instead you want to add an array value to the `value` type, see #4881 .

What I see is you want the value to be a number and the label to be a string. Think of the types in that way.

The types here are not totally correct, and since so much changed at once it's hard to fix. This is why I would suggest starting a new PR to just do one first. Get the logic working without coercing the types (i.e. no `as number`, `as string`, etc.), add the design later. You can use the Material UI multiselect example here:

https://mui.com/material-ui/react-select/#multiple-select

Do a feature branch if you need to. Then it can be one PR to review the logic, then one PR to review design.

If it's easier, you can make a new shared MultiSelect component, rather than change the existing one.

I think it's a bit dangerous to merge this as is, as you're coercing the types `as number[]`, etc. when that's not really what they are. This could cause the app to error out. 

See example here: https://github.com/Couchers-org/couchers/pull/4881

Feel free to take over that PR, or copy from it and then close it.

app/web/components/Select.tsx
bakeiro: wait, I still didn't read the link you sent me, I just need more time 😅

app/web/components/Select.tsx
bakeiro: native it's an option for Material components, to make it look like native components in the phone (like when openning a dropdown for example) still didn't confirm it, that's something I wanted to do in the testing of this PR

app/web/features/search/FilterDialog.tsx
bakeiro: you are absolutely right, the thing is, I got this example from the oficial Material UI page, they have it like this and it works (something that confuse me)

bakeiro: image
image

bakeiro: I mean this PR it's a bit WIP I have to say, since some testing was pending and some things I wanted to double check (this is one of them), isn't like I'm absolutely ready with this PR everything is clear xd

bakeiro:
splitting in 2 PRs

https://github.com/Couchers-org/couchers/pull/4884
https://github.com/Couchers-org/couchers/pull/4881

Issue #4823, opened 2024-09-16
Create all events page


bakeiro:
This PR represents the entire feature branch.
**Context:**
User clicks on the events page, can see it's events, the discover section but wants to see it's events, so clicks the "see all" button

**Page behavior:**

- tab “upcoming”.
all events created by the user or to be participated (and that have not started yet), cancelled or not

- tab "past":
all events created by the user or that he/she has participated in but that have already happened, cancelled or not

![image](https://github.com/user-attachments/assets/7a606a5a-61bf-4043-986e-28353e5fb423)

bakeiro:
how every event should look like

![events_page_phase_1 - copia - copia](https://github.com/user-attachments/assets/b26d87e7-128a-4c1c-9510-22a339896c89)

bakeiro:
note: 
same event component as here https://github.com/Couchers-org/couchers/issues/4784 ignore the UI difference, it's just due the software I'm using to create this mockups

Issue #4784, opened 2024-09-10
Create new events page - Part I


nabramow:
Use design mockup from #3989.

This is the first iteration of the new events page. It should replace the existing events page in the top main navigation bar.

This first iteration should include your events, create an event, filter canceled, etc. The discover section will be part II.

Create a feature branch for events and make smaller PRs off of it.

bakeiro:
suggestion:
one iteration could be implementing this mockup without the pills (image attached)
[you can do it in any other way, this is just a suggestion]

![events_page_phase_1](https://github.com/user-attachments/assets/2f617319-c38b-4b35-9513-31b7b6396700)

bakeiro:
also, behavior detail, add a "see all" button at the botton of  "Your events" list when this is list is longer than 10 events (to show the user that the list is longer)  
![309244499-8af9497c-c6d1-4efc-bcda-7844b02da031](https://github.com/user-attachments/assets/4542a773-79f7-46f2-8d65-8e226b89499e)

bakeiro:
also, this is a mockup to show the page and the idea, don't copy the look & feel, we should use our components and our UI style (just in case U know)

bakeiro:
how every event should look like

![events_page_phase_1 - copia - copia](https://github.com/user-attachments/assets/a6cc9220-b645-4630-8b96-aead9d4a78e5)

notes:
- this mockup it's only to show the logic, ignore the UI difference with our page, should be implemented using our components/UI style
- event date can be an interval
- title can be long & contain emojis

bakeiro:
note about event image:
- when there is no event image we use this (image attached) we can keep doing this

![image](https://github.com/user-attachments/assets/3e967fc5-d41a-4a9f-a498-b2e6767c0ae5)

bakeiro:
for the events in Discovery we can keep using the old component :)

![image](https://github.com/user-attachments/assets/cc0cb950-1b57-46a4-b8e0-c0f98e23e491)

PR #4863, opened 2024-09-22
New Events Page - DESIGN ONLY

nabramow:
This is the DESIGN ONLY for the EventsList and the EventItem card. This will merge into the "web/feature/new-events-page" feature branch, not prod, so no need to worry. I am breaking them up as it will be a big feature and it's easier to review in pieces.

Please only review the design, the loading and functionality will come in the next PR. 

Right now I'm just using mock data from app/web/test/fixtures/events.json. If you want to test different UI like a very long title, online event, etc. you can add or adjust the data there.

Design mocks in these tickets: #4823, #4784

**Web frontend checklist**
- [ WILL DO ONCE DESIGN CONFIRMED ] Added any new components to storybook
- [ WILL DO IN FUTURE PR ] Added tests where relevant
- [ ] All tests pass
- [ DESIGN ONLY ] Clicked around my changes running locally and it works
- [ X ] Checked Desktop, Mobile and Tablet screen sizes
- 
Screenshot 2024-09-22 at 11 05 44 AM


app/web/features/communities/newEvents/EventsList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  const renderEventItems = () => {
+    return mockEvents.map((event) => (
+      <EventItem key={event.eventId} event={event} />
+    ));
+  };
+
+  return (
+    <div>
+      <Typography variant="h3">
```
bakeiro, line 24: IMHO feels redundant (if we have the tabs title)

app/web/features/communities/newEvents/EventItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+  },
+}));
+
+const EventItem = ({ event }: { event: Event.AsObject }) => {
+  const classes = useStyles({
+    eventImageSrc: event.photoUrl || eventImagePlaceholderUrl,
+  });
+  const { t } = useTranslation([GLOBAL, COMMUNITIES]);
+
+  const startTime = dayjs(timestamp2Date(event.startTime!)).format("llll");
```
bakeiro, line 78: what about:
- not showing the year when the event is upcoming
- show it when passed already?

not a request, but what do you think about it?

bakeiro: another thing (this could be a petition xd)

what about this formatting?
image

day_of_week, day, month **at** hour

app/web/features/communities/newEvents/EventItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const endTime = dayjs(timestamp2Date(event.endTime!));
+
+  const formattedEventDates = `${startTime} - ${endTime.format(
+    endTime.isSame(startTime, "day") ? "LT" : "llll"
+  )}`;
+
+  const {
```
bakeiro, line 80: since this is in every event item, maybe makes sense to receive it from it's parent? (by prop) [proporsal, not request]

app/web/features/communities/newEvents/EventItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const isCancelled = event.isCancelled;
+
+    return (
+      <div className={classes.tags}>
+        {isCreatedByMe && (
+          <Pill variant="rounded">{t("communities:created_by_me")}</Pill>
+        )}
+        {isOnline && <Pill variant="rounded">{t("communities:online")}</Pill>}
+        {isCancelled && (
+          <Pill variant="rounded">{t("communities:cancelled")}</Pill>
```
bakeiro: red color when cancelled? (as we do now)

image

bakeiro:
overall good, didn't spot nothing improveable in the code, tomorrow I will finish the review ;)

app/web/features/communities/newEvents/EventsList.tsx
nabramow: It was in the design, do you prefer I remove it? It doesn't really take much performance to render it either way.

app/web/features/communities/newEvents/EventItem.tsx
nabramow: IMHO makes more sense for each event to be responsible for its own data and logic. We're already passing the event in the props. It's not something that's being fetched from the DB so just parsing stuff so it's not a heavy calculation.

app/web/features/communities/newEvents/EventItem.tsx
nabramow: Done!

app/web/features/communities/newEvents/EventItem.tsx
nabramow: So I like this, but we're using the `LocalizedFormat` plugin to localize the dates for dayjs and it doesn't support that:

https://day.js.org/docs/en/display/format#list-of-localized-formats

I think it's not worth all the work it would take to get it to work, but I do agree I like that more (I'd also love to use the Calendar plugin to do `Today`, `Tomorrow`, `Yesterday` etc) but no localization support.

nabramow: Oh also, I decided to just show the `startTime` in the `EventsList` I think it's good just to have the most important data in the short version, then they can see the extended data in the event. What do you think?

bakeiro: fine with that :)

bakeiro:
petition about the tabs, can we use the original tabs? to maintain consistency (as the picture attached)
image

I think was a misunderstading with the mockups, my mockups are just to understand the abstract idea what I wanted to achieve, but we should keep using the original UI style/guidelines as we have in other pages, see comments in the attached pictures (maybe it's my fault because wasn't totally clear, but yeah for these mockups I'm using an AI to generate them, that's why they don't follow the Couchers UI guidelines 😅)
image
image

app/web/features/communities/newEvents/EventsList.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+interface EventListProps {
+  eventType: EventsType;
+}
+
+const EventsList = ({ eventType }: EventListProps) => {
+  const { t } = useTranslation([GLOBAL, COMMUNITIES]);
+
+  const renderEventItems = () => {
+    return mockEvents.map((event) => (
```
bakeiro, line 17: mockEvents in the list component?

bakeiro (approved):

bakeiro:
new issue created https://github.com/Couchers-org/couchers/issues/4877
PR approved

app/web/features/communities/newEvents/EventsList.tsx
nabramow: Yep as I described in the description of the PR this PR is just the design, so using mock events. Logic comes next.

PR #4881, opened 2024-09-23
Add multiselect example with no design [WIP]

nabramow:
This is just an example for adding multiselect, as the types are very tricky. **Don't merge this as is, as it has no proper design yet.**

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/Select.tsx
+++ b/app/web/components/Select.tsx
@@ -31,7 +31,9 @@ export default function Select<T extends Record<string | number, string>>({
 }: Omit<SelectProps, "children"> & {
   id: string;
   options: Extract<keyof T, string | number>[];
-  value?: T extends undefined ? string | number : keyof T;
+  value?: T extends undefined // can be undefined
+    ? string | number | number[] // strng number or array of numbers (for multiselect)
+    : keyof T | Array<keyof T>; // whatever type T is, or array of whatever type T is for multiselect
   optionLabelMap: T;
 }) {
   const classes = useStyles();
--- a/app/web/features/search/SearchPage.tsx
+++ b/app/web/features/search/SearchPage.tsx
@@ -4,9 +4,9 @@ import { Coordinates } from "features/search/constants";
 import { useTranslation } from "i18n";
 import { GLOBAL, SEARCH } from "i18n/namespaces";
 import { LngLat, Map as MaplibreMap } from "maplibre-gl";
-import { User } from "proto/api_pb";
+import { HostingStatus, User } from "proto/api_pb";
 import { UserSearchRes } from "proto/search_pb";
-import { useEffect, useRef, useState } from "react";
+import { ChangeEvent, useEffect, useRef, useState } from "react";
 import {
   QueryClient,
   QueryClientProvider,
@@ -86,7 +86,9 @@ export default function SearchPage({
     "location"
   );
   const [lastActiveFilter, setLastActiveFilter] = useState(0);
-  const [hostingStatusFilter, setHostingStatusFilter] = useState(0);
+  const [hostingStatusFilter, setHostingStatusFilter] = useState<
+    Exclude<HostingStatus, HostingStatus.HOSTING_STATUS_UNKNOWN>[] | undefined
+  >(undefined);
   const [numberOfGuestFilter, setNumberOfGuestFilter] = useState(0);
   const [completeProfileFilter, setCompleteProfileFilter] = useState(false);
   const [selectedResult, setSelectedResult] = useState<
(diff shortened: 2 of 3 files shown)


app/web/features/search/SearchPage.tsx
```diff
@@ -86,7 +86,9 @@ export default function SearchPage({
     "location"
   );
   const [lastActiveFilter, setLastActiveFilter] = useState(0);
-  const [hostingStatusFilter, setHostingStatusFilter] = useState(0);
+  const [hostingStatusFilter, setHostingStatusFilter] = useState<
+    Exclude<HostingStatus, HostingStatus.HOSTING_STATUS_UNKNOWN>[] | undefined
```
nabramow, line 90: The `Exclude` is because in the `Select` options in `FilterDialog` you're not including the `HostingStatus.HOSTING_STATUS_UNSPECIFIED` value.

If you don't want them to be able to choose unspecified, you need to exclude it from the type for `hostingStatusFilter` as well.

nabramow: Second question though, since that's a value that might be saved in the DB, do you really want to exclude it? Maybe people want to search for people who did not specify their hosting status? Or is it required now?

app/web/features/search/SearchPage.tsx
```diff
@@ -168,6 +166,31 @@ export default function SearchPage({
...
+
+      if (hostingStatusFilter.length === 0) {
+        setHostingStatusFilter(undefined);
+      } else {
+        setHostingStatusFilter(newHostingStatusFilter);
+      }
+    } else {
+      setHostingStatusFilter([...hostingStatusFilter, value]);
+    }
+  };
```
nabramow, line 192: In plain language, when the user chooses a value, since it's a multiselect we want to:

```javascript
    if (hostingStatusFilter === undefined) { // if no values yet, add the first one in an array
      setHostingStatusFilter([value]);
    } else if (hostingStatusFilter.includes(value)) { // if the value is already there, remove it
      const newHostingStatusFilter = hostingStatusFilter.filter(
        (existingValue) => existingValue !== value
      );

      if (hostingStatusFilter.length === 0) { // if removing the last value, set to undefined again
        setHostingStatusFilter(undefined);
      } else {
        setHostingStatusFilter(newHostingStatusFilter); // otherwise set to existing values with selected one removed
      }
    } else {
      setHostingStatusFilter([...hostingStatusFilter, value]); // otherwise, just add it to existing values
    }
  };
```

app/web/features/search/SearchPage.tsx
```diff
@@ -220,8 +243,8 @@ export default function SearchPage({
           setLocationResult={setLocationResult}
           lastActiveFilter={lastActiveFilter}
           setLastActiveFilter={setLastActiveFilter}
+          onChange={handleHostingStatusFilterChange}
           hostingStatusFilter={hostingStatusFilter}
-          setHostingStatusFilter={setHostingStatusFilter}
```
nabramow, line 224: We will handle this function in `SearchPage` since that's where the `hostingStatusFilter` state is, then just pass the function.

In general, good to keep the logic for the state in the same place the state is so you're not searching around for it and everything is consistent.

app/web/features/search/SearchPage.tsx
bakeiro: it's not saved in the DB, since the queries are generated dynamically, this is a SELECT in SQL, not an UPDATE, DELETE or any other write operation

bakeiro: so `HostingStatus.HOSTING_STATUS_UNSPECIFIED` would be (almost) the same as not filtering by hosting status at all

bakeiro: so from the UI perspective you are offering a option to the user, which IMHO would confuse him more rather than giving him an useful filter

app/web/components/Select.tsx
```diff
@@ -31,7 +31,9 @@ export default function Select<T extends Record<string | number, string>>({
 }: Omit<SelectProps, "children"> & {
   id: string;
   options: Extract<keyof T, string | number>[];
-  value?: T extends undefined ? string | number : keyof T;
+  value?: T extends undefined // can be undefined
+    ? string | number | number[] // strng number or array of numbers (for multiselect)
+    : keyof T | Array<keyof T>; // whatever type T is, or array of whatever type T is for multiselect
```
bakeiro, line 36: brilliant!

app/web/features/search/SearchPage.tsx
nabramow: Ah no I don't mean the Select value is stored in the DB. I mean the hostingStatus is stored in the DB in the `users` table.

What if the person us searching for someone to meet up with or my username and they don't care about their hosting status?

nabramow:
Closing as was just for an example.

PR #4884, opened 2024-09-24
types and minor changes

bakeiro:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/search/FilterDialog.tsx
+++ b/app/web/features/search/FilterDialog.tsx
@@ -94,8 +94,8 @@ interface FilterDialogProps {
   setHostingStatusFilter: Dispatch<SetStateAction<number>>;
   completeProfileFilter: boolean;
   setCompleteProfileFilter: Dispatch<SetStateAction<boolean>>;
-  numberOfGuestFilter: number;
-  setNumberOfGuestFilter: Dispatch<SetStateAction<number>>;
+  numberOfGuestFilter: string;
+  setNumberOfGuestFilter: Dispatch<SetStateAction<string>>;
 }
 
 export default function FilterDialog({
@@ -198,7 +198,7 @@ export default function FilterDialog({
                 id="last_active_filter"
                 className={classes.marginBottom}
                 value={lastActiveFilter}
-                onChange={(e) => setLastActiveFilter(e.target.value as number)}
+                onChange={(e) => setLastActiveFilter(parseInt(e.target.value as string))}
                 label={t("search:form.host_filters.last_active_field_label")}
                 optionLabelMap={getLastActiveOptions(t)}
                 options={[
@@ -253,8 +253,7 @@ export default function FilterDialog({
                 id="num-guests-filter"
                 value={numberOfGuestFilter}
                 onChange={(e) => {
-                  const tempNumOfGuest = parseInt(e.target.value);
-                  setNumberOfGuestFilter(tempNumOfGuest);
+                  setNumberOfGuestFilter(e.target.value);
                 }}
                 inputRef={register({
                   valueAsNumber: true,
--- a/app/web/features/search/FilterDialog.tsx
+++ b/app/web/features/search/FilterDialog.tsx
@@ -134,7 +134,7 @@ export default function FilterDialog({
   };
 
   const handleLastActiveChange = (
-    event: React.ChangeEvent<{ value: unknown }>
+    event: React.ChangeEvent<{value: unknown}>
   ) => {
     const value = parseInt(event.target.value as string);
     setLastActiveFilter(value);
--- a/app/web/features/search/FilterDialog.tsx
+++ b/app/web/features/search/FilterDialog.tsx
@@ -123,6 +123,23 @@ export default function FilterDialog({
     theme.breakpoints.down("sm")
   );
 
+  const handleNumGuestsChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const convertedValue = parseInt(e.target.value);
+    const tempNumOfGuest =
+      !Number.isNaN(convertedValue) && convertedValue > 0
+        ? convertedValue
+        : undefined;
+
+    setNumberOfGuestFilter(tempNumOfGuest);
+  };
+
+  const handleLastActiveChange = (
+    event: React.ChangeEvent<{ value: unknown }>
+  ) => {
+    const value = parseInt(event.target.value as string);
+    setLastActiveFilter(value);
+  };
+
   return (
     <Dialog
       open={isOpen}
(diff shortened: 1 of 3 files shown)


bakeiro:
why changing num of guest? this is the explanation
image

app/web/features/search/FilterDialog.tsx
```diff
@@ -198,7 +198,7 @@ export default function FilterDialog({
                 id="last_active_filter"
                 className={classes.marginBottom}
                 value={lastActiveFilter}
-                onChange={(e) => setLastActiveFilter(e.target.value as number)}
+                onChange={(e) => setLastActiveFilter(parseInt(e.target.value as string))}
```
bakeiro, line 201: why changing num of guest? this is the explanation
image

app/web/features/search/FilterDialog.tsx
```diff
@@ -252,8 +252,11 @@ export default function FilterDialog({
                 variant="standard"
                 id="num-guests-filter"
                 value={numberOfGuestFilter}
+                inputProps={{ min: 0 }}
                 onChange={(e) => {
-                  const tempNumOfGuest = parseInt(e.target.value);
+                  const convertedValue = parseInt(e.target.value);
+                  const tempNumOfGuest = !Number.isNaN(convertedValue) && convertedValue > 0 ? convertedValue : undefined;
+
```
bakeiro: Explanation:

- ParseInt("") => NaN
- Also, if user types -5 will cause an error
- Aaand, if the type it's always number and NOT undefined, this input always contains a number, so isn't empty, which (IMHO) is annoying, plus when the user types a digit (ie: '1') result will be '01'.

With this solution I solve all the mentioned problems

app/web/features/search/FilterDialog.tsx
nabramow: Please update the type in the parent component `SearchPage` for this value rather than using `as string` to coerce it. The reason it's giving you an error is that the `lastActiveFilter` and the `numberOfGuestsFilter` being passed to this component has a different type in the parent component `SearchPage`.

`as [TYPE]` should be used very rarely and usually signals something is defined wrong with the types.

app/web/features/search/FilterDialog.tsx
nabramow: I think if you fix the types in the parent component you won't need so much logic, so let's see.

app/web/features/search/FilterDialog.tsx
bakeiro: hmm, the thing is, I want the type to be a number, because the indexes of `lastActiveOptions` are numbers, plus the function which sends this value to the backend needs a number (image attached) (plus before sending I want to compare if it's equally 0).  

The `as string` I needed because (if I'm not wrong) because `e.target.value` is unknown, so TS complains and I need to type it, and since actually (regardless of the TS type) it's an string I need to convert it to a number (if I'm following the conversation correctly)

app/web/features/search/FilterDialog.tsx
bakeiro: but I need it to be a number, also being an string I can't compare if it's lower than 0 or feels wrong doing it, plus string are not designed to be used with math operators (`> 0`)...

app/web/features/search/FilterDialog.tsx
bakeiro: image

bakeiro: basically TS complains because I can't convert an unknown typed variable to a number (even if it's not an unknown and its an actual string)

nabramow: Ahhh okay, I see what you mean.

Could you do it like this so it's more clear?

In reusable `Select` component add onChange prop with type like this:

```javascript
export default function Select({
    onChange
}: {
    onChange?: (event: React.ChangeEvent<{ value: unknown }>) => void;
}) =>

     <MuiSelect
        onChange={onChange}
      >
```

That makes it more clear it's supposed to be that way. Right now it's hidden in `otherProps`.

And would you mind doing the coercion in a `handleLastActiveFilterChange function like this?:

```javascript
const handleLastActiveChange = (
  event: React.ChangeEvent<{ value: unknown; }>
) => {
  const value = event.target.value as number // This way you're casting it to a new value rather than doing it to event.target.value directly
  setLastActiveFilter(value);
};
```

app/web/features/search/FilterDialog.tsx
nabramow: Okay got it, I made a comment below for the other one, could you do the coercion for these others like that in a separate function variable like `handelNumGuestsChange`?

It will be more performant by preventing re-renders, especially since calculations are happening here.

bakeiro: sure ✅

app/web/features/search/FilterDialog.tsx
bakeiro: ok, the thing is

I need to parse it to an number so I can compare it later, so `parseInt()` is something needed, if I do `as number` it's only for typescript right? but I need a number in JS the TS is only for the typing...

for using the parseInt it needs to receive an string, so if it's of type unknown complains, what I can do it's to modify the prop in the general select, to `event: React.ChangeEvent<{ value: string | unknown}>` I guess that would be fine? (feels to much chaning the prop just for this, but I guess it's the solution to avoid the `as string`)

bakeiro: and this in the `handleLastActiveChange`, I guess with that it's fine? (since I need the `parseInt()`

nabramow: I think in this specific case it is a number, since the options you're giving the 1`Select` are numbers. The issue is the Material-UI `Select` type for the `onChange` says the value is unknown, and doesn't let us change it (I tried).

You can test this by logging `console.log('event.target.value type', typeof event.target.value)`  or `console.log('is event.target.value number?:, Number.isInteger(event.target.value)` and that should tell you the type.

bakeiro: :D as I said, it's an string
image

nabramow: I think I'm getting confused with the `hostingStatus` and `numGuests` 🙈.

nabramow:
I think this one should merge first before the other one since it effects its types?

app/web/features/search/FilterDialog.tsx
bakeiro: I think it's fine, I think all of this it's due an `as string` which I believe I need it, I suggest to merge this, and afterwards you can check if you find something improvable (but I want me who does it, if possible, since I want to improve by myself)

bakeiro: also I feel we are blocking it bc of that, and this is a small PR

nabramow: @bakeiro happy to approve this one once it's passing the gitlab check!

bakeiro:
@nabramow approve for this PR? :D

nabramow (approved):
Sweeeet, thanks for this!! Sorry for my delay!

PR #4886, opened 2024-09-24
mutiple option select hosting status

bakeiro:
before:
image

now:
image

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [x] All tests pass


app/web/features/search/FilterDialog.tsx
```diff
@@ -212,15 +238,39 @@ export default function FilterDialog({
...
-                  setHostingStatusFilter(e.target.value as number);
-                }}
+                id="can_host_status_filter_1"
+                variant="outlined"
                 label={t("search:form.host_filters.hosting_status_field_label")}
+                multiple={true}
+                menuItems={true}
+                value={hostingStatusFilter}
+                onChange={(e) =>
+                  setHostingStatusFilter(e.target.value as number[])
```
bakeiro: why this onChange function? I did based on the MUI documentation, basically they are doing this to show as demo

image
image

bakeiro: https://v4.mui.com/components/selects/#multiple-select

app/web/features/search/SearchPage.tsx
```diff
@@ -112,8 +114,6 @@ export default function SearchPage({
     ({ pageParam }) => {
       // @ts-ignore @TODO David fixing these in a separate PR
       const lastActiveComparation = parseInt(lastActiveFilter);
```
bakeiro: https://github.com/Couchers-org/couchers/pull/4884

bakeiro:
note to myself
- [ ] check `native` option

app/web/features/search/FilterDialog.tsx
bakeiro: ah wait, of course [I understand now why this works]

it receives directly an `string[]`, so I it always replace the current state's value by the event's value (even if the values don't change).  

It's how it's done in the official docs... so I guess it's fine doing it this way (the only downside I can think it's about re-updating even when the values are the same, but the dropdown doesn't allow the user to do it [in the UI], so I don't think we need the complex function which searches, compares and updates the value)

nabramow: Ah okay I see it is sending an array in this case.

Please fix the types though. You should not need `as number[]` on line 248 or on line 256. That means the type is defined wrong somewhere higher up.

Also better to make a `const handleChange = () => {}` function since it's changing the state. This will prevent the value from being regenerated every time the function runs.

nabramow: Also what happens if the user wants to search for someone to meet up with and doesn't care about hosting status?

nabramow:
Just a small design thing, can you add some padding to the left of the options?:

Screenshot 2024-09-24 at 8 50 25 AM

Would also be nice to have little `x`'s on the tags eventually so you can just delete them by clicking on them, but maybe that's more a future feature request,

app/web/components/Select.tsx
```diff
@@ -31,10 +34,16 @@ export default function Select<T extends Record<string | number, string>>({
...
+  value?: T extends undefined
+    ? string | number | number[]
+    : keyof T | Array<keyof T>;
+  menuItems?: boolean;
   optionLabelMap: T;
 }) {
   const classes = useStyles();
+  const OptionComponent = menuItems
+    ? MenuItem
+    : ("option" as unknown as React.ElementType);
```
nabramow: Please do not use `as unknown as React.ElementType`, to me that signals something is wrong with the type and you are coercing it.

I think you want something like:

```javascript
const OptionComponent: React.ElementType = menuItems ? MenuItem : "option";
```

app/web/features/search/FilterDialog.tsx
```diff
@@ -90,8 +111,8 @@ interface FilterDialogProps {
   setLocationResult: Dispatch<SetStateAction<GeocodeResult>>;
   lastActiveFilter: number;
   setLastActiveFilter: Dispatch<SetStateAction<number>>;
-  hostingStatusFilter: number;
-  setHostingStatusFilter: Dispatch<SetStateAction<number>>;
+  hostingStatusFilter: typeHostingStatusOptions;
```
nabramow, line 114: Types should generally be in all caps, so `HostingStatusOptionsType`

app/web/features/search/FilterDialog.tsx
```diff
@@ -212,15 +233,45 @@ export default function FilterDialog({
...
-                id="can_host_status_filter"
-                value={hostingStatusFilter}
-                onChange={(e) => {
-                  setHostingStatusFilter(e.target.value as number);
-                }}
+                id="can_host_status_filter_1"
+                variant="outlined"
                 label={t("search:form.host_filters.hosting_status_field_label")}
+                multiple={true}
+                menuItems={true}
```
nabramow, line 272: For boolean values that are true, you can just write `multiple` (without the ={true}` and it will work.

app/web/features/search/FilterDialog.tsx
```diff
@@ -212,15 +233,45 @@ export default function FilterDialog({
...
+                  <Input
+                    className={classes.inputHostingStatus}
+                    id="select-multiple-chip"
+                  />
+                }
+                className={classes.marginBottom}
+                native={false}
+                renderValue={(selected) => (
+                  <div className={classes.chips}>
+                    {(selected as number[]).map(
```
nabramow: Please remove the `as number[]` and fix the issues with the type instead. Check the options and the value in the parent component `SearchBox` for clues.

app/web/features/search/FilterDialog.tsx
```diff
@@ -212,15 +233,45 @@ export default function FilterDialog({
...
+                    )}
+                  </div>
+                )}
+                MenuProps={MenuProps}
                 optionLabelMap={getHostingStatusOptions(t)}
                 options={[
-                  HostingStatus.HOSTING_STATUS_UNSPECIFIED,
                   HostingStatus.HOSTING_STATUS_CAN_HOST,
                   HostingStatus.HOSTING_STATUS_MAYBE,
                   HostingStatus.HOSTING_STATUS_CANT_HOST,
```
nabramow, line 312: What if the user is just looking to meet up with someone and doesn't care about their hosting status?

app/web/features/search/SearchPage.tsx
```diff
@@ -15,10 +15,17 @@ import {
...
 import FilterDialog from "./FilterDialog";
 import MapWrapper from "./MapWrapper";
 import SearchResultsList from "./SearchResultsList";
 
+export type typeHostingStatusOptions = Exclude<
+  HostingStatus,
+  | HostingStatus.HOSTING_STATUS_UNKNOWN
+  | HostingStatus.HOSTING_STATUS_UNSPECIFIED
+>[];
+
```
nabramow: Types should generally be all caps so `HostingStatusOptionsType`.

nabramow:
A few things here. For all the `onChange` values in `FilterDialog`:

Since this type is being stored in `SearchPage` in a `useState` you should write the functions there. I'd suggest make a `handleNumGuestFilterChange` function in that component, then pass it in the props as `onNumGuestsFilterChange` and call it in this function.

So in `searchPage`:

```javascript
const handleNumGuestsFilterChange = (event) => {
// logic you need once fixing the type
setNumGuestsFilter(event.target.value)
}

<FilterDialog
    onNumGuestsFilterChange={handleNumGuestsFilterChange}
/>
```

in `FilterDialog`:

``` javascript
<Select
    onChange{onNumGuestsFilterChange}
/>
```


This is a better practice as it will keep the logic together with the values in the parent component. This will make it easier to keep track of the types, which is what's causing so much confusion with the different type definitions for the same value in different components.

Same with the other Select onChanges: `setActiveFilter`, `hostingStatusFilter`.

app/web/components/Select.tsx
nabramow, line 46:
```suggestion
const OptionComponent: React.ElementType = menuItems ? MenuItem : "option";
```

app/web/features/search/FilterDialog.tsx
bakeiro: then leaves the dropdown unselected, and since it's undefined value isn't sent to the server

app/web/components/Select.tsx
```diff
@@ -23,6 +24,8 @@ const useStyles = makeStyles((theme) => ({
 export default function Select<T extends Record<string | number, string>>({
   id,
   className,
+  native = true,
```
nabramow, line 27: I think `native` should be false when `menuItems` is true, is that right? So when you use `<option>`, `native` should be false.

Not 100% sure though.

So you might want it default to `false` then in this case since most of the existing Selects are not multiselect.

app/web/components/Select.tsx
bakeiro: done ✅

app/web/features/search/SearchPage.tsx
bakeiro: done ✅

app/web/features/search/FilterDialog.tsx
bakeiro: done ✅

app/web/features/search/FilterDialog.tsx
nabramow: Ah perfect thanks just wanted to make sure it was possible.

app/web/components/Select.tsx
bakeiro: I mean, the thing it's that is true by default, if I set it to false all the components that use <select> will have `native=false` which basically I don't think is retro compatible

bakeiro:
> A few things here. For all the `onChange` values in `FilterDialog`:
> 
> Since this type is being stored in `SearchPage` in a `useState` you should write the functions there. I'd suggest make a `handleNumGuestFilterChange` function in that component, then pass it in the props as `onNumGuestsFilterChange` and call it in this function.
> 
> So in `searchPage`:
> 
> ```js
> const handleNumGuestsFilterChange = (event) => {
> // logic you need once fixing the type
> setNumGuestsFilter(event.target.value)
> }
> 
> <FilterDialog
>     onNumGuestsFilterChange={handleNumGuestsFilterChange}
> />
> ```
> 
> in `FilterDialog`:
> 
> ```js
> <Select
>     onChange{onNumGuestsFilterChange}
> />
> ```
> 
> This is a better practice as it will keep the logic together with the values in the parent component. This will make it easier to keep track of the types, which is what's causing so much confusion with the different type definitions for the same value in different components.
> 
> Same with the other Select onChanges: `setActiveFilter`, `hostingStatusFilter`.

ok, but in the searchPage all the components have a lot of props and a lot of setValueFunction, this changes feels like a conversion of that page, did you saw how many setters are passed in the props?
image

also if we do it only for `FilterDialog` and not for the others feels confusing, and all because of the typing... do you think it's possible to make all onChange functions outside the render function as you said and the rest in other PR? feels outside the scope of this PR

nabramow:
> > A few things here. For all the `onChange` values in `FilterDialog`:
> > Since this type is being stored in `SearchPage` in a `useState` you should write the functions there. I'd suggest make a `handleNumGuestFilterChange` function in that component, then pass it in the props as `onNumGuestsFilterChange` and call it in this function.
> > So in `searchPage`:
> > ```js
> > const handleNumGuestsFilterChange = (event) => {
> > // logic you need once fixing the type
> > setNumGuestsFilter(event.target.value)
> > }
> > 
> > <FilterDialog
> >     onNumGuestsFilterChange={handleNumGuestsFilterChange}
> > />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in `FilterDialog`:
> > ```js
> > <Select
> >     onChange{onNumGuestsFilterChange}
> > />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is a better practice as it will keep the logic together with the values in the parent component. This will make it easier to keep track of the types, which is what's causing so much confusion with the different type definitions for the same value in different components.
> > Same with the other Select onChanges: `setActiveFilter`, `hostingStatusFilter`.
> 
> ok, but in the searchPage all the components have a lot of props and a lot of setValueFunction, this changes feels like a conversion of that page, did you saw how many setters are passed in the props? image
> 
> also if we do it only for `FilterDialog` and not for the others feels confusing, and all because of the typing... do you think it's possible to make all onChange functions outside the render function as you said and the rest in other PR? feels outside the scope of this PR

Yeah fair I guess it can be considered tech debt. I think after events I will do some tech debt things like upgrade Material UI, etc. so maybe we can do it then.

app/web/service/search.ts
```diff
@@ -29,6 +29,16 @@ export async function userSearch(
   }: UserSearchFilters,
   pageToken = ""
 ) {
+
+  console.debug(`-----v-----`);
```
bakeiro, line 33: will delete after approval :)

app/web/features/search/SearchBox.tsx
```diff
@@ -139,7 +139,7 @@ export default function SearchBox({
           <RadioGroup
             className={classes.justifyContent}
             row
-            onChange={(event) => handleOnChangeRadioButton(event, searchType)}
+            onChange={(event, value) => handleOnChangeRadioButton(event, value as "location" | "keyword")}
```
bakeiro: 👀 https://github.com/Couchers-org/couchers/issues/4955

bakeiro:
> Yeah fair I guess it can be considered tech debt. I think after events I will do some tech debt things like upgrade Material UI, etc. so maybe we can do it then.

https://github.com/Couchers-org/couchers/pull/4942 :)

app/web/features/search/SearchBox.tsx
```diff
@@ -139,7 +139,9 @@ export default function SearchBox({
           <RadioGroup
             className={classes.justifyContent}
             row
-            onChange={(event) => handleOnChangeRadioButton(event, searchType)}
+            onChange={(event, value) =>
+              handleOnChangeRadioButton(event, value as "location" | "keyword")
```
nabramow: Could you add a comment here that material-ui has this type as `unknown` hence the type coercion?

app/web/features/search/FilterDialog.tsx
```diff
@@ -229,15 +258,47 @@ export default function FilterDialog({
...
+                  <Input
+                    className={classes.inputHostingStatus}
+                    id="select-multiple-chip"
+                  />
+                }
+                className={classes.marginBottom}
+                native={false}
+                renderValue={(selected) => (
+                  <div className={classes.chips}>
+                    {(selected as TypeHostingStatusOptions).map(
```
nabramow, line 289: Could you leave a comment here that the type coercion is because the selected value type is unknown in the material ui Select component?

app/web/features/search/SearchPage.tsx
```diff
@@ -112,18 +119,13 @@ export default function SearchPage({
...
       return service.search.userSearch(
         {
           query: queryName,
           bbox: locationResult.bbox,
           lastActive: lastActiveFilter === 0 ? undefined : lastActiveFilter,
           hostingStatusOptions:
-            hostingStatusFilterComparation === 0
-              ? undefined
-              : [hostingStatusFilter],
+            hostingStatusFilter.length === 0 ? undefined : hostingStatusFilter,
```
nabramow, line 128: Since you have an empty array as the default state in the `useState` above, shouldn't the `hostingStatusFilter.length === 0` case also be an empty array (rather than `undefined`)? Either way they should match the default states.

nabramow:
Just a couple small comments and a small design thing. Feel free to make a separate ticket for the design one. For the selected state, it's not so obvious it's selected IMO. What about making the selected color our light green? Kinda like this Material-UI example:

Screenshot 2024-10-14 at 2 20 02 PM


Otherwise it's working great for me!

bakeiro:
> Just a couple small comments and a small design thing. Feel free to make a separate ticket for the design one. For the selected state, it's not so obvious it's selected IMO. What about making the selected color our light green? Kinda like this Material-UI example:
> 
> Screenshot 2024-10-14 at 2 20 02 PM
> Otherwise it's working great for me!

true!, I think it's worth doing it :) will implement it here directly

app/web/features/search/SearchPage.tsx
bakeiro: ah I got what you mean, well basically everything that is `undefined` isn't sent to the server, so basically following that rule (this rule applies for this function, isn't something standard in our code) when is empty basically don't send anything to the server, as we were doing before.

bakeiro: like maybe sending an empty array to the server is the same as undefined, but I rather don't risk to do it, also maybe if the backends changes (by mistake for example) will the frontend in that scenario.  
If we don't send anything (like before) then we make it more retro-compatible

bakeiro: basically I debug what we sent to the server every time I change the state, to proof doesn't make a difference, and keep it retro-compatible

nabramow: Ah sorry you're right I thought this was the value for `hostingStatusFilter` but it's actually the value for `hostingStatusOptions`

bakeiro:
@nabramow 
Done :)    [default gray, but I can update it for other one :)   ]
image

nabramow (approved):
Excellent, thanks for making those changes! Looks good to me! 🚀

PR #4894, opened 2024-09-25
Add showCancelled to ListAllEvents and filter on backend

nabramow:
This adds `showCancelled` to the `ListAllEvents` function on the backend, to the protos and switches out the existing events frontend to filter on the backend instead of the frontend (same as `ListMyEvents` works).

**TO TEST:**

- Regenerate your protos locally and restart your docker containers on the backend
- Go to the main events page
- Make sure you have a canceled event in the events list. Create an event and cancel it if not.
- Toggle the "show cancelled". It should add and remove the cancelled events from the list.

**Backend checklist**
- [ x ] All tests pass

**Web frontend checklist**
- [ x ] All tests pass


aapeliv:
Try this hash: `9e66f9be792b3f27b73c6d8f7b2150f84406948edca85252134fc212a3fd14e8`

app/proto/events.proto
```diff
@@ -294,6 +294,8 @@ message ListAllEventsReq {
 
   uint32 page_size = 2;
   string page_token = 3;
+  
+  bool include_cancelled = 4;
```
aapeliv: Please move this above the `page_{size,token}`

aapeliv (commented):
One minor comment.

app/proto/events.proto
nabramow: Done! Just out of curiosity, what does moving the location up do?

aapeliv: Nothing, but it's just clearer since the pagination stuff should come after all the content :)

aapeliv (dismissed):
LGTM

aapeliv (approved):

PR #4910, opened 2024-09-25
Get rid of protos sha business

aapeliv:


nabramow (approved):
Looks good, I was able to build the container and successfully run the app!

PR #4912, opened 2024-09-26
New Events Page - Functionality

nabramow:
This PR adds all the functionality for the events page.

**NOTE:** I did not add the "Friends" filters under `Discover` as there is no backend functionality for it yet.

There will be one final PR where I add the tests and the storybook components.

**TO TEST:**
- You need to make a variety of events - canceled, online, make 1-2 in the past by setting to 1 min from now (and waiting 1 min lol), different locations. If possible log in to a second account and make events from another user to test the "my events" part (user that's not you should only show under discover if you're not attending).
- Click around the filters, upcoming, past, use the search input, etc.
- Test pagination, i.e. if you are on last page and change filter, it should go back to first page.
- Make sure "My Events" on the dashboard page not changed

**Web frontend checklist**
- [ NEXT PR ] Added any new components to storybook
- [ NEXT PR ] Added tests where relevant
- [ x ] All tests pass


bakeiro:
image
mouse pointer missing
image

bakeiro:
image
[For UI team]
I was thinking they were toggles, not like checkboxes.... suggestion, if we implement the same pills on the map and these will be toggles, then I suggest toggles here, if they will be like checkboxes, checkboxes here

bakeiro:
[For UI team]
do we need this?
image

bakeiro:
[For UI]
I was thinking like a button of "see all" when user clicks there leads to a different page can see all it's created/attended/interested events, either upcoming or in the past...
for me everything in one page [upcoming, past, attended & discover] feels a bit overloaded (personal opinion) thoughs?

309244499-8af9497c-c6d1-4efc-bcda-7844b02da031

- note: ignore the missing pills here and the page subtitle :)

app/web/features/communities/newEvents/EventsList.tsx
```diff
@@ -1,15 +1,92 @@
...
+        {events.map((event) =>
+          isVerticalStyle ? (
+            <EventCard key={event.eventId} event={event} />
+          ) : (
+            <EventItem key={event.eventId} event={event} userId={userId} />
+          )
+        )}
+      </div>
+      <div className={classes.seeMoreContainer}>
+        {hasNextPage && (
```
bakeiro: this seems to be cached in react query, so once the user clicks the next time is displaying all the results... [anyway I'd wait until UI feedback about this, I was thinking about a new page but we can see what's they view]

bakeiro:
to contextualize

tickets
https://github.com/Couchers-org/couchers/issues/4784
https://github.com/Couchers-org/couchers/issues/4823

app/web/features/communities/newEvents/EventsList.tsx
nabramow: It should display 10 more at a time as I set the pageSize to 10 in this ReactQuery hook. Or am I misunderstanding?

nabramow:
Fixed the mouseover pointer on the pills, I think everything else was for UI to answer. If not let me know!

bakeiro:
> Fixed the mouseover pointer on the pills, I think everything else was for UI to answer. If not let me know!

yeah I need some time to re-check (I'm not that fast sadly 😅) today I will give it another review :)

app/web/features/communities/newEvents/EventsList.tsx
bakeiro: like, in principle it should lead to another page [UI liked this idea, I guess we can chat the tuesday about this] if not I'm expecting it like a toggle, so next time don't show all the events until I click again on the button [that's what me personally as a user would expect the page to behave]

bakeiro: maybe other people expect other thing, but feels like should be like a toggle in the state, and not something stored in cache until it's cleared [for me, but I suspect other people would feel the same]

nabramow: A separate page I understand, if we decide we wanna do that, would it happen on 1st click of see more? Or we add that additional "see more" button to the top right and that one opens a new page, but also have the pagination at the bottom?

The toggle I'm not understanding. It's not showing all events on the button click, just 10 more. It's only showing the rest of the events if you have less than 10 more.

Or you mean when they refresh the page you expect the list to reset to 10 items?

What about pagination like this? Then it doesn't get too long: https://mui.com/material-ui/react-pagination/

bakeiro: like both buttons that are labeled as "see all" (doesn't matter if it's on the top or the botton of the list) leads user to a new page, there can see upcoming & past events, for created, participated or interested events

bakeiro: anyway I think tomorrow we can discuss this again in the meeting

app/web/features/communities/newEvents/EventsPage.tsx
```diff
@@ -45,27 +112,137 @@ const EventsPage = () => {
...
+
+  const handleFilterNearMeClick = async () => {
+    const user = await getCurrentUser();
+
+    if (nearMeLocation) {
+      setNearMeLocation(undefined);
+    } else {
+      setIsOnlineOnly(false);
+      setLocationResult("");
+      setNearMeLocation(new LngLat(user.lng, user.lat));
```
bakeiro: about the near me, as I said, was the original idea in the mockup, but I though about replacing it by the location pill (which will open a pop-up to type location) or the location bar as you suggested, the thing is, [for me] since user can type its location directly having the near me button feels redundant.  

Maybe makes sense to use it and asks the browser the current user's location, we can speak about this in the meeting tomorrow

bakeiro:
untranslated token?
image

app/web/features/communities/newEvents/EventsPage.tsx
bakeiro: regardless about the near me vs searchbar or whatever else, seems that selecting the results doesn't affect the results grid
image

bakeiro:
I feel that not showing this part when there are no events makes the page more clear, and less loaded
image

app/web/features/communities/newEvents/EventsList.tsx
bakeiro: Just to share, this one of my firsts mockups

image

bakeiro: image

bakeiro: image

bakeiro (changes requested):
.

app/web/features/communities/newEvents/EventsPage.tsx
nabramow: Should be fixed now!

nabramow:
> I feel that not showing this part when there are no events makes the page more clear, and less loaded image

Hmm but what happens if they select a filter that does not have any events, the section disappears, then they want to undo it and the section and filters are gone? They'd have to refresh the page, which IMO not great UI.

Also more consistent if it's always there, then they know there WOULD be events there if they had any. Also more confirmation if there are any events for the filters they selected.

nabramow:
Okay so I checked with the pageNumber pagination and backend doesn't support it right now. Since no one feels super strongly either way, I will do the new page for see more events since it's easier than adjusting the backend right now.

bakeiro:
> > I feel that not showing this part when there are no events makes the page more clear, and less loaded image
> 
> Hmm but what happens if they select a filter that does not have any events, the section disappears, then they want to undo it and the section and filters are gone? They'd have to refresh the page, which IMO not great UI.
> 
> Also more consistent if it's always there, then they know there WOULD be events there if they had any. Also more confirmation if there are any events for the filters they selected.

Sure, I meant when that section doesn't have any events at all, not even events for filtering. if after filtering is empty should be there, of course.
> then they know there WOULD be events there if they had any  

yeah I think that's right, let's keep it that way then :)

app/web/features/communities/newEvents/MyEventsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+          className={showCancelled ? classes.selectedFilter : classes.filter}
+          variant="body2"
+          onClick={handleFilterShowCancelledClick}
+        >
+          {t("communities:show_cancelled_events")}
+        </Typography>
+      </div>
+      {!hasEvents && !isLoading && (
+        <TextBody className={classes.emptyState}>
+          {t("communities:events_empty_state")}
```
bakeiro, line 125: [small detail, suggestion only]

What about
`events that you have created, or that you participate in will appear here.`

instead of 
`No events at the moment`

bakeiro (approved):
LGTM

bakeiro:
- [ ] do we need to create/updat unit tests for the events page?

bakeiro:
small issue here
https://github.com/Couchers-org/couchers/issues/4979

app/web/service/events.ts
```diff
@@ -249,19 +249,40 @@ export async function listAllEvents({
...
+  if (pageNumber) {
+    req.setPageNumber(pageNumber);
+  }
   if (pageSize) {
     req.setPageSize(pageSize);
   }
   if (pageToken) {
     req.setPageToken(pageToken);
   }
+  if (showCancelled !== undefined) {
```
bakeiro, line 283: `if (showCancelled)` = `if (showCancelled !== undefined)` since it's a boolean?

app/web/features/communities/newEvents/DiscoverEventsList.tsx
```diff
@@ -0,0 +1,212 @@
...
+      fieldError={errors.location?.message}
+      fullWidth={isMobile}
+    />
+  );
+
+  return (
+    <>
+      <div className={classes.column}>
+        <Typography variant="h2">
+          {t("communities:discover_events_title")}
```
bakeiro, line 161: H1 instead? feels a bit small

image

app/web/features/communities/newEvents/MyEventsList.tsx
nabramow: Hmm I would argue that should be an additional description that's always there? Because even if they have events we might want that to be clear which events they are.

app/web/service/events.ts
nabramow: Then it won't set the showCancelled value in the request if it's `false`...

app/web/features/communities/newEvents/DiscoverEventsList.tsx
nabramow: UI wise and SEO for google crawling wise a page should only have one H1 per page.

I can do H2s but had it originally and they felt too big. Maybe I do the final ticket with tests and I assume Jesse and co will test before release and we can do all the UI preferences then.

app/web/service/events.ts
bakeiro: 🤦‍♂️👍

PR #4939, opened 2024-10-03
Add page number pagination to ListMyEvents and EventSearch

nabramow:
This PR adds pageNumber pagination to `ListMyEvents` and `EventSearch` while maintaining the existing `pageToken` pagination.

Discussed with @bakeiro and decided to do it this way if the backend part won't take too long.

**Backend checklist**
- [ x ] All tests pass

Closes #4939


nabramow:
Also halp why is my deployment failing, I'm not able to see it since we don't have an account for me there.

app/backend/src/couchers/servicers/events.py
```diff
@@ -1043,13 +1048,19 @@ def ListMyEvents(self, request, context, session):
...
+        occurrences = (
+            occurrences.offset(offset).limit(page_size) if request.page_number else occurrences.limit(page_size + 1)
+        )
         occurrences = session.execute(occurrences).scalars().all()
 
         return events_pb2.ListMyEventsRes(
             events=[event_to_pb(session, occurrence, context) for occurrence in occurrences[:page_size]],
             next_page_token=str(millis_from_dt(occurrences[-1].end_time)) if len(occurrences) > page_size else None,
+            page_number=page_number,
+            total_items=total_items,
```
bakeiro, line 1062: I believe (I can be wrong here, just an possible idea)

changing the return of this endpoint, could lead to a different proto object, leading to a different type in the frontend, and maybe that could be the issue of the build failing?

bakeiro:
> Also halp why is my deployment failing, I'm not able to see it since we don't have an account for me there.

the result of the build it's in Gitlab... do you need an account to see it? anyway, by executing `yarn build` in the directory `app/web` is causing some issues... maybe that's the root of the bug

nabramow:
> > Also halp why is my deployment failing, I'm not able to see it since we don't have an account for me there.
> 
> the result of the build it's in Gitlab... do you need an account to see it? anyway, by executing `yarn build` in the directory `app/web` is causing some issues... maybe that's the root of the bug

No, when I click through it says I need an account and Aapeli said it costs to add seats. `yarn build` was a good idea though, that gave me a clue, thanks!

app/backend/src/couchers/servicers/events.py
nabramow: The issue ended up being that the `MyEvents` tests was throwing an error as it was missing the `pageNumber` and `totalItems` in the api response. I handled this by making these keys optional in the response in the protos. Then we can just use them if we want to, but don't have to.

app/backend/src/couchers/servicers/events.py
```diff
@@ -1043,13 +1048,19 @@ def ListMyEvents(self, request, context, session):
...
+        occurrences = (
+            occurrences.offset(offset).limit(page_size) if request.page_number else occurrences.limit(page_size + 1)
+        )
         occurrences = session.execute(occurrences).scalars().all()
 
         return events_pb2.ListMyEventsRes(
             events=[event_to_pb(session, occurrence, context) for occurrence in occurrences[:page_size]],
             next_page_token=str(millis_from_dt(occurrences[-1].end_time)) if len(occurrences) > page_size else None,
+            page_number=page_number if request.page_number else None,
+            total_items=total_items if request.page_number else None,
```
aapeliv: Please return this regardless of if `page_number` is set, given that it's computed anyway.

app/backend/src/couchers/servicers/search.py
```diff
@@ -660,10 +665,14 @@ def EventSearch(self, request, context, session):
...
+        total_items = session.execute(select(func.count()).select_from(statement.subquery())).scalar()
+        # Apply pagination by page number
+        statement = statement.offset(offset).limit(page_size) if request.page_number else statement.limit(page_size + 1)
         occurrences = session.execute(statement).scalars().all()
 
         return search_pb2.EventSearchRes(
             events=[event_to_pb(session, occurrence, context) for occurrence in occurrences[:page_size]],
             next_page_token=(str(millis_from_dt(occurrences[-1].end_time)) if len(occurrences) > page_size else None),
+            page_number=page_number if request.page_number else None,
+            total_items=total_items if request.page_number else None,
```
aapeliv: ditto

app/proto/events.proto
```diff
@@ -280,12 +280,15 @@ message ListMyEventsReq {
 
   uint32 page_size = 5;
   string page_token = 6;
+  optional uint32 page_number = 9;
```
aapeliv: Everything is `optional`, no need for it in proto3.

Please change to

```proto3
oneof pagination {
  string page_token = 6;
  uint32 page_number = 9;
}
```

app/proto/events.proto
```diff
@@ -280,12 +280,15 @@ message ListMyEventsReq {
...
 }
 
 message ListMyEventsRes {
   repeated Event events = 1;
 
   string next_page_token = 2;
+  optional uint32 page_number = 3;
```
aapeliv: Do you need to return the page number if it's just what was in the input?

app/proto/search.proto
```diff
@@ -198,10 +198,13 @@ message EventSearchReq {
 
   uint32 page_size = 10;
   string page_token = 11;
+  optional uint32 page_number = 18;
```
aapeliv: Again `oneof` thing

app/proto/search.proto
```diff
@@ -198,10 +198,13 @@ message EventSearchReq {
...
 }
 
 message EventSearchRes {
   repeated org.couchers.api.events.Event events = 1;
 
   string next_page_token = 2;
+  optional uint32 page_number = 3;
+  optional uint32 total_items = 4;
```
aapeliv: Please remove all `optional`s.

So proto2 syntax had this required/optional thing. proto3 has everything be optional. It's a bit silly, optional/required was a good thing, but they simplified the language. In proto3 if something's missing it gets a default value (e.g. 0, empty string, etc, depending on type). proto2 was more powerful, you could define the default if it was missing.

aapeliv (changes requested):
Looking good! I put in a couple small change requests.

In general I think pagination can cause some performance issues, but there shouldn't be that many events so it should be fine. But if we wanted to do this with e.g. users, we might need to think about it some more!

app/proto/events.proto
nabramow: Good point, probably not. Fixed.

app/proto/search.proto
nabramow: Thanks for the background! I've made the changes now and updated the tests.

app/proto/events.proto
```diff
@@ -280,12 +280,14 @@ message ListMyEventsReq {
 
   uint32 page_size = 5;
   string page_token = 6;
+  uint32 page_number = 9;
```
aapeliv: Please change to

```
oneof pagination {
  string page_token = 6;
  uint32 page_number = 9;
}
```

aka just wrap both inside the `oneof pagination`

app/proto/search.proto
```diff
@@ -198,10 +198,12 @@ message EventSearchReq {
 
   uint32 page_size = 10;
   string page_token = 11;
+  uint32 page_number = 18;
```
aapeliv: Same here

aapeliv (changes requested):
just one small thing!

app/proto/events.proto
nabramow: Okay just pushed this! Had to adjust the events.py and seach.py a bit for the oneof so the tests would still pass :-)

app/backend/src/couchers/servicers/events.py
```diff
@@ -978,9 +978,23 @@ def SetEventAttendance(self, request, context, session):
...
+        # Check which field in the 'pagination' oneof is set
+        pagination_type = request.WhichOneof("pagination")
         page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
-        # the page token is a unix timestamp of where we left off
-        page_token = dt_from_millis(int(request.page_token)) if request.page_token else now()
+
+        if pagination_type == "page_token":
+            # the page token is a unix timestamp of where we left off
+            page_token = dt_from_millis(int(request.page_token))
+            page_number = 1
```
aapeliv: I don't think these `page_number`s are used now outside the `pagination_type == "page_number"` case

app/backend/src/couchers/servicers/search.py
```diff
@@ -660,10 +673,13 @@ def EventSearch(self, request, context, session):
                 EventOccurrence.start_time.desc()
             )
 
-        statement = statement.limit(page_size + 1)
+        total_items = session.execute(select(func.count()).select_from(statement.subquery())).scalar()
+        # Apply pagination by page number
+        statement = statement.offset(offset).limit(page_size) if request.page_number else statement.limit(page_size + 1)
```
aapeliv, line 670: It seems `offset` is not defined if you have `pagination_type == "page_token"`

app/backend/src/couchers/servicers/events.py
nabramow: You are right, apologies. I realize I ran the tests before regenerating the protos and assumed the error was this.

app/backend/src/couchers/servicers/search.py
nabramow: I changed it back, turns out I did not need these changes.

aapeliv (approved):
LGTM if the tests pass!

PR #4951, opened 2024-10-06
Switch to getLiteUser where applicable on frontend

### LINKED ISSUE #4936: Use LiteUser in lots of places in the frontend
The frontend currently loads the user's whole profile in most cases, even if all it needs is the name + avatar. We discussed this at a few dev meetings, and for now I've implemented a `GetLiteUser` RPC that returns just the minimal info needed in the majority of cases in the backend. The main places where this does NOT work are the profile itself, and the map (that's a separate exercise).

In addition to the lite user RPC, there is also the `GetLiteUsers` RPC that gets many users: this allows fetching up to 200 users at a time in one backend call, which should significantly reduce the number of calls required e.g. when opening the map page.

This issue requires:
1. Creating a parallel `useLiteUser(s)` hook similat to the existing `useUser(s)`  hook, and
2. Updating many uses of the latter hook to the new former hook.

https://github.com/Couchers-org/couchers/blob/711efe78776efad8ccb331fda32eb47d6b1af13c/app/proto/api.proto#L22-L28

nabramow:
This PR switches out `useUser` and `useUsers` with `useLiteUser` and `useLiteUsers`, which returns a much smaller user object. Backend code was already merged for this.

**HOW TO TEST**
- Check anywhere that it getLiteUser or getLiteUsers effects. This is anywhere you just have user avatar, name, etc (almost everywhere except profile and map). So references, chat, comments on communities and events, friend requests, host requests, etc.

Closes #4936 

**Web frontend checklist**
- [ x ] All tests pass
- [ N/A ] Checked Desktop, Mobile and Tablet screen sizes


bakeiro:
wow such a nice job, looking good!

bakeiro:
I'll test to see if everything works good on the page

nabramow:
@aapeliv @bakeiro just checking up on this one. I did click around the app already for this but if anyone wants to take a second look I couldn't check the reference user info as I don't have any references in my test account (and can't add as I'm blocked by not having a profile photo and doesn't let me upload one on localdev).

Although we had pretty good tests for most of these components so hopefully it would catch anything.

bakeiro:
sorry, totally forget, if the test was good then is a ready to go for me, I´ll check it briefly, if all good I approve it :)

nabramow:
@bakeiro  @aapeliv Okay whoops realizing I missed a big part of this by not using the GetLiteUsers new api so give me some time to adjust the hooks and re-write the tests!

darrenvong:
This is exciting to see - the original `useUsers` hook implementation has been a hack all along to work around the limitation of not being able to get multiple users in one trip! I'd go as far to say a good measure of this new API doing its work is that there should be no `useEffect`s and the weird workaround with the ref to store the list of user IDs once you've fully switched over to the new RPC!

nabramow:
> This is exciting to see - the original `useUsers` hook implementation has been a hack all along to work around the limitation of not being able to get multiple users in one trip! I'd go as far to say a good measure of this new API doing its work is that there should be no `useEffect`s and the weird workaround with the ref to store the list of user IDs once you've fully switched over to the new RPC!

Thanks for this, I was getting hung up on refactoring the existing hook but this made me realize I should just re-write and use my own common sense haha, since the whole api call works differently.

Done now, re-wrote the tests. Could be missing some test cases, but I have most of the old ones covered (except the invalidate param since we don't need to pass that anymore).

app/web/features/connections/friends/useFriendRequests.test.tsx
```diff
@@ -172,13 +164,10 @@ describe("when the listFriendRequests succeeds", () => {
...
   });
 
-  it("returns isError as true with the error and missing friend data in request object of response if some getUser queries failed", async () => {
-    getUserMock.mockImplementation((userId: string) => {
-      return userId === "3"
-        ? Promise.reject(new Error(`Error fetching user ${userId}`))
-        : getUser(userId);
-    });
-    jest.spyOn(console, "error").mockReturnValue(undefined);
+  it("returns isError as true with the error if some getLiteUsers query failed", async () => {
```
darrenvong:
```suggestion
  it("returns isError as true with the error if the getLiteUsers query failed", async () => {
```
With the refactored endpoint, isn't this always going to be a single query? The phrasing in the old test ("some getUser queries") is alluding to the fact it was making multiple requests which I believe is no longer possible?

app/web/features/connections/friends/useFriendRequests.ts
```diff
@@ -41,10 +45,10 @@ export default function useFriendRequests(
           ...friendRequest,
           friend: usersData.get(friendRequest.userId),
         }))
-      : void 0;
+      : [];
 
   return {
-    data,
+    data: isLoading ? undefined : data,
```
darrenvong: Wouldn't it be simpler to have the `undefined` (or `void 0`, we were going through a phase of using all the cool stuff we knew about JS lol, as in practice `undefined` will never be re-declared to warrant the need to use `void 0` and it only makes this more obscure perhaps 😅 ) on line 48 as before?

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -138,8 +138,8 @@ export default function NewHostRequest({
         )}
       </Typography>
       {error && <Alert severity="error">{error.message}</Alert>}
-      {hostError ? (
-        <Alert severity={"error"}>{hostError}</Alert>
+      {hostError?.message ? (
+        <Alert severity={"error"}>{hostError?.message}</Alert>
```
darrenvong:
```suggestion
      {hostError ? (
        <Alert severity={"error"}>{hostError.message}</Alert>
```
Would it not be simpler to do this so the check on `hostError` narrows the type and eliminates the need for optional chaining?

app/web/features/queryKeys.ts
```diff
@@ -15,6 +15,12 @@ export function userKey(userId?: number) {
   return userId === undefined ? "user" : ["user", userId];
 }
 
+export function liteUserKey(userId?: number) {
+  return userId === undefined ? "liteUser" : ["liteUser", userId];
+}
```
darrenvong, line 20: This is one of those where it's been a while so I don't remember why we did it like this... but upon fresh perspective, it feels weird/unintuitive to allow `userId` to be undefined? 🤔 Or is this perhaps a way to ensure fetching deleted users won't crash the hook?

app/web/features/userQueries/useLiteUsers.ts
```diff
@@ -0,0 +1,46 @@
...
+          ])
+        );
+
+  return {
+    ...query,
+    data: usersById,
+  };
+}
+
+// should is returned when stale if subsequent refetch queries fail
```
darrenvong: I have trouble understanding what this comment is trying to explain 🙈  please elaborate?

app/web/features/userQueries/useLiteUsers.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      // Ensure the query function was NOT called
+      expect(mockGetLiteUsers).not.toHaveBeenCalled();
+    });
+
+    it("Should not run the query function when ids is undefined", () => {
+      const ids: (number | undefined)[] = [undefined];
+
+      renderHook(() => useLiteUsers(ids), { wrapper });
+
+      expect(mockGetLiteUsers).not.toHaveBeenCalled();
```
darrenvong: If someone screw something up in the future and this IS being called by accident, wouldn't this not catch the problem since we're not awaiting for next update? i.e. you could write the same logic in the `Should return loading state correctly` test in the first part and this would still pass (at least from my understanding of reading the code)

app/web/features/userQueries/useLiteUsers.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      const { result, waitForNextUpdate } = renderHook(
+        () => useLiteUsers([1, 2, 3]),
+        { wrapper: sharedClientWrapper }
+      );
+
+      await waitForNextUpdate();
+
+      expect(result.current).toEqual(
+        expect.objectContaining({
+          data: undefined,
```
darrenvong: Should this NOT be `undefined` based on the description of the test? 🤔

app/web/features/userQueries/useLiteUsers.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    it("Uses cached data instead of refetching", async () => {
+      const { result } = renderHook(() => useLiteUsers(ids), {
+        wrapper: sharedClientWrapper,
+      });
+
+      expect(mockGetLiteUsers).not.toHaveBeenCalled();
+
+      expect(result.current.isFetching).toBe(false);
+      expect(result.current.isLoading).toBe(false);
+    });
```
darrenvong, line 206: I think this should also assert `result.current.data` exist to prove it has data and is reading from the cache

app/web/features/userQueries/useLiteUsers.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      await waitForNextUpdate();
+
+      expect(result.current.isLoading).toBe(false);
+      expect(result.current.error).toBe(error);
+      expect(result.current.data).toBeUndefined();
+    });
+
+    it("Should not run the query function when id is undefined", () => {
+      // @ts-expect-error
+      renderHook(() => useLiteUser(), { wrapper });
```
darrenvong:
```suggestion
      renderHook(() => useLiteUser(undefined), { wrapper });
```
A bit more verbose but then you won't need the `ts-expect-error` comment? I think it's because it's defined with `| undefined` instead of `?:` in the type annotation, so it becomes a required argument even when the value is `undefined`

app/web/features/userQueries/useLiteUsers.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+    it("Uses cached data instead of refetching", async () => {
+      const { result } = renderHook(() => useLiteUser(3), {
+        wrapper: sharedClientWrapper,
+      });
+
+      expect(mockGetLiteUser).not.toHaveBeenCalled();
+      expect(await sharedClient.getQueryData(liteUserKey(3))).toEqual(
+        mockLiteUsers[2]
+      );
```
darrenvong: I think it'd be better to assert on `result.current.data` to prove the hook works here rather than interacting on the `sharedClient`, since we just set the query data in the `beforeEach` so I'd expect to have data there. We're testing React Query is doing its job here instead of the hook otherwise?

darrenvong (commented):
I haven't tested so all comments are based on the code. It looks like a lot of comments (of small things) but I want to say this is generally great, given the minimal changes you're able to make across the board!

app/web/features/connections/friends/useFriendRequests.test.tsx
nabramow: Ah yeah copy pasta good catch!

app/web/features/connections/friends/useFriendRequests.ts
nabramow: Was wondering about the `void 0` haha. Okay so this made me find a lot of mistakes I made, but neither your comment or mine ended up being the answer 😅.

I had to adjust some behavior since I wanted `useLiteUsers` to return `undefined` if there's no data (instead of empty Map, a holdover from earlier). I also needed to return an empty array for `useFriendRequests` here if `useLiteUsers` returned `undefined` data.

I was incorrectly hacking it before I now realize, as I had something wrong in `userLiteUsers`.

app/web/features/profile/view/NewHostRequest.tsx
nabramow: Oh yeah that makes sense!

app/web/features/queryKeys.ts
nabramow: It seems we pass ids from other queries that could be undefined (if the query fails I guess). For example in app/web/features/communities/discussions/DiscussionPage.tsx.

I agree it's kinda silly but on first glance we'd need to do a bunch of refactoring for this change. Probably makes more sense in a separate PR IMO.

app/web/features/userQueries/useLiteUsers.test.tsx
nabramow: Oh wow good catch! I've added in a waitFor until `isLoading` is false!

app/web/features/userQueries/useLiteUsers.test.tsx
nabramow: I ended up going about this another way, by calling the hook twice and checking that the second time used cache data (rather than setting it manually in the queryClient).

app/web/features/userQueries/useLiteUsers.test.tsx
nabramow: I totally redid these tests after reading through the `useQuery` docs more. The way I had it wasn't actually using the cached data if there was an error as I had to set `retry: false` in the `useQuery` hook for that. In the new version I am checking that data is not empty though :-).

app/web/features/userQueries/useLiteUsers.test.tsx
nabramow: Yes, you are correct! Fixed!

app/web/features/userQueries/useLiteUsers.test.tsx
nabramow: Fixed!

app/web/features/userQueries/useLiteUsers.ts
nabramow: Ahah sorry I copied that from the old file. I fixed it now to say "React Query typically retains the last successful data until the next successful fetch". This took me ages to figure out but seems to be how the `useQuery` hook is supposed to work?!

app/web/features/userQueries/useLiteUsers.ts
```diff
@@ -0,0 +1,64 @@
...
+      const result = service.user.getLiteUsers(nonFalseyIds);
+      return result;
+    },
+    staleTime: userStaleTime,
+    enabled: nonFalseyIds.length > 0, // run only if there are valid liteUserIds
+    retry: (failureCount, error) => {
+      // don't retry if the user isn't found
+      return (
+        error.code !== StatusCode.NOT_FOUND && failureCount < reactQueryRetries
+      );
```
nabramow, line 25: I added this in as I realized it was probably a good idea, no need to call again if user not found, and defining the retryQueries allows me to reference that in the test (we seem to be doing this elsewhere).

nabramow:
@darrenvong Thanks so much for the thorough code review! I did a bit of refactoring as your comments made me realize I was not fully understanding how the `useQuery` hook should be working. It's ready for re-review now!

bakeiro (approved):
Code & testing looks great

PR #4956, opened 2024-10-06
docs: swap to curl for downloading protos in frontend repo

darrenvong:
A tiny PR to swap to using the curl equivalent command for downloading protos for the frontend.

According to Claude:
> curl comes pre-installed on macOS and are available by default in many popular Linux distributions

So for lazy people like me who don't want to install another tool or create a temp Docker container to fetch the files, this feels useful. Hopefully not too controversial of a change for making things easier for others (less pre-requisite download steps)!

**Backend checklist**
N/A

**Web frontend checklist**
N/A

### DIFF
--- a/app/web/readme.md
+++ b/app/web/readme.md
@@ -9,7 +9,7 @@ Communication with the backend is via [protobuf messages](https://github.com/pro
 We recommend `nvm` (the [node version manager](https://github.com/nvm-sh/nvm)). You can install it with:
 
 ```sh
-wget -qO- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash
+curl -sL https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.1/install.sh | bash
 ```
 
 **You will need to restart your terminal before `nvm` becomes available.**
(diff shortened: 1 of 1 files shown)


app/web/readme.md
```diff
@@ -43,6 +44,8 @@ npm install --global yarn
 
 ## Download & extract the latest protos
 wget -qO- https://develop--protos.preview.couchershq.org/ts.tar.gz | tar xz
+## Or run the following (if you don't have wget or don't want to install it)
+curl -SL https://develop--protos.preview.couchershq.org/ts.tar.gz | tar xz
```
aapeliv: Do you want to just switch out the `wget` to `curl` to keep it simple? I'd be happy with that, I just thought wget is more common!

Otherwise please comment this line: I'd like this code snippet to just be runnable as-is so experienced devs can just look at the non-commented lines and see what's going on.

darrenvong: I think given how common `curl` is, it should be safe to swap out, I'll do that!

app/web/readme.md
```diff
@@ -24,6 +24,7 @@ If you don't want `nvm` to be run when you open your shell, you can add `--no-us
 On macOS you can alternatively install dependencies via homebrew. You can get nvm with `brew install nvm`, or `nodejs` directly with `brew install node@20`.
 
 You also need `git`, `wget`, and `tar`, they should be preinstalled or be available from the usual places.
```
aapeliv: Please also update this line and any other instances!

darrenvong: I'll leave the ones that are actually in build scripts since they are already working, even though things shouldn't break in theory... but since people shouldn't need to interact directly with those during dev, I think it's fine?

aapeliv (approved):
LGTM!

PR #4992, opened 2024-10-15
Backend/feature/user search mega issue

Krishnag09:
added filter for language qualifying [fluent, conversational] for language abilities values as deemed to return user profiles.  

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/backend/readme.md
```diff
@@ -9,19 +9,19 @@ You can run the whole thing through Docker and docker-compose (see the readme in
 You can run all tests in docker with the following command, executed in the `app` folder:
 
 ```sh
-docker compose -f docker-compose.test.yml up --build
+docker-compose -f docker-compose.test.yml up --build
```
aapeliv: `compose` is now a sub-command of `docker`. If you update docker you should be able to do `docker compose`.

app/backend/readme.md
```diff
@@ -9,19 +9,19 @@ You can run the whole thing through Docker and docker-compose (see the readme in
...
 1. If you haven't already, run the first two steps from the [readme in the `app/` folder](https://github.com/Couchers-org/couchers/blob/develop/app/readme.md).
 
-**Remember** that if you've made changes to any *.proto files, you may need to re-compile the protocol buffers (step 2).
+**Remember** that if you've made changes to any \*.proto files, you may need to re-compile the protocol buffers (step 2).
 
 2. Make sure the postgres_tests container is running:
 
 ```sh
-docker compose -f docker-compose.test.yml up postgres_tests
+docker-compose -f docker-compose.test.yml up postgres_tests
```
aapeliv: Ditto

app/backend/readme.md
```diff
@@ -45,7 +45,7 @@ export DATABASE_CONNECTION_STRING=postgresql://postgres:06b3890acd2c235c41be0bbf
 6. Run `pytest` in the `app/backend/src/` folder.
 
 ```sh
-cd src
+cd src..
```
aapeliv: I think this should be just `cd src`?

app/backend/src/couchers/servicers/search.py
```diff
@@ -55,6 +58,11 @@
 TRI_SIMILARITY_THRESHOLD = 0.6
 TRI_SIMILARITY_WEIGHT = 5
 
+conversational_fluent_filter_values = [
```
aapeliv: Could you add a comment here as to what this means?

app/backend/src/couchers/servicers/search.py
```diff
@@ -55,6 +58,11 @@
...
+    fluency2sql[api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL],
+    fluency2sql[api_pb2.LanguageAbility.Fluency.FLUENCY_FLUENT],
```
aapeliv: No need to go through `fluency2sql` here, you can just have `LanguageFluency.conversational` and `LanguageFluency.fluent` from `models.py` (already imported).

app/backend/src/couchers/servicers/search.py
```diff
@@ -444,6 +453,22 @@ def UserSearch(self, request, context, session):
             statement = statement.where(
                 User.parking_details.in_([parkingdetails2sql[det] for det in request.parking_details_filter])
             )
+            # limits/default could be handled on the front end as well
+        min_age = request.age_min.value if request.HasField("age_min") else 18
+        max_age = request.age_max.value if request.HasField("age_max") else 200
+
+        statement = statement.where((User.age >= min_age) & (User.age <= max_age))
```
aapeliv: It might be better to have these as `if request.HasField():` then only include the `where` in case they are specified. Adding extra `where`s can possibly confuse the query planner?

I don't think it's a big deal!

app/backend/src/tests/test_search.py
```diff
@@ -158,6 +159,54 @@ def test_user_filter_meetup_status(db):
...
+        existing_language = session.query(Language).filter_by(code="dan").one_or_none()
+
+        if not existing_language:
+            # Insert the language record if it doesn't exist
+            session.add(Language(code="dan", name="Danish"))
+
+        existing_language = session.query(Language).filter_by(code="ces").one_or_none()
+
+        if not existing_language:
+            session.add(Language(code="ces", name="Czech"))
```
aapeliv: No need to do this: these are defined in `languages.json` and are included in the `db` test dependency

app/proto/search.proto
```diff
@@ -205,3 +218,8 @@ message EventSearchRes {
 
   string next_page_token = 2;
 }
+
+message SearchRequest {
+    int32 min_age = 6;
+    int32 max_age = 7;
+}
```
aapeliv: I think this is not needed? Maybe leftover from earlier?

app/proto/search.proto
```diff
@@ -101,6 +101,18 @@ message SearchRes {
   string next_page_token = 2;
 }
 
+enum LanguageFluency {
+    BEGINNER = 0;
+    CONVERSATIONAL = 1;
+    FLUENT = 2;
+}
```
aapeliv: Please import `api.proto`, then you should be able to use `org.couchers.api.core.LanguageAbility.Fluency` I believe.

app/backend/src/couchers/servicers/search.py
aapeliv: Actually it seems this is unused now?

I think it's better the frontend deals with what fluency level should be matched in the backend.

app/backend/src/couchers/servicers/search.py
```diff
@@ -444,6 +453,22 @@ def UserSearch(self, request, context, session):
...
+
+        if len(request.language_ability_filter) > 0:
+            for ability_filter in request.language_ability_filter:
+                fluency_enum_value = LanguageFluency(ability_filter.fluency)
```
aapeliv: You should use the `fluency2sql` here

aapeliv (commented):
OK sorry, lots of comments that don't all make sense. Please ask further questions as needed! Also apologies for the slow review time!

app/backend/src/tests/test_search.py
Krishnag09: when i dont add the condition, it throws me an foreign key violation error. something like this 

`E       sqlalchemy.exc.IntegrityError: (psycopg2.errors.ForeignKeyViolation) insert or update on table "language_abilities" violates foreign key constraint "fk_language_abilities_language_code_languages"
E       DETAIL:  Key (language_code)=(dan) is not present in table "languages".
E       
E       [SQL: INSERT INTO language_abilities (user_id, language_code, fluency) SELECT p0::BIGINT, p1::VARCHAR, p2::languagefluency FROM (VALUES (%(user_id__0)s, %(language_code__0)s, %(fluency__0)s, 0), (%(user_id__1)s, %(language_code__1)s, %(fluency__1)s, 1)) AS imp_sen(p0, p1, p2, sen_counter) ORDER BY sen_counter RETURNING language_abilities.id, language_abilities.id AS id__1]
E       [parameters: {'fluency__0': 'fluent', 'user_id__0': 1, 'language_code__0': 'dan', 'fluency__1': 'conversational', 'user_id__1': 2, 'language_code__1': 'ces'}]
E       (Background on this error at: https://sqlalche.me/e/20/gkpj)
`

app/backend/src/couchers/servicers/search.py
Krishnag09: was trying to have a default value assigned as 18 min and 200 max when they are empty.

app/backend/src/tests/test_search.py
aapeliv: Ahh sorry, my bad. Try using some from this list:

https://github.com/Couchers-org/couchers/blob/0fb8910a3bc0c3cf46b8251fdda7d1e7b12d9e78/app/backend/src/tests/test_fixtures.py#L178

Krishnag09: done, thanks.

app/backend/src/couchers/servicers/search.py
```diff
@@ -444,6 +452,29 @@ def UserSearch(self, request, context, session):
...
+            for ability_filter in request.language_ability_filter:
+                fluency_sql_value = fluency2sql.get(ability_filter.fluency)
+
+                if fluency_sql_value is None:
+                    continue
+
+                statement = statement.where(
+                    User.language_abilities.any(
+                        (LanguageAbility.language_code == ability_filter.code)
+                        & (LanguageAbility.fluency.in_(conversational_fluent_filter_values))
```
aapeliv:
```suggestion
                        & (LanguageAbility.fluency.in_(fluency_sql_value))
```

app/backend/src/couchers/servicers/search.py
```diff
@@ -55,6 +57,11 @@
 TRI_SIMILARITY_THRESHOLD = 0.6
 TRI_SIMILARITY_WEIGHT = 5
 
+conversational_fluent_filter_values = [
+    fluency2sql[api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL],
+    fluency2sql[api_pb2.LanguageAbility.Fluency.FLUENCY_FLUENT],
+]
+
```
aapeliv: See comment above, this will no longer be needed!

app/backend/src/tests/test_search.py
```diff
@@ -158,6 +160,45 @@ def test_user_filter_meetup_status(db):
         assert [result.user.user_id for result in res.results] == [user_does_not_want_to_meet.id]
 
 
+def test_user_filter_language(db):
+    """
+    Make sure the language filter returns the rigth profiles
+    """
+    recreate_database()
```
aapeliv: This should not be needed, the `db` fixture does it already.

app/backend/src/tests/test_search.py
```diff
@@ -158,6 +160,45 @@ def test_user_filter_meetup_status(db):
...
+                api_pb2.LanguageAbility(code="jpn", fluency=api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL)
+            ]
+        )
+
+        result_1 = api.UserSearch(search_request_1)
+        result_2 = api.UserSearch(search_request_2)
+
+        assert [result.user.user_id for result in result_1.results] == [user_with_german_fluent.id]
+        assert [result.user.user_id for result in result_2.results] == [user_with_japanese_conversational.id]
+
```
aapeliv, line 211: If you'd like, it'd be good to have a test where the user is, say `beginner` in a language but you search for `conversational` and they wouldn't show up.

app/proto/search.proto
```diff
@@ -101,6 +101,14 @@ message SearchRes {
   string next_page_token = 2;
 }
 
+
+message LanguageAbility {
+    int64 user_id = 1;          // Corresponds to the user ID (optional, may not be needed for search)
+    string language_code = 2;   // ISO code for the language (e.g., "dan" for Danish)
+    org.couchers.api.core.LanguageAbility.Fluency fluency = 3; // The language fluency level
+}
```
aapeliv: Not needed anymore.

app/proto/search.proto
```diff
@@ -123,17 +131,19 @@ message UserSearchReq {
   repeated org.couchers.api.core.SleepingArrangement sleeping_arrangement_filter = 7;
   repeated org.couchers.api.core.ParkingDetails parking_details_filter = 8;
   repeated org.couchers.api.core.MeetupStatus meetup_status_filter = 32;
+  repeated org.couchers.api.core.LanguageAbility language_ability_filter = 34;
+  repeated org.couchers.api.core.LanguageAbility.Fluency fluency = 35;
```
aapeliv: This fluency line (field 35) is not needed anymore.

aapeliv (changes requested):
Hey, sorry for the really long delay!

This is looking really close, just one minor change and there's a few unused bits of code here and there. But should be ready to merge otherwise!

aapeliv:
@Krishnag09 let me know when you've addressed all review comments and this is good to go!

app/backend/src/couchers/servicers/search.py
Krishnag09: i think its better to implement this is backend rather than someone explaining everything again to front end. IN that case should i create more like sql objects based version of conversational_fluent_filter_values ?

app/backend/readme.md
```diff
@@ -29,7 +29,7 @@ This will spin up a complete copy of the database, backend, and proxies needed t
 
 This will not currently run the frontend, to do that, please follow the instructions in [app/web/readme.md](../web/readme.md) under *Quick Start*, then *Running against a local backend*.
 
-### Running tests in docker
+**Remember** that if you've made changes to any \*.proto files, you may need to re-compile the protocol buffers (step 2).
```
aapeliv: Please put the heading back

app/backend/readme.md
```diff
@@ -51,7 +51,9 @@ source venv/bin/activate
 pip install -r requirements.txt
 
 ## Now run the test
-DATABASE_CONNECTION_STRING="postgresql://postgres:06b3890acd2c235c41be0bbfe22f1b386a04bf02eedf8c977486355616be2aa1@localhost:6544/postgres" pytest src/tests/
+DATABASE_CONNECTION_STRING="postgresql://postgres:06b3890acd2c235c41be0bbfe22f1b386a04bf02eedf8c977486355616be2aa1@localhost:6544/postgres"
```
aapeliv: Do you need an `export DATABASE_...` if you are doing it in a separate command?

app/backend/readme.md
```diff
@@ -70,6 +72,13 @@ DATABASE_CONNECTION_STRING="postgresql://postgres:06b3890acd2c235c41be0bbfe22f1b
...
 3. Re-run `docker compose up --build`
 
+### Q: I'm seeing issues with proto!
+
+**A**: Generate protos using
+
+```sh
+docker run --pull always --rm -w /app -v $(pwd):/app registry.gitlab.com/couchers/grpc ./generate_protos.sh
+```
 If you have **any trouble**, send Aapeli a message on Slack. He's more than happy to spend a bit of time helping you set things up!
```
aapeliv: Please put this above the new Q (at the end of the old one)

app/backend/readme.md
```diff
@@ -80,4 +89,5 @@ If you are using the live dev api ("next"/staging), it will send you real emails
 
 ### Q: How do I receive emails like the signup confirmation email in local dev?
 
-**A**: We run [MailDev](https://github.com/maildev/maildev) with the docker compose setup: it will receive emails and let you browse them. To view emails, visit <http://localhost:1080>.
+A: First doublecheck what port the DB is listening on - run `docker compose up postgres` and it should say something like `listening on IPv6 address "::", port 6545`. Then doublecheck you have the right password. There are TWO passwords - one for the test db and one for the normal db! See app/postgres.dev.env and app/postgres.test.env
```
aapeliv: Please check this bit. I don't think you meant to remove the answer.

app/backend/src/couchers/servicers/api.py
```diff
@@ -67,6 +67,19 @@
...
+    api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL: LanguageFluency.conversational,
+    api_pb2.LanguageAbility.Fluency.FLUENCY_FLUENT: LanguageFluency.fluent,
+}
+
+fluency2api = {
+    None: api_pb2.LanguageAbility.Fluency.FLUENCY_UNKNOWN,
+    LanguageFluency.beginner: api_pb2.LanguageAbility.Fluency.FLUENCY_BEGINNER,
+    LanguageFluency.conversational: api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL,
+    LanguageFluency.fluent: api_pb2.LanguageAbility.Fluency.FLUENCY_FLUENT,
+}
```
aapeliv: Aren't these already about 70 lines down?

app/backend/src/couchers/servicers/search.py
```diff
@@ -55,6 +58,11 @@
 TRI_SIMILARITY_THRESHOLD = 0.6
 TRI_SIMILARITY_WEIGHT = 5
 
+conversational_fluent_filter_values = [
+    LanguageFluency.conversational,
+    LanguageFluency.fluent,
+]
```
aapeliv: This is no longer required, please remove

app/backend/src/tests/test_search.py
```diff
@@ -13,7 +13,9 @@
     db,
     events_session,
     generate_user,
+    recreate_database,
```
aapeliv: I don't think you ened this import?

app/backend/src/tests/test_search.py
```diff
@@ -158,6 +160,51 @@ def test_user_filter_meetup_status(db):
...
+            search_pb2.UserSearchReq(
+                language_ability_filter=[
+                    api_pb2.LanguageAbility(
+                        code="deu",
+                        fluency=api_pb2.LanguageAbility.Fluency.FLUENCY_CONVERSATIONAL,
+                    )
+                ]
+            )
+        )
+        assert [result.user.user_id for result in res.results] != [user_with_german_beginner.id]
```
aapeliv: I think better to asset that it's `== []`

app/proto/api.proto
```diff
@@ -5,8 +5,7 @@ package org.couchers.api.core;
 import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
 import "google/protobuf/wrappers.proto";
-
-import "annotations.proto";
+import  "annotations.proto";
```
aapeliv: Please undo this change

app/proto/search.proto
```diff
@@ -101,6 +101,8 @@ message SearchRes {
   string next_page_token = 2;
 }
 
+
+
```
aapeliv: And these lines

aapeliv (changes requested):
This is good except for a couple of cleanup bits!

app/backend/src/tests/test_search.py
Krishnag09: you wrote this part few weeks ago - If you'd like, it'd be good to have a test where the user is, say beginner in a language but you search for conversational and they wouldn't show up.

aapeliv: Yes I think the test is good: I just meant that in a test it’s better to test that the list is exactly what you expect, not that individual properties. If I understand correctly, the search should output an empty list since that user isn’t returned, right?

app/backend/readme.md
aapeliv: It seems you didn't put the heading back ("Running tests in docker")

aapeliv (approved):
LGTM

aapeliv:
Please merge when you are happy with this @Krishnag09

PR #5007, opened 2024-10-19
Final - New Events Page 

### LINKED ISSUE #4784: Create new events page - Part I

### LINKED ISSUE #4823: Create all events page

nabramow:
This is the final PR for the new events feature. The individual pieces have all already been reviewed in smaller PRs, so here we're looking for any final design or usability issues before it goes to prod.

Closes #4823 

**Web frontend checklist**
- [ x ] All tests pass


jesseallhands:
Should we change the "communities" and "online" pills to "in-person" and "online"? That's the primary differentiator, right? I don't think we have any in-person events that wouldn't be associated with a community.

Screenshot_20241019_103726_Firefox

bakeiro:
> Should we change the "communities" and "online" pills to "in-person" and "online"? That's the primary differentiator, right? I don't think we have any in-person events that wouldn't be associated with a community.

I think there is a difference, because the community pill shows the events only of the communities that I belong, not just any community

jesseallhands:
>I think there is a difference, because the community pill shows the events only of the communities that I belong, not just any community

In that case the pill should be "my community" and not just "community".

bakeiro:
> > I think there is a difference, because the community pill shows the events only of the communities that I belong, not just any community
> 
> In that case the pill should be "my community" and not just "community".

gotcha... guess that `my communities` it's even better maybe? :D

bakeiro:
not the UI guy here but
image

I feels this text a bit small... right? (page wasn't zoomed out), if you feel the same, maybe we can increase it :)

bakeiro:
about the code, was something new added to review? (this is just the merged of the reviewed PRs right?)

nabramow:
> about the code, was something new added to review? (this is just the merged of the reviewed PRs right?)

No, no new functionality! This is just where all those PRs merged into.

bakeiro:
[suggestion here, no needed to approve the PR or something like that]

I was thinking about the events component in the community page... maybe makes sense to standarize these components? (make them more similar, re-using the same component...)

nabramow:
> [suggestion here, no needed to approve the PR or something like that]
> 
> I was thinking about the events component in the community page... maybe makes sense to standarize these components? (make them more similar, re-using the same component...)

I agree! Let's do it in a separate ticket? I'd want more feedback about how the pagination should work for that and any filters, and I think this PR is big enough.

nabramow:
Okay I think I addressed everything now!

- Changed 'Communities' -> 'My Communities' @jesseallhands 
- Made the font size for the event info a little bit bigger @bakeiro 
- Put the pills rendering logic in the body of the component instead of a render function @darrenvong 

Is there anything else?

bakeiro:
I'm just wondering...

how the event location search works? its by area? (2 coordinates and then search ~50km around it?) if that's the case what happens when user types a country/state/region, would be only the center of that region? (circle area instead of a polygon)

if that's the case we should allow only filter by city... (or smaller 'nodes') I can create another ticket for this if it's needed

let me research about this and create a proper ticket... I already faced similar issues while working on the map

bakeiro:
> I'm just wondering...
> 
> how the event location search works? its by area? (2 coordinates and then search ~50km around it?) if that's the case what happens when user types a country/state/region, would be only the center of that region? (circle area instead of a polygon)
> 
> if that's the case we should allow only filter by city... (or smaller 'nodes') I can create another ticket for this if it's needed
> 
> let me research about this and create a proper ticket... I already faced similar issues while working on the map

BTW, this doens't block the PR, since it's already too big

darrenvong:
I'm not quite sure how it's related... but on the prod couchers site, I noticed the "remember me" switch label doesn't render on a different line like in this branch:
Screenshot 2024-10-22 at 01 05 09

darrenvong:
One more quirk I noticed - it's unexpected to me that when you got a location filter on (by typing something in the input), then adding and removing "my communities" and "online" filters would reset it. See recording:

https://github.com/user-attachments/assets/a2f1c96c-0408-4a4e-bf9a-19edf4cae171

In the scenario shown, when I click back to filter on events based on "my communities", I would expect it to only show the event in Sydney since that's what I got in the location input still filled in.

Interestingly, pressing only the searching glass button is enough to put the filter back on without needing to trigger another nomatim search -> select it from the location list before it doing anything, which suggests me that the filter might be still stored in state somewhere but not correctly sent as filters to the backend with the way I used it?

https://github.com/user-attachments/assets/0f3514c3-8c27-47d6-aec5-76810ac2f767

nabramow:
> I'm just wondering...
> 
> how the event location search works? its by area? (2 coordinates and then search ~50km around it?) if that's the case what happens when user types a country/state/region, would be only the center of that region? (circle area instead of a polygon)
> 
> if that's the case we should allow only filter by city... (or smaller 'nodes') I can create another ticket for this if it's needed
> 
> let me research about this and create a proper ticket... I already faced similar issues while working on the map

So if you look at `EventSearchReq` or you look at the backend function `def EventSearch` there's a few options for how search can work.

@bakeiro You can "search in rectangle", that's what I chose here. With than you give it a `RectArea` which is a latitude min, latitude max, longitude min and longitude max to search in. So you can calculate that by taking the exact location of the search coordinates and adding some "buffer" around them.

You can also "search in area", with that one you give a latitude, longitude and then a radius.

I don't think it's bad to also allow filter by country or state? What if you live in a tiny country like Monaco or something, haha, or a small US state. Maybe you're willing to drive a bit if you're not in a city? What's the downside or making it flexible?

nabramow:
> I'm not quite sure how it's related... but on the prod couchers site, I noticed the "remember me" switch label doesn't render on a different line like in this branch: Screenshot 2024-10-22 at 01 05 09

Wow that was a shockingly good catch! I've removed all the changes to the Switch now. I was originally using it here, but ended up changing the design anyway so it's back to how it was before.

nabramow:
> One more quirk I noticed - it's unexpected to me that when you got a location filter on (by typing something in the input), then adding and removing "my communities" and "online" filters would reset it. See recording:
> 

@darrenvong I thiiiink I fixed this now. I believe the problem was I was clearing the search result after selecting the online only filter, but that wasn't clearing the visible search input value, so when you remove the filter it's gone but the text is still there.

For the life of me cannot find an accessible `onClear` function for the `LocationAutocomplete` component. Can't even see how that little clear `x` is getting there. So I just took out that invisible logic so at least what you see matches now.

Thank you for such a thorough test!

bakeiro:
> > I'm just wondering...
> > how the event location search works? its by area? (2 coordinates and then search ~50km around it?) if that's the case what happens when user types a country/state/region, would be only the center of that region? (circle area instead of a polygon)
> > if that's the case we should allow only filter by city... (or smaller 'nodes') I can create another ticket for this if it's needed
> > let me research about this and create a proper ticket... I already faced similar issues while working on the map
> 
> So if you look at `EventSearchReq` or you look at the backend function `def EventSearch` there's a few options for how search can work.
> 
> @bakeiro You can "search in rectangle", that's what I chose here. With than you give it a `RectArea` which is a latitude min, latitude max, longitude min and longitude max to search in. So you can calculate that by taking the exact location of the search coordinates and adding some "buffer" around them.
> 
> You can also "search in area", with that one you give a latitude, longitude and then a radius.
> 
> I don't think it's bad to also allow filter by country or state? What if you live in a tiny country like Monaco or something, haha, or a small US state. Maybe you're willing to drive a bit if you're not in a city? What's the downside or making it flexible?

The point is, I search "thailand" probably will look within an square, [so not respecting the frontiers], this can lead to overshow/undershow all the events wihtin that territory. We can solve this by using 'geometries' or polygons (or whatever are called) these polygons are available in open-street map, I was thinking about implementing then in map search (something that finally I didn't)

Without polygons, user will think "Oh I can search in a whole country/state" but in reality it's only either to small or to big

bakeiro:
this is what I meant, my suggestion here is:
In this iteration we should allow only the small 'nodes' (city, neighborhood, province), once we implement polygons search, we 
could search within any node regardless of its size.

Expected result when searching in a big node
image

I know how to implement the polygons, I can give you all the information if you need it.

bakeiro:
@nabramow 

image
France even shows results within the US and south Spain.. but Portugal seems works good, and Spain works "OK"... I wonder how that rectagle is calculated 🤔😅

I mean, it's better than don't having it? idk I can create another ticket or you can limit search nodes in this PR as you wish :)

nabramow:
>France even shows results within the US and south Spain.. but Portugal seems works good, and Spain works "OK"... I > > wonder how that rectagle is calculated 🤔😅
> 
> I mean, it's better than don't having it? idk I can create another ticket or you can limit search nodes in this PR as you wish :)

Oooo this was a great find, I didn't have the right test data to see it!

I've fixed this now. What I did: I checked if `isRegion = true`. If so, I search by query using location name (i.e. "France"). If it's not a region, I use the bbox rectangle search.

I tried it for Canada, Germany, France, Portugal, Thailand, United States. All seem to be working with my test data:

Screenshot 2024-10-23 at 8 13 47 PM

Screenshot 2024-10-23 at 8 14 00 PM

Screenshot 2024-10-23 at 8 14 10 PM

@bakeiro let me know if it works for you too!

app/web/service/search.ts
```diff
@@ -98,12 +98,20 @@ export async function EventSearch({
...
-    const location = new RectArea();
-    location.setLatMin(searchLocation?.bbox[1] || 0);
-    location.setLatMax(searchLocation?.bbox[3] || 0);
-    location.setLngMin(searchLocation?.bbox[0] || 0);
-    location.setLngMax(searchLocation?.bbox[2] || 0);
-    req.setSearchInRectangle(location);
+    // If it's a region (i.e. "France" or "United States") use query search by name
+    // This will search for events in the region by that name
+    if (searchLocation?.isRegion) {
+      req.setQuery(new StringValue().setValue(searchLocation.name));
```
bakeiro, line 104: this works by I'm trying to understand why... this simply does something like a regex? like if the searched name is inside the event location name?

bakeiro:
ok, with Valencia from Spain works and even from Valencia from Venezuela works... [they don't get mixed] I still don't understand how that `setQuery` works but it works

bakeiro (approved):

app/web/service/search.ts
bakeiro: explain before merging pls, might be useful for the map in the future (even if not want to understand it :D)

nabramow: Sure! It goes like this:

- We type a region like "France" in the search box
- That gets set in `LocationAutocomplete` component
- `LocationAutocomplete` component sends it in an api call to openstreetmap via the `useGeocodeQuery` hook
- Openstreetmap processes it, and gives us back the `GeocodeResult` object, standardizing the name, simplified name, bbox, lat and longitude, etc.
- We send that to our backend to to the `EventSearch` function to try to find matching events
- The backend checks what type of search: `search_in_area`, `search_in_rectangle` or `query` in this case. We define which search in the `search.ts` file with a grpc request
- We decide what to pass in the query. I am passing the location `name` passed back from the openstreet map response. We could also send `simplifiedName`, or something else.
- The backend checks if we want to search only the event title, or the event title, content AND address
- I chose the one with title, content and address
- Our event addresses have the city, state, country in them, that's why it works to search by country or city, as that's always in there. When you create an event, it also standardizes it with `LocationAutocomplete` via the openstreetmap api, so it's pretty safe that it will always match. So no chance someone makes a typo, leaves out the city or country, etc.

Screenshot 2024-10-24 at 7 30 02 PM

The ONLY thing I could think that might go wrong...since the query also searches the content of the event, if the user adds the name of a different city in the event description, that might make it show up in a search.

So say I make an event in Paris, France, but in the description I write, "David and I met in Madrid and decided to organize this event together", the event might also show up when searching Madrid.

I think that's a rare edge case though, so it's probably okay.

Looks like @spreeni wrote this code on the backend, please correct me if I got anything wrong!

PR #5033, opened 2024-10-24
Add manifest.json to HTML head

aapeliv:
This should make it possible to pin the app to the Home Screen on iOS and enable push notifications.

### DIFF
--- a/app/web/pages/_document.tsx
+++ b/app/web/pages/_document.tsx
@@ -8,8 +8,9 @@ export default class MyDocument extends Document {
     return (
       <Html lang={this.props.locale ?? "en"}>
         <Head>
-          {/* PWA primary color */}
           <meta name="theme-color" content={theme.palette.primary.main} />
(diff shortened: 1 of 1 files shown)


app/web/pages/_document.tsx
```diff
@@ -8,8 +8,9 @@ export default class MyDocument extends Document {
     return (
       <Html lang={this.props.locale ?? "en"}>
         <Head>
-          {/* PWA primary color */}
           <meta name="theme-color" content={theme.palette.primary.main} />
+          <link rel="manifest" href="/manifest.json" />
```
bakeiro, line 12: just this? LGTM

bakeiro (approved):

app/web/pages/_document.tsx
aapeliv: Yep! Seems we already had the manifest.json in the file, just wasn't linked.

PR #5049, opened 2024-10-26
Fix gitlab failing if protos doesn't need to run

nabramow:
Right now GitLab will fail if we didn't update anything that makes the protos job run, for example adding meeting notes or adjusting settings.json.

This is because the `preview:protos` job needs the `protos` job. To make the `preview:protos` job conditional on whether the `protos` job ran, we should use a when condition in the GitLab CI/CD configuration. This setting can prevent the `preview:protos` job from failing if the `protos` job did not produce new artifacts, such as when the only changes are in files like .eslint or settings.json.

### DIFF
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -516,6 +516,8 @@ preview:web-next:
 
 preview:protos:
   needs: ["protos"]
+    - job: "protos"
+      artifacts: true # Only run if artifacts exist
   stage: preview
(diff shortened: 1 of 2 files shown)


aapeliv:
Hey, sorry: I fixed this yesterday in https://github.com/Couchers-org/couchers/pull/5039 but didn't mention on slack! Sorry for the wasted effort!

nabramow:
Fixed in #5039

PR #5051, opened 2024-10-26
Fix backend issue with local image upload

nabramow:
I was getting a server error due to cors policy and unable to upload an image, which made it annoying to create events etc during testing.

**Backend checklist**
- [ N/A ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass


aapeliv (approved):
LGTM!

PR #5055, opened 2024-10-27
Fix message screen size issue on mobile

### LINKED ISSUE #4995: Messaging on mobile is buggy
**Problem**
Currently if I send a message on mobile, the message thread does not always advance to display the message I just sent. See below:

Screenshot_20241016_000443_Firefox

**Desired behavior:**

The message thread should automatically advance to display the message I just sent. I shouldn't need to scroll to see the message I just sent.

### LINKED ISSUE #5028: Mobile: Messages need more space
While using Couchers in my mobile Browser i realized it's almost impossible to read messages as there is lots of space occupied by other elements and very little space left for the chat view itself. This is on an Samsung S9, so not the largest phone.

I think we could leave out some elements or make them more compact to make the site more mobile-friendly.

nabramow:
On mobile there's not enough room to see messages, especially in host and surf request view. This PR makes some mobile adjustments for the messages.

Closes #4995 
Closes #5028 

**Web frontend checklist**
- [ ] There are no console warnings when running the app (no new ones)
- [ x ] All tests pass

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -48,7 +48,6 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   "@global html": {
     scrollPaddingTop: `calc(${theme.shape.navPaddingXs} + ${theme.spacing(2)})`,
     height: "100%",
-    overflow: "hidden",
   },
   [theme.breakpoints.up("sm")]: {
     "@global html": {
@@ -59,7 +58,6 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   },
   "@global body": {
     height: "100%",
-    overflow: "hidden",
   },
   "@global #__next": {
     display: "flex",
@@ -89,6 +87,8 @@ export default function AppRoute({
 
   const isNativeEmbed = useIsNativeEmbed();
   const isMobile = useMediaQuery(theme.breakpoints.down("sm")); // isMobile is browser mobile
+  const shouldShowFooter = !noFooter && !isNativeEmbed && !isMobile;
+  const shouldShowCookieBanner = !isPrivate && !isNativeEmbed && !isMobile;
 
   //there must be the same loading state on auth'd pages on server and client
   //for hydration matching, so we will display a loader until mounted.
@@ -134,10 +134,10 @@ export default function AppRoute({
             {/* Have to wrap this in a fragment because of https://github.com/mui-org/material-ui/issues/21711 */}
             <>{children}</>
           </Container>
-          {!noFooter && !isNativeEmbed && !isMobile && <Footer />}
+          {shouldShowFooter && <Footer />}
         </>
       )}
-      {!isPrivate && !isNativeEmbed && !isMobile && <CookieBanner />}
+      {shouldShowCookieBanner && <CookieBanner />}
     </ErrorBoundary>
   );
 }
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -1,4 +1,4 @@
-import { Container } from "@material-ui/core";
+import { Container, useMediaQuery } from "@material-ui/core";
 import classNames from "classnames";
 import CircularProgress from "components/CircularProgress";
 import CookieBanner from "components/CookieBanner";
@@ -9,6 +9,7 @@ import { useRouter } from "next/router";
 import { useIsNativeEmbed } from "platform/nativeLink";
 import { ReactNode, useEffect, useState } from "react";
 import { jailRoute, loginRoute } from "routes";
+import { theme } from "theme";
 import makeStyles from "utils/makeStyles";
 
 import Navigation from "./Navigation";
@@ -47,6 +48,7 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   "@global html": {
     scrollPaddingTop: `calc(${theme.shape.navPaddingXs} + ${theme.spacing(2)})`,
     height: "100%",
+    overflow: "hidden",
   },
   [theme.breakpoints.up("sm")]: {
     "@global html": {
@@ -57,6 +59,7 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   },
   "@global body": {
     height: "100%",
+    overflow: "hidden",
   },
   "@global #__next": {
     display: "flex",
@@ -85,6 +88,7 @@ export default function AppRoute({
   const isJailed = authState.jailed;
 
   const isNativeEmbed = useIsNativeEmbed();
+  const isMobile = useMediaQuery(theme.breakpoints.down("sm")); // isMobile is browser mobile
 
   //there must be the same loading state on auth'd pages on server and client
   //for hydration matching, so we will display a loader until mounted.
@@ -130,10 +134,10 @@ export default function AppRoute({
             {/* Have to wrap this in a fragment because of https://github.com/mui-org/material-ui/issues/21711 */}
             <>{children}</>
           </Container>
-          {!noFooter && !isNativeEmbed && <Footer />}
+          {!noFooter && !isNativeEmbed && !isMobile && <Footer />}
         </>
       )}
-      {!isPrivate && !isNativeEmbed && <CookieBanner />}
+      {!isPrivate && !isNativeEmbed && !isMobile && <CookieBanner />}
     </ErrorBoundary>
   );
 }
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -1,4 +1,4 @@
-import { Container, useMediaQuery } from "@material-ui/core";
+import { Container } from "@material-ui/core";
 import classNames from "classnames";
 import CircularProgress from "components/CircularProgress";
 import CookieBanner from "components/CookieBanner";
@@ -9,7 +9,6 @@ import { useRouter } from "next/router";
 import { useIsNativeEmbed } from "platform/nativeLink";
 import { ReactNode, useEffect, useState } from "react";
 import { jailRoute, loginRoute } from "routes";
-import { theme } from "theme";
 import makeStyles from "utils/makeStyles";
 
 import Navigation from "./Navigation";
@@ -86,9 +85,6 @@ export default function AppRoute({
   const isJailed = authState.jailed;
 
   const isNativeEmbed = useIsNativeEmbed();
-  const isMobile = useMediaQuery(theme.breakpoints.down("sm")); // isMobile is browser mobile
-  const shouldShowFooter = !noFooter && !isNativeEmbed && !isMobile;
-  const shouldShowCookieBanner = !isPrivate && !isNativeEmbed && !isMobile;
 
   //there must be the same loading state on auth'd pages on server and client
   //for hydration matching, so we will display a loader until mounted.
@@ -134,10 +130,10 @@ export default function AppRoute({
             {/* Have to wrap this in a fragment because of https://github.com/mui-org/material-ui/issues/21711 */}
             <>{children}</>
           </Container>
-          {shouldShowFooter && <Footer />}
+          {!noFooter && !isNativeEmbed && <Footer />}
         </>
       )}
-      {shouldShowCookieBanner && <CookieBanner />}
+      {!isPrivate && !isNativeEmbed && <CookieBanner />}
     </ErrorBoundary>
   );
 }
--- a/app/web/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/features/messages/groupchats/GroupChatView.tsx
@@ -1,3 +1,4 @@
+import { useMediaQuery } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 import { Skeleton } from "@material-ui/lab";
 import Alert from "components/Alert";
@@ -43,6 +44,7 @@ import {
 } from "react-query";
 import { groupChatsRoute } from "routes";
 import { service } from "service";
+import { theme } from "theme";
 
 import { GROUP_CHAT_REFETCH_INTERVAL } from "./constants";
 
@@ -51,14 +53,34 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
     marginTop: "auto",
     flexGrow: 0,
     paddingBottom: theme.spacing(2),
+    paddingLeft: theme.spacing(2),
+    paddingRight: theme.spacing(2),
+
+    [theme.breakpoints.down("sm")]: {
+      paddingLeft: theme.spacing(1),
+      paddingRight: theme.spacing(1),
+    },
   },
   header: {
-    paddingTop: theme.spacing(1),
+    padding: theme.spacing(1, 2),
+    borderBottom: `1px solid ${theme.palette.divider}`,
     alignItems: "center",
     display: "flex",
     flexGrow: 0,
     "& > * + *": {
       marginInlineStart: theme.spacing(2),
+    },
+
+    [theme.breakpoints.down("sm")]: {
+      paddingLeft: theme.spacing(1),
+      paddingRight: theme.spacing(1),
+    },
+  },
+  messageLoader: {
+    padding: theme.spacing(2, 2),
+
+    [theme.breakpoints.down("sm")]: {
+      padding: theme.spacing(1, 1),
     },
   },
   pageWrapper: {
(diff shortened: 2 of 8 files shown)


app/web/components/AppRoute.tsx
```diff
@@ -47,6 +48,7 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   "@global html": {
     scrollPaddingTop: `calc(${theme.shape.navPaddingXs} + ${theme.spacing(2)})`,
     height: "100%",
+    overflow: "hidden",
```
nabramow: FYI both of these are fixing an issue I noticed where the whole app scrolls. We don't want that for mobile nor regular app (i.e. the nav bar scrolling with the whole page). Top layer should not scroll, that should be responsibility of the child components.

Example of what I mean on mobile (also relevant to non-Mobile):

ScreenRecording2024-10-27at11 38 40AM-ezgif com-video-to-gif-converter

jesseallhands (commented):
LGTM!

Tested in firefox using various mobile resolutions and appears to fix the issues!

app/web/features/messages/requests/HostRequestView.tsx
```diff
@@ -154,6 +156,40 @@ export default function HostRequestView({
 
   const handleBack = () => router.back();
 
+  const renderUserSummary = () => (
```
darrenvong: A minor thing - since you don't need to pass anything to the function, did you know you can assign the JSX directly to the variable, so then you can reference it directly (as opposed to needing to call it when assigned to a function)?

app/web/components/AppRoute.tsx
```diff
@@ -130,10 +134,10 @@ export default function AppRoute({
             {/* Have to wrap this in a fragment because of https://github.com/mui-org/material-ui/issues/21711 */}
             <>{children}</>
           </Container>
-          {!noFooter && !isNativeEmbed && <Footer />}
+          {!noFooter && !isNativeEmbed && !isMobile && <Footer />}
         </>
       )}
-      {!isPrivate && !isNativeEmbed && <CookieBanner />}
+      {!isPrivate && !isNativeEmbed && !isMobile && <CookieBanner />}
```
darrenvong: Minor thing again - but I wonder if it's worth extracting these to readable variable names? Maybe it's at the point where it's fine to leave for now unless we add one more `&&`

app/web/components/AppRoute.tsx
```diff
@@ -57,6 +59,7 @@ export const useAppRouteStyles = makeStyles((theme) => ({
   },
   "@global body": {
     height: "100%",
+    overflow: "hidden",
```
darrenvong: The whole app seems to be unscrollable apart from the message pages, and even the chat box is buggy where it's completely hidden for me 🤔

https://github.com/user-attachments/assets/1090d51a-a721-44a8-bdd1-893d122e3f7a

(Excuse the weird light in the top lol, I need a better screen reader app that doesn't show my room ceiling 😆)

Edit: it's a bit odd that the issue doesn't seem to be reproducible if you use the mobile viewport from the dev tool, but it's a problem against a real mobile device's browser (on an Android OnePlus 9 but I doubt that's the problem? Would be interested to see how it fares on iPhones)

darrenvong (changes requested):

app/web/components/AppRoute.tsx
darrenvong: It seems to break scrolling even on desktop (no scrollbar for home page)
Screenshot 2024-10-28 at 01 04 35

app/web/features/messages/requests/HostRequestView.tsx
nabramow: Ooo I actually did not know that, but it makes sense now that you said it! You're right I guess there's nothing in there that's a state change that wouldn't just be a page reload.

app/web/components/AppRoute.tsx
nabramow: Yeah that makes sense!

app/web/components/AppRoute.tsx
nabramow: Hmmm I guess that was a bit ambitious of me to sneak into this PR...I'll take it out because you're right it probably needs a separate thorough testing.

app/web/features/messages/requests/HostRequestView.tsx
darrenvong: It makes more sense if you think of it as (since JSX is syntactic sugar of the below!):
```typescript
const userSummary = React.createElement(...)
```

otherwise intuitively it's like "why am I allowed to assign HTML to JS variables?" 😄 

(A bit nitpicky but I would get rid of `render` in the variable name since it's no longer a function)

app/web/components/AppRoute.tsx
nabramow: Just made a separate ticket #5070 . That's probably a good first issue for one of the new people as well!

app/web/features/messages/requests/HostRequestView.tsx
nabramow: Nah totally fair I was thinking it as I did it but was like gah there's already a "UserSummary" so I named it "userSummarySection". Done!

darrenvong:
One more tiny spacing thing I spotted after this fix - I wonder if there's an easy fix to stop the top bar being stuck right underneath the nav bar (unless you scroll all the way to the beginning of the chat which seems to scroll it back to the right position)?
IMG_20241028_022604

app/web/features/messages/requests/HostRequestView.tsx
```diff
@@ -158,17 +157,27 @@ export default function HostRequestView({
...
-              ).format("LL")}`}
+              {`${
+                isMobile
+                  ? dayjs(hostRequest.fromDate).format("ll")
+                  : dayjs(hostRequest.fromDate).format("LL")
+              } - ${
+                isMobile
+                  ? dayjs(hostRequest.fromDate).format("ll")
+                  : dayjs(hostRequest.toDate).format("LL")
+              }`}
```
nabramow: Makes the shortened form of the month name on mobile as it was wrapping to a second line otherwise.

app/web/features/messages/requests/HostRequestView.tsx
```diff
@@ -186,7 +195,6 @@ export default function HostRequestView({
           </div>
         )}
       </UserSummary>
-      {isMobile && <Divider spacing={1} />}
```
nabramow: Decided to make these a border instead otherwise it had a little floating bit of padding after the divider.

app/web/pages/messages/[[...slug]].tsx
```diff
@@ -58,4 +58,4 @@ export default function LeaveReferencePage() {
   );
 }
 
-LeaveReferencePage.getLayout = appGetLayout();
+LeaveReferencePage.getLayout = appGetLayout({ variant: "full-width" });
```
Comment (nabramow) [inline app/web/pages/messages/[[...slug]].tsx]: This was the magic one here to stop the weird scrolling, but then I had to add it to all the inner components. As it should be, I think.

nabramow:
> One more tiny spacing thing I spotted after this fix - I wonder if there's an easy fix to stop the top bar being stuck right underneath the nav bar (unless you scroll all the way to the beginning of the chat which seems to scroll it back to the right position)? 

Done! This ended up with me setting the main component to "full-width" variant so you'll see I had to go in and add a lot of paddings to the child components.

That also solved the weird navbar scrolling (for the chat page at least) without adjusting the global css, win-win!

While I was there also fixed some other stuff like non-centered loading spinner and made the date format for the host/surf request shorter for mobile so it stays to one line.

bakeiro:
not sure if it's only on the mobile version of the desktop browser (sometimes it's different from an actual phone) but I'm seeing this (brave, chromium browser)
image

app/web/pages/messages/[[...slug]].tsx
Comment (bakeiro) [reply app/web/pages/messages/[[...slug]].tsx]: oh I used no-footer = true on the map 🤔 
`SearchPage.getLayout = appGetLayout({ noFooter: true });`

bakeiro:
suggestion [this should be a UI/UX decision not mine but]
what about this?
image

jesseallhands:
>suggestion [this should be a UI/UX decision not mine but]
what about this?

My suggestion is to leave the dates (very relevant for the request) but remove the user location (interesting, but not directly relevant for the request). This will free up some screen real estate.

jesseallhands:
If we want to retain at least some of the requesting user's location info, then we could add the country name after the requesting user's age. For example, the following would work well: 

`Claudio, 39, France`

Perhaps we can apply this just to the mobile version since it doesn't cause problems for desktop?

But on mobile, long names could still lead to text wrap on small screens, so perhaps we just don't display the requesting user's location on mobile (only) since it's not directly relevant to the request.

app/web/pages/messages/[[...slug]].tsx
Comment (nabramow) [reply app/web/pages/messages/[[...slug]].tsx]: @bakeiro Do we want the footer on the desktop version though for this page? Or only no footer on the mobile version?

bakeiro:
> > suggestion [this should be a UI/UX decision not mine but]
> > what about this?
> 
> My suggestion is to leave the dates (very relevant for the request) but remove the user location (interesting, but not directly relevant for the request). This will free up some screen real estate.

yeah... just to clarify my suggestion was to move the cancel button there, not something like removing the user's location 😅 

in the ticket for avoiding the location taking to much space my suggestion was to trim the text if it's to long (to make it occupy one line max) i.e: `comunidad Valenciana, Valencia, Sp...`

app/web/pages/messages/[[...slug]].tsx
Comment (bakeiro) [reply app/web/pages/messages/[[...slug]].tsx]: Nevermind! (U can ignore my comment xd)

nabramow:
> not sure if it's only on the mobile version of the desktop browser (sometimes it's different from an actual phone) but I'm seeing this (brave, chromium browser) 

Hmm I am not able to replicate this at all. I even downloaded the Brave browser for you David! Chromium browser I tried but every time I attempted to install, it said the download was damaged.

I did try every phone simulator on Google Chrome and in Safari and no issue with that broken nav bar. 

I'm gonna say that's good enough, as these are sort of niche browsers.

app/web/pages/messages/[[...slug]].tsx
Comment (nabramow) [reply app/web/pages/messages/[[...slug]].tsx]: I actually just did it as probably we don't want the footer on the messages page when I thought about it more...

nabramow:
> If we want to retain at least some of the requesting user's location info, then we could add the country name after the requesting user's age. For example, the following would work well:
> 
> `Claudio, 39, France`
> 
> Perhaps we can apply this just to the mobile version since it doesn't cause problems for desktop?
> 
> But on mobile, long names could still lead to text wrap on small screens, so perhaps we just don't display the requesting user's location on mobile (only) since it's not directly relevant to the request.

@bakeiro @jesseallhands The host request info no longer sticks to the top of the screen, it's scrollable. So I think it's okay as it is? They can just scroll down now!

app/web/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -98,11 +121,27 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
...
+    width: "100%",
+  },
+  userSummary: {
+    borderBottom: `1px solid ${theme.palette.divider}`,
+    padding: theme.spacing(1) + " !important",
+
+    [theme.breakpoints.down("sm")]: {
+      borderBottom: `1px solid ${theme.palette.divider}`,
+      paddingBottom: theme.spacing(1) + " !important",
+    },
```
darrenvong: Curious why we need `important`s if we're opening up the `UserSummary` component for style extension. Are there no other nice ways to achieve this without it? 🤔

darrenvong:
I played around on Android Brave as well and I don't see the weird glitches David was mentioning... so I think best to leave it until we get more reports as it's clearly not that easy to reproduce? In most cases, it should practically be the same as Chrome (since it runs on Chromium under the hood, but with more aggressive tracking protections/built-in ad blocks etc)

app/web/features/messages/groupchats/GroupChatView.tsx
nabramow: So in `UserSummary.tsx`, even though I passed my `className` through, the padding in that component's `classes.root` is overriding my `paddingBottom`. If I take out the `padding: 0` in `classes.root` in `UserSummary` then my css works.

Any idea best way to handle that then? I tried a bunch but got stumped so added the `! important` 😅. I'm trying not to change style in `UserSummary` since it's reused all over.

bakeiro:
> > not sure if it's only on the mobile version of the desktop browser (sometimes it's different from an actual phone) but I'm seeing this (brave, chromium browser)
> 
> Hmm I am not able to replicate this at all. I even downloaded the Brave browser for you David! Chromium browser I tried but every time I attempted to install, it said the download was damaged.
> 
> I did try every phone simulator on Google Chrome and in Safari and no issue with that broken nav bar.
> 
> I'm gonna say that's good enough, as these are sort of niche browsers.

oh, but it's chromium based! they use webkit, I don't think you need to download it...
I'm thinking that might be because of the name length (it's a long name)so maybe hasn't nothing to do with the browser... anyway feels like a small detail to me, not a big thing, still if you wanna resolve it... (I was just trimming it can be enough)

bakeiro:
> > If we want to retain at least some of the requesting user's location info, then we could add the country name after the requesting user's age. For example, the following would work well:
> > `Claudio, 39, France`
> > Perhaps we can apply this just to the mobile version since it doesn't cause problems for desktop?
> > But on mobile, long names could still lead to text wrap on small screens, so perhaps we just don't display the requesting user's location on mobile (only) since it's not directly relevant to the request.
> 
> @bakeiro @jesseallhands The host request info no longer sticks to the top of the screen, it's scrollable. So I think it's okay as it is? They can just scroll down now!

I know I know, that why I suggested to put the botton there, since that content is scrollable and it's next to the hosting info [and that button it's to cancel that request] and right now that button it's next to the message field which not sure if makes much sense...

but this is just dumping my thoughts, I'm not the UI person :)

app/web/pages/messages/[[...slug]].tsx
Comment (bakeiro) [reply app/web/pages/messages/[[...slug]].tsx]: yeah that's the thing, I though about do we need the footer after loggin in? also I though, maybe we use the same solutions for the same problems, instead of different solutions for the same problem, but I don't know which is the difference between `full-width` and `noFooter`...

in case we need the footer in desktop I guess full-width? but since I dind't want to change more than needed I asked you to ignore my comment lol

Comment (bakeiro) [reply app/web/pages/messages/[[...slug]].tsx]: 🤔🤔🤔 I'm still thinking, I don't think we need the footer at all in the messages page

Screenshot_2024-10-29-10-58-27-78_e4424258c8b8649f6e67d283a50a2cbc

Comment (bakeiro) [reply app/web/pages/messages/[[...slug]].tsx]: do we need it at all after the user logs in? [map, chat, dashboard, events...] we can discuss this in the today's meeting

bakeiro (dismissed):
LGTM

app/web/features/messages/groupchats/GroupChatView.tsx
nabramow: Hey @darrenvong just pinging you for this, it's the last thing blocking the PR ;-). I could also take it out, it doesn't really look so bad without it.

darrenvong: Hmm yeah sounds like order of CSS injection is reversed somehow - i.e. `UserSummary` styling is injected after the chat view ones here?

The only thing I can think of that's slightly better is specificity hack it by applying the styles against the same class twice with `'&&'`... At least it makes it slightly clearer it's a local override rather than a stack of importants?

Or since you're saying it's not too bad without the 0 padding, maybe it's not the worse thing to take it out!

nabramow:
@darrenvong @jesseallhands Okay I realized a fatal flaw with making the request info at the top non-sticky...if the users have a lot of messages they would have to scroll allllll the way up to see it.

So I made it sticky on mobile again, but made a much smaller version as suggested in comments above. It looks like this:

Screenshot 2024-10-30 at 11 49 49 PM

Let me know if we're happy with that!

And Darren, I solved the css classname inheritance issue by wrapping `UserSummary` in a div and targeting that with the className instead.

Also I hope it's okay I made a `localStyles`. It was starting to get too confusing with the styles in another component and since these ones were only for the mobile view I thought better make separate styles only for this one.

darrenvong (approved):
Looks good to me, sorry for the slow re-review!

PR #5073, opened 2024-10-28
fix empty discussion post in events communities

### LINKED ISSUE #4172: Empty discussion posts in events and communities are allowed
It is currently possible to submit an empty comment in community and event discussion threads:

image

You should not be able to post a comment if it is empty or only contains `&nbsp;` or other nonvisible characters. Reddit has a simple solution to this with an error text:

image

I don't really like reddit's wording; I think something like "You need to type something first!" is a bit more clear.

katemoser:
Closes #4172 

Prevent empty comments and empty replies in events and communities.

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [x] All tests pass


katemoser:
Sorry, half my comment seems to have been deleted, I think I filled out the template wrong!

This is for #4172 , added front end validation and backend check as well. Started writing tests but would like to chat with someone for 15-20 minutes about how to best write tests for this!

nabramow:
> Sorry, half my comment seems to have been deleted, I think I filled out the template wrong!
> 
> This is for #4172 , added front end validation and backend check as well. Started writing tests but would like to chat with someone for 15-20 minutes about how to best write tests for this!

Leaving a comment here about tests in case we don't overlap soon!

Code looks great so far.

Another good frontend test would be to make a new test file `CommentForm.test.tsx` (since no one is testing that component yet) and test that when the comment form is empty or only filled with whitespace characters, tabs, etc. and user hits submit, it doesn't call the `postReply` function. You could do this by mocking the service like:

```javascript
const postReplyMock = service.threads.postReply as jest.Mock;
```

Then check it wasn't called in the individual test:

```javascript
expect(postReplyMock).not.toHaveBeenCalled();
```

We can assume that if the function that picks up the change wasn't called, then we successfully protected it via the frontend.

For backend, you probably wanna add a test case in `test_threads.py`, the `def test_threads_errors(db)` function looks right. Probably you'd make the content in the arguments equal to `""` and then check the right error showed. Looks like there's some similar functions in there you could base it on.

Instructions for testing locally on the backend are in `app/backend/readme.md`. You can run tests for only one function like `pytest test_threads.py::test_threads_errors` (path might be different to folder).

app/backend/src/couchers/errors.py
```diff
@@ -74,6 +74,7 @@
 INCOMPLETE_PROFILE_SEND_REQUEST = "You have to complete your profile before you can send a request."
 INSECURE_PASSWORD = "The password is insecure. Please use one that is not easily guessable."
 INVALID_BIRTHDATE = "You must be at least 18 years old to sign up."
+INVALID_COMMENT = "You cannot post an empty comment."
```
bakeiro, line 77: don't we have translations for the backend errors? (not related with this PR, generally asking)

aapeliv (commented):
Backend looks good to me!

bakeiro:
I've run the `lint` & `format` to make the build pass, @aapeliv says the backend look good... guess it's fine to merge? (and to leave the PR in this state...)

bakeiro (dismissed):
LGTM

bakeiro (dismissed):

nabramow (commented):
Hey I am unapproving it as gitlab tests are failing.

nabramow (changes requested):
Sorry now I am unapproving.

app/web/features/communities/EditCommunityInfoPage.tsx
```diff
@@ -148,11 +148,6 @@ export default function EditCommunityPage({
                 imageUpload
                 required={t("communities:page_content_required")}
               />
-              {errors.content && (
-                <Typography color="error" variant="body2">
-                  {errors.content.message}
-                </Typography>
-              )}
```
nabramow, line 155: I moved the error functionality to be within the MarkdownInput since we are passing the errorText in multiple components already but not displaying it.

Now that that's done I can remove this extra error logic.

app/web/components/MarkdownInput/MarkdownInputNoSSR.tsx
```diff
@@ -50,6 +50,10 @@ const useStyles = makeStyles((theme) => ({
       border: "2px solid red",
     },
   },
+  errorText: {
+    color: theme.palette.error.main,
+    marginTop: "8px",
```
aapeliv: Should this be in `theme.spacing`?

app/backend/src/couchers/servicers/threads.py
```diff
@@ -116,6 +116,9 @@ def GetThread(self, request, context, session):
         return threads_pb2.GetThreadRes(replies=replies, next_page_token=next_page_token)
 
     def PostReply(self, request, context, session):
+        if request.content.strip() == "":
```
aapeliv: Might be good to do `content = request.content.strip()`, then use `content` below on lines 124/126. That way it'll strip the surrounding whitespace from the comments/replies.

aapeliv (dismissed):
LGTM, two small optional things

Issue #5087, opened 2024-11-02
Cant upload profile picture


CouchersBot:
Subject: Cant upload profile picture
Description:
Upon uploading profile picture, website displays "Couldnt connect to server". Attempted downsizing to 60kb and different pictures, same result each time

Results:

Backend version: develop-182774fb
Frontend version: develop-182774fb
User Agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Mobile Safari/537.36
Page: https://couchers.org/user/felixthecat
User: georgericks (47701)

hueldoeu:
i didn't know couchers.org is US american, i though it's european like bewelcome.org.
https://couchers.org/foundation
i had some bad experiences with airbnb and in my opinion, US american companies should stay away of Europe, especially if they are causing problems.

aapeliv:
> i didn't know couchers.org is US american, i though it's european like bewelcome.org. https://couchers.org/foundation i had some bad experiences with airbnb and in my opinion, US american companies should stay away of Europe, especially if they are causing problems.

I am not sure what this has to do with anything. Indeed Couchers.org is legally set up in the US: it’s where we could easily and safely incorporate a full charity. That said, the majority of the board are not Americans and the majority live outside the US. We are also a non-profit, unlike Airbnb and other US for profit corporations.

aapeliv:
Can replicate and confirmed to be an issue.

aapeliv:
Fixed in 7618e865b0c98e098bb076e895ff962c8b1d6421

PR #5088, opened 2024-11-02
Move media CORS origin config into media service

aapeliv:
We merged #5051 so that we could have CORS even in the case when there's no nginx in front of the media container. (Different ports on `localhost` count as different origins, so this is needed.) That unfortunately introduced a bug because we now had two of the CORS origin headers. 7618e865b0c98e098bb076e895ff962c8b1d6421 is a hotfix to remove it from nginx. This PR completes the rest of the change and moves it to media to make it fully configurable.

See also #5087.

**Backend checklist**
- [ ] All tests pass


app/media/src/tests/test_server.py
```diff
@@ -68,6 +68,7 @@ def client_with_secrets(tmp_path):
         main_server_address="localhost:8088",
         main_server_use_ssl=False,
         media_upload_location=tmp_path,
+        media_cors_origin="*",
```
nabramow, line 71: This is allowing upload from anywhere right? I wonder if it should be more specific to our media server?

We could do something like 

```
const allowedOrigins = ['https://couchers.org', 'http://localhost:3000']; // etc
```

Or am I misunderstanding?

aapeliv: Yes, that's right. When running this properly, it'll pull the origin from an environment variable. We'll use something quite restrictive in prod, but in staging it can be `*`.

nabramow (approved):
LGTM!

PR #5119, opened 2024-11-07
Finish maplibre upgrade

nabramow:
Closes #5106 

Upgrading maplibre to the most recent stable version 4.7.1.

v3 had no breaking changes for us.

v4 had these breaking changes that were addressed:
Guide: https://github.com/maplibre/maplibre-gl-js/blob/main/CHANGELOG.md#400

* ⚠️ Removed callback usage from map.loadImage in continue to below change (https://github.com/maplibre/maplibre-gl-js/pull/3422)
* ⚠️ Changed the GeoJSONSource's getClusterExpansionZoom, getClusterChildren, getClusterLeaves methods to return a Promise instead of a callback usage (https://github.com/maplibre/maplibre-gl-js/pull/3421)
* ⚠️ Changed the map.loadImage method to return a Promise instead of a callback usage (https://github.com/maplibre/maplibre-gl-js/pull/3233)

Besides that, I:

* Changed MapMouseEvent & { features: etc } to `MapLayerMouseEvent` as it's the same thing and shorter
* Removed `MaplibreEvent` as it no longer exists in the package and I removed it from related types. It didn't seem to have any effect so assuming this got deprecated and they forgot to document it.

**Web frontend checklist**
- [ x ] Formatted my code with `yarn format`
- [ x ] There are no warnings from `yarn lint --fix`
- [ x ] There are no console warnings when running the app
- [ x  ] All tests pass
- [ x ] Clicked around my changes running locally and it works
- [ x ] Checked Desktop, Mobile and Tablet screen sizes

PR #5143, opened 2024-11-10
Bug - Fix bbox value

nabramow:
I was working on another branch, clicking around to make sure I didn't break anything. I noticed when I searched a location on the map, it jumped to a random unrelated place.

I replicated this on prod, so seems it's a bug. Since I recently upgraded maplibre in #5119 , I thought maybe I broke something.

But in the end seems like we were putting the bbox values in the wrong order on search?! So either this has been a longstanding issue we didn't notice or ???.

I guess the NominatumPlace could have also changed via that api call, but that also does not make sense tbh.

But hey...it works?!

**Definitely test this one though in case I'm going crazy!**. I tested a lot myself but I still don't understand why we had this bug to begin with.

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/Dockerfile.prod
+++ b/app/web/Dockerfile.prod
@@ -1,4 +1,4 @@
-FROM node:20-bullseye as builder
+FROM node:20-bullseye AS builder
 
 ARG environment=development
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


bakeiro:
I think this is new 🙃 I tested with locations that were working nicely before and now seems they don't... I gonna check this as well!

bakeiro:
I pushed a small commit, the dashboard needed a small fix as well

bakeiro (dismissed):

bakeiro (approved):

PR #5200, opened 2024-11-19
Fix mobile typing bug in messages

### LINKED ISSUE #5127: [messges] expired request improvements
Context:
I receive a request, I open the chat, I see it's expired and see a few flaws

ToDo:
- If it's expired, don't show [accept] [decline] buttons ;)
- Change text to: `This request expired. To continue chatting, send the other person a normal message.
- Fix the `undefined` string...

384617809-0170d3bd-2a71-43bc-a321-bda16571eff4

### LINKED ISSUE #5147: Request dates shown incorrectly on small devices
Subject: Request dates shown incorrectly on small devices
Description:
If you enter a request chat and change screen size to a small size, the dates will show up as "From - From" instead of "From - To".

Results:

Backend version: develop-1843cb28
Frontend version: develop-1843cb28
User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:133.0) Gecko/20100101 Firefox/133.0
Page: https://couchers.org/dashboard
User: aapeli (1)

### LINKED ISSUE #5185: When composing message in new chat, user can't see what they are typing
### Problem
When a user creates a new chat and begins to compose a message, by default it is not possible for them to see what they are typing while they are composing the message. See video below that demonstrates the issue:

ezgif-6-a267b48abb

### Solution
Message box where the message is being composed should be fixed to above the keyboard so the user can see what they are typing. Here is what that behavior looks like on trustroots:
ezgif-6-114b75b3d8

nabramow:
The typing area was getting cut off on mobile. This was because we weren't accounting for the mobile browser url bar. This isn't shown in Chrome responsive view, so we missed it.

Also fixed a mobile date issue here from #5127 .

Closes #5185

**TESTING:** It's best if you can test this on a real phone, i.e. use the Vercel link on your phone. This bug was only possible to replicate on mobile as the dev tools responsive view wasn't 100% accurate in this case.

**Web frontend checklist**
- [ x ] All tests pass


app/web/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -84,29 +67,13 @@ export const useGroupChatViewStyles = makeStyles((theme) => ({
...
-    width: "100%",
-    display: "flex",
-    alignItems: "center",
-    marginInlineEnd: theme.spacing(2),
-    marginInlineStart: theme.spacing(2),
-    "& > *": { marginInlineEnd: theme.spacing(2) },
 
     [theme.breakpoints.down("sm")]: {
-      fontSize: "0.9rem",
+      height: `calc(100vh - ${theme.shape.navPaddingXs} - 80px)`, // 80 is space for mobile browser url bar
```
nabramow: This is what actually fixed it. If you have a non iPhone I especially want you to test this. I'm not totally sure if the url bar sizing will be weird on some phones.

app/web/features/messages/requests/HostRequestView.tsx
```diff
@@ -308,28 +336,14 @@ export default function HostRequestView({
...
-                  />
-                </div>
-              )}
+              <div className={classes.footer}>
+                <HostRequestSendField
+                  hostRequest={hostRequest}
+                  sendMutation={sendMutation}
+                  respondMutation={respondMutation}
+                />
+              </div>
```
nabramow: I made it sticky again but made the header bar a bit smaller on mobile.

app/web/theme.ts
```diff
@@ -110,28 +110,6 @@ const themeOptions: ThemeOptions = {
...
-        borderRadius: "20px",
-      },
-      "&::-webkit-scrollbar-thumb:hover": {
-        background: "rgba(0,0,0,0.3)",
-      },
-      overflow: "auto",
-      paddingInlineEnd: spacing(1),
-      scrollbarHeight: "thin",
-      scrollbarWidth: "thin",
-    },
```
nabramow: This is only used in the `InfiniteMessageLoader` so doesn't make sense to have it in the theme overrides globally.

Moving it as it makes it easier to tell in that component what style is set.

app/web/features/messages/groupchats/GroupChatView.tsx
nabramow: Also Github is showing this weird, I didn't replace the fontSize with this, hah.

jesseallhands:
Seems to be working now! 

Though on chrome I am seeing a gap between the text field and the nav bar on my phone:


https://github.com/user-attachments/assets/cc4045be-d1a7-41e5-870c-ef6f69947b14

jesseallhands:
I noticed this now: On the one hand it's better cause the text I'm composing isn't completely hidden, but on the other hand it still partially is hidden. 

The other thing is that when I scroll down there is space below the text field until I scroll back up then I lose my ability to scroll down and see the gap, until I start typing again:

https://github.com/user-attachments/assets/52a01a24-7af8-4371-bd48-1cc9509698b9

jesseallhands:
Here is my latest test, I'd say it's good enough for now:

This is on android/firefox

https://github.com/user-attachments/assets/96426836-2b2c-477a-ad4f-4aa7ef1d79ae

jesseallhands:
I confirmed that on Chrome it is completely fixed, working as expected.

nabramow:
> Here is my latest test, I'd say it's good enough for now:
> 
> This is on android/firefox
> 

Okay so I just pushed a fix that might fix firefox on Android, so give that a try. BUT if not, since I don't have any device to replicate it, and even the firefox Android mobile simulator plugin can't replicate it, I'm going to consider it good enough for now! They can still type and it's better than now. I think probably the percentage of Android users using Firefox on the phone is low. Based on the number of Stack Exchange posts in this, it seems to be an Android bug.

jesseallhands:
@nabramow it's fixed now on firefox too! Thank you!!!

nabramow:
> @nabramow it's fixed now on firefox too! Thank you!!!

Oh wow what luck, that was a total stab in the dark! Excellent we are all good then! Thanks for testing so many times!

nabramow:
@bakeiro if you have a moment this bug fix is ready too!

app/web/features/messages/groupchats/ChatContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+  if (!messages || (isHostRequest && !hostRequest)) {
+    return null;
+  }
+
+  return (
+    <StyledInfiniteMessageLoader
+      earliestMessageId={
+        messages.pages[messages.pages.length - 1].lastMessageId
+      }
+      latestMessage={messages.pages[0].messagesList[0]}
```
aapeliv, line 64: Are you always guaranteed that if `messages` is not undefined then `pages` has at least one page?

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
```diff
@@ -0,0 +1,285 @@
...
+        aria-label={t("messages:chat_view.back_button.a11y_label")}
+        size={isMobile ? "small" : "medium"}
+      >
+        <BackIcon fontSize={isMobile ? "small" : "medium"} />
+      </HeaderButton>
+
+      {groupChat?.isDm ? (
+        <StyledTitleBox>
+          <Link
+            href={`/user/${getDmUsername(
```
aapeliv: It'd be better to use `routeToUser` from `routes.ts` here.

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
```diff
@@ -0,0 +1,285 @@
...
+              groupChatMembersQuery,
+              currentUserId
+            )}`}
+          >
+            <a>
+              <PageTitle
+                sx={{
+                  [theme.breakpoints.down("sm")]: {
```
aapeliv, line 126: Use `isMobile` here?

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
```diff
@@ -0,0 +1,285 @@
...
+                fontSize={isMobile ? "small" : "large"}
+              />
+            )
+          )}
+        </StyledTitleBox>
+      ) : (
+        <StyledTitleBox>
+          <PageTitle
+            sx={{
+              [theme.breakpoints.down("sm")]: {
```
aapeliv, line 151: ditto

app/web/features/messages/requests/HostRequestUserSummarySection.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <StyledShortUserInfo>
+        <Typography component="p" variant="body2">
+          {!otherUser ? (
+            <Skeleton />
+          ) : (
+            `${
+              (otherUser?.name.length ?? 0) < 25
+                ? otherUser?.name
+                : otherUser?.name.substring(0, 25) + "..."
+            }, ${otherUser?.age}, ${otherUser?.city.split(",")[2]}` // get only country
```
aapeliv: The user `city` isn't guaranteed to be of the format `City, State, Country`. I would say keep it as is, maybe truncate if it's too long.

app/web/features/messages/requests/HostRequestUserSummarySection.tsx
```diff
@@ -0,0 +1,127 @@
...
+      )}
+      <StyledShortUserInfo>
+        <Typography component="p" variant="body2">
+          {!otherUser ? (
+            <Skeleton />
+          ) : (
+            `${
+              (otherUser?.name.length ?? 0) < 25
+                ? otherUser?.name
+                : otherUser?.name.substring(0, 25) + "..."
```
aapeliv: That said, maybe move this to a utility function so we can just do `truncateTextEllipsis(text, max_length)` or similar.

app/web/features/messages/requests/HostRequestUserSummarySection.tsx
```diff
@@ -0,0 +1,127 @@
...
+          )}
+        </Typography>
+        {hostRequest && (
+          <Typography
+            component="p"
+            variant="h3"
+            sx={{ paddingRight: theme.spacing(1) }}
+          >
+            {`${dayjs(hostRequest.fromDate).format("ll")} - ${dayjs(
+              hostRequest.fromDate
```
aapeliv: This should be `toDate`, which then closes #5147 as well :)

aapeliv (changes requested):
This is looking great! Thanks for the big cleanup! I added a bunch of mostly minor comments.

app/web/features/messages/groupchats/ChatContent.tsx
nabramow: Yes, just tested it writing a new person for the first time and it automatically creates a "Nicole created the chat" author message there. Also I think if it's not undefined it will just be an empty array.

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
nabramow: It's user for both, just I'm making the fontSize smaller than default for mobile. The CSS key needs the `theme.breakpoints.down("sm")` as it break down to `@media (max-width:599.95px)` for example when the function is run, versus a boolean in `useMediaQuery` hook.

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
nabramow: Same as above

app/web/features/messages/requests/HostRequestUserSummarySection.tsx
nabramow: Done!

app/web/features/messages/requests/HostRequestUserSummarySection.tsx
nabramow: Ooo good catch, done!

app/web/features/messages/groupchats/GroupChatHeaderBar.tsx
aapeliv: OK, makes sense

aapeliv (approved):
LGTM!

PR #5215, opened 2024-11-22
Backend/bugfix/4098 expired pending request showing new

Krishnag09:
**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


aapeliv:
Which issue is this addressing?

I think it's better not to keep expiration state in the database explicitly. Then we'd need to keep updating it and have some background job to do it. It should be a computed field: if it's now past the end of the request, then it's expired.

nabramow:
@Krishnag09 I think we can close this since we're doing it on the front end - is that right?

nabramow:
Closing as we handled it on the frontend.

PR #5157, opened 2024-11-11
Material UI upgrade to v5

nabramow:
Migrating Material-UI to v5! There are a TON of changes here, the best you can do is really click around the app and make sure everything works and looks good. I did a thorough testing myself, but this migration touched almost every file so it's worth having as many people test as possible.

This is my plan for the full migration:
1. Upgrade to v5 using adaptorV4 theme functions **[[ YOU ARE HERE ]]**
2. Migrate to the new Material-UI style system, which styles by component and removes 'makeStyles`. I started some components in here. Remove the adapter theme stuff as it prevents migration to v6 later.
3. Upgrade Next.js to v13 (at least) and React to v18 
4. Migrate Material-UI to the latest version

Some notable things:

* I upgraded `msw` and `eslint-plugin-storybook` packages as well as they were causing conflicts with Typescript types in MUI v5.
* The next router in `UserPage.test.tsx` and `ProfilePage.test.tsx` I could not get working as written. The real app seems to work though, so I commented out part of the tests after spending many hours on it. I *think* the Material-UI Next package and Next.js support might have something to do with it, but we need to upgrade to Next.js v13 before we can use that so 🤷‍♀️.

The full migration guide is here FYI:
https://mui.com/material-ui/migration/migration-v4/

**Web frontend checklist**
- [ x ] Formatted my code with `yarn format`
- [ x ] There are no warnings from `yarn lint --fix`
- [ x ] There are no console warnings when running the app
- [ x ] All tests pass
- [ x ] Clicked around my changes running locally and it works
- [ x ] Checked Desktop, Mobile and Tablet screen sizes

PR #5225, opened 2024-11-25
Bug fixes mobile search

### LINKED ISSUE #5091: Search doesn't present a list of hosts / search function on mobile is broken
Subject: Search doesn't present a list of hosts 
Description:
Search for any city 

Results:
You only see a map with a bunch of icons for hosts, but you do not see any hosts. You have to click on each icon after zooming in in order to see hosts. 

Backend version: develop-7618e865
Frontend version: develop-7618e865
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.1 Mobile/15E148 Safari/604.1
Page: https://couchers.org/search?location=New+York%2C+New+York%2C+United+States&hostingStatusOptions=2&hostingStatusOptions=3&bbox=-73.7002330&bbox=40.9176300&bbox=-74.2588430&bbox=40.4765780#loc=8.88/40.6975/-73.9795
User: brooklyngarson (12838)

### LINKED ISSUE #5219: Breakpoint issues in Mui upgrade
In this commit: 91fbf5b70bfcd120106d876793f4a4e1361c3bcd

* `SearchResult.tsx` should display more info on big screen, less on small. This behavior was inverted.
* It seems also at least `[ReferenceOverview.tsx](https://github.com/Couchers-org/couchers/commit/91fbf5b70bfcd120106d876793f4a4e1361c3bcd#diff-6ffc1f6a5b5ad21a4dd26293c65c973b0b1a262c44d180ed86257954468991da)` got inverted

I think?

On desktop: this should also show ref count and last active

Screenshot from 2024-11-23 21-03-23

nabramow:
A few issues here:
 
 1. Some conditional rendering for mobile vs. desktop got inverted in #5157. Fixing it here.
 2. We forgot to fix the FilterDialog bbox value when we fixed the rest in #5143 (this was also an issue on Desktop we just didn't notice as it has the other search box there)
 3. We changed the value of when to show the search results on mobile to be too strict on mobile in #4521 . Previously it showed if there were any filters, location query or user pin selected. In this PR we changed it to only show if user pin selected. I changed it back to the previous condition to show if there are any filters, etc. (Please correct me if I am wrong and that was intentional).

Closes #5091 , Closes #5219 

**Web frontend checklist**
- [ x  ] There are no warnings from `yarn lint --fix`
- [ x ] All tests pass

### DIFF
--- a/app/web/features/profile/view/leaveReference/formSteps/submit/ReferenceOverview.tsx
+++ b/app/web/features/profile/view/leaveReference/formSteps/submit/ReferenceOverview.tsx
@@ -26,14 +26,14 @@ export default function ReferenceOverview({
   const { t } = useTranslation([GLOBAL, PROFILE]);
   const classes = useReferenceStyles();
   const user = useProfileUser();
-  const isAboveMedium = useMediaQuery(theme.breakpoints.up("md"));
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
 
   return (
     <>
       <TextBody className={classes.text}>
         {t("profile:leave_reference.thank_you_message")}
       </TextBody>
-      {!isAboveMedium && (
+      {isMobile && (
         <>
           <TextBody className={classes.text}>
             {t("profile:leave_reference.writing_for_text")}
--- a/app/web/features/search/SearchPage.tsx
+++ b/app/web/features/search/SearchPage.tsx
@@ -104,14 +104,6 @@ export default function SearchPage({
     Pick<User.AsObject, "userId" | "lng" | "lat"> | undefined
   >();
 
-  const hasSearchFilters = !!(
-    hostingStatusFilter.length ||
-    lastActiveFilter !== 0 ||
-    (locationResult.location.lng !== 0 && locationResult.location.lat !== 0) ||
-    numberOfGuestFilter !== undefined ||
-    queryName !== ""
-  );
-
   const [isFiltersOpen, setIsFiltersOpen] = useState(false);
 
   // Loads the list of users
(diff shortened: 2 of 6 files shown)


app/web/features/profile/view/leaveReference/formSteps/submit/ReferenceOverview.tsx
```diff
@@ -26,14 +26,14 @@ export default function ReferenceOverview({
...
-  const isAboveMedium = useMediaQuery(theme.breakpoints.up("md"));
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
 
   return (
     <>
       <TextBody className={classes.text}>
         {t("profile:leave_reference.thank_you_message")}
       </TextBody>
-      {!isAboveMedium && (
+      {isMobile && (
```
bakeiro, line 36: giphy

app/web/features/search/SearchPage.tsx
```diff
@@ -103,6 +103,15 @@ export default function SearchPage({
   const [selectedResult, setSelectedResult] = useState<
     Pick<User.AsObject, "userId" | "lng" | "lat"> | undefined
   >();
+
+  const hasSearchFilters = !!(
```
bakeiro: image

maybe makes sense to merge the logic? [line 150]

jesseallhands:
Tested and seems to be working!

I noticed that clicking  "show USERNAME on map" does not always highlight the pin if the pin is in a cluster. It just centers around the cluster rather than zooming into the "pin-level" to highlight the pin in question. I think this issue probably should be addressed in a separate issue since the primary problems are  already solved in this ticket and we need to push the fix asap.

app/web/features/search/SearchPage.tsx
nabramow: @bakeiro Oh good catch! Fixed it now!

nabramow:
> Tested and seems to be working!
> 
> I noticed that clicking "show USERNAME on map" does not always highlight the pin if the pin is in a cluster. It just centers around the cluster rather than zooming into the "pin-level" to highlight the pin in question. I think this issue probably should be addressed in a separate issue since the primary problems are already solved in this ticket and we need to push the fix asap.

Yeah sounds like a separate issue, maybe with the map. Feel free to throw up a separate ticket with the steps to replicate and how it should work.

aapeliv (approved):
LGTM!

PR #5254, opened 2024-11-26
Add useLiteUsersList and use lite users for friends list

aapeliv:
Mini refactoring here.

**Web frontend checklist**
- [ ] Formatted my code with `yarn format`
- [ ] There are no warnings from `yarn lint --fix`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #5253, opened 2024-11-26
Update UserSummary

aapeliv:
This makes a few small modifications to the `UserSummary` component:

* Adds the Strong Verifiation badge next to the name
* Slightly changes the text sizing and positioning to make it more balanced
* Removes `nameOnly` from a bunch of places, making event attendee lists etc a bit nicer to look at! (With `nameOnly` they only contain one line and it looks weird).

**Web frontend checklist**
- [x] Formatted my code with `yarn format`
- [x] There are no warnings from `yarn lint --fix`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #5259, opened 2024-11-27
Create a UsersList component and refactor

aapeliv:
~~This builds on top of #5253 and #5254.~~

Creates a unified `UsersList` component that can be used in a bunch of places.

This is what it looks like:

https://github.com/user-attachments/assets/b3de39a0-bba2-4607-9ab4-b70a4f0fbca5

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/UsersList.tsx
+++ b/app/web/components/UsersList.tsx
@@ -45,25 +45,27 @@ export default function UsersList({
   } = useLiteUsers(userIds || []);
 
   // this is undefined if userIds is undefined or users hasn't loaded, otherwise it's an actual list
-  const foundUserIds =
+  const foundUserIds: number[] | undefined =
     userIds &&
     (userIds.length > 0 ? userIds?.filter((userId) => users?.has(userId)) : []);
 
-  return (
-    <ContainingDiv>
-      {error ? (
-        <Alert severity="error">{error.message}</Alert>
-      ) : usersError ? (
-        <Alert severity="error">{usersError.message}</Alert>
-      ) : !userIds ? (
-        <CircularProgress />
-      ) : isLoadingLiteUsers ? (
+  const inner = () => {
+    if (error) {
+      return <Alert severity="error">{error.message}</Alert>;
+    } else if (usersError) {
+      return <Alert severity="error">{usersError.message}</Alert>;
+    } else if (!userIds) {
+      return <CircularProgress />;
+    } else if (isLoadingLiteUsers) {
+      return (
         <StyledUsersDiv>
           {userIds.map((userId) => (
             <UserSummary headlineComponent="h3" key={userId} user={undefined} />
           ))}
         </StyledUsersDiv>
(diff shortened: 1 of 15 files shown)


app/web/components/UsersList.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const {
+    data: users,
+    isLoading: isLoadingLiteUsers,
+    error: usersError,
+  } = useLiteUsers(userIds || []);
+
+  // this is undefined if userIds is undefined or users hasn't loaded, otherwise it's an actual list
+  const foundUserIds =
+    userIds &&
+    (userIds.length > 0 ? userIds?.filter((userId) => users?.has(userId)) : []);
```
nabramow: Rather than add a comment I would just add a type with typescript here!

```typescript
const foundUserIds: User.AsObject[] | undefined; // or whatever it is
```

app/web/components/UsersList.tsx
```diff
@@ -0,0 +1,82 @@
...
+              user={users?.get(userId)}
+            />
+          ))}
+          <>{endChildren}</>
+        </StyledUsersDiv>
+      ) : (
+        <>{emptyListChildren}</>
+      )}
+    </ContainingDiv>
+  );
```
nabramow: This is a wild ternary operator 😅. I'd say once it gets past the complicated of `is it this ? render this : otherwise render this`, it's time to restructure a bit. It's not very readable here to tell what's happening. Can you break some of these out to run separately? You could use render functions here for each "state", or break it into separate components.

It looks like this breaks into:

* errors states
* loading states
* user states (within this then the end chidlren and the empty states)

It seems these three will never happen at the same time, i.e. it will either be an error, loading, or have users so I don't think they need to depend on each other?

app/web/features/communities/events/EventAttendeesDialog.tsx
```diff
@@ -52,23 +47,7 @@ export default function EventAttendeesDialog({
...
-              <UserSummary
-                headlineComponent="h3"
-                key={id}
-                smallAvatar
-                user={user}
-              />
-            ) : null;
-          })
-        ) : null}
+        <UsersList error={error} userIds={attendeesIds} />
```
nabramow: Oooo so much cleaner I love it!

app/web/components/UsersList.tsx
aapeliv: I've added that type, but the comment was more for me to keep track of when it's undefined and when not :P

app/web/components/UsersList.tsx
aapeliv: OK, I've moved it into a function with `if-else`: the order of which one catches is important. I agree the terniary makes it hard to read (it took a while to get used to it), so I hope this is better. It still takes a moment to figure out the logic but I think it makes sense and is decently readable with the change? Let me know!

nabramow (approved):
Just tested and looks good. Thanks for making those changes!

PR #5250, opened 2024-11-25
Users by badge

aapeliv:
Continues the frontend part of #4389

**Web frontend checklist**
- [ ] Formatted my code with `yarn format`
- [ ] There are no warnings from `yarn lint --fix`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #4389, opened 2024-05-28
Badge pages

schradert:
Closes #4318 

- [x] backend: paginated ListBadgeUsers for retrieving badge's user list
- [x] frontend: badge index page
- [ ] frontend: badge user page
- [ ] frontend: tests
- [ ] frontend: storybook
- [ ] frontend: UI/UX improvements

**Backend checklist**
- [x] Formatted my code by running `autoflake --exclude src/proto -r -i --remove-all-unused-imports src && isort . && black .` in `app/backend`
- [x] Added tests for any new code or added a regression test if fixing a bug
- [x] All tests pass
- [x] Run the backend locally and it works
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Web frontend checklist**
- [x] Formatted my code with `make format`
- [x] There are no warnings from `make lint`
- [x] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ ] All tests pass
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

PR #5261, opened 2024-11-27
Users by badge

aapeliv:
~~This builds on top of #5259, which itself builds on #5253 and #5254~~!

Taking over from #5250. Continues the frontend part of https://github.com/Couchers-org/couchers/pull/4389.

TODO:
- ~~making the badges pretty on the user page~~

**Web frontend checklist**
- [ ] All tests pass

**Mobile Design**
The badges will be aligned and wrapped around. Also made the text white for easier reading against harsh BG colors. Filtering badges only work one at a time as of now. Pls refer to the desktop version for the mini bio of each badge!
Badges _ Mobile View 2


nabramow:
@aapeliv Is this ready for review? I can't tell! Feel free to move the status to "PR" if so!

aapeliv:
@nabramow: it's blocked by https://github.com/Couchers-org/couchers/pull/5259 which this builds on! I've marked this as blocked for now.

aapeliv:
This is ready for review now!

app/web/features/badges/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+import { useInfiniteQuery, useQuery } from "react-query";
+import { service } from "service";
+
+export const useBadges = () => {
+  const { data, ...rest } = useQuery(badgesKey, () =>
+    service.resources.getBadges().then((result) =>
+      result.badgesList.reduce(
+        (badgesResult, badge) => {
+          badgesResult.badges[badge.id] = badge;
+          return badgesResult;
```
nabramow: Could you switch this over to async await syntax instead? The `.then` is a bit outdated IMO. 

It means you'll need to turn the first into its own variable but that also is more readable.

app/web/features/profile/view/Badges.tsx
```diff
@@ -1,45 +1,29 @@
...
-const useStyles = makeStyles((theme) => ({
-  badgeContainer: {
-    marginTop: theme.spacing(1),
-  },
-  badge: {
-    marginInlineStart: theme.spacing(1),
-    marginBottom: theme.spacing(1),
-  },
+const StyledContainer = styled("div")(({ theme }) => ({
+  marginTop: theme.spacing(1),
```
nabramow, line 11: Yaaay!

nabramow:
Okay finally tested the branch and on Desktop it looks great! But on mobile it looks like this:

Screenshot 2024-12-19 at 9 12 56 PM

I'd suggest making a mobile view that has all the badges up top in a flexbox with wrap, then the userslists rendering below that. Or if you're feeling fancy, put the badges in a select on mobile (tbh wouldn't be bad on desktop either) as there will be quite a few badges.

nabramow:
FYI @aapeliv Shirley just added the mobile designs to the description!

aapeliv:
@nabramow: This is now ready for re-review! I addressed your comment and implemented a mobile view

app/web/features/badges/BadgesPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+import PageTitle from "components/PageTitle";
+import UsersList from "components/UsersList";
+import Badge from "features/badges/Badge";
+import { useBadges, useBadgeUsers } from "features/badges/hooks";
+import { useTranslation } from "i18n";
+import { GLOBAL, PROFILE } from "i18n/namespaces";
+
+const BadgeList = styled("div")(({ theme }) => ({
+  // display: "inline-block"
+}));
```
nabramow: Looks like you're not using this one?

aapeliv: Yes, good catch! Thanks :)

app/web/features/badges/BadgesPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+      }
+    />
+  );
+}
+
+export interface BadgesPageProps {
+  badgeId?: string;
+}
+
+export default function BadgesPage({ badgeId = undefined }: BadgesPageProps) {
```
nabramow, line 58: Might make sense to make BadgesPage a separate file? Easier to scroll through I think.

app/web/features/badges/BadgesPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { t } = useTranslation([GLOBAL, PROFILE]);
+  const { badges, isLoading: isBadgesLoading } = useBadges();
+  const theme = useTheme();
+
+  const isBelowMd = useMediaQuery(theme.breakpoints.down("md"));
```
nabramow: Can you call this `isMobile`? Switching over to that as it's easier to understand at a glance.

app/web/features/badges/BadgesPage.tsx
```diff
@@ -0,0 +1,145 @@
...
+              <>Badge not found</>
+            )}
+          </ContentDiv>
+        ) : (
+          <CenteredDiv>
+            <Typography variant="subtitle1">
+              {t("profile:badges.click_on_left")}
+            </Typography>
+          </CenteredDiv>
+        )}
```
nabramow: There are a lot of ternary operators in here. Is there a way to break up this logic to make it more readable?

nabramow:
Okay functionality-wise looks good and works well on both mobile and desktop! Just some small comments, and noticed one mobile design thing:

Screenshot 2025-01-26 at 11 09 33 AM

Maybe we need to add some more margin for long usernames? Or `overflow:hidden` on the top parent component with a defined width?

app/web/features/badges/BadgesPage.tsx
aapeliv: I tried to think about it a bit; to my eye this hits a good balance with not being *too* conditional but also being readable and concise. I'm not sure how to break it apart in a way that doesn't just separate the logic too much.

Open to suggestions though!

app/web/features/badges/BadgesPage.tsx
aapeliv: What do you mean, sorry?

nabramow: Ah instead of having two components in one file, break it into a separate file.

app/web/features/badges/BadgesPage.tsx
nabramow: Okay I see what you mean, I think it's okay for now as probably this logic won't get more complicated or expanded on much.

aapeliv:
@nabramow this is now ready for review again! I think it should be done, fingers crossed :)

nabramow (approved):
Looks good and works well for me now on desktop and mobile!

PR #5283, opened 2024-12-01
Update zh_Hans translation

HankAviator:
Update zh_Hans translation


bakeiro (approved):
I trust these translations :D

PR #5294, opened 2024-12-04
frontend - feat: add new list view of users on map #4848

### LINKED ISSUE #4848: [map] [mobile] add new list view of users on map
**Current:**
image

**Desired situation**
image

**Notes:**
image

marco-digennaro:
Closes #4848 

**Web frontend checklist**
- [x ] There are no warnings from `yarn lint --fix`
- [x] All tests pass
- [ xx] Clicked around my changes running locally and it works


jesseallhands:
@marco-digennaro this is looking really good! I'm super excited to this feature to be rolled out as is should improve mobile QoL by a lot.

Here are a few things I noticed in the preview (from a UI/UX perspective, I'm not a FE dev):

The spacing between the cards is too much and takes up a lot of screen real estate:
Screenshot_20241204_193034_Firefox

When the user selects a single pin, it should not show the expand drawer arrow ^ because there is no point of expanding the drawer upward for a single user card. Also, the spacing above and below the card is too much:
Screenshot_20241204_192531_Firefox

When no pin is selected, the drawer still appears, which should not happen:
Screenshot_20241204_192819_Firefox

After clicking on a pin, if the user zooms in and clicks on "search in this area" the card for the previously clicked pin persists, even if that pin is no longer in the search area:
https://github.com/user-attachments/assets/e32ba8e2-1ccf-444a-81c8-49f5864811c6

It's possible some of the map behavior issues may be caused by the next instance, but I'm not sure. Perhaps @nabramow and/or @darrenvong can assist here in helping assess the causes.

bakeiro:
this is so cool feature 🤩 , thanks for this!

bakeiro:
code looks great, once the UI feedback is done, and the build isn't red I can re-check and most likely approve :)

app/web/features/search/SearchResultsList.tsx
```diff
@@ -132,62 +139,72 @@ export default function SearchResultsList({
...
-          <TextBody className={classes.baseMargin}>
-            {t("search_result.no_user_result_message")}
-          </TextBody>
-        )}
+          {!isLoading && hasAtLeastOnePageResults && (
+            <Paper>
+              {isLoading ||
+                (selectedUserData.isLoading && (
+                  <CircularProgress className={classes.baseMargin} />
+                ))}
```
nabramow: Nit - turn the two isLoading conditions into a variable to make it more readable. Took me a minute to figure out what was happening here.

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,113 @@
...
+    position: "sticky",
+    top: 0,
+    backgroundColor: theme.palette.background.default,
+    borderRadius: 0,
+    zIndex: theme.zIndex.drawer,
+  },
+  icon: {
+    fontSize: "3rem",
+  },
+}));
```
nabramow: Since this is a new component, could you please use the new MUI styled way of doing the CSS? We're slowly trying to remove all of this way as it's not compatible with the next MUI version.

Just search "Styled" for some examples or you can look at: app/web/components/Footer/Footer.tsx.

marco-digennaro:
> It's possible some of the map behavior issues may be caused by the next instance, but I'm not sure. Perhaps @nabramow and/or @darrenvong can assist here in helping assess the causes.

yes, i tried to investigate, it's cause of this piece of code
`  const wasResultFound =
    resultsList?.find(
      (value) => value.user?.userId === selectedResult?.userId
    ) !== undefined;

  if (!wasResultFound && selectedUserData.data) {
    resultsList = [{ user: selectedUserData.data, rank: 0, snippet: "" }];
  }`
  which persist che selected user in the search result 
 @nabramow is it necessary?

marco-digennaro:
@bakeiro @nabramow @jesseallhands made some adjustments, please check :)

jesseallhands:
The spacing issues are completely resolved! Thanks!

While testing, I ran into an issue where when I clicked on certain pins, it would not center the card within the drawer (it would land somewhere in between cards) see video below:


https://github.com/user-attachments/assets/9d38d230-832d-4b0c-baf7-98de7d1d1005

There is also some strange behavior when drawer initially opens. It's almost as if it is showing my own card (and the Chevron for the drawer)  just before opening or just before closing. See video:

https://github.com/user-attachments/assets/889f8566-0feb-432f-9094-44ff4f803909

In addition to resolving that problem, I would also recommend using mountOnEnter and unmountOnExit so the drawer slides in from the bottom upward, which also hints to the user that it is a slidable drawer. I think the same can be used to slide down the drawer when minimizing from the "full screen" list view to the Single-Card view and also when closing the drawer altogether.

Thanks once again for working on this feature! I'm very excited about it and I think mobile users will really love it!

bakeiro:
>  Screen_Recording_20241205_201332_Firefox.mp4 
> In addition to resolving that problem, I would also recommend using mountOnEnter and unmountOnExit so the drawer slides in from the bottom upward, which also hints to the user that it is a slidable drawer. I think the same can be used to slide down the drawer when minimizing from the "full screen" list view to the Single-Card view and also when closing the drawer altogether.

Although I agree on this I am not sure how we can solve this as first it is nothing new, it is something old [and in production already....]

The behavior we see on the video is the difference between openning a result which is on the loaded list vs opening a result which isn't in this list, (you may wonder which list is the one on the left in the desktop version).

the background for this is, at the beginning I implemented to show only a subset of results, plus the pagination, what happened is that we wanted to implement to show all users at the first render, because it was something that sold very well.

Since loading all users in the list isn't possible (cause it's to big) therefore, when you click on a user, if this is within the list, it moves and puts the user in the center, otherwise it clears the list temporarily and shows specifically that particular user.

Edit:
I thinking about 2 possible solutions:

- Not to show a list at the beginning, in the first render, then whenever you click on a pin, it shows that particular user.
And once you do a search then it shows that result in the list.

- Use pagination, and display on the map only what's on the list, plus to prevent the map seeing empty, display the user's surroundings (without much zoom).  

apart of this we can create a investigation ticket to find a better solution

marco-digennaro:
> There is also some strange behavior when drawer initially opens. It's almost as if it is showing my own card (and the Chevron for the drawer) just before opening or just before closing. See video:

i think there's a different behaviour of the map between mobile and desktop, and apparently i can't reproduce that bug on desktop browser with mobile view.
if i understood the problem on mobile the results change on map move (?), if so i added a (maybe) fix, but again, i couldn't reproduce it so i hope it would work.

PS: where can i find the vercel previews?

bakeiro:
> > There is also some strange behavior when drawer initially opens. It's almost as if it is showing my own card (and the Chevron for the drawer) just before opening or just before closing. See video:
> 
> i think there's a different behaviour of the map between mobile and desktop, and apparently i can't reproduce that bug on desktop browser with mobile view. if i understood the problem on mobile the results change on map move (?), if so i added a (maybe) fix, but again, i couldn't reproduce it so i hope it would work.
> 
> PS: where can i find the vercel previews?

Vercel preview :)
image

marco-digennaro:
> > There is also some strange behavior when drawer initially opens. It's almost as if it is showing my own card (and the Chevron for the drawer) just before opening or just before closing. See video:
> 
> i think there's a different behaviour of the map between mobile and desktop, and apparently i can't reproduce that bug on desktop browser with mobile view. if i understood the problem on mobile the results change on map move (?), if so i added a (maybe) fix, but again, i couldn't reproduce it so i hope it would work.
> 
> PS: where can i find the vercel previews?

it didnt work, i need to investigate further why it happens, or if @bakeiro knows already where to look, please write me so i'll try to fix :)

bakeiro:
> > > There is also some strange behavior when drawer initially opens. It's almost as if it is showing my own card (and the Chevron for the drawer) just before opening or just before closing. See video:
> > 
> > 
> > i think there's a different behaviour of the map between mobile and desktop, and apparently i can't reproduce that bug on desktop browser with mobile view. if i understood the problem on mobile the results change on map move (?), if so i added a (maybe) fix, but again, i couldn't reproduce it so i hope it would work.
> > PS: where can i find the vercel previews?
> 
> it didnt work, i need to investigate further why it happens, or if @bakeiro knows already where to look, please write me so i'll try to fix :)

I believe that both issues can be solved if we solve the list issue, as I say before depends how we want to solve it

- One solution could be not to show a list at the beginning, then when you click on a pin, it shows that particular user
- Use Pagination + subsets, and in the first render, show the user surrondings, plus don't allow the user to see the map zommed out

if other person has other suggestion/solution please say, I personally think the best one it's to do the first solution [not show the list at the beginning] and then create a re-search ticket for the future.  

If this solution is outside the scope of this PR/ticket I can create another ticket for this :) [then we can merge both PRs together, or you can do a PR from this PR as you prefer...]

write me on Slack if you need more info about this!

bakeiro:
Update: Marco and I talked about this, our plan it's to merge this PR (unless a new bug is found, code suggestions or whatever) and then we will do pair programming to see how to address these bugs [in another PR/ticket]

bakeiro (dismissed):

bakeiro:
I've approved it, but we don't have to merge it right now... just in case someone has other feedback/comment to made :)

nabramow:
> Update: Marco and I talked about this, our plan it's to merge this PR (unless a new bug is found, code suggestions or whatever) and then we will do pair programming to see how to address these bugs [in another PR/ticket]

Just checking - these bugs already existed yeah? Or were they created in this PR?

bakeiro:
> > Update: Marco and I talked about this, our plan it's to merge this PR (unless a new bug is found, code suggestions or whatever) and then we will do pair programming to see how to address these bugs [in another PR/ticket]
> 
> Just checking - these bugs already existed yeah? Or were they created in this PR?

The inconsistencies between clicking a user which is on the list vs clicking a user which isn't on the list is something old, now we have this new view we see the same problem a bit different

nabramow (changes requested):
Okay just tested! This works on initial load of the map, if I click the "Search this area" button, then it appears. That part looks great!

However, if I click to search a location via the filters, that results list doesn't have the new scrollable feature.

Here's a video showing what I mean:

https://github.com/user-attachments/assets/941b5fd7-3854-40f7-b0f6-1d5beb0c4609

I assume the search results is a separate component maybe and we need to add this functionality there?

bakeiro:
> Okay just tested! This works on initial load of the map, if I click the "Search this area" button, then it appears. That part looks great!
> 
> However, if I click to search a location via the filters, that results list doesn't have the new scrollable feature.
> 
> Here's a video showing what I mean:
> 
>  Screen.Recording.2024-12-08.at.11.19.27.AM.mov 
> I assume the search results is a separate component maybe and we need to add this functionality there?

I think this is due it's only 1 results... so only displays that specific result in the new div

app/web/features/search/SearchPage.tsx
```diff
@@ -180,46 +179,21 @@ export default function SearchPage({
     <QueryClientProvider client={queryClient}>
       <HtmlMeta title={t("global:nav.map_search")} />
       <div className={classes.container}>
-        {/* Desktop */}
```
bakeiro, line 183: unnecesary code

bakeiro:
ready to re-review :)

bakeiro:
to give context:
> Screenshot_20241204_192819_Firefox

this is not related with this PR, will be solved once this gets merged :)

app/web/features/search/SearchPage.tsx
```diff
@@ -174,52 +172,35 @@ export default function SearchPage({
     locationResult.location.lat,
   ]);
 
+  /**
```
bakeiro, line 175: fixes the bug when a selected result remains even after searching

app/web/features/search/SearchResultsList.tsx
```diff
@@ -123,6 +128,10 @@ export default function SearchResultsList({
     .flatMap((page) => page.resultsList)
     .filter((result) => result.user);
 
+  if (isMobile && !wasSearchPerformed && !searchFromDashboard) {
```
bakeiro, line 131: fix blinking

nabramow (changes requested):
Okay two things here:

1. The up and down arrows are really big, thick and have a lot of padding around them. That seems non-standard to me. Especially since mobile space is limited, could you make the arrow smaller and less pixels of padding above and below?

2. Can we just always show this version on mobile even if there is only one result? It seems confusing that it switches when there's more results. I think it's fine if it shows even with one result for consistency.

Otherwise works great functionality-wise so just those 2 design things!

bakeiro:
> 2. Can we just always show this version on mobile even if there is only one result? It seems confusing that it switches when there's more results. I think it's fine if it shows even with one result for consistency.

@nabramow 

I totally agree here, the problem is that we show all the users the first time (bc is a big selling point apparently) and this creates issues, since we need to change the behavior after the user performs a search/applies a filter, I've been complaining about this during long time, I've created this ticket to explore different solutions https://github.com/Couchers-org/couchers/issues/5348

we can talk about this (and we should) but is basically this switch between the first render and after the search which causes a lot of issues, but apparently is better this behavior inconsistency than not showing all the users at the first render...

bakeiro:
ok, I think a solution can be, when a user searchs from dashboard, we return the subset, so we display the small section with the arrow, and only display the one result box when we click "map search" link in the navbar??

bakeiro:
I'm thinking about a solution for this... I came up with this idea:
Clicking a result on the map will always show a small bubble on the botton, regardless we display all the users on the map or only a subset, to avoid the weird behaviour, and we display the box with the arrow to toggle the users list only when there is a search/filter (aka subset)

I feel this way it's better than re-locating the list on the button to display the selected user... [basically something like airbnb is doing!]

more info #5405

bakeiro:
small CSS change pushed :)

nabramow:
@bakeiro This looks way better, thanks for making that change! Way more consistent this way.

Just that one comment about the arrow then, would it be possible to make the arrow a bit smaller? Especially the up arrow seems huge and on mobile it's usually thinner and smaller.

bakeiro:
making arrow smaller again, please, let me know if its good enough now @nabramow  :)

app/web/features/search/SearchPage.tsx
nabramow: Sorry I just saw this. I think it makes more sense to set this in the same place the search happens rather than a separate useEffect? So where you set the new location, then set this there instead.

app/web/features/search/SearchPage.tsx
nabramow: Why was it unnecessary? Looks like a big chunk removed.

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,115 @@
...
+const StyledCloseButton = styled(IconButton)(({ theme }) => ({
+  width: "100%",
+  marginLeft: 0,
+  position: "sticky",
+  maxHeight: "60px",
+  top: 0,
+  backgroundColor: theme.palette.background.default,
+  borderRadius: 0,
+  zIndex: theme.zIndex.drawer,
+  "& svg": { fontSize: "3rem" },
```
nabramow, line 57: I would remove this here

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,115 @@
...
+}));
+
+const StyledOpenButton = styled(IconButton)(({ theme }) => ({
+  width: "100%",
+  marginLeft: 0,
+  borderRadius: "15px 15px 0 0",
+  boxShadow: "0px -4px 5px 0px rgba(17, 17, 26, 0.08)",
+  backgroundColor: theme.palette.background.default,
+  maxHeight: "50px",
+  "& svg": { fontSize: "4rem" },
```
nabramow, line 38: and remove this here (they are different sizes by the way). see below

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,115 @@
...
+      height: "auto",
+      "& .MuiCardContent-root": {
+        height: "auto",
+      },
+      "& .MuiCardActionArea-root": {
+        height: "100%",
+      },
+    },
+  },
+}));
```
nabramow, line 79:
```suggestion
const StyledVerticalList = styled(List)(({ theme }) => ({
  display: "flex",
  flexDirection: "column",
  alignItems: "center",
  gap: theme.spacing(2),
  padding: theme.spacing(2),
  
  "&:first-child": {
    paddingTop: 0,
  },
  "& .MuiCard-root": {
    height: "auto",
    width: "100%",
    [theme.breakpoints.down("md")]: {
      height: "auto",
      "& .MuiCardContent-root": {
        height: "auto",
      },
      "& .MuiCardActionArea-root": {
        height: "100%",
      },
    },
  },
}));
```

nabramow: Add this `&:first-child` to remove the padding from the top item so there's not so much space

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  useEffect(() => {
+    setShouldShowOpenButton(resultsSnippet.length > 1 && !open);
+  }, [resultsSnippet, open]);
+
+  return (
+    <StyledDiv>
+      {shouldShowOpenButton && (
+        <StyledOpenButton onClick={toggleDrawer(true)}>
+          <ExpandLess />
```
nabramow, line 100:
```suggestion
          <ExpandLess sx={{ fontSize: "24px"}}/>
```

Do size here so you're not targeting all svgs. I think for icons a static size is okay, especially as this is only mobile.

app/web/features/search/SearchResultsMobileVerticalList.tsx
```diff
@@ -0,0 +1,115 @@
...
+        </StyledOpenButton>
+      )}
+
+      <StyledDrawer open={open}>
+        {open && (
+          <StyledCloseButton onClick={toggleDrawer(false)}>
+            <ExpandMore />
```
nabramow, line 107:
```suggestion
          <ExpandLess sx={{ fontSize: "24px"}}/>
```

Same here.

app/web/theme.ts
```diff
@@ -47,6 +47,9 @@ const themeOptions: ThemeOptions = {
           textDecoration: "none",
           color: "inherit",
         },
+        '.maplibregl-ctrl-bottom-right': {
+          bottom: '7px !important',
+        }
```
nabramow, line 52: I'm a little nervous about all these fixes mixed in here. Could we keep this PR to this feature and put the fixes in another PR?

nabramow:
Looks like prettier needs to run FYI.

app/web/theme.ts
nexagoai: hmmm it's related 😅 I can explain where and how (but lately I don't have much time for this 😞)

nabramow:
Closing as full map re-write is in progress.

PR #5047, opened 2024-10-26
[needs attention] - frontend/feature: add guest menu on landing page, refactor cookie banner

marco-digennaro:
- It adds a guest menu on navbar with links to sign-in, sign-up, report bug, help
- it refactors the cookie banner fixing it at bottom

About last point check:
https://github.com/Couchers-org/couchers/issues/4127#issuecomment-2439445903

Closes #4127

**Backend checklist**
- [ ] Formatted my code by running `ruff check --select I --fix . && ruff check . && ruff format .` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [ ] All tests pass
- [ ] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Web frontend checklist**
- [ x] Formatted my code with `yarn format`
- [ x] There are no warnings from `yarn lint --fix`
- [ ] There are no console warnings when running the app
- [ ] Added any new components to storybook
- [ ] Added tests where relevant
- [ x] All tests pass
- [x ] Clicked around my changes running locally and it works
- [ x] Checked Desktop, Mobile and Tablet screen sizes

PR #5302, opened 2024-12-08
Web/feature/report a bug (issue #5274)

### LINKED ISSUE #5274: Add "report a bug" feature to the navbar
### Problem
The bug reporting feature that previously existed on the website has disappeared. There used to be a button in the upper-right corner of the screen next to the hamburger-menu button.

Because of this, users cannot easily report bugs like they used to be able to. Users are only given the opportunity to report a bug if an error occurs and we have error handling for it. In all other situations the user only has the option to email support or open a ticket on github (if they event know how to do that). 

Here is what it looked like:
image

### Solution
The quickest solution would be to restore the bug-reporting feature we used to have (red button next to the hamburger-menu button in the upper-right of screen). I'm not sure when this feature was removed, but it was fairly recently (within the past 6 months). We should be able to use the old code to bring it back quickly.

Related to #5266 

Edit: I am splitting the Sentry-based bug-reporting "upgrade" into a separate issue #5291

graceegal:
Closes #5274 

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/web/components/Navigation/LoggedInMenu.tsx
+++ b/app/web/components/Navigation/LoggedInMenu.tsx
@@ -6,6 +6,7 @@ import useCurrentUser from "features/userQueries/useCurrentUser";
 import React, { Dispatch, ReactNode, SetStateAction } from "react";
 import { theme } from "theme";
 import makeStyles from "utils/makeStyles";
+
 import ReportButton from "./ReportButton";
 
 const useStyles = makeStyles((theme) => ({
--- a/app/web/components/Navigation/LoggedInMenu.tsx
+++ b/app/web/components/Navigation/LoggedInMenu.tsx
@@ -6,6 +6,7 @@ import useCurrentUser from "features/userQueries/useCurrentUser";
 import React, { Dispatch, ReactNode, SetStateAction } from "react";
 import { theme } from "theme";
 import makeStyles from "utils/makeStyles";
+import ReportButton from "./ReportButton";
 
 const useStyles = makeStyles((theme) => ({
   menu: {
@@ -34,6 +35,9 @@ const useStyles = makeStyles((theme) => ({
     width: "2rem",
     marginLeft: theme.spacing(1),
   },
+  reportBtnContainer: {
+    padding: theme.spacing(2),
+  },
 }));
 
 export default function LoggedInMenu({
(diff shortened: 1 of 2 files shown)


jesseallhands:
Yey! Bug reporting button is back, and I confirmed via test that it works. Thank you!!!

The only thing I noticed is that the bug report button is no longer on the landing page prior to log-in. It used to be in the spot circled in the screenshot below. I think the bug report button was there prior to log-in to allow people to report issues in case of problems logging in. @aapeliv can you confirm?

image

nabramow:
> The only thing I noticed is that the bug report button is no longer on the landing page prior to log-in. It used to be in the spot circled in the screenshot below. I think the bug report button was there prior to log-in to allow people to report issues in case of problems logging in. @aapeliv can you confirm?

Ah yes I can confirm that! It was removed in this PR #5047 @graceegal if you wanna use that to undo it. See the `Navigation` component in that PR where the `ReportButton` is commented out.

graceegal:
@jesseallhands @nabramow sure! Actually, It looks like the ReportButton is still embedded in the GuestMenu but just as a menu button link (see screenshot), rather than the red button. Do you want me to remove it as a menu link and add it as a button to the left of the hamburger button, like it is in the logged in view?  Or I can add it in and leave the menu link remaining as well. 
Screenshot 2024-12-08 at 09 04 35

jesseallhands:
> Do you want me to remove it as a menu link and add it as a button to the left of the hamburger button, like it is in the logged in view? Or I can add it in and leave the menu link remaining as well.

On the one hand, it doesn't have much visibility if it is in the hamburger menu, but on the other hand, the functionality is there if people need it! I just tested it and it works from the hamburger menu, so I'd say let's just leave it like that for now.

graceegal:
> > Do you want me to remove it as a menu link and add it as a button to the left of the hamburger button, like it is in the logged in view? Or I can add it in and leave the menu link remaining as well.
> 
> On the one hand, it doesn't have much visibility if it is in the hamburger menu, but on the other hand, the functionality is there if people need it! I just tested it and it works from the hamburger menu, so I'd say let's just leave it like that for now.

Agreed on visibility! I will leave it for now but can easily change it later

marco-digennaro (dismissed):

nabramow (approved):
Approving again since the last got lost after lint fix.

PR #5306, opened 2024-12-08
Mobile style fixes for references list

nabramow:
I was clicking around the references list to answer some questions about it and noticed the reference text and tab scrollbar where overflowing the view on mobile. Decided to quickly throw up a fix here.

* Shows the tab scrollable arrows also on mobile view. Before IMO it wasn't clear there were more tabs on mobile.
* Prevent reference text from overflowing view on mobile

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/components/TabBar/TabBar.tsx
+++ b/app/web/components/TabBar/TabBar.tsx
@@ -5,7 +5,7 @@ import makeStyles from "utils/makeStyles";
 export const useStyles = makeStyles((theme) => ({
   messagesTab: {
     [theme.breakpoints.down("md")]: {
-      overflow: "visible",
+      // overflow: "visible",
       margin: `0 ${theme.spacing(2)}`,
     },
   },
@@ -34,7 +34,8 @@ export default function TabBar<T extends Record<string, React.ReactNode>>({
       onChange={handleChange}
       indicatorColor="primary"
       textColor="primary"
-      scrollButtons="auto"
+      scrollButtons={true}
+      allowScrollButtonsMobile
       variant="scrollable"
     >
       {Object.entries(labels).map(([value, label]) => (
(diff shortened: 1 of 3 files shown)


jesseallhands:
We should probably add the scrollability of the tabs when it is required (eg. a small mobile screen) but on my screen there should be plenty of room to display all of the tabs without need for scrolling:

Screenshot_20241208_200031_Firefox

jesseallhands:
@nabramow This might be unrelated, but I also noticed that the navigation jumps to the top of the screen when I click in certain areas of the reviews cards. See video:

https://github.com/user-attachments/assets/e8cc58e1-2d37-4394-ab19-398613d13aae

nabramow:
> We should probably add the scrollability of the tabs when it is required (eg. a small mobile screen) but on my screen there should be plenty of room to display all of the tabs without need for scrolling:

Fixed!

nabramow:
> @nabramow This might be unrelated, but I also noticed that the navigation jumps to the top of the screen when I click in certain areas of the reviews cards. See video:

I don't think this is related to my PR, but regardless I wasn't able to replicate this. Where specifically are you clicking that causes this? It's not totally clear form the video, it looks like you're clicking links in the footer?

If you click the "UserSummary", that will take you to that user's profile, just FYI.

jesseallhands:
Tabs look great now! Thank you!!!

> I don't think this is related to my PR, but regardless I wasn't able to replicate this. Where specifically are you clicking that causes this? It's not totally clear form the video, it looks like you're clicking links in the footer?

I just tested again and it doesn't happen from clicking. It happens from swiping, and only on mobile. If I swipe any area of the references it pops me up to the top of the page automatically. If I just tap the area of the references, nothing happens. Very strange! I think we open a separate ticket on this issue as it seems unrelated.

jesseallhands (approved):

PR #5314, opened 2024-12-10
Upgrade React to v18 and react-hook-form to v7.

nabramow:
Upgrading React to v18. Big change to note here is React now has automatic batching. Batching is when React groups multiple state updates into a single re-render for better performance. [Full release info for React 18 here.](https://react.dev/blog/2022/03/29/react-v18)

This conflicted with react-hook-form v6, and since we use that everywhere for the main users flows, I upgraded that to v7 too.

**So when testing - testing the forms is the most important!**

The major change with `react-hook-form` is how refs are handled. Hence why you see some forwarding refs, etc. added.

I also updated all the `react-testing-library` packages because I needed the tests to support the automatic batching. This had some breaking changes for `userEvent`, which is now awaited.

Except not always...it doesn't work with the Datepicker. I'm hoping/assuming that will be fixed once we update MUI further, or at least mui-x-datepickers. For now I just went with the partially awaited partially not chaos.

**Web frontend checklist**
- [ x ] All tests pass


app/web/package.json
```diff
@@ -110,5 +111,8 @@
     "timezone-mock": "^1.3.2",
     "typescript": "~5.7.2",
     "whatwg-fetch": "^3.6.20"
+  },
+  "resolutions": {
+    "@types/react": "^18.3.1"
```
nabramow, line 116: I added this as `react-simple-maps`'s  sub-dependencies are pulling in a conflicting version of @types/react (v19). 

You can see this by running `yarn list @types/react`. Running dedupe didn't work so enforcing v18 of types for now.

Already on the latest version of `react-simple-maps`, which doesn't seem like it's maintained anymore...but that's a problem for another day.

aapeliv: I was playing around with `react-simple-maps` the other day, we might want to just ditch it at some point. @polemius also had issue with that on mobile, so we'd want a cross-platform solution.

nabramow: @aapeliv @polemius I just made a ticket for this here: #5318 !

@polemius if you want to add a few details to that ticket about what the issues we should be aware of on mobile were (or tell me I can add), that would be great!

bakeiro:
awesome job 💯

app/web/components/Datepicker.tsx
```diff
@@ -30,7 +31,7 @@ interface DatepickerProps {
   testId?: string;
 }
 
-export default function Datepicker({
+const Datepicker = ({
```
aapeliv, line 34: Why'd you change this from `export default` and move it to the bottom?

app/web/components/Datepicker.tsx
```diff
@@ -45,63 +46,72 @@ export default function Datepicker({
...
-          renderDay={(day, selectedDates, pickersDayProps) => {
-            return (
-              <PickersDay
-                {...pickersDayProps}
-                style={{
-                  ...(pickersDayProps.selected && {
-                    backgroundColor: theme.palette.primary.main, // make selected day our primary color
-                  }),
+      render={({ field }) => {
+        return (
```
aapeliv: Any reason for doign `{return(...)}` instead of `(...)`?

app/web/features/auth/logins/LoginCard.tsx
```diff
@@ -79,28 +79,38 @@ export default function LoginsPage({
...
-              Near{" "}
-              <strong>
-                {{ approximate_location: session.approximateLocation }}
-              </strong>
+            <Trans
+              t={t}
+              i18nKey="auth:active_logins.location"
+              values={{ approximate_location: session.approximateLocation }}
+            >
+              {`Near `}
```
aapeliv, line 87: Is there a reason for these bing like this instead of just `Near <strong>...`?

app/web/features/communities/events/EventForm.tsx
```diff
@@ -118,18 +120,24 @@ export default function EventForm({
...
+
+  const timeDelta = useRef(60);
+  const defaultEndTime = useMemo(
+    () =>
+      dayjs()
+        .add(1, "hour")
+        .add(timeDelta.current, "minutes")
+        .format("HH:[00]"),
+    []
+  );
```
aapeliv: I see this is used a couple of times, maybe worth moving to a utility function?

Also I'm not entirely following the default date/time logic here. For `defaultDate`, why does it not suffice to add 1 hour, won't that roll over to the next day?

Or maybe it's just simpler to make the default start date/time 24 hours ahead from now?

aapeliv:
Here is the list of forms changed:

* ContributorForm (when signing up)
* Date picker (signup birthdate?, host request dates?)
* Map?
* Report button
* Flag button
* Community guidelines
* Delete account
* Change email
* Location selection in jail
* Login form
* Current logins page
* Change password
* Reset password
* Change phone #
* Sign up form
* Edit community info page
* Create discussion
* Event creation
* Event editing (in particular dates/times)
* Donation
* Group chats: create/edit/send/invite
* Profile editing (me and home)
* New host req
* Leave ref
* Search + filtering

app/web/features/auth/signup/AccountForm.tsx
```diff
@@ -45,7 +45,10 @@ type SignupAccountInputs = {
   gender: string;
```
aapeliv, line 45: (I'm adding testing things as comments so we can keep track of them and resolve them easier)

When signing up, the location screen no longer highlights the place to enter the location if it's missing.

---
Screenshot from 2024-12-12 12-41-07
---

app/web/features/auth/signup/AccountForm.tsx
```diff
@@ -201,14 +205,8 @@ export default function AccountForm() {
         <Datepicker
```
aapeliv, line 205: The signup form now (not sure if it's this PR introducing it) defaults to the current date 18 years ago as the default. It should default to today (or some other invalid date that forces the user to select something real).

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -89,18 +89,16 @@ export default function NewHostRequest({
```
aapeliv, line 89: The host request form doesn't clear after sending one:

1. Go to a profile. click request
2. Write a host req and "send"
3. Click on request again

The values from 2. will be there, they should be cleared.

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -89,18 +89,16 @@ export default function NewHostRequest({
 
   const {
```
aapeliv, line 90: The request form allows picking dates in the past (but this is caught by the backend):

This branch:

Screenshot from 2024-12-12 12-52-37

Develop:

Screenshot from 2024-12-12 12-52-44

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -39,8 +39,17 @@ import {
 } from "./constants";
```
aapeliv, line 39: Issues with the map:

1. When you first load the page and make changes to the map (e.g. change radius with the slider, or double click to pick location like that): the map will refresh/redraw/flicker.
2. Having an empty "Display location" does not stop you from submitting the form.

This screencap shows both:

https://github.com/user-attachments/assets/4718fdf7-034b-4d30-a8aa-877d6e8e3962

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -39,8 +39,17 @@ import {
...
 import useStyles from "./styles";
```
aapeliv, line 40: Empty name does not have a validation error message (not introduced in thsi branch):

Screenshot from 2024-12-12 12-58-42

app/web/components/Navigation/ReportButton.tsx
```diff
@@ -239,33 +239,32 @@ export default function ReportButton({
               </DialogContentText>
```
aapeliv, line 239: Not introduced in this PR; but there are no errors on the bug report form, it doesn't say "this field is required" or highlight in red. It just moves focus to the first missing field.

OK it turns out a whole bunch of forms are missing validation errors or red highlighting. I'm not sure if this is something we want to implement at some point?

app/web/features/auth/email/ChangeEmail.tsx
```diff
@@ -58,8 +58,13 @@ export default function ChangeEmail({ className, email }: ChangeEmailProps) {
       <Typography variant="h2">{t("auth:change_email_form.title")}</Typography>
```
aapeliv, line 58: This form is also missing validation errors.

app/web/features/auth/phone/ChangePhone.tsx
```diff
@@ -80,7 +80,8 @@ export default function ChangePhone({
     handleSubmit: verifyHandleSubmit,
```
aapeliv, line 81: Translations are broken (not sure if introduced in this PR):

Screenshot from 2024-12-12 13-14-45
---
Screenshot from 2024-12-12 13-39-30
---
Screenshot from 2024-12-12 13-39-48

app/web/features/donations/Donations.tsx
```diff
@@ -150,14 +150,18 @@ export default function Donations() {
```
aapeliv, line 150: I can't donate a custom amount:

Screenshot from 2024-12-12 13-37-56

app/web/features/profile/view/leaveReference/formSteps/Appropriate.tsx
```diff
@@ -42,7 +42,12 @@ export default function Appropriate({
   const theme = useTheme();
```
aapeliv, line 42: Not sure if we should just move the error next to the Q?

Screenshot from 2024-12-12 13-47-22

app/web/features/profile/edit/EditProfile.tsx
aapeliv: Map is also flickering on the map search page

app/web/components/Map.tsx
```diff
@@ -84,9 +84,6 @@ export default function Map({
   useEffect(() => {
     if (!containerRef.current) return;
 
-    //don't create a new map if it exists already
-    if (mapRef.current) return;
-
```
aapeliv, line 89: Might this be related to the flickering?

app/web/components/Map.tsx
```diff
@@ -84,9 +84,6 @@ export default function Map({
   useEffect(() => {
```
aapeliv, line 84: Map search page issues:

1. The map flickers when it first loads
2. The map flickers when you click on "Show XX on the map"
3. I'm unable to click on pins on the map! Only mouse interaction I can do on the map itself is panning.
4. Clicking on "Search this area" causes the "Ooops... that's an unknown error" screen!
5. Searching for a location also causes the oopsie screen

aapeliv (commented):
OK! I've done some manual testing. List:

Not tested:
* Location selection in jail
* Edit community info page

Tested:
* Sign up form
* Community guidelines
* ContributorForm (when signing up)
* Date picker (signup birthdate?, host request dates?)
* New host req
* Report button
* Flag button
* Change email
* Current logins page
* Change password
* Event creation
* Event editing (in particular dates/times)
* Donation
* Change phone #
* Group chats: create/edit/send/invite
* Login form
* Reset password
* Profile editing (me and home)
* Leave ref
* Create discussion
* Map?
* Search + filtering
* Delete account

jesseallhands:
This join community button does something funky when you leave the community and click the button again to rejoin:

Screenshot_20241213_212058_Firefox

nexagoai:
small UI detail, not a blocker for me
image

nexagoai:
looks way to big
`app/web/features/messages/requests/HostRequestListItem.tsx`
image

nexagoai:
any filter on the map breaks it
image

nexagoai:
[profile page]
not important
weird squares appears when clicking on a map, [if takes to much work I don't think it's important for this PR]
image

app/web/features/profile/view/NewHostRequest.tsx
nexagoai: also happens when creating an event, you can specify a past date

app/web/components/Datepicker.tsx
nabramow: Ah just a style preference TBH. I needed to add forwardRef to a lot of components as well and it just looked nicer.

nabramow: I mean there is a technical difference between a function declaration and an arrow function assigned to a variable in terms of scope, hoisting, etc. Technically I'd say a block-scoped variable has less chance of side effects but that wasn't really my reason.

app/web/components/Datepicker.tsx
nabramow: Ah I was probably console.logging in there and forgot to remove it haha. Will do now.

app/web/features/auth/logins/LoginCard.tsx
nabramow: Yes typescript didn't like it that way! I upgraded the `@types/react` package as part of this upgrade. It also stopped working the old way. It wanted the translation variable on its own line and then to define the value of the variables to be translated now.

app/web/features/communities/events/EventForm.tsx
nabramow: I didn't change the logic, just moved it, so I rather leave that for now and make a separate ticket.

What I did do was set defaultValues in the useForm and move the defaultValues up to the parent component. It was throwing an error about the `.startOf` dates being undefined in the child component. I'm guessing the order of stuff rendering caused this with the new automatic batching for React v18. At least that's my best guess.

app/web/features/auth/signup/AccountForm.tsx
nabramow: Should be fixed now!

app/web/features/auth/signup/AccountForm.tsx
nabramow: lolz in my head I was looking for something like 1975 for 18 years ago and then realized 2006 is 18 years ago how horrifying...

nabramow: Looks like we're doing that on purpose and it's also on prod so I'll leave that for now.

Really convenient for the 18 year olds...


Screenshot 2024-12-14 at 6 38 12 PM

app/web/features/profile/view/NewHostRequest.tsx
nabramow: So according to the code this should also be happening on prod, but all of you have "Can't host" there lol so I can't test.

In any case I fixed it!

app/web/features/profile/view/NewHostRequest.tsx
nabramow: Fixed!

app/web/features/profile/edit/EditProfile.tsx
nabramow: Map should be fixed now.

The allows you to submit without display location is also on prod, so feel free to make a separate ticket for that if you want. Trying to keep this PR only to upgrade changes.

app/web/features/profile/edit/EditProfile.tsx
nabramow: This was an easy fix so I did it here.

app/web/components/Navigation/ReportButton.tsx
nabramow: Feel free to throw up a ticket, the bug report one should be pretty straight-forward!

app/web/features/donations/Donations.tsx
nabramow: Okay fixed!

app/web/features/auth/phone/ChangePhone.tsx
nabramow: I think fixed but I didn't donate so can't go all the way to the verification so please check me.

Translation fixed though.

app/web/components/Map.tsx
nabramow: So this one nearly killed me. I literally woke up this morning thinking about this FROM A DEAD SLEEP lol.

The culprit: In React 18, if you have Strict Mode enabled, the useEffect hook runs twice in development mode. This was causing this to return before the map was initialized so it was blank. So I am storing an `isMapInitialized` value in state now instead and referencing it and that seems to do the trick.

app/web/features/profile/view/leaveReference/formSteps/Appropriate.tsx
nabramow: Done

app/web/components/Map.tsx
nabramow: Should be fixed now.

app/web/components/Map.tsx
darrenvong: I think this type of `useEffect`s is exactly what the Strict Mode change is trying to catch - asymmetrical effects that need clean ups!

Would moving the separate cleanup below on line 126 into this effect instead negate the need to use the `isMapInitialized` trick? There's actually a warning in the official docs that [advise against doing this to prevent the effect from firing twice](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development) if possible so we're not hiding subtle bugs.

nabramow: @darrenvong Not the React docs calling me out...lol that they literally have a map example in the description. Thank you so much this really broke my brain. Last time I was a frontend engineer in my day job we were using React v17 hah.

I did almost this with a few adjustments. I needed to make the useEffect only load on first render via empty dependencies array otherwise it rerenders whenever I type in the location input. And needed to set `mapRef.current = undefined` in addition t removing the mapRef in the return.

I checked the MapSearch map and that's still working too so 🤞.

app/web/features/auth/phone/ChangePhone.tsx
```diff
@@ -168,25 +158,15 @@ export default function ChangePhone({
...
-                  }}
-                  international
-                  placeholder={t("auth:change_phone.phone_label")}
-                  id="phone"
-                />
-              )}
+              international
+              placeholder={t("auth:change_phone.phone_label")}
+              id="phone"
             />
```
nabramow: In case anyone asks me why I did this:

https://gitlab.com/catamphetamine/react-phone-number-input#react-hook-form

Then it handles forwarding the ref for react-hook-form without me writing tons of code to do it.

nabramow:
OMG I FINALLY GOT THE GREEN CHECK!

Okay @aapeliv @jesseallhands @bakeiro @darrenvong feel free to take a look and mark the comment as resolved if you're happy with it. I believe I addressed everything now (or told you to make a separate ticket if it's not related to this PR).

jesseallhands:
LGTM! Couldn't find anything broken in testing!

app/web/features/auth/phone/ChangePhone.tsx
aapeliv: You can use the stripe test cards on next, 4242424242424242 (and anything for the other fields) will succeed for instance. Just fyi for the future!

app/web/components/Map.tsx
darrenvong: Right, yeah there's some nasty stuff to handle if we were to follow the exhaustive deps rule which might be tricky to deal with without making some values more stable, and that'd probably complicate the code more. Without knowing too much about this whole thing works (the Map is the area of app I didn't really work with as closely 😅), this seems like a reasonable tradeoff until we refactor further I reckon 👍

darrenvong: Or if you somehow know which variables may make sense to still be reactive over from more recent working knowledge of this beast, I'd recommend following this part of the doc to pull out the non-reactive variables into an "effect event" - https://react.dev/learn/separating-events-from-effects#choosing-between-event-handlers-and-effects

Although this whole technique is still a bit clunky imo, since the `useEffectEvent` hook doesn't actually exist as an official API, so you'd have to add a [rough userland implementation of it](https://gist.github.com/AndrewIngram/2e5d530eb91b96c6a1b009b21c59db52), I'll leave it up to you on this one on whether you want to go that far to follow the official advice!

app/web/components/ContributorForm/ContributorForm.test.tsx
```diff
@@ -40,36 +40,46 @@ describe("contributor form", () => {
...
+
+    const user = userEvent.setup();
+
+    await user.type(screen.getByLabelText(IDEAS_LABEL), "I have great ideas");
+    await user.type(
       screen.getByLabelText(FEATURES_LABEL),
       "I want all the features"
     );
-    userEvent.click(screen.getByRole("radio", { name: "Yes" }));
-    // For some reason, checking boxes makes tests flakey when running jest
```
darrenvong, line 52: Has the upgrade somehow magically fixed the performance issue?

app/web/components/ImageInput/ImageInput.tsx
```diff
@@ -77,7 +77,8 @@ const useStyles = makeStyles((theme) => ({
 
 interface ImageInputProps {
   className?: string;
-  control: Control;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  control: Control<any>;
```
darrenvong, line 81: I wonder if making this extends from `FieldValues` and let it passed in as a generic param would work instead to avoid the `any`? Appreciate the type golf with react-hook-form can get hairy though...

app/web/features/communities/CommunityPage/JoinCommunityButton.tsx
```diff
@@ -59,17 +58,6 @@ export default function JoinCommunityButton({
         loading={isLoading}
         variant={community.member ? "outlined" : "contained"}
         onClick={() => (community.member ? leave.mutate() : join.mutate())}
-        sx={{
```
darrenvong, line 62: Have the styling for this somehow been baked in/implemented somewher else?

app/web/features/communities/discussions/DiscussionPage.test.tsx
```diff
@@ -237,22 +237,25 @@ describe("Discussion page", () => {
...
-        screen.getByRole("button", {
+      const user = userEvent.setup();
+
+      user.click(
+        await screen.findByRole("button", {
           name: t("communities:load_earlier_comments"),
         })
       );
-      await waitForElementToBeRemoved(screen.getByRole("progressbar"));
+
```
darrenvong, line 248: Move the comment where it was next to assertion to here so it's clear why we're asserting the function has been called 6 times?

app/web/features/communities/events/EventTimeChanger.tsx
darrenvong, line 1: Somehow this is not quite working as it is in prod - when the user updates the end time such that it's more than 1 hour from the start time, if they then update the start time, the end time is meant to shift by the same duration.

So if start time is 1pm, and I change the end time to 4pm (3 hours apart instead of defaulting to 1 hour apart which would be 2pm), if I then change the start time to 3pm, the end time should become 6pm instead of 4pm

darrenvong (commented):
Only found a few more things, but otherwise from a very brief click around (and a not so brief code "skim" that still end up taking me a while 😂) , it looks ok to me! So hopefully between us we caught most things 🤞

app/web/features/auth/phone/ChangePhone.tsx
nabramow: Ohhhh yes I forgot you can do that with Stripe, thanks!

app/web/components/Map.tsx
nabramow: So I actually think everything in that useEffect is for initial load, so it makes sense to be in a useEffect. The onClick, zoom, etc. etc. are in separate functions.

I'm also a bit overwhelmed by the map TBH. I think it could use a refactor to make it more readable, but since we're likely going to do that as part of a map redesign soon-ish it doesn't seem like a priority.

What makes it complicated IMO is that we're using this Map component multiple places, but we've put in all this conditional logic depending on where it's being rendered.

Probably this gets way cleaner if we opt to duplicate *some* logic and just break them out into separate map components by use case. Then we don't need different map wrapper components either I assume, which is also causing rerenders for example when typing in the location input even before we hit enter.

So for now just trying to get it to work in a decent, not-too-hacky way until we decide on the map design then I will go at it with a 🔪 if I have anything to say about it.

app/web/components/ContributorForm/ContributorForm.test.tsx
nabramow: Yes! But also I think probably using `waitFor` back then might have fixed it too. Since these return promises now it's easier to know it's done clicking before you check the assertion.

app/web/components/ImageInput/ImageInput.tsx
nabramow: Not gonna lie, I spent full DAYS on these types during the Thanksgiving break 😅.

I did try `FieldValues`. Every time I get one component's types working it breaks the parent components. Or I get both props passing TS checks and the useController types get mad.

I cannot figure it out and I spent so, so much time on it before I decided it wasn't worth that much effort.

Totally happy if you wanna push a commit to this PR if something obvious stands out to you though! Or can do after.

app/web/features/communities/CommunityPage/JoinCommunityButton.tsx
nabramow: I copy pasta'd button style when I did the MUI upgrade and Jesse pointed out this one looked weird. When I took that style out it looks better again. I think it's just a leftover mistake from when I did the MUI upgrade.

app/web/features/communities/discussions/DiscussionPage.test.tsx
nabramow: Oh yes good catch!

app/web/features/auth/signup/AccountForm.tsx
aapeliv: OK, Jesse made an issue here: #5375

app/web/features/communities/events/EventTimeChanger.tsx
nabramow: @darrenvong Okay I fixed this up! I put the old code back as I had the wrong idea. Then refactored a bit to take things out of the useEffects and refs as they are actually tied to onChange events so useEffect doesn't make sense to me.

Also put the old tests back, and everything is passing now. So I think it's working!

nabramow:
@aapeliv @darrenvong Okay I think I addressed everything now!

aapeliv (approved):
If you addressed my comments, I’m happy :)

we can do any residual bug squashing on develop!

PR #5329, opened 2024-12-13
Add Admin ListUserIds endpoint (#5174)

### LINKED ISSUE #5173: Admin API call: list users by messages/reqs sent in interval
This should take an interval (start time and end time), and produce a list of `(user_id, messages_sent, host_requests_sent)` ordered by messages sent descending. Messages sent is number of all messages sent (every host req includes one message and host requests sent is number of host reqs sent.

The admin console can then show most messages sent by users in last week or day or whatever.

### LINKED ISSUE #5174: Admin API call: list of new accounts
Given an interval (start time, end time), should return user_ids of users who signed up in that interval, ordered by time signed up, desc.

jasonwvh:
add Admin ListUserIds endpoint to return a set of userIds based on a `start_time` and `end_time`

closes #5174

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/backend/src/tests/test_admin.py
```diff
@@ -525,4 +525,24 @@ def test_DeleteEvent(db):
...
+    normal_user, normal_token = generate_user()
+
+    with real_admin_session(super_token) as api:
+        res = api.ListUserIds(
+            admin_pb2.ListUserIdsReq(
+                start_time=Timestamp_from_datetime(datetime(2000, 1, 1)), end_time=Timestamp_from_datetime(now())
+            )
+        )
+        assert len(res.user_ids) > 1
+        assert any(user == normal_user.id for user in res.user_ids)
```
aapeliv: Could you please assert that the returned list is exactly what you expect, I think in this case you would do

```py
assert res.user_ids == [super_user.id, normal_user.id]
```

It'll strengthen the test to make it catch more stuff.

app/backend/src/tests/test_admin.py
```diff
@@ -525,4 +525,24 @@ def test_DeleteEvent(db):
...
+
+    with real_admin_session(super_token) as api:
+        res = api.ListUserIds(
+            admin_pb2.ListUserIdsReq(start_time=Timestamp_from_datetime(now()), end_time=Timestamp_from_datetime(now()))
+        )
+        assert len(res.user_ids) < 1
```
aapeliv: Same here, just assert the returned value is what you expect (empty list?)

aapeliv (commented):
I think this is basically perfect! I added a tiny comment on the test.

Also the issue number is off by one, should be #5174!

aapeliv (approved):
LGTM! (Just wait for tests to pass before merging please.)

PR #5336, opened 2024-12-14
Make dialogs prettier

aapeliv:
* Fixes the dialog title font being tiny.
* Makes action buttons centered instead of spread apart

---

Before:

Screenshot from 2024-12-13 19-20-40

---

After:

Screenshot from 2024-12-13 19-20-16


nabramow:
I think actually we should just put the old font sizes back. I had removed them in the MUI PR, because develop looked bigger font than prod. But I actually think it was just a difference with the local version. Hold on, I'll see if I can do a quick PR just reverting that change, as it effects all the headings.

nabramow (approved):

PR #5338, opened 2024-12-14
MUI - Put old heading sizes back in theme & fix breakpoints

nabramow:
When I upgraded MUI to v5 in #5157, I adjusted the heading sizes in the theme. Once it merged to prod, it looked too small. For some reason there seems to be a difference in how things appear on locahost vs. prod. Could also be had screens at different zoom levels.

I'm just putting it back to the old values. We have comments in the original file with what the font sizes should be. I checked the computed font sizes in the Chrome browser, and they match now off this branch...so I am pretty sure this is right!

I'm also fixing a breakpoints I missed that were in props or using theme.breakpoints.values.md so the codemod didn't get them. MUI changed the meanings of the breakpoints to mean "bellow small" instead of "below and including small" so that needed to be fixed. This effects:
* Dialog
* Login
* Signup

In theme H1 font size:
Screenshot 2024-12-13 at 7 05 49 PM

H1 computed fontsize for medium and up screen:
Screenshot 2024-12-13 at 7 06 47 PM

In theme H2 font size:
Screenshot 2024-12-13 at 7 13 40 PM

On medium up screen H2:
Screenshot 2024-12-13 at 6 21 45 PM

H6 font size in theme:
Screenshot 2024-12-13 at 6 27 20 PM

H6 computed fontsize for medium up screen:
Screenshot 2024-12-13 at 6 28 40 PM

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/components/Dialog.tsx
+++ b/app/web/components/Dialog.tsx
@@ -54,7 +54,7 @@ export interface AccessibleDialogProps extends Omit<DialogProps, "className"> {
 }
 
 export function Dialog(props: AccessibleDialogProps) {
-  return <MuiDialog {...props} fullWidth maxWidth="md" scroll="body" />;
+  return <MuiDialog {...props} fullWidth maxWidth="sm" scroll="body" />;
 }
 
 export function DialogActions(props: Omit<DialogActionsProps, "className">) {
--- a/app/web/components/Dialog.tsx
+++ b/app/web/components/Dialog.tsx
@@ -54,7 +54,7 @@ export interface AccessibleDialogProps extends Omit<DialogProps, "className"> {
 }
 
 export function Dialog(props: AccessibleDialogProps) {
-  return <MuiDialog {...props} fullWidth maxWidth="sm" scroll="body" />;
+  return <MuiDialog {...props} fullWidth maxWidth="md" scroll="body" />;
 }
 
 export function DialogActions(props: Omit<DialogActionsProps, "className">) {
--- a/app/web/components/Dialog.tsx
+++ b/app/web/components/Dialog.tsx
@@ -19,7 +19,7 @@ import makeStyles from "utils/makeStyles";
 const useStyles = makeStyles((theme) => ({
   actions: {
     display: "flex",
-    justifyContent: "space-around",
+    justifyContent: "center",
     margin: 0,
     padding: theme.spacing(2),
     paddingTop: 0,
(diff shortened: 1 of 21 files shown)


app/web/components/Dialog.tsx
```diff
@@ -54,7 +54,7 @@ export interface AccessibleDialogProps extends Omit<DialogProps, "className"> {
 }
 
 export function Dialog(props: AccessibleDialogProps) {
-  return <MuiDialog {...props} fullWidth maxWidth="sm" scroll="body" />;
+  return <MuiDialog {...props} fullWidth maxWidth="md" scroll="body" />;
```
aapeliv: I think it looks better with `sm`

aapeliv:
The font changes are great and the login/signup width changes. But not sure why you changed a bunch of the other page widths? E.g. TOS page to me looks better narrower (maybe even narrower than it's on prod at the moment)?

app/web/components/Dialog.tsx
nabramow: Will change back. I thought the default changed with the MUI changes but maybe not for the maxWidth.

It's confusing because for anything `theme.breakpoints.down` we needed to change the value to keep it the same.

aapeliv: Ah I see

nabramow:
> The font changes are great and the login/signup width changes. But not sure why you changed a bunch of the other page widths? E.g. TOS page to me looks better narrower (maybe even narrower than it's on prod at the moment)?

How the breakpoints resolve for that value changed with the MUI upgrade, and I forgot to update these types of values. 

I changed anything with the `theme.breakpoints.values` as I believe it was affected.

If there are certain ones you like better as they are now though, let me know I will change that back. You want the TOS back the current on prod way?

aapeliv:
> If there are certain ones you like better as they are now though, let me know I will change that back. You want the TOS back the current on prod way?

I think all the ones on prod have been fine? I didn't track down all the ones that have changed here tho.

nabramow:
@aapeliv Here's the blurb from the v5 migration guide for reference:
https://mui.com/material-ui/migration/v5-component-changes/#%E2%9C%85-breakpoint-behavior

This definitely also affected Login/Signup as Darren pointed out in the original review, which is why I changed the theme headers originally as I didn't notice these other breakpoint values were the issue.

app/web/components/Dialog.tsx
nabramow: Okay dialog is changed back now. As were the others that weren't related to signup or login.

aapeliv (approved):
LGTM!

Issue #5050, opened 2024-10-26
Add character count to user edit profile text input - solve user not knowing why profile incomplete warning


aapeliv:
We now have a requirement of minimum 150 characters in "About Me" + a profile picture in order to do certain things. (See: https://help.couchers.org/hc/couchersorg-help-center/articles/1725919152-why-do-i-need-to-complete-my-profile-to-use-some-features)

However, we are receiving a bunch of bug reports of people not knowing what to fill in in their profile.

**We should add some kind of error message with a counter: "You need to write 136 more characters about yourself to be able to send requests, messages, or create events." (136 should of course change and count to 0, then this message should disappear.)**

bakeiro:
yes, this actually have happened to me 😅

bakeiro:
I was one of those users xd

PR #5367, opened 2024-12-19
Incomplete profile helper

aapeliv:
We get a bunch of bug reports from people not knowing what to fill in (#5050), this attempts to fix that:

Screenshot from 2024-12-19 10-21-58

That all being said: I'm pretty sure I've done it in *the wrong way*, and that there's some big performance bug here... so could some React pro please help me figure out what I've done wrong? (I think one way to fix this is to use a `<Controller/>` then do this with `control`/`render`/`onChange`..? But that seems rather complicated compared to this... Maybe there's a better way?)

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/Alert.tsx
+++ b/app/web/components/Alert.tsx
@@ -1,7 +1,7 @@
 import { Alert as MuiAlert, AlertProps as MuiAlertProps } from "@mui/material";
 import { grpcErrorStrings, ObscureGrpcErrorMessages } from "appConstants";
 import classNames from "classnames";
-import React from "react";
+import React, { ReactNode } from "react";
 import makeStyles from "utils/makeStyles";
 
 const useStyles = makeStyles((theme) => ({
@@ -12,7 +12,7 @@ const useStyles = makeStyles((theme) => ({
 
 interface AlertProps extends MuiAlertProps {
   severity: MuiAlertProps["severity"];
-  children: string;
+  children: string | ReactNode;
 }
 
 export default function Alert({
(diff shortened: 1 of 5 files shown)


app/web/components/Alert.tsx
```diff
@@ -12,7 +12,7 @@ const useStyles = makeStyles((theme) => ({
 
 interface AlertProps extends MuiAlertProps {
   severity: MuiAlertProps["severity"];
-  children: string;
+  children: string | ReactNode;
```
darrenvong, line 15: `ReactNode` includes `string` as well so you can replace it altogether!

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -101,6 +109,36 @@ export default function EditProfileForm() {
...
+                }),
+              }}
+            </strong>{" "}
+            in this section to complete your profile. You also need to upload a
+            profile picture.
+          </Trans>
+        </Alert>
+      );
+    }
+  };
```
darrenvong, line 140: The way this is written feels like it could be its own component to me, and I'd consider using [`useWatch`](https://www.react-hook-form.com/api/usewatch/) in that separate component instead as according to the docs, it helps to isolate re-renders only to this bit of the component instead of this entire `EditProfile` component every time the `aboutMe` field updates, which will help with performance if that's a concern

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -391,6 +429,7 @@ export default function EditProfileForm() {
               defaultValue={user.aboutMe || DEFAULT_ABOUT_ME_HEADINGS}
               control={control}
               className={classes.field}
+              otherChildren={charsRemainingAboutMe()}
```
darrenvong, line 432: I'd consider updating this to `children` instead since it's not being used yet, then once you break it up into a separate component, it can read more "naturally" like this:
```tsx
// Intentionally leaving out all of the props for brevity
<ProfileMarkdownInput>
  <IncompleteProfileAlert />
</ProfileMarkdownInput>
```

instead of having to do `otherChildren={...}` (children is a shortcut for allowing more "natural" composition of components for these particular use cases)

app/web/features/profile/edit/EditProfile.tsx
darrenvong: But otherwise, I think what you've done here is fine and no need for Controllers!

aapeliv:
Superseded by #5378

PR #5377, opened 2024-12-22
Firefox scrolling issues

### LINKED ISSUE #5257: Page scrolls halfway down on load
Video:

https://github.com/user-attachments/assets/7668a205-260b-49f3-891e-afcc2adc8d14

The culprit is this line:

https://github.com/Couchers-org/couchers/blob/1338c1c8df756237dc2eac18edf1ad09994feac2/app/web/pages/_app.tsx#L67

I don't entirely know what that code is doing, and maybe it will be sorted out with the Next.js upgrade. Maybe the `navigator.userAgent` should match something else?

### LINKED ISSUE #5361: upscrolling
Subject: upscrolling
Description:
Upscroll

Results:
Insteadof upscrolling one bit, it goes all the way up

Backend version: develop-1dbd5d77
Frontend version: develop-1dbd5d77
User Agent: Mozilla/5.0 (Android 12; Mobile; rv:133.0) Gecko/133.0 Firefox/133.0
Page: https://couchers.org/user/anaximanderjh
User: [@ezeq](https://couchers.org/user/ezeq) (30773)

nabramow:
So this seems to be only an Android AND Firefox problem. I can't replicate on Mac/iOS with Firefox, so stabbing in the dark here.

Trying to fix two issues here that I believe are related to the same code introduced in #5200:
* On Android phone with Firefox browser, swiping causes random scroll all the way to the top
* In Firefox desktop browser on non-Mac, page scrolls halfway down on initial page load.

Closes #5361
Closes #5257 

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/messages/groupchats/GroupChatView.tsx
+++ b/app/web/features/messages/groupchats/GroupChatView.tsx
@@ -18,6 +18,7 @@ import { RpcError } from "grpc-web";
 import { useTranslation } from "i18n";
 import { GLOBAL, MESSAGES } from "i18n/namespaces";
 import { GetGroupChatMessagesRes, GroupChat } from "proto/conversations_pb";
+import { useEffect } from "react";
 import {
   useInfiniteQuery,
   useMutation,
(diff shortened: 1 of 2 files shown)


nabramow:
@jesseallhands @aapeliv So this is a fun bug in that I can't replicate on iOS or Mac with Firefox and I only have Apple devices, so apologies in advance y'all are gonna have to be my manual testers. I believe both your bugs are related though!

I took an initial theoretical stab at this. Can you tell me if it's fixed? Also that messages scrolling isn't broken again!

jesseallhands (approved):
LGTM! I tested on Firefox mobile, which is the only place I had previously encountered the problem, and I wasn't able to replicate so it seems to be fixed! Thank you!

PR #5378, opened 2024-12-22
Add aboutMe profile character count

### LINKED ISSUE #5050: Add character count to user edit profile text input - solve user not knowing why profile incomplete warning

nabramow:
User's aren't realizing the reason they can't send messages, etc. is because they need a longer profile. This PR adds a warning text on the EditProfile page aboutMe section if the user is under the limit and same with the missing profile photo:

Screenshot 2024-12-22 at 7 42 54 PM

Screenshot 2024-12-22 at 8 08 19 PM

Screenshot 2024-12-22 at 8 09 46 PM

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/features/profile/edit/EditProfile.tsx
+++ b/app/web/features/profile/edit/EditProfile.tsx
@@ -89,7 +89,6 @@ export default function EditProfileForm() {
     formState: { errors, isDirty, isSubmitted },
   } = useForm<EditProfileFormValues>({
     defaultValues: {
-      avatarKey: user?.avatarUrl,
       location: {
         city: user?.city,
         lat: user?.lat,
@@ -101,9 +100,9 @@ export default function EditProfileForm() {
     shouldFocusError: true,
   });
 
-  const [avatarKeyField, aboutMeField] = useWatch({
+  const aboutMeField = useWatch({
     control,
-    name: ["avatarKey", "aboutMe"],
+    name: "aboutMe",
   });
 
   useUnsavedChangesWarning({
@@ -168,7 +167,7 @@ export default function EditProfileForm() {
   const handleSubmitButtonClick = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
 
-    if (aboutMeField?.length < ABOUT_ME_MIN_LENGTH || !avatarKeyField) {
+    if (aboutMeField?.length < ABOUT_ME_MIN_LENGTH || !user?.avatarUrl) {
       setShowIncompleteProfileDialog(true);
     } else {
       onSubmit();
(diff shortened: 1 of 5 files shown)


jesseallhands:
This looks to be working as expected and can be merged.

I have a feeling it will be overlooked, especially on mobile.  

Could we add a warning when the user clicks "save" if the character count is still below the limit?

It could say something like "Your "About Me" section is still too empty. Save anyway?" [Yes] [Cancel] and if they click cancel, it can bounce them up to the "About me" section so they can continue editing.

Obviously this would have to be a new issue. Thoughts?

jesseallhands (dismissed):

aapeliv (commented):
Tested and seems to be working well.

I suggest using the dialog from #5367 for the helper text. Also I think maybe instead of "Cancel" it's better to say "Continue editing"

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -134,13 +143,26 @@ export default function EditProfileForm() {
...
     },
     // All field validation errors should scroll to their respective field
     // Except the avatar, so this scrolls to top on avatar validation error
     (errors) =>
       errors.avatarKey && window.scroll({ top: 0, behavior: "smooth" })
   );
 
+  const handleSubmitButtonClick = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (aboutMeField?.length < 150) {
```
aapeliv: Maybe we can move 150 into a constant. (It's something we've changed before too.)

app/web/features/profile/locales/en.json
```diff
@@ -34,7 +34,15 @@
     "gender_verification_verified": "This user has verified their gender",
     "gender_verification_mismatch": "User provided gender does not match verified gender"
   },
+  "helper_text": {
+    "characters_remaining": "You need to write {{count}} more characters about yourself to be able to send requests, messages, or create events."
+  },
   "section_tabs_a11y_label": "tabs for user's details",
+  "incomplete_dialog": {
+    "title": "Profile Incomplete",
+    "message": "Your \"Who I Am\" section is still too empty. You won't be able to send requests, messages or create events until this section is at least 150 characters long.",
```
aapeliv: "still too short" maybe?

nabramow:
@aapeliv @jesseallhands Okay lots of big changes here. I added the alert for the profile photo as well, added a dialog when they click submit that conditionally lists what's missing, stole some of Aapeli's design from his other PR, and updated and added tests.

One thing I wasn't able to test super well - I already have a profile photo and there is no delete option. I tested the UI for that but didn't test it with a real missing photo.

Anyway it's ready again!

aapeliv (approved):
Tested and code looks good!

aapeliv:
(Also I did test without a picture.)

PR #5325, opened 2024-12-12
Streamlined Direct Message Flow: Pop-up Dialog for Messaging

mrahmed2026:
## Issue Reference:
Closes #4301

## Description
This PR implements a streamlined flow for direct messaging from user profiles. Instead of redirecting to the messages page and requiring the user to refresh to see new messages, a pop-up dialog is now presented directly on the user profile page. This dialog allows users to type and send messages seamlessly.

**Key Changes:**
- Added a dialog box component in the user profile for entering messages.
- Integrated logic to check for existing chats and create new ones if necessary.
- Enhanced error handling for thread creation and message sending (i.e cannot send empty / blank message).

**Backend checklist**
- [x] Run the backend locally and it works

**Web frontend checklist**
- [x] Formatted my code with `yarn format`
- [x] There are no warnings from `yarn lint --fix`
- [x] There are no console warnings when running the app
- [x] Clicked around my changes running locally and it works
- [x] Checked Desktop, Mobile and Tablet screen sizes

PR #5392, opened 2024-12-24
Streamlined message flow

### LINKED ISSUE #4301: Streamlined Direct Message flow from user page
Currently, to send someone a DM, you go to their profile, click "Message", and it opens the messagse page with a user selector. This is cumbersome and non-ideal.

We should implement a backend feature called `SendDirectMessage` in `conversations.proto` that both creates a group chat (like `CreateGroupChat`) if there is none, and sends a message (like `SendMessage`).

This can then be used for a better frontend implementation.

---

Instead of popping up the messages page and making a new conversation, it should pop up a window to enter a message and the frontend should automatically send to the right user.

We might want to implement a streamlined backend call for this if it doesn't exist yet!

Subject: When you make a DM it doesn’t show up
Description:
Profile > Message > It goes ot messages and pops up the new message box with the user’s name filled in > confirm the dialog

Results:
The dm does not show up. I need to somehow refresh before it shows up.

Backend version: develop-60580153
Frontend version: develop-60580153
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 17_4_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4.1 Mobile/15E148 Safari/604.1
Page: https://couchers.org/messages/chats?to=msieglo
User: aapeli (1)

aapeliv:
Continues from the abandoned  #5325. Closes #4301.

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/profile/actions/MessageUserButton.tsx
+++ b/app/web/features/profile/actions/MessageUserButton.tsx
@@ -1,14 +1,22 @@
+import {
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  TextField,
+} from "@mui/material";
 import Button from "components/Button";
-import ProfileIncompleteDialog from "components/ProfileIncompleteDialog/ProfileIncompleteDialog";
-import useAccountInfo from "features/auth/useAccountInfo";
 import { useTranslation } from "i18n";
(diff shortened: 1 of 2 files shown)


aapeliv:
Actually this is pretty unusable.

PR #5397, opened 2024-12-26
Refactor Datepicker to allow null, Signup Form birthdate and EventTimeChanger changes

### LINKED ISSUE #4102: Update birthdate validation to prevent users from selecting impossibly old birthdate
Users shouldn't be able to say that they are 823 years old like this user:
image

We should update the birthdate validation so we don't accept ages above 99 years old (the threshold is somewhat arbitrary but the issue is quite common. There are reasons why we require users to provide their real age).

### LINKED ISSUE #5375: During signup, birthdate defaults to 18 years old based upon current date
### Problem
Currently during signup the system default's the user's age to 18 years old as of today's date. There have been numerous reports by users that their age is wrong and that they aren't 18 but X age instead. We likely have (and will continue to have) a significant number of new users with the incorrect birthdate.

### Solution
Default date should be empty and force user to use date picker to enter their actual birthdate.

nabramow:
There are several tickets to improve the birthdate validation in the SignupForm to now allow impossibly old ages and to have it be null by default so the user needs to choose a birthdate.

In order to fulfill the latter I needed to refactor the Datepicker to allow null values. IMO this is how it should be anyway, since it's a reusable component it should be more flexible, and the validation gets handled in the parent components so it can be catered to each use case.

Those changes required me to also make changes to EventTimeChanger, so this effects the EditEvent, CreateEvent, SignupForm, and HostRequest form datepickers.

I made some other improvements:

- Show errors on input blur so rather than waiting for onClick for feedback
- More specific errors for too old, too young, blank date, etc.

**Web frontend checklist**
- [ x ] All tests pass


app/web/components/Datepicker.tsx
```diff
@@ -63,10 +63,8 @@ const Datepicker = ({
           minDate={minDate}
           maxDate={maxDate}
           onChange={(date) => {
-            if (date?.isValid()) {
-              field.onChange(date);
-              onPostChange?.(date);
-            }
+            field.onChange(date);
+            onPostChange?.(date);
           }}
```
nabramow, line 68: This is the major change here, this allows a null value. We need that to set the SignupForm birthdate to null as a default. 

This also gives us more flexibility to customize validation more in parent components, i.e. SignupForm, Event forms, etc.

app/web/components/Datepicker.tsx
```diff
@@ -91,7 +89,9 @@ const Datepicker = ({
               fullWidth
               id={id}
               error={error}
-              helperText={helperText}
+              helperText={
+                <span data-testid={`${name}-helper-text`}>{helperText}</span>
+              }
```
nabramow, line 94: So we can use `findByTestId` rather than `document.getElementById` in tests to check the rendered helperText which was a bit fussy that way it was before.

app/web/components/TextField.tsx
```diff
@@ -30,20 +30,29 @@ type AccessibleTextFieldProps = Omit<TextFieldProps, "variant"> & {
...
-});
+    return (
+      <MuiTextField
+        {...otherProps}
+        inputRef={ref}
+        name={name}
+        variant={variant}
+        helperText={
+          <span data-testid={`${name}-helper-text`}>{helperText}</span>
+        }
```
nabramow, line 48: Same here, lets us write better tests without referencing the `document`.

app/web/features/auth/locales/en.json
```diff
@@ -77,7 +77,8 @@
     "birthday": {
       "field_label": "Birthday",
       "required_error": "Enter your birthdate",
-      "validation_error": "Must be a valid date in the past."
+      "validation_error": "Must be a valid date in the past.",
+      "too_young_error": "You must be at least 18 years old to use Couchers"
```
nabramow: Confirmed with @jesseallhands users must be at least 18 so made a specific error.

app/web/features/auth/signup/AccountForm.tsx
```diff
@@ -180,6 +188,7 @@ export default function AccountForm() {
           }}
           helperText={errors?.username?.message ?? " "}
           error={!!errors?.username?.message}
+          autoComplete="username"
```
nabramow, line 191: We were getting console errors about this and it was annoying me so I fixed it.

app/web/features/auth/signup/AccountForm.tsx
```diff
@@ -206,19 +216,35 @@ export default function AccountForm() {
...
+              const birthDate = dayjs(stringBirthDate);
+              const age = Math.abs(dayjs().diff(birthDate, "year"));
+
+              if (age < 18) {
+                return t("auth:account_form.birthday.too_young_error");
+              }
+
+              if (age > 120) {
+                return t("auth:account_form.birthday.validation_error");
+              }
```
nabramow: The oldest know human is 122 so 🤷‍♀️. Let me know if I should make it lower.

aapeliv: Yeah I think the bound is fine, but maybe the error should say something else, we can just say "Please pick a real birthdate".

app/web/features/auth/signup/AccountForm.tsx
```diff
@@ -206,19 +216,35 @@ export default function AccountForm() {
...
+          helperText={errors?.birthdate?.message}
           id="birthdate"
           rules={{
             required: t("auth:account_form.birthday.required_error"),
-            validate: (stringDate: string) =>
-              validatePastDate(stringDate) ||
-              t("auth:account_form.birthday.validation_error"),
+            validate: (stringBirthDate: string) => {
+              const birthDate = dayjs(stringBirthDate);
+              const age = Math.abs(dayjs().diff(birthDate, "year"));
```
aapeliv: Could you just add a comment here that `// dayjs does the difference correctly by counting months`, or something please. I was afraid at first they'd just compute second diff and convert that to years, and their documentation is unclear; but I checked source code and they do it right! (Also tested to work...)

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -73,36 +73,54 @@ export default function EventTimeChanger({
   const handleStartTimeChange = (e: {
```
aapeliv, line 55: I've found two minor things while testing this out:

1) I think on `develop`, when you select a start date, the date picker doesn't allow you to pick an end date that's older, like the `minDate` is set to `startDate` or similar. I actually think I prefer how it works on this branch; BUT
2) When I change the start date/time, the end date/time *always* changes to 1 hour after start time. I think this is quite frustrating and people might not even notice, so I think the end date/time should only change in case you select a start datetime that's after the selected end datetime.

nabramow: @aapeliv I'm actually thinking maybe I should set these null by default too! 

But it's not always one hour, just that by default. Try this:

1. startTime 9am by default (endTime 10am)
2. Change endTime to 1pm
3. Change startTime to 10am, endTime will change to 2pm (keeping the same difference i.e. "dateDelta").

Do you think I should just remove all the "helping" with the dates and just keep the validation if it's after the startTime?

nabramow: @aapeliv Okay I pushed some changes like this. Let me know if you're happy with it and I'll update the tests.

@jesseallhands tagging you here as an FYI since this is a design/functionality change, but I think it kinda goes along with the birthdate change as well where we do null by default so they have to consciously pick. That prevents accidental wrong dates and times.

darrenvong: > Do you think I should just remove all the "helping" with the dates

Since I implemented it and yet can't quite remember why we did it that way... I wonder if it's because Couchsurfing works that way? 🤔 Or are there well known event scheduling tools that give users this level of "help"? Google Calendar perhaps?

nabramow: @darrenvong I wish I knew haha I'm sure it was cool at the time!

What we found is people were just going with the defaults and then getting confused why it had those dates, especially the birthdate hah so trying to force them to think about it.

nabramow:
@jesseallhands @aapeliv let me know if you're happy with both the birthdate and setting null as default for the event creation. If it seems okay I will go in and fix the failing tests (didn't wanna spend time yet in case y'all don't like it).

app/web/features/communities/events/EventTimeChanger.tsx
aapeliv: I actually like the idea of removing all the "smartness". Just let start and end date/times to be anything in the future, and do errors underneath them if they are incorrect.

aapeliv:
> @jesseallhands @aapeliv let me know if you're happy with both the birthdate and setting null as default for the event creation. If it seems okay I will go in and fix the failing tests (didn't wanna spend time yet in case y'all don't like it).

Yes I think this PR is looking good!

jesseallhands:
> @jesseallhands @aapeliv let me know if you're happy with both the birthdate and setting null as default for the event creation. If it seems okay I will go in and fix the failing tests (didn't wanna spend time yet in case y'all don't like it).

Yes! Sounds great to have it start at null!

app/web/features/communities/events/CreateEventPage.test.tsx
```diff
@@ -71,6 +71,8 @@ describe("Create event page", () => {
       expect(titleInput).toHaveValue("Test event");
     });
 
+    // @TODO(NA): Need to add start and end dates now but it's triple tested in EventTimeChanger and I can't seem to type in them from here.
+
```
nabramow: @darrenvong Cry for help lol. Now that EventTimeChanger doesn't have default dates, I can't submit CreateEventPage. Even though I see the `startDate` field in `screen.debug`, it seems to be marked read-only (I assume because it's nested 2 levels deep: CreateEventPage => EventForm => EventTimeChanger)?

I *think* the answer is to just test the interaction that's actually happening in this component and strip out the submit functionality but I'm not sure?

Does anything jump out at you as a proper way to handle this? I'm stuck with this one.

With `EditEventForm` I could just pass default dates but obviously with a new event it's blank.

nabramow: omg nevermind it was that damn mui-x-datepickers mock FML

nabramow:
@aapeliv okay this is finally ready and passing tests!

aapeliv (commented):
If I create a new event, it all is working well and as expected.

However, when I now edit that event, I can change the start time to be after the end time and the form will be POSTed (after which I get a backend error):

Screenshot from 2025-01-06 17-00-55

I suppose this means it's not catching it correctly in the frontend?

nabramow:
> If I create a new event, it all is working well and as expected.
> 
> However, when I now edit that event, I can change the start time to be after the end time and the form will be POSTed (after which I get a backend error):
> 
> I suppose this means it's not catching it correctly in the frontend?

Hey @aapeliv when I do this locally on the branch, it works. I even wrote a test and pushed it that's passing. However, when I check Vercel, it doesn't work. We had a similar issue in Gracee's clear filters branch. Could something be wrong with vercel?

Here's what I see locally:

https://github.com/user-attachments/assets/339dadac-7374-469e-ac78-f769b7e11f27

jesseallhands:
Specifically for the birthday datepicker during signup, I saw we are using a combination of the calendar-based datepicker and the manual date entry to give the user the option of whichever to use. Is it possible to make the manual-entry option more clear by having a box with the dd/mm/yyyy helper text?

nabramow:
> Specifically for the birthday datepicker during signup, I saw we are using a combination of the calendar-based datepicker and the manual date entry to give the user the option of whichever to use. Is it possible to make the manual-entry option more clear by having a box with the dd/mm/yyyy helper text?

Let's stick to the scope of this ticket ;-). Nothing design-wise has changed for the manual date/calendar combo as far as I know. This ticket is about setting null as the default date.

jesseallhands (approved):
LGTM

PR #5401, opened 2024-12-27
Upgrade jest and related

### LINKED ISSUE #5317: Upgrade jest and related plugins
We're a couple years behind in the jest package, let's upgrade this.

Migration guide for v28 to v29:
https://jestjs.io/docs/upgrading-to-jest29

Also upgrade `@types/jest` and `jest-junit` with it.

nabramow:
While upgrading Next I was having some issues with tests, and some next v13+ features are only available in later versions of jest. Jest was quite out of date so updating this now so I can be sure when upgrading next the versions on up to date.

Closes #5317 

**Web frontend checklist**
- [ x ] There are no warnings from `yarn lint:fix`
- [ x ] There are no console warnings when running the app (No new ones)
- [ x ] All tests pass


app/web/__mocks__/nextFontMock.js
```diff
@@ -0,0 +1,12 @@
...
+  {
+    get: function getter() {
+      return () => ({
+        className: "className",
+        variable: "variable",
+        style: { fontFamily: "fontFamily" },
+      });
+    },
+  }
+);
```
nabramow, line 12: Next v13 has some new font features in next/font so just adding this now.

app/web/jest.config.ts
```diff
@@ -0,0 +1,61 @@
+import nextJest from "next/jest";
+import type { Config } from "jest";
+
+// Providing the path to your Next.js app which will enable loading next.config.js and .env files
+const createJestConfig = nextJest({ dir: "./" });
+
+const customJestConfig: Config = {
+  verbose: true, // Shows detailed test results
```
nabramow, line 8: It was driving me nuts I couldn't see which tests failed when a lot ran at once.

app/web/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
nabramow, line 3: We in the future now

aapeliv (approved):
Looks good to me, but I'm not a jest expert!

app/web/__mocks__/nextFontMock.js
darrenvong: Out of interest, are there any reason you're using a Proxy over an object with an getter on it?

app/web/package.json
```diff
@@ -100,15 +100,16 @@
...
+    "jest": "^29.7.0",
+    "jest-environment-jsdom": "^29.7.0",
+    "jest-junit": "^16.0.0",
     "msw": "^1.2.1",
     "next-router-mock": "^0.7.4",
     "prettier": "^2.5.1",
     "require-from-string": "^2.0.2",
     "sentry-testkit": "^3.3.4",
     "timezone-mock": "^1.3.2",
+    "ts-node": "^10.9.2",
```
darrenvong, line 112: What's `ts-node` being used for?

app/web/tsconfig.json
darrenvong: I wonder if we can be even bolder and have the target at something like `es2020`? Since `es6` is still pretty old (equivalent to `es2015` I believe?) and since the death of IE, pretty much most consumers are on newish version of Chromium via Edge too, which would support most recent-ish features of JS. I believe the case would be similar for Safari users on MacOS/iOS

PR #5402, opened 2024-12-28
Upgrade Next.js to v13

### LINKED ISSUE #5154: Upgrade next to v13
We're quite behind, let's get this updated.

Migration guide to v13: https://nextjs.org/docs/pages/building-your-application/upgrading/version-13

Also keep in mind, it looks like there's a fix to this issue with newer versions:

app/web/utils/useStablePush.ts

```javascript

// from https://github.com/vercel/next.js/issues/18127#issuecomment-950907739
export default function useStablePush(): NextRouter["push"] {
  const router = useRouter();
  const routerRef = useRef(router);

  routerRef.current = router;

  const [{ push }] = useState<Pick<NextRouter, "push">>({
    push: (path) => routerRef.current.push(path),
  });

  return push;
}
```

See if you can remove this hack-around after updating!

nabramow:
Closes #5154 

Next v13 upgrade. I will do the v14 upgrade and conversion to the appRouter in a separate PR.

Migration Guide:
https://nextjs.org/docs/pages/building-your-application/upgrading/version-13

I needed to upgrade these packages as well for this:

- i18next
- next-i18next
- react-i18next
- @sentry/nextjs
- @sentry/react
- sentry-testkit

This means you'll want to pay attention to translations and sentry while testing.

**Web frontend checklist**
- [ x ] All tests pass


app/web/test/setupTests.ts
```diff
@@ -45,9 +45,8 @@ global.crypto = {
   },
 };
 
-//sentry testing was causing OOM for some reason
-//const { testkit, sentryTransport } = sentryTestkit();
-//global.testKit = testkit;
+const { testkit } = sentryTestkit();
+global.testKit = testkit;
```
nabramow, line 49: Their next upgrade was focused mostly on performance and reducing bundle size so my guess was the OOM was happening due to that.

app/web/test/utils.tsx
```diff
@@ -1,8 +1,5 @@
 import { fireEvent, screen } from "@testing-library/react";
 import mediaQuery from "css-mediaquery";
-import { TFunction } from "i18n";
-import { StringMap, TOptions } from "i18next";
-import i18n from "test/i18n";
```
nabramow, line 5: They stopped exposing `StringMap` purposely (says in their docs). I tried to re-write this function but it was getting pretty gnarly, which is why I decided to just use the test instance of i18n in the tests and use the t function from there rather than mocking a separate one.

jesseallhands (commented):
I couldn't find any issues during testing. The only issue I encountered I think is unrelated to this PR:

Sending a request with a duration of over 1 year clears both dates AND clears the request message you typed (the entire text is deleted). This would really suck if you typed a long request:
image

nabramow:
> I couldn't find any issues during testing. The only issue I encountered I think is unrelated to this PR:
> 
> Sending a request with a duration of over 1 year clears both dates AND clears the request message you typed (the entire text is deleted). This would really suck if you typed a long request: image

Oh interesting but yeah that's unrelated let me throw that in another ticket here: https://github.com/Couchers-org/couchers/issues/5432

aapeliv (dismissed):
Looks good to me codewise! Is the empty `<a>` thing something you just picked up here or is it raising errors now in the new version of Next? Either way great you cleaned it up!

nabramow:
> Looks good to me codewise! Is the empty <a> thing something you just picked up here or is it raising errors now in the new version of Next? Either way great you cleaned it up!

Ah it was part of the upgrade to remove the `<a>` tags, see here (second to last point):
https://nextjs.org/docs/pages/building-your-application/upgrading/version-13#v13-summary

I imagine something changed under the hood there so that we no longer need it.

nabramow:
FYI Merging as aapeli approved but it went away when I fixed the merge conflicts.

darrenvong:
> v14 upgrade and conversion to the appRouter in a separate PR

To my best knowledge, I don't think they have plans to deprecate/remove support for the pages router, so an entire app conversion might be unnecessary? I'd probably start trying out the app routers for new routes only, although that probably need an entire app architecture re-think on how to use/not use React Query together with the app router. As things are, it feels like they conflict with each other 😅 

Also soz for being a bit useless at reviewing PRs lately, but it seems like most of the changes are around link updates and tweaking the translation to work? Great job for powering through it! 🚀

nabramow:
> > v14 upgrade and conversion to the appRouter in a separate PR
> 
> To my best knowledge, I don't think they have plans to deprecate/remove support for the pages router, so an entire app conversion might be unnecessary? I'd probably start trying out the app routers for new routes only, although that probably need an entire app architecture re-think on how to use/not use React Query together with the app router. As things are, it feels like they conflict with each other 😅
> 
> Also soz for being a bit useless at reviewing PRs lately, but it seems like most of the changes are around link updates and tweaking the translation to work? Great job for powering through it! 🚀

No worries, I really appreciate you popping in at key moments! This was a good call out. What I liked about app router was it meant we could get rid of the next-18n-next package and there were some hackarounds we have that we'd be able to remove with it. But after thinking about it, there's so much important feature work to be done (like the map re-design that's blocking tons of stuff), so might be better to do less for now and return to it later if I want.

The actual conversion to the app router was pretty straight-forward tbh! It was the i18 stuff that got really complicated, and made it hard to do incrementally.

PR #5406, opened 2024-12-29
Remove some links to the forum

aapeliv:
As discussed.


jesseallhands (approved):
LGTM, though we need the issues related to fixing community discussions to be fixed asap since that is basically the replacement for removing references to the forum. 

The three main outstanding issues for community discussions are notifications, reporting flag on all discussions & (sub-)comments, and the API for removing comments & discussions.

Issue #5341, opened 2024-12-14
EPIC - New Map


nabramow:
This is the main feature ticket for the planned map improvements. See subtasks below. The map has been completed re-written as it was hard to maintain the previous architecture.

The new map should fix usability issues, add filters, and make it easier to search and find people, as well as offer an improved experience on mobile.

PR #5413, opened 2024-12-30
frontend - add clear filter button #5275

### LINKED ISSUE #5078: clear filter of the map button

### LINKED ISSUE #5275: Add "clear filters" button to map and change "search this area" to "search here"
### Problem
Right now it is very difficult for users to "reset" all search filters so that all pins are shown on the map again after they have performed a search with any filter. A common problematic situation is this:

1. User performs search on map
2. Limited number of users are shown on the screen due to the search
3. User wants to expand their search beyond the limits of their initial search (eg. they want to see everyone in the area now).
4. The user zooms out and very few pins are shown, leading them to think there are very few users on Couchers

### Solution
The new buttons should look like this on mobile (update- change "show all" to "clear filters"):
image
- [x] If the user has applied a filter, display a "clear filters" button which reverts all search parameters so that all pins are shown on the map. If no filters are applied, show disabled button. 
- [x] To ensure there is enough space on mobile, the button "Search in this area" should be changed to "Search here" 
- [x] Remove the spinner to give more room. The spinner is being added as an overlay to the map in #5026 
- [x] Update border-radius and margin of buttons (so there is enough space in between the buttons)
- [x] Test

graceegal:
closes #5275
part of #5341

**Web frontend checklist**
- [x] All tests pass


nabramow:
Screenshot 2024-12-29 at 8 15 38 PM

Needs some margin around the right side of the button here. I'd make it the same as the distance between search in this area and this.

nabramow:
When I click clear filters, it also clears my selected location. IMO it shouldn't clear location but everything else. But deferring to @jesseallhands what do you think?

app/web/features/search/MapWrapper.tsx
```diff
@@ -160,6 +181,29 @@ export default function MapWrapper({
...
+    setAreFiltersApplied(filtersActive);
+  }, [
+    queryName,
+    lastActiveFilter,
+    hostingStatusFilter,
+    numberOfGuestFilter,
+    completeProfileFilter,
+    locationResult,
+  ]);
+
```
nabramow: Okay first of all, this was a great first attempt, and it pretty much works! Well done taking this on as one of your first tickets as honestly the map is confusing even for me!

A few notes on things I'd change due to the approach:

First a caveat - this isn't how I'd ideally do this in i.e. a day job, normally I'd just send an api call to the backend to clear everything and that would return an updated response, which would update the UI.

BUT we cache the result and filter it on the frontend, which is why that doesn't work here.

Given we will refactor the map soon anyway, this is how I'd go about it as a stop gap:

We already have a very similar useEffect to the one you created for `wasSearchPerformed` in the parent component SearchPage so IMO we want to avoid another one.

Also I'd generally try to keep this functionality in the same place the values are set in useState, which is in SearchPage.

Since you are correctly clearing the values there, we just need a variable to pass down to say "hey the filters changed refilter the user pin results in this component. Your function is already updating the state, which is re-running the api call and regenerating "result", which gets passed down to `MapWrapper`.

So IMO we just need to tell MapWrapper the search was cleared as the actual functionality is already there.

I'd simplify it a bit like so:

- In SearchPage create a useState variable `areFiltersCleared, setAreFiltersCleared` and set it to true as a default value
- When you run the handleClearFilters function in SearchPage, set this to true
- In SearchPage, in the useEffect that checks if `if (!wasSearchPerformed)` create an else to the inner if clause that checks if everything is 0 or undefined, and `setAreFiltersCleared` false there
- In MapWrapper, in the useEffect that re-renders users list on map, change `wasSearchedPerformed` to `{wasSearchPerformed || areFilltersCleared)`

I thiiiiink that should do it but let me know if not!

app/web/features/search/MapWrapper.tsx
```diff
@@ -267,6 +311,13 @@ export default function MapWrapper({
     }
   };
 
+  /**
+   * Clicks on 'clear filters' button
+   */
+  const handleClearFiltersOnClick = () => {
+    handleClearFilters();
```
nabramow: Just as a standard practice, I'd change the prop name from `handleClearFilters to `onClearFiltersClick`.

Generally the handle functions update the state in that same component. If passing down to a child component, name it the action it does instead.

bakeiro:
> When I click clear filters, it also clears my selected location. IMO it shouldn't clear location but everything else. But deferring to @jesseallhands what do you think?

The thing is, this is not "clear filters" (AFAIK) this is show all the users, and to do that we need to delete everything (including the location) because we can't apply filters/search a location when we display all the users on the map...

jesseallhands:
> When I click clear filters, it also clears my selected location. IMO it shouldn't clear location but everything else. But deferring to @jesseallhands what do you think?

It should clear location as well, but not reset the map position. 

If we don't clear location we still have the problem that users cannot see all of the pins on the map which is one of the main problems we are trying to resolve with this button.

graceegal:
> Screenshot 2024-12-29 at 8 15 38 PM
> 
> Needs some margin around the right side of the button here. I'd make it the same as the distance between search in this area and this.

@nabramow hmmm that's weird the border radius is rounded for you. On my vercel preview and on my localhost mine looks like this (with the filter buttons intentionally squashed together but no border radius where they're touching) 

UPDATE: I re-ran all of my environments locally and now mine is showing like your screenshot... vercel preview still shows as intended though (weird). I will defer to initial design of equal border radius and margin since it is the most simple fix, unless you say otherwise. 

Screenshot 2024-12-30 at 13 27 14
Screenshot 2024-12-30 at 13 27 21

app/web/features/search/MapWrapper.tsx
graceegal: Updated! That makes sense thanks for the info.

app/web/features/search/MapWrapper.tsx
graceegal: Awesome, thank you so much for the detailed feedback and explanation Nicole! Really appreciate it! 

I've made the changes you outlined, however I just want to call out that updating the useEffect that checks if `if (!wasSearchPerformed)` in SearchPage differed a bit from what you proposed... this was to ensure that the areFiltersCleared state was properly updated all the time, not just upon the first search. LMK if this makes more sense now.

graceegal:
> When I click clear filters, it also clears my selected location. IMO it shouldn't clear location but everything else. But deferring to @jesseallhands what do you think? 

@nabramow @jesseallhands @bakeiro 
updated so that it does reset any location that was set, but does not reset the map position (the current bbox stays the same). LMK if this is the correct functionality yall had in mind!

app/web/features/search/MapWrapper.tsx
```diff
@@ -267,6 +283,24 @@ export default function MapWrapper({
...
+        currentBbox[0][0],
+        currentBbox[0][1],
+        currentBbox[1][0],
+        currentBbox[1][1],
+      ];
+      handleClearFilters(bbox);
+    } else {
+      handleClearFilters();
+    }
+  };
```
nabramow: Ah sorry it's the other way. Name the original function "handleClearFiltersClick", but when you pass it as a prop, name it 'onClickFiltersClick'.

That's because the function is "handling" it, but when you pass it you want to see the action it's doing.

app/web/features/search/MapWrapper.tsx
```diff
@@ -267,6 +283,24 @@ export default function MapWrapper({
     }
   };
 
+  /**
+   * Clicks on 'clear filters' button
+   */
+  const onClearFiltersClick = () => {
```
nabramow:
```suggestion
  const handleClearFiltersClick = () => {
```

app/web/features/search/MapWrapper.tsx
```diff
@@ -267,6 +283,24 @@ export default function MapWrapper({
...
+    if (currentBbox) {
+      const bbox: [number, number, number, number] = [
+        currentBbox[0][0],
+        currentBbox[0][1],
+        currentBbox[1][0],
+        currentBbox[1][1],
+      ];
+      handleClearFilters(bbox);
+    } else {
+      handleClearFilters();
```
nabramow:
```suggestion
      onClearFiltersClick(bbox);
    } else {
      onClearFiltersClick();
```

app/web/features/search/MapWrapper.tsx
```diff
@@ -103,13 +110,17 @@ export default function MapWrapper({
   setIsFiltersOpen,
   wasSearchPerformed,
   setWasSearchPerformed,
+  areFiltersCleared,
+  handleClearFilters,
```
nabramow:
```suggestion
  onClearFiltersClick,
```

app/web/features/search/SearchPage.tsx
```diff
@@ -247,6 +271,8 @@ export default function SearchPage({
             isLoading={isLoading || isFetching}
             setWasSearchPerformed={setWasSearchPerformed}
             wasSearchPerformed={wasSearchPerformed}
+            areFiltersCleared={areFiltersCleared}
+            handleClearFilters={handleClearFilters}
```
nabramow:
```suggestion
            onClearFiltersClick={handleClearFilters}
```

jesseallhands:
> updated so that it does reset any location that was set, but does not reset the map position (the current bbox stays the same). LMK if this is the correct functionality yall had in mind!

Yes, sounds correct!

I tested it on mobile but it doesn't appear to be working. See video below where you can see the "clear filters" button never is clickable despite a location filter (Berlin) being applied. Also, when I click "search here" the button also is not clickable despite the "search here" filter parameter being applied:

https://github.com/user-attachments/assets/7ab64b04-f7b6-4bd7-8b84-0364c842b6cd

app/web/features/search/MapWrapper.tsx
graceegal: aha! Okay just updated all of that. Thank you :)

graceegal:
@nabramow @jesseallhands Clear filters button functionality should now match what we discussed on the meeting today. LMK if not!

A few other things I noticed while working on this and wanted to address: 
1) SearchResults List Mobile View: On mobile only, SearchResultsList is collapsed down and hidden when no filters are applied (this occurs in 2 situations - when they open the map from dashboard without an initial location search query AND every time that clear filters button is clicked). Is that okay behavior or would you prefer it be shown even when no filters are applied? 

2) LocationAutocomplete Sync Issue: When clearing filters and when a location is searched, there's an inconsistency between two instances of LocationAutocomplete (one in SearchBox and one in FilterDialog). The location input might show different values in each place - for example, SearchBox might still display "London" from an initial Dashboard search while FilterDialog shows empty or a different location. Should this synchronization issue be addressed in this PR or as a separate ticket?

app/web/features/search/SearchPage.tsx
```diff
@@ -105,7 +107,9 @@ export default function SearchPage({
   >();
 
   const [isFiltersOpen, setIsFiltersOpen] = useState(false);
-  const [areFiltersCleared, setAreFiltersCleared] = useState(true);
+  const [areFiltersCleared, setAreFiltersCleared] = useState(
+    locationResult.name === ""
```
nabramow: What happens if locationResult is not finished loading? It's risky to make one state variable dependent on another in the same component. Curious why not just set it to true by default?

app/web/features/search/SearchPage.tsx
```diff
@@ -160,9 +162,9 @@ export default function SearchPage({
       numberOfGuestFilter !== undefined ||
       completeProfileFilter !== false ||
       queryName !== "" ||
-      (locationResult.location.lng !== 0 && locationResult.location.lat !== 0);
+      locationResult.name != "" ||
```
nabramow: I think you want `locationResult.name !== ""` here (two equal signs)

app/web/features/search/SearchPage.tsx
```diff
@@ -175,27 +177,27 @@ export default function SearchPage({
...
 
   /**
    * Handler for clearing all filters
    */
-  const handleClearFilters = (bbox?: [number, number, number, number]) => {
+  const handleClearFilters = () => {
     setQueryName("");
     setLocationResult({
-      bbox: bbox || [0, 0, 0, 0],
+      bbox: [390, 82, -173, -66],
```
nabramow, line 189: Why these hardcoded numbers?

nabramow:
> @nabramow @jesseallhands Clear filters button functionality should now match what we discussed on the meeting today. LMK if not!
> 
> A few other things I noticed while working on this and wanted to address:
> 
> 1. SearchResults List Mobile View: On mobile only, SearchResultsList is collapsed down and hidden when no filters are applied (this occurs in 2 situations - when they open the map from dashboard without an initial location search query AND every time that clear filters button is clicked). Is that okay behavior or would you prefer it be shown even when no filters are applied?
> 2. LocationAutocomplete Sync Issue: When clearing filters and when a location is searched, there's an inconsistency between two instances of LocationAutocomplete (one in SearchBox and one in FilterDialog). The location input might show different values in each place - for example, SearchBox might still display "London" from an initial Dashboard search while FilterDialog shows empty or a different location. Should this synchronization issue be addressed in this PR or as a separate ticket?

Yeah these are both outstanding issues! David is working on 1 in #5294 and 2 I've come across before. Definitely things we should fix, but they existed before this PR so you don't need to address here.

app/web/features/search/SearchPage.tsx
graceegal: They are the hardcoded numbers that set the original map position on initial render, so upon clearing filters I set them back to the same numbers to remain consistent

app/web/features/search/SearchPage.tsx
graceegal: If it is set to true by default then the clear filters button will be disabled on initial render, even if the map is initially rendered with a location filter (ie: someone searched Berlin from the homepage) which is an issue. I see what you mean about the risk of making one state variable dependent on another though... I am thinking to change it to make it dependent on the locationName prop instead. Does that seem like a good fix?

app/web/features/search/SearchPage.tsx
nabramow: Ahhh gotcha okay!

app/web/features/search/SearchPage.tsx
nabramow: It looks like locationName and bbox are passed in the props from the parent component to `SearchPage`. I assume that's coming from the dashboard (please check). If so you could use that instead, something like `locationName.length > 0` as the default value.

app/web/features/search/SearchPage.tsx
```diff
@@ -81,7 +81,9 @@ export default function SearchPage({
   const isMobile = useMediaQuery(theme.breakpoints.down("md"));
 
   // State
-  const [wasSearchPerformed, setWasSearchPerformed] = useState(false);
+  const [wasSearchPerformed, setWasSearchPerformed] = useState(
```
nexagoai, line 84: [David here] maybe a boolean it's better than empty string?

nexagoai: [I'm david BTW]

app/web/features/search/SearchPage.tsx
```diff
@@ -105,6 +107,9 @@ export default function SearchPage({
   >();
 
   const [isFiltersOpen, setIsFiltersOpen] = useState(false);
+  const [areFiltersCleared, setAreFiltersCleared] = useState(
```
nexagoai, line 110: same

app/web/features/search/SearchPage.tsx
nexagoai: I feel like setting the `wasSearchPerformed` to false can show all the users on the map?

nexagoai: sorry if I review this late, I've been busy lately 😅

nabramow: @nexagoai `locationName !== ""` does return a boolean.

app/web/features/search/SearchPage.tsx
graceegal: Hey David @nexagoai! If it is set to true by default then the clear filters button will be disabled on initial render, even if the map is initially rendered with a location filter (ie: someone searched Berlin from the homepage) which is an issue. That's why it's currently set to a conditional that checks whether it should be true or false

app/web/features/search/MapWrapper.tsx
```diff
@@ -103,13 +110,17 @@ export default function MapWrapper({
...
+  onClearFiltersClick,
 }: mapWrapperProps) {
   const { t } = useTranslation([SEARCH]);
   const [areClustersLoaded, setAreClustersLoaded] = useState(false);
   const [isMapStyleLoaded, setIsMapStyleLoaded] = useState(false);
   const [isMapSourceLoaded, setIsMapSourceLoaded] = useState(false);
   const previousResult = usePrevious(selectedResult);
   const classes = useStyles();
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down("sm"));
```
nabramow:
```suggestion
  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
```

Medium here to represent mobile as it doesn't include the size defined. So this would mean "below medium" rather than "medium and below".

app/web/features/search/MapWrapper.tsx
```diff
@@ -1,5 +1,6 @@
 import ReplayIcon from "@mui/icons-material/Replay";
 import TuneIcon from "@mui/icons-material/Tune";
+import { useMediaQuery, useTheme } from "@mui/material";
```
nabramow: Rather than `useTheme` you can just import our customized theme:

`import { theme } from "theme";`

nabramow:
Okay functionality-wise this is working great for me on both mobile and web. Happy to approve just those two small changes!

graceegal:
> Okay functionality-wise this is working great for me on both mobile and web. Happy to approve just those two small changes!

@nabramow Awesome, thanks! I've made the final 2 updates so should be ready for approval. Note- the vercel deployment failed (for what seemed like no reason just a weird hiccup) so the last 2 commits were just to fix it and have it regenerate the vercel preview (I just moved an import 1 line down and then back again because of linting). All good now though and all checks pass :)

nabramow (approved):
Looks good to me! Thanks for your persistence on this one! 🚀

PR #5414, opened 2024-12-31
Remove Storybook

aapeliv:
As per various discussions at the weekly meetings and on Slack, we've come to the conclusion to remove storybook due to lack of use and the effort required for upkeep and in particular the immediate work to make it compatible with upgrades.

This PR removes the stories and storybook configuration in such a way that it would be easy in the future to roll back.


app/web/package.json
```diff
@@ -68,18 +66,13 @@
...
-    "@storybook/builder-webpack5": "^6.5.10",
-    "@storybook/manager-webpack5": "^6.5.10",
-    "@storybook/react": "^6.5.10",
     "@testing-library/dom": "^10.4.0",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.1.0",
     "@testing-library/user-event": "^14.5.2",
     "@types/css-mediaquery": "^0.1.4",
     "@types/geojson": "^7946.0.10",
+    "@types/glob": "^8.1.0",
```
nabramow, line 75: Oh interesting what happened here?

nabramow (approved):
👋

app/web/package.json
aapeliv: I think it was being loaded as a transitive dep of some storybook package!

Issue #5299, opened 2024-12-07
Switch over MUI components to new Styled format so we can upgrade


nabramow:
**NOTE:** Best to do a few components at a time, then open a PR. Then we can more easily test those few changes and make sure nothing broke.

------

Before we can upgrade MUI to v6 and above, we need to finish changed over our style from `makeStyles` to the new `Styled`.

If you search `Styled` you should find many examples already done.

This ticket should be relatively straight-forward though it will be a lot of changes. Feel free to do PRs in sections, as it makes it easier to test, i.e. do a group of 10 components in 1 PR, open it. Do another PR with another 10.

** Make sure to import `styled` from `@mui/material` (it comes up from other packages in the suggestions).**

Please name the new styled components starting with `Styled`, i.e. `StyledAlert`, `StyledCollapse`, etc..

Here is the migration from JSS guide from MUI if it helps:
https://mui.com/material-ui/migration/migrating-from-jss/

Old way:

```javascript
const useStyles = makeStyles((theme) => ({
  root: {
    marginBottom: theme.spacing(2),
  },
}));
```

New way:

```javascript
import { theme } from "theme";

const StyledAlert = styled("div", {
    marginBottom: theme.spacing(2),
});
```

PR #5445, opened 2025-01-06
Web - Switch over MUI components to new Styled format

jkurlowicz:
Change styling from JSS to Emotion so we can upgrade MUI.

part of #5299 

**Web frontend checklist**
- [x] All tests pass


nabramow (approved):
This looks good to me! Tested and the CircularProgress seems to be displaying as normal across the app.

The IconButton loading state also seems okay. We only use that prop in `AdminsDialog` in the GroupChatHeaderBar and it looks good.

PR #5446, opened 2025-01-06
GroupChatView - Fix console warning and select behavior

### LINKED ISSUE #5379: Bug from React 18 upgrade - cannot be given refs
Making a ticket so I can handle it later. I got this console error when I loaded the messages page. I probably forgot to handle the new ref passing in one of the components there. This is the error:

Image

nabramow:
Closes #5379 

Bug introduced during the React v18 upgrade. Fixing here. There were two issues:

- When it's a multiselect spreading the `field` messes with the `options` being passed. For this specific component passing explicit `field.value` and `field.onChange` is enough.
- The values weren't being shown as selected (i.e. highlighted in green in dropdown). This is because they were objects and MUI didn't know the value. Adding `isOptionEqualToValue` fixes it.

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/features/messages/groupchats/CreateGroupChat.tsx
+++ b/app/web/features/messages/groupchats/CreateGroupChat.tsx
@@ -146,8 +146,10 @@ export default function CreateGroupChat({ className }: { className?: string }) {
                   render={({ field }) => {
                     return (
                       <Autocomplete
-                        {...field}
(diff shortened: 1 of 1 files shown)


jkurlowicz (approved):
lgtm :) Checked locally no related errors anymore.

PR #5452, opened 2025-01-09
Convert some more MUI to styled

nabramow:
Another chunk of conversions to new styled format.

Part of #5299 

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/components/IconText.tsx
+++ b/app/web/components/IconText.tsx
@@ -1,19 +1,18 @@
-import { Typography } from "@mui/material";
+import { styled, Typography } from "@mui/material";
 import { OverridableComponent } from "@mui/material/OverridableComponent";
 import { SvgIconTypeMap } from "@mui/material/SvgIcon";
 import React, { ReactNode } from "react";
-import makeStyles from "utils/makeStyles";
+import { theme } from "theme";
 
-const useStyles = makeStyles((theme) => ({
-  label: {
-    marginInlineStart: theme.spacing(1),
-  },
-  root: {
-    alignItems: "center",
-    display: "flex",
-    marginBottom: theme.spacing(1),
-    marginTop: theme.spacing(1),
-  },
+const StyledWrapper = styled("div")(({ theme }) => ({
+  alignItems: "center",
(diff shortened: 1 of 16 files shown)


app/web/components/IconText.tsx
```diff
@@ -1,19 +1,18 @@
...
-  },
+const StyledWrapper = styled("div")(({ theme }) => ({
+  alignItems: "center",
+  display: "flex",
+  marginBottom: theme.spacing(1),
+  marginTop: theme.spacing(1),
+}));
+
+const StyledLabel = styled("div")(({ theme }) => ({
+  marginLeft: theme.spacing(1),
```
jkurlowicz: why do we change marginInlineStart to marginLeft here?

app/web/components/LabelAndText.tsx
```diff
@@ -27,21 +21,29 @@ export interface LabelAndTextProps {
...
+          display: "flex",
+          alignItems: "center",
+        }}
       >
         {label}
       </Typography>
       {typeof text === "string" ? (
-        <TextBody className={classes.flexItem}>{text}</TextBody>
+        <TextBody
+          sx={{ flex: "1 1 50%", display: "flex", alignItems: "center" }}
```
jkurlowicz, line 40: I wonder if it makes sense to keep those properties as a separate object instead of having the same properties in 3 different places?

app/web/components/IconText.tsx
nabramow: Ooo good catch I think Github Copilot sabotaged me there 😅.

app/web/components/LabelAndText.tsx
nabramow: Yeah I thought about that. In general I'm a fan of keeping things separate even if it's a bit repetitive. It makes it easier to refactor one thing without introducing bugs to another (learned this the hard way). I think it's okay for now since this component is pretty small.

nabramow:
@jkurlowicz Are you happy with this PR? If so would you mind approving?

PR #5475, opened 2025-01-15
Revert "frontend - add clear filter button #5275"

nabramow:
Reverts Couchers-org/couchers#5413

This is causing a major bug with the map where after clearing filters it doesn't reset to the initial map with all the pins when there are multiple pages of uses.

We didn't notice it testing as there's not enough users for multiple pages with the stage data.

I tried to quickly fix, but wasn't able to come up with an immediate solution, so reverting this PR as it's a big bug.


nabramow:
Okay nvm I think I fixed it.

aapeliv (approved):
LGTM

PR #5485, opened 2025-01-18
Upgrade prettier and eslint-config-prettier

### LINKED ISSUE #5484: Upgrade prettier and eslint-config-prettier
We're 3 years since last upgrade of these 🫣.

Upgrade to latest version:

- prettier
- eslint-config-prettier

https://prettier.io/blog/2023/07/05/3.0.0.html#javascript

nabramow:
Closes #5484 

We haven't upgraded these packages for years so getting them up to date! The number of files makes this look much scarier than it is. There are so many changes as default format settings for prettier changed with the upgrade, so a lot of code formatting ran. **Almost all the changes are trailing commas being added.** So...not very risky ;-).

There shouldn't be any changes to any functionality here. So the only thing to look for will be if you have strong feelings about any of the new formatting.

See here for breaking changes: https://prettier.io/blog/2023/07/05/3.0.0.html#javascript

**Web frontend checklist**
- [ x ] All tests pass


app/web/cms/index.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
```
aapeliv, line 1: I think this is historical legacy stuff and technically better to be uppercase.

aapeliv (changes requested):
Please add the commit to `.git-blame-ignore-revs`.

Also the doctype comment is not very important if it takes you more than 1 min.

app/web/cms/index.html
nabramow: According to html standards doc looks like it's case insensitive:

https://html.spec.whatwg.org/multipage/syntax.html#the-doctype

nabramow: Also here's the discussion on this change in the prettier repo, looks like no one reported any bugs:

https://github.com/prettier/prettier/issues/15096

nabramow:
> Please add the commit to `.git-blame-ignore-revs`.
> 
> Also the doctype comment is not very important if it takes you more than 1 min.

Done! By the way, what's the purpose of this?

app/web/cms/index.html
aapeliv: OK no worries if it's not causing issues.

aapeliv:
> > Please add the commit to `.git-blame-ignore-revs`.
> > Also the doctype comment is not very important if it takes you more than 1 min.
> 
> Done! By the way, what's the purpose of this?

When you do a `git blame` it won't show this commit: makes it easier to blame if there's formatting changes in between.

aapeliv (approved):
LGTM

PR #5500, opened 2025-01-20
Upgrade glob

### LINKED ISSUE #5497: Update glob to from v8 to v11
We only use glob in this file: app/web/utils/markdownPages.ts

But it's quite out of date and we get a warning that versions.

Changelog to refer to for any possible breaking changes:
https://github.com/isaacs/node-glob/blob/main/changelog.md

nabramow:
Closes #5497 

Upgraded from v8 to v11. Only one change in those migrations effects us from v10: 

* [No default exports, only named exports](https://github.com/isaacs/node-glob/blob/main/changelog.md#1000)

We only use this in one file (app/web/utils/markdownPages.ts), but this effects all the markdown pages. That said, if one works, they all do, as they're all handled in that function.

So to test try going to volunteer page, blog, open source page, mission, terms of service, etc.

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/package.json
+++ b/app/web/package.json
@@ -39,7 +39,7 @@
     "dayjs": "^1.11.13",
     "frontmatter-markdown-loader": "^3.7.0",
     "geojson": "^0.5.0",
-    "glob": "^8.0.3",
+    "glob": "^11.0.1",
     "google-protobuf": "^3.21.4",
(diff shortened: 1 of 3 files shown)


aapeliv (approved):
LGTM

PR #5505, opened 2025-01-20
Didn't meet up backend

### LINKED ISSUE #5507: Add `did_not_stay` and reason to backend for HostRequest
We need some adjustments to the backend so the frontend can add UI to the Leave Reference flow in order to mark that though the Host Request was accepted, the surfer didn't end up staying.

Acceptance Criteria:
- Add did_stay: boolean and did_not_stay_reason: string to the HostRequest object on the backend
- Allow passing these it params in the leave reference api call

aapeliv:
Closes #5507.

I haven't updated the email templates yet as the frontend isn't ready.

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/references.py
+++ b/app/backend/src/couchers/servicers/references.py
@@ -5,6 +5,7 @@
 * References become visible after min{2 weeks, both reciprocal references written}
 """
 
+from google.protobuf import empty_pb2
 from types import SimpleNamespace
 
 import grpc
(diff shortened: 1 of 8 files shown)


aapeliv:
@jasonwvh could you have a look at this and review please?

app/backend/src/couchers/servicers/references.py
```diff
@@ -258,6 +278,18 @@ def WriteHostRequestReference(self, request, context, session):
...
+    def HostRequestIndicateDidntMeetup(self, request, context, session):
+        host_request, surfed = get_host_req_and_check_can_write_ref(session, context, request.host_request_id)
+
+        reason = request.reason_didnt_meetup.strip()
+
+        if surfed:
+            host_request.surfer_reason_didnt_meetup = reason
+        else:
+            host_request.host_reason_didnt_meetup = reason
+
```
jasonwvh, line 290: I'm not most familiar with sqlalchemy but do we need a `session.commit()` here to commit the changes?

aapeliv: We have this `session_scope` context manager (that's `context` in the args to the API call):

https://github.com/Couchers-org/couchers/blob/2d98600263c414ce7dc99885777690ddfccb95a0/app/backend/src/couchers/db.py#L62-L88

We wrap each req in that context manager in this "interceptor" (middleware):

https://github.com/Couchers-org/couchers/blob/2d98600263c414ce7dc99885777690ddfccb95a0/app/backend/src/couchers/interceptors.py#L289-L307

(You can see where this is set up here:

https://github.com/Couchers-org/couchers/blob/2d98600263c414ce7dc99885777690ddfccb95a0/app/backend/src/couchers/server.py#L67-L79

)

aapeliv: So yes, in vanilla SQLAlchemy you do need it, but we make sure to commit everything at the end of an API call in our middleware so here you don't!

jasonwvh (approved):
Ah got it, lgtm!

PR #5522, opened 2025-01-22
Fix backend readme

nabramow:
This got changed recently and confused me for a good 20 minutes while I was checking some python tests I wrote. The way the README was changed makes it look like it's two separate commands one-after-the-next, but really it's all one line to be submit at once.

**Backend checklist**
- [ ] All tests pass


aapeliv (approved):
Ah you're right, thanks for picking that up.

Krishnag09 (approved):
Good catch, thanks.

PR #5552, opened 2025-01-28
Remove unused AuthHeader

nabramow:
I noticed we're not using this component anywhere while I was doing MUI conversions. Removing it here as no point having it sit around then!

**Web frontend checklist**


aapeliv (dismissed):
LGTM

aapeliv (approved):
still LGTM :)

PR #5566, opened 2025-01-30
App: Add Friend button to profile

polemius:
Add Friend button to profile. 

image

Use could also Accept or Decline request.
image
image


aapeliv (approved):
Great!! LGTM

PR #5603, opened 2025-02-13
fix card alignment in sidebar of map search it was incorrect when cli…

### LINKED ISSUE #5269: [bug / desktop] Card alignment in sidebar of map search is incorrect when clicking on pin
### Problem
On desktop, instead of the card in the sidebar lining up with the top navbar, it is partially covered by the navbar:
image

### Solution
It appears the alignment of the card is set to align with the top of the browser window instead of the bottom of the navbar.

joachimageron:
…cking on pin

closes #5269 

**Web frontend checklist**
- [x] All tests pass 

### DIFF
--- a/app/web/features/search/SearchResultsList.tsx
+++ b/app/web/features/search/SearchResultsList.tsx
@@ -48,7 +48,7 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.background.paper,
     boxShadow: "0 0 4px rgba(0,0,0,0.25)",
     marginBottom: theme.spacing(3),
-    scrollMarginTop: "200px",
+    scrollMarginTop: theme.shape.navPaddingSmUp,
     overflow: "initial",
     "&:last-child": {
       marginBottom: 0,
--- a/app/web/features/search/SearchResultsList.tsx
+++ b/app/web/features/search/SearchResultsList.tsx
@@ -48,6 +48,8 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.background.paper,
     boxShadow: "0 0 4px rgba(0,0,0,0.25)",
     marginBottom: theme.spacing(3),
+    scrollMarginTop: "200px",
+    overflow: "initial",
     "&:last-child": {
       marginBottom: 0,
     },


app/web/features/search/SearchResultsList.tsx
```diff
@@ -48,6 +48,8 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.background.paper,
     boxShadow: "0 0 4px rgba(0,0,0,0.25)",
     marginBottom: theme.spacing(3),
+    scrollMarginTop: "200px",
```
nabramow: We have a variable for the navbar height you can use actually rather than hardcoding it to `200px` which might be too far down on mobile, for example.

You can use `theme.shape.navPaddingXs` for smalls screens and `theme.shape.navPaddingSmUp` for larger.

Try searching those variables in the codebase in your IDE and you'll see some examples of how to handle the different screen sizes.

joachimageron:
Ok I'll watch

app/web/features/search/SearchResultsList.tsx
```diff
@@ -48,7 +48,7 @@ const useStyles = makeStyles((theme) => ({
     backgroundColor: theme.palette.background.paper,
     boxShadow: "0 0 4px rgba(0,0,0,0.25)",
     marginBottom: theme.spacing(3),
-    scrollMarginTop: "200px",
+    scrollMarginTop: theme.shape.navPaddingSmUp,
```
nabramow, line 51: What happens if the screen size is small?

Take a closer look at other places this variable is being used and how it's handled on different screen sizes.

We usually have one case for Xs and one for SmUp.


Also just FYI, the map is currently being re-written, so this code will be obsolete soon. Totally fine if you want to work on it as a learning exercise, but if you want to pick up something else no worries there. Or if you need help picking something just ping me!

joachimageron: When the screen is small the profiles are horizontal so this property doesn't apply.

Ha ok I didn't knew I'll do something else

nabramow: @joachimageron Do you wanna go ahead and close this PR then? Or do you plan to add more to it?

PR #5604, opened 2025-02-13
Disables save button in edit profile until profile image is done uploading

### LINKED ISSUE #5184: Clicking "save" before profile image has finished upload terminates upload
**Nicole Note:** Before adding a pop up, see if it's possible to give better user feedback that the profile photo is still uploading. That gets to the core of the issue more. This would look like:

- Disabling the "Save" button and chucking a spinner on it while the profile photo is uploading. 
- Making sure the spinner shows the whole time on the actual photo (seems it just shows after "confirm" click?)
- Bonus: removing the confirm check/X and just uploading as default (can be in a separate ticket if that's too much).

If it turns out there's some technical problem preventing doing the above...we can do the below as a stop gap.

----------------------

Problem
When a user is editing their profile and their profile image is taking a while to upload (after clicking the checkmark) and the spinner is spinning, if the user clicks "save" it will kill the upload and navigate back to the "my profile" page without having saved your new profile picture. There is no warning.

### Solution
A warning should alert the user that their profile picture has not finished uploading yet and they should wait until it is complete before saving their profile. 

Warning text:
`Your profile picture is not uploaded yet! Please wait and click "save" after it has finished uploading.`
[Okay]

If the user clicks "okay" on the aforementioned warning and then tries to click away to navigate somewhere else on the site, they will still get the same warning as currently appears:
image

shane-jeon:
Closes #5184 
- Disables save button when profile image is still uploading
- Displays progress circle on save button when disabled during image upload

~~currently (as of 2/13/25) working on removing confirm check/X. intending to write tests after this is figured out.~~
4/10/25: Determined removing confirm check/X should be a separate ticket.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [X] Formatted my code with `yarn format`
- [X] There are no warnings from `yarn lint --fix`
- [x] All tests pass

### DIFF
--- a/app/web/features/profile/edit/EditProfilePage.test.tsx
+++ b/app/web/features/profile/edit/EditProfilePage.test.tsx
@@ -77,7 +77,7 @@ describe("Edit profile", () => {
       () => expect(mockRouter.pathname).toBe(routeToProfile("about")),
       { timeout: 5000 },
     );
-  }, 60000);
+  }, 8000);
 
   it(`should not submit the default headings for the '${t(
     "profile:heading.who_section",
--- a/app/web/features/profile/edit/EditProfilePage.test.tsx
+++ b/app/web/features/profile/edit/EditProfilePage.test.tsx
@@ -77,7 +77,7 @@ describe("Edit profile", () => {
       () => expect(mockRouter.pathname).toBe(routeToProfile("about")),
       { timeout: 5000 },
     );
-  });
+  }, 60000);
 
   it(`should not submit the default headings for the '${t(
     "profile:heading.who_section",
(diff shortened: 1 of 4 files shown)


shane-jeon:
After mulling it over, I think it best to make the removal of the confirm check/x for image upload should be a separate ticket.

app/web/features/profile/edit/EditProfilePage.test.tsx
```diff
@@ -77,7 +77,7 @@ describe("Edit profile", () => {
       () => expect(mockRouter.pathname).toBe(routeToProfile("about")),
       { timeout: 5000 },
     );
-  });
+  }, 60000);
```
nabramow: Does this need to be quite so high?

nabramow:
@shane-jeon just one little comment about the test timeout length change otherwise everything looks good!

app/web/features/profile/edit/EditProfilePage.test.tsx
shane-jeon: oopsie, added an extra 0

nabramow (approved):
Okay you're good to go! Merge away!

PR #5610, opened 2025-02-16
Add backend filter

nabramow:
**Backend checklist**
- [ ] All tests pass


app/backend/src/couchers/servicers/search.py
```diff
@@ -540,8 +542,6 @@ def UserSearch(self, request, context, session):
         # TODO:
         # google.protobuf.StringValue language = 11;
         # bool friends_only = 13;
-        # google.protobuf.UInt32Value age_min = 14;
-        # google.protobuf.UInt32Value age_max = 15;
```
nabramow, line 544: Looks like you already did this below but correct me if I'm wrong!

Krishnag09: I was using them to  add default filter values for age but didn't up completing the original age filter. So removal looks good for now.

Krishnag09 (approved):
Thanks for formatting and tests.

PR #5626, opened 2025-02-22
Add members tab to community page

nabramow:
I was planning to make a ticket for this, but then I looked at the code and it was super easy to implement so I just did it.

Nothing fancy, we can add stuff as we go, but I think it provides a lot of value as a starting point.

**Web frontend checklist**
- [ x ] All tests pass


aapeliv:
Screenshot from 2025-02-22 18-35-32

Would it be better to hide overflow than wrap?

I think it would look weird then as it's just cut off in the middle. I could add an ellipsis (...) after a certain amount of letters instead?

aapeliv (approved):
See comment, otherwise LGTM!

PR #5632, opened 2025-02-23
Fix up SV filter in user search

aapeliv:
**Backend checklist**
- [ ] All tests pass


nabramow (approved):
Thanks for fixing my mistake!

PR #5643, opened 2025-02-25
Update Other pages “roadmap”

c-kreuz:
Automatically generated by Netlify CMS


aapeliv:
This is how it looks right now:

https://couchers-git-cms-generalroadmap-couchers-org.vercel.app/roadmap

I think maybe it needs a little more styling?

nabramow:
> This is how it looks right now:
> 
> https://couchers-git-cms-generalroadmap-couchers-org.vercel.app/roadmap
> 
> I think maybe it needs a little more styling?

My only this is would it be possible to add column borders? I'm not so sure it's clear the staggered points are columns on first glance.

aapeliv (approved):

PR #5651, opened 2025-02-28
edit community info

### LINKED ISSUE #5120: No clear way to exit "edit community" page
Problem:
When editing a community on desktop (for example [here](https://couchers.org/community/3215/los-angeles/info/edit)), you can update the image and text and then click the update" button and it will make the update, but there is no "done" button or other button to exit out of the editing page back to the standard page. The only option is to navigate somewhere else.

Fix:
Change the "update" button to a "save" button and after clicking the button, return the user to the standard version of that page. (same behavior as editing the profile page).

**Nicole Hint:** Don't take this as law, but it might be as simple as doing a `router.push` to `/[community-name]/info` from `[community-name]/info/edit`.

ColleenLillian:
closes #5120 
changed button text to "save" and redirects to main community page

**Web frontend checklist**
- [ x] There are no warnings from `yarn lint --fix`
- [ ] All tests pass

### DIFF
--- a/app/web/features/communities/EditCommunityInfoPage.tsx
+++ b/app/web/features/communities/EditCommunityInfoPage.tsx
@@ -11,6 +11,7 @@ import { communityKey } from "features/queryKeys";
 import { RpcError } from "grpc-web";
 import { useTranslation } from "i18n";
 import { COMMUNITIES, GLOBAL } from "i18n/namespaces";
+import { useRouter } from "next/router";
 import { Community } from "proto/communities_pb";
 import { Page } from "proto/pages_pb";
 import { useForm } from "react-hook-form";
@@ -20,8 +21,6 @@ import { service } from "service";
 import makeStyles from "utils/makeStyles";
 
 import CommunityBase from "./CommunityBase";
-import { useRouter } from "next/router";
-
 
 const useStyles = makeStyles((theme) => ({
   root: {
--- a/app/web/features/communities/EditCommunityInfoPage.tsx
+++ b/app/web/features/communities/EditCommunityInfoPage.tsx
@@ -20,6 +20,8 @@ import { service } from "service";
 import makeStyles from "utils/makeStyles";
 
 import CommunityBase from "./CommunityBase";
+import { useRouter } from "next/router";
+
 
 const useStyles = makeStyles((theme) => ({
   root: {
@@ -58,6 +60,7 @@ export default function EditCommunityPage({
   const { t } = useTranslation([GLOBAL, COMMUNITIES]);
   const classes = useStyles();
   const queryClient = useQueryClient();
+  const router = useRouter();
   const {
     control,
     handleSubmit,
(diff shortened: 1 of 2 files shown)


app/web/features/communities/EditCommunityInfoPage.tsx
```diff
@@ -176,7 +176,7 @@ export default function EditCommunityPage({
           </>
         ) : (
           <Redirect
-            to={routeToCommunity(community.communityId, community.slug, "info")}
+            to={routeToCommunity(community.communityId, community.slug)}
```
nabramow: I'd leave this how it was. This one seems to be more for security, so that if someone manually goes to `community/edit`. but they're not an Admin, it will redirect them back.

nabramow:
You got the idea but I'd suggest a different approach!

- You can add a router into EditCommunity page, i.e. `const router = useRouter()` (from `next/router`)
- Console log that out to see what's in there, when you're in edit mode you should see edit in the pathname
- Since we want this to redirect the user after they submit the community description change, find where the save/submit is happening, in which function
- In that function, get the current path from the router variable you made
- Make a new variable that has the newPath you want to redirect them to (i.e. same path but without `/edit` at the end
- Push the new path to the router in the submit function after the update function
- You should be able to test it by saving in a community you can edit. If you're logged into aapeli's localdev account, for example, he's an admin of everything.
- You will need to run against the local database for that (usually the app uses stage data)
- To do that copy everything from `.env.localdev` and replace what's in `.env.development`, then login again as aapeli (you can find the login in app/backend/readme.md, or search `Aapeli's password` in VSCode it'll come up) (don't commit this file, you'll want to change it back before pushing to the branch) 
- Go to a community, hit local info, edit, check your code works


Just ping me if any questions! Sorry that is a novel!

ColleenLillian:
This seems to work locally now but showed this error btw
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node

nabramow:
> This seems to work locally now but showed this error btw NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node

That error is unrelated to you, no worries! It's a long-standing thing that only happens locally we need to eventually sort out.

Looks like you got some failing formatting, linting and tests though. The first two just follow the instructions in the PR checklist above. The latter looks like it's because the button string was changed to "save" here but not in some tests.

If you click in `ci/gitlab/gitlab.com for the CI test checks above and then click into the failing "web:test" you can see which tests failed.

Or run `yarn test` locally.

ColleenLillian:
all tests pass now.
Got a bit weird with save conflicts but it just kept wanting to reorder the imports that's all

nabramow (approved):
Excellent, this is working great now, well done! Merge away 🚀!

ColleenLillian:
umm why does it say merged now?! that was not my intention :/

PR #5675, opened 2025-03-03
Backend/language preference cookie #5468

### LINKED ISSUE #5468: Language preference cookie
Once #5467 is complete, the backend should send the language preference in a cookie. It should be overwritten when the user is logged in, but otherwise left untouched.

laurabotel:
Creates cookie & corresponding utils for ui language preference. Closes #5468 

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/interceptors.py
```diff
@@ -192,28 +202,37 @@ def user_unaware_function(req, context):
...
-                    context.send_initial_metadata(
-                        [("set-cookie", cookie) for cookie in create_session_cookies(token, context.user_id, expiry)]
-                    )
-                except ValueError as e:
-                    logger.info("Tried to send initial metadata but wasn't allowed to")
+                cookies.extend(
+                    [("set-cookie", cookie) for cookie in create_session_cookies(token, context.user_id, expiry)]
+                )
+
+            if context.ui_language_preference != cookie_ui_lang:
```
aapeliv: Maybe we only want to set this for logged in users and non-API requests? You could break up the if above on line 222.

app/backend/src/couchers/servicers/account.py
```diff
@@ -248,6 +249,16 @@ def ChangeLanguagePreference(self, request, context, session):
...
+        # setting this on context will update the cookie (via interceptors)?
+        context.ui_language_preference = request.ui_language_preference
+        expiry = now() + timedelta(days=400)
+
+        try:
+            context.send_initial_metadata(
+                [("set-cookie", cookie) for cookie in create_lang_cookie(context.ui_language_preference, expiry)]
+            )
+        except ValueError as e:
+            logger.info("Tried to send initial metadata but wasn't allowed to")
```
aapeliv: I think you don't need to send the cookie: the interceptor will do it anyway.

app/backend/src/couchers/utils.py
```diff
@@ -218,6 +218,10 @@ def create_session_cookies(token, user_id, expiry) -> list[str]:
     ]
 
 
+def create_lang_cookie(lang, expiry):
+    return [_create_tasty_cookie("couchers-lang", lang, expiry, httponly=False)]
```
aapeliv: Please make a new constant in `constants.py`, something like `LANGUAGE_PREFERENCE_COOKIE_EXPIRY = timedelta(days=3650)` and use that in place of the argument here.

aapeliv (changes requested):
This is really good! I left a few minor comments.

app/backend/src/couchers/servicers/account.py
laurabotel: it may be an issue with my [test](https://github.com/Couchers-org/couchers/pull/5675/files#diff-85b0024f6edf69c506935f6158c45beafd59da91355c9066488f29ac71310591R531), but invoking the `ChangeLanguagePreference` request didn't seem to send a cookie, or at the very least it didn't show up in the initial metadata 🤔   maybe I should be testing this in test_interceptors.py instead?

aapeliv: Yeah you'll need to have the interceptor there (mimicking `test_interceptors.py`) and the way the metadata comes out of there, otherwise it's not going to work. I can help, let me know if you want to? Also down to chat about it!

app/backend/src/couchers/interceptors.py
```diff
@@ -192,28 +202,37 @@ def user_unaware_function(req, context):
...
             res = user_aware_function(req, context)
+            cookies = []
 
-            # check the two cookies are in sync
-            if context.user_id and not context.is_api_key and cookie_user_id != str(context.user_id):
-                try:
-                    token, expiry = context.token
-                    context.send_initial_metadata(
+            # check the two cookies are in sync & that language preference cookie is correct
+            token, expiry = context.token
```
aapeliv: Please move this one line down into the if statement

app/backend/src/couchers/interceptors.py
```diff
@@ -192,28 +202,37 @@ def user_unaware_function(req, context):
...
-                    logger.info("Tried to send initial metadata but wasn't allowed to")
+                if context.ui_language_preference != cookie_ui_lang:
+                    cookies.extend(
+                        [("set-cookie", cookie) for cookie in create_lang_cookie(context.ui_language_preference)]
+                    )
+
+            try:
+                context.send_initial_metadata(cookies)
+            except ValueError as e:
+                logger.info("Tried to send initial metadata but wasn't allowed to")
```
aapeliv: And indent this one indent so it is also in the if statement

app/backend/src/couchers/interceptors.py
```diff
@@ -192,28 +202,37 @@ def user_unaware_function(req, context):
...
         headers = dict(handler_call_details.invocation_metadata)
         cookie_user_id = parse_user_id_cookie(headers)
+        cookie_ui_lang = parse_ui_lang_cookie(headers)
 
         handler = continuation(handler_call_details)
         user_aware_function = handler.unary_unary
 
         def user_unaware_function(req, context):
             res = user_aware_function(req, context)
+            cookies = []
```
aapeliv: And put this in the if as well :)

aapeliv (changes requested):
Just one minor thing and this is ready to go in!

aapeliv (approved):
Awesome, looks good!

PR #5703, opened 2025-03-11
Backend: Reference text editing API

### LINKED ISSUE #5547: Edit reference text API
We need an api to be able to edit the text of a reference hat has been left.

Why? In case a user posts illegal content or private/personal information unrelated to the hosting/surfing/meetup experience.

Who? Only admins should be able to call this api

After the API has been created, it needs to be added to the admin console (new, separate ticket).

rafaelferreiranvr:
Implements an API that allow admins to edit the text from a reference. Closes #5547.

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/servicers/admin.py
```diff
@@ -531,3 +532,17 @@ def ListUserIds(self, request, context, session):
...
+    def EditReferenceText(self, request, context, session):
+        reference = session.execute(select(Reference).where(Reference.id == request.reference_id)).scalar_one_or_none()
+
+        if reference is None:
+            context.abort(grpc.StatusCode.NOT_FOUND, errors.REFERENCE_NOT_FOUND)
+
+        if request.new_text.strip() == "":
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.REFERENCE_NO_TEXT)
+
+        reference.text = request.new_text
```
aapeliv: Please strip also here

app/backend/src/couchers/servicers/admin.py
```diff
@@ -531,3 +532,17 @@ def ListUserIds(self, request, context, session):
...
             next_page_token=str(user_ids[-1]) if len(user_ids) > page_size else None,
         )
+
+    def EditReferenceText(self, request, context, session):
+        reference = session.execute(select(Reference).where(Reference.id == request.reference_id)).scalar_one_or_none()
+
+        if reference is None:
+            context.abort(grpc.StatusCode.NOT_FOUND, errors.REFERENCE_NOT_FOUND)
+
+        if request.new_text.strip() == "":
```
aapeliv: You can also do `if not request.new_text.strip()` in python (an empty string is falsey), but that's minor

app/backend/src/couchers/servicers/admin.py
```diff
@@ -531,3 +532,17 @@ def ListUserIds(self, request, context, session):
...
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.REFERENCE_NO_TEXT)
+
+        reference.text = request.new_text
+        session.commit()
```
aapeliv: No need to `session.commit`, the whole request is wrapped in a `session_scope()` context manager that will auto-commit at exit. See here:

https://github.com/Couchers-org/couchers/blob/db5ca2e55d34bfcf20f16be88aeebf95a9a9fe02/app/backend/src/couchers/interceptors.py#L289-L307

https://github.com/Couchers-org/couchers/blob/db5ca2e55d34bfcf20f16be88aeebf95a9a9fe02/app/backend/src/couchers/db.py#L81-L82

app/proto/admin.proto
```diff
@@ -268,3 +271,8 @@ message ListUserIdsRes {
 
   string next_page_token = 2;
 }
+
+message EditReferenceTextReq {
+  int64 reference_id = 1;
+  string new_text = 6; //plain text
```
aapeliv: Please do `= 2` here. It's the gRPC tag number (grpc doesn't send the field name, just the identifier and type)

app/backend/src/tests/test_admin.py
```diff
@@ -645,4 +645,34 @@ def test_ListUserIds(db):
...
+            res = api.EditReferenceText(
+                admin_pb2.EditReferenceTextReq(reference_id=reference.id, new_text=test_new_text)
+            )
+
+        session.expire_all()
+
+        modified_reference = session.execute(select(Reference).where(Reference.id == reference.id)).scalar_one_or_none()
+        assert modified_reference.text == test_new_text
+
+
```
aapeliv, line 676: This test is good. I guess in general it's slightly better to do the testing via the API (make ref via API, edit like here via API, then list/get via API). That way you're testing it all. But this is good too, no need to change.

aapeliv (changes requested):
A few minor things! I tried to be descriptive to help you figure out idiosyncrasies of the codebase!

aapeliv (approved):
Awesome, looks good to me!

PR #5707, opened 2025-03-13
Backend: Reference Deletion API

### LINKED ISSUE #5548: Delete reference API
We need an api to be able to delete a reference hat has been left.

Why? 
1. A reference is left by mistake (eg. the user didn't actually stay with the person)
2. In case a user posts illegal content or private/personal information unrelated to the hosting/surfing/meetup experience to the extent that the entire reference needs to be removed

Who? Only admins should be able to call this api

After the API has been created, it needs to be added to the admin console (new, separate ticket).

rafaelferreiranvr:
Implements an API that allow admins to remove references. Closes #5548 

**Backend checklist**
- [x] All tests pass


app/proto/references.proto
```diff
@@ -56,6 +56,8 @@ message Reference {
 
   // if the reference is about the current user, this tells which host request it's about
   int64 host_request_id = 7;
+
+  bool is_deleted = 8;
```
aapeliv: No need for this here since it's not returned

app/backend/src/tests/test_admin.py
```diff
@@ -674,4 +674,33 @@ def test_EditReferenceText(db):
...
+                session.expire_all()
+
+                modified_reference = session.execute(
+                    select(Reference).where(Reference.id == reference.reference_id)
+                ).scalar_one_or_none()
+
+            reference_list = api.ListReferences(references_pb2.ListReferencesReq(from_user_id=user1.id)).references
+
+        assert modified_reference.is_deleted
+        assert not reference_list
```
aapeliv: You can simplify to something like this:

```py
    with references_session(user1_token) as api:
        reference = api.WriteFriendReference(
            references_pb2.WriteFriendReferenceReq(
                to_user_id=user2.id, text="Old Text", private_text="", was_appropriate=True, rating=1
            )
        )

    with references_session(user1_token) as api:
        assert reference_list = api.ListReferences(references_pb2.ListReferencesReq(from_user_id=user1.id)).references

    with real_admin_session(super_token) as admin_api:
        admin_api.DeleteReference(admin_pb2.DeleteReferenceReq(reference_id=reference.reference_id))

    with references_session(user1_token) as api:
        assert not reference_list = api.ListReferences(references_pb2.ListReferencesReq(from_user_id=user1.id)).references

        modified_reference = session.execute(
            select(Reference).where(Reference.id == reference.reference_id)
        ).scalar_one_or_none()

        assert modified_reference.is_deleted
```

app/backend/src/couchers/servicers/references.py
```diff
@@ -109,7 +109,7 @@ def ListReferences(self, request, context, session):
 
         to_users = aliased(User)
         from_users = aliased(User)
-        statement = select(Reference)
+        statement = select(Reference).where(Reference.is_deleted == False)
```
aapeliv, line 112: Please also add this check in `user_model_to_pb` in `api.py`, line 796.

It might be good to add a `@hybrid_property` for `is_visible` in the Reference model and use that, but I don't think that's so important

aapeliv (approved):
LGTM!

PR #5710, opened 2025-03-13
Fix discussion reply notification issue

### LINKED ISSUE #5698: Exception: Tried to send push notification to User(id=141, email=...@gmail.com, username=...) but didn't have push info
Sentry Issue: [BACKEND-DW](https://couchers.sentry.io/issues/6378448910/?referrer=github_integration)

```
Exception: Tried to send push notification to User(id=141, email=...+test@gmail.com, username=...) but didn't have push info
  File "couchers/jobs/worker.py", line 72, in process_job
    ret = func(message_type.FromString(job.payload))
  File "couchers/notifications/background.py", line 177, in handle_notification
    _send_push_notification(session, user, notification)
  File "couchers/notifications/background.py", line 120, in _send_push_notification
    raise Exception(f"Tried to send push notification to {user} but didn't have push info")

Tried to send push notification to User(id=141, email=...+test@gmail.com, username=...) but didn't have push info
```

aapeliv:
Closes #5698

I was checking for which part of the proto was there incorrectly: this fixes that. (`data.event` and `data.discussion` both evaluate to truthy things, I had to explicilty check with `HasField`). See the docs for details:

https://protobuf.dev/reference/python/python-generated/#oneof

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/notifications/render.py
+++ b/app/backend/src/couchers/notifications/render.py
@@ -634,10 +634,11 @@ def render_notification(user, notification) -> RenderedNotification:
                 push_url=discussion_link,
             )
     elif notification.topic_action.display == "thread:reply":
-        if data.HasField("event"):
+        parent = data.WhichOneof("reply_parent")
+        if parent == "event":
             title = data.event.title
             view_link = urls.event_link(occurrence_id=data.event.event_id, slug=data.event.slug)
-        elif data.HasField("discussion"):
+        elif parent == "discussion":
             title = data.discussion.title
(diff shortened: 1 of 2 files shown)


rafaelferreiranvr:
Since those two fields return true when evaluating and both are part of the same oneof field, it would make sense doing this:
```
parent = data.WhichOneof("reply_parent")
if parent == "event":
    ...
elif parent == "discussion":
    ...
else:
    raise Exception("Can only do replies to events and discussions")
```

But your solution works too

rafaelferreiranvr (dismissed):

rafaelferreiranvr (approved):
Looks good!

PR #5713, opened 2025-03-14
Remove smoking_allowed

### LINKED ISSUE #5711: Smoking location filter bug
There is a bug right now where the smoking filter doesn't work and this shows up in the Map POC (#5701). Basically `smoking_allowed` is not actually a bool, it's this enum:

https://github.com/Couchers-org/couchers/blob/3b8252289e77874709553ed5a3845db192540ccc/app/backend/src/couchers/models.py#L35-L39

So the question is: how do we fix this? Do we just fix on the backend and pick which of those should be included in `True`, or do we change the API and frontend to choose?

What are your thoughts from the frontend side @nabramow? What's your preference? I'm happy to implement.

nabramow:
I misunderstood this filter. I added it so just removing it again. I will use one of the other existing smoking filters instead for the frontend.

Closes #5711 

**Backend checklist**
- [ ] All tests pass


aapeliv (approved):
Oh I didn't see we had another one already.

PR #5726, opened 2025-03-17
Fix style for signup and cookie banner

nabramow:

- Open an incognito window in the browser (so you have no cookies yet)
- In dev tools - put it in mobile simulator, a larger mobile screen like iPhone XR
- Go to signup page and go til the end
- You should see and be able to scroll to and click the signup button
- You should be able to close the cookie banner

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/CookieBanner.tsx
+++ b/app/web/components/CookieBanner.tsx
@@ -27,7 +27,7 @@ const StyledWrapper = styled("div")(({ theme }) => ({
 
 const StyledCloseButton = styled(IconButton)(({ theme }) => ({
   position: "absolute",
-  top: "50%",
+  top: theme.spacing(4),
   transform: "translateY(-50%)",
   right: theme.spacing(1),
 }));
--- a/app/web/features/auth/signup/Signup.tsx
+++ b/app/web/features/auth/signup/Signup.tsx
@@ -35,7 +35,6 @@ const StyledScrollingContent = styled("div")(({ theme }) => ({
   position: "relative",
   zIndex: 2,
   justifyContent: "center",
-  minHeight: `calc(100vh - ${theme.shape.navPaddingXs})`,
   display: "flex",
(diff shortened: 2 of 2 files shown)


app/web/components/CookieBanner.tsx
```diff
@@ -27,7 +27,7 @@ const StyledWrapper = styled("div")(({ theme }) => ({
 
 const StyledCloseButton = styled(IconButton)(({ theme }) => ({
   position: "absolute",
-  top: "50%",
+  top: theme.spacing(4),
   transform: "translateY(-50%)",
   right: theme.spacing(1),
 }));
```
nabramow, line 33: This is just to move the close x to the top right of the cookie banner as that seems more standard.

app/web/features/auth/signup/Signup.tsx
```diff
@@ -35,7 +35,6 @@ const StyledScrollingContent = styled("div")(({ theme }) => ({
   position: "relative",
   zIndex: 2,
   justifyContent: "center",
-  minHeight: `calc(100vh - ${theme.shape.navPaddingXs})`,
```
nabramow, line 38: This was leftover from before I made style changes to the main AppRoute style. The issue was it conflicted with those changes and caused issues. We can remove it as the height is being handled there now.

aapeliv (approved):
LGTM!

One thought: what if clicking on the "Create an account" button made the cookie banner disappear? It says underneath that clicking on it you accept the cookie policy.

nabramow:
> LGTM!
> 
> One thought: what if clicking on the "Create an account" button made the cookie banner disappear? It says underneath that clicking on it you accept the cookie policy.

I rather keep this to the bug for now. I think the way it is is more standard behavior but we can def talk about it for a future ticket! Let's see if this pushes down the bug reports.

PR #5730, opened 2025-03-19
Fix layout

nabramow:
This is fixing some layout issues that I will need done in the map POC. Breaking it into a separate PR since it effects multiple pages so it's easier to see any bugs.

- Click around and make sure no pages layout looks weird

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -39,7 +39,7 @@ const globalStyles = (
 const PageWrapper = styled(Box)({
   display: "flex",
   flexDirection: "column",
-  flexGrow: 1,
+  flex: 1,
   overflowY: "auto",
 });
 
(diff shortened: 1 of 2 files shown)


app/web/components/AppRoute.tsx
```diff
@@ -39,7 +39,7 @@ const globalStyles = (
 const PageWrapper = styled(Box)({
   display: "flex",
   flexDirection: "column",
-  flexGrow: 1,
+  flex: 1,
   overflowY: "auto",
 });
```
bakeiro, line 45: context? (just to understand and know how to test and what)

nabramow: @bakeiro Sure! So when we did the original app layout in `AppRoute` we weren't setting the main height there. This meant we had to hack around it and set the height in the pages instead. It caused weird scrolling behavior, i.e. scrolling over the navbar, etc.

This caused issues on other pages, so I fixed it a couple months back. But I forgot some things, and noticed while working on the map which has a bit of a complex layout.

So I'm fixing it here! This is just making sure the screen stretches to the full height, so the full-screen components can also have height 100% when specified.

Otherwise map looks like this:

Screenshot 2025-03-19 at 7 49 06 AM


Main thing here is to make sure it didn't mess up other pages. That's why I did it in a separate PR.

I already clicked around myself but just in case I missed something good to have a second set of eyes!

bakeiro: all good!! 👌

bakeiro (approved):
all in order :)

PR #5735, opened 2025-03-20
create timepicker and add to EventTimeChanger

### LINKED ISSUE #5483: Make Timepicker selected time background match Couchers style
The selected time background is a bright blue for the Timepicker, which is not our style:

Image

This should be the green "theme.primary.main" color instead.

This is in the `EventTimeChanger` component.

How you probably want to do this is add special style to the `TextField` component when the  `type` prop is "time".

couchers/app/web/components/TextField.tsx

ColleenLillian:
closes #5483
-
-


ColleenLillian:
definitely still some things to fix! but i'll leave it here to have a look and offer some suggestions :)

app/web/components/Timepicker.tsx
```diff
@@ -0,0 +1,92 @@
+import TextField from "@mui/material/TextField";
+import { TimePicker } from "@mui/x-date-pickers";
+import { useTranslation } from "i18n";
+import { Control, Controller, UseControllerProps } from "react-hook-form";
+import { theme } from "theme";
+import dayjs, { Dayjs } from "utils/dayjs";
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
nabramow: I wouldn't put this rule for the whole file, rather just for that one line with Control. We want Typescript to ban `any` in general. There should be an option to do it just for that line.

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -52,44 +51,13 @@ export default function EventTimeChanger({
...
-      return;
-    }
-    const newStartTime = dayjs(e.target.value, TIME_FORMAT);
-
-    setValue("startTime", newStartTime.format(TIME_FORMAT), {
-      shouldDirty: true,
-      shouldValidate: true,
-    });
-  };
-
```
nabramow: I think you might still need to `setValue` for the new Timepicker start and end times. I assume that function is passing it up to the parent form component? Similar to what the Datepicker is doing, it has the Controller, but is still setting the value.

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -153,16 +124,17 @@ export default function EventTimeChanger({
                 t("communities:past_time_error")
               );
             },
-          })}
-          defaultValue={eventStartTime || null}
```
nabramow, line 157: Is defaultValue being passed in the field now?

For example if you go to the `EditEvent` component (where this component is also used) and it already has a default start and end time, are you seeing them passed through when you go to edit?

ColleenLillian: ok i'm not sure! I will have another look next week, I'm back on Thursday :)

ColleenLillian:
The timepicker works again and pops up in the correct position. No console or lint warnings.
Not sure about the default value issue but it works locally for me now

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -124,9 +92,12 @@ export default function EventTimeChanger({
...
         />
-        <TextField
-          id="startTime"
-          {...register("startTime", {
+
+        <Controller
+          control={control}
+          defaultValue={eventStartTime || ""}
+          name="startTime"
+          rules={{
```
nabramow: Okay taking a closer look at this now!

So I see here you have wrapped this in a controller and are also passing `control` and wrapping the Timepicker in a Controller again in the Timepicker component.

You only need one, otherwse you're doubling it which might cause some issues.

I'd recommend removing this Controller and just pass Timepicker, since you already have a Controller within Timepicker. You can pass the rules to the Timepicker component in `EventTimeChanger` instead.

nabramow:
Once you remove the double controller, you can make these changes to get the values passing:

In the function `splitTimestampDateAndTime` change time to `time: dayjsDate,`.

Since it was previously a `TextField` we converted it to a string, but since we're using a Timepicker now, it can be a dayjs date.

Then for `defaultValue` in the startTime and endTime, you can put `eventStartTime || null` and `eventEndTime || null`.

Then you can change the type in `EventForm` for startTime and endTime to Dayjs to match the startDate and endDate since they are no longer strings.

nabramow:
When you change the endTime to be before the startTime, it should immediately show a red error warning `The event end time must be after the start time.`.

I think this is not showing as you remove the `handleStartTimeChange` and `handleEndTimeChange` functions. You can add a simplified version of these back.

For example just this, similar now to the startDate and endDate ones:

```javascript
  const handleStartTimeChange = (newStartTime: Dayjs) => {
    setValue("startTime", newStartTime, {
      shouldDirty: true,
      shouldValidate: true,
    });
  };
```

I think once all these changes are made you should be good to go! It's close just some adjustments for it being a date now!

nabramow:
Okay I think I have an idea about the weird test failing in `EventTimeChanger`! Two things:

1. Add the `TimePicker` to the mock at the top of the file as well (this forces it to use the desktop version which allows typing in the time, otherwise it doesn't let you type and fails:

```javascript
jest.mock("@mui/x-date-pickers", () => {
  return {
    ...jest.requireActual("@mui/x-date-pickers"),
    DatePicker: jest.requireActual("@mui/x-date-pickers").DesktopDatePicker,
  };
});
```

2. There seems to be a weird quirk with the MUI Timepicker that it adds a space after the time values. More into here:
https://github.com/mui/mui-x/issues/9410

If you add a space after the expect values for the failing tests related to times, they should pass. 

For example if you `it.only` this test you can try it in one of them at a time (then remove the only at the end). These are just the changes lines as an example:

```javascript
  it.only("should make user manually fix times when the start date/time updates", async () => {
      await waitFor(() => expect(startTime).toHaveValue("00:00 ")); // instead of "00:00", add an extra space

    const endTime = screen.getByLabelText(t("communities:end_time"));

    // Increases time difference between start and end time to 3 hours
    user.type(endTime, "0300");

    await waitFor(() => expect(endTime).toHaveValue("03:00 ")); // same here
    
    //etc, etc.
```

That should do it I think!

nabramow:
Sorry to bring the bearer of bad news haha. Found this error. To replicate:

- Go to Create an Event (on fresh page load)
- Add title and start date
- Add start time
- Add an end time before the start time
- Throws error

I think you need to check if `!endDate` first like you're doing with the startTime.

Then you can probably remove the "start of day" logic from the `endDate` variable as you want it to compare the exact endDate combined with endTime and if it's before or after the startDate and time.

app/web/components/Timepicker.tsx
```diff
@@ -0,0 +1,112 @@
...
+      id,
+      rules,
+      label,
+      name,
+      onPostChange,
+      testId,
+    }: TimepickerProps,
+    ref,
+  ) => {
+    const { t } = useTranslation();
```
nabramow:
```suggestion
    const { t } = useTranslation([GLOBAL]);
```

app/web/components/Timepicker.tsx
```diff
@@ -0,0 +1,112 @@
...
+                  }
+                  data-testid={testId}
+                  InputLabelProps={{
+                    shrink: true,
+                  }}
+                  InputProps={{
+                    ...props.InputProps,
+                    className,
+                    inputRef: ref || inputRef,
+                    "aria-label": t("components.timepicker.change_time"),
```
nabramow:
```suggestion
                    "aria-label": t("global:change_time"),
```

nabramow: I think you want one that just says "Change time"? If you so that's the one in this file:

app/web/resources/locales/en.json

nabramow:
Okay just a couuuuuple more things, so close!

Getting this error for `Timepicker`:

Screenshot 2025-04-13 at 10 38 15 AM

I believe you just need to remove the `defaultValue` prop for the TextField on line 68.


Second the `Timepicker` shouldn't need the manual ref behavior and the `forwardRef` anymore since the Controller is handling that and passing the ref in the field.

You should be able to:

- remove the `forwardRef` wrapper function from the Timepicker
- remove the `inputRef` and `anchorEl` variables and their references

Any questions just shout!

ColleenLillian:
Not getting the forward ref console error any more when I remove it so it must have been related to something else :)

app/web/components/Timepicker.tsx
```diff
@@ -0,0 +1,93 @@
...
+      control={control}
+      defaultValue={defaultValue}
+      name={name}
+      rules={rules}
+      render={({ field }) => (
+        <TimePicker
+          data-testid={testId}
+          {...field}
+          label={label}
+          value={field.value ? dayjs(field.value) : null}
```
nabramow: I think this can just be `field.value`!

nabramow (dismissed):
Just change that one thing about `field-value` and then go ahead and merge!

It's working great for me, all the proper errors showing, can create and edit an event with times successfully. 

Well done!

nabramow (approved):
All good to merge! 🚀

PR #5773, opened 2025-04-01
Upgrade protobuf to v6.30.2

aapeliv:
Continued from #5768 and #5764


aapeliv:
OK apparently some deps aren't ready for this.

PR #5785, opened 2025-04-06
Return exactly a given set of user_ids in search

aapeliv:


nabramow (approved):
Looks good to me! Thanks for doing this!

PR #5817, opened 2025-04-14
Automatically upload image without confirm

### LINKED ISSUE #5804: Remove check/x when uploading profile photo
Users don't see the check/ X to confirm or remove an uploaded photo. They hit save and then think it's not working.

Let's just remove the check / X. It should automatically just upload the photo.

Acceptance Criteria

- Remove the check / X behavior when uploading a photo and automatically just run the current check behavior
- The Save button should be disabled with a `loading` prop button spinner until the upload is done.

Image

nabramow:
Closes #5804 

- Go to profile and click "edit"
- Click image, upload a new image
- Should show spinner while uploading, then see image uploaded
- Click save
- New image should be uploaded

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/components/ImageInput/ImageInput.test.tsx
+++ b/app/web/components/ImageInput/ImageInput.test.tsx
@@ -1,12 +1,9 @@
 import { render, screen, waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import {
-  CANCEL_UPLOAD,
-  CONFIRM_UPLOAD,
   COULDNT_READ_FILE,
   getAvatarLabel,
   SELECT_AN_IMAGE,
-  UPLOAD_PENDING_ERROR,
 } from "components/constants";
 import { StatusCode } from "grpc-web";
 import { InitiateMediaUploadRes } from "proto/api_pb";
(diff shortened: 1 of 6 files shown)


jesseallhands:
I tested this on desktop and encountered the following:

1. I tried uploading a 2.9MB photo (which is pretty small by today's smartphone's standards) and it took over 30 seconds to upload. This happened on desktop. The upload time might be a backend issue, but if it's going to take that long in some cases, could we implement some kind of a progress bar? Or is that a lot of work?

2. The following issue I only encountered on **_desktop_**: I was able to navigate from the page while the uploading was happening without getting a warning message. We need this warning message because lots of users are overlooking the save button at the bottom, navigating away without saving, and then complaining that "the couchers system didn't save my profile changes!".  There have been a bunch of tickets about this before we implemented the warning. Here is the warning I am referring to:
image

3. Currently, if you make any changes while editing your profile and try to navigate away, you get the warning (pictured above) telling you your changes will not be saved. However, this is not the case with photos. If you upload a photo and try to navigate away without saving (eg. maybe you decide against the photo) then it still saves the change to the photo despite not having clicked "save".  In fact, even if you interrupt the upload process by navigating away before upload is complete (spinner is still spinning) the photo still somehow uploads in the background (I was able to replicate this on both mobile and desktop).

jesseallhands (approved):
Tested and working as expected!

PR #5820, opened 2025-04-15
Add link to new blog on dashboard

### LINKED ISSUE #5806: Update dashboard to include text and link to announce new blog post
To highlight the new blog post, we should add a text and link to the dashboard as appears below. The word "here" is not pictured with a link in the screenshot below, but the word "here" should be a link directly to the new blog post:
https://couchers.org/blog/2025/04/12/50k-couchers-milestone-reached

Image

Note: This is a temporary solution for blog posts until #5686  is finished, where we can include blog post announcements.

ColleenLillian:
closes #5806 

just a temp fix so it ain't fancy

- [ x] There are no warnings from `yarn lint --fix`
- [ x] There are no console warnings when running the app
- [ ] All tests pass
- [ x] Clicked around my changes running locally and it works

### DIFF
--- a/app/web/dashboardBlog.json
+++ b/app/web/dashboardBlog.json
@@ -1,3 +1,0 @@
-{
-  "2025-04-12": "50k Couchers Milestone Reached!"
-}
--- a/app/web/dashboardBlog.json
+++ b/app/web/dashboardBlog.json
@@ -1,0 +1,3 @@
+{
+  "2025-04-12": "50k Couchers Milestone Reached!"
+}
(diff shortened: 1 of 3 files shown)


app/web/dashboardBlog.json
```diff
@@ -0,0 +1,3 @@
+{
+  "2025-04-12": "50k Couchers Milestone Reached!"
+}
```
nabramow: What's this for? Curious why a new one rather than it going together with `dashboardNews`?

nabramow:
@ColleenLillian Just a style comment. The two bold fonts in a row looks a lil weird to me. What if we do this?

"New Blog Post! 50k Couchers Milestone Reached (Read here)!" (and link the "Read here")

Screenshot 2025-04-15 at 12 08 35 PM

ColleenLillian:
We already had a styled link component with the style Jesse suggested so I've used that now

nabramow (approved):
Looks good to me!

Issue #5177, opened 2024-11-14
Update sort order of communities list on dashboard


jesseallhands:
I am not sure if this issue can/should be handled on the frontend or backend.

### Problem:
Communities listed on the dashboard are ranked from biggest to smallest rather than from smallest to biggest, with no consideration of the person's home community.  For example, instead of me seeing my "home town" (Berlin) first, I see the following:

![image](https://github.com/user-attachments/assets/75e87455-99a9-41a8-b0e2-e9b22745ae65)

### Solution:
List the user's "home communities" at the top (with the children at the top, parent communities at the bottom), then the rest in alphabetical order. 

The "home communities" are those that the user's geo.json location falls within. The user is automatically added to these communities if their geo.json coordinate falls within the polygon of the community when their account is created (or the community is created).

This is somewhat related to the sorting problem on the "browse communities" feature (issue #5124 )

aapeliv:
Dupe of #1471, which I just closed as not a priority :). But if you think it's a priority, happy to keep it!

nabramow:
Closed by #5828

PR #5828, opened 2025-04-16
sort community names alphabetically on "browse communities" page

### LINKED ISSUE #5124: sort community names alphabetically on "browse communities" page
**Problem**
When using the "browse communities" link on the dashboard, communities are currently sorted in the order they were created. This makes communities hard to find. See screenshot below:

image

We do plan to implement a text search feature and refactor the "browse communities" page, but we need a temporary fix until then.

**Solution**
Sort communities alphabetically in the list pictured above.

I'm not sure if this should be handled on the frontend or backend (probably BE?)

Note: The sorting of the "Your communities" section of the dashboard is also messed up. That's a separate issue ( #5177 ), but somewhat related to this issue.

### LINKED ISSUE #5807: Add pageNumber pagination and option to order by created_at to ListMembers query
We want to be able to use pagination by pageNumber like for events for the communities "Members" tab.

We also want to be able to pass an order from the frontend to the backend to sort with newest members up top as well.

Acceptance Criteria
- Add `total_items` to the response for `ListMembers` for communities. It should return the total number of items for that query regardless of pagination (so we know how many pages)
- Add `page_number` with an `offset` (visually we show page number 1 as first but it's really 0 logic-wise) to the query logic in `ListMembers`
- You can use `ListMyEvents` query as an example
- Add an option to request for `ListMembers` to sort the list by `created_at` `descending` (so newest members show first). Come up with a name that makes it clear `show_new__members_first` or something?
- Add backend tests

pcolt:
It closes #5124.
At the end I am not sure if it was a good idea to sort the communities by name in the backend instead of in the frontend. But if we see it is needed we can  always add an index on the name attribute later on.
Regarding the somehow related issue #5177 I leave it for another time because the intended ordering logic is more complex and it requires work in the frontend.

- no particular steps needed

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/communities.py
+++ b/app/backend/src/couchers/servicers/communities.py
@@ -404,7 +404,7 @@ class Communities(communities_pb2_grpc.CommunitiesServicer):
 
     def ListUserCommunities(self, request, context, session):
         page_size = min(MAX_PAGINATION_LENGTH, request.page_size or MAX_PAGINATION_LENGTH)
-        offset = int(request.page_token) if request.page_token else 0
+        next_node_id = int(request.page_token) if request.page_token else 0
         user_id = request.user_id or context.user_id
         nodes = (
             session.execute(
@@ -413,9 +413,9 @@ class Communities(communities_pb2_grpc.CommunitiesServicer):
                 .join(ClusterSubscription, ClusterSubscription.cluster_id == Cluster.id)
                 .where(ClusterSubscription.user_id == user_id)
                 .where(Cluster.is_official_cluster)
-                .order_by(Cluster.name)
+                .where(Node.id >= next_node_id)
+                .order_by(Cluster.id)
                 .limit(page_size + 1)
-                .offset(offset)
             )
             .scalars()
             .all()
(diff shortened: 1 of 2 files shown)


jesseallhands (dismissed):
Appears to be working as intended, though I couldn't test too thoroughly because there are so few communities in the test environment.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -411,7 +414,7 @@ def ListUserCommunities(self, request, context, session):
                 .where(ClusterSubscription.user_id == user_id)
                 .where(Cluster.is_official_cluster)
                 .where(Node.id >= next_node_id)
-                .order_by(Node.id)
+                .order_by(Cluster.id)
```
aapeliv: This should be reverted, right?

aapeliv (changes requested):
Looks good, just minor comment

app/backend/src/couchers/servicers/communities.py
pcolt: Oh sure, my apologies

pcolt: Just reverted)

aapeliv (approved):
LGTM!

PR #5830, opened 2025-04-16
Implement notifications seen

### LINKED ISSUE #5823: Backend - add notification `isSeen`
Rather than replying on cookies, we want to keep track of if a user saw a notification on the backend instead.

Acceptance Criteria
- Add `is_seen: boolean` to each notification returned by `ListNotifications` that shows whether that notification was seed by the current user.
- Determine what counts as `is_seen`
- Add functionality for when user clicks "mark all as seen"
- How can frontend see if there are any new notifications (do we do this on backend or frontend)?
- Pass friendRequestId for the accept/decline functionality like this:

Image

aapeliv:
Closes #5823 

- Unit tests should suffice for this one as it's thoroughly tested

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -2458,6 +2458,9 @@ class Notification(Base):
 
     data = Column(Binary, nullable=False)
 
+    # whether the user has marked this notification as seen or not
+    seen = Column(Boolean, nullable=False, server_default=text("false"))
```
nabramow: Nit - what about `is_seen` to indicate it's a boolean?

app/proto/api.proto
```diff
@@ -113,6 +113,8 @@ message PingRes {
   uint32 unseen_sent_host_request_count = 5;
   uint32 unseen_received_host_request_count = 6;
   uint32 pending_friend_request_count = 4;
+
+  uint32 unseen_notification_count = 7;
```
nabramow, line 117: Is the idea that I will check `if unseedNotificationCount = 0` don't show red badge?

nabramow:
@aapeliv any chance you can add the `friendRequestId` to the friend_request sent notification? (This is for the "Accept" or "Decline" functionality in the feed)? Or you rather we do this later after beta?

Checking my understanding for the frontend - is the idea that when they click on a notification I will call the backend to mark it as seen? Or when they open the notification feed or scroll I mark certain types as seen? Some of them don't really have clicking actions, like "your password was reset" so trying to brainstorm the logic here.

app/proto/api.proto
aapeliv: Yes if that works for you, better than counting on the frontend, right?

aapeliv:
> @aapeliv any chance you can add the `friendRequestId` to the friend_request sent notification? (This is for the "Accept" or "Decline" functionality in the feed)? Or you rather we do this later after beta?

I was hoping to keep it general: we could implement "actions" that basically render as buttons. That way the frontend would be cleaner and we can easier introduce new notification types?


> Checking my understanding for the frontend - is the idea that when they click on a notification I will call the backend to mark it as seen? Or when they open the notification feed or scroll I mark certain types as seen? Some of them don't really have clicking actions, like "your password was reset" so trying to brainstorm the logic here.

I'm not sure. For now I think we can just have the "mark all as read" button and leave it at that?

nabramow (approved):
Yaaaaay thanks for doing this! 🚀

PR #5839, opened 2025-04-18
Mark host and surf requests past now expired

### LINKED ISSUE #5515: Adjust text for expired host/surf request in messages
Right now a request in the past will still say "pending". This is a bit confusing. We want to check the event date on the frontend and see if it's in the past.

Acceptance Criteria
- In messages list, where it says "This request is pending" change it to "This request is expired" if it's in the past (and make it color red)
- In the 1:1 message thread in the header bar, change "Pending" to "Expired"
- For messages that are expired, change the "..." icon to Material Ui History clock icon: https://mui.com/material-ui/material-icons/?query=clock&selected=History

**Hint:** You can test this by making a second account in Incognito window and host requesting yourself. Or manually go in and change the conditional props in React.

How it looks now:

Image
Image

nabramow:
Closes #5515

Previously expired requests still said "confirmed" and it's confusing. We want to check the date and mark as expired if the endDate is past now.

- Go to messages
- Check host and surf request that are expired (send yourself one from a second account or ask someone to request you on stage)
- It should say "this request is expired" for pending requests in the past
- For past requests it should say "(Past)" at the end with the normal status
- In the message it should say "(Past)" at the top
- Past request items should be dark grey entirely and have a history icon

Screenshot 2025-04-24 at 7 29 05 PM

Screenshot 2025-04-24 at 7 29 10 PM

**Web frontend checklist**
- [ x ] All tests pass


jesseallhands:
Not sure if this is ready for review yet, but I tested it and it appears to be working properly! Looks good!

I did discover a minor, unrelated issue while testing and created a separate issue for it here: #5841

aapeliv:
Would it be possible to have both the request status and the fact that it's in the past in the overview? The red makes it look a bit sad; and I'm now not able to see who I've hosted and who I haven't, which were useful indicators.

Also I think it should say "This request is in the past" instead of "expired": the latter sounds like something is wrong or action was not taken.

jesseallhands:
> Would it be possible to have both the request status and the fact that it's in the past in the overview? The red makes it look a bit sad; and I'm now not able to see who I've hosted and who I haven't, which were useful indicators.
> 
> Also I think it should say "This request is in the past" instead of "expired": the latter sounds like something is wrong or action was not taken.

I agree, but im not sure if it should be in a different issue or not. UI/UC could definitely could do some mockups. But in general I agree that we should include the status, followed by if expired or not. 

So basically, we have the following:

For active requests:
* ✅ Accepted
* ❌ Declined
* ❓Pending

For past requests:
* Accepted (✅ 🔄past)
* Declined (❌ 🔄past)
* Ghosted i.e. unanswered (⁉️🔄 Expired!)

I like the MUI icons chosen thus far, the above are just placeholders.

nabramow:
@jesseallhands @aapeliv I've gone ahead and made the changes we talked about now. Let me know what you think!

aapeliv (dismissed):
LGTM.

My preference would be to keep the original icon even in past reqs (so one can see easily/visually if it was declined/confirmed/pending even when inpast).

nabramow:
> LGTM.
> 
> My preference would be to keep the original icon even in past reqs (so one can see easily/visually if it was declined/confirmed/pending even when inpast).

Okay as a happy medium I changed it back to the original icons, but when it's in the past, I made the color grey to match the past grey of everything else.

Screenshot 2025-04-26 at 10 54 27 AM

aapeliv (approved):
Thanks!

PR #5842, opened 2025-04-20
Add only_unread to ListNotifications

aapeliv:


app/backend/src/couchers/servicers/notifications.py
```diff
@@ -83,6 +84,7 @@ def ListNotifications(self, request, context, session):
                 select(Notification)
                 .where(Notification.user_id == context.user_id)
                 .where(Notification.id <= next_notification_id)
+                .where(or_(request.only_unread == False, Notification.is_seen == False))
```
nabramow, line 87: What happens here if `only_unread` is undefined?

aapeliv: Proto3 does not have undefined, it's `false` by default (and only `true` if it's explicitly set to `true`).

nabramow (approved):

PR #5862, opened 2025-04-25
Frontend: Stripe billing portal

aapeliv:
Also fixed some translation string issues. Is there no better way than numbered placeholders in i18n, it seems so brittle.

- Go to donations page (/donate): make sure the link in the donation box works
- Go to settings -- make sure manage donation button works

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/service/donations.ts
+++ b/app/web/service/donations.ts
@@ -1,3 +1,4 @@
+import { Empty } from "google-protobuf/google/protobuf/empty_pb";
 import { InitiateDonationReq } from "proto/donations_pb";
 
 import client from "./client";
@@ -10,4 +11,9 @@ export async function initiateDonation(amount: number, recurring: boolean) {
 
   const res = await client.donations.initiateDonation(req);
   return res.getStripeCheckoutSessionId();
+}
+
+export async function getDonationPortalLink() {
+  const res = await client.donations.getDonationPortalLink(new Empty());
+  return res.getStripePortalUrl();
 }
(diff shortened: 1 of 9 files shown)


app/web/service/donations.ts
```diff
@@ -11,3 +12,8 @@ export async function initiateDonation(amount: number, recurring: boolean) {
   const res = await client.donations.initiateDonation(req);
   return res.getStripeCheckoutSessionId();
 }
+
+export async function getDonationPortalLink() {
+  const res = await client.donations.getDonationPortalLink(new Empty());
+  return res.getStripePortalUrl();
```
nabramow, line 18: We normally define the req and then pass that to the res no?

aapeliv: You mean `const req = new Empty();`? I'm seeing both when searching for `new Empty()` in `//app/web`

nabramow:
When I go to account settings and then click "go to donations management portal" nothing happens.

I think you need to await the route in a variable and then pass it to the `router.push` maybe?

nabramow:
When I go to the `/donate` page and click the donation portal link, it throws this error:

Screenshot 2025-04-25 at 6 29 46 PM

Probably the same thing where you need to wait it first and then pass the value.

app/web/service/donations.ts
nabramow: Oh I see now there is no request type in the protos.

nabramow:
Ahhh wait never mind, it looks like I needed to generate the protos. Seems to be working now! Just write that in the instructions next time if you can.

nabramow:
Okay works well! Only thing I see is when I load the `/donate` page I'm getting these stripe errors in the console:

Screenshot 2025-04-25 at 6 38 33 PM


Might just be because localhost is http and not https but just in case...

aapeliv:
I think it's just localhost stuff

nabramow (approved):

PR #5883, opened 2025-04-30
Fix and simplify backend debugging with vscode

### LINKED ISSUE #5880: Backend debugging using Dev Containers extension for VS Code
Backend debugging using Dev Containers extension for VS Code appears to not be working anymore. 

It looks that the problem is caused by the missing _customizations_ and nested _vscode_ keys in the `devcontainer.json` file as it is clearly stated in the [official documentation](https://containers.dev/supporting#visual-studio-code). 

This change may be happened because of a newer feature introduced in Dev Containers after late 2022. In older versions, these settings were at the root level of the configuration (this explanation is coming from my AI assistant based on model Claude 3.5 Sonnet).

I have also simplified and clarified how things are set up by updating the documentation file and by adding a default `launch.json` file needed to start the backend in debugging mode right away.

#1838 appears to be a related issue to this topic.

pcolt:
It closes #5880

VS Code Ide is needed to check this out. Also please follow the instructions updated in the `/docs/backend-in-vscode.md` file.

**Backend checklist**
- [x] All tests pass

### DIFF
--- a/app/.devcontainer/devcontainer.json
+++ b/app/.devcontainer/devcontainer.json
@@ -19,16 +19,23 @@
 	// connected. This is typically a file mount in .devcontainer/docker-compose.yml
 	"workspaceFolder": "/workspace",
 
-	// Set *default* container specific settings.json values on container create.
-	"settings": {
-		"python.defaultInterpreterPath": "/workspace/venv/bin/python3.13",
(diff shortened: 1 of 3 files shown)


aapeliv (approved):
LGTM! Feel free to merge.

PR #5884, opened 2025-04-30
User summary long text

kimalena9:

-
-
-

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/components/SafeLinesEllipsis.tsx
+++ b/app/web/components/SafeLinesEllipsis.tsx
@@ -1,0 +1,51 @@
+import React, { useEffect, useState } from "react";
+
+interface SafeLinesEllipsisProps {
+  text: string;
+  maxLine?: number;
+  ellipsis?: string;
+  trimRight?: boolean;
+  basedOn?: string;
+}
+
+const SafeLinesEllipsis: React.FC<SafeLinesEllipsisProps> = ({
+  text,
+  maxLine = 1,
+  ellipsis = "...",
+  trimRight = true,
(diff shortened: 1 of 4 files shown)


nabramow:
Okay just took a quick cursory look!

This is a solid attempt but I think it can be waaaaay simpler.

What I'd suggest:

- For name, you have a conditional value like this ` {nameOnly ? user.name : ${user.name}, ${user.age}}` right now.
- You can set this to a variable like `nameValue` above the title variable.
- Then check the length, so `nameValue.length > 20`
- If it's greater than 20, slice nameValue to be 20 characters and add a `...` to it
- Otherwise just return nameValue
- Do the same for user location (called `city` here I think) - but that one is a 45 char limit
- Import the `Tooltip` component from `@mui/material` and wrap the value(s) in that with the text at the full value so it shows on Desktop on hover

Alternatively I think you could wrap the name conditional in `LinesEllipsis` set to `maxLines={1}` and the location to max 2 lines.

That would look a bit like this:

```typescript
        <>
        <LinesEllipsis
          maxLine={1}
          text= {nameOnly ? user.name : `${user.name}, ${user.age}`}
         />
          {user.hasStrongVerification ? <StrongVerificationBadge /> : null}
        </>
```

I did not test either of these myself though so you'd need to check on localhost and see if it looks okay for the long values!

kimalena9:
Replaced by #6003 and hit 'Close pull request"?

Issue #5776, opened 2025-04-01
Add pagination to UserDetails component


nabramow:
In bug report #5762: 

"On community tab i've clicked on member link and then i've clicked on load. When i click on load for the 5th time it gave me an error saying we cant show that many people, so i couldnt see the whole list in the end."

Right now UserDetals component loads ALL the members of the community at once. If there are a lot it throws an error.

Acceptance Criteria
- Add pagination to the list in the UserDetails component to only load, say, 50 or 100 members at a time.

nabramow:
This is done for the purpose of V1. Communities member list is fixed. If there's issues in other areas we should make separate tickets.

PR #5899, opened 2025-05-02
Mod tools + new users list

### LINKED ISSUE #5511: Moderation tooling (console): “New users” feed
We want to add a "New users" feed to the console (different codebase than the main app) that allows admins to see a feed of new users. This helps identify anything suspicious, i.e. robo accounts being rapidly created, banned users making new accounts, etc.

aapeliv:
Closes #5511

- Be a mod (so test against local with `is_superuser = 1` on that acct
- Go to `/mod/`: you should see a list of users, ordered by recently signed up
- Go to `/mod/user/2`: you should see the admin console

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/mod/ModPage.tsx
+++ b/app/web/features/mod/ModPage.tsx
@@ -1,4 +1,4 @@
-import { Typography, useTheme } from "@mui/material";
+import { Typography } from "@mui/material";
 import Button from "components/Button";
 import HtmlMeta from "components/HtmlMeta";
 import PageTitle from "components/PageTitle";
@@ -10,7 +10,6 @@ import { useNewUsers } from "./hooks";
 
 export default function ModPage() {
   const { t } = useTranslation([GLOBAL, MOD, PROFILE]);
-  const theme = useTheme();
 
   const { userIds, error, hasNextPage, isFetchingNextPage, fetchNextPage } =
     useNewUsers();
--- a/app/web/features/mod/ModPage.tsx
+++ b/app/web/features/mod/ModPage.tsx
@@ -1,0 +1,39 @@
+import { Typography, useTheme } from "@mui/material";
+import Button from "components/Button";
+import HtmlMeta from "components/HtmlMeta";
+import PageTitle from "components/PageTitle";
+import UsersList from "components/UsersList";
+import { useTranslation } from "i18n";
+import { GLOBAL, MOD, PROFILE } from "i18n/namespaces";
+
+import { useNewUsers } from "./hooks";
+
+export default function ModPage() {
+  const { t } = useTranslation([GLOBAL, MOD, PROFILE]);
+  const theme = useTheme();
+
+  const { userIds, error, hasNextPage, isFetchingNextPage, fetchNextPage } =
+    useNewUsers();
+
+  return (
+    <>
+      <HtmlMeta title={t("mod:title")} />
+      <PageTitle>{t("mod:title")}</PageTitle>
+      <h1>{t("mod:list_users.heading")}</h1>
+      <UsersList
+        error={error}
+        userIds={userIds}
+        emptyListChildren={
+          <Typography variant="body1">
+            {t("mod:list_users.none_found")}
+          </Typography>
+        }
+      />
+      {hasNextPage && (
+        <Button loading={isFetchingNextPage} onClick={() => fetchNextPage()}>
+          {t("mod:list_users.load_more")}
+        </Button>
+      )}
+    </>
+  );
+}
--- a/app/web/features/mod/hooks.ts
+++ b/app/web/features/mod/hooks.ts
@@ -26,7 +26,6 @@ export const useNewUsers = () => {
     },
   );
   const userIds = query.data?.pages.flatMap((page) => page.userIdsList);
-  console.log(userIds);
   return { ...query, userIds };
 };
 
--- a/app/web/features/mod/hooks.ts
+++ b/app/web/features/mod/hooks.ts
@@ -1,0 +1,24 @@
+import { newUsersListKey } from "features/queryKeys";
+import { RpcError } from "grpc-web";
+import { ListUserIdsRes } from "proto/admin_pb";
+import { useInfiniteQuery } from "react-query";
+import { service } from "service";
+
+export const useNewUsers = () => {
+  const query = useInfiniteQuery<ListUserIdsRes.AsObject, RpcError>(
+    newUsersListKey,
+    ({ pageParam }) =>
+      service.admin.listUserIds({
+        startTime: new Date(1970, 1, 1),
+        endTime: new Date(),
+        pageToken: pageParam,
+      }),
+    {
+      getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
+      refetchInterval: 60_000,
+    },
+  );
+  const userIds = query.data?.pages.flatMap((page) => page.userIdsList);
+  console.log(userIds);
+  return { ...query, userIds };
+};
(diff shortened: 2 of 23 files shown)


app/web/features/mod/hooks.ts
```diff
@@ -0,0 +1,67 @@
...
+        pageSize: 50,
+        pageToken: pageParam,
+      }),
+    {
+      getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
+      refetchInterval: 60_000,
+    },
+  );
+  const userIds = query.data?.pages.flatMap((page) => page.userIdsList);
+  console.log(userIds);
```
nabramow: Leftover console.log ;-)

app/web/features/profile/view/UserCard.tsx
```diff
@@ -54,6 +56,7 @@ export default function UserCard({
         <TabPanel classes={{ root: classes.tabPanel }} value="about">
           <About user={user} />
         </TabPanel>
+        {modPanel}
```
nabramow, line 82: We use this component all over do we really want to already render it? Should there be some check if they are a mod just to be safe?

app/web/features/profile/view/AdminPanelUserButton.tsx
```diff
@@ -25,6 +25,7 @@ export default function AdminPanelUserButton({
           aria-label={t("view_in_admin_console")}
           onClick={() => router.push(adminPanelUserLink(username))}
           color="primary"
+          size="large"
```
nabramow: Do we want this to be large everywhere? It seems it's used in a couple places.

app/web/service/admin.ts
```diff
@@ -0,0 +1,49 @@
...
+} from "proto/admin_pb";
+import { User } from "proto/api_pb";
+
+import client from "./client";
+
+export async function getUser(user: string): Promise<User.AsObject> {
+  const req = new GetUserReq();
+  req.setUser(user || "");
+  return (await client.admin.getUser(req)).toObject();
+}
```
nabramow: Don't we already have a `useUser` hook?

nabramow:
Maybe add the `isProfileLink` prop so it can also be a link and you can click on the name?

Screenshot 2025-05-11 at 6 59 40 PM

nabramow:
It looks like this is addressing #5776 too?

nabramow:
I tried to go to the mod page from a non-mod account and I see this:

Screenshot 2025-05-11 at 7 06 44 PM

While it's not showing the data, probably better id non-mods can't see this at all. Why don't we just redirect them to the dashboard if they're not a mod?

One way would be to add something like this:

```javascript
ModPage.getLayout = appGetLayout({ isModOnly: true});
```

and then handling it in `AppRoute` similar to how we're handling `isPrivate`.

app/web/features/mod/ModPage.tsx
```diff
@@ -0,0 +1,39 @@
...
+import PageTitle from "components/PageTitle";
+import UsersList from "components/UsersList";
+import { useTranslation } from "i18n";
+import { GLOBAL, MOD, PROFILE } from "i18n/namespaces";
+
+import { useNewUsers } from "./hooks";
+
+export default function ModPage() {
+  const { t } = useTranslation([GLOBAL, MOD, PROFILE]);
+  const theme = useTheme();
```
nabramow: I don't think you're using this

nabramow:
Okay this works well for me! I tested from both a mod and a non-mod account. Just some smaller comments, assign that other ticket to yourself if it applies and looks like you need to run `yarn lint` and fix some things and address the merge conflicts then you're good to go!

Maybe @jesseallhands should have a look too since he'll use it the most?

jesseallhands:
I don't have a moderation account on there. Can someone make jesse@couchers.org a moderator?

app/web/features/profile/view/UserCard.tsx
aapeliv: If you aren't in the mod view, then `modPanel` is just undefined. It's only when `UserCard` is used in `ModUserPage` that this is something.

app/web/service/admin.ts
aapeliv: This uses `admin.getUser` which bypasses banned etc

aapeliv:
> It looks like this is addressing #5776 too?

It's not. Though I will maybe give that a try in a separate PR

app/web/features/mod/ModPage.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <PageTitle>{t("mod:title")}</PageTitle>
+      <h1>{t("mod:list_users.heading")}</h1>
+      <UsersList
+        error={error}
+        userIds={userIds}
+        emptyListChildren={
+          <Typography variant="body1">
+            {t("mod:list_users.none_found")}
+          </Typography>
+        }
```
nabramow, line 29: Rather than pass a node, I'd just remove this prop (make it conditional) and check if it's undefined within `UsersList` then show the Typography no users found there. That keeps that component responsible for its own logic.

app/web/service/admin.ts
```diff
@@ -0,0 +1,49 @@
...
+  ListUserIdsReq,
+  UserDetails,
+} from "proto/admin_pb";
+import { User } from "proto/api_pb";
+
+import client from "./client";
+
+export async function getUser(user: string): Promise<User.AsObject> {
+  const req = new GetUserReq();
+  req.setUser(user || "");
```
nabramow:
```suggestion
  if (user) {
    req.setUser(user);
  }
```

app/web/service/admin.ts
```diff
@@ -0,0 +1,49 @@
...
+  return (await client.admin.getUser(req)).toObject();
+}
+
+export async function getUserDetails(
+  user: string,
+): Promise<UserDetails.AsObject> {
+  const req = new GetUserDetailsReq();
+  req.setUser(user || "");
```
nabramow:
```suggestion
  if (user) {
    req.setUser(user);
  }
```

app/web/features/mod/hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+    ({ pageParam }) =>
+      service.admin.listUserIds({
+        startTime: new Date(1970, 0, 0, 0, 0, 1),
+        endTime: new Date(2050, 0, 0),
+        pageSize: 50,
+        pageToken: pageParam,
+      }),
+    {
+      getNextPageParam: (lastPage) => lastPage.nextPageToken || undefined,
+      refetchInterval: 60_000,
```
nabramow, line 25: Should it have the underscore? The docs say the type is just `number` and that doesn't seem very javascripty?

aapeliv: It's just for clarity, it's called "numeric separators" I think

app/web/features/mod/ModPage.tsx
aapeliv: There is a minor difference in behavior though: with `emptyListChildren`, it will show an empty list if the users are blocked out for some reason. E.g. in your blocking PR, if `userIds` consisted of exactly one blocked user, then this would display the `emptyListChildren` node (because it runs it through `useLiteUsers`).

aapeliv:
> While it's not showing the data, probably better id non-mods can't see this at all. Why don't we just redirect them to the dashboard if they're not a mod?
> 
> One way would be to add something like this:
> 
> ```js
> ModPage.getLayout = appGetLayout({ isModOnly: true});
> ```
> 
> and then handling it in `AppRoute` similar to how we're handling `isPrivate`.

OK, I got halfway there, then I realized we don't hand out info about whether the user is mod or not where we otherwise do auth stuff. This means I'd have to add extra API calls and add it to the auth state, or modify the backend...

Maybe we could just merge this as is for now and worry about it later?

nabramow (approved):
Okay I think this is good enough to go!

PR #5901, opened 2025-05-03
Map filters: Add meetup filter, change lastActiveOptions and adjust filter order

### LINKED ISSUE #5900: Add "Wants to meet" to map filters + adjustments
We already have "hosting status", but we should add "meetup status" to the map filters.

We also want to adjust the categories.

Acceptance Criteria
- Move can host filter up
- Remove host filters, put age under accommodation filters
- In `FilterDialog` component add `Wants to meet`. It can be same style and logic as hosting status
- Put it in the UI below hosting status
- Add it to the  map reducer in`app/web/features/search/state/mapSearchReducers.ts` logic, which manages the state under the `SET_FILTERS` action
- You'll need to add it to `initialState` in the reducer and `UserFilters` type
- Add it to the request in `app/web/service/search.ts` to pass it to the backend api

nabramow:
Addresses some map UI issues from the last core meeting:

- Get rid of hosting filters and move hosting status up
- Add meetup status
- Move age into accommodation filters
- Change order of lastActiveOptions and change to: 1 week, 1 month, 3 months, 6 months, 1 year

- Go to map search
- Check the new order of the filters looks good
- Use the meetup status filter and check it works

**Web frontend checklist**
- [ x  ] There are no warnings from `yarn lint --fix`
- [ x  ] There are no console warnings when running the app
- [ x ] All tests pass


aapeliv (dismissed):
LGTM

jesseallhands (changes requested):
This looks REALLY good!!! The changes you made will help the UX a LOT! I do have a few change requests:

_**Meetup Status Filter Order**_
The order of the "can't meet | open to meet | wants to meet" needs to be flipped to match the "can host | may host | can't host":
image

_**Disallow applying Hosting Status and Meetup Status Filters simultaneously**_
We should disallow applying the hosting status and meetup status filters at the same time because some users will try to apply both at the same time (since they are right next to each other) without realizing the implications. Furthermore, there isn't a common use case for doing this (searching for a host and a meetup person simultaneously). If a user attempts to apply both filter types (host status filter + meetup status filter), we should disallow the second filter type applied and display an error message that says the following:

> You cannot apply a hosting status filter and a meetup status filter at the same time.

**_"Last Active" filter:_** 
1. Would it be hard to add "last 6 months" as a filter option? Last 3 months seems a bit strict for people who live in rural or less-well-travelled areas, especially with the current community size/user activity level.
2. In terms of the sort order of the dropdown menu, I would suggest starting with "Any" but then sorting from least-restrictive to most restrictive. This is because users are more likely to choose something at the top of the menu, and those are extremely restrictive with few results (leading users to believe we have less active users than we do)

nabramow:
> This looks REALLY good!!! The changes you made will help the UX a LOT! I do have a few change requests:
> 
> _**Meetup Status Filter Order**_ The order of the "can't meet | open to meet | wants to meet" needs to be flipped to match the "can host | may host | can't host"

Ooo good catch. Fixed!

> _**Disallow applying Hosting Status and Meetup Status Filters simultaneously**

Let's keep it simple for now, as this involves adding in more logic and error handling and this ticket was just to address immediate UI issues.

> 1. Would it be hard to add "last 6 months" as a filter option? Last 3 months seems a bit strict for people who live in rural or less-well-travelled areas, especially with the current community size/user activity level.

I agree, but we need some backend done for this so I'll add to my agenda for Sunday to tap Aapeli about it, for the sake of keeping this ticket straight-forward for now.

> 2. In terms of the sort order of the dropdown menu, I would suggest starting with "Any" but then sorting from least-restrictive to most restrictive. This is because users are more likely to choose something at the top of the menu, and those are extremely restrictive with few results (leading users to believe we have less active users than we do)

Done!

c-kreuz:
These are great improvements! May I suggest to change "Age" to "Host age" to eliminate ambiguity.

jesseallhands (dismissed):
LGTM!

I did find something new but unrelated to this ticket: the strong verification tooltip (i) symbol doesn't work on mobile or mouse-less devices (I think it only works onmouseover). We probably should create a ticket for adding a bunch of tooltips in various places since I think there are a bunch we could add that are missing and some that are there but not working.
Screenshot_20250505_140110_Firefox

nabramow:
> These are great improvements! May I suggest to change "Age" to "Host age" to eliminate ambiguity.

Ooo good idea - done!

PR #5959, opened 2025-05-06
Only run migration test if ENV var is set

aapeliv:
This test will only run succesfully if you have `pg_dump` installed and everything set up, which only happens if you run
the test within the container

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/tests/test_db.py
+++ b/app/backend/src/tests/test_db.py
@@ -1,8 +1,8 @@
-import pytest
 import difflib
 import re
 import subprocess
 
+import pytest
 from sqlalchemy.sql import func
 
 from couchers.config import config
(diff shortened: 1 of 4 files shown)


app/backend/src/tests/test_db.py
```diff
@@ -143,7 +144,11 @@ def test_sort_pg_dump_output():
 def strip_leading_whitespace(lines):
     return [s.lstrip() for s in lines]
 
-
+"""
+This test will only run succesfully if you have `pg_dump` installed and everything set up, which only happens if you run
+the test within the container
```
pcolt: Should this be explained better or I am doing something wrong? Because when I run the tests following the instructions "Running tests in docker" I get same error with pd_dump() as running them locally.

pcolt (commented):
Just little clarification in the comment below. 
Also I see the deploying pipeline failing with the ruff check

app/backend/src/tests/test_db.py
aapeliv: Yes that's a good point, it doesn't actually run in the container, you have to separately install `pg_dump`. I've clarified the comment. Thanks!

pcolt (approved):
Everything looks good to me

PR #5967, opened 2025-05-07
Add duplicate accounts backend feature including api call to append, remove and get duplicated users 

### LINKED ISSUE #4355: add duplicate_accounts to UserDetails and create api call to append and remove usernames to this field
We need a way to link duplicate accounts (eg. when a user is banned and creates a new account, we need to be able to link them so the various accounts can be found easily again).

At the same time, a related planned feature is to allow users to link their profile with housemates and travel buddies. The idea is that if you live with someone, then you could request to link your profile with theirs and they would show up on each others' profiles as housemates. Similarly, if you are traveling with someone, you can link profiles as travel buddies and this would show up in requests you send. This would also enable the possibility of allowing people to send and receive linked requests. This solves a lot of problems discussed here on the forum: https://community.couchers.org/t/profiles-for-couples-and-friend-groups/431/27

Users could have multiple link types simultaneously housemate, travel_buddy, duplicate_account AND multiple linked accounts (eg. multiple housemates, multiple travel buddies, multiple banned accounts, etc. So I suggest we add housemates, travel_buddies, and duplicate_accounts to UserDetails which would be null by default but have one or more user_id or usernames each.

Then we need an api to update the fields of both user accounts whenever they are linked. 

**For now, we just need to be able to link duplicate accounts of banned users, so the api can be kept very simple, but we should keep the concept of linked accounts in mind when building this on the backend.**

- [ ] add duplicate_accounts to UserDetails
- [x] create API to append and remove APIs
- [ ] update Admin.GetUserDetails so that duplicate accounts are displayed when called (similar to AdminNotes) - though this step may need to be a different issue if aapeli isn't assigned to this issue 😅

pcolt:
This partially closes #4355.

- It adds a new many-to-many linking table to link duplicated users together (therefore includes a migration). The table has been designed to easily support in the future other kinds of links between users a part of the "duplicated" type (just adding a new value in the enumerator `UserLinkType`).
- It adds three new endpoints to link together two users at a time, remove a link and get all users linked to one given user.
- What I left out is the part regarding the `_user_to_details` function that I think was mentioned in the ticket. It was not easy and clear to me how to implement this part and in the ticket was suggested to leave this part for a following issue.

No specific steps required to review. I have mainly used unit tests to check everything.

Being new to the project and to some of the technologies used I obviously open to get any suggestion for change.

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/models.py
```diff
@@ -2667,3 +2667,32 @@ class AccountDeletionReason(Base):
...
+class UserLink(Base):
+    """
+    Links between users for various types
+    """
+
+    __tablename__ = "user_links"
+
+    id = Column(BigInteger, primary_key=True)
+    user1_id = Column(ForeignKey("users.id"), nullable=False)
+    user2_id = Column(ForeignKey("users.id"), nullable=False)
```
aapeliv: Add `index=True` to automatically create a lookup index for these

app/backend/src/couchers/servicers/admin.py
```diff
@@ -578,3 +581,76 @@ def EditReply(self, request, context, session):
...
+            user1_id=user1.id,
+            user2_id=user2.id,
+            link_type=UserLinkType.duplicate_account,
+        )
+
+        try:
+            session.add(user_link)
+            session.commit()
+        except exc.IntegrityError:
+            context.abort(grpc.StatusCode.ALREADY_EXISTS, errors.USER_LINK_ALREADY_EXISTS)
```
aapeliv: Switch this to a lookup: `IntegrityError` will catch other stuff too.

app/proto/admin.proto
```diff
@@ -93,6 +99,8 @@ message UserDetails {
 
   int64 pending_mod_notes_count = 16;
   int64 acknowledged_mod_notes_count =17;
+
+  repeated UserLink links = 18;
```
aapeliv: Please use `user_links`

app/proto/admin.proto
```diff
@@ -296,3 +304,32 @@ message EditReplyReq {
...
   string new_content = 2;
 }
+
+message LinkUsersDuplicatedReq {
+    string user1 = 1;
+    string user2 = 2;
+}
+
+message LinkUsersDuplicatedRes {
+    int64 link_id = 1;
```
aapeliv: `user_link_id`, etc

app/proto/admin.proto
pcolt: Oh, my apologies I forgot to revert this. I was trying to add the duplicate links to the _user_to_details administration function, but I got a few errors and wasn't sure what was the right way to implement this.

app/proto/admin.proto
pcolt: Not 100% sure about this, cause I see this a bit of a repetition. Anyway I changed all `link_id` to `user_link_id` and all `link_type` to `user_link_type` in the protos and in the consequent requests/responses but I didn't change the models/column tables. Hope this was what you intended?

pcolt:
Following Aapeli (and Jesse) suggestions 

> "The one thing that was discussed earlier was about having 'groups' of linked users, instead of just 1-to-1 links. So e.g. with dupe accounts you'll often see people make many accounts, and then it'd be hard to link each one together." 

I changed a bit the architecture adding a new association table `UserGroupMember` which groups users together instead of linking them one by one. 

The only downside of this approach is that a check before grouping has to be made ensuring that there are not two or more UserGroupMember associated with same group type (to avoid having repetitions of duplicated groups that should be unique instead).

I also added some methods to the sqlachemy models like `has_user` and `get_group_duplicated` that I honestly never used before but they were somehow a pattern used in the codebase already.

An UML of the new tables using plantuml:

Image

- [x] Formatted my code by running ruff check --select I --fix . && ruff check . --fix && ruff format . in app/backend
- [x] All tests pass
- [x] Added migrations if there are any database changes, rebased onto develop if necessary for linear migration history

app/backend/src/couchers/models.py
```diff
@@ -2667,3 +2673,37 @@ class AccountDeletionReason(Base):
     reason = Column(String, nullable=True)
 
     user = relationship("User")
+
+
+class UserGroupType(enum.Enum):
+    duplicate_account = enum.auto()
+
+
+class UserGroup(Base):
```
aapeliv: Let's specialize this and drop the `UserGroupType`, instead calling this `ModerationUserList` and do it just for duplciate users. "Group" is too overloaded here, since it's a concept in our communities side as well.

app/backend/src/couchers/models.py
```diff
@@ -2667,3 +2673,37 @@ class AccountDeletionReason(Base):
...
+
+class UserGroupMember(Base):
+    """
+    Association table for many-to-many relationship between users and groups
+    """
+
+    __tablename__ = "user_group_members"
+
+    user_id = Column(ForeignKey("users.id"), primary_key=True)
+    group_id = Column(ForeignKey("user_groups.id"), primary_key=True)
```
aapeliv: You should add a database constraint here to make sure each user is in each group at most once.

I see you implemented a restriction of one user can be in at most one group. You can ask Jesse what his preference is: in my view it might be useful to allow mods to have multiple user lists and use them for whatever is needed: e.g. duplicate accounts, but also to link them together otherwise (if they are related in some other activity, etc).

For each user being able to be in at most one group, you can add:

```py
    __table_args__ = (UniqueConstraint("user_id"),)
```

For each user being able to be listed in each group at most once:

```py
    __table_args__ = (UniqueConstraint("user_id", "group_id"),)
```

app/backend/src/couchers/servicers/admin.py
```diff
@@ -584,6 +586,77 @@ def EditReply(self, request, context, session):
...
+            user_group_duplicated = UserGroup(group_type=UserGroupType.duplicate_account)
+            session.add(user_group_duplicated)
+            session.flush()
+
+        # Add users to the duplicate account group only if they are not already in it
+        for user in users:
+            if user not in user_group_duplicated.users:
+                user_group_duplicated.users.append(user)
+
+        session.commit()
```
aapeliv: No need for a commit here: it autocommits when it leaves the `session_scope` contextmanager (defined in `interceptors.py:310` (see `SessionInterceptor`)

app/backend/src/couchers/servicers/admin.py
```diff
@@ -584,6 +586,77 @@ def EditReply(self, request, context, session):
...
+        user = session.execute(select(User).where_username_or_email_or_id(request.user)).scalar_one_or_none()
+        if not user:
+            context.abort(grpc.StatusCode.NOT_FOUND, errors.USER_NOT_FOUND)
+
+        duplicate_group = user.get_group_duplicated()
+
+        if not duplicate_group:
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.USER_NOT_IN_ANY_DUPLICATE_GROUP)
+
+        duplicate_group.users.remove(user)
```
aapeliv: If it's the last user in the group, then we should remove the group too, right?

app/backend/src/couchers/models.py
```diff
@@ -427,6 +429,10 @@ def phone_code_expired(self):
     def __repr__(self):
         return f"User(id={self.id}, email={self.email}, username={self.username})"
 
+    def get_group_duplicated(self):
+        """Returns the duplicate account group if user belongs to one, otherwise returns None."""
+        return next((group for group in self.groups if group.group_type == UserGroupType.duplicate_account), None)
```
aapeliv: This is better to move into the code where it runs.

app/backend/src/couchers/models.py
```diff
@@ -2667,3 +2673,37 @@ class AccountDeletionReason(Base):
...
+    id = Column(BigInteger, primary_key=True)
+    group_type = Column(Enum(UserGroupType), nullable=False)
+    created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
+
+    # Relationships
+    users = relationship("User", secondary="user_group_members", back_populates="groups")
+
+    def has_user(self, user):
+        """Check if user is in group"""
+        return user.id in {u.id for u in self.users}
```
aapeliv: This kind of check is better done in SQL. Right now when you call this function, it will pull the set of users for this group into python, construct the set, then check membership. With a SQL index, you can do that in very little computation on the database side.

aapeliv (changes requested):

app/backend/src/couchers/models.py
jesseallhands: >I see you implemented a restriction of one user can be in at most one group. You can ask Jesse what his preference is: in my view it might be useful to allow mods to have multiple user lists and use them for whatever is needed: e.g. duplicate accounts, but also to link them together otherwise (if they are related in some other activity, etc).

I agree with allowing mods to have multiple user lists

app/backend/src/couchers/models.py
pcolt: done

app/backend/src/couchers/models.py
pcolt: done

app/backend/src/couchers/servicers/admin.py
pcolt: OK, done

app/backend/src/couchers/servicers/admin.py
pcolt: yes you are right. Done

app/backend/src/couchers/models.py
pcolt: done

app/backend/src/couchers/models.py
pcolt: I moved it where it is used in the tests' code

pcolt:
This time I created only two endpoints: one for adding and another for removing users from the new moderation lists. I thought it was better to do one thing at a time. Also because the part of getting the lists is not yet clear to me and in the original issue was stated that @aapeliv would manage it. Of course I am open to implement it afterwards if needed

aapeliv (commented):
I think what's here looks good! Just one more thing is that I think we need a `ListUserModerationLists` or similar to get the moderation list the user is in.

app/proto/admin.proto
```diff
@@ -353,6 +355,15 @@ message AddUsersToModerationUserListReq {
   int64 moderation_list_id = 2;
 }
 
+message ListModerationUserListsReq {
+  // username, email, or user id
+  string user = 1;
+}
+
+message ListModerationUserListsRes {
+  repeated int64 moderation_list_ids = 1;
```
aapeliv: Maybe make a

```proto3
message ModerationList {
  int64 moderation_list_id = 1;
  repeated uint32 member_ids = 2;
}
```

and return repeated objects like that?

pcolt: This is done. I have now to undo the migration, merge develop and reapply the migrations. @aapeliv may you review this in the meantime?

jesseallhands:
@aapeliv did you finish your review on this?

aapeliv (dismissed):
LGTM

aapeliv (approved):

PR #6022, opened 2025-05-16
Fix: Landing on middle of page after clicking a link in footer

### LINKED ISSUE #5991: Landing on middle of page after clicking a link in footer
Subject: Landing on middle of blog page after clicking blog link in footer
Description:
I clicked on the "blog" link in the footer. I tried this on both firefox and chrome and it happens in both.

Results:
Instead of landing in the middle of the blog page the user should land at the top of the blog page.

Backend version: develop-a29eba26
Frontend version: develop-a29eba26
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:138.0) Gecko/20100101 Firefox/138.0
Page: https://couchers.org/blog
User: [@jesse](https://couchers.org/user/jesse) (57)

banflam:
closes #5991 

Note: I have solved this issue by deleting references to overflow. While this fixed the issue, I don't know why those references were there in the first place and it's possible this approach might introduce some other bug.

I've tested things myself but I've not found any newly-introduced bug.

If someone who wrote that code could comment it would be helpful. Otherwise we can go ahead and merge for now and see if there's any impact later.

**Web frontend checklist**

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -24,7 +24,6 @@ const globalStyles = (
     styles={{
       "html, body": {
         margin: 0,
-        overflow: "hidden", // Prevents whole-page scrolling
       },
       "#__next": {
         height: "calc(var(--vh, 1vh) * 100)", // Use the dynamic --vh value from _app
@@ -39,7 +38,6 @@ const PageWrapper = styled(Box)({
   display: "flex",
   flexDirection: "column",
   flex: 1,
-  overflowY: "auto",
 });
 
 const ContentWrapper = styled(Container, {


aapeliv:
The navbar now disappears when you scroll down

app/web/components/AppRoute.tsx
```diff
@@ -24,7 +24,6 @@ const globalStyles = (
     styles={{
       "html, body": {
         margin: 0,
-        overflow: "hidden", // Prevents whole-page scrolling
```
nabramow, line 27: Styles in this component affect the entire app, so changing here is quite risky. 

This is fixing it by accident, as you are saying "don't allow scrolling in main app page". Of course banning scrolling will fix any weird scrolling, but it will also break scrolling where we should have it.

You're going to want to find the specific component causing the weird scrolling behavior and fix it there instead.

banflam: Thanks for the feedback. I might not be able to look at this bug again until Friday 23rd if that's fine. Please let me know if fixing this bug is part of any release

nabramow:
@banflam are you still working on this? No worries if not, I just want to put it back on the board if you're no longer working on it so it doesn't get blocked for too long.

banflam:
@nabramow Sorry about the delay. I will work on it this weekend and post an update by Monday midnight EST. Is that alright?

banflam:
I don't know how to fix this bug @nabramow . I would be grateful if someone could work through this one with me, so that I can gain a deeper understanding of the codebase and be able to fix bugs like these in the future.

Here are the things that I've tried:

1 - Checked for relevant components having scrollIntoView somewhere, and if those scrollIntoView are causing the issue -- LandingPage.tsx, UserPage.tsx, useMapSearchActions.tsx

2 – Confirmed it was not a simple anchor/hash problem
Footer <Link> elements contain plain paths (/plan, /faq, …) – no #fragment.

3 – Eliminated obvious JavaScript scroll calls
Commented-out every scrollIntoView in Comment.tsx.

Deleted the moreContentRef scroll in LandingPage.

4 – Removed global “wiggle” hack in _app.tsx

window.scrollTo(0, scrollY+1); window.scrollTo(0, scrollY-1);
window.addEventListener("focusout", resetScroll);
Effect: confirmed via console trace that hack was gone, but mid-page landing persisted.

5 – Console-traced all scroll & focus calls
Added monkey-patch for scrollTo / scrollIntoView / focus.

After navigation the only call is Next.js’ own handleSmoothScroll → scrollTo(0,0).

No further scroll/focus calls occur; therefore JS is not pushing us down.

6 – Forced top scroll after route change (several variants)

router.events.on("routeChangeComplete", () => window.scrollTo(0,0))

Same with setTimeout(...,150) and with scrollRestoration="manual"

7 – Ruled out scroll={false}
Added scroll={true} to our StyledLink wrapper so every internal link explicitly requests top scroll – still jumps.

8 – Disabled viewport-height CSS hack
Commented-out updateVH() effect that rewrites --vh (mobile 100 vh workaround) → no change.

Observation: each route always lands at the same fixed offset (e.g., page A ≈75 %, page B bottom).

nabramow:
Closing as handled in #6216

PR #6023, opened 2025-05-16
Implement warning emails and blocking limits on host requests/friend requests/chat initiations

### LINKED ISSUE #5512: Moderation tooling: Messages rate limit trigger to notify support (console)
We want to add a trigger to the console (different codebase from main app), that notifies support if someone creates a non-human account of messages within a certain period of time. This might signal bots, etc that we want to be brought to our attention.

Acceptance Criteria
- Decide on what a non-human rate limit it (100 messages within 10 min?)
- Implement this trigger in the console for admins

spreeni:
closes #5512
**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/errors.py
```diff
@@ -67,6 +67,7 @@
 GROUP_OR_COMMUNITY_NOT_FOUND = "Group or community not found."
 HOST_REQUEST_CLOSED = "This host request is closed, use a normal message instead."
 HOST_REQUEST_IN_PAST = "This host request is in the past."
+HOST_REQUEST_SPAM_USER_BAN = "Due to excessive host requests, the account has been suspended until further notice. The moderation team has been notified."
```
aapeliv: Ah, this doesn't suspend the account, right? It just stops you from sending more reqs?

Maybe change it to `You've sent too many host requests in the 24 hours, try again later.`

app/backend/src/couchers/servicers/requests.py
```diff
@@ -164,6 +177,43 @@ def CreateHostRequest(self, request, context, session):
...
 
+        # Check if user has been sending host requests excessively
+        count_host_requests_last_24h = session.execute(
+            select(func.count())
+            .select_from(HostRequest)
+            .join(Conversation, HostRequest.conversation_id == Conversation.id)
+            .where(HostRequest.surfer_user_id == context.user_id)
+            .where(Conversation.created >= now() - timedelta(hours=24))
+        ).scalar_one()
+        if count_host_requests_last_24h > HOST_REQUEST_DAILY_BAN_QUOTA - 2:
```
aapeliv: So since the user isn't actually stopped from sending more reqs, this will send up to ~`HOST_REQUEST_SPAM_USER_BAN-HOST_REQUEST_DAILY_BAN_QUOTA` to the mods, right?

app/backend/src/couchers/errors.py
spreeni: In this case it would ban the user, so it would stop the user indefinitely until a mod review

app/backend/src/couchers/servicers/requests.py
spreeni: No, inside the action is aborted with an error, and the user is banned, so they can't send more. But this is also a draft, not sure if this should be final - waiting for yours and Jesses opinions still :)

aapeliv: I don’t think we should ever auto-ban, especially without warning. Let’s discuss on slack.

spreeni: Removed the banning - only blocking further requests now 👍

app/backend/src/couchers/errors.py
```diff
@@ -34,6 +34,9 @@
 CANT_WRITE_REFERENCE_FOR_REQUEST = "You can't write a reference for that host request, or it wasn't found."
 CANT_WRITE_REFERENCE_INDICATED_DIDNT_MEETUP = "You can't write a reference for that host request because you indicated that you didn't meet up."  # fmt: skip
 CHAT_NOT_FOUND = "Couldn't find that chat."
+CHAT_INITIATION_THRESHOLD = (
+    "You have messaged a lot of users today. To avoid spam, you can't contact any more users for today."
```
aapeliv: "Today" is ambiguous, use "24 hours" somehow

app/backend/src/couchers/constants.py
```diff
@@ -44,3 +44,11 @@
 ACTIVENESS_PROBE_TIME_REMINDERS = [timedelta(days=0), timedelta(days=4, hours=8)]
 # total time from initiation after which to expire the probe
 ACTIVENESS_PROBE_EXPIRY_TIME = timedelta(days=14)
+
+# request quotas
+HOST_REQUEST_DAILY_WARNING_QUOTA = 20
+HOST_REQUEST_DAILY_BLOCKING_QUOTA = 80
```
aapeliv: Don't use the word `blocking` since the user is not blocked, call it a `limit`, or hard limit, whatever

app/backend/src/couchers/servicers/api.py
```diff
@@ -657,6 +671,38 @@ def SendFriendRequest(self, request, context, session):
...
+            send_friend_request_spam_report_email(
+                session=session,
+                user=user,
+                friend_requests=friend_requests,
+                threshold=FRIEND_REQUEST_DAILY_BLOCKING_QUOTA,
+                time_interval_str="24 hours",
+                user_is_blocked=True,
+            )
+            context.abort(grpc.StatusCode.RESOURCE_EXHAUSTED, errors.FRIEND_REQUEST_THRESHOLD)
+        if count_friend_requests_last_24h == FRIEND_REQUEST_DAILY_WARNING_QUOTA - 1:
```
aapeliv: You are not guaranteed to catch this with equality. There can be race conditions, especially if someone is spamming. So you should update this somehow to use inequality.

app/backend/src/couchers/servicers/api.py
```diff
@@ -657,6 +671,38 @@ def SendFriendRequest(self, request, context, session):
...
             context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.FRIENDS_ALREADY_OR_PENDING)
 
+        # Check if user has been sending friend requests excessively
+        count_friend_requests_last_24h = session.execute(
+            select(func.count())
+            .select_from(FriendRelationship)
+            .where(FriendRelationship.from_user_id == context.user_id)
+            .where(FriendRelationship.time_sent >= now() - timedelta(hours=24))
+        ).scalar_one()
+        if count_friend_requests_last_24h >= FRIEND_REQUEST_DAILY_BLOCKING_QUOTA - 1:
```
aapeliv: Why the `-1`?

app/backend/src/couchers/servicers/api.py
```diff
@@ -657,6 +671,38 @@ def SendFriendRequest(self, request, context, session):
         ):
             context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.FRIENDS_ALREADY_OR_PENDING)
 
+        # Check if user has been sending friend requests excessively
+        count_friend_requests_last_24h = session.execute(
```
aapeliv: Please make the time interval a constant as well.

aapeliv (changes requested):
Let's make this a bit more generic. I suggest you:
- Move the logic on reporting emails to a background job where you give it the `user_id`, and the limiting reason as well as if they hit a soft limit or hard limit.
- Make a single system email template that includes all the information you currently have. That way you can fetch all that info in the background and give a full picture to mods.
- That way you only need to do a `func.count(HostRequests).where(...)` or similar on the hot path.
- I would move the rate limit logic into somewhere like `rate_limit.py` and import from there.
- Ideally you could write these rate limit events into the database somewhere. That way we can do queries against it, as well as rate limit the rate limit emails (this would also fix the race condition issue in one of my comments).

I think that will make the system easier to extend and work with, without cluttering the critical paths of API calls.

Regarding wording:
- Don't use the word "block" in this context. It has very specific meaning in our codebase. You can use "limit", "rate limited", "hard limit", word like that.
- Let's also not call these emails "spam_report"s. Again, "spam" has a very specific meaning and I would rather we use some other wording around it. You could call them "rate_limit_notification" reports, or "excessive_{something}", etc.

app/backend/src/couchers/servicers/api.py
spreeni: So it blocks on the 50th request if FRIEND_REQUEST_DAILY_BLOCKING_QUOTA is 50, as the count are just the already made requests.

aapeliv: Well if there is a limit of n requests, you would assume the user will be limited on their n+1th req? So that they can do exactly n and no more?

app/backend/src/couchers/servicers/api.py
spreeni: Done 👍

app/backend/src/couchers/servicers/api.py
spreeni: Changed it now, yeah I guess that makes sense 👍

app/backend/src/couchers/servicers/api.py
spreeni: Is this something we have encountered with Couchers? But adjusted it to inequality, and storing RateLimitViolations in the DB

app/backend/src/couchers/constants.py
spreeni: Adjusted the wording everywhere now!

app/backend/src/couchers/errors.py
spreeni: I set a string representation of the interval constant as another constant and am using it everywhere now. This lets us adjust it flexibly.

spreeni:
@aapeliv Thanks for the good feedback! I implemented most of your points, have a look again. It looks a lot nicer now I think, was just a bit lazy the first time I guess 😅

app/backend/src/couchers/models.py
```diff
@@ -2756,3 +2756,21 @@ class AccountDeletionReason(Base):
     reason = Column(String, nullable=True)
 
     user = relationship("User")
+
+
+class RateLimitAction(enum.Enum):
+    host_request = "host request"
```
aapeliv, line 2797: Please make these `enum.auto()`

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,214 @@
...
+def _get_rate_limit_for_action(action: RateLimitAction) -> tuple[int, int]:
+    match action:
+        case RateLimitAction.host_request:
+            return HOST_REQUEST_WARNING_LIMIT, HOST_REQUEST_HARD_LIMIT
+        case RateLimitAction.friend_request:
+            return FRIEND_REQUEST_WARNING_LIMIT, FRIEND_REQUEST_HARD_LIMIT
+        case RateLimitAction.chat_initiation:
+            return CHAT_INITIATION_WARNING_LIMIT, CHAT_INITIATION_HARD_LIMIT
+        case _:
+            raise ValueError(f"Unknown rate limit action: {action}")
```
aapeliv: Maybe it's cleaner to turn this into a dictionary in `constants.py`?

```py
{
  RateLimitAction.host_request: (20,80),
  #...
}
```

Right now this is duplicating stuff in `constants.py` for not much value: you have the structure here but the values there. It would be good to keep everything in one consise and mostly self-explanatory place, so that when you add/modify them you need to only change one place.

---

On second thought, I would go one step further: I would define a structure in this file that contains *all* the info and logic for a particular rate limit:

```py
RATE_LIMIT_DEFINITIONS = {
  RateLimitAction.host_request: {
    "soft_limit": 20,
    "hard_limit": 80,
    # counts number of times the rate-limitable thing has been done in the interval
    "count_actions_query": lambda (session, user_id, interval): session.execute(
        select(func.count())
        .select_from(HostRequest)
        .join(Conversation, HostRequest.conversation_id == Conversation.id)
        .where(HostRequest.surfer_user_id == user_id)
        .where(Conversation.created >= now() - interval)
    ).scalar_one(),
    # 
    "mod_email_information_query": ...
  },
  # other rate limits
}
```

I think right now the code is a bit too nested and has a lot of imperative bits *per rate limit*. I don't think there's need to duplicate all the copy-paste logic per-limit in this file, you could write something slightly more complicated based on such a definition as above.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -530,6 +534,10 @@ def CreateGroupChat(self, request, context, session):
             ).scalar_one_or_none():
                 context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.ALREADY_HAVE_DM)
 
+        # Check if user has been initiating chats excessively
+        if has_reached_chat_initiation_limit(session=session, user_id=context.user_id):
```
aapeliv: It's not clear now that this is not a pure function based on the name. I don't know what to call it. A really bad idea is `process_rate_limit_return_should_abort`

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,214 @@
...
+    """Get all events for the user in the last rate limit interval for the mod email."""
+    match action:
+        case RateLimitAction.host_request:
+            return _get_user_host_requests_in_past_time_interval(session, user_id)
+        case RateLimitAction.friend_request:
+            return _get_user_friend_requests_in_past_time_interval(session, user_id)
+        case RateLimitAction.chat_initiation:
+            return _get_user_initiated_chats_in_past_time_interval(session, user_id)
+        case _:
+            raise ValueError(f"Unknown rate limit action: {action}")
```
aapeliv: I would just return *all* events in the email, regardless of the rate limit type that the user hit. That gives mods a better view.

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,214 @@
...
+) -> bool:
+    warning_limit, hard_limit = _get_rate_limit_for_action(action=action)
+    for limit, is_hard_limit in [(warning_limit, False), (hard_limit, True)]:
+        if count_last_interval >= limit:
+            if (
+                _user_rate_limit_violations_in_past_time_interval(
+                    session=session, user_id=user_id, action=action, hard_limit=is_hard_limit
+                )
+                == 0
+            ):
```
aapeliv: Please change `_user_rate_limit_violations_in_past_time_interval` to be `_user_has_violated_rate_limit_in_past_time_interval` if this is only used to check existence.

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,214 @@
...
+            return CHAT_INITIATION_WARNING_LIMIT, CHAT_INITIATION_HARD_LIMIT
+        case _:
+            raise ValueError(f"Unknown rate limit action: {action}")
+
+
+def _has_reached_rate_limit(
+    session: "Session", user_id: int, action: RateLimitAction, count_last_interval: int
+) -> bool:
+    warning_limit, hard_limit = _get_rate_limit_for_action(action=action)
+    for limit, is_hard_limit in [(warning_limit, False), (hard_limit, True)]:
```
aapeliv: Does this run twice if `count_last_interval>hard_limit`? if so, you can swap the order to make it run only once, right?

aapeliv (changes requested):
Thanks for making it more abstract! I think it's much cleaner now in a separate module. I added some more comments :)

app/backend/src/couchers/rate_limit.py
spreeni: Having these queries in `constants.py` would lead to circular imports between `models.py` and `constants.py`, but I like the first approach! Added it 👍

app/backend/src/couchers/models.py
spreeni: I intentionally chose to set a string here so I can use it in the email template. Or do you have another reason why this is not a good idea?

app/backend/src/couchers/rate_limit.py
spreeni: True, changed it 👍

app/backend/src/couchers/rate_limit.py
spreeni: Done! 👍

app/backend/src/couchers/servicers/conversations.py
spreeni: Yeah, this is a good point. Was thinking to break it up, but I think from the flow it makes sense like this, so just renamed it.

app/backend/src/couchers/rate_limit.py
spreeni: Was a bit afraid of overloading the moderators with excessive information, but yeah, probably makes sense to rather serve more info. Added it 👍

spreeni (commented):
Thanks for the great feedback! Have a look again @aapeliv :)

app/backend/src/couchers/constants.py
```diff
@@ -44,3 +46,27 @@
...
 # total time from initiation after which to expire the probe
 ACTIVENESS_PROBE_EXPIRY_TIME = timedelta(days=14)
+
+
+class RateLimitAction(Enum):
+    """Possible user actions which can be rate limited."""
+
+    host_request = "host request"
+    friend_request = "friend request"
+    chat_initiation = "chat initiation"
```
aapeliv: I would like this in `models.py` given that it's a DB enum.

Simultaneously I would like `constants.py` to not contain any real imports...

The former I think is more important than the latter. Or I would suggest you just move the rate limit definitions into the rate limit file.py

app/backend/src/couchers/models.py
```diff
@@ -2756,3 +2757,15 @@ class AccountDeletionReason(Base):
     reason = Column(String, nullable=True)
 
     user = relationship("User")
+
+
+class RateLimitViolation(Base):
+    __tablename__ = "rate_limit_violations"
+
+    id = Column(BigInteger, primary_key=True)
+    created = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
```
aapeliv, line 2806: `time`? More descriptive than `created`?

app/backend/src/couchers/models.py
```diff
@@ -2756,3 +2757,15 @@ class AccountDeletionReason(Base):
...
+    user_id = Column(ForeignKey("users.id"), nullable=False)
+    action = Column(Enum(RateLimitAction), nullable=False)
+    hard_limit = Column(Boolean, nullable=False)
```
aapeliv: `is_hard_limit`

app/backend/src/couchers/models.py
aapeliv: OK yea that's good. I think sqlalchemy uses the labels anyway

app/backend/src/couchers/rate_limit.py
aapeliv: See my comment on imports :upside_down_face: 

I think it would be cleaner to keep it all in `rate_limits.py` -- you could also make a module for the whole subsystem and have `rate_limits/constants.py` since this will probably grow.

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,182 @@
...
+    session.flush()
+    return violation
+
+
+def _user_has_violated_rate_limit_in_past_time_interval(
+    session: "Session", user_id: int, action: RateLimitAction, hard_limit: bool
+) -> bool:
+    """Check if a RateLimitViolation for the user for the given action exists in the last RATE_LIMIT_INTERVAL."""
+    return (
+        session.execute(
```
aapeliv: Use `EXISTS`

app/backend/src/couchers/rate_limit.py
```diff
@@ -0,0 +1,182 @@
...
+            select(func.count())
+            .select_from(RateLimitViolation)
+            .where(RateLimitViolation.user_id == user_id)
+            .where(RateLimitViolation.action == action)
+            .where(RateLimitViolation.created >= now() - RATE_LIMIT_INTERVAL)
+            .where(RateLimitViolation.hard_limit == hard_limit)
```
aapeliv: Add an index into the `RateLimitViolation` table as follows:

```py
    __table_args__ = (
        # Fast lookup for rate limits in interval
        Index(
            "ix_rate_limits_by_user",
            user_id,
            action,
            hard_limit,
            created
        ),
    )
```

aapeliv (changes requested):

app/backend/src/couchers/rate_limit.py
spreeni: I created a new module as discussed for rate_limits, have a look :)

app/backend/src/couchers/constants.py
spreeni: Done!

app/backend/src/couchers/models.py
spreeni: To be consistent with the rest of the code base, I left it as `created` for now. Let me know if it bothers you a lot 😅

app/backend/src/couchers/models.py
spreeni: Done!

app/backend/src/couchers/rate_limit.py
aapeliv: What about the queries @spreeni ?

spreeni: Added them to the definition!

spreeni:
@aapeliv Have a look again, I packed a lot now into RATE_LIMIT_DEFINITIONS and left an instruction on how to add future rate limits.

aapeliv (approved):

PR #6030, opened 2025-05-18
Add API to remove friends

### LINKED ISSUE #6028: Add backend api to remove a friend connection
Right now there is no way to remove a friend connection.

We need a backend API to allow this.

aapeliv:
Closes #6028

-
-
-

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/backend/src/couchers/servicers/api.py
```diff
@@ -517,9 +517,6 @@ def UpdateProfile(self, request, context, session):
             else:
                 user.camping_ok = request.camping_ok.value
 
-        # save updates
-        session.commit()
-
```
nabramow, line 522: Why remove this?

aapeliv: It's not needed, when it leaves the `session_scope` it's wrapped in, it automatically commits the transaction. It's form a bygone era :P

app/backend/src/couchers/servicers/api.py
```diff
@@ -543,6 +540,27 @@ def ListFriends(self, request, context, session):
...
+            .where(FriendRelationship.status == FriendStatus.accepted)
+        ).scalar_one_or_none()
+
+        if not rel:
+            context.abort(grpc.StatusCode.FAILED_PRECONDITION, errors.NOT_FRIENDS)
+
+        session.delete(rel)
+
+        return empty_pb2.Empty()
+
```
nabramow, line 563: Does the FriendRelationship table store friends in both directions?

If so we need to make sure both directions are removed.

Since it says `scalar_one_or_none` seems it's only going to do one.

aapeliv: It's only one way, from the `from` is the requester, `to` is the requestee.

nabramow (approved):
Looks good to me! Thanks for doing this so quickly!

PR #6037, opened 2025-05-18
Remove feedback form

aapeliv:

- Try signing up
- Make sure feedback form does not show up
- Make sure you can signup otherwise

**Web frontend checklist**
- [x] All tests pass


nabramow:
Now that this screen is the last one, we should change the button text to "Submit":

Screenshot 2025-05-18 at 7 12 47 PM


Otherwise everything looks good and signup works!

nabramow (approved):
Looks good to me

nabramow (changes requested):
Just blocking so @jesseallhands can review since we're removing the whole last page here.

TBH I don't think it's bad as it makes the process shorter and since we're going to put this stuff in a more appropriate place after beta this seems okay?

jesseallhands (approved):
On Firefox the placement of the cookies banner and the Firefox auto-fill prompt blocking the button to continue. This isn't caused by this PR, but we should address it because it will definitely prevent signups (even though it is possible to "get around" the cookie banner and auto-fill issues):

On chrome it's a bit less problematic, but the cookie banner is still very obtrusive and far from standard behavior on websites:
Screenshot_20250519_183227_Chrome

jesseallhands:
Here is the problem with the browser auto-fill and the cookie banner:
https://github.com/user-attachments/assets/76b6487b-bab6-4ab2-98c6-dfbf536759c9

PR #6051, opened 2025-05-19
Robots.txt: disallow all for non-prod instances

aapeliv:
Our next instance is showing up on search, this should fix it.

Testing:

Make sure this says "Disallow: /":

https://couchers-git-frontend-bugfixrobots-off-prod-couchers-org.vercel.app/robots.txt


jesseallhands (approved):
LGTM

PR #6054, opened 2025-05-20
Fix backend RemoveFriend query

nabramow:
The `RemoveFriend` query that finds the friend relationship to remove was returning more than one row and throwing an error since it's `scalar_one_or_none`.

When I logged it, I realized it was returning any relationships that contained the given `user_id`.

We need to make it more specific, tell it we only want the friend relationship with `request.user_id` and `context.user_id`.

- There's no frontend yet, but I have tested it against my frontend branch successfully.
- Otherwise we can test it after merging
- Make sure my logic makes sense

**Backend checklist**
- [ x ] Formatted my code by running `ruff check --select I --fix . && ruff check . --fix && ruff format .` in `app/backend`
- [ x ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -547,8 +547,14 @@ class API(api_pb2_grpc.APIServicer):
             .where_users_column_visible(context, FriendRelationship.to_user_id)
             .where(
                 or_(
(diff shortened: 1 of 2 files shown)


aapeliv (approved):
Thanks for picking this up! Good pythoning :)

LGTM, feel free to merge when tests pass!

PR #6055, opened 2025-05-20
Add numReferences to references tab

### LINKED ISSUE #5948: Add numReferences to References tab of profile
It would be helpful upon first clicking into a profile to see that person's number of references right away.

Acceptance Criteria
* Add a small circle in `theme.palette.primary.main` color with the `numReferences` in bold white text inside.
* In the `EditProfile` component, we already have `user.numReferences`. I *think* the easiest way to do this would be to just move the state for the `useCurrentUser` response up into `EditProfilePage` where the tabs are, then just pass `user` down through the props to `EditProfile` instead.

nabramow:
Closes #5948 

- Go to a few users' profiles pages
- You should see the number of the references on the tab
- It should match the actual references within the tab

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/features/profile/constants.ts
+++ b/app/web/features/profile/constants.ts
@@ -13,12 +13,6 @@ export const contactLink = "mailto:support@couchers.org";
 export const hostRequestReferenceSuccessDialogId =
   "hostRequestReferneceSuccessDialog";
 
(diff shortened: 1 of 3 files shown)


aapeliv (approved):
LGTM

PR #6063, opened 2025-05-20
Language selector frontend functionality

### LINKED ISSUE #5470: Language selector functionality
Blocked by #5468 (backend cookie setting) and #5495 (creation of select)

Right now the language is auto-detected and there's no clear way for the user to choose their language. Given that our users travel a lot, this is inconvenient if you want a different language than the one for the country you're in.

We want to create a way to select your language with this PR.

Acceptance Criteria
- Set default selected language by getting the couchers-language-preference cookie from #5468 
- Add the select to the navigation bar (see design below)
- Add language select to the account settings page as well, with a link "Help translate Couchers into your language" in bold (see design).
- On submit click, we should: 1. set the new `couchers-preferred-language` to the cookie from #5468 , 2. send to the backend to be stored in the profile once that is implemented (#5467)
- There should be code built into i18n by this point that resets the language when the cookie changes, but if not, we'll need to do this too. (@nabramow will know the answer to this, ask her any questions)
- Write some tests for the main functionality (check cookie is set, default value is set on load, value is changed on click (and sent to the backend, i.e. proper function called)

Select in NavBar and account settings page for language
Image

Image

laurabotel:
Closes #5470 

- Log in
- Open  DevTools: command + option + i
- Go to Application => Cookies ==> Localhost/vercel
- Watch the NEXT_LOCALE cookie, value (if you have one)
- Change the language in the language picker
- The NEXT_LOCALE cookie should change /appear with that new value
- Log out
- Change the language picker to a different language that's not your cookie saved language
- Log in
- It should redirect to your saved cookie language after log in
- Go to account settings and find language picker, make sure it works there

Screenshot 2025-06-08 at 10 02 47 PM

**Web frontend checklist**
- [x] All tests pass


nabramow:
Okay first initial look - I am not 100% certain of this but this is my theory of what's happening.

It looks like you're using the Router to pass the new language. Routing happens a layer below i18n, which is a wrapper around the whole app. So when you log in, i18next default settings will hit first and then the routing. 

This would explain the weird login behavior, I think.

What I believe we want to do is handle the language change via i18next itself.

I was googling a bit and this package might be a good solution, it's from i18next itself, and seems to have the exact logic we want:

https://www.npmjs.com/package/i18next-browser-languagedetector

Using that model ^^, under "Adding own detection functionality", the cookie lookup would go under their lookup function, then if there's no cookie we default to the browser language seems to make sense.

I thiiiink we'd add this detection to `app/web/next-i18next.config.d.ts` and edit that to have the `detection` key. That file is called in `next.config.js` and importing via the `i18n` key.

That should have i18n handle the detection then, which should hopefully solve the issue. I think the problem is we need to handle the language switching and detection a level higher since i18next is wrapper around the entire app.

I can probably throw up a quick PR with the setup for that package if it would be helpful and put you as the reviewer, or I can push it to this PR, just let me know.

nabramow:
Just a smaller thing before I forget as I was testing in the incognito window. We should probably add the language selector to the logged out navigation bar too:

Screenshot 2025-05-23 at 8 50 20 AM

aapeliv:
This branch now deploys here as well: https://language-selector.vercel.couchershq.org

app/web/features/auth/login/Login.test.tsx
```diff
@@ -13,54 +13,61 @@ const passwordLoginMock = service.user.passwordLogin as jest.MockedFunction<
...
-
-  const user = userEvent.setup();
+describe("Login", () => {
+  it("shows the known gRPC error from the API", async () => {
+    const errorMessage = "Couldn't find that user.";
+    passwordLoginMock.mockRejectedValue({
+      code: 5,
+      message: errorMessage,
+    });
+    render(<Login />, { wrapper });
```
nabramow, line 23: I just wrapped this in a `describe` but otherwise didn't change anything.

app/web/features/auth/login/LoginForm.tsx
```diff
@@ -1,5 +1,4 @@
-import { FormControlLabel } from "@mui/material";
-import { styled } from "@mui/styles";
+import { FormControlLabel, styled } from "@mui/material";
```
nabramow, line 1: Just fixed where `styled` is imported from as it was causing an error.

app/web/features/landing/locales/es.json
```diff
@@ -1 +1,8 @@
-{}
+{
+  "introduction_title": "La nueva alternativa a Couchsurfing™",
+  "introduction_subtitle": "Gratis para siempre. Dirigido por la comunidad. Sin ánimo de lucro. Moderna.",
+  "signup_header": "Regístrate en Couchers.org hoy",
+  "signup_description": "Únete a {{ user_count }} couchsurfers en la plataforma de couchsurfing de más rápido crecimiento.",
+  "create_an_account": "Crear una cuenta",
+  "signup_continue": "Finalizar registro"
```
nabramow, line 7: @brunagt help I got a strong assist from AI here. I didn't want the Spanish on the main page to be empty though.

app/web/resources/locales/es.json
```diff
@@ -1 +1,9 @@
-{}
+{
+  "language_preference": {
+    "current_preferred_language": "Tu idioma preferido actualmente es ",
+    "form_title": "Idioma",
+    "help_translate": "Ayuda a traducir couchers a tu idioma.",
+    "select_language": "Selecciona un idioma",
+    "success_message": "Tu cambio ha sido guardado."
```
nabramow, line 7: @brunagt Also here, please correct me if it's wrong 😅.

aapeliv:
I need to click on it twice for some reason:

[Screencast from 2025-06-17 18-40-06.webm](https://github.com/user-attachments/assets/e4de9583-a183-4265-84cc-dd5724be4a0f)

aapeliv (approved):
LGTM

PR #6070, opened 2025-05-21
Add info for signup page

aapeliv:
Screenshot from 2025-05-21 17-22-58

-
-
-

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/public.py
+++ b/app/backend/src/couchers/servicers/public.py
@@ -4,7 +4,7 @@ import grpc
 from sqlalchemy.sql import func, union_all
 
 from couchers import errors
-from couchers.models import ProfilePublicVisibility, Reference, User
+from couchers.models import Cluster, Node, ProfilePublicVisibility, Reference, User
 from couchers.servicers.api import fluency2api, hostingstatus2api, meetupstatus2api, user_model_to_pb
 from couchers.servicers.gis import _statement_to_geojson_response
 from couchers.sql import couchers_select as select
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/servicers/public.py
```diff
@@ -107,3 +107,36 @@ def GetPublicUser(self, request, context, session):
...
+            .all()
+        )
+
+        if len(communities) <= 1:
+            # either no community or just global community
+            last_location = "the World"
+        else:
+            # probably global, continent, region, city
+            last_location = f"{communities[-1]}, {communities[-2]}"
+
```
nabramow, line 135: Why do we need this part if we're already grabbing the most recent user created?

app/backend/src/couchers/servicers/public.py
```diff
@@ -107,3 +107,36 @@ def GetPublicUser(self, request, context, session):
                     badges=[badge.badge_id for badge in user.badges],
                 )
             )
+
+    def GetSignupPageInfo(self, request, context, session):
+        # last user who signed up
+        last_signup, geom = session.execute(
+            select(User.joined, User.geom).where(User.is_visible).order_by(User.id.desc())
```
nabramow, line 114: Is there no `created_at` column on the User table? I guess ids are reliable but that seems even more so?

aapeliv: The `joined` is `created_at`. But there's a database index on `id`, that's why I'm sorting by that (and later ids should have later joined times, probably)

app/backend/src/couchers/servicers/public.py
aapeliv: We need to grab the names of the communities, it's not in the user table

nabramow: Ah duh that makes sense.

nabramow (approved):
Looks good to me!

Issue #5400, opened 2024-12-27
Bug: Dashboard events doesn't show events for users' communities


nabramow:
On the Dashboard page, the "My Events" section doesn't show events for the communities a user is in, only events attending.

First step - replicate this. Join a city community, make an event in it (but don't attend the event) and see if it shows up on Dashboard.

If bug is confirmed, fix this.

Hint: My guess is we might need to switch which api call we're doing, or adjust the api call, but please confirm. For example in the "Events" tab under "Discover Events" there is a "Communities" pill to select. 

Acceptance Criteria:

- We want both the events the user is attending and the events that are in their joined communities.
- The dashboard should display any upcoming events within the user's city-level or lower communities (in addition to any events they are attending). The definition of city-level would be any community with >=3 parents (communities above it) when calling api.communities.GetCommunityReq that the user belongs to (which can be called with api.communities.ListUserCommunitiesReq)

![Image](https://github.com/user-attachments/assets/793dd377-d6a2-4ea5-a85e-4a61904371ea)

nabramow:
@jesseallhands wrote this one up real quick as I was thinking about it. Feel free to edit the description directly if I got anything wrong, then move to "Ready". Or give me the a-ok and I can!

bakeiro:
I feel we could implement also the behavior of the "near me" pill we had on events page, like by default showing all the events that are near the profile location, maybe makes sense?

c-kreuz:
I can confirm this bug. This will be a great improvement once implemented.

PR #6082, opened 2025-05-25
Fast user search

aapeliv:
This PR replaces `UserSearch` that returnes `Result`s of type `User` with a new `UserSearchV2` that returns simply lists of `SearchUser` objects. The `SearchUser` object contains the subset of information required by the search screen.

This also simplifies a few small things (like passing a boolean for whether the user has completed their profile in the API).

- Run backend & frontend
- Make sure user search on map is working as expected.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/search.py
+++ b/app/backend/src/couchers/servicers/search.py
@@ -7,8 +7,9 @@ from datetime import timedelta
 import grpc
 from sqlalchemy.sql import and_, func, or_
 
-from couchers import errors
+from couchers import errors, urls
 from couchers.crypto import decrypt_page_token, encrypt_page_token
+from couchers.materialized_views import lite_users, user_response_rates
 from couchers.models import (
     Cluster,
     ClusterSubscription,
@@ -29,9 +30,13 @@ from couchers.models import (
 from couchers.servicers.account import has_strong_verification
 from couchers.servicers.api import (
     fluency2sql,
+    get_num_references,
+    hostingstatus2api,
     hostingstatus2sql,
+    meetupstatus2api,
     meetupstatus2sql,
     parkingdetails2sql,
+    response_rate_to_pb,
     sleepingarrangement2sql,
     smokinglocation2sql,
     user_model_to_pb,
@@ -42,8 +47,10 @@ from couchers.servicers.groups import group_to_pb
 from couchers.servicers.pages import page_to_pb
 from couchers.sql import couchers_select as select
 from couchers.utils import (
+    Timestamp_from_datetime,
     create_coordinate,
     dt_from_millis,
+    get_coordinates,
     last_active_coarsen,
     millis_from_dt,
     now,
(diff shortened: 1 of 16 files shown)


app/backend/src/couchers/servicers/search.py
```diff
@@ -338,6 +345,196 @@ def _search_clusters(
...
+    # This is a bit of a hacky feature to help with the frontend map implementation
+    if len(request.exactly_user_ids) > 0:
+        statement = statement.where(User.id.in_(request.exactly_user_ids))
+    else:
+        # Apply all the normal filters
+        if request.HasField("query"):
+            if request.query_name_only:
+                statement = statement.where(
+                    or_(User.name.ilike(f"%{request.query.value}%"), User.username.ilike(f"%{request.query.value}%"))
+                )
```
nabramow, line 364: I don't think we give an option on the frontend to say nameOnly for the search?

app/backend/src/couchers/servicers/search.py
```diff
@@ -338,6 +345,196 @@ def _search_clusters(
...
+                        4326,
+                    ),
+                )
+            )
+        if request.HasField("search_in_community_id"):
+            # could do a join here as well, but this is just simpler
+            node = session.execute(select(Node).where(Node.id == request.search_in_community_id)).scalar_one_or_none()
+            if not node:
+                context.abort(grpc.StatusCode.NOT_FOUND, errors.COMMUNITY_NOT_FOUND)
+            statement = statement.where(func.ST_Contains(node.geom, User.geom))
```
nabramow, line 500: Ooo is this new? Will this fix that bug with the dashboard communities? #5400

app/backend/src/couchers/servicers/search.py
aapeliv: There's a bunch of unused filters implemented in the backend.

I didn't modify this bit.

app/proto/search.proto
```diff
@@ -23,6 +24,11 @@ service Search {
     // Search for users in particular with extra filters, e.g. hosts
   }
 
+  rpc UserSearchV2(UserSearchReq) returns (UserSearchV2Res) {
+    // Search for users in particular with extra filters, e.g. hosts
+    // only returns used fields
+  }
+
```
nabramow, line 31: Is there supposed to be something in the brackets?

app/web/components/Avatar/Avatar.tsx
```diff
@@ -42,7 +42,7 @@ const StyledSkeleton = styled(Skeleton)(() => ({
 
 export interface AvatarProps {
   children?: React.ReactNode;
-  user?: LiteUser.AsObject;
+  user?: Pick<LiteUser.AsObject, "username" | "name" | "avatarThumbnailUrl">;
```
nabramow, line 45: We should change this to `SearchUser.AsObject` now probably.

app/backend/src/couchers/servicers/search.py
aapeliv: I didn't modify this bit, so not new. This should only be for user search, I'm not sure if this would help with 5400

app/web/features/profile/view/userLabels.tsx
```diff
@@ -47,7 +47,14 @@ export const ReferencesLastActiveLabels = ({ user }: Props) => {
   );
 };
 
-export const ResponseRateText = ({ user }: { user: User.AsObject }) => {
+export const ResponseRateText = ({
+  user,
+}: {
+  user: Pick<
+    User.AsObject,
```
nabramow, line 54: This is also `SearchUser.AsObject` now?

app/proto/search.proto
aapeliv: No, empty. We have comments in there sometimes (like here) and if using JSON transcoding or other options, they go in there as well.

It's funky syntax!

app/web/components/Avatar/Avatar.tsx
aapeliv: This is used in a whole bunch of places though, right? Not just with SearchUser

app/web/features/profile/view/userLabels.tsx
aapeliv: This is shared in places that user `User.AsObject` as well

app/web/features/search/utils/mapUtils.ts
```diff
@@ -17,8 +17,7 @@ const usersToGeoJSON = (pins: User.AsObject[]): FeatureCollection => ({
     },
     properties: {
       id: pin.userId,
-      hasCompletedProfile:
-        pin.avatarUrl && pin.aboutMe && pin.aboutMe.length >= 150,
+      hasCompletedProfile: pin.hasCompletedProfile,
```
nabramow, line 20: Ooo nice

nabramow (approved):
This seem seems to be okay to me! Could we deploy to stage first though to try it with a bigger data set?

PR #6085, opened 2025-05-25
Frontend perf stuff

### LINKED ISSUE #5415: Don't load fonts from google
We are loading something from `fonts.googleapis.com`. I think we already load it from fontsource, and I don't like relying on Big G for this kind of thing. So I believe all it requires is removing the import and checking the fonts are still loaded correctly.

aapeliv:
The main thing this does is replacing the `next/image` component with a current one. This skips next's "image optimization", which seems to be an extortion-style feature from them. Basically it's supposed to load the image and optimize it on the server side (seems to work well and be expensive on their hosted service), but for us it just creates unnecessary contention for the container. This replaces it with a minimal version that uses 3 different sizes based on screen width that load from the CDN.

We only use next/image in this one single place.

Closes #5415

- Load the frontend on different screen sizes and make sure the hero image works fine.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass


app/web/package.json
```diff
@@ -22,8 +23,6 @@
   "dependencies": {
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.0",
-    "@fontsource/mansalva": "^5.2.5",
-    "@fontsource/ubuntu": "^5.2.5",
```
nabramow, line 26: What's up here?

app/web/package.json
```diff
@@ -67,6 +66,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.26.9",
+    "@next/bundle-analyzer": "^15.4.0-canary.51",
```
nabramow, line 69: Isn't a canary a beta version?

app/web/pages/_document.tsx
```diff
@@ -8,12 +8,17 @@ export default class MyDocument extends Document {
...
       <Html lang={this.props.locale ?? "en"}>
         <Head>
+          <link rel="preconnect" href="https://cdn.couchers.org" />
           <meta name="theme-color" content={theme.palette.primary.main} />
           <link rel="manifest" href="/manifest.json" />
           <link rel="apple-touch-icon" href="/logo512.png" />
           <link
             rel="stylesheet"
-            href="https://fonts.googleapis.com/css2?family=Ubuntu:wght@400;500;700&display=swap"
+            href="https://cdn.couchers.org/fonts/ubuntu/ubuntu.css"
```
nabramow, line 17: Ah is this solving this one? https://github.com/Couchers-org/couchers/issues/5415

app/web/package.json
aapeliv: Loading them from our cdn

nabramow:
Do the other images need the three sizes as well? The one for the signup and login form for example.

app/web/package.json
aapeliv: I’m not sure, this installed with yarn add. It’s just a dev dep tho

app/web/pages/_document.tsx
aapeliv: Oh yeah

aapeliv:
> Do the other images need the three sizes as well? The one for the signup and login form for example.

Those ones seem to be fine

nabramow (approved):
Okay looks good to me!

Issue #4165, opened 2021-10-05
Add ability to block/unblock users


CouchersBot:
Add the ability to block/unblock users. This is necessary for user safety.

The backend is already done for this.The API is here: 
https://github.com/Couchers-org/couchers/blob/develop/app/proto/blocking.proto

**BLOCKING FLOW:**
- On a user's profile page, user clicks on teal flag icon:
<img width="321" alt="Image" src="https://github.com/user-attachments/assets/d3259a1b-a32d-43da-b68c-211a284008e7" />

- Add text "Report or Block" to the right of the flag icon

-  A modal pops up with this design:

<img width="517" alt="Image" src="https://github.com/user-attachments/assets/d444870b-69dc-4ae0-ba08-5e9f676f2af0" />

- The report checkbox should be activated by default, and the block user checkbox should be deactivated by default.

- If a user deactivates the report checkbox, the dropdown menu and input box collapse/disappear and it looks like this:

<img width="645" alt="Image" src="https://github.com/user-attachments/assets/01355ee8-97a8-4a9d-aa6d-5758ce8a5853" />

- Upon submitting, a basic confirmation dialog must be shown to prevent accidental reports or blocks.

- If the user unchecks both boxes, then the "Submit" button should become inactive (greyed out) since there is nothing to submit in that case. Alternatively, we could use an error alert to warn the user if they try to click the button when neither checkbox is activated. In this case, use this error alert text:
`To submit, at least one checkbox must be selected`

Notes:
If the API supports it, we could display "Block [name]" instead of the generic "Block this person".

**UNBLOCKING FLOW:**
-  Add to "my connections" page (https://couchers.org/connections). 
- The blocked users should appear in the bottom-right quadrant of the matrix (next to "friend requests you sent" and below "your friends"). 
- On mobile it should appear at the very bottom of the page. 

========================================
Subject: Blocking
Description:
Used website

Results:
It's essential to be able to block some members e.g. who may have been a previous bad experience on Couchsurfing...

Backend version: develop-d04e9dbe
Frontend version: develop-d04e9dbe
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:78.0) Gecko/20100101 Firefox/78.0
Page: https://app.couchers.org/
User: smersh (1396)

darrenvong:
Are we leaving this for support or should this actually be implemented in the frontend? Looks like there are some APIs available in `blocking.proto`

[...]

PR #6086, opened 2025-05-25
Add more data to GetBlockedUsers response

nabramow:
Part of #4165 

- Run backend
- Use .env.localdev if running against frontend
- You can probably just log the response for now, and make sure tests pass I did test against the frontend though)

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/blocking.py
+++ b/app/backend/src/couchers/servicers/blocking.py
@@ -2,7 +2,7 @@ import grpc
 from google.protobuf import empty_pb2
 from sqlalchemy.sql import union
 
-from couchers import errors
+from couchers import errors, urls
 from couchers.models import Upload, User, UserBlock
 from couchers.sql import couchers_select as select
 from proto import blocking_pb2, blocking_pb2_grpc
@@ -85,7 +85,7 @@ class Blocking(blocking_pb2_grpc.BlockingServicer):
                 blocking_pb2.BlockedUser(
                     username=blocked_user.username,
                     name=blocked_user.name,
-                    avatar_thumbnail_url=Upload(filename=blocked_user.filename).thumbnail_url
+                    avatar_thumbnail_url=urls.media_url(filename=blocked_user.filename, size="thumbnail")
                     if blocked_user.filename
                     else None,
                 )
--- a/app/backend/src/couchers/servicers/blocking.py
+++ b/app/backend/src/couchers/servicers/blocking.py
@@ -3,7 +3,7 @@ from google.protobuf import empty_pb2
 from sqlalchemy.sql import union
 
 from couchers import errors
-from couchers.models import User, UserBlock
+from couchers.models import Upload, User, UserBlock
 from couchers.sql import couchers_select as select
 from proto import blocking_pb2, blocking_pb2_grpc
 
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/blocking.py
```diff
@@ -72,17 +72,23 @@ def UnblockUser(self, request, context, session):
...
+            .where(UserBlock.blocking_user_id == context.user_id)
+        ).all()
 
         return blocking_pb2.GetBlockedUsersRes(
-            blocked_usernames=[blocked_user.username for blocked_user in blocked_users],
+            blocked_users=[
+                blocking_pb2.BlockedUser(
+                    username=blocked_user.username,
+                    name=blocked_user.name,
+                    avatar_thumbnail_url=Upload(filename=blocked_user.filename).thumbnail_url
```
aapeliv: Replace with `urls.media_url("thumbnail", blocked_user.filename)`

aapeliv (changes requested):
One minor thing!

app/backend/src/couchers/servicers/blocking.py
nabramow: Wait which part? Can you do a suggest changes?

aapeliv:
```suggestion
                    avatar_thumbnail_url=urls.media_url(filename=blocked_user.filename, size="thumbnail")
```

app/backend/src/tests/test_blocking.py
```diff
@@ -111,3 +122,7 @@ def test_relationships_userblock_dot_user(db):
         ).scalar_one_or_none()
         assert block.blocking_user.username == user1.username
         assert block.blocked_user.username == user2.username
+        assert block.blocking_user.name == user1.name
+        assert block.blocked_user.name == user2.name
+        assert block.blocking_user.avatar.thumbnail_url == user1.avatar.thumbnail_url
+        assert block.blocked_user.avatar.thumbnail_url == user2.avatar.thumbnail_url
```
nabramow: @aapeliv Do you have any idea what I need to do to get these tests passing? Both are failing due to these errors:

```
FAILED src/tests/test_blocking.py::test_GetBlockedUsers - sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <User at 0...
FAILED src/tests/test_blocking.py::test_relationships_userblock_dot_user - sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <User at 0...
```

Do I need to run a function similar to `generate_user` but to generate an upload?

app/backend/src/tests/test_blocking.py
```diff
@@ -91,12 +91,25 @@ def test_GetBlockedUsers(db):
...
+        blocked_avatar_urls = [
+            user.avatar_thumbnail_url if user.avatar_thumbnail_url else None for user in blocked_user_list.blocked_users
+        ]
+
+        assert user2.username in blocked_usernames
+        assert user3.username in blocked_usernames
+        assert user2.name in blocked_names
+        assert user3.name in blocked_names
+        assert user2.avatar.thumbnail_url in blocked_avatar_urls if user2.avatar else None
+        assert user3.avatar.thumbnail_url in blocked_avatar_urls if user3.avatar else None
```
aapeliv: Okay so the issue is this: when we create the user above with `user3, token3 = generate_user()`, we "detach" the user from the session. So normally the User object is attached to a "session" i.e. a database transaction. So that the ORM (that maps objects <> rows in the db) knows what `UPDATE` instruction to issue when you do `user.whatever = something`. When we make the user, we detach it from the session so we can use it outside of the `session_scope` (database transaction) that created it. That allows us still to access all the things that were loaded while it was attached (e.g. `username`) --- but we can no longer modify it or access things that require further SQL queries. SQLAlchemy by default lazy loads things like the avatar (because it refers to another database table, to avoid lots of joins everywhere). So when you try to do `user.avatar.thumbnail_url`, it tries to do a `SELECT filename FROM uploads WHERE key = (whatever)`, but now it's not connected to a database transaction where it could do that.

So that's the error you're getting, `Parent instance <User at 0x79f74af87610> is not bound to a Session`: that `user2` (or whichever) is not bound to a session: it's the parent of the `avatar` object. So it can't issue the SQL to grab the avatar: `lazy load operation of attribute 'avatar' cannot proceed`.

aapeliv: TLDR: since the avatars here are all null, the easiest way is to remove the checks of avatar.

Alternatively, if you give the users avatars, you can fetch the filename from the database manually. I would just do the former.

nabramow: That makes sense, thank you for the explanation! I've removed the avatar part of the tests for now.

aapeliv (approved):

PR #6093, opened 2025-05-26
Update sleeping arrangement meanings

### LINKED ISSUE #5679: Clarify sleeping arrangement meanings
Clarify what the sleeping arrangements mean on the frontend.

- Add it to the EditProfileDropdown (probably best just adjust the translation which is being used)
- Also add it to the Profile page "Sleeping arrangement" value.
- Filter dialog for map search (on new map if it's done)

Text should be:

- "Private space (e.g. own room or similar)"
- "Common room (e.g. couch in living room)" (change existing common space to common room also)
- "Shared room (e.g. shared bedroom with own sleeping surface)"

ColleenLillian:
Closes #5679 


ColleenLillian:
I just updated the translations, but I guess we want to leave the map search filter as it is though so just the short title (private/common/shared) is on the button?

nabramow:
Okay now that we see how it looks, I think what we want to do is put the little description below the sleeping arrangement row instead in small font, so it's not so squished:

Probably a Typography with variant="caption" would be a good size.

Screenshot 2025-05-26 at 8 57 02 PM

Let's see if that looks better!

nabramow:
For the explanations, rather than put it in the toggle boxes themselves, which looks a bit cramped, I'd add a key below that line with explanations in small font.

Common room - e.g. couch in living room
Private space - e.g. own room or similar
Shared room - e.g. shared bedroom with own sleeping surface

ColleenLillian:
thanks! yeah it looks better in a separate box below. I tried positioning under each button but it looked too cramped so it's just a small list

nabramow (dismissed):
Okay this is good to go!

app/web/features/profile/locales/en.json
```diff
@@ -105,9 +105,9 @@
     "yes": "Yes"
   },
   "sleeping_arrangement": {
-    "private": "Private space",
-    "common": "Common area",
-    "shared_room": "Shared room"
+    "private": "Private space (e.g. own room or similar)",
```
jesseallhands, line 111: I think "private_ space" is a bit too ambiguous. We should be explicit here to avoid any confusion:

```suggestion
    "private": "Private room (a room to yourself)",
```

app/web/features/search/locales/en.json
```diff
@@ -40,9 +40,18 @@
...
       "sleeping_arrangement_filters": {
-        "shared_room": "Shared room",
-        "private": "Private room",
-        "common": "Common room"
+        "shared_room": {
+    "label": "Shared room",
+    "description": " - e.g. shared bedroom with own sleeping surface"
+  },
+  "private": {
+    "label": "Private space",
```
jesseallhands, line 48:
```suggestion
    "label": "Private room",
```

app/web/features/search/locales/en.json
```diff
@@ -40,9 +40,18 @@
...
+    "description": " - e.g. own room or similar"
```
jesseallhands, line 49:
```suggestion
    "description": " - e.g. own bedroom or similar"
```

jesseallhands (changes requested):
This is looking great! 

I requested an update of the wording for `Private space` because I think that is too ambiguous. For example, someone might consider the other side of a paper room divider or curtain a "private space" when in reality it is actually a shared room with a divider that adds some additional privacy.  However, with this option what we mean is a room of some kind where the person will be sleeping alone and have the room entirely to themselves.

I noticed on the "my home" profile editing page on mobile, the "shared room" text doesn't wrap so it's not fully readable:
image

On desktop search filters I think the descriptions should be below the toggles because that is where users will be looking when they make their selection:
image

On mobile search filters page I think there might be an issue with the padding for small devices. This is what I see in the firefox mobile device viewer for some iphone 11 and 12 models, but maybe this is just firefox rendering weird:

image

On my HUGE phone (galaxy s24 ultra) it looks fine on firefox. On chrome it looks like this, which is pretty good but still has some padding issues:
image

nabramow:
@jesseallhands We talked about this in the meeting and decided to keep it as is for now.

nabramow:
Okay just to summarize what still needs to be done here as it's a bit confusing. From what I can see:

1. The key in the FilterDialog needs to move to the right side on bigger screens (and center on smaller screens)
2. The text overflows the select. It's not possible to wrap text in a select as Jesse suggested. so we need a solution to that.  Why suggestion? Take out the parentheses parts so it's shorter. Add the same key that we used in FilterDialog above or below that dropdown in the EditProfile MyHome section (whatever looks best).

Things we won't change in this PR:

- We'll keep the room descriptions as is for now, nothing to do there
- The padding is not related to sleeping arrangement, so we should make a separate ticket (unless Colleen can fix it quickly)

ColleenLillian:
I moved the key for the map filter but not sure how to fix the select dropdown. Is this in the EditHostingPreference form?

nabramow:
> I moved the key for the map filter but not sure how to fix the select dropdown. Is this in the EditHostingPreference form?

Yeah so in EditHostingPreference there's a variable `sleepingArrangementLabels` and that has the translations listed which are in `app/web/features/profile/locales/en.json`

These are the translations you originally added to 😅, like this:

```json
  "sleeping_arrangement": {
    "private": "Private space (e.g. own room or similar)",
    "common": "Common room (e.g. couch in living room)",
    "shared_room": "Shared room (e.g. shared bedroom with own sleeping surface)"
  },
  ```
  
 They're used in the Home and EditHostingPreferences section. I guess maybe we want to make a new "short" set. So like:
 
 ```json
  "sleeping_arrangement": {
    "private": "Private space (e.g. own room or similar)",
    "common": "Common room (e.g. couch in living room)",
    "shared_room": "Shared room (e.g. shared bedroom with own sleeping surface)",
    "private_short": "Private space",
    "common_short": "Common room",
    "shared_room": "Shared room"
  },
  ```
  
  And then make a new `sleepingArrangementLabelsShort` variable using the short ones, and switch that out in the `HostingPreferenceForm`.

ColleenLillian:
thanks! placed the list below the dropdown, it's quite small but I think still readable on mobile. lmk if that's ok for now!

nabramow:
I pushed a few small changes to make sure the profile non-edit view has the long version and the edit view has the short version. 

Once the tests pass this is good to go!

nabramow (approved):
Okay this is good to go! 🚀

nabramow (dismissed):

jesseallhands (changes requested):
This is looking good!

Here are the things I found in my testing:

1. Since there are so many different font sizes we should try to minimize the number of different sizes we have. My suggestion is to have the font size of the bed descriptions match the font size of the helper text:
Screenshot_20250628_213816_Firefox

I also noticed the descriptions have center alignment on mobile but on desktop they are right-aligned. They both should be right-aligned so they are directly under the options.

Mobile
Screenshot_20250628_223635_Firefox

Desktop
Screenshot_20250628_223645_Chrome

ColleenLillian:
do we not want it centered on mobile? that's an easy fix if so - I just took Nicole's suggestion literally :D 
p.s. I do agree that the edit profile section is a bit chaotic looking overall, maybe we can revisit the whole design of that section later

aapeliv (approved):

jesseallhands (approved):
LGTM

PR #6094, opened 2025-05-27
Block & Unblock User Functionality

### LINKED ISSUE #4165: Add ability to block/unblock users

### LINKED ISSUE #5962: Add "Report" next to flag on profile page
It's not clear what the flag is so we want to add the text "Report" next to the flag for the profile page only.

The comments should still only keep the flag.

The font size for this text should be small so it doesn't draw too much attention, but is clearly labelled.

nabramow:
Closes #4165 
Closes #5962 

- Go to your profile => my connections
- Block a friend by clicking the ... and "block user"
- That user should appear in your "Blocked users" and be removed from your friends
- Now unblock them
- They should be removed from your blocked users and go back to your friends
- Go to a user's profile
- Click "Report or block" and check to block them, then confirm
- Go back to your connections, they should be in your blocked users
- Block someone that has discussions, messages, etc. with you
- Check the map search, discussions, host requests, messages etc. to make sure you can no longer see them anywhere
- Go to a message with that user, you should no longer be able to write them a message
- Check group chat with the blocked user, should say "Unknown user"

**Web frontend checklist**
- [ x ] All tests pass


aapeliv:
If you block someone, then immediately go to their profile, it'll still show stuff. Maybe we need to purge the user from the query cache on block?

aapeliv:
Also the avatar in the unblock list links to the user's profile (which is no longer visible)

aapeliv:
Screenshot from 2025-06-01 09-37-59

- There's no snackbar/toast when you do block.
- If you go back to the user’s page it’s still there from cache (with error though), see above

app/web/resources/locales/en.json
```diff
@@ -69,14 +69,19 @@
       "success_message": "Thank you for your report and making Couchers a better platform for everybody!"
     },
     "flag": {
+      "block_user": "Block {{username}}",
+      "block_user_explainer": "Messages and requests from this person will be blocked, and your profile, events and comments will not be visible to this person.",
```
jesseallhands:
```suggestion
      "block_user_explainer": "Messages and requests from this person will be blocked, and your profile, events and comments will not be visible to this person. They will not be informed that you blocked them.",
```

jesseallhands (changes requested):
The functionality works on the profile page, but it doesn't work in the messages or requests/hosting pages:

image

I think it should probably work wherever flags are -- but of course we'd only have the "Report or Block" text on the profile pages.

app/web/features/connections/locales/en.json
```diff
@@ -14,5 +22,11 @@
     "title": "Remove friend",
     "message": "Are you sure you want to remove {{name}}? This action cannot be undone.",
     "confirm": "Remove"
+  },
+  "unblock_user": "Unblock user",
+  "unblock_user_confirmation_dialog": {
+    "title": "Are you sure you want to unblock {{name}}?",
+    "message": "Unblocking will allow them to see your profile and send you requests. If you were previously friends, they will go back into your friends list.",
```
jesseallhands:
```suggestion
    "message": "Unblocking will allow them to see your profile and send you requests. If you were previously friends, they will go back into your friends list. They will not be notified.",
```

app/web/features/connections/locales/en.json
```diff
@@ -1,10 +1,18 @@
 {
   "add_friend": "Add friend",
+  "blocked_list_title": "Blocked users",
+  "block_user": "Block user",
+  "block_user_confirmation_dialog": {
+    "title": "Are you sure you want to block {{name}}?",
+    "message": "This will prevent you both from seeing each others profiles. You will not be able to send requests or messages.",
```
jesseallhands:
```suggestion
    "message": "Messages and requests from this person will be blocked, and your profile, events and comments will not be visible to this person. They will not be informed that you blocked them.",
```

jesseallhands (changes requested):
I added a couple more things since I forgot to check out the connections page and the functionality there (which looks great and works well!) One thing I'm not sure of: @aapeliv will have to confirm whether events and discussions/replies authored by the person will be hidden or not. If not, those two words can be left out of my suggestions. EDIT: Aapeli confirmed they cannot see their events or discussions.

When blocking someone, it seems like a report is sent to support sometimes even if the person doesn't choose to also report the person. Here's an example of what I mean (note that there is no "reason" given):

https://chatwoot.couchershq.org/app/accounts/2/conversations/3379

nabramow:
> The functionality works on the profile page, but it doesn't work in the messages or requests/hosting pages:

> I think it should probably work wherever flags are -- but of course we'd only have the "Report or Block" text on the profile pages.

Hmm, this is not what was decided in the ticket. I also don't think that's usual behavior for blocking functionality. Let's leave it as is for now, we can discuss if we want to expand it later.

nabramow:
> When blocking someone, it seems like a report is sent to support sometimes even if the person doesn't choose to also report the person. Here's an example of what I mean (note that there is no "reason" given):

Oooo good catch, I fixed this now!

nabramow:
> * There's no snackbar/toast when you do block.

So this is trickier than it seems. Since after they block, the user profile is no longer visible to them, I can't render a snackbar there. I'd need to build some sort of higher up Toast at the app level, then pass it to there via React Context or something, but that's a bit of a feature in itself. I could redirect them to their connections page instead of the dashboard maybe?  Then they see the blocked user there.

> * If you go back to the user’s page it’s still there from cache (with error though), see above

This is fixed, I'm now routing them to the `NotFound` page if they go to a blocked profile.

aapeliv:
> > * If you go back to the user’s page it’s still there from cache (with error though), see above
> 
> This is fixed, I'm now routing them to the `NotFound` page if they go to a blocked profile.

I think it's better to leave as it was before. I don't understand react query well enough to know why it doesn't work and what corner of the cache the data is left to be in. But I think it's better not to have all that extra logic on the frontend just for this corner case (now I think this cause us to load blocked users every time you load the profile page).

nabramow:
> > > * If you go back to the user’s page it’s still there from cache (with error though), see above
> > 
> > 
> > This is fixed, I'm now routing them to the `NotFound` page if they go to a blocked profile.
> 
> I think it's better to leave as it was before. I don't understand react query well enough to know why it doesn't work and what corner of the cache the data is left to be in. But I think it's better not to have all that extra logic on the frontend just for this corner case (now I think this cause us to load blocked users every time you load the profile page).

@aapeliv Okay I fixed the blocked user issue, but I think we need to filter out the messages from the blocked user in the MessagesList on the backend. I changed the username and avatar to "Unknown user", but it seems we're filtering out the user, but not their messages:

Screenshot 2025-06-07 at 10 06 00 AM

aapeliv (dismissed):

jesseallhands (dismissed):
lgtm

aapeliv (approved):
LGTM

PR #6095, opened 2025-05-27
Add report flag to event cards

### LINKED ISSUE #6008: Add ability to report events
We have the ability to report comments, users, etc. We want to add the report flag to report events, as sometimes people make spammy ones.

Acceptance Criteria
- Add the `ReportFlag` to both types of events, the long event and the short event:

Image
Image

- Make sure this also shows within a community, the events tab

- For the long event card, the ReportFlag can overlay the event image at the bottom left with a white circle background.
- For tall event card the ReportFlag can go at the bottom right

ColleenLillian:
Closes #6008 
Updated to use styled


app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -6,6 +6,7 @@ import {
   Tooltip,
   Typography,
 } from "@mui/material";
+import { useTheme } from "@mui/material/styles";
```
nabramow: Why this switch? The other theme should do the same thing so better to keep that one if you don't mind!

nabramow:
Let's make the icon the small size on this one (the my event card):

Screenshot 2025-05-27 at 8 51 16 AM

nabramow:
Looks like the long event card image size got changed somehow in this PR. This is how it looks on prod:

Screenshot 2025-05-27 at 8 56 02 AM

and this is how it looks in this branch:

Screenshot 2025-05-27 at 8 54 02 AM


Probably something went awry when switching to styled format.

app/web/features/communities/events/LongEventCard.tsx
ColleenLillian: got it! just needed to make a styled component for the pill so this isn't needed any more

ColleenLillian:
> Let's make the icon the small size on this one (the my event card):
> 
do you mean on the long event card? or a separate component for events created by the user?

nabramow:
> do you mean on the long event card? or a separate component for events created by the user?

Yes on the LongEventCard! The one that shows under "My Events".

app/web/features/communities/events/LongEventCard.tsx
nabramow: Sweet yeah so doing `import theme from 'theme'` and `import { useTheme } from @mui/material` are the same thing, but the former is preferred!

nabramow:
@ColleenLillian Okay I pushed the fixed up events changes to this branch, so hopefully that fixes things!

This was my bad, it's one of the first features I did at Couchers and I did the tests in such a way that rather than return mock events, I was mocking the whole hook and not returning anything.

This was causing your ReportFlag to fail as it relies on there being an `eventId` and `eventCreatorId`. The way I did the tests originally meant these were undefined, causing the tests to fail and ReportFlag to throw errors for undefined values.

So I updated the tests to mock the way we usually mock in other tests!

nabramow:
> Looks like the long event card image size got changed somehow in this PR. This is how it looks on prod:
> 
> Screenshot 2025-05-27 at 8 56 02 AM
> and this is how it looks in this branch:
> 
> Screenshot 2025-05-27 at 8 54 02 AM
> Probably something went awry when switching to styled format.

@ColleenLillian Looks like there's still this issue with the image being stretched out compared to prod. Probably something got lost switching to styled.

Besides that could we make the whole circle with the flag on that style card a big smaller? So there's less padding between the flag and the edge of the circle. Trying to avoid announcing the report flag so strongly ;-).

app/web/features/FlagButton.tsx
```diff
@@ -76,6 +77,7 @@ export default function FlagButton({
   };
 
   const onSubmit = handleSubmit((data) => {
+    console.log("REPORT CONTENT RAN");
```
nabramow: Don't forget to remove this!

app/web/features/communities/events/EventCard.tsx
```diff
@@ -17,65 +17,66 @@ import { useMemo } from "react";
...
-    lineClamp: 5,
-    overflow: "hidden",
+  }),
+}));
+
+const CardWrapper = styled("div", {
+  shouldForwardProp: (prop) => prop !== "isCancelled",
+})({
+  position: "relative",
+});
```
nabramow: You are not passing `isCancelled` in the props to this element, not using it, so I think you can remove the `isCancelled` logic here. I'm also unsure if you need this element? We didn't have an additional wrapper before.

You can probably set the `CardMedia` element itself to `position: absolute` and/or give it `top/left` values.

app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -104,58 +135,70 @@ const LongEventCard = ({
...
-                : t("communities:virtual_event_location_placeholder")}
-
-              <div>{startTime}</div>
-            </div>
-            <div className={classes.attendees}>
-              {t("communities:attendees_count", {
-                count: event.goingCount + event.maybeCount,
+    <StyledCard data-testid="event-item">
+      <Link href={routeToEvent(event.eventId, event.slug)} passHref>
+        <CardLayout>
```
nabramow: Instead of adding an additional component here, which might affect layout, you can make the Link a `StyledLink` and pass the style there (since it's not a mui element with sx props).

app/web/features/communities/events/LongEventCard.tsx
```diff
@@ -104,58 +135,70 @@ const LongEventCard = ({
...
-        </CardContent>
+            />
+            <FlagWrapper>
+              <FlagButton
+                contentRef={`event/${event.eventId}`}
+                authorUser={event.creatorUserId}
+              />
+            </FlagWrapper>
+          </ImageWrapper>
+
```
nabramow: Okay I think this image wrapper is the culprit here. Remove the `ImageWrapper` added element. Then grab the `cardMedia` style from the former `useStyles` and put it in the `sx` of `CardMedia` intstead. That should fix the stretched image.

Then remove the `position: "absolute"` from the `FlagWrapper` that will make the flag visible again, then it's just a matter of adjusting the location of it in a different way.

For `FlagWrapper`, I think you could just do this and it looks okay. It's a bit different than what was originally described but since there's space to the left of the image...

```typescript
const FlagWrapper = styled("div")(({ theme }) => ({
  display: "flex",
  alignItems: "flex-end",
  cursor: "pointer",
  marginLeft: theme.spacing(1),
  marginBottom: theme.spacing(1),
  "& svg": {
    fontSize: 22,
  },
}));
```

ColleenLillian: thank you! Hopefully that fixes it, I don't see the image being stretched out though so maybe I was looking at a different view? I'm looking at the main Events page and Past Events

ColleenLillian:
btw I reduced the flag size on both and kept it to the side of the image on the long event card so it's more discreet.

nabramow:
Unfortunately it looks like the images are still messed up:

This is the long one:

Screenshot 2025-06-17 at 8 44 17 AM

And now the short one has new style issues:

Screenshot 2025-06-17 at 8 44 24 AM


It looks like maybe you made the style changes to the wrong event element? The short ones looked fine before, it was just the long ones at top that had issues. I'd undo whatever changes were made to the short ones undo Discover Events.

nabramow (approved):
Okay this looks good to go now! Thanks for bearing with me!

PR #6099, opened 2025-05-29
Switch NotificationFeed to update via mutations

### LINKED ISSUE #6056: Fix NotificationsFeed mark read - use mutation instead
When I originally did the NotificationsFeed, I am using `refetch` to cause the api to query again.

I since learned `react-query` better and realized that's not the best practice way.

Since we already called the API, we don't want to call it again. We just want to update the cache.

Acceptance Criteria
- In NotificationsFeed when marking a single item as read, create a mutation and update the cache for `is_seen` for just that specific item in the cache. Remove `refetch`.
- For "Mark all as read" , create a mutation and update the cache to update all items in the cache as `is_seen` true. Remove `refetch`.

nabramow:
Closes #6056 

- Mark a notification unread, you should see it change right away
- Mark notification as read by clicking on it (it will go to stage)
- Go back to previous page
- Open notifications feed and check that notification is marked read
- Mark all notifications as read via top right menu
- They should quickly change

**Web frontend checklist**
- [ x ] All tests pass

### DIFF
--- a/app/web/features/notifications/NotificationsFeed/NotificationItem.test.tsx
+++ b/app/web/features/notifications/NotificationsFeed/NotificationItem.test.tsx
@@ -1,8 +1,8 @@
 import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Empty } from "google-protobuf/google/protobuf/empty_pb";
+import mockRouter from "next-router-mock";
 import { service } from "service";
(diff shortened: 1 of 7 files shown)


aapeliv (commented):
Code looks good, didn't have a chance to test yet

aapeliv (approved):
LGTM and works well

PR #6138, opened 2025-06-03
Add notifications for nested replies

### LINKED ISSUE #5961: Notification for replies to a reply (nested)
We need a push notification when there are replies to a reply so the conversation doesn't get lost. This is a backend ticket.

spreeni:
closes #5961

**Backend checklist**
- [x] All tests pass


nabramow:
@aapeliv @spreeni feel free to tag me if this needs a frontend part! I think it might be okay as we already have the replies section now.

spreeni:
@nabramow Yes, this just changes the notify triggers and wording slightly, but should not require anything new from the frontend!

aapeliv:
Edit: oop, wrong PR, sorry

app/backend/src/couchers/servicers/threads.py
```diff
@@ -133,16 +133,20 @@ def generate_reply_notifications(payload: jobs_pb2.GenerateReplyNotificationsPay
             reply = session.execute(select(Reply).where(Reply.id == database_id)).scalar_one()
             # the comment we're replying to
             parent_comment = session.execute(select(Comment).where(Comment.id == reply.comment_id)).scalar_one()
+            thread_replies_author_user_ids = (
+                session.execute(select(Reply.author_user_id).where(Reply.comment_id == parent_comment.id))
```
aapeliv: Use `.where_users_column_visible(context, Reply.author_user_id)` here; and see next comment.

app/backend/src/couchers/servicers/threads.py
```diff
@@ -133,16 +133,20 @@ def generate_reply_notifications(payload: jobs_pb2.GenerateReplyNotificationsPay
...
             author_user = session.execute(select(User).where(User.id == reply.author_user_id)).scalar_one()
 
-            if are_blocked(session, parent_comment.author_user_id, reply.author_user_id):
-                return
-
-            if parent_comment.author_user_id == reply.author_user_id:
-                return
+            user_ids_to_notify = set(thread_user_ids) - {reply.author_user_id}
+            user_ids_to_notify = {
+                user_id for user_id in user_ids_to_notify if not are_blocked(session, user_id, reply.author_user_id)
```
aapeliv: ...then remove `are_blocked` from here (you do need it for the `reply.author_user_id` though).

Otherwise you'll do $n$ DB queries.

(Also currently this is broken: it checks whether the users above have blocked each other, not whether they've blocked the user posting now.)

aapeliv (changes requested):

app/backend/src/couchers/servicers/threads.py
spreeni: Oh nice, yeah that is helpful - added it!

app/backend/src/couchers/servicers/threads.py
spreeni: Removed it 👍  But from what I can see it would be fine to check if just one of them blocked the other which the function seems to do, no? But definitely agree that it is inefficient doing single DB queries here

aapeliv (approved):
LGTM!

PR #6141, opened 2025-06-07
Add screen resolution to bug report information

### LINKED ISSUE #5851: add screen resolution to bug report sent to github
### Problem:

When bugs are reported, the user agent is included, but not the screen resolution. This information would be helpful when reviewing bug reports, especially frontend-related bugs where the screen resolution can make a big difference.

### Solution:

When a bug is reported, the issue created in github should include the screen resolution data for the reporting user. The github issue would list "Screen Resolution:" followed by the resolution in pixels below the "User Agent:" label.

Image

spreeni:
closes #5851

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [x] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/bugs.py
+++ b/app/backend/src/couchers/servicers/bugs.py
@@ -43,6 +43,7 @@ class Bugs(bugs_pb2_grpc.BugsServicer):
             f"Backend version: {self._version()}\n"
             f"Frontend version: {request.frontend_version}\n"
             f"User Agent: {request.user_agent}\n"
+            f"Screen resolution: {request.screen_resolution.width}x{request.screen_resolution.height}\n"
             f"Page: {request.page}\n"
             f"User: {user_details}"
         )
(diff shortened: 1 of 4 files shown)


spreeni:
@nabramow Are `window.screen.width` and `window.screen.height` valid to retrieve screen size or is there a better way? Went ahead and quickly added the lines in the frontend as well.

aapeliv:
See also: https://stackoverflow.com/questions/37443482/what-is-the-difference-between-window-innerwidth-and-screen-width

Not sure which we should use?

nabramow:
> See also: https://stackoverflow.com/questions/37443482/what-is-the-difference-between-window-innerwidth-and-screen-width
> 
> Not sure which we should use?

@aapeliv @spreeni I usually use `innerWidth`! There's some other examples in the codebase too if you search `innerWidth`.

I think width is the width of the screen, so also outside the browser. Since we are in the browser, we'd want innerWidth.

spreeni:
@nabramow That makes sense, adjusted it 👍

aapeliv (approved):
LGTM

PR #6147, opened 2025-06-08
Update backend cookie logic to work better on dev

nabramow:
This PR updates the `_create_tasty_cookie` utility to only set the domain attribute on cookies when `COOKIE_DOMAIN` is explicitly configured.

Setting the domain to something like localhost (commonly configured in dev) makes the cookie invalid and causes it to be rejected by browsers.

By not setting the domain when unnecessary, cookies will still work on localhost, Vercel preview URLs, and other dynamic environments.

This improves reliability of cookie-based features (e.g., language preference, session management) in development and staging, while preserving cross-subdomain support in production.

**Backend checklist**
- [ x ] Formatted my code by running `ruff check --select I --fix . && ruff check . --fix && ruff format .` in `app/backend`
- [ ] All tests pass


nabramow:
Closing as found a better solution.

PR #6160, opened 2025-06-09
Couchers mission section of new landing page

### LINKED ISSUE #6016: Make Couchers mission section of new landing page
This PR is just for the Couchers mission section of the new landing page. Full designs can be seen here:

https://www.figma.com/design/5buI930COXegHPxOyHEwKo/Couchers-Landing-Page?node-id=67-6&t=xj3atKTlm5s7ZxLw-0

Acceptance Criteria
- Make a new branch off of the `web/new-landing-page` branch.
- Make a new component in `web/features/landing` called `<CouchersMission/>` and do your work there
- Follow the Figma designs above, you should do the design from the "Couchers mission" heading up to and including the "Volunteer" and "Donate" buttons
- When each green blurb is clicked on, it should change the text of the grey box below it.
- For any without text in the designs, just write "Fill in some text here" for now
- Put your component in the `LandingPage` component
- Open your PR off of the `web/new-landing-page` branch, ping @nabramow , @jesseallhands and @aapeliv to confirm final text.

---
Text: 
Title: Our plan/long term mission/values (?)

Then a text for each item (if this is what we want to show):
Non-profit structure
Community-first framework
Member accountability
An improved review system
Better host matching
Build it right
Want to help us grow?
Join as a volunteer or chip in with a donation – every bit helps keep Couchers.org running and improving.
For more details on out progress check our Roadmap

pcolt:
This should close #6016
Not sure if this will be included in the final landing page.

Very open to reviews being my second PR on the front-end, in particular:
- I do not understand why `spacing` prop is not working for Grid-container component even if the [documentation about Material UI v5](https://v5.mui.com/material-ui/react-grid/#spacing) says the opposite. I was forced to use `gap` and I was also forced to use strange values for `xs` and `sm` items. 
- I ended up using `sx` prop from System UI to style the custom Grid components in this section. I [read](https://v5.mui.com/material-ui/react-grid/#spacing) that in certain scenarios it may be better to use sx instead of styled() components. It also resembles more the fine grained component based styling I was accustomed with Tailwind CSS, but of course I don't want to break with any consistency or rule here.
- Test are not testing the translations itself (those are mocked) but only the front-end logic about clicking and displaying different content. Is that the right approach?
- Some texts in the descriptions are still missing @aapeliv @jesseallhands 

**Web frontend checklist**
- [x] All tests pass


app/web/features/landing/CouchersMission.tsx
```diff
@@ -0,0 +1,177 @@
...
+            transform: "translateX(-50%)",
+            width: 0,
+            height: 0,
+            borderLeft: "10px solid transparent",
+            borderRight: "10px solid transparent",
+            borderTop: `10px solid ${selectedItem === itemName ? theme.palette.primary.main : theme.palette.primary.light}`,
+            display: selectedItem === itemName ? "block" : "none",
+            [theme.breakpoints.down("md")]: {
+              display: "none",
+            },
```
nabramow: I would definitely make this a `StyledGridBubble` (or similar name) as it's so large. It helps to make it more readable in the actual component so we can see what's happening at a glance.

app/web/features/landing/CouchersMission.tsx
```diff
@@ -0,0 +1,177 @@
...
+        <Link href={volunteerRoute} passHref legacyBehavior>
+          <StyledButton component="a" variant="contained" color="secondary">
+            {t("global:nav.volunteer")}
+          </StyledButton>
+        </Link>
+        <Link href={donationsRoute} passHref legacyBehavior>
+          <StyledButton component="a" variant="contained">
+            {t("global:nav.donate")}
+          </StyledButton>
+        </Link>
```
nabramow: Does this work without the `legacyBehavior`? We have this in some places due to a package upgrade migration for old components, but when we make new ones, it's better to do it the "new" way.

It might work if you just take it out, but if not, see if you can figure out the new way they want.

nabramow:
The arrow/pointy bit is a little bit bigger in the design, which gives it a little more space between that and the text box. The text should also be white.

This is yours:
Screenshot 2025-06-09 at 9 54 38 PM

This is the design:

Screenshot 2025-06-09 at 9 54 27 PM

nabramow:
Looks like that first one is missing the spaces between the paragraphs. I also can't tell, is the font slightly bigger in the design? Hard to see, but I think regardless we can make this variant="body1" size.

Yours:
Screenshot 2025-06-09 at 9 56 30 PM

Design:
Screenshot 2025-06-09 at 9 58 22 PM

nabramow:
Just a few comments but overall this is looking really good!

app/web/features/landing/CouchersMission.tsx
pcolt: done

app/web/features/landing/CouchersMission.tsx
pcolt: After reading the [Next.js v15 documentation](https://nextjs.org/docs/pages/api-reference/components/link#legacybehavior), I think the correct behavior is to remove both `passHref` and `legacyBehavior` from the Link component together with the `component=a` from the nested button, otherwise it is mistakenly transformed in a nested `<a>` element, which is exactly what Next v15 Link does not want.

pcolt:
@nabramow I did all the fixes requested

app/web/features/landing/CouchersMission.tsx
nabramow: Thanks for looking into this!

nabramow (approved):
Okay this looks good to go! Nice work!

PR #6163, opened 2025-06-10
Add translator job description

nabramow:
Adding translator job description so we can link it next to the account settings language selector!

### DIFF
--- a/app/web/markdown/volunteer/translator.md
+++ b/app/web/markdown/volunteer/translator.md
@@ -32,7 +32,7 @@ Couchers.org is seeking native speakers to help us translate our website into:
 
 ### Duties
 
-- Our translations our done online via [Weblate](https://translate.couchershq.org/)
+- Our translations are done online via [Weblate](https://translate.couchershq.org/)
 - You can translate as much or as little as you want, you will see a list of sentences and phrases there to translate.
 - We communicate via Slack. There is a Slack channel for translation if you need to ask questions or get support!
 
--- a/app/web/markdown/volunteer/translator.md
+++ b/app/web/markdown/volunteer/translator.md
@@ -1,0 +1,49 @@
+---
+title: "Translator"
+description: "Help us translate Couchers.org!"
+---
+
+## [Apply Here](/volunteer/form)
+
+## Position Description
+
+**This is a remote position**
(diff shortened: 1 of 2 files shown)


app/web/markdown/volunteer/translator.md
```diff
@@ -0,0 +1,49 @@
...
+description: "Help us translate Couchers.org!"
+---
+
+## [Apply Here](/volunteer/form)
+
+## Position Description
+
+**This is a remote position**
+
+Couchers.org is seeking native speakers to help us translate our website into:
```
aapeliv, line 12: Personally I take a bit of issue with "native", and prefer "fluent" for reasons :P

app/web/markdown/volunteer/translator.md
```diff
@@ -0,0 +1,49 @@
...
+- Portuguese (Brazil)
+- Portuguese (Portugal)
+- Russian
+- Spanish (Spain)
+- Spanish (Latin America)
+- Want to translate a different language? Let us know!
+
+### Duties
+
+- Our translations our done online via [Weblate](https://translate.couchershq.org/)
```
aapeliv: "Are done"

aapeliv (changes requested):

app/web/markdown/volunteer/translator.md
nabramow: I swear I get dyslexic as I get older.

app/web/markdown/volunteer/translator.md
nabramow: I sort of think translators should be native speakers though? For example I am fluent in German but I'm not sure I should be a serious translator into German.

aapeliv: Then are you really fluent? :thinking:

aapeliv (approved):

app/web/markdown/volunteer/translator.md
nabramow: I think fluent can be like B2/C1, meaning can speak fluently without struggling to keep up, but there's still C2 above that.

I did used to translate from German to English but it's common that the native speakers do the target language.

PR #6170, opened 2025-06-10
Backend/feature/block blocked user messages

### LINKED ISSUE #6144: Filter out blocked users from messages & group messages list
Right now if a user is blocked, the user name will not be visible in the messages and group chats list, but the messages will still be visible.

We should filter out the blocked user's messages from the messages list and the group chats list

spreeni:
closes #6144

- New messages can't be sent if no other non-blocked/-deleted/-banned participant is in the chat
- Introduced new flag in GroupChats `can_message`
- Changed `are_blocked` to be `is_not_visible` (so also check deleted/banned)

**Backend checklist**
- [x] All tests pass


app/backend/src/couchers/errors.py
```diff
@@ -22,6 +22,7 @@
 CANT_INVITE_SELF = "You can't invite yourself to a chat."
 CANT_INVITE_TO_DM = "You can't invite other users to a direct message."
 CANT_MAKE_SELF_ADMIN = "You can't make yourself an admin."
+CANT_MESSAGE_IN_CHAT = "You can't message in this chat."
```
aapeliv: "You can't send a message in this chat."

app/backend/src/couchers/servicers/blocking.py
```diff
@@ -8,14 +8,18 @@
 from proto import blocking_pb2, blocking_pb2_grpc
 
 
-def are_blocked(session, user1_id, user2_id):
+def is_not_visible(session, user1_id, user2_id):
     blocked_users = (
         select(UserBlock.blocked_user_id)
+        .join(User, User.id == UserBlock.blocked_user_id)
+        .where(User.is_visible)
```
aapeliv: You need to do the opposite here (filter out users who are not visible). It might be easiest to add another query to the `union` below; so something like `select(User.id).where(User.id.in_([user1_id,user2_id])).where(not_(User.is_visible))`

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -119,6 +118,24 @@ def _get_visible_admins_for_subscription(subscription):
         ]
 
 
+def _user_can_message(session, context, group_chat: GroupChat) -> bool:
+    """
+    User can message in a group chat they are part of if a chat participant exists that:
```
aapeliv: I like what you've done here.

On the other hand, I think the logic should probably be uniform for DMs vs Real Group Chats. I.e. if I'm in a group chat with 2 banned users I should still be able to message, because it's an RGC and not a DM.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -119,6 +118,24 @@ def _get_visible_admins_for_subscription(subscription):
...
+    - Has not been blocked by the user or is blocking the user
+    - Has not left the chat
+    """
+    return session.execute(
+        func.exists(
+            select(GroupChatSubscription)
+            .where_users_column_visible(context=context, column=GroupChatSubscription.user_id)
+            .where(GroupChatSubscription.user_id != context.user_id)
+            .where(GroupChatSubscription.group_chat_id == group_chat.conversation_id)
+            .where(GroupChatSubscription.left == None)
```
aapeliv, line 136: Also would be good to add the joined check here just for completeness.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -136,13 +153,12 @@ def generate_message_notifications(payload: jobs_pb2.GenerateMessageNotification
...
             return []
 
-        subscriptions = (
+        context = make_user_context(user_id=message.author_id)
+        subscription_user_ids = (
             session.execute(
-                select(GroupChatSubscription)
-                .join(User, User.id == GroupChatSubscription.user_id)
+                select(GroupChatSubscription.user_id)
+                .where_users_column_visible(context=context, column=GroupChatSubscription.user_id)
```
aapeliv, line 162: Thanks for cleaning this up!

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -156,19 +172,18 @@ def generate_message_notifications(payload: jobs_pb2.GenerateMessageNotification
         else:
             msg = f"{message.author.name} sent a message in {group_chat.title}"
 
-        for subscription in subscriptions:
-            if are_blocked(session, subscription.user_id, message.author.id):
-                continue
+        user_ids_to_notify = set(subscription_user_ids) - {message.author_id}
```
aapeliv: It's a bit cleaner IMO if you put the `.where(GroupChatSubscription.user_id != message.author_id)` back, then you don't need to worry about the sets etc.

app/backend/src/couchers/servicers/conversations.py
spreeni: Hmm yeah, was thinking about that, if it makes sense to be able to chat in a group chat where nobody can interact with you. But changed it so you can always message in a group chat now..

app/backend/src/couchers/errors.py
spreeni: Done 👍

app/backend/src/couchers/servicers/blocking.py
spreeni: Oh, good spot! Adjusted it

app/backend/src/couchers/servicers/conversations.py
spreeni: What do you mean? What would you like me to check here?

app/backend/src/couchers/servicers/conversations.py
spreeni: Sure! 😊

app/backend/src/couchers/servicers/conversations.py
spreeni: True - for some reason I thought I needed another join with the User table, but you're right 👍

spreeni (commented):
Implemented all proposed adjustments except for the joined condition, where I am unsure what exactly you have in mind.

app/backend/src/couchers/servicers/conversations.py
aapeliv: Actually, I think it's fine, ignore this!

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -119,6 +118,26 @@ def _get_visible_admins_for_subscription(subscription):
...
+def _user_can_message(session, context, group_chat: GroupChat) -> bool:
+    """
+    User can message in a group chat they are part of if it's not a DM, or if a chat participant exists that:
+    - Is not deleted/banned
+    - Has not been blocked by the user or is blocking the user
+    - Has not left the chat
+    """
+    if not group_chat.is_dm:
+        return True
+    return session.execute(
```
aapeliv, line 130: For us thick ones, could you just add a comment saying "if it's not a DM (a true group chat), we can always message", then "otherwise... etc"

also i think the comment above is wrong, says "if it's NOT a DM"

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -569,21 +589,27 @@ def CreateGroupChat(self, request, context, session):
...
             .where(GroupChatSubscription.user_id == context.user_id)
             .where(GroupChatSubscription.left == None)
-        ).scalar_one_or_none()
-        if not subscription:
+        ).one_or_none()
+        if not result:
             context.abort(grpc.StatusCode.NOT_FOUND, errors.CHAT_NOT_FOUND)
 
+        subscription, group_chat = result
+        if _user_can_message(session, context, group_chat) is False:
```
aapeliv: `if not _user_can_message...`

aapeliv (changes requested):

app/backend/src/couchers/servicers/conversations.py
spreeni: No, I think the criticism is fair 😆 Adjusted it, have a look again.

app/backend/src/couchers/servicers/conversations.py
spreeni: Done 👍

app/backend/src/couchers/servicers/blocking.py
```diff
@@ -1,14 +1,17 @@
...
 from couchers import errors, urls
 from couchers.models import Upload, User, UserBlock
 from couchers.sql import couchers_select as select
 from proto import blocking_pb2, blocking_pb2_grpc
 
 
-def are_blocked(session, user1_id, user2_id):
+def is_not_visible(session, user1_id, user2_id) -> bool:
+    """
+    Check if user2 is not visible to user1 (due to block from either direction or because account is deleted/banned).
```
aapeliv: I would make this function symmetric, just need to add the `where` clause in `hidden_users` as before. Otherwise it's a bit dangerous and you'd need to audit the usages carefully.

Otherwise please make it very explicit with keyword only args:

```python3
def is_not_visible(session, *, this_user=user1_id, other_user=user2_id) -> bool:
```

aapeliv (changes requested):

app/backend/src/couchers/servicers/blocking.py
spreeni: Fair, I guess there is not really a use case where there can be a notification outgoing from an invisible user. Made it symmetric 👍

aapeliv (approved):
LGTM!

PR #6176, opened 2025-06-14
Add Makefile for backend commands, ListMembers pagination

### LINKED ISSUE #5776: Add pagination to UserDetails component

### LINKED ISSUE #5807: Add pageNumber pagination and option to order by created_at to ListMembers query

### LINKED ISSUE #6175: Change ListMembers to pagination and desc
* Change ListMembers to use `page_number` pagination on the backend
* Returns members in descending order so newer members first
* Add a Makefile with backend commands because I'm tired of having to search the ruff and test commands every time ;-)

nabramow:
Closes #6175 

Switching the community members list to use the nextPageToken and memberCount to trick it into pageNumber pagination. It will also fix that error when there's a lot of users.

Users are now returned in descending order.

I added a `Makefile` so we can run backend commands for formatting and testing, etc. I got tired of looking up the commands every time. ;-)

- Run backend locally `docker compose up --build`
- Replace contents of `.env.development` with contents of `.env.localdev`
- Log in with a local user
- Go to communities members page, it should have pageNumber pagination now
- Go into `backend/` folder, run `make format` and it should format
- Run `make test file=test_communities.py`, it should run just the communties tests
- Run `make test` should run all tests
- Run `make test file=test_communities.py::TestCommunities::test_ListMembers` should run just that test function

**Backend checklist**
- [x] All tests pass

**Web frontend checklist**
- [x] All tests pass


app/backend/src/couchers/servicers/communities.py
```diff
@@ -244,26 +244,39 @@ def RemoveAdmin(self, request, context, session):
...
-                .where(User.id >= next_member_id)
-                .order_by(User.id)
-                .limit(page_size + 1)
+                .order_by(User.id.desc())  # Show the most recent members first, given ids increment
+                .offset(offset)
+                .limit(page_size)
             )
             .scalars()
             .all()
         )
```
nabramow: I know it's not quite as performant as we're running the query twice. But except for the global community we're not going to have millions in there so I think it's worth it?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -244,26 +244,39 @@ def RemoveAdmin(self, request, context, session):
...
+            select(func.count(User.id))
+            .select_from(User)
+            .join(ClusterSubscription, ClusterSubscription.user_id == User.id)
+            .where_users_visible(context)
+            .where(ClusterSubscription.cluster_id == node.official_cluster.id)
+        ).scalar_one()
+
         members = (
             session.execute(
                 select(User)
```
nabramow: Can we use `LiteUser` on the backend here?

docs/contributing.md
```diff
@@ -58,7 +58,7 @@ All python code should live in the `couchers` namespace (i.e. a folder within th
...
 
-`ruff` is installed automatically if you install the requirements on your computer (or you can install it with `pip install ruff`, e.g. if you work with Docker). Run `ruff check --select I --fix . && ruff check . && ruff format .` in the `//app/backend` folder before you commit (or before asking for review) so that it picks up the config in `pyproject.toml`.
+`ruff` is installed automatically if you install the requirements on your computer (or you can install it with `pip install ruff`, e.g. if you work with Docker). Run `make format` in the `//app/backend` folder before you commit (or before asking for review) so that it picks up the config in `pyproject.toml`.
 
 You can run `ruff` linting and autoformatting automatically before each commit via `pre-commit` (It comes with the dependencies, or you can install it via `pip install pre-commit`). For this you have to once run `pre-commit install`. If you don't want to run the pre-commit hook, you can skip commit hooks with the `--no-verify` flag: `git commit --no-verify -m "commit message"`.
```
nabramow, line 63: I know I didn't follow my own rule of mixing two things together, haha. I can break it into a separate PR if we want.

app/backend/src/couchers/servicers/communities.py
nabramow: Another option would be if I pass `last_user_id` and then query ` < last_user_id` instead of using offset. That would be more efficient on large datasets.

app/backend/src/couchers/servicers/communities.py
aapeliv: Yes

app/backend/src/couchers/servicers/communities.py
aapeliv: Yes could we not do pagination here please? I think we can get basically the same experience by just listing users in the chosen order with infinite pages. You can also show the number of users (which is approximate), which you can get from `member_count` in the community object (it doesn't count people you blocked/who blocked you though).

Otherwise this will be slow, the `total_items` query now is ~150 ms on prod for the global community and that will get worse.

You will also have issues when new users sign up with the page shifting unless you do it cursor-style.

docs/contributing.md
aapeliv: Which rule? This is fine!

aapeliv (changes requested):

app/backend/src/couchers/servicers/communities.py
nabramow: I think I will keep it for now as I want to change the community member cards to have more info soon

app/backend/src/couchers/servicers/communities.py
```diff
@@ -244,26 +244,25 @@ def RemoveAdmin(self, request, context, session):
...
+            .where(ClusterSubscription.cluster_id == node.official_cluster.id)
         )
+        if next_member_id is not None:
+            query = query.where(User.id < next_member_id)
+        members = session.execute(query.order_by(User.id.desc()).limit(page_size + 1)).scalars().all()
+
         return communities_pb2.ListMembersRes(
             member_user_ids=[member.id for member in members[:page_size]],
-            next_page_token=str(members[-1].id) if len(members) > page_size else None,
+            next_page_token=str(members[page_size - 1].id) if len(members) > page_size - 1 else None,
```
aapeliv: The original here still works. This will now give an incorrect answer though if the length of the list is exactly the number of items left (i.e. you list 10 members out of 10, it will return a next page token even if there is no next page)

app/web/features/communities/members/CommunityMembersList.tsx
```diff
@@ -46,50 +22,73 @@ const LoadMoreButtonWrapper = styled("div")(({ theme }) => ({
...
-  );
+  const { data, isFetching, isLoading, error, fetchNextPage } = useListMembers({
+    communityId,
+    pageSize: PAGE_SIZE,
+  });
+
+  const memberUserIdsList = data?.pages
+    ? data.pages[pageNumber - 1]?.memberUserIdsList
+    : [];
+  const numPages = Math.ceil((memberCount ?? 0) / PAGE_SIZE);
```
aapeliv: The member count is approximate, so this will not give the exact right number.

app/web/features/communities/members/CommunityMembersList.tsx
```diff
@@ -46,50 +22,73 @@ const LoadMoreButtonWrapper = styled("div")(({ theme }) => ({
...
+
+  const handlePageNumberChange = (
+    event: React.ChangeEvent<unknown>,
+    value: number,
+  ) => {
+    if (value > pageNumber) {
+      fetchNextPage();
+    }
```
aapeliv: OK, I don't understand the page numbering (how you are computing where to jump if I click on a given page number).

I think it's better if we have infinite pages for the time being, given we can't compute the exact page numbers.

On prod we'll have a bunch of gaps due to deleted/banned users, so we can't reliably compute the right number. On next it has "1, 2, 3, ... 9" in global, but the last page never loads (I know you changed the order, but the API didn't change, it still does the same thing, just opposite order, I think?)

aapeliv (changes requested):

app/web/features/communities/members/CommunityMembersList.tsx
nabramow: Just to clarify how react-query does the page numbering. They return multiple pages at a time. So if the pageSize was 2, they return:

```json
{ 
    "pages": [
        [9,8],
        [7,6],
        [5,4],
        [3,2],
        [1]
    ]
}
```

So the way it's working is I'm keeping track of the pageNumber in useState and then referencing `pages[pageNumber]`.

I see what you mean about deleted/banned users though. I didn't realize how they were connected when you first explained that. So basically memberCount will include the blocked/deleted users even though the query does not return them?

aapeliv: `memberCount` does not include deleted/banned users but it *does* include banned users. That is one of the very few things we cache in the backend (for a few minutes) due to the latency of a full table scan to compute it on the fly otherwise. Because we cache it globally, it also does not include the users visible to others but not this user.

The problem with the page numbers though is that if users 7,3, and 2 are gone (blocked/banned/deleted), then I guess when you query those pages, you will receive something like this:

```json
{ 
    "pages": [
        [9,8],
        [6,5],
        [5,4],
        [1],
        [1]
    ]
}
```

(e.g second last line: if I search for users where id <= 3, I'll get just 1. But the last page will also be "two users with id <= 1".)

aapeliv (approved):
LGTM

PR #6178, opened 2025-06-15
feat(events): added optional should_notify flag in UpdateEventReq, updated proto.

### LINKED ISSUE #4414: Make event update notifications optional
Right now when host changes anything in the event it emails everyone attending. We want to allow the host to choose whether everyone gets notified or not. This is the backend piece to allow that.

* Add to `EventUpdateReq`: `shouldNotify`
* Change the update event API handler to not call notify() if this is false

andrei-kolesnik-california:
Description

Fixed logic in UpdateEvent to handle should_notify correctly and adjusted handling of UpdateEventReq in the backend service.
Also updated the proto definition to reflect the new optional flag.
Closes #4414

Basic functionality is unchanged: if the value is missing or set to true, notifications are sent. Only if the value is explicitly false, no notification is sent. Because of this, additional tests may not be necessary.

How to test
    1. Send a valid UpdateEventReq with different values for should_notify using Kreya or gRPCurl.
    2. Confirm expected behavior:
        ◦ If should_notify=true or the field is not provided, a notification email is triggered.
        ◦ If should_notify=false, no notification is sent.
    3. Check the logs or the database to verify that the event was updated correctly.
    4. (Optional) Log in as another user and verify the changes.

### DIFF
--- a/app/proto/events.proto
+++ b/app/proto/events.proto
@@ -212,7 +212,7 @@ message UpdateEventReq {
   google.protobuf.StringValue timezone = 11;
 
   // whether to notify all attendees about the change
-  optional bool should_notify = 12;
+  bool should_notify = 12;
 }
 
 message GetEventReq {
--- a/app/proto/events.proto
+++ b/app/proto/events.proto
@@ -212,7 +212,7 @@ message UpdateEventReq {
   google.protobuf.StringValue timezone = 11;
 
   // whether to notify all attendees about the change
-  // bool notify_attendees = 12;
+  optional bool should_notify = 12;
 }
 
 message GetEventReq {
--- a/app/backend/src/couchers/servicers/events.py
+++ b/app/backend/src/couchers/servicers/events.py
@@ -674,17 +674,20 @@ class Events(events_pb2_grpc.EventsServicer):
         session.flush()
 
         if notify_updated:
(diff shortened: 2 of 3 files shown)


app/proto/events.proto
```diff
@@ -212,7 +212,7 @@ message UpdateEventReq {
   google.protobuf.StringValue timezone = 11;
 
   // whether to notify all attendees about the change
-  // bool notify_attendees = 12;
+  optional bool should_notify = 12;
```
aapeliv: `optional` is a bit of a mess in protobuf. We don't use it elsewhere in the codebase (because when we started it wasn't a thing in proto3). I'm not opposed to using it, but we should test it though.

So proto3 originally had this somewhat stupid thing where on the wire it would not track presence: missing values would just be their default (the default being 0/false/empty/etc). I guess now they added it back in, so I suppose they must be serializing missing values as well.

aapeliv (commented):
Looks good but please add tests. You can just adapt existing ones as needed, no need to write from scratch.

app/proto/events.proto
andrei-kolesnik-california: Changed should_notify to required, since defaulting to False could result in users missing important updates.
Frontend must be updated to always send this field!
Also added a test to verify:
    No background queue changes when should_notify is False
    A job is queued when should_notify is True

aapeliv (approved):
LGTM!

PR #6211, opened 2025-06-20
Backend/feature/send direct message

andrei-kolesnik-california:
Implements SendDirectMessage: checks if a DM chat exists between two users; if not, creates a conversation and chat, subscribes both users, and sends the message. Returns the group_chat_id.
The created pytest is the best way to verify functionality.


app/backend/src/couchers/servicers/conversations.py
```diff
@@ -619,6 +619,72 @@ def SendMessage(self, request, context, session):
...
+            .group_by(GroupChatSubscription.group_chat_id)
+            .having(func.count(GroupChatSubscription.user_id) == 2)
+        )
+
+        chat = (
+            session.execute(
+                select(GroupChat).where(GroupChat.is_dm == True).where(GroupChat.conversation_id.in_(dm_chat_ids))
+            )
+            .scalars()
+            .first()
```
aapeliv: Just a word of warning: SQLalchemy does not render `LIMIT 1` with `.first()`! I found this the hard way and I suggest we don't use `first()` in the codebase because it's evil.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -619,6 +619,72 @@ def SendMessage(self, request, context, session):
...
+            .where(GroupChatSubscription.user_id.in_([user_id, recipient_id]))
+            .group_by(GroupChatSubscription.group_chat_id)
+            .having(func.count(GroupChatSubscription.user_id) == 2)
+        )
+
+        chat = (
+            session.execute(
+                select(GroupChat).where(GroupChat.is_dm == True).where(GroupChat.conversation_id.in_(dm_chat_ids))
+            )
+            .scalars()
```
aapeliv: I think it might be better to extract out a `_get_direct_message_subscription(session, this_user, other_user)` or similar from lines 533-550 above, then reuse it here. That would need some minor tweaking: I think the other case where that's used does not do anything with it, so you might want to change it to return the subscription id so you could use it below.

The SQL queries here are a bit intricate and therefore error prone, I think having them centralized is a good idea.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -619,6 +619,72 @@ def SendMessage(self, request, context, session):
...
+                )
+            session.flush()
+
+            # Retrieve the sender's active subscription to the chat
+        subscription = session.execute(
+            select(GroupChatSubscription)
+            .where(GroupChatSubscription.group_chat_id == chat.conversation_id)
+            .where(GroupChatSubscription.user_id == user_id)
+            .where(GroupChatSubscription.left == None)
+        ).scalar_one()
```
aapeliv: And here you could extract lines 553-577 into something like `_create_group_chat(session, this_user, recipient_user_ids)` that returns the subscription so you can pass it directly to `_add_message_to_subscription`

app/backend/src/couchers/servicers/conversations.py
andrei-kolesnik-california: Totally agree.

app/backend/src/couchers/servicers/conversations.py
andrei-kolesnik-california: Thank you very much!

app/backend/src/couchers/servicers/conversations.py
andrei-kolesnik-california: Let me do that.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -218,6 +218,46 @@ def _add_message_to_subscription(session, subscription, **kwargs):
...
+
+def _get_direct_message_subscription(session, user_id, conversation_id):
+    subscription = session.execute(
+        select(GroupChatSubscription)
+        .where(GroupChatSubscription.group_chat_id == conversation_id)
+        .where(GroupChatSubscription.user_id == user_id)
+        .where(GroupChatSubscription.left == None)
+    ).scalar_one_or_none()
+
+    return subscription
```
aapeliv: Please change also the existing functions to use these new helper functions

aapeliv (approved):
Nice!!

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -619,14 +637,44 @@ def SendMessage(self, request, context, session):
...
-    def EditGroupChat(self, request, context, session):
-        subscription = session.execute(
-            select(GroupChatSubscription)
-            .where(GroupChatSubscription.user_id == context.user_id)
-            .where(GroupChatSubscription.group_chat_id == request.group_chat_id)
-            .where(GroupChatSubscription.left == None)
+    def SendDirectMessage(self, request, context, session):
+        if request.text == "":
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, errors.INVALID_MESSAGE)
+
```
aapeliv: Please add the check on incomplete profile

aapeliv (changes requested):
Nice!!

aapeliv (dismissed):
LGTM!

aapeliv (approved):
LGTM!

andrei-kolesnik-california:
Thank you so much!

PR #6235, opened 2025-06-26
Rework report reasons

### LINKED ISSUE #6169: Rework report reasons
See comment from Jesse below. Update the frontend report list to reflect this.

nabramow:
Closes #6169 

- Safety, other or rules violation should not allow submit until filled out
- The others should not be required
- The report should actually send to moderation

**Web frontend checklist**
- [ x ] All tests pass


aapeliv:
If it's easy: would you be able to also make sure the reason gets sent in english (right now on prod it will send in the language of the translation)

nabramow:
> If it's easy: would you be able to also make sure the reason gets sent in english (right now on prod it will send in the language of the translation)

Okay this should work now!

aapeliv (changes requested):
The test is failing on gitlab, also I'm getting a bunch of missing strings:

image

aapeliv (approved):
LGTM

jesseallhands (commented):
To make reporting as easy as possible, users should not be required to input text unless they select the "other" option. All other options should have the text field be optional. 

If possible, when "other" is selected, the text input box should be labelled with "required", otherwise the label should be "optional".

PR #6236, opened 2025-06-26
Add upcoming event reminders with templates, migration, and tests

### LINKED ISSUE #6158: Upcoming event reminders
In backend:

* When a user is set to attend an event, and
* The event is 24 hours away

Then we should send them a reminder that it's coming up.

We should be careful that if you set yourself to "attending" < 24 hours before the event, then no reminder is sent.

Requires:

* New email template
* New notification type
* New background job
* Adding a migration

andrei-kolesnik-california:
This PR close #6158 - adds email reminders for upcoming events. Emails are sent to users who have marked themselves as attending an event that starts in the next 24 hours. Includes:
   - New notification type: event:reminder
   - Email templates (.mjml, .txt, and precompiled .html)
   - Migration to register the new notification type
   - Test for verifying email delivery content
Manual testing instructions:
   - Create an event starting ~24 hours from now.
   - Mark a user as “going” to the event.
   - Trigger the background job.
   - Confirm that the reminder email is generated and sent.
   - Email subject and body should include event title and formatted start time.
   
🟡 PR Requires review of email templates by the person responsible for external communication and mail content.


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,61 @@ def gen_randomized_coords(user_id, lat, lng):
...
+        )
+
+        attendees = []
+        for occurrence in occurrences:
+            attendees = (
+                session.execute(
+                    select(User)
+                    .join(EventOccurrenceAttendee, EventOccurrenceAttendee.user_id == User.id)
+                    .where(EventOccurrenceAttendee.occurrence_id == occurrence.id)
+                    .where(EventOccurrenceAttendee.attendee_status == "going")
```
aapeliv: The two options are going/maybe. I think we can send to both

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,61 @@ def gen_randomized_coords(user_id, lat, lng):
...
+def send_event_reminders(payload: empty_pb2.Empty):
+    """
+    Sends reminders for events that are 24 hours away to users who marked themselves as attending.
+    """
+    logger.info("Sending event reminder emails")
+
+    with session_scope() as session:
+        now_time = now()
+        lower_bound = now_time + timedelta(hours=23.5)
+        upper_bound = now_time + timedelta(hours=24.5)
```
aapeliv: The background job is not guaranteed to run exactly every 60 minutes (and one cannot in general guarantee that either), so this should use some kind of mechanism to guarantee idempotency.

I suppose the two options are: add a flag to the event, or add a flag to the user occurrence. The flag means keeping track of if we've sent a reminder yet.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,61 @@ def gen_randomized_coords(user_id, lat, lng):
...
+                    topic_action="event:reminder",
+                    data=notification_data_pb2.EventReminder(
+                        occurrence_id=occurrence.id,
+                        start_time=occurrence.start_time.isoformat(),
+                        event=events_pb2.Event(
+                            event_id=occurrence.id,
+                            slug=occurrence.event.slug,
+                            title=occurrence.event.title,
+                            start_time=Timestamp(seconds=int(occurrence.start_time.timestamp())),
+                        ),
```
aapeliv: Please use `event_to_pb(session, occurrence, context)` here

app/proto/notification_data.proto
```diff
@@ -178,6 +178,12 @@ message ReferenceReminder {
   uint32 days_left = 3;
 }
 
+message EventReminder {
+  int64 occurrence_id = 1;
+  string start_time = 2;
+  org.couchers.api.events.Event event = 3;
```
aapeliv: I think the event object should have all the info you need

aapeliv (changes requested):

app/backend/templates/v2/event_reminder.txt
```diff
@@ -0,0 +1,11 @@
+Hi {{ user.name }},
+
+This is a reminder for the upcoming event:
```
jesseallhands:
```suggestion
This is a reminder that the upcoming event is happening soon:
```

app/backend/templates/v2/event_reminder.txt
```diff
@@ -0,0 +1,11 @@
...
+
+{{ event_title }}
+Starts at {{ start_time }} UTC
```
jesseallhands: I don't think anyone except for tech people will understand UTC. Can we show the start time in the event's local time? Could say something like `Starts at {{ start_time }} local time`

app/backend/templates/v2/event_reminder.txt
```diff
@@ -0,0 +1,11 @@
...
+
+Don’t forget to join on time!
```
jesseallhands:
```suggestion
Don’t forget to attend!
```

jesseallhands (changes requested):
I didn't test the functionality, but assuming it uses the same template as the other emails we should be good. I requested a few wording changes.

I also think we should change the timing from 24 hours before the event to 12 hours before the event.

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: Agreed, makes sense to include both 'going' and 'maybe' – will update the query accordingly.

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: You're right — will add a flag to track whether a reminder was already sent.

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: Will do

app/proto/notification_data.proto
andrei-kolesnik-california: Yes, agreed — no need to duplicate fields

app/backend/templates/v2/event_reminder.txt
andrei-kolesnik-california: Yes, sounds better — clearer and more natural.

app/backend/templates/v2/event_reminder.txt
andrei-kolesnik-california: Yes, good point. I'll update the text to remove UTC and clarify it's local.

app/backend/templates/v2/event_reminder.txt
andrei-kolesnik-california: Agreed, 'attend' is more general and appropriate.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,56 @@ def gen_randomized_coords(user_id, lat, lng):
...
+
+        occurrences = (
+            session.execute(
+                select(EventOccurrence)
+                .where(EventOccurrence.start_time >= lower_bound)
+                .where(EventOccurrence.start_time <= upper_bound)
+            )
+            .scalars()
+            .all()
+        )
```
aapeliv: Look for all events happening in the next 24 hours, that way if this bg job doesn't run for some reason for a bit, it will still work (and the `reminder_sent` flag will stop multiple emails.

```python
        occurrences = (
            session.execute(
                select(EventOccurrence)
                .where(EventOccurrence.start_time <= func.now() + EVENT_REMINDER_TIMEDELTA)
                .where(EventOccurrence.start_time >= func.now())
            )
            .scalars()
            .all()
        )
```

bonus points if you put `EVENT_REMINDER_TIMEDELTA` in `constants.py`

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,56 @@ def gen_randomized_coords(user_id, lat, lng):
...
+
+        for occurrence in occurrences:
+            results = session.execute(
+                select(User, EventOccurrenceAttendee)
+                .join(EventOccurrenceAttendee, EventOccurrenceAttendee.user_id == User.id)
+                .where(EventOccurrenceAttendee.occurrence_id == occurrence.id)
+                .where(EventOccurrenceAttendee.attendee_status.in_(["going", "maybe"]))
```
aapeliv: Better to use `AttendeeStatus.going` and `AttendeeStatus.maybe` instead of strings. Also these are the only options, so you don't really need this `where`

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1055,3 +1058,56 @@ def gen_randomized_coords(user_id, lat, lng):
...
+                    data=notification_data_pb2.EventReminder(
+                        event=event_to_pb(session, occurrence, context),
+                        user=user_model_to_pb(user, session, context),
+                        start_time=occurrence.start_time,
+                    ),
+                )
+
+                attendee.reminder_sent = True
+
+    session.flush()
```
aapeliv: Change this to be a `session.commit()` right after the `attendee.reminder_sent = True`. That way if the bg job gets interrupted, we don't send multiple reminders.

app/backend/src/couchers/migrations/versions/14585a4e1868_add_event__reminder_to_.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "14585a4e1868"
+down_revision = "3937eb39a4ca"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute("ALTER TYPE notificationtopicaction ADD VALUE IF NOT EXISTS 'event__reminder'")
```
aapeliv, line 20: Please merge these into one migration

app/backend/src/couchers/models.py
```diff
@@ -2165,6 +2165,8 @@ class EventOccurrenceAttendee(Base):
     user = relationship("User")
     occurrence = relationship("EventOccurrence", backref=backref("attendances", lazy="dynamic"))
 
+    reminder_sent = Column(Boolean, nullable=False, default=False, server_default=expression.false())
```
aapeliv, line 2168: Oooohhh, this is nice!! I didn't know about `expression.false()`

app/backend/templates/v2/event_reminder.mjml
```diff
@@ -0,0 +1,32 @@
...
+        <mj-text font-size="16px" font-family="Helvetica">
+          This is a reminder that the upcoming event is happening soon:
+        </mj-text>
+
+        <mj-text font-size="18px" font-family="Helvetica" font-weight="bold">
+          {{ event.title }}
+        </mj-text>
+
+        <mj-text font-size="16px" font-family="Helvetica">
+          Starts at {{ start_time }} local time
```
aapeliv: This needs to be `{{ start_time|v2timestamp(user) }}`

app/proto/notification_data.proto
```diff
@@ -178,6 +178,12 @@ message ReferenceReminder {
   uint32 days_left = 3;
 }
 
+message EventReminder {
+  org.couchers.api.events.Event event = 1;
+  org.couchers.api.core.User user = 2;
+  google.protobuf.Timestamp start_time = 3;
```
aapeliv: This is included in `event.start_time`, so no need to add it separately.

app/backend/src/couchers/notifications/render.py
```diff
@@ -625,6 +625,26 @@ def render_notification(user, notification) -> RenderedNotification:
                 push_icon=v2avatar(data.author),
                 push_url=event_link,
             )
+        elif notification.action == "reminder":
+            start_time = v2timestamp(data.start_time, user)
```
aapeliv: Use `data.event.start_time`

aapeliv (commented):
The backend logic is almost there! `expression.false()` is also very nifty!

Are you OK if I push a slightly different email template to this branch @andrei-kolesnik-california ?

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: I like this approach. Let me follow this path.

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: Sure, I’ll take it out.

app/backend/src/couchers/jobs/handlers.py
andrei-kolesnik-california: Will commit right after reminder_sent = True to avoid duplicates.

app/backend/src/couchers/migrations/versions/14585a4e1868_add_event__reminder_to_.py
andrei-kolesnik-california: Good point, I’ll merge the two into a single migration.

app/backend/templates/v2/event_reminder.mjml
andrei-kolesnik-california: Will update to {{ start_time|v2timestamp(user) }}.

app/proto/notification_data.proto
andrei-kolesnik-california: You're right, thanks — I’ll remove the extra start_time field.

app/backend/src/couchers/notifications/render.py
andrei-kolesnik-california: I’ll update the render as well.

aapeliv (approved):
LGTM

jesseallhands (approved):
LGTM!

PR #6244, opened 2025-06-27
Fix missing pins

aapeliv:


app/web/components/StaticMap.tsx
```diff
@@ -41,10 +55,12 @@ const StaticMap = () => {
         }}
         mapStyle="https://cdn.couchers.org/maps/couchers-basemap-style-v1.json"
         interactiveLayerIds={clusterLayer.id ? [clusterLayer.id] : []}
-        hash={true}
+        hash={false}
```
nabramow, line 58: What does the hash do actually? I was trying to figure this out.

app/web/features/search/utils/mapLayers.ts
```diff
@@ -22,7 +22,7 @@ const clusterLayer: LayerProps = {
     ],
     "circle-radius": ["step", ["get", "point_count"], 20, 100, 30, 750, 40],
   },
-  source: SOURCE_CLUSTERED_USERS_ID,
+  source: USERS_SOURCE_ID,
```
nabramow, line 25: This is used in the regular map search too, do we really wanna change it?

nabramow (approved):
This looks good! I checked the landing page map and the regular search map and everything is working. 

Thanks for figuring this one out!

app/web/components/StaticMap.tsx
aapeliv: It puts in the URL fragment that tells you where the map is zoomed/panned to:

image

app/web/features/search/utils/mapLayers.ts
aapeliv: Yes, just a name change to make the constant descriptive.

PR #6249, opened 2025-06-28
Add functionality to tell if some refs are hidden

### LINKED ISSUE #3530: Common issue: leaving friend references after stay

aapeliv:
Does two things:

* Closes #3530 by changing the button label to say "Leave friend reference" since this is the only kind of ref you can leave with that button.
* Adds a note saying that some refs are hidden because this user hasn't written one for the recipient yet if some refs are indeed hidden (due to having surfed/hosted and only one ref was written). This has caused some confusion.

- Create a host req, accept, then backdate it.
- Leave a ref from one user to the next.
- The user's profile who didn't yet leave a ref should have the text on their profile.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/servicers/references.py
+++ b/app/backend/src/couchers/servicers/references.py
@@ -8,7 +8,7 @@
 import grpc
 from google.protobuf import empty_pb2
 from sqlalchemy.orm import aliased
-from sqlalchemy.sql import and_, func, literal, or_, union_all
+from sqlalchemy.sql import and_, func, literal, not_, or_, union_all
 
 from couchers import errors
 from couchers.context import make_background_user_context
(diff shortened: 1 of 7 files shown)


nabramow:
I think this is also from this branch?

I'm not sure how I feel about a host request meetup option? Is there a reason we're formalizing that?

Screenshot 2025-07-26 at 10 57 22 PM

nabramow:
If we will keep this, on mobile I think the checkboxes should take up the entire first row, then the dates below it:

Screenshot 2025-07-26 at 10 59 47 PM

nabramow:
Can you make this alert full width? I think it would look better!

Screenshot 2025-07-26 at 11 05 29 PM

Besides that, it seems a bit vague. As a user, I don't think I'd really know what to make of it.

Can we change the text to be more specific? Something like "You have pending references. Users have 14 days to leave a reference after hosting/surfing". Or something like that? I bet @jesseallhands would have a better idea for text.

nabramow:
Okay tested this properly and all is looking good! Just had that comment above about adjusting the wording.

nabramow:
Closing as this was handled in #6814 .

PR #6258, opened 2025-06-29
Make community guidelines translatable

nabramow:
These were coming straight from the backend in English. I used the titles as keys to make them translatable.

**Web frontend checklist**
- [ ] All tests pass

### DIFF
--- a/app/web/features/auth/CommunityGuidelines.tsx
+++ b/app/web/features/auth/CommunityGuidelines.tsx
@@ -68,8 +68,6 @@ export default function CommunityGuidelines({
     queryFn: () => service.resources.getCommunityGuidelines(),
   });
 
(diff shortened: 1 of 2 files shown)


aapeliv (changes requested):
This is too much of a hack. The community guidelines are defined in the backend in

https://github.com/Couchers-org/couchers/blob/develop/app/backend/resources/community_guidelines.json

We have two choices:

1. Not translate these for now and wait until the backend has translations.
2. Straight up move them from the backend to the frontend.

We keep track of the version of community guidelines agreed to by users in the backend, and that is why it's fetched from the backend. Option 2 would break that, but that's OK probably, just need to be careful as to which version people agreed to.

nabramow:
Closing as we need to handle this another way.

PR #6279, opened 2025-06-30
Address misc design fixes

nabramow:
* Remove extra hero section top padding on mobile
* Change signup form text to "Join the Couchers.org community" etc.
* Fix Next.js SSR issues for Fade 
* Center map animation on mobile
* Move "join us" button on mobile to below map
* Add "report a problem" link to footer rather than header so it's findable but not in your face
* Fix "last signup" text alignment on mobile


aapeliv:
I'm now seeing this with javascript disabled (which is what crawlers will see)
image

aapeliv (approved):

PR #6285, opened 2025-07-01
More landing page design fixes

nabramow:
- Remove Couchers map attribution on mobile to save space
- Make map smaller on mobile
- Add commas to userCount number, e.g. "56,232 members" instead of "56232 members".


aapeliv (approved):
LGTM!

PR #6286, opened 2025-07-01
Add shouldNotify checkbox to EditEventPage

### LINKED ISSUE #5719: Edit Event - Add Checkbox 'notify attendees'
**Note from Nicole:** The `shouldNotify` needs to be added to the backend first in #4414 

We want to add a checkbox to the Edit Event page so that when organizers edit the event, they can choose whether or not to notify the event attendees. This is especially useful, for example, if the time or location changes so people can be notified.

Right now everyone gets notified, so we want to give the option to turn off, for example if the host is just fixing a typo or something.

Acceptance Criteria
- Add a checkbox (default unchecked) that says "notify event attendees"
- This should send it to the backend in the `updateEvent` api via the `req.shouldNotify` key

banflam:
Closes #5719 

Adds a checkbox to the Edit Event form that allows organizers to choose whether or not to notify attendees of changes made.

**Web frontend checklist**
- [x ] Formatted my code with `yarn format`
- [x ] There are no console warnings when running the app


nabramow:
Two things here:

1. We only want this to show when editing an event, as creating an event should always notify. The `EventForm` is used by both the `CreateEventPage` and `EditEventPage`. 

2. When I leave `shouldNotify` checkbox blank and update the event, it still sends the notification. You can see this if you turn on your push notifications in Account Settings (then Notification Settings, find the edit event you are attending one). Then edit the event you created and save, see if you get the notification.


I'd suggest adding a prop to EventFrom like `isEdit` and set it to true for `EditEventPage`. Use that to conditionally render the checkbox.

In `EditEventPage` to fix the notification, check the mutation for updating the event, find the `shouldNotify` value. You can see it's hardcoded to `true`. You need to set it to the `data.dirtyFields.shouldNotify` value that's being passed instead.

banflam:
Thank you for the feedback. I won't be able to work on this issue for the foreseeable future, so I'm unassigning myself. Please re-assign it to somebody else.

nabramow (approved):
Necessary changes made and tested successfully so approving!

PR #6351, opened 2025-07-06
Add hosted web client

aapeliv:


nabramow (dismissed):

nabramow:
@aapeliv what's this and do we still need it?

PR #6357, opened 2025-07-06
Backend/feature/6173 how you heard abt us

Krishnag09:

-
-
-

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/models/rest.py
+++ b/app/backend/src/couchers/models/rest.py
@@ -187,6 +187,10 @@ class SignupFlow(Base):
 
     opt_out_of_newsletter: Mapped[bool | None] = mapped_column(Boolean, nullable=True)
 
+    heard_about_us_source = mapped_column(String, nullable=True)
+    what_you_want_to_get_out_of_couchers = mapped_column(String, nullable=True)
+    have_you_couchsurfed_before = mapped_column(Boolean, nullable=True)
+
     ## Feedback (now unused)
     filled_feedback: Mapped[bool] = mapped_column(Boolean, default=False)
     ideas: Mapped[str | None] = mapped_column(String, nullable=True)
--- a/app/backend/src/couchers/models/users.py
+++ b/app/backend/src/couchers/models/users.py
@@ -245,6 +245,10 @@ class User(Base):
 
     recommendation_score: Mapped[float] = mapped_column(Float, server_default="0")
 
+    heard_about_us_source = mapped_column(String, nullable=True)
+    what_you_want_to_get_out_of_couchers = mapped_column(String, nullable=True)
+    have_you_couchsurfed_before = mapped_column(Boolean, nullable=True)
+
     # Columns for verifying their phone number. State chart:
     #                                       ,-------------------,
     #                                       |    Start          |
--- a/app/backend/src/couchers/servicers/auth.py
+++ b/app/backend/src/couchers/servicers/auth.py
@@ -279,6 +279,7 @@ class Auth(auth_pb2_grpc.AuthServicer):
                 flow.geom_radius = request.account.radius
                 flow.accepted_tos = TOS_VERSION
                 flow.opt_out_of_newsletter = request.account.opt_out_of_newsletter
+                flow.have_you_couchsurfed_before = request.account.have_you_couchsurfed_before
                 session.flush()
 
             if request.HasField("feedback"):
@@ -326,6 +327,9 @@ class Auth(auth_pb2_grpc.AuthServicer):
                 last_onboarding_email_sent=func.now(),
                 opt_out_of_newsletter=flow.opt_out_of_newsletter,
                 invite_code_id=flow.invite_code_id,
+                heard_about_us_source=flow.heard_about_us_source,
+                what_you_want_to_get_out_of_couchers=flow.what_you_want_to_get_out_of_couchers,
+                have_you_couchsurfed_before=flow.have_you_couchsurfed_before,
             )
 
             session.add(user)
--- a/app/backend/src/tests/test_auth.py
+++ b/app/backend/src/tests/test_auth.py
@@ -165,6 +165,9 @@ def test_signup_incremental(db):
                     lng=-73.9778,
                     radius=500,
                     accept_tos=True,
+                    heard_about_us_source="friend",
+                    what_you_want_to_get_out_of_couchers="meet new people",
+                    have_you_couchsurfed_before=True,
                 ),
             )
         )
@@ -222,6 +225,9 @@ def _quick_signup() -> int:
                     lng=-73.9778,
                     radius=500,
                     accept_tos=True,
+                    heard_about_us_source="friend",
+                    what_you_want_to_get_out_of_couchers="meet new people",
+                    have_you_couchsurfed_before=True,
                 ),
                 feedback=auth_pb2.ContributorForm(),
                 accept_community_guidelines=wrappers_pb2.BoolValue(value=True),
@@ -344,6 +350,9 @@ def test_login_part_signed_up_not_verified_email(db):
                     lng=-73.9778,
                     radius=500,
                     accept_tos=True,
+                    heard_about_us_source="friend",
+                    what_you_want_to_get_out_of_couchers="meet new people",
+                    have_you_couchsurfed_before=True,
                 ),
             )
         )
@@ -565,6 +574,9 @@ def test_signup_without_password(db):
                         lng=1,
                         radius=100,
                         accept_tos=True,
+                        heard_about_us_source="friend",
+                        what_you_want_to_get_out_of_couchers="meet new people",
+                        have_you_couchsurfed_before=True,
                     ),
                     feedback=auth_pb2.ContributorForm(),
                 )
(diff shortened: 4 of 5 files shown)


app/docker-compose.yml
```diff
@@ -16,7 +16,6 @@ services:
     restart: unless-stopped
     ports:
       - 6545:6545
-    platform: linux/amd64
```
nabramow, line 19: Can you put this back? I don't think we should change the docker compose file in this PR.

app/backend/src/couchers/models.py
```diff
@@ -167,6 +179,13 @@ class User(Base):
     joined = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
     last_active = Column(DateTime(timezone=True), nullable=False, server_default=func.now())
 
+    heard_about_us_source = Column(
+        Enum(HeardAboutUs),
+        server_default=text("'source_friend'"),  # Use the string representation of the Enum value
+        nullable=False,
+    )
```
nabramow: If this is `nullable=False`, how will we handle the existing rows in the DB?

nabramow:
@Krishnag09 we probably need an alembic migration as well, no? Since we are adding columns to the table. We can just set them to null but we need to make sure they exist.

app/proto/account.proto
```diff
@@ -269,6 +272,27 @@ message LogOutOtherSessionsReq {
...
+    SOURCE_ADVERTISEMENT = 8;
+  }
+
+  Source source = 1;
+
+  // If the source is "other", this field can be used to specify it
+  string other_source = 2;
+}
+
+
```
nabramow: I think it makes more sense to define one message and use it once in GetAccountInfoRes. Replace the two fields the PR added with one:

```grpc
message HeardAboutUs {
  // The source of the information about Couchers.org
  enum Source {
    SOURCE_FRIEND = 0;
    SOURCE_SOCIAL_MEDIA = 1;
    SOURCE_SEARCH_ENGINE = 2;
    SOURCE_OTHER = 3;
    SOURCE_EVENT = 4;
    SOURCE_COUCHSURFING = 5;
    SOURCE_OTHER_HOSTING_PLATFORM = 6;
    SOURCE_NEWS_ARTICLE = 7;
    SOURCE_ADVERTISEMENT = 8;
  }

  // Required
  Source source = 1;

  // Optional: when source == SOURCE_OTHER (may be empty otherwise)
  string other_source = 2;
}
```

Then in `getAccountInfo`:

```grpc
message GetAccountInfoRes {
  // ... existing fields ...

  // Replace the two fields the PR added with a single cohesive one:
  HeardAboutUs heard_about_us = 19;
}
```

app/backend/src/couchers/models.py
```diff
@@ -167,6 +179,13 @@ class User(Base):
...
+    heard_about_us_source_other = Column(String, nullable=True, server_default=text("''"))
```
nabramow:
```suggestion
    heard_about_us_source_other = Column(String, nullable=True)
```

app/backend/src/couchers/models.py
nabramow: Can't it just be:

```suggestion
    heard_about_us_source = Column(Enum(HeardAboutUs), nullable=True)
```

That seems more in line with our existing pattern. Also for existing users we have no way of knowing how they heard about us. Which means it needs to be nullable as there's no way we can backfill it.

I also don't think we should require this to sign up. It's more a "nice to have".

nabramow:
Do we need api mapping for the enum? It seems like that's our existing pattern, for example with the enum `profilepublicitysetting2sql` in `account.py`. Something like this:

```python
from couchers.models import HeardAboutUs as HeardAboutUsModel

heardaboutus2api = {
    HeardAboutUsModel.source_friend: account_pb2.HeardAboutUs.SOURCE_FRIEND,
    HeardAboutUsModel.source_social_media: account_pb2.HeardAboutUs.SOURCE_SOCIAL_MEDIA,
    HeardAboutUsModel.source_search_engine: account_pb2.HeardAboutUs.SOURCE_SEARCH_ENGINE,
    HeardAboutUsModel.source_other: account_pb2.HeardAboutUs.SOURCE_OTHER,
    HeardAboutUsModel.source_event: account_pb2.HeardAboutUs.SOURCE_EVENT,
    HeardAboutUsModel.source_couchsurfing: account_pb2.HeardAboutUs.SOURCE_COUCHSURFING,
    HeardAboutUsModel.source_other_hosting_platform: account_pb2.HeardAboutUs.SOURCE_OTHER_HOSTING_PLATFORM,
    HeardAboutUsModel.source_news_article: account_pb2.HeardAboutUs.SOURCE_NEWS_ARTICLE,
    HeardAboutUsModel.source_advertisement: account_pb2.HeardAboutUs.SOURCE_ADVERTISEMENT,
}
```

I dunno @aapeliv correct me if I'm wrong.

nabramow:
Also seems we need to add this to `getAccountInfoRes`?

app/backend/src/couchers/models.py
Krishnag09: ok changes to nullable = True, i was getting a pre hook error for not having a server default value

Krishnag09: marked nullable = true

app/proto/account.proto
Krishnag09: the `other source` field is to take a text input when the user chooses the `SOURCE_OTHER` option..

app/backend/src/couchers/models.py
Krishnag09: i was getting a built error without server default value..

app/backend/src/couchers/models.py
```diff
@@ -94,6 +94,18 @@ class ParkingDetails(enum.Enum):
...
+class HeardAboutUs(enum.Enum):
+    source_friend = 1
+    source_social_media = 2
+    source_search_engine = 3
+    source_other = 4
+    source_event = 5
+    source_couchsurfing = 6
+    source_other_hosting_platform = 7
+    source_news_article = 8
+    source_advertisement = 9
```
aapeliv: Let's keep these as a freeform input **on the backend side**. (The frontend will do multiple choice.) This will make it easier to modify down the line.

This will also need migrations, I can help with that when it's otherwise ready

aapeliv (changes requested):

nabramow:
Hey @Krishnag09 sorry to be annoying. Would it be possible to add two more questions?

* What do you want to do on Couchers? Options: Host/surf/events
* Have you couch surfed before? Options: Yes/No

app/backend/src/couchers/models.py
Krishnag09: the idea was to use enums was to make it better for analysis. but will change to single field with text

Krishnag09:
> Hey @Krishnag09 sorry to be annoying. Would it be possible to add two more questions?
> 
> * What do you want to do on Couchers? Options: Host/surf/events
> * Have you couch surfed before? Options: Yes/No

@nabramow -added those two fields, let me know if they look in terms of def
@aapeliv reduced the HeardAboutUs field to string

once the field type etc is approved, i can proceed to tests

app/backend/src/couchers/models.py
```diff
@@ -174,6 +179,10 @@ class User(Base):
     last_notified_message_id = Column(BigInteger, nullable=False, default=0)
     # same as above for host requests
     last_notified_request_message_id = Column(BigInteger, nullable=False, server_default=text("0"))
+    heard_about_us_source = Column(String, nullable=True)
+    what_you_want_to_get_out_of_couchers = Column(GetOutofCouchers, nullable=True)
```
aapeliv: Again, make it a string please

aapeliv (changes requested):

app/backend/src/couchers/models.py
Krishnag09: did you want both as enums? changes `what_you_want_to_get_out_of_couchers ` to string

nabramow:
Hey @Krishnag09 looks like your protos are failing the CI when you get a sec. Are you just waiting for a review otherwise or you still have more changes to make?

app/backend/src/couchers/models.py
aapeliv: Yes please use strings for everything, no enums!

nabramow:
Hey @Krishnag09 you assigned me to review but looks like you're still failing some backend tests.

aapeliv:
@Krishnag09 you would have to pass these detail in during signup and then stored in the signup flow + user model (once the signup completes). They don't need to be available in the account options.

Krishnag09:
> @Krishnag09 you would have to pass these detail in during signup and then stored in the signup flow + user model (once the signup completes). They don't need to be available in the account options.


@aapeliv - moved it to signup flow. should i remove the fields from both account.proto and account.py?

aapeliv:
@Krishnag09 sorry about the delay, I updated now to what was on this branch (removed the copied files). It seems maybe it's missing some stuff from earlier?

aapeliv:
Taken over in https://github.com/Couchers-org/couchers/pull/7550

PR #6379, opened 2025-07-09
Add hosting location fields to HostRequest and update service, tests,…

### LINKED ISSUE #4398: In hosting/surfing references, place where the hosting/surfing happened should be displayed with the reference
Backend summary:

* Need to add new columns to `host_requests`: `hosting_geom`,  `hosting_geom_radius`, and `hosting_city` (for consistency with user model)
* Need to modify the `CreateHostRequest` call to set the above two fields based on the host user's info
* Need to add migration
* Need to add `hosting_city` to `HostRequest`

---

In hosting/surfing/event references, the person who left the reference's profile photo, name and age are shown along with their **current** location rather than the place where the person hosted or surfed. 

We should be storing the place where the person was hosted/surfed/event took place. This adds important context if the person fails to mention it in their reference - especially if the person is referring to the person's living situation in their reference.

See example of what is (incorrectly) happening right now:
image

In the example above, instead of "Berlin, Germany" it should say, "Host in Madison, USA". If it were a surfing reference, it should say "Guest in Madison, USA".

I suspect this will require a backend update (store the location of the request at the time of request, eg. HostRequestLocation) and then a frontend update so that we are displaying the location that the hosting/surfing happened.

andrei-kolesnik-california:
Description
✨ Added hosting location fields (hosting_city, hosting_lat, hosting_lng, hosting_radius) to the HostRequest model and protobufs.
✨ Updated backend APIs (CreateHostRequest, GetHostRequest, ListHostRequests) to handle and return these fields for richer location details in host requests.
🛠️ Added corresponding database migration to add these new columns to host_requests table.
📋 Added test cases for new hosting location fields.

Testing
🔹 Verify new fields in Create/Get/ListHostRequest API responses
🔹 Run backend tests

Closes #4398


app/backend/src/couchers/models.py
```diff
@@ -1461,6 +1461,11 @@ class HostRequest(Base):
     surfer_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
     host_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
 
+    hosting_city = Column(String, nullable=True)
+    hosting_lat = Column(Float, nullable=True)
+    hosting_lng = Column(Float, nullable=True)
```
aapeliv: Small note: this should be a postgis GEOM object (that contains both lat/lon in one).

andrei-kolesnik-california: Totally agree. It's a good idea to use the same approach as in the user model.

app/backend/src/couchers/models.py
```diff
@@ -1461,6 +1461,10 @@ class HostRequest(Base):
     surfer_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
     host_user_id = Column(ForeignKey("users.id"), nullable=False, index=True)
 
+    hosting_city = Column(String, nullable=True)
+    hosting_location = Column(Geometry("POINT", srid=4326), nullable=True)
+    hosting_radius = Column(Float, nullable=True)
```
aapeliv: We should make these non-nullable and back-populate based on the user's data. For now we can just copy the current host's values into these fields.

Once we merge this I'll write a script that gets the actual right info from somewhere.

app/backend/src/couchers/servicers/requests.py
```diff
@@ -89,6 +91,8 @@ def host_request_to_pb(host_request: HostRequest, session, context):
         .limit(1)
     ).scalar_one()
 
+    point = to_shape(host_request.hosting_location) if host_request.hosting_location else None
```
aapeliv: Please use `lat, lng = get_coordinates(host_request.hosting_location)` (function from app/backend/src/couchers/utils.py)

app/backend/src/couchers/servicers/requests.py
```diff
@@ -103,6 +107,10 @@ def host_request_to_pb(host_request: HostRequest, session, context):
             else host_request.host_last_seen_message_id
         ),
         latest_message=message_to_pb(latest_message),
+        hosting_city=host_request.hosting_city or "",
+        hosting_lat=point.y if point else 0.0,
+        hosting_lng=point.x if point else 0.0,
+        hosting_radius=host_request.hosting_radius or 0.0,
```
aapeliv: If you make them nullable then no need for `or` stuff

app/backend/src/couchers/servicers/requests.py
```diff
@@ -195,6 +203,9 @@ def CreateHostRequest(self, request, context, session):
             surfer_last_seen_message_id=message.id,
             # TODO: tz
             # timezone=host.timezone,
+            hosting_city=host.city,
+            hosting_location=from_shape(Point(host.coordinates[1], host.coordinates[0]), srid=4326),
```
aapeliv: And use `create_coordinate` here, also from `utils.py`

aapeliv (changes requested):

app/backend/src/couchers/models.py
andrei-kolesnik-california: Great! It will be done.

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Have not noticed those functions. Thank you!

andrei-kolesnik-california: Have not noticed those functions. Thank you!

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Leave it as it is, because nullable = false now.

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Yes, thank you!

app/backend/src/couchers/servicers/requests.py
aapeliv: These will not be nullable and I'll run a query to update the locations, so please get rid of the `or` here!

app/backend/src/couchers/servicers/requests.py
```diff
@@ -195,6 +204,9 @@ def CreateHostRequest(self, request, context, session):
             surfer_last_seen_message_id=message.id,
             # TODO: tz
             # timezone=host.timezone,
+            hosting_city=host.city,
+            hosting_location=create_coordinate(host.coordinates[0], host.coordinates[1]),
```
aapeliv: I think you can just do `host.geom` here?

app/backend/src/couchers/servicers/requests.py
```diff
@@ -308,6 +320,14 @@ def ListHostRequests(self, request, context, session):
...
                 ),
                 latest_message=message_to_pb(result.Message),
+                hosting_city=result.HostRequest.hosting_city or "",
+                hosting_lat=get_coordinates(result.HostRequest.hosting_location)[0]
+                if result.HostRequest.hosting_location
+                else 0.0,
+                hosting_lng=get_coordinates(result.HostRequest.hosting_location)[1]
+                if result.HostRequest.hosting_location
+                else 0.0,
+                hosting_radius=result.HostRequest.hosting_radius or 0.0,
```
aapeliv: Again, no defaults please!

app/backend/src/couchers/servicers/requests.py
```diff
@@ -89,6 +91,9 @@ def host_request_to_pb(host_request: HostRequest, session, context):
         .limit(1)
     ).scalar_one()
 
+    coords = get_coordinates(host_request.hosting_location)
+    lat, lng = coords if coords else (None, None)
```
aapeliv: Here as well, no need for else, just use `lat, lng = get_coordinates(...)`

app/backend/src/tests/test_requests.py
```diff
@@ -30,7 +30,16 @@ def _(testconfig):
...
 def test_create_request(db):
     user1, token1 = generate_user()
-    user2, token2 = generate_user()
+    hosting_city = "Morningside Heights, New York City"
+    hosting_lat = 40.8086
+    hosting_lng = -73.9616
+    hosting_radius = 500
+    user2, token2 = generate_user(
+        city=hosting_city,
+        geom=f"POINT({hosting_lng} {hosting_lat})",
```
aapeliv: you can also use `create_coordinate(hosting_lat, hosting_lon)` here if you wish; but it's in a test so up to you!

app/backend/src/couchers/migrations/versions/8c878b177151_add_hosting_location_to_host_request.py
```diff
@@ -0,0 +1,43 @@
+"""add hosting location to host request
+
+Revision ID: 8c878b177151
+Revises: 14585a4e1868
```
aapeliv, line 4: Please update to `Revises: d345174cb002`

app/backend/src/couchers/migrations/versions/8c878b177151_add_hosting_location_to_host_request.py
```diff
@@ -0,0 +1,43 @@
...
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from geoalchemy2 import Geometry
+
+# revision identifiers, used by Alembic.
+revision = "8c878b177151"
+down_revision = "14585a4e1868"
```
aapeliv: Please update to `down_revision = "d345174cb002"`

aapeliv (changes requested):
Almost there! Just minor comments, and then it's good to go!

app/backend/src/couchers/migrations/versions/8c878b177151_add_hosting_location_to_host_request.py
aapeliv: Actually don't worry about these, I can update it when it's ready to go

app/backend/src/couchers/migrations/versions/8c878b177151_add_hosting_location_to_host_request.py
aapeliv: Actually don't worry about these, I can update it when it's ready to go (given `develop` is getting lots of migrations right now)

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Absolutely agree. Thank you.

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Will remove defaults.

app/backend/src/couchers/servicers/requests.py
andrei-kolesnik-california: Yes

app/backend/src/tests/test_requests.py
andrei-kolesnik-california: The code will look cleaner indeed, thanks!

aapeliv (approved):

PR #6419, opened 2025-07-14
Adds script for adding translation languages and more

### LINKED ISSUE #6386: Add Swedish as a language and hook into Weblate
Add Swedish as a language and get it working in Weblate.

Make all the json files too.

### LINKED ISSUE #6392: LanguagePickerSelect - default to browser language
See if it's possible to have the LanguagePickerSelect to default to the browser language. Right now it defaults to English for everyone.

It might be the `localeDetection` setting we need to change to true for i18n, but the select has to match this.

These docs might help:

https://nextjs.org/docs/pages/guides/internationalization

nabramow:
Closes #6386 
Closes #6392 

* Adds script to add or remove translation languages to create all the json files in features/locales, add to LANGUAGE_MAP for flags, and to allLanguages.ts
* Adds script to remove a language (in case we mess up mostly)
* Adds Swedish (sv) and Ukranian (uk)
* Sets localeDetection to true, this will default to the browser language instead of English. Once the user selects a language, it will set a cookie and keep using that.
* Removes two unused English translation keys


app/web/next-i18next.config.js
```diff
@@ -13,7 +13,7 @@ const fallbackLng = {
 module.exports = {
   i18n: {
     defaultLocale: "en",
-    localeDetection: false,
+    localeDetection: true,
```
aapeliv, line 16: What does this do?

nabramow: It's in the description ;-)

nabramow: This was a request from Chris

aapeliv (approved):
LGTM

PR #6492, opened 2025-07-18
Map style improvements and fixes

### LINKED ISSUE #6361: Mobile map user card - haiku cuts off bottom text
For the empty profile haiku cards on mobile, the bottom line of text is getting cut off:

Image

Acceptance Critera:

- For mobile, decrease the padding/margin above and below the haiku so text doesn't get cut off
- Make sure everything still looks good on other screen sizes and desktop

### LINKED ISSUE #6362: Mobile map pagination - iphone cuts previous/next
On iOS mobile (on the phone only not google chrome dev view), the previous.next bar is cut off on the bottom right and left because the iPhone screen is rounded.

Acceptance Criteria
- Make the bottom bar height a bit bigger for mobile so the text isn't obviously cut off.
- You'll need to test it on a real phone, follow the instructions in: docs/run-local-app-on-mobile.md

### LINKED ISSUE #6519: Filter empty profiles during zoom switch
Replicate:

Open map -> manually zoom to any location -> map switches modes (i.e. shows user tiles) but doesn't filter out empty profiles

Solution:

Right now the filter empty profiles works when searching from the dashboard and when searching on the search input.

We need to target it to also by default filter out empty profiles when the zoom switches passed the query users level.

nabramow:
* Make previous/next buttons more obvious and visually differentiated
* Make "search this area" button more visible
* Make list width expand icon change color when you hover over it
* Make zoom controls change color when you hover over it
* Fix issue with + in Zoom Control not working
* Fix map empty haiku container bottom card text getting cut off
* Default to empty filters can/maybe host should also work on the zoom in threshold user search
* Changed the level to switch to user search to be higher (city level)
* On zoom out past that threshold again, resets the filters and search value

Closes #6361 
Closes #6519 

Hopefully also Closes #6362 since the bottom bar is now bigger and buttons slightly more inline


jesseallhands (dismissed):
LGTM, only change I might suggest is having the next and previous buttons match the same style as all other buttons on the website (eg the "show all profiles" button).

jesseallhands (approved):
Lgtm!

PR #6526, opened 2025-07-22
Invite code integration

### LINKED ISSUE #6449: Invitations backend
Users should be able to create unique invite links to share with friends/family/online, and we should keep track of where users came from in that case.

The idea is that when a user clicks on an invite code, we generate a link of this type:

https://couchers.org/invite?code=[codehere]

This should then take the visitor to a customized signup page that says "Aapeli invited you to join Couchers!", along with Aapeli's avatar and username (this is the frontend part).

This requires the following from the backend:
1. A new `InviteCode` table, containing `id` (a unique 6-character alphanumeric code), datetimes `created`/`disabled` (initially null), `creator_user_id`
2. Add a new field in `users` called `invite_code_id` that is a nullable foreign key to the `InviteCode` telling where the user came from
3. An API in `account` called `CreateInviteCode` that generates a new unique 8-char code
4. An API in `account` called `DisableInviteCode` that disables an invite code
5. An API in `account` called `ListInviteCodes` that lists the links as well as counts on how many people joined with that link!
6. An API in `auth` called `GetInviteCodeInfo` that returns the user's name, username, and avatar URL.
7. The `auth` signup flow should be amended to send back the invite code with the signup request and store it in `invite_code_id`: this probably requires changes to the `signup_flow` object as well.

andrei-kolesnik-california:
This PR adds support for invite codes in the signup flow. Users can now be invited with a unique code, which is stored during the signup process and linked to both the user and the signup flow.
This helps track invitations and enables future features like referral tracking.

Closes #6449

Steps to check PR:
- Create an invite code using CreateInviteCode in the account API.
- Use the invite code during the signup basic step via SignupFlow.
- Complete signup normally.
- Check in the DB that the new user has invite_code_id set.
- Check that invalid invite codes are rejected.
- Run all tests (backend).


app/backend/src/couchers/models.py
```diff
@@ -124,6 +124,21 @@ class TimezoneArea(Base):
     )
 
 
+class InviteCode(Base):
+    __tablename__ = "invite_codes"
+
+    id = Column(String(8), primary_key=True)  # 8-char unique code
```
aapeliv: Let's use `String` here: no need to use a strict length varchar here.

app/backend/src/couchers/models.py
```diff
@@ -124,6 +124,21 @@ class TimezoneArea(Base):
...
+    creator_user_id = Column(Integer, ForeignKey("users.id", ondelete="CASCADE"), nullable=False)
```
aapeliv: No need to add an `ondelete`

app/backend/src/couchers/models.py
```diff
@@ -124,6 +124,21 @@ class TimezoneArea(Base):
...
+    created = Column(DateTime, nullable=False, default=func.now())
+    disabled = Column(DateTime, nullable=True)
+
+    creator = relationship(
+        "User",
+        foreign_keys=[creator_user_id],
```
aapeliv: I don't think you need to define the foreign keys given that there's only one option.

app/backend/src/couchers/models.py
```diff
@@ -124,6 +124,21 @@ class TimezoneArea(Base):
     )
 
 
+class InviteCode(Base):
```
aapeliv: Please move this later in the file, let's keep `User` at the top (I know this file is a hot mess...)

app/backend/src/couchers/models.py
```diff
@@ -326,6 +341,20 @@ class User(Base):
 
     age = column_property(func.date_part("year", func.age(birthdate)))
 
+    # ID of the invite code used to sign up (if any)
+    invite_code_id = Column(String(8), ForeignKey("invite_codes.id"), nullable=True)
+    invite_code = relationship(
+        "InviteCode",
+        foreign_keys=[invite_code_id],
```
aapeliv: Again, no need for foreign keys explicitly.

app/backend/src/couchers/models.py
```diff
@@ -326,6 +341,20 @@ class User(Base):
...
+    )
+
+    # List of invite codes created by user
+    created_invite_codes = relationship(
+        "InviteCode",
+        foreign_keys="[InviteCode.creator_user_id]",
+        back_populates="creator",
+    )
```
aapeliv: SQLAlchemy automatically creates this other side, so no need to do it explicitly.

app/backend/src/couchers/models.py
```diff
@@ -326,6 +341,20 @@ class User(Base):
 
     age = column_property(func.date_part("year", func.age(birthdate)))
 
+    # ID of the invite code used to sign up (if any)
+    invite_code_id = Column(String(8), ForeignKey("invite_codes.id"), nullable=True)
```
aapeliv: No need to specify the type again

app/backend/src/couchers/models.py
```diff
@@ -1013,6 +1042,8 @@ class SignupFlow(Base):
     contribute_ways = Column(ARRAY(String), nullable=True)
     expertise = Column(String, nullable=True)
 
+    invite_code_id = Column(ForeignKey("invite_codes.id"), nullable=True)
```
aapeliv, line 1056: This is a dupe from above

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +641,52 @@ def SetProfilePublicVisibility(self, request, context, session):
         user.has_modified_public_visibility = True
         return empty_pb2.Empty()
 
+    def CreateInviteCode(self, request, context, session):
+        user = session.execute(select(User).where(User.id == context.user_id)).scalar_one()
+
+        code = urlsafe_secure_token()[:8]
```
aapeliv: Please make a new function in crypto called `generate_invite_code` maybe?

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +641,52 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+        invite = InviteCode(id=code, creator_user_id=user.id)
```
aapeliv: Since you're only using the `id` for `user`, you can just put `context.user_id` here, no need to pull the user object and make that query

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +641,52 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+        session.add(invite)
+        session.commit()
```
aapeliv: You can flush here instead of commit

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +641,52 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+
+        return account_pb2.CreateInviteCodeRes(code=code)
+
+    def DisableInviteCode(self, request, context, session):
+        user = session.execute(select(User).where(User.id == context.user_id)).scalar_one()
```
aapeliv: Again, no need to fetch the user

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +641,52 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+    def ListInviteCodes(self, request, context, session):
+        user = session.execute(select(User).where(User.id == context.user_id)).scalar_one()
+
+        invites = session.execute(select(InviteCode).where(InviteCode.creator_user_id == user.id)).scalars().all()
+
+        codes = []
+        for invite in invites:
+            count = session.execute(
+                select(func.count()).select_from(User).where(User.invite_code_id == invite.id)
+            ).scalar_one()
```
aapeliv: Better to do this query in the database, join on the users above and group by invite code, then `func.count(User.id)` or similar. Maybe order them by invited users desc, disabled

app/backend/src/couchers/servicers/auth.py
```diff
@@ -633,3 +644,22 @@ def AntiBotPolicy(self, request, context, session):
...
 
         return auth_pb2.AntiBotPolicyRes(should_antibot=False)
+
+    def GetInviteCodeInfo(self, request, context, session):
+        invite = session.execute(select(InviteCode).where(InviteCode.id == request.code)).scalar_one_or_none()
+
+        if not invite:
+            context.abort(grpc.StatusCode.NOT_FOUND, errors.INVITE_CODE_NOT_FOUND)
+
+        user = session.execute(select(User).where(User.id == invite.creator_user_id)).scalar_one()
```
aapeliv: Again, can combine into one query here

app/backend/src/couchers/servicers/auth.py
```diff
@@ -633,3 +644,22 @@ def AntiBotPolicy(self, request, context, session):
...
+
+        if user.avatar:
+            avatar_url = media_url(filename=user.avatar.key, size="thumbnail")
+        else:
+            avatar_url = ""
+
+        return auth_pb2.GetInviteCodeInfoRes(
+            name=user.name,
+            username=user.username,
+            avatar_url=avatar_url,
```
aapeliv: Use something like this here:

```py
user.avatar.thumbnail_url if user.avatar else None
```

app/proto/account.proto
```diff
@@ -145,6 +158,28 @@ message ChangeLanguagePreferenceReq {
   string ui_language_preference = 1;
 }
 
+message CreateInviteCodeReq {}
+
+
+message CreateInviteCodeRes {
+  string code = 1;
+}
```
aapeliv: Please add a `string url = 2;` here that's simply the public URL + `/invite?code={code}`

app/proto/account.proto
```diff
@@ -145,6 +158,28 @@ message ChangeLanguagePreferenceReq {
...
+
+message DisableInviteCodeReq {
+  string code = 1;
+}
+
+message InviteCodeInfo {
+  string code = 1;
```
aapeliv: Same here with the URL

aapeliv (changes requested):

app/backend/src/couchers/models.py
andrei-kolesnik-california: Got it. Length is enforced in migrations anyway, and this keeps the DB schema more flexible.

app/backend/src/couchers/models.py
andrei-kolesnik-california: Got it. Agree.

app/backend/src/couchers/models.py
andrei-kolesnik-california: There are actually two options: either only keep creator_user_id and perform an extra DB query each time we need the User object, or explicitly define the relationship with foreign_keys, as I did. In the latter case, SQLAlchemy sees multiple potential foreign key paths and requires the correct one to be specified explicitly.

app/backend/src/couchers/models.py
andrei-kolesnik-california: I tried to find a better place within the existing structure. Please check it out.

app/backend/src/couchers/models.py
andrei-kolesnik-california: In this case, SQLAlchemy sees multiple possible foreign key paths. I specify foreign_keys explicitly to make the relationship unambiguous and avoid errors.

app/backend/src/couchers/models.py
andrei-kolesnik-california: Thanks, corrected!

app/backend/src/couchers/models.py
andrei-kolesnik-california: Agreed, thanks!

app/backend/src/couchers/models.py
andrei-kolesnik-california: This is not a duplicate — this invite_code_id is for storing the invite code during the signup flow before the actual User is created.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Thanks, I moved the code to a new generate_invite_code() function in crypto.py, ensuring it generates URL-safe token.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Thanks a lot, I removed the extra database query and now just use context.user_id directly.

andrei-kolesnik-california: Thanks a lot, I removed the extra database query and now just use context.user_id directly.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Done.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Agreed.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: I simplified it to a single query with join and group by as suggested. Thank you very much!

app/backend/src/couchers/servicers/auth.py
andrei-kolesnik-california: Agred.

app/backend/src/couchers/servicers/auth.py
andrei-kolesnik-california: Thanks, good catch.

app/proto/account.proto
andrei-kolesnik-california: Done, added url field and made corresponding adjustments in the server response.

app/proto/account.proto
andrei-kolesnik-california: Done, added url field and made corresponding adjustments in the server response.

app/backend/src/couchers/crypto.py
```diff
@@ -193,3 +194,11 @@ def asym_decrypt(private_key: bytes, encrypted_data: bytes) -> bytes:
 def generate_asym_keypair():
     skey = PrivateKey.generate()
     return skey.encode(), skey.public_key.encode()
+
+
+def generate_invite_code(length=8):
+    """
+    Generates a secure, URL-safe invite code of the given length.
+    """
+    alphabet = string.ascii_letters + string.digits
```
aapeliv: This is very elegant, didn't know `string` had these!

Let's use `string.ascii_lowercase` here though

app/backend/src/couchers/models.py
aapeliv: Ah sorry, I didn't notice it was in a different model

app/backend/src/couchers/models.py
```diff
@@ -1300,6 +1306,17 @@ def __repr__(self):
         return f"GroupChatSubscription(id={self.id}, user={self.user}, joined={self.joined}, left={self.left}, role={self.role}, group_chat={self.group_chat})"
 
 
+class InviteCode(Base):
+    __tablename__ = "invite_codes"
+
+    id = Column(String, primary_key=True)
+    creator_user_id = Column(Integer, ForeignKey("users.id"), nullable=False)
+    created = Column(DateTime, nullable=False, default=func.now())
+    disabled = Column(DateTime, nullable=True)
```
aapeliv: These should be `DateTime(timezone=True)`, otherwise it's not guaranteed to be stored in UTC!

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +642,55 @@ def SetProfilePublicVisibility(self, request, context, session):
...
 
+    def CreateInviteCode(self, request, context, session):
+        code = generate_invite_code()
+        invite = InviteCode(id=code, creator_user_id=context.user_id)
+        session.add(invite)
+        session.flush()
+
+        return account_pb2.CreateInviteCodeRes(
+            code=code,
+            url=f"{config['BASE_URL']}/invite?code={code}",
```
aapeliv: Please add a new function in `urls.py` and use that one:

```
def invite_code_link(*, code):
    return f"{config['BASE_URL']}/invite?code={code}"
```

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +642,55 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+            select(
+                InviteCode.id,
+                InviteCode.disabled,
+                func.count(User.id).label("num_users"),
+            )
+            .outerjoin(User, User.invite_code_id == InviteCode.id)
+            .where(InviteCode.creator_user_id == context.user_id)
+            .group_by(InviteCode.id, InviteCode.disabled)
+            .order_by(func.count(User.id).desc(), InviteCode.disabled)
+        ).all()
```
aapeliv: Great!!

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +642,55 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+
+        return account_pb2.ListInviteCodesRes(
+            invite_codes=[
+                account_pb2.InviteCodeInfo(
+                    code=code_id,
+                    disabled=Timestamp_from_datetime(disabled) if disabled else None,
+                    uses=len_users,
+                    url=f"{config['BASE_URL']}/invite?code={code_id}",
```
aapeliv: Again, use `url.invite_code_link(code=code_id)`

app/backend/src/couchers/servicers/auth.py
```diff
@@ -205,10 +206,19 @@ def SignupFlow(self, request, context, session):
 
                 flow_token = cookiesafe_secure_token()
 
+                invite = None
+                if request.basic.invite_code:
+                    invite = session.execute(
+                        select(InviteCode).where(InviteCode.id == request.basic.invite_code)
```
aapeliv: add `.where(or_(InvitedCode.disabled == None, InvitedCode.disabled>func.now()))`

also I think you can do `select(InviteCode.id)` and then you don't need to do `invite.id if invite else None` below!

app/backend/src/couchers/servicers/auth.py
```diff
@@ -633,3 +643,18 @@ def AntiBotPolicy(self, request, context, session):
                     return auth_pb2.AntiBotPolicyRes(should_antibot=True)
 
         return auth_pb2.AntiBotPolicyRes(should_antibot=False)
+
+    def GetInviteCodeInfo(self, request, context, session):
+        invite = session.execute(select(InviteCode).where(InviteCode.id == request.code)).scalar_one_or_none()
```
aapeliv: `.where(or_(InvitedCode.disabled == None, InvitedCode.disabled>func.now()))`

app/backend/src/couchers/servicers/auth.py
```diff
@@ -633,3 +643,18 @@ def AntiBotPolicy(self, request, context, session):
...
+        if not invite:
+            context.abort(grpc.StatusCode.NOT_FOUND, errors.INVITE_CODE_NOT_FOUND)
+
+        user = session.execute(select(User).where(User.id == invite.creator_user_id)).scalar_one()
+
+        return auth_pb2.GetInviteCodeInfoRes(
+            name=user.name,
+            username=user.username,
+            avatar_url=user.avatar.thumbnail_url if user.avatar else None,
+            url=f"{config['BASE_URL']}/invite?code={request.code}",
```
aapeliv: Use the URL function again please

app/backend/src/couchers/servicers/account.py
```diff
@@ -640,6 +642,55 @@ def SetProfilePublicVisibility(self, request, context, session):
...
+            .where(InviteCode.creator_user_id == context.user_id)
+            .group_by(InviteCode.id, InviteCode.disabled)
+            .order_by(func.count(User.id).desc(), InviteCode.disabled)
+        ).all()
+
+        return account_pb2.ListInviteCodesRes(
+            invite_codes=[
+                account_pb2.InviteCodeInfo(
+                    code=code_id,
+                    disabled=Timestamp_from_datetime(disabled) if disabled else None,
```
aapeliv, line 667: Your protobuf also has a `created`, so you should add it here!

app/backend/src/tests/test_auth.py
```diff
@@ -1137,3 +1140,85 @@ def test_banned_username(db):
...
+                    invite_code="INV12345",
+                )
+            )
+        )
+        flow_token = res.flow_token
+        assert flow_token
+
+    with session_scope() as session:
+        flow = session.execute(select(SignupFlow).where(SignupFlow.flow_token == flow_token)).scalar_one()
+        assert flow.invite_code_id == "INV12345"
```
aapeliv: Please add a full test to make sure that the invite code is attached to the user at the end, you can copy-paste the body of `test_opt_out_of_newsletter` and adapt

aapeliv (changes requested):
Just a few more small things!

app/backend/src/couchers/crypto.py
andrei-kolesnik-california: Will do, thanks for the suggestion.

app/backend/src/couchers/models.py
andrei-kolesnik-california: Will do.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Sure, will move it to urls. Thanks!

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Totally agree — it's a few lines, but I'm sure it's worth it for a single DB query.

app/backend/src/couchers/servicers/auth.py
andrei-kolesnik-california: Good point — will update the query with the .where() condition.

app/backend/src/couchers/servicers/auth.py
andrei-kolesnik-california: Will do.

app/backend/src/couchers/servicers/auth.py
andrei-kolesnik-california: Will use func here and in the tests as well — thanks for pointing it out.

app/backend/src/couchers/servicers/account.py
andrei-kolesnik-california: Will do.

app/backend/src/tests/test_auth.py
andrei-kolesnik-california: That definitely does make sense to do and is a very good point, especially considering the fact that manual registration check is not the fastest process.

aapeliv (approved):
LGTM, feel free to merge when tests pass!

PR #6567, opened 2025-07-28
Specify yarn version

fbalda:
Specify the yarn version used to allow people with newer/incompatible versions to automatically install and switch to the correct one. (This is not optional if you have a newer version installed and I currently have this change locally to be able to work with the project at all)

The version chosen (1.22.22) is the latest stable version of yarn classic and in theory should work automatically for everyone who worked on the project until now if they weren't using an old yarn version. If their version is incompatible they will receive a prompt to install the correct one and can then continue working as before.

**Other**

### DIFF
--- a/app/web/package.json
+++ b/app/web/package.json
@@ -112,5 +112,6 @@
   },
   "resolutions": {
     "@types/react": "^18.3.1"
-  }
+  },
+  "packageManager": "yarn@1.22.22"
 }


nabramow:
Hey @fbalda can you tell me a little more about the issue you had here?

I'm just wondering if this is a band-aid and maybe we should fix the source of the problem instead.

Currently we tell people to install yarn globally like so:

```
## Set up node & yarn
# install the requires node version
nvm install
# install yarn
npm install --global yarn
```
app/web/readme.md

I'm thinking rather than require a version, we could change this to:

`npx yarn install`

Which would locally install the right version instead of polluting things globally.

What do you think?

fbalda:
> Hey @fbalda can you tell me a little more about the issue you had here?
> 
> I'm just wondering if this is a band-aid and maybe we should fix the source of the problem instead.
> 
> Currently we tell people to install yarn globally like so:
> 
> ```
> ## Set up node & yarn
> # install the requires node version
> nvm install
> # install yarn
> npm install --global yarn
> ```
> 
> app/web/readme.md
> 
> I'm thinking rather than require a version, we could change this to:
> 
> `npx yarn install`
> 
> Which would locally install the right version instead of polluting things globally.
> 
> What do you think?

Hey @nabramow,
non-classic yarn versions (>= 2.0.0) use a different mechanism for tracking dependencies by default so new untracked files will be created every time you run a command with a new version in a repository that uses a classic version (or has no version specified). 
I have a non-classic yarn version installed globally and it is used by default. To make yarn switch to the correct version,
you specify the package manager version in the `package.json` of the project and I don't think this is a bandaid fix.
I just checked and here are the first 4 daily and weekly trending Github repositories using yarn I found, all specifying the version (both classic and non-classic):
- https://github.com/makeplane/plane/blob/preview/package.json
- https://github.com/software-mansion/react-native-reanimated/blob/main/package.json
- https://github.com/toss/es-toolkit/blob/main/package.json
- https://github.com/outline/outline/blob/main/package.json

(In fact the version gets added to the `package.json` when you use the official guide I linked below)

As you seem to be using classic versions we should require everyone working on the project to also use classic versions so everything works as expected. Alternatively we could upgrade to a new yarn version but we'd have to add some additional configuration in that case. Long-term this should be done though, as classic yarn is in maintenance mode since 2020 and will eventually reach EOL [according to the website](https://classic.yarnpkg.com/lang/en/docs/install).

I assume your global yarn defaults to a classic version (probably because you don't have a newer version installed). You can check with `yarn -v`. If that is the case you can replicate my issue by installing a non-classic global version as described [here](https://yarnpkg.com/migration/guide). Once you do this for any repository, the new version should be used by default.

Sorry for the wall of text, I hope I'm making sense

nabramow:
> > Hey @fbalda can you tell me a little more about the issue you had here?
> > I'm just wondering if this is a band-aid and maybe we should fix the source of the problem instead.
> > Currently we tell people to install yarn globally like so:
> > ```
> > ## Set up node & yarn
> > # install the requires node version
> > nvm install
> > # install yarn
> > npm install --global yarn
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > app/web/readme.md
> > I'm thinking rather than require a version, we could change this to:
> > `npx yarn install`
> > Which would locally install the right version instead of polluting things globally.
> > What do you think?
> 
> Hey @nabramow, non-classic yarn versions (>= 2.0.0) use a different mechanism for tracking dependencies by default so new untracked files will be created every time you run a command with a new version in a repository that uses a classic version (or has no version specified). I have a non-classic yarn version installed globally and it is used by default. To make yarn switch to the correct version, you specify the package manager version in the `package.json` of the project and I don't think this is a bandaid fix. I just checked and here are the first 4 daily and weekly trending Github repositories using yarn I found, all specifying the version (both classic and non-classic):
> 
> * https://github.com/makeplane/plane/blob/preview/package.json
> * https://github.com/software-mansion/react-native-reanimated/blob/main/package.json
> * https://github.com/toss/es-toolkit/blob/main/package.json
> * https://github.com/outline/outline/blob/main/package.json
> 
> (In fact the version gets added to the `package.json` when you use the official guide I linked below)
> 
> As you seem to be using classic versions we should require everyone working on the project to also use classic versions so everything works as expected. Alternatively we could upgrade to a new yarn version but we'd have to add some additional configuration in that case. Long-term this should be done though, as classic yarn is in maintenance mode since 2020 and will eventually reach EOL [according to the website](https://classic.yarnpkg.com/lang/en/docs/install).
> 
> I assume your global yarn defaults to a classic version (probably because you don't have a newer version installed). You can check with `yarn -v`. If that is the case you can replicate my issue by installing a non-classic global version as described [here](https://yarnpkg.com/migration/guide). Once you do this for any repository, the new version should be used by default.
> 
> Sorry for the wall of text, I hope I'm making sense

Thanks for taking the time to answer this @fbalda ! This makes sense. I'll go ahead and approve your PR.

nabramow (approved):

PR #6569, opened 2025-07-29
Require reference rating

### LINKED ISSUE #6435: Make reference slider undefined by default
Sometimes people just leave the reference review slider as it is, at neutral, giving a person a meh rating by accident.

Acceptance Criteria
- Make the `ReferenceSlider` undefined by default
- Make it required to move to go to the next step

fbalda:
Makes the value of the rating slider on the reference page undefined by default and requires the user to change the value to go to the next step.

Also included are styling changes to better communicate the need to change the slider value:
- Made track grey by default
- Made thumb icon a question mark by default
- Made rail transparent with a black outline (The default color clashes with the grey. As a bonus, this works better with changing colors of the track)

~~As the rating is now required, the form will display an error message if it isn't set. I defined the key for the translation (`profile:leave_reference.rating_required`), but the actual texts are still needed~~
Edit: Text added

closes #6435

I tested this with the local backend by temporarily adding the following code to `LeaveReferencePage.tsx` right under the `useListAvailableReferences` call:

```
if (availableReferences) {
    availableReferences.canWriteFriendReference = true;
}
```

Logging in with the Aapeli test account and then going to http://localhost:[port]/leave-reference/friend/2 opens the relevant page.

(There is probably a better way to test this, but this worked for me)

Regarding automated tests: I did not add any, as all of the files I touched up until now either had no tests at all or had no tests that were applicable to the code I changed. If desired, I will add some

**Web frontend checklist**
- [x] All tests pass

**Other**


app/web/components/RatingsSlider/RatingsSlider.tsx
```diff
@@ -92,19 +94,17 @@ const StyledSlider = styled(Slider, {
...
-      value={value}
+      sliderColor={getSliderColor(value)}
+      value={value ?? marks[1].value}
       min={0}
       max={1}
       step={0.01}
       marks={marks}
       valueLabelDisplay="on"
-      valueLabelFormat={(value) => <SliderLabel value={value} />}
+      valueLabelFormat={() => <SliderLabel value={value} />}
```
fbalda, line 107: As this change might be a bit confusing: I added a separate `SliderLabel` icon when the value is undefined, however the value passed to the `valueLabelFormat` callback is always a number (as a slider can't have an undefined value). So I'm using the value from the props. AFAIU this doesn't change anything, as the slider re-renders on value change anyway

app/web/features/profile/view/leaveReference/formSteps/PrivateFeedback.tsx
```diff
@@ -181,64 +181,71 @@ export default function PrivateFeedback({
             render={({ field }) => (
               <RatingsSlider onChange={field.onChange} value={field.value} />
             )}
+            rules={{
+              required: t("profile:leave_reference.rating_required"),
```
jesseallhands, line 192: I'm not sure where the text is stored for `leave_reference.rating_required` (it's in some json file), but here is the text we should use:

`Please move the slider and rate your experience.`

jesseallhands (changes requested):
seems to be working as expected! Looks fantastic too! I included my suggestion for the error message text (but I'm not sure which json file the text should be stored in - Nicole would know!)

app/web/features/profile/view/leaveReference/formSteps/PrivateFeedback.tsx
fbalda: Will add tomorrow (I know the relevant translation files). For new text like this one, do we merge into develop with only the English version or do we wait for more translations?

jesseallhands: You merge into develop with English only and translations are merged separately later on a periodic basis.

jesseallhands (dismissed):
This is fantastic! Love the new colors too!

LGTM, but have @nabramow take a look before merging

app/web/features/profile/view/leaveReference/formSteps/PrivateFeedback.tsx
```diff
@@ -181,64 +181,71 @@ export default function PrivateFeedback({
...
-                      "& > .MuiInputBase-root": { width: "100%", marginTop: 1 },
-                      marginTop: 2,
-                    }}
-                  />
-                )}
-              />
-            </PrivateTextContainer>
-          )}
+          {rating !== undefined &&
+            (wasAppropriate === "false" || rating < 0.33) && (
```
nabramow: Might be good to extract this into a variable for readability?

nabramow (dismissed):
Just that one small nit, but I tested it with a real reference and was able to properly leave a review and trigger the error message. This looks great!

app/web/features/profile/view/leaveReference/formSteps/PrivateFeedback.tsx
```diff
@@ -35,6 +35,8 @@ import {
 } from "routes";
 import { theme } from "theme";
 
+const ACCEPTABLE_RATING_THRESHOLD = 0.33;
```
fbalda, line 38: @nabramow I took a bit of an issue with that magic number so I extracted it into a constant. However I was not sure which case to use, there doesn't seem to be a consistent style for constants like this in the codebase. Lmk if this works for you, maybe ideally we should have a short style guide and/or stricter ESlint rules (e.g. [naming-convention](https://typescript-eslint.io/rules/naming-convention/))? I personally love not having to think about stuff like that while coding 😄

nabramow: Yeah I agree building out the eslint rules more has been on my list for awhile! Just has always been more pressing matters ;-). The all caps is good! Feel free to throw that into a small PR for the naming-convention if you want!

nabramow (approved):
LGTM!

PR #6577, opened 2025-07-30
Improved rating slider color interpolation

fbalda:
Changed the color interpolation on the rating slider from RGB to HSL. This prevents the brownish colors in the middle of the spectrum and instead transitions nicely through the hues.

Branched off from #6569, should be merged after it

To keep using the colors values defined in the theme I had to bring in a function that converts the RGB color strings to the HSL representation. To prevent adding an additional dependency I just adapted a function from the [color-convert](https://www.npmjs.com/package/color-convert) package. Other options would be to add the package as a dependency or to directly define the slider extreme values as HSL (instead of taking them from the theme)

For testing see #6569 

**Web frontend checklist**
- [x] All tests pass

**Other**


app/web/utils/color.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  try {
+    const { values } = decomposeColor(colorString);
+    rgb = values as [number, number, number];
+  } catch {
+    console.warn(`Tried to decompose invalid color string '${colorString}'`);
+  }
+
+  const r = rgb[0] / 255;
+  const g = rgb[1] / 255;
```
nabramow: Would you mind adding some short comment in here explaining what problem this is solving? Since we have a lot of engineers come and go, that helps so a future person is not like 🧐. :-)

app/web/utils/math.ts
```diff
@@ -0,0 +1,2 @@
+export const lerp = (start: number, end: number, alpha: number) =>
+  start + (end - start) * alpha;
```
nabramow, line 3: Is it possible to give this a name that makes it clearer what this is doing?

fbalda: I'd say it is basically convention to use that name (In fact, except for GLSL which calls it 'mix', I've only ever seen it called 'lerp' - [here](https://threejs.org/docs/#api/en/math/Vector3]) [are](https://animejs.com/documentation/utilities/lerp/) [a](https://p5js.org/reference/p5/lerp/) [few](https://developer.download.nvidia.com/cg/lerp.html) [examples](https://en.cppreference.com/w/cpp/numeric/lerp.html)). Personally I look specifically for that name in a new codebase/api. Though I understand that it could sound a bit arcane if one isn't familiar with it. Maybe a comment? Alternatively I could rename it 'linearInterpolation'? Your call 🫡

nabramow: Yeah adding a comment would be good! Since we have a lot of new people in the code base decently often, that helps!

nabramow (dismissed):
Good to go!

nabramow (dismissed):
Approving again, not sure why the last didn't go through!

Ahh I think @jesseallhands needs to approve.

jesseallhands (dismissed):
LGTM!

fbalda:
> Approving again, not sure why the last didn't go through!
> 
> Ahh I think @jesseallhands needs to approve.

@nabramow It went through but I added more commits, so it got dismissed again. Added the last comment now, so after the next approval I will merge it

nabramow (approved):

PR #6543, opened 2025-07-27
Upgrade react-query from v3 to v5

nabramow:
**Describe briefly what this PR is doing and why**

Closes #6353 

This PR updates react-query to the latest version. This affects all the queries in the app. It also unblocks us to start working on the mobile app as there was a conflict between the old version of react-query and expo.

**Please give clear steps for how the reviewer can best test this PR**

*Please include any necessary dev environment, .env, etc. adjustments.*

- Test as many queries as you can. This is anywhere data is pulled from the database, or updated in the SB.
- For example: creating/editing events, discussions, sending friend requests, edit profile, updating account settings, signing up, sending a host request, sending a message, etc.

**Other**
*Untick the following if you'd prefer that maintainers don't push commits/merge your branch.*
- [x] A maintainer can push commits to my branch
- [x] A maintainer can merge my PR (you can also merge after approval)

PR #6593, opened 2025-08-02
Allow users to appoint/remove event co-organizers

### LINKED ISSUE #5680: Ability to make someone an event co-organizer
We already have the backend APIs for this, and it's possible in the console.

Let's make this possible for regular users.

Acceptance Criteria
- For the existing event attendees only, add a the ellipsis "..." icon to the bottom right of the card
- On click, it should open a little dropdown menu (you can copy NotificationsFeed which has a similar example)
- The menu should have one option: PlusIcon + "Make event co-organizer"
- On click this should pop up a dialog that asks "are you sure you want to make {{firstName}} a co-organizer?" Confirm/Cancel
- Then add that user as a co-organizer

- For the event creator only,  add a the ellipsis "..." icon to the bottom right each card for the existing event co-organizers (except for the creator themselves)
- On click, it should open a little dropdown menu (you can copy NotificationsFeed which has a similar example)
- The menu should have one option: MinusIcon and "Remove co-organizer" 
- On click, this should show a confirmation dialog: "Are you sure you want to remove NAME as co-organizer?" (Confirm/Cancel)
- On confirm, remove that userId from the event co-organizers
- On cancel, close the dialog and do nothing

fbalda:
Added the following functionality:
- Event organizers can appoint other organizers and remove themselves as organizers
- Event creators can appoint and remove organizers other than themselves

Attendees eligible to be added/removed as organizers by the user have a drop-down menu on the right side of their card with the respective option.

Closes #5680

**Web frontend checklist**
- [x] All tests pass

**Other**


nabramow:
@fbalda Backend from #6599 was just merged! If you pull `develop` into this branch you should have access to it now!

You'll need to `generate_protos` as well to see the new apis.

app/web/utils/menuOption.ts
```diff
@@ -0,0 +1,6 @@
+// @TODO(FB): Find suitable location
+export interface MenuOption {
+  icon: JSX.Element;
+  title: string;
+  onClick: () => unknown;
+}
```
fbalda, line 6: @nabramow I wasn't quite sure where to put this or if we want to got this route. Personally I prefer to be restrictive with our core components, and I don't think it is necessary to add arbitrary children to a menu or menu item. This makes it super straightforward to make menus. However it probably doesn't belong in the `utils` folder. WDYT?

app/web/test/fixtures/events.json
```diff
@@ -2,7 +2,7 @@
   {
     "attendanceState": 2,
     "content": "*Be there* or be square!",
-    "creatorUserId": 2,
+    "creatorUserId": 1,
```
fbalda, line 5: Changed this so that the mock current user from the fixtures is the creator of the first event. It doesn't matter yet, but I might add a test to make sure they can't remove themselves or whatever route we want to go in the future

app/web/features/queryKeys.ts
```diff
@@ -123,7 +123,7 @@ export const eventOrganizersBaseKey = "eventOrganizers";
 export const eventOrganizersKey = ({ eventId, type }: EventUsersInput) => [
   eventOrganizersBaseKey,
   eventId,
-  { type },
+  ...(type === "summary" ? ["summary"] : []),
```
fbalda, line 126: @nabramow I think we should probably change all the keys to work like this. This way, invalidating `all` actually also invalidates all `summary` queries, but not the other way round

app/web/utils/menuOption.ts
```diff
@@ -0,0 +1,8 @@
+import { SvgIconComponent } from "@mui/icons-material";
+
+// @TODO(FB): Find suitable location
+export interface MenuOption {
+  icon: SvgIconComponent;
+  title: string;
+  onClick: () => unknown;
+}
```
fbalda: @nabramow I wasn't quite sure where to put this or if we want to got this route. Personally I prefer to be restrictive with our core components, and I don't think it is necessary to add arbitrary children to a menu or menu item. This makes it super straightforward to make menus. However the type probably doesn't belong in the utils folder. WDYT?

app/web/features/queryKeys.ts
nabramow: Ah yeah are we passing this an object then? I'm not even sure if it will recognize changes in an object so this seems fine.

I am doing a lot of refactoring for react-query in #6543. In that I'm refactoring all the query keys to be arrays.

app/web/utils/menuOption.ts
nabramow: FYI We have a component already `EllipsisMenu` that can be used for the menu item design/functionality. In that component we have the types already there.

app/web/components/EllipsisMenu.tsx

I'd say in general good to have the types with the reusable component then just export it out from there.

For stuff used in all different features' folders, we could add something to app/web/types maybe? Or make an `internalTypes` file?

app/web/features/queryKeys.ts
fbalda: [Objects in the array are serialized](https://tanstack.com/query/v4/docs/framework/react/guides/query-keys), so react-query will recognize changes. It's just that invalidating [a,b] invalidates [a,b,c] too, so we should attach a value to the more specific query instead of having both queries be "siblings" IMO

app/web/utils/menuOption.ts
fbalda: Yeah, I am using the `EllipsisMenu` in the actual rendered card, down the hierarchy, it's just that I have to prop drill the menu options and I much prefer writing
```ts
{
  icon: Add,
  onClick: () => {
    foo();
  },
  title: "bar",
}
```
over
```tsx
<MenuItem
  onClick={() => {
    foo();
  }}
>
  <Add fontSize="small" />
  <Typography variant="body2">bar</Typography>
</MenuItem>;

```
every time, where the user has many opportunities to make mistakes, add inconsistent custom styling etc. My way makes the interface obvious, easy to use, very hard to abuse and we have a single, central component for menu styling (which, I'd argue, should be 100% consistent across the app).
I didn't add the type to the `EllipsisMenu` file (yet), because it currently doesn't use my system, but I'd gladly refactor it to do in a separate PR, if you agree with my sentiment.

nabramow: Hmm I do see your point. I feel like we have an established pattern the first way though and the second is a bit more readable.

What if we refactor `EllipsisMenu` to be a bit more opinionated instead? Perhaps taking in an array of options that are something like `[ {icon: <Add/>, text: "Title":, {icon: <Subtract/>, text: "Another title} ]` and then having `EllipsisMenu` iterate through them like:

```typescript
{
menuOptions.map((option) => (
<MenuItem>
{icon}
<Typography>{text}</Typography>
</MenuItem>
}
```

Then it guarantees the style more. I don't think we're using it a ton of places yet, either, just in the my connections page.

app/web/features/queryKeys.ts
nabramow: I'm trying to think why we did it the first way even to begin with haha. Usually we do do it the way you described. Would we not want the `eventId` in the array too though to keep it specific to the event?

nabramow: Also good to know about the objects I learned something today!

fbalda: The `eventId` is still in the array: it's `[eventOrganizersBaseKey, eventId]` for an `all` query and `[eventOrganizersBaseKey, eventId, "summary"]` for summaries

app/web/utils/menuOption.ts
fbalda: Sorry, I feel like I explained my thought process poorly:
My idea is that `EllipsisMenu` should take an array of `MenuOption` eventually, I just didn't want to immediately refactor it (though as you said, it currently isn't used much anyway, so maybe I could), so I took the intermediate step of rendering the menu items in `UserSummary` - eventually this should be done in `EllipsisMenu` directly. If we want to refactor `EllipsisMenu` this way, I would put the `MenuOption` interface there, too

fbalda: But I definitely think we need a way to prevent consumers of `EllipsisMenu` from manually rendering and styling `Typography` and icon components for the menu items each time, that would just be so much duplication and a nightmare to maintain/keep in sync

fbalda: Maybe I do the refactoring first in another PR and we discuss there, so we can be a bit more specific?

nabramow: Yeah let's do that!

nabramow:
< Event organizers can appoint other organizers and remove themselves as organizers

The organizers appointing other organizers and removing them is working great. I don't see how someone can remove themselves as an co-organizer though?

fbalda:
> < Event organizers can appoint other organizers and remove themselves as organizers
> 
> The organizers appointing other organizers and removing them is working great. I don't see how someone can remove themselves as an co-organizer though?

Note that event creators currently can't remove themselves as organizers. We have to implement the ability to transfer the "ownership" of an event to someone else first. Only co-organizers can remove themselves. 
Please let me know if it still doesn't work as expected for you

nabramow:
> Note that event creators currently can't remove themselves as organizers. We have to implement the ability to transfer the "ownership" of an event to someone else first. Only co-organizers can remove themselves. Please let me know if it still doesn't work as expected for you

Oops you are right. I was logged into the wrong account last time. It works!

app/web/features/communities/events/EventAttendees.tsx
```diff
@@ -27,12 +74,44 @@ export default function EventAttendees({ eventId }: EventAttendeesProps) {
...
+                        onClick: () => {
+                          setUserToPromote(user);
+                          setIsCoOrganizerDialogOpen(true);
+                        },
+                        label: t("communities:make_co_organizer:title"),
+                      },
+                    ]
+                  : undefined
+            : undefined
+        }
```
nabramow: Just a nit - maybe break this up into a function or put the undefineds first to make it a bit more readable?

Could it be:

* If isCoOrganizedByCurrentUser and the currentUser is not the eventOwner, return menuItems otherwise undefined?

app/web/features/communities/events/EventOrganizers.tsx
```diff
@@ -29,12 +70,42 @@ export default function EventOrganizers({ eventId }: EventOrganizersProps) {
...
       />
       <EventOrganizersDialog
-        eventId={eventId}
+        eventId={event.eventId}
         open={isDialogOpen}
         onClose={() => setIsDialogOpen(false)}
       />
+
+      <RemoveAsCoOrganizerDialog
+        username={organizerToRemove?.name ?? ""}
```
nabramow, line 95: Weird is it possible that the name is undefined?

app/web/features/communities/events/EventPage.test.tsx
```diff
@@ -206,14 +219,12 @@ describe("Event page", () => {
         name: t("communities:going_to_event"),
       });
 
-      const user = userEvent.setup();
-      // @TODO should be awaited but doesn't work, try again after more package upgrades
-      user.click(attendanceMenuButton);
+      const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
```
nabramow, line 222: Ahhh this was the fix then! I feel like I used this somewhere else then immediately forgot it haha. Good call.

nabramow (dismissed):
Works great! Just one nit but I won't block this for it.

app/web/features/communities/events/EventOrganizers.tsx
fbalda: Not the name, but the `organizerToRemove` can be `undefined`, it holds the user whose "remove as organizer" menu item was last clicked. But if none was clicked yet, it can be undefined.

Also note: I can't just determine the dialog open state by checking if `organizerToRemove` is set, as the dialog has a fade out animation and the name would disappear during that. So I had to add the `isDialogOpen` state

app/web/features/communities/events/EventAttendees.tsx
fbalda: > Could it be:
> 
>     * If isCoOrganizedByCurrentUser and the currentUser is not the eventOwner, return menuItems otherwise undefined?

The problem is that there is 2 "levels" of undefined, if the current user isn't an organizer, they don't get any menus at all. If they are, then the function is passed to run for every attendee. That one then checks if the attendee is the current user, in which case they don't get the option. But I'm sure I'll find a way to make it more readable 

But your comment made me realize a different thing: The function should actually check if the user is an organizer already, in which case they probably either shouldn't show the menu, or show the "remove as organizer" option 🤔 

Showing "remove as organizer" for attendees who are also organizers seems natural, but as attendees and organizers are in different lists, it might be confusing and people might be removed as organizers by accident.
So I'd probably just not show any menu on attendees who are organizers as well.
I think there is probably a lot of UX improvements to be made here, maybe we shouldn't worry too much about it now. 
@jesseallhands WDYT?

Edit: Hid the menu on attendees who are organizers as well for now

nabramow:
Thanks for cleaning up that code - looks great. I noticed one new thing:

When I make someone a co-organizer, then remove them, the menu gets stuck at the top left on Desktop:

Screenshot 2025-08-12 at 8 01 34 AM

I think probably you just need to close the menu after selecting?

fbalda:
> Thanks for cleaning up that code - looks great. I noticed one new thing:
> 
> When I make someone a co-organizer, then remove them, the menu gets stuck at the top left on Desktop:
> Screenshot 2025-08-12 at 8 01 34 AM
> 
> I think probably you just need to close the menu after selecting?

Oops! I now changed the ellipsis menu items to close the menu by default, as I think it is almost always the desired behavior, but also gave the option to turn that behavior off

nabramow:
Hey @jesseallhands did you still have some comments here?

jesseallhands (approved):
> I think there is probably a lot of UX improvements to be made here, maybe we shouldn't worry too much about it now.
@jesseallhands WDYT?

I'm still a proponent of allowing all organizers remove other organizers (except the creator, who shouldn't be able to be removed), but that and other things like the ability to add organizers from your friends list (a very common use-case) can always be added later. 

I think the added functionality of this PR will be a huge improvement! LGTM!

PR #6596, opened 2025-08-02
Convert more MUI to styled

fbalda:
#5299

**Web frontend checklist**
- [x] All tests pass

**Other**


fbalda:
This one is a bit larger because of all the dependencies in the communities page. Also IMO there is an uncomfortable amount of style repetition now. I think it is probably fine to get the upgrade to MUI 6+ done, but should probably be cleaned up/streamlined after

nabramow:
> This one is a bit larger because of all the dependencies in the communities page. Also IMO there is an uncomfortable amount of style repetition now. I think it is probably fine to get the upgrade to MUI 6+ done, but should probably be cleaned up/streamlined after

Yeah for sure. If there are certain ones we are repeating exactly over and over, we can ID those and extract them out.

app/web/features/communities/CommunityPage/PlacesSection.tsx
```diff
@@ -8,24 +9,55 @@ import { useTranslation } from "i18n";
...
 
+const StyledCardContainer = styled(HorizontalScroller)(() => ({
+  [theme.breakpoints.down("sm")]: {
+    left: "50%",
+    marginLeft: "-50vw",
+    marginRight: "-50vw",
+    position: "relative",
+    right: "50%",
+    width: "100vw",
+  },
```
nabramow: I don't think this is being used anywhere so can be deleted. I believe there was some features here that we never implemented years ago.

nabramow (dismissed):
Just a nit about `PlacesSection` which seems it can be deleted but I pulled the branch and checked all of these and everything looks good!

nabramow (approved):
Good to go!

PR #6597, opened 2025-08-02
Convert more MUI to styled

fbalda:
#5299

**Web frontend checklist**
- [x] All tests pass

**Other**


app/web/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -63,28 +61,21 @@ export default function PlaceCard({
...
           />
+
           <CardContent>
-            <LinesEllipsis
-              text={place.title}
-              maxLine={2}
-              component="h3"
-              className={classes.title}
-            />
-            <LinesEllipsis
```
nabramow: Was LinesEllipsis not behaving? ;-) Or why the switch?

app/web/features/communities/CommunityPage/PlaceCard.tsx
```diff
@@ -3,41 +3,41 @@ import {
...
-  useTheme,
 } from "@mui/material";
 import Link from "next/link";
 import { Page } from "proto/pages_pb";
 import React, { useMemo } from "react";
 import LinesEllipsis from "react-lines-ellipsis";
 import { routeToPlace } from "routes";
-import makeStyles from "utils/makeStyles";
+import { theme } from "theme";
 import stripMarkdown from "utils/stripMarkdown";
```
nabramow: Please sanity check me but you might be able to just delete this component and its parent: `PlacesSection`. 

It seems this is only used in `PlacesSection` and that is not used anywhere.

app/web/features/communities/discussions/CreateDiscussionForm.tsx
```diff
@@ -104,13 +102,13 @@ export default function CreateDiscussionForm({
...
-        <div className={classes.actionButtonsContainer}>
+        <StyledActionButtonsContainer>
           <Button loading={isLoading} type="submit">
             {t("communities:post")}
           </Button>
           <Button onClick={handleCancel}>{t("global:cancel")}</Button>
-        </div>
-      </form>
-    </Card>
+        </StyledActionButtonsContainer>
```
nabramow: I know this is unrelated, but would you mind switching the Post and Cancel button order so Post is on the right? I think that's more standard. Probably we should make `Cancel` a grey[50] color instead too to differentiate them:

Screenshot 2025-08-02 at 5 04 32 PM

I'm trying to update these as I see them!

nabramow:
I checked the design for everything and it's looking good! Just the comment about the ones that can be deleted and then I'm happy to approve!

app/web/features/communities/CommunityPage/PlaceCard.tsx
fbalda: I might be blind from all the style refactoring, but I'm not quite sure what you mean 🙈 
I replaced the `LinesEllipsis` with styled versions, I just named them after their previous CSS class to have a more meaningful name. Or am I completely missing the issue?

app/web/features/communities/CommunityPage/PlaceCard.tsx
fbalda: Yeah, they aren't used, but I decided against deleting them. Trying to state my reasons just now I realized they are all bad 🙈 Will delete

app/web/features/communities/discussions/CreateDiscussionForm.tsx
```diff
@@ -1,37 +1,44 @@
...
-    justifyContent: "flex-end",
+  display: "flex",
+  justifyContent: "flex-end",
+}));
+
+const StyledCancelButton = styled(Button)(() => ({
+  backgroundColor: theme.palette.grey[50],
+  color: theme.palette.grey[800],
+  "&:hover": {
+    backgroundColor: theme.palette.grey[100],
```
nabramow, line 39: Thanks for adding this! I would just add a little bit of marginRight as the buttons are touching each other now when I check:
Screenshot 2025-08-03 at 1 38 35 PM

Then I think this is good to go!

fbalda: Thanks, totally missed that. Not sure why the transition to styled broke it though. Changed it to `gap` now, which is cleaner IMO anyway

nabramow: Yeah I was trying to figure that out too 😅

nabramow (approved):
Okay this is good to go now!

PR #6598, opened 2025-08-02
Fixed docker services not working on Apple silicon

fbalda:
The docker services for local dev don't work on apple silicon (tested on M1 MacBook Pro). Adding the target platform fixes the issue. 

- [x] Test on MacOS ARM
- [ ] Test on MacOS x86 
- [x] Test on Linux x86
- [ ] Test on Windows x86

**Other**

### DIFF
--- a/app/docker-compose.test.yml
+++ b/app/docker-compose.test.yml
@@ -20,10 +20,8 @@ services:
       - postgres_tests
     links:
       - postgres_tests
-    platform: linux/amd64
   media_tests:
     build: media
     env_file: media.test.env
     command: pytest src
     restart: unless-stopped
-    platform: linux/amd64
--- a/app/docker-compose.test.yml
+++ b/app/docker-compose.test.yml
@@ -8,6 +8,7 @@ services:
     restart: unless-stopped
     ports:
       - 6544:6544
+    platform: linux/amd64
   backend_tests:
     build: backend
     env_file: backend.test.env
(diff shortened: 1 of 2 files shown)


aapeliv:
What's the error you are getting? I'm OK with this change, and I don't have an apple silicon mac around; but as far as I understand, this will slow down a bit on such a mac? Given it'll have to run an x86_64 machine in a vm?

fbalda:
> What's the error you are getting? I'm OK with this change, and I don't have an apple silicon mac around; but as far as I understand, this will slow down a bit on such a mac? Given it'll have to run an x86_64 machine in a vm?

So it is a common issue on Apple Silicon, that certain images don't natively run on ARM (see [here](https://docs.docker.com/desktop/troubleshoot-and-support/troubleshoot/known-issues/)), however I misinterpreted the error messages a bit and overreacted by adding the platform spec to all services. In fact only the `postgis/postgis` image [doesn't support ARM64](https://hub.docker.com/r/postgis/postgis/tags), the rest work fine. So I removed the platform spec from all the other services. 
This shouldn't change anything for anybody who used docker compose on x86 until now (presumably everyone but me), but I'm gonna check tomorrow on my Desktop Linux and Windows x86 machines just to be safe
(I don't have an x86 Mac, but realistically if it runs on the rest, it will work too)

aapeliv (approved):
Works fien on linux x86

PR #6599, opened 2025-08-03
Backend: Allow event owner to remove co-organizer

nabramow:

Backend to fulfill the ACs for #5680

- Run the backend locally
- Tests should pass
- This is not used on frontend yet so not especially risky.

**Backend checklist**
- [ x ] Formatted my code by running `make format` in `app/backend`
- [ x ] All tests pass

**Other**


app/backend/src/couchers/servicers/events.py
```diff
@@ -1156,18 +1156,32 @@ def RemoveEventOrganizer(self, request, context, session):
...
+        # Determine which user to remove
+        user_id_to_remove = request.user_id if request.HasField("user_id") else context.user_id
+
+        # Check permissions: either removing yourself OR you're the event owner
+        can_remove = (
+            context.user_id == user_id_to_remove  # Self-removal
+            or _is_event_owner(event, context.user_id)  # Event owner removing someone else
+        )
+
+        if not can_remove:
```
aapeliv: This check should be

```py3
        if not _can_edit_event(session, event, context.user_id):
```

app/proto/events.proto
```diff
@@ -352,4 +352,5 @@ message InviteEventOrganizerReq {
 
 message RemoveEventOrganizerReq {
   int64 event_id = 1;
+  optional int64 user_id = 2;
```
aapeliv: Use a `google.protobuf.Int64Value` here instead of optional

aapeliv (changes requested):

app/proto/events.proto
nabramow: Are you sure? I don't see any `google.protobuf.UInt64Value` in the codebase  already only `google.protobuf.UInt32Value` thus far.

aapeliv: Yeah it should work the same way

app/backend/src/couchers/servicers/events.py
nabramow: If I understand correctly, `_can_edit_event` currently only includes event owners and moderators, but not co-organizers.

That means this change would block permission from a co-organizer removing themselves.

nabramow: Am I correct that `_can_edit_event` does NOT include co-organizers currently?

That would mean a co-organizer can no longer remove themselves once someone else makes them a co-organizer.

At least the test fails when I do this and that seems to be why.

app/proto/events.proto
nabramow: Done

app/backend/src/couchers/servicers/events.py
aapeliv: We should probably change it then so that co-organizers can edit the event, right?

Just change `_is_event_owner` to `_is_event_organizer`, and use change `_can_edit_event` to use that.

app/backend/src/tests/test_events.py
```diff
@@ -1989,12 +1989,43 @@ def test_RemoveEventOrganizer(db):
         with pytest.raises(grpc.RpcError) as e:
             api.RemoveEventOrganizer(events_pb2.RemoveEventOrganizerReq(event_id=event_id))
         assert e.value.code() == grpc.StatusCode.FAILED_PRECONDITION
-        assert e.value.details() == errors.EVENT_NOT_AN_ORGANIZER
+        assert e.value.details() == errors.EVENT_EDIT_PERMISSION_DENIED
 
         res = api.GetEvent(events_pb2.GetEventReq(event_id=event_id))
         assert not res.organizer
```
nabramow, line 2001: These were tests that were here before me that started failing once I did the `_can_edit_event` change so please check me if this is right as I don't feel entirely certain.

I think the first one is checking if the person is not an organizer. The second is testing is someone is trying to remove a person who is not an organizer.

Since `_can_edit_event` handles this now, both these fall into that error so it has to be more vague.

At least that's what I'm understanding.

aapeliv (approved):
Pushed a few commits. LTGM if tests pass!

PR #6602, opened 2025-08-03
Only have VSCode autoformat the frontend

nabramow:
Sorry @aapeliv I didn't mean to enforce autosave formatting on the backend as well ;-).

This should make the auto-format only affect the frontend file types.

I noticed this as it was auto-formatting the mjml files weirdly in another branch.


.vscode/settings.json
```diff
@@ -8,40 +8,49 @@
   "files.insertFinalNewline": true,
   "[python]": {
     "editor.tabSize": 4,
-    "editor.rulers": [120]
```
nabramow, line 11: It looks like I added this in a previous PR. I didn't want to change around anything for python so just removing it again.

.vscode/settings.json
```diff
@@ -8,40 +8,49 @@
...
+  "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit",
     "source.formatDocument": "explicit"
   },
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "esbenp.prettier-vscode",
+    "editor.formatOnSave": true
```
nabramow, line 25: Moving it out of the global settings and into each file type.

nabramow:
Throwing you on here as a reviewer as well @fbalda in case you have a better way of doing it. Goal is to keep auto-format for the frontend without causing chaos in the backend files (it wasn't doing it correctly for some files there).

.vscode/settings.json
aapeliv: I added it (your PR reformatted it), please keep it in!

aapeliv (changes requested):

fbalda:
> Throwing you on here as a reviewer as well @fbalda in case you have a better way of doing it. Goal is to keep auto-format for the frontend without causing chaos in the backend files (it wasn't doing it correctly for some files there).

I honestly would force as few workflow/editor behavior settings as possible in the project settings and just have everyone use their own user settings - we can recommend certain ones of course. 
[Project settings override user settings](https://code.visualstudio.com/docs/configure/settings#_settings-precedence), so if someone doesn't like a certain setting they now have to change the project settings and have a modified file sitting in their local repo at all times.

If we want to ensure nobody pushes unformatted/unlinted code (not sure if needed, our PRs prevent merging in those cases anyway), I'd suggest a git-hook manager like [husky](https://www.npmjs.com/package/husky) or [lefthook](https://www.npmjs.com/package/lefthook).
If we still want to force it, maybe [multiple workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces) are a more complete/less cumbersome solution.

Somewhat unrelated, but I think we should also remove `"typescript.preferences.importModuleSpecifier": "relative"` and handle this with ESLint (e.g. [no-relative-import-paths](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths) can be configured to allow certain depths. E.g. with depth=1 `../foo` passes, but `../../bar` requires an absolute import path. And it is also auto-fixable). I volunteer to do this if desired 🤓

nabramow:
> > Throwing you on here as a reviewer as well @fbalda in case you have a better way of doing it. Goal is to keep auto-format for the frontend without causing chaos in the backend files (it wasn't doing it correctly for some files there).
> 
> I honestly would force as few workflow/editor behavior settings as possible in the project settings and just have everyone use their own user settings - we can recommend certain ones of course. [Project settings override user settings](https://code.visualstudio.com/docs/configure/settings#_settings-precedence), so if someone doesn't like a certain setting they now have to change the project settings and have a modified file sitting in their local repo at all times.
> 
> If we want to ensure nobody pushes unformatted/unlinted code (not sure if needed, our PRs prevent merging in those cases anyway), I'd suggest a git-hook manager like [husky](https://www.npmjs.com/package/husky) or [lefthook](https://www.npmjs.com/package/lefthook). If we still want to force it, maybe [multiple workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces) are a more complete/less cumbersome solution.
> 
> Somewhat unrelated, but I think we should also remove `"typescript.preferences.importModuleSpecifier": "relative"` and handle this with ESLint (e.g. [no-relative-import-paths](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths) can be configured to allow certain depths. E.g. with depth=1 `../foo` passes, but `../../bar` requires an absolute import path. And it is also auto-fixable). I volunteer to do this if desired 🤓

We've debated this before and I will say and the issue is our CI tests take quite awhile to run. So if you forget to `yarn format`, it takes 20+ minutes to see it. It can get quite annoying especially if a volunteer just did a quick PR, and then you need to nag them to go back for days, run `yarn format` and re-push before you can merge.

We do also have a lot of juniors come through so the more stuff that can be automated, the better IMO.

So I would like the auto-format in the frontend, but I'm open to moving specific rules around.

Happy to handle more things with eslint, for example.

fbalda:
> > > Throwing you on here as a reviewer as well @fbalda in case you have a better way of doing it. Goal is to keep auto-format for the frontend without causing chaos in the backend files (it wasn't doing it correctly for some files there).
> > 
> > 
> > I honestly would force as few workflow/editor behavior settings as possible in the project settings and just have everyone use their own user settings - we can recommend certain ones of course. [Project settings override user settings](https://code.visualstudio.com/docs/configure/settings#_settings-precedence), so if someone doesn't like a certain setting they now have to change the project settings and have a modified file sitting in their local repo at all times.
> > If we want to ensure nobody pushes unformatted/unlinted code (not sure if needed, our PRs prevent merging in those cases anyway), I'd suggest a git-hook manager like [husky](https://www.npmjs.com/package/husky) or [lefthook](https://www.npmjs.com/package/lefthook). If we still want to force it, maybe [multiple workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces) are a more complete/less cumbersome solution.
> > Somewhat unrelated, but I think we should also remove `"typescript.preferences.importModuleSpecifier": "relative"` and handle this with ESLint (e.g. [no-relative-import-paths](https://www.npmjs.com/package/eslint-plugin-no-relative-import-paths) can be configured to allow certain depths. E.g. with depth=1 `../foo` passes, but `../../bar` requires an absolute import path. And it is also auto-fixable). I volunteer to do this if desired 🤓
> 
> We've debated this before and I will say and the issue is our CI tests take quite awhile to run. So if you forget to `yarn format`, it takes 20+ minutes to see it. It can get quite annoying especially if a volunteer just did a quick PR, and then you need to nag them to go back for days, run `yarn format` and re-push before you can merge.
> 
> We do also have a lot of juniors come through so the more stuff that can be automated, the better IMO.
> 
> So I would like the auto-format in the frontend, but I'm open to moving specific rules around.
> 
> Happy to handle more things with eslint, for example.

This is understandable, and I suspected this is the reason. I'd still prefer git-hooks in this case, but we don't have to debate this here again of course. Your changes will be a decent fix for now at the very least 👍

fbalda (dismissed):
LGTM 🚀

.vscode/settings.json
nabramow: Oops - just added it back @aapeliv !

aapeliv (approved):

tristanlabelle:
Additional datapoint: As a new dev I ran into what @nabramow pointed out and it would have saved me a few CI roundtrips to have the editor configured as in https://github.com/Couchers-org/couchers/compare/backend/refactor/format-python-on-save-ruff?expand=1 , which makes me run into the reverse issue mentioned by @fbalda :

> so if someone doesn't like a certain setting they now have to change the project settings and have a modified file sitting in their local repo at all times.

At my last company we had CI push a formatting commit to the branch as necessary, and then restart, but I don't know if this is implementable with the gitlab setup (happy to look into it). Maintaining a private workspace file is a burden for new folks, but another solution would be to not override default settings, and to check in a workspace file that configures the editor more thoroughly.

PR #6611, opened 2025-08-04
Convert more MUI to styled

fbalda:
#5299

**Web frontend checklist**
- [x] All tests pass

**Other**


fbalda:
~Forgot to check the tests, need to update one of them~
Edit: Done

nabramow:
From the `Button` in `AttendanceMenu`, would you remind removing this style:

```
          "&:hover": {
            borderColor: theme.palette.grey[300],
            backgroundColor: "#3135390A",
          },
```

It's not from you - but I think that got added by accident and I noticed it while testing.

app/web/features/dashboard/DashboardBanners.tsx
```diff
@@ -1,4 +1,4 @@
-import { Alert as MuiAlert, Typography } from "@mui/material";
+import { Alert as MuiAlert, styled, Typography } from "@mui/material";
 import Alert from "components/Alert";
 import Button from "components/Button";
```
nabramow, line 3: Since we're already importing our `Alert`, do we need the MUI one? Probably we can just switch this to use the one from `components.Alert` as well?

nabramow:
When I go in an event in the `AttendanceMenu` and click "Join event" then change it back to "Not going", I see this warning in the console:

```
event/346/test-event-to-test-adding-the-checkbox-to-edit-event:1 Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
Element with focus: <li.MuiButtonBase-root MuiMenuItem-root MuiMenuItem-gutters MuiMenuItem-root MuiMenuItem-gutters css-1l2ahug-MuiButtonBase-root-MuiMenuItem-root>
Ancestor with aria-hidden: <div.MuiPopover-root MuiMenu-root MuiModal-root css-10nakn3-MuiModal-root-MuiPopover-root-MuiMenu-root#event-page-attendance-menu> <div role=​"presentation" id=​"event-page-attendance-menu" class=​"MuiPopover-root MuiMenu-root MuiModal-root css-10nakn3-MuiModal-root-MuiPopover-root-MuiMenu-root" aria-hidden=​"true">​…​</div>​
```

It doesn't seem to be breaking anything, but if the solution is obvious, might be good to fix here. Not sure if this is from you or not, might have been there before.

app/web/features/dashboard/Hero/HeroLinks.tsx
```diff
@@ -30,60 +31,35 @@ const useStyles = makeStyles((theme) => ({
...
+          {t("find_a_host")}
+        </StyledDefaultTab>
 
-            <StyledLink
-              underline="none"
-              href={routeToEditProfile("home")}
-              className={classes.tabAppearance}
-            >
-              {t("become_a_host")}
-            </StyledLink>
```
nabramow, line 75: I don't think this is from you, but this is very weird that they are tabs. Since when you click one, it re-routes, so you can never actually change the tab 😂.

I'm going to assume someone put in tabs just for the underline and hover style? 🤷‍♀️

Anyway, it's on my mental to-do list to refactor.

I noticed because you added all this `isSelected` logic and I was like...what's being selected actually?!

fbalda: I mean they are tabs in name only, I think that makes some sense, as they at least present as tabs even though they don't behave like tabs. For the same reason I just went with `isSelected` for the prop, even though it is just static and could actually be baked in with the HOC. I'm open for naming suggestions of course.

I think this might actually be more of a design issue - don't know how common fake tabs are, but maybe we should consider not presenting them as equal options? It looks sleek and modern though

(Also in general whenever I encounter decisions like these while converting components to styled, I err on the side of keeping it the same. I'll gladly refactor later, I actually quite enjoy cleaning up)

app/web/features/dashboard/DashboardBanners.tsx
fbalda: Our Alert only allows string children, don't know if changing that breaks other code. Quite scared to do that for a styled upgrade PR. Also something I'll gladly clean up and refactor later if we don't do it now

app/web/features/dashboard/Hero/HeroLinks.tsx
nabramow: No totally I think you did the right thing since we're not fully refactoring here. I just hadn't looked in this particular part of the code before and was like 🧐.

app/web/features/dashboard/DashboardBanners.tsx
nabramow: Ah okay, that makes sense.

fbalda:
> When I go in an event in the `AttendanceMenu` and click "Join event" then change it back to "Not going", I see this warning in the console:
> 
> ```
> event/346/test-event-to-test-adding-the-checkbox-to-edit-event:1 Blocked aria-hidden on an element because its descendant retained focus. The focus must not be hidden from assistive technology users. Avoid using aria-hidden on a focused element or its ancestor. Consider using the inert attribute instead, which will also prevent focus. For more details, see the aria-hidden section of the WAI-ARIA specification at https://w3c.github.io/aria/#aria-hidden.
> Element with focus: <li.MuiButtonBase-root MuiMenuItem-root MuiMenuItem-gutters MuiMenuItem-root MuiMenuItem-gutters css-1l2ahug-MuiButtonBase-root-MuiMenuItem-root>
> Ancestor with aria-hidden: <div.MuiPopover-root MuiMenu-root MuiModal-root css-10nakn3-MuiModal-root-MuiPopover-root-MuiMenu-root#event-page-attendance-menu> <div role=​"presentation" id=​"event-page-attendance-menu" class=​"MuiPopover-root MuiMenu-root MuiModal-root css-10nakn3-MuiModal-root-MuiPopover-root-MuiMenu-root" aria-hidden=​"true">​…​</div>​
> ```
> 
> It doesn't seem to be breaking anything, but if the solution is obvious, might be good to fix here. Not sure if this is from you or not, might have been there before.

Interesting, I can't reproduce this with the local backend 🤔 

Edit: Can reproduce in Chromium but not in Firefox, will take a look

fbalda:
I fixed the warning now, though I am not entirely sure why it happened in the first place, [other people seem to have the same issue with various MUI components](https://github.com/mui/material-ui/issues/43106#issuecomment-2314809028). I took one of their solutions and it works fine, AFAIU the warning can be ignored though

nabramow (approved):
LGTM!

PR #6625, opened 2025-08-07
Bugfix - Stop propagation on open profile new tab click

### LINKED ISSUE #6623: Some filters (like "Last active") resets after every profile opening
Subject: Some filters (like "Last active") resets after every profile opening
Description:
Using the search with filters active, opening people's profiles.

Results:
Last Active filter resets every time.. 

Backend version: develop-fc594181
Frontend version: develop-fc594181
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36
Screen resolution: 1512x823
Page: https://couchers.org/search#8.85/48.2292/9.245
User: [@vlesyk](https://couchers.org/user/vlesyk) (49020)

nabramow:
Closes #6623 

There was a bug, where if you search a location on the map, then click either the user name link or the open in new tab icon, it would select that user and reset the filters every time.

I added stop propagation to those elements so it's fixed.

You should be able to click to open a profile from map search results in a new tab now without it messing up any existing filters.

**To Test:**

- Go to map search and search a location
- Set a filter, i.e. "last active" or something
- Click either the username link or the icon to pen profile at the top right
- It should open the profile in a new tab and not mess with any filters
- If you click elsewhere on the user card, it will select that user and show them on the map as usual (previously it was also doing this on open in new tab profile click)

### DIFF
--- a/app/web/features/search/SeachResultUserCard.tsx
+++ b/app/web/features/search/SeachResultUserCard.tsx
@@ -173,6 +173,7 @@ const SearchResultUserCard = ({
                 target="_blank"
                 rel="noopener noreferrer"
                 sx={{ fontSize: "1.1rem", overflow: "hidden" }}
+                onClick={(e) => e.stopPropagation()}
               >
                 <Typography
                   variant="h2"
@@ -193,6 +194,7 @@ const SearchResultUserCard = ({
               href={routeToUser(user.username)}
               target="_blank"
               rel="noopener noreferrer"
+              onClick={(e) => e.stopPropagation()}
             >
               <Tooltip title={t("profile:open_profile_new_tab")}>
                 <StyledOpenInNewIcon


fbalda:
I have no issues with the changes, I think they make sense to prevent an additional user card click action, however:
I can't reproduce the original bug to begin with, and I'm struggling to figure out how a card click (which I assume is causing the issue) could affect the filters. I added console logs to the `clearSearchFilters` and `setSearchFilters` reducer actions and even on develop they just don't get called for me when clicking on the links or the card directly, and my filters don't get reset. Tested on Firefox and Chromium. Just mentioning this to make sure we are not overlooking the underlying issue
Edit: Also can't reproduce on production site, I might be missing something

fbalda (approved):

jesseallhands:
I also wasn't able to replicate the bug. Seems to be fixed from my testing

jesseallhands (approved):
Lgtm

nabramow:
@jesseallhands @fbalda Here's a video. What's happening is it's selecting that user when they click to open the profile and it's making all the other pins disappear.


https://github.com/user-attachments/assets/90d492c0-fc61-45e2-a49c-d9b848837300

PR #6633, opened 2025-08-08
Refactor user_model_to_pb and lite_user_to_pb for ghost profiles

### LINKED ISSUE #4297: Ghost profile issues
When people get banned/delete their accounts, the UI ends up looking broken:

Screenshot from 2024-05-07 18-24-03

We should do something about this. I'm not 100% sure what exactly. Ideas:

* Return username, but everything else set to "ghost"/"This user is no longer on Couchers"/empty.
* Return the user ID/username of a "ghost" profile?

(We want some content to remain after a user deletes their account, but also often we just want the content to be removed.)

andrei-kolesnik-california:
PR is doing:
Refactored user_model_to_pb and lite_user_to_pb to ensure consistent handling of deleted and banned users (ghost profiles).
Previously, ghost profile data was not fully aligned between these two functions. This PR unifies the approach and ensures that all relevant fields are set to safe default values for deleted/banned users.

Tests:
Added test_user_model_to_pb_ghost_user and test_lite_user_to_pb_ghost_user to verify that both functions return the correct proto for ghost profiles.

PR closes #4297.

How to test PR:
    - Create a test user.
    - Mark the user as is_deleted = True or is_banned = True in the database.
    - Call user_model_to_pb and verify the returned proto contains only ghost profile defaults.
    - Call lite_user_to_pb and verify the returned proto contains only ghost profile defaults.
    - Run the new tests — all should pass.

Closes #4297

### DIFF
--- a/app/backend/src/couchers/constants.py
+++ b/app/backend/src/couchers/constants.py
@@ -71,6 +71,9 @@ COMMUNITIES_SEARCH_FUZZY_SIMILARITY_THRESHOLD = 0.35
 
 UNKNOWN_ERROR_MESSAGE = "An unknown backend error occurred. Please consider filing a bug!"
 
+GHOST_USER_DISPLAY_NAME = "Deleted user"
+GHOST_USERNAME = "ghost"
+
 # Donation drive start date - set to None to disable donation drive banner
 # When set, users who haven't donated since this date will see a donation banner
 DONATION_DRIVE_START = pytz.UTC.localize(datetime(2025, 11, 1))
--- a/app/backend/src/couchers/constants.py
+++ b/app/backend/src/couchers/constants.py
@@ -71,7 +71,6 @@ COMMUNITIES_SEARCH_FUZZY_SIMILARITY_THRESHOLD = 0.35
 
 UNKNOWN_ERROR_MESSAGE = "An unknown backend error occurred. Please consider filing a bug!"
 
-GHOST_USER_DISPLAY_NAME = "Deleted user"
 GHOST_USERNAME = "ghost"
 
 # Donation drive start date - set to None to disable donation drive banner
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -7,6 +7,7 @@ from sqlalchemy.sql import and_, delete, distinct, func, intersect, or_, union
 
 from couchers import urls
 from couchers.config import config
+from couchers.constants import GHOST_USER_DISPLAY_NAME, GHOST_USERNAME
 from couchers.crypto import b64encode, generate_hash_signature, random_hex
 from couchers.helpers.strong_verification import get_strong_verification_fields
 from couchers.materialized_views import LiteUser, UserResponseRate
@@ -39,6 +40,7 @@ from couchers.proto import api_pb2, api_pb2_grpc, media_pb2, notification_data_p
 from couchers.rate_limits.check import process_rate_limits_and_check_abort
 from couchers.rate_limits.definitions import RATE_LIMIT_INTERVAL_STRING
 from couchers.resources import get_badge_dict, language_is_allowed, region_is_allowed
+from couchers.servicers.blocking import is_not_visible
 from couchers.sql import couchers_select as select
 from couchers.sql import is_valid_user_id, is_valid_username
 from couchers.utils import (
@@ -217,9 +219,7 @@ class API(api_pb2_grpc.APIServicer):
         )
 
     def GetUser(self, request, context, session):
-        user = session.execute(
-            select(User).where_users_visible(context).where_username_or_id(request.user)
-        ).scalar_one_or_none()
+        user = session.execute(select(User).where_username_or_id(request.user)).scalar_one_or_none()
 
         if not user:
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "user_not_found")
@@ -228,15 +228,13 @@ class API(api_pb2_grpc.APIServicer):
 
     def GetLiteUser(self, request, context, session):
         lite_user = session.execute(
-            select(LiteUser)
-            .where_users_visible(context, table=LiteUser)
-            .where_username_or_id(request.user, table=LiteUser)
+            select(LiteUser).where_username_or_id(request.user, table=LiteUser)
         ).scalar_one_or_none()
 
         if not lite_user:
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "user_not_found")
 
-        return lite_user_to_pb(lite_user)
+        return lite_user_to_pb(session, lite_user, context)
 
     def GetLiteUsers(self, request, context, session):
         if len(request.users) > MAX_USERS_PER_QUERY:
@@ -246,11 +244,7 @@ class API(api_pb2_grpc.APIServicer):
         ids = {u for u in request.users if is_valid_user_id(u)}
 
         users = (
-            session.execute(
-                select(LiteUser)
-                .where_users_visible(context, table=LiteUser)
-                .where(or_(LiteUser.username.in_(usernames), LiteUser.id.in_(ids)))
-            )
+            session.execute(select(LiteUser).where(or_(LiteUser.username.in_(usernames), LiteUser.id.in_(ids))))
             .scalars()
             .all()
         )
@@ -271,7 +265,7 @@ class API(api_pb2_grpc.APIServicer):
                 api_pb2.LiteUserRes(
                     query=user,
                     not_found=lite_user is None,
-                    user=lite_user_to_pb(lite_user) if lite_user else None,
+                    user=lite_user_to_pb(session, lite_user, context) if lite_user else None,
                 )
             )
 
@@ -904,9 +898,18 @@ def get_num_references(session, user_ids):
     )
 
 
-def user_model_to_pb(db_user, session, context):
+def user_model_to_pb(db_user, session, context, is_admin_should_show_invisible=False):
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
+
+    if not is_admin_should_show_invisible and is_not_visible(session, context.user_id, db_user.id):
+        # Return an anonymized "ghost" user profile for deleted, banned, and blocked users
+        return api_pb2.User(
+            user_id=db_user.id,
+            username=GHOST_USERNAME,
+            name=GHOST_USER_DISPLAY_NAME,
+        )
+
     num_references = get_num_references(session, [db_user.id]).get(db_user.id, 0)
 
     # returns (lat, lng)
@@ -1081,7 +1084,15 @@ def user_model_to_pb(db_user, session, context):
     return user
 
 
-def lite_user_to_pb(lite_user: LiteUser):
+def lite_user_to_pb(session, lite_user: LiteUser, context):
+    if is_not_visible(session, context.user_id, lite_user.id):
+        # Return an anonymized "ghost" user profile for deleted, banned, and blocked users
+        return api_pb2.LiteUser(
+            user_id=lite_user.id,
+            username=GHOST_USERNAME,
+            name=GHOST_USER_DISPLAY_NAME,
+        )
+
     lat, lng = get_coordinates(lite_user.geom) or (0, 0)
 
     return api_pb2.LiteUser(
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -914,10 +914,7 @@ def user_model_to_pb(db_user, session, context, *, is_admin_see_ghosts=False, is
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
 
-    # If admin mode, skip visibility check and return full user data
-    if is_admin_see_ghosts:
-        pass
-    elif is_not_visible(session, context.user_id, db_user.id):
+    if not is_admin_see_ghosts and is_not_visible(session, context.user_id, db_user.id):
         # User is not visible (deleted, banned, or blocked)
         if is_get_user_return_ghosts:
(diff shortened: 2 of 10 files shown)


app/backend/src/tests/test_db.py
```diff
@@ -125,7 +124,7 @@ def strip_leading_whitespace(lines):
     return [s.lstrip() for s in lines]
 
 
-@pytest.mark.skipif(not run_migration_test(), reason="Migration test disabled")
+# @pytest.mark.skipif(not run_migration_test(), reason="Migration test disabled")
```
nabramow: Do you mean to leave this commented out?

app/backend/src/tests/test_api.py
```diff
@@ -167,6 +169,54 @@ def test_get_user(db):
...
+    assert user_pb.about_place == ""
+    assert user_pb.additional_information == ""
+    assert list(user_pb.language_abilities) == []
+    assert list(user_pb.regions_visited) == []
+    assert list(user_pb.regions_lived) == []
+    assert list(user_pb.badges) == []
+    assert user_pb.friends == api_pb2.User.FriendshipStatus.NOT_FRIENDS
+    assert user_pb.avatar_url == ""
+    assert user_pb.avatar_thumbnail_url == ""
+    assert user_pb.has_strong_verification is False
```
nabramow:
```suggestion
    assert not user_pb.has_strong_verification
```

app/backend/src/tests/test_db.py
andrei-kolesnik-california: Nicole, thank you so much for catching it! The test failed in the CI pipeline. To run it locally, it was necessary to temporarily disable the fixture. Thanks!

app/backend/src/tests/test_api.py
andrei-kolesnik-california: Let me make these changes.

app/backend/src/couchers/servicers/api.py
```diff
@@ -902,6 +902,12 @@ def get_num_references(session, user_ids):
 def user_model_to_pb(db_user, session, context):
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
+    caller_is_admin = False
```
nabramow: Won't this never be true then?

Also, since this is immediately overridden below, do we need this variable?

app/backend/src/couchers/servicers/api.py
```diff
@@ -967,51 +973,89 @@ def user_model_to_pb(db_user, session, context):
...
-        .scalars()
-        .all(),
-        **get_strong_verification_fields(session, db_user),
-        **response_rate_to_pb(response_rate),
-    )
+    if (db_user.is_deleted or db_user.is_banned) and not caller_is_admin:
+        return api_pb2.User(
+            user_id=db_user.id,
+            username=f"ghost{db_user.id}",
+            name="Deleted user",
```
nabramow: Let's move these to a constant above like:

```python
GHOST_USER_DISPLAY_NAME = "Deleted user"
GHOST_USERNAME_PREFIX = "ghost"
```

app/backend/src/couchers/servicers/api.py
```diff
@@ -967,51 +973,89 @@ def user_model_to_pb(db_user, session, context):
...
+            avatar_url=None,
+            avatar_thumbnail_url=None,
+            badges=[],
+            **get_strong_verification_fields(session, db_user),
+            **response_rate_to_pb(None),
+        )
+    else:
+        user = api_pb2.User(
+            user_id=db_user.id,
+            username=db_user.username,
```
nabramow: For readability, so it's easy to immediately tell the difference, what about extracting these out into separate functions? Something like:

```python
    if (db_user.is_deleted or db_user.is_banned) and not caller_is_admin:
        return _create_ghost_user_pb(db_user, session, context)
    
    return _create_full_user_pb(db_user, session, context)
```

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: It’s initialized to False as a default, so the variable is always defined when no caller_id is present.

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: Let me do that.

app/backend/src/tests/test_api.py
```diff
@@ -167,6 +169,54 @@ def test_get_user(db):
...
+    assert list(user_pb.language_abilities) == []
+    assert list(user_pb.regions_visited) == []
+    assert list(user_pb.regions_lived) == []
+    assert list(user_pb.badges) == []
+    assert user_pb.friends == api_pb2.User.FriendshipStatus.NOT_FRIENDS
+    assert user_pb.avatar_url == ""
+    assert user_pb.avatar_thumbnail_url == ""
+    assert not user_pb.has_strong_verification
+
+
```
nabramow, line 298: I think it would be good to also add tests:

* Admin Viewing Ghost Users
* Ghost User Viewing Themselves
* Test for user that is both deleted and banned

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: These parameters aren’t reused anywhere else, so extracting into separate functions feels unnecessary. Instead, I’ll add a short inline comment before each if and else block to make the intent clear.

app/backend/src/tests/test_api.py
andrei-kolesnik-california: Test for user user that is both deleted and banned: The current tests already cover both is_deleted=True and is_banned=True cases individually. Adding a separate test for deleted && banned would be redundant, as any change of the condition from or to and would already break the existing tests. To keep the test suite minimal and avoid duplication, I’d prefer not to add an extra case here.

andrei-kolesnik-california: Let me add test: Admin Viewing Ghost Users. Thanks.

andrei-kolesnik-california: Backend logic about ghost users behaviour did not change overall. Let's formulate and make adjustments within another ticket.

app/backend/src/couchers/servicers/api.py
```diff
@@ -902,6 +903,12 @@ def get_num_references(session, user_ids):
 def user_model_to_pb(db_user, session, context):
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
+    caller_is_admin = False
+    caller_id = getattr(context, "_user_id", None) or getattr(context, "user_id", None)
+    if caller_id:
+        caller = session.get(User, caller_id)
+        caller_is_admin = bool(caller and caller.is_superuser)
```
aapeliv: Change the check to an input flag called `is_admin_should_show_invisible`, then just set that to default False, but true in the admin.py

app/backend/src/couchers/servicers/api.py
```diff
@@ -967,51 +974,91 @@ def user_model_to_pb(db_user, session, context):
...
+            language_abilities=[],
+            regions_visited=[],
+            regions_lived=[],
+            additional_information="",
+            friends=api_pb2.User.FriendshipStatus.NOT_FRIENDS,
+            avatar_url=None,
+            avatar_thumbnail_url=None,
+            badges=[],
+            **get_strong_verification_fields(session, db_user),
+            **response_rate_to_pb(None),
```
aapeliv: You odn't need to fill these in, pb will set missing values to their defaults.

app/backend/src/couchers/servicers/api.py
```diff
@@ -967,51 +974,91 @@ def user_model_to_pb(db_user, session, context):
...
-        .scalars()
-        .all(),
-        **get_strong_verification_fields(session, db_user),
-        **response_rate_to_pb(response_rate),
-    )
+    if (db_user.is_deleted or db_user.is_banned) and not caller_is_admin:
+        # Return an anonymized "ghost" user profile for deleted or banned users
+        return api_pb2.User(
+            user_id=db_user.id,
+            username=f"{GHOST_USERNAME_PREFIX}{db_user.id}",
```
aapeliv: I think we should use the username `ghost` for all of them, not include the user id.

aapeliv (changes requested):

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: Good catch—proto3 returns default values for unset fields. I’ll drop the manual empty/zero assignments and rely on that. Thanks!

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: Sounds good—I'll switch to a fixed GHOST_USERNAME for all ghost users and update the tests.

app/backend/src/couchers/constants.py
```diff
@@ -66,3 +66,7 @@
 ANTIBOT_FREQ = timedelta(hours=48)
 
 EVENT_REMINDER_TIMEDELTA = timedelta(hours=24)
+
+GHOST_USER_DISPLAY_NAME = "Deleted user"
+
+GHOST_USERNAME = "Ghost"
```
aapeliv: Make this lowercase please, so it adheres to username regex!

app/backend/src/couchers/servicers/api.py
aapeliv: This still needs to be done. The way you're doing the check here is both expensive as well as hacky.

app/backend/src/tests/test_api.py
```diff
@@ -167,6 +170,86 @@ def test_get_user(db):
...
+    user2, _ = generate_user()
+
+    with session_scope() as session:
+        u2 = session.merge(user2)
+        setattr(u2, flag, True)
+        session.commit()
+
+    with session_scope() as session:
+        db_user = session.execute(select(User).where(User.id == user2.id)).scalar_one()
+        context = SimpleNamespace(user_id=user1.id)
```
aapeliv: Please don't use `SimpleNamespace`, there's now functions to create the context. It's better to test this via the API as well (which will auto-setup the context).

app/backend/src/tests/test_api.py
```diff
@@ -167,6 +170,86 @@ def test_get_user(db):
...
+    with session_scope() as session:
+        a = session.merge(admin)
+        t = session.merge(target)
+        setattr(t, flag, True)
+        a.is_superuser = True
+        session.commit()
+
+    with session_scope() as session:
+        db_user = session.execute(select(User).where(User.id == target.id)).scalar_one()
+        context = SimpleNamespace(user_id=admin.id)
```
aapeliv: ditto

aapeliv (changes requested):

app/backend/src/couchers/constants.py
andrei-kolesnik-california: Thanks, fixed to lowercase "ghost" so it matches the username regex.

app/backend/src/couchers/servicers/api.py
andrei-kolesnik-california: Thanks a lot 🙏 I’ve optimized the query and added the is_admin_should_show_invisible flag.

app/backend/src/tests/test_api.py
andrei-kolesnik-california: I've replaced the SimpleNamespace context with an API call using api_session and GetUser.

app/backend/src/tests/test_api.py
andrei-kolesnik-california: ditto

app/backend/src/couchers/servicers/api.py
```diff
@@ -216,9 +217,7 @@ def Ping(self, request, context, session):
         )
 
     def GetUser(self, request, context, session):
-        user = session.execute(
-            select(User).where_users_visible(context).where_username_or_id(request.user)
-        ).scalar_one_or_none()
+        user = session.execute(select(User).where_username_or_id(request.user)).scalar_one_or_none()
```
aapeliv, line 231: You will also need to modify the functions below to get lite user and plurals (getliteusers, etc)

app/backend/src/couchers/servicers/api.py
```diff
@@ -899,9 +898,11 @@ def get_num_references(session, user_ids):
     )
 
 
-def user_model_to_pb(db_user, session, context):
+def user_model_to_pb(db_user, session, context, is_admin_should_show_invisible=False):
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
+    caller_is_admin = is_admin_should_show_invisible
```
aapeliv: You used to have another check here. Could you please add it back as a temporary check: I'm a little antsy about this change, so if you add back the check that makes sure those users are only ever shown if the user is admin and API is admin call, that would assuade the fears.

app/backend/src/couchers/servicers/api.py
```diff
@@ -967,51 +968,62 @@ def user_model_to_pb(db_user, session, context):
...
-        parking_details=parkingdetails2api[db_user.parking_details],
-        avatar_url=db_user.avatar.full_url if db_user.avatar else None,
-        avatar_thumbnail_url=db_user.avatar.thumbnail_url if db_user.avatar else None,
-        badges=session.execute(select(UserBadge.badge_id).where(UserBadge.user_id == db_user.id).order_by(UserBadge.id))
-        .scalars()
-        .all(),
-        **get_strong_verification_fields(session, db_user),
-        **response_rate_to_pb(response_rate),
-    )
+    if (db_user.is_deleted or db_user.is_banned) and not caller_is_admin:
```
aapeliv: Use `db_user.is_visible`, or there's a function like that

aapeliv (changes requested):

aapeliv:
@andrei-kolesnik-california waiting on changes from your side

nabramow:
@aapeliv is this still in progress or did you end up doing it in another PR?

Issue #6232, opened 2025-06-25
Add key markdown pages to Weblate for translation


nabramow:
**Blocked:** The english version of these pages should be reviewed, updated and okayed before we do translations.

-------

It seems this is possible here: https://docs.weblate.org/en/weblate-5.3.1/formats/markdown.html

Let's implement it.

We don't need everything, for example it's way too much work to do every blog post.

**NOTE: Before translation actually happens, we should review the English content**

I'm thinking we should do:

- Donate: https://couchers.org/donate
- Our Plan: https://couchers.org/plan 
- Mission & Values: https://couchers.org/mission
- Volunteer: https://couchers.org/volunteer
- Open Source: https://couchers.org/open-source
- FAQ: https://couchers.org/faq
- Our Team: https://couchers.org/team
- Contact Us: https://couchers.org/contact
- Foundation: https://couchers.org/foundation
- Roadmap (at least the text at the bottom if not image text): https://couchers.org/roadmap

nabramow:
Looks like this is possible:

https://docs.weblate.org/en/weblate-5.3.1/formats/markdown.html

nabramow:
Moving this to backlog as we need to sit down and narrow down which pages we realistically want translated.

PR #6637, opened 2025-08-08
Prep markdown file structure for translation

nabramow:
Preps the markdown folder to allow us to add files for translation one-by-one. For now I am just adding the volunteer page as the text has been reviewed. This means the markdown files need to be reachable via the route `/en/volunteer`, `/de/volunteer`, etc.

It's a bit messy for now, because we need to support the older future structure and allow adding files one-by-one to the new structure.

I didn't want Weblate to automatically import stuff that's not ready. There will probably also be markdown files we don't want to translate. That's why I didn't move everything into the new folder structure yet.

Part of #6232.

**TO TEST**: 
- Go to volunteer page and switch languages via the language picker to a few different things. It should work. (new folder structure)
- Go to other markdown pages like FAQ, Mission, etc. and switch languages to a few different things. Should also work. (old folder structure)
- Make sure regular pages still translate: i.e. Profile, Messages, Dashboard, etc. Switch to German, etc.

### DIFF
--- a/app/web/markdown/locales/cs/volunteer.md
+++ b/app/web/markdown/locales/cs/volunteer.md
@@ -1,0 +1,92 @@
+---
+title: "Volunteer: be part of bringing the next generation of couch surfing to life"
+crumb: Volunteer
+description: "Join the global Couchers.org team. Work with passionate professionals and make a huge impact on the couch surfing community."
+type: general
+---
+
+## Our Team
+
+Couchers.org is built and run by a passionate team of volunteers, with everyone contributing their professional skills in their spare time.
+
+We are always looking for more people to join us, in positions such as software engineering, data science, operations, design, marketing, finance, events, volunteer coordination and more!
+
+## Why volunteer with Couchers? The Volunteering Experience.
+
+Volunteering with Couchers means contributing to something meaningful — **building a platform that helps people connect in genuine, non-transactional ways**. We get to focus on projects that help real people, with priorities shaped by our community — not by shareholders or profit motives. We care deeply about what we're doing and about the people we do it with. We're always glad to welcome volunteers and do our best to make your experience enriching and worthwhile. We focus especially on:
+
+- **Professional & personal development**: Volunteering with Couchers offers a chance to build your skills alongside a diverse, global team of smart, committed volunteers. Our volunteers come from a wide range of backgrounds — from top tech companies to startups, and from all over the world. Whether you're looking to deepen your expertise or explore something new, you'll find opportunities to learn, contribute, and grow in a collaborative, international, real-world environment.
+
+- **Have fun & grow your network**: Our volunteers share a passion for the project and a belief in its mission, and come from a wide range of backgrounds and experiences. We create space for connection through casual virtual events and community chats, and have big plans for in-person meetups to bring volunteers and alumni together. It's a chance to connect with thoughtful, like-minded people while contributing to something meaningful.
+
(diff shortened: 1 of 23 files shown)


app/web/markdown/locales/cs/volunteer.md
```diff
@@ -0,0 +1,92 @@
...
+crumb: Volunteer
+description: "Join the global Couchers.org team. Work with passionate professionals and make a huge impact on the couch surfing community."
+type: general
+---
+
+## Our Team
+
+Couchers.org is built and run by a passionate team of volunteers, with everyone contributing their professional skills in their spare time.
+
+We are always looking for more people to join us, in positions such as software engineering, data science, operations, design, marketing, finance, events, volunteer coordination and more!
```
nabramow, line 12: According to what I read about translating Markdown in Weblate, we add the english for now, then that gets used for other translations. I'll need to play around and make sure it works once this is merged. But I needed the filepath to exist first before I can do the setup in Weblate.

app/web/utils/markdownPages.ts
```diff
@@ -1,26 +1,37 @@
...
+// Support both old and new formats:
+// - Old: markdown/<slug>.md (default to 'en')
+// - New: markdown/locales/<lang>/<slug>.md
 export function getAllMarkdownFiles(): Array<string> {
-  return glob.sync("markdown/**/*.md");
+  const oldFormat = glob.sync("markdown/*.md");
+  const oldLangFormat = glob.sync("markdown/*/*.md");
+  const newFormat = glob.sync("markdown/locales/*/**/*.md");
+  return [...oldFormat, ...oldLangFormat, ...newFormat];
 }
```
fbalda, line 12: A slight nitpick, but this part is a bit confusing to me. First thought was: "Why don't we use `markdown/**/*.md`?" but I guess we don't want to include the ones in `/blog`? In which case I think I'd add a comment explaining this and change the glob to `markdown/{*.md,!(blog)/**/*.md}` to make that intention (if it is the intention) clear
Edit: Also the name implies that it should return all, not just the ones we want to translate, so maybe we could adjust that as well, to be more self-explanatory

app/web/utils/markdownPages.ts
```diff
@@ -1,26 +1,37 @@
...
-
-markdown/issues/communities-and-trust.md
-
-into
-
-['issues', 'communities-and-trust']
+Converts filename to { locale, slug } for both formats.
 */
-export function filenameToSlug(filename: string) {
+export function filenameToLocaleAndSlug(filename: string) {
```
fbalda, line 17: Just an observation: That function isn't used anywhere. Disregard if we need it in the future
Edit: You removed the only call in `getAllMarkdownPathsWithLocales`, looks like duplicated logic

app/web/utils/markdownPages.ts
```diff
@@ -29,15 +40,41 @@ export function getAllMarkdownPathsWithLocales(): {
...
   locale: string;
 }[] {
-  const baseSlugs = getAllMarkdownFiles().map(filenameToSlug);
+  // Collect all unique slugs from all formats
+  const files = getAllMarkdownFiles();
+  const slugs = new Set();
+  for (const file of files) {
+    // Use the existing logic to extract slug (ignore locale)
+    let slugArr;
+    if (file.startsWith("markdown/locales/")) {
```
fbalda, line 49: There's a some repetition of string literals, I'd probably extract `markdown/`, `markdown/locales/` and `.md` into variables

nabramow:
Closing as we decided this is not a priority for now and need to sit down and re-define which pages absolutely should be translated.

PR #6640, opened 2025-08-09
Convert more MUI to styled

fbalda:
#5299

**Web frontend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/web/features/donations/Donations.tsx
+++ b/app/web/features/donations/Donations.tsx
@@ -1,6 +1,4 @@
-import { Link, Typography } from "@mui/material";
-import makeStyles from "@mui/styles/makeStyles";
-import classNames from "classnames";
+import { Link, styled, Typography } from "@mui/material";
 import HtmlMeta from "components/HtmlMeta";
 import Markdown from "components/Markdown";
 import Landscape from "features/donations/resources/landscape.jpeg";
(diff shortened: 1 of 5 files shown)


fbalda:
This one is only 2 components, as they were a bit convoluted in their styling and structure, so I decided to refactor them a bit more than usual

app/web/features/donations/Donations.tsx
```diff
@@ -1,238 +1,178 @@
...
 
-const useStyles = makeStyles((theme) => ({
-  donationsImage: {
-    position: "absolute",
-    width: "100%",
-    height: "100%",
-    objectFit: "cover",
-    opacity: 0.3,
-  },
+const LATEST_FINANCIALS_YEAR = latestFinancialsURL.slice(-4);
```
fbalda, line 14: I decided to take the latest financials year from the link URL, so we don't have to keep the URL and year in sync manually. The way I did it, I just slice the last 4 chars off the URL, so we need to make sure future URLs ends with the year. For that reason I added a comment in `routes.ts` and a test to make sure we don't accidentally use an incompatible URL in the future

app/web/features/donations/DonationsBox.tsx
fbalda, line 1: This one especially had a lot of duplicated/redundant code, so I made a lot of changes. The final style is exactly the same AFAICT, except for some default margin, that I decided not to change, to keep the code simpler (would require another styled component). If desired, I can adjust that one as well

app/web/features/donations/Donations.tsx
nabramow: Let's see if we have our shit together enough to not mess that up 😉. We do blog posts in this separate self-hosted CMS thing that opens PRs to the repo, and I assume it just makes the slug the title of the blog post.

In any case, not any more risky than what we're doing now.

nabramow (approved):
Looks good, donations still work!

app/web/features/donations/Donations.tsx
fbalda: I mean, at least the test will fail in this case (assuming it runs on PRs from the CMS) and we can adjust it. Or are we married to a URL once we publish? 😄

PR #6643, opened 2025-08-10
Redirect authenticated users from index to dashboard

### LINKED ISSUE #6464: Going to Couchers.org homepage when I am logged in does not take me to the dashboard
Subject: Going to Couchers.org homepage when I am logged in does not take me to the dashboard
Description:
went to Couchers.org when logged in

Results:
It took me to the marketing landing page.
I should be redirected to the dashboard, I am already logged in.

Backend version: develop-9d1e6fe7
Frontend version: develop-9d1e6fe7
User Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Safari/537.36
Screen resolution: 2344x1209
Page: https://couchers.org/
User: [@itsi](https://couchers.org/user/itsi) (2)

fbalda:
Redirect authenticated users from index to `/dashboard`, otherwise redirect them to `/landing` (which renders the landing page that's currently on index).

I decided to do this with `getServerSideProps`, to not affect our SEO (index permanently redirects to `/landing` for unauthenticated users, so crawlers won't be affected at all) and for a better UX (no page is rendered at all before the user is redirected to the correct page, reducing load times and flickering). 
It does increase server load, though I assume the increase will be negligible.
As we handle auth with our Python backend, I can't actually verify the auth cookie, and instead just check for its existence. This is probably desired though, as it reduces server load and redirects previously authenticated users (aka users with expired cookies) to `/login` (via `/dashboard`) instead of to `/landing`, which feels more natural

> [!IMPORTANT]  
> Doesn't work on preview builds, as the session cookie isn't present

Closes #6464

**Web frontend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/web/appConstants.ts
+++ b/app/web/appConstants.ts
@@ -17,3 +17,5 @@ export const grpcErrorStrings = {
 };
 
 export type ObscureGrpcErrorMessages = keyof typeof grpcErrorStrings;
+
+export const sessionCookieName = "couchers-sesh";
--- a/app/web/components/Footer/Footer.tsx
+++ b/app/web/components/Footer/Footer.tsx
@@ -18,7 +18,6 @@ import { GLOBAL } from "i18n/namespaces";
 import Link from "next/link";
 import { ReactNode } from "react";
 import {
-  baseRoute,
   blogRoute,
   builtWithRoute,
   contactRoute,
@@ -31,6 +30,7 @@ import {
   githubURL,
   helpCenterURL,
   instagramURL,
+  landingRoute,
   missionRoute,
   planRoute,
   redditURL,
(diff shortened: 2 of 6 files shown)


nabramow:
We probably want to remove the `Landing Page` link from the `Footer` component too, as I imagine that will no longer work now!

Is there some way to test this that I'm missing or do we YOLO it and test on prod? :-) Code itself looks good!

fbalda:
> We probably want to remove the `Landing Page` link from the `Footer` component too, as I imagine that will no longer work now!
> 
> Is there some way to test this that I'm missing or do we YOLO it and test on prod? :-) Code itself looks good!

Good catch! Updated the landing page link, so you can still visit it. 
You can test with the local backend. Also we can still test on staging before we release (the cookies are the same), we don't have to YOLO all the way. It just doesn't work on preview

nabramow (approved):
This works great! I tested as both a logged in and logged out user, works as expected. Also the Landing Page link works for the logged in users. 🚀

PR #6658, opened 2025-08-11
fix: handle empty unseen_messages in send_message_notifications

### LINKED ISSUE #6571: ValueError: max() iterable argument is empty
Sentry Issue: [BACKEND-HM](https://couchers.sentry.io/issues/6777104617/?referrer=github_integration)

```
ValueError: max() iterable argument is empty
  File "couchers/jobs/worker.py", line 93, in process_job
    ret = func(message_type.FromString(job.payload))
  File "couchers/jobs/handlers.py", line 262, in send_message_notifications
    user.last_notified_message_id = max(message.id for _, message, _ in unseen_messages)

max() iterable argument is empty
```

andrei-kolesnik-california:
PR fixes a crash in send_message_notifications when unseen_messages is empty by adding a guard before calling max().
Added a regression test to ensure the function works correctly when there are no unseen messages. This test does not interact with a real database because the original code path requires waiting 5 minutes before sending notifications, and such a timeout is not practical in automated testing. Instead, it uses the test database context and fixtures to simulate the scenario instantly.

PR closes #6571

How to test this PR:
   - Run the backend tests locally.
   - Check test_send_message_notifications_empty_unseen_simple in tests/test_notifications.py.
   - The test creates two users, makes them friends, ensures no unseen messages exist, and calls send_message_notifications.
   - Verify that no exceptions are thrown and the function exits gracefully.

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -259,6 +259,9 @@ def send_message_notifications(payload):
                 .order_by(subquery.c.message_id.desc())
             ).all()
 
+            if not unseen_messages:
+                continue
+
             user.last_notified_message_id = max(message.id for _, message, _ in unseen_messages)
 
             def format_title(message, group_chat, count_unseen):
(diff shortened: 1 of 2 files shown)


nabramow (approved):
This looks good to go! I followed the test steps and everything is passing. Ran the backend locally with no issues and loaded the app locally and nothing crashed.

PR #6665, opened 2025-08-13
Tooling Update

fbalda:
Update our frontend tooling, specifically ESLint and prettier, making them stricter and more consistent. Saves code review work and helps prevent errors.
- Upgrade ESLint to v.9 for better monorepo support of the VSCode ESLint plugin with the new [flat config files](https://eslint.org/docs/latest/use/configure/configuration-files).
- Move stylistic rules to prettier if possible (`simple-import-sort` -> `prettier-plugin-sort-imports`).
- Add more, stricter ESLint rules.

This is a first draft and I want to agree on the rules first before I refactor all the files (A LOT of files) to match those rules.

Some reasoning:

- Removed `"@next/next/no-img-element": "off"` (default: "warn"): We should probably refactor to use `<Image>` tags
- Removed `"react/no-unescaped-entities": "off"` (default: `"error"`): Current reasoning: "This is ugly". This only triggers when you legitimately want to include `>`, `"`, `'` or `}` inside a string in JSX. However, it can easily be fixed by just writing the string inside a JS interpolation (`{">"}`) in those cases. IMO that slight annoyance in edge cases doesn't warrant turning off a warning that's more often than not legitimate.
- Removed `"@typescript-eslint/explicit-module-boundary-types": "off"` (default: `"off"`): Off by default
- Removed `"@typescript-eslint/no-non-null-assertion": "off"` (default: `"error"`): IMO this should almost always be avoided and actual undefined handling be used instead. There are exceptions, but they are few and far between and our code contains quite a few sketchy usages already IMO (rule can always be turned off for a file/line, if there is a legitimate exception)

There is currently no real system for what is a warning and what is an error. I did like the previous file and just made all explicit rules warnings. Open to suggestions, however AFAIU it currently doesn't make much of a difference (we don't treat warnings and errors differently).

In general I tend to prefer rules super strict, so I don't have to think when writing my own code or reviewing other people's code. So I am super happy to add more, now or in the future. Though might as well add them now, if I commit to refactoring a lot of code anyway

**Other**


fbalda:
@nabramow Feel free to add additional rules/plugins you like and/or give feedback. Once we agree on a set of rules, I will start refactoring our codebase

.vscode/settings.json
```diff
@@ -54,5 +54,5 @@
   "editor.tabSize": 2,
   "editor.insertSpaces": true,
   "files.trimTrailingWhitespace": true,
-  "eslint.useFlatConfig": false
+  "eslint.useFlatConfig": true
```
nabramow: Ooooo nice this has been on my todo list I think it was blocking our upgrade for another package! Forgot what now though haha.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
+// import reactRefresh from "eslint-plugin-react-refresh";
```
nabramow: What's this for?

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+    },
+    rules: {
+      // Add a space after comments for consistency.
+      // Stylistic rule that isn't handled by prettier
+      "spaced-comment": ["warn", "always"],
+
+      // Console logs should only be used for debugging.
+      // Use custom logging function for actual logging
+      "no-console": "warn",
+
```
nabramow: Guess we need some custom logging then, ahaha. Though we do have Sentry for errors.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+      "@typescript-eslint/no-empty-function": "warn",
+
+      // Force promises to be awaited/handled with .catch,
+      // can be bypassed by adding "void" in front of call
+      "@typescript-eslint/no-floating-promises": [
+        "warn",
+        { ignoreVoid: true },
+      ],
+
```
nabramow: Is there a way to force them to all be awaited rather than with .catch? I actually don't remember seeing much .catch but I would support that.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+        {
+          vars: "all",
+          varsIgnorePattern: "^_",
+          args: "all",
+          argsIgnorePattern: "^_",
+          caughtErrors: "all",
+          caughtErrorsIgnorePattern: "^_",
+          ignoreRestSiblings: true,
+        },
+      ],
```
nabramow: So basically this is making it a warning instead of an error? Turning the typescript one off and having the others set to "warn" I mean.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+        },
+        // Allow "&" styling for styled components
+        {
+          selector: "objectLiteralProperty",
+          filter: {
+            regex: "^&",
+            match: true,
+          },
+          format: null,
+        },
```
nabramow: What's this one for?

app/web/package.json
```diff
@@ -15,8 +16,8 @@
     "create-translation-files": "node scripts/create-translation-files.js",
     "delete-translation-files": "node scripts/delete-translation-files.js",
     "jest": "jest --modulePaths=.",
-    "lint": "next lint",
-    "lint:fix": "next lint --fix .",
+    "lint": "eslint",
+    "lint:fix": "eslint --fix .",
```
nabramow, line 20: Unrelated but how do you feel if we make `format` run both prettier and linting? Most codebases I've worked in this is the case.

app/web/package.json
```diff
@@ -108,10 +111,8 @@
     "timezone-mock": "^1.3.2",
     "ts-node": "^10.9.2",
     "typescript": "^5.8.3",
+    "typescript-eslint": "^8.39.1",
     "whatwg-fetch": "^3.6.20"
   },
-  "resolutions": {
-    "@types/react": "^18.3.1"
-  },
```
nabramow, line 111: I think we need this until we upgrade React to v19. This was also in the migration guide for the MUI upgrade I just did that if we're not on v19 yet I need to add this.

nabramow:
A few questions but overall looks good! I have some suggestions of additional rules, let me know what you think:

Add `eslint-plugin-jest` and then add the rules:

    'jest/no-test-prefixes': 
    'jest/no-disabled-tests': 
    'jest/consistent-test-it': ['error', { fn: 'it' }],
    
 While I was doing the MUI upgrade I realized I had disabled an entire test suite ages ago while trying to fix a test so this comes in handy. We use it at work and prevents me from forgetting about unskipping tests.
 
 
 Means if we have tests we want to deactivate temporarily we either need to comment them out, fix them, or make a decision but I think that's good tbh.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+  compat.config({
+    extends: ["next"],
+    settings: {
+      next: {
+        rootDir: ".",
+      },
+    },
+  }),
+  eslint.configs.recommended,
+  tseslint.configs.strict,
```
nabramow: So this replaces a lot of our previously explicit ones with the defaults if I understand correctly?

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+  {
+    languageOptions: {
+      parser: eslintParser,
+      parserOptions: {
+        ecmaVersion: 2021,
```
nabramow:
```suggestion
        ecmaVersion: 'latest',
```

Might be time to get with the future? This is quite out of date now.

app/web/eslint.config.js
nabramow: Do we need to add `prettierConfig` from `eslint-config-prettier` here to avoid them fighting?

app/web/prettier.config.js
```diff
@@ -0,0 +1,13 @@
+const config = {
+  importOrder: ["<THIRD_PARTY_MODULES>", "^[./]"],
+  importOrderSeparation: true,
+  importOrderSortSpecifiers: true,
+  printWidth: 79,
```
nabramow: This is such a random number haha.

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+          filter: {
+            regex: "^&",
+            match: true,
+          },
+          format: null,
+        },
+        ...baseNamingConventions,
+      ],
+    },
+  },
```
nabramow: Would also be a fan of `bracketSpacing: true,`:

e.g. { foo: bar } vs {foo: bar}

app/web/prettier.config.js
```diff
@@ -0,0 +1,13 @@
+const config = {
+  importOrder: ["<THIRD_PARTY_MODULES>", "^[./]"],
+  importOrderSeparation: true,
+  importOrderSortSpecifiers: true,
```
nabramow: Would it make sense to make this more specific? I usually have something like:

```
importOrder: [
  "^@?\\w",           // External packages
  "^[^.]",            // Relative imports from root
  "^\\./",            // Same directory
  "^\\.\\./"          // Parent directories
],
importOrderGroupBy: "module", // Group by type
```

app/web/.eslintignore
```diff
@@ -1 +0,0 @@
-proto/**
```
nabramow, line 1: Why do we want to remove this?

app/web/eslint.config.js
```diff
@@ -0,0 +1,200 @@
...
+];
+
+export default tseslint.config([
+  {
+    ignores: ["node_modules/**/*", ".yarn/**/*", "proto/**", ".next/**"],
+  },
+  // Next config needs compatibility layer for new ESLint versions
+  compat.config({
+    extends: ["next"],
```
nabramow: I wonder if we should use `next/core-web-vitals` instead since it's a bit stricter and we seem to be using next suboptimaly. But not sure how much work this would be.

Probably we want `next/typescript` as well since we're using typescript.

Do we need prettier here too?

app/web/eslint.config.js
fbalda: Yeah, I'd just implement a simple adapter for `console.log` for now. This makes it so that you can use `console.log` for debugging and be reminded to remove it later. We could also add something like a `critical` level, that additionally logs with sentry.

app/web/eslint.config.js
fbalda: That would be nice, will take a look

app/web/eslint.config.js
fbalda: No, the `typescript-eslint` rule [needs to be disabled](https://www.npmjs.com/package/eslint-plugin-unused-imports) for `unused-imports` to work, because the `typescript-eslint` one doesn't distinguish between imports and other variables.  I just set everything to `warn`, because that's how it was before and there is no distinction between the 2 for the purpose of testing with our setup

app/web/eslint.config.js
fbalda: It allows object properties to start with a `&`, which would be disabled by our rules otherwise, because we use that for our css-in-js with MUI. I'd prefer to move away from those if possible, but there might not be alternatives in some cases

app/web/package.json
fbalda: I agree and also do it like that usually 👍

app/web/package.json
fbalda: Ah ok, will add it back then. Wasn't quite sure what it was for, but I guess I would have discovered it while refactoring haha

app/web/eslint.config.js
fbalda: Yes, this is stricter/as-strict as all our previous explicit ones

fbalda:
> A few questions but overall looks good! I have some suggestions of additional rules, let me know what you think:
> 
> Add `eslint-plugin-jest` and then add the rules:
> 
> ```
> 'jest/no-test-prefixes': 
> 'jest/no-disabled-tests': 
> 'jest/consistent-test-it': ['error', { fn: 'it' }],
> ```
> 
> While I was doing the MUI upgrade I realized I had disabled an entire test suite ages ago while trying to fix a test so this comes in handy. We use it at work and prevents me from forgetting about unskipping tests.
> 
> Means if we have tests we want to deactivate temporarily we either need to comment them out, fix them, or make a decision but I think that's good tbh.

I like it 👍

app/web/prettier.config.js
fbalda: Yeah, that should be changed to 80/default. For some reason I needed that to be 1 less in a previous project to play nice with ESLint. But they shouldn't be conflicting here as we are only using stylistic ESLint rules that prettier doesn't handle

app/web/prettier.config.js
fbalda: Yes, I would love that. I actually removed some patterns for this PR, but I'd love to make it stricter for sure. 
I would also prefer to have the absolute (relative from root) imports be prefixed with `@/`, to distinguish them from external packages, and we could then sort `@foo` and `@/bar` differently, too

app/web/.eslintignore
fbalda: ~~Ahh, sorry. I confused it with the `app/proto` directory, which doesn't contain any lintable files 🙈  Will add back~~
Edit: Nvm, got confused by my own PR, that's what you get for switching between branches a lot haha. `.eslintignore` is deprecated now and the ignores are in the `eslint.config.js` file, including `proto/**`

app/web/eslint.config.js
fbalda: - I think `next/core-web-vitals` only upgrades warnings to errors, but can't hurt to use it. 
- Will add `next/typescript`!
- I thought `prettier` won't be needed as the stylistic ESLint rules shouldn't be in the pre-made configs anymore, but I guess next is a bit behind in that regard. I will add it back to be safe

app/web/prettier.config.js
fbalda: I can't find anything about the `importOrderGroupBy` rule, is it a different plugin? I think the `<THIRD_PARTY_MODULES>` specifier handles that functionality with the plugin I'm using
Edit: I read through the docs again. Third-party modules are moved to the top by default, also we don't allow imports from parent directories any longer, so if you agree about the `@/` prefix for absolute imports, I'd just make it this: `["^@/(.*)$", "^[.]"]`, aka Third-party first, then `@/` prefixed/absolute, then relative

app/web/eslint.config.js
fbalda: This is handled by prettier

app/web/eslint.config.js
fbalda: Sure, let's try it 👍

fbalda:
An additional tooling suggestion, which we might or might not want to add here:
I really like something like [typebox](https://www.npmjs.com/package/@sinclair/typebox) to validate and type env variables.
Would allow us to do something like:
```
const configSchema = Type.Object({
  DEV_BACKEND_PORT: Type.Optional(Type.Integer()),
  TEST_BACKEND_PORT: Type.Optional(Type.Integer()),
  PORT: Type.Optional(Type.Integer()),
  SESSION_SECRET: Type.String({
    minLength: MIN_SECRET_LENGTH,
    default: "0".repeat(MIN_SECRET_LENGTH),
  }),
});
```
and then export the typed and validated config, so we don't have multiple `process.env` accesses scattered in our code. Need to check how to best do this with next though, as I think it just string replaces accesses to `process.env.x` and I think we can't actually parse `process.env`

app/web/eslint.config.js
fbalda: Actually adding `next/typescript` breaks the config, because it defines the `typescript-eslint` plugin again. So I checked the rules and they are actually more lenient than `tseslint.configs.strict`, which I was using, so I removed it again

fbalda: And I added the prettier config to the main array and not inside the compat, which is just needed for the next configs

fbalda:
I removed all the standard prettier settings, as they were all set to their defaults, plus one superfluous setting that was a leftover from a plugin that we aren't using

nabramow:
> An additional tooling suggestion, which we might or might not want to add here: I really like something like [typebox](https://www.npmjs.com/package/@sinclair/typebox) to validate and type env variables. Would allow us to do something like:
> 
> ```
> const configSchema = Type.Object({
>   DEV_BACKEND_PORT: Type.Optional(Type.Integer()),
>   TEST_BACKEND_PORT: Type.Optional(Type.Integer()),
>   PORT: Type.Optional(Type.Integer()),
>   SESSION_SECRET: Type.String({
>     minLength: MIN_SECRET_LENGTH,
>     default: "0".repeat(MIN_SECRET_LENGTH),
>   }),
> });
> ```
> 
> and then export the typed and validated config, so we don't have multiple `process.env` accesses scattered in our code. Need to check how to best do this with next though, as I think it just string replaces accesses to `process.env.x` and I think we can't actually parse `process.env`

Hmm we actually do have the env types in app/web/env.d.ts, but not sure if we're properly using that.

app/web/.eslintignore
nabramow: Ahhhh good to know, thanks!

app/web/prettier.config.js
nabramow: Yeah I am cool with the `@` prefix!

fbalda:
> > An additional tooling suggestion, which we might or might not want to add here: I really like something like [typebox](https://www.npmjs.com/package/@sinclair/typebox) to validate and type env variables. Would allow us to do something like:
> > ```
> > const configSchema = Type.Object({
> >   DEV_BACKEND_PORT: Type.Optional(Type.Integer()),
> >   TEST_BACKEND_PORT: Type.Optional(Type.Integer()),
> >   PORT: Type.Optional(Type.Integer()),
> >   SESSION_SECRET: Type.String({
> >     minLength: MIN_SECRET_LENGTH,
> >     default: "0".repeat(MIN_SECRET_LENGTH),
> >   }),
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and then export the typed and validated config, so we don't have multiple `process.env` accesses scattered in our code. Need to check how to best do this with next though, as I think it just string replaces accesses to `process.env.x` and I think we can't actually parse `process.env`
> 
> Hmm we actually do have the env types in app/web/env.d.ts, but not sure if we're properly using that.

The advantage with my approach is that it validates the values as well, so we can fail even at build time (if run inside the `next.config.js`) if the env variables aren't set or set to some unexpected value. Plus we can set default values, too. Then we can derive the types from the same file. Zero duplication.
But either way we maybe don't have to add that in this PR, as there are already a lot of changes

fbalda:
@nabramow If you have no more suggestions, I'd start refactoring

nabramow:
> The advantage with my approach is that it validates the values as well, so we can fail even at build time (if run inside the `next.config.js`) if the env variables aren't set or set to some unexpected value. Plus we can set default values, too. Then we can derive the types from the same file. Zero duplication. But either way we maybe don't have to add that in this PR, as there are already a lot of changes

Yeah that seems good to me! I'm cool whether or not it goes in this PR. Up to you!

nabramow:
> @nabramow If you have no more suggestions, I'd start refactoring

Go for it, I think everything is clarified!

fbalda:
@nabramow 
[Should we maybe enforce a consistent definition style for function components](https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/function-component-definition.md)? I think we are mostly using function expressions but are sometimes using arrow functions as well. I'm partial towards arrow functions, as imo they look more idiomatic in a functional codebase such as ours, but I think I'd still prefer consistent use of function expressions over the mix we have now. Wdyt?

nabramow:
> @nabramow [Should we maybe enforce a consistent definition style for function components](https://github.com/jsx-eslint/eslint-plugin-react/blob/HEAD/docs/rules/function-component-definition.md)? I think we are mostly using function expressions but are sometimes using arrow functions as well. I'm partial towards arrow functions, as imo they look more idiomatic in a functional codebase such as ours, but I think I'd still prefer consistent use of function expressions over the mix we have now. Wdyt?

Ah yes I have been doing arrow functions as I make new components. I am fine with that, and since it's mostly us two doing the frontend I think we can decide then ahaha. Go for it!

PR #6679, opened 2025-08-15
Remove unused dependencies

nabramow:
Trying to clear unused dependencies to see if it helps decrease the next server memory usage.

That is suggested here:

https://nextjs.org/docs/app/guides/memory-usage


app/web/package.json
```diff
@@ -40,8 +40,6 @@
     "@tanstack/react-query-devtools": "^5.83.0",
     "@tanstack/react-query-persist-client": "^5.83.0",
     "@toast-ui/editor": "^3.2.2",
-    "classnames": "^2.5.1",
-    "country-flag-icons": "^1.5.19",
```
nabramow, line 44: We load this from the CDN now

app/web/package.json
```diff
@@ -98,17 +96,13 @@
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-simple-import-sort": "^12.1.1",
     "eslint-plugin-unused-imports": "^4.1.4",
-    "i18next-resources-to-backend": "^1.0.0",
     "jest": "^29.7.0",
-    "jest-environment-jsdom": "^29.7.0",
```
fbalda, line 103: Are you sure we don't need this? Might be missing something, but we need jsdom to test react components, no?

nabramow: Yeah sorry that was dumb. I need this and also ts-node for tests.

nabramow:
Okay I also decided I don't like the `webpackMemoryOptimization`, so I took it out. It was making the next server restart all the time and seemed to freeze my computer even worse.

PR #6700, opened 2025-08-19
End-to-end tests

fbalda:
Adds the foundation for end-to-end testing plus a few auth tests.

Tests can be run from `app/e2e` with `yarn test`, after installing dependencies and browsers with (`yarn && yarn deps:install && yarn browsers:install`). They run against the local dev server by default (see `app/e2e/readme.md`).
Also includes a Dockerfile for the test runner, in case we want to make e2e testing part of our CD/CI - to be discussed.

Also renamed and moved a few `.env` variables, to make the naming consistent across the repository, while playing around with a potential docker compose setup. Can be reverted if desired, as that is technically out of scope for this PR.

**Web frontend checklist**
- [x] All tests pass

**Other**


app/web/features/auth/login/LoginForm.tsx
```diff
@@ -80,7 +80,7 @@ export default function LoginForm() {
 
   return (
     <>
-      <StyledForm onSubmit={onSubmit}>
+      <StyledForm onSubmit={onSubmit} aria-label="Login">
```
fbalda: Forms need an accessible name to count as `form` landmarks and be findable by role. See [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/form_role#description)

app/web/features/auth/signup/SignupFormContent.tsx
```diff
@@ -24,9 +25,7 @@ export default function SignupFormContent() {
   useEffect(() => {
     const fetchSignupInfo = async () => {
       try {
-        const response = await fetch(
-          "https://couchers.org/api/public/signup-page-info",
-        );
+        const response = await fetch(`${API_BASE_URL}/public/signup-page-info`);
```
fbalda, line 28: I didn't find a reason for the base URL to be hardcoded here. Being able to fetch everything locally makes e2e testing easier/possible

app/web/platform/usePersistedState.ts
```diff
@@ -11,12 +11,21 @@ export function usePersistedState<T>(
...
+    const saved =
+      typeof window !== "undefined" ? window[storage].getItem(key) : null;
+
+    if (!saved) {
+      return;
+    }
+
+    _setState(JSON.parse(saved) as T);
+  }, [key, storage]);
+
```
fbalda, line 28: Our code here wasn't SSR compatible and always failed in e2e testing. While I couldn't reproduce this with manual testing, having the code fully SSR compatible is at least best practice and probably mandatory to prevent certain (potential) issues for actual users, too

app/web/features/auth/login/LoginForm.tsx
fbalda: Actually the label should be translated. Will fix tomorrow

nabramow:
```suggestion
      <StyledForm onSubmit={onSubmit} aria-label={t("global:login")}>
```

app/web/features/auth/signup/SignupFormContent.tsx
```diff
@@ -1,4 +1,5 @@
 import { Link, Typography } from "@mui/material";
+import { API_BASE_URL } from "components/Map";
```
nabramow, line 2: I think we shouldn't export this from the map component since it doesn't have anything to do with the map. The variable is same as `process.env.NEXT_PUBLIC_API_BASE_URL` anyway.

app/web/features/landing/SocialProof.tsx
```diff
@@ -1,6 +1,7 @@
 import { Favorite, Language, Star } from "@mui/icons-material";
 import { Box, Skeleton, Typography, useMediaQuery } from "@mui/material";
 import Divider from "components/Divider";
+import { API_BASE_URL } from "components/Map";
```
nabramow, line 4: Same here

app/web/platform/usePersistedState.ts
nabramow, line 28: Will this fix https://github.com/Couchers-org/couchers/issues/4191 by chance?

I think maybe that's something else but would be nice.

app/e2e/readme.md
```diff
@@ -0,0 +1,27 @@
+# End-to-end tests
+
+## How to run
+
+If you need help setting up the repository, [please refer to the web dev readme](web/readme.md) first
+
```
nabramow: Would probably be good to link to this README from the main README as well to draw attention to the fact that there are E2E tests.

app/e2e/readme.md
```diff
@@ -0,0 +1,27 @@
...
+```sh
+# cd into the repository_root/app/e2e directory
+# assuming you are currently in the repository root:
+cd app/web
+
+# install Playwright
```
nabramow, line 14: This confused be a bit. I was looking for an e2e folder within web. Maybe just say `cd app/e2e`?

app/e2e/readme.md
```diff
@@ -0,0 +1,27 @@
...
+# run tests
+yarn test
+```
+
+> [!NOTE]
+> Tests run against local frontend server by default.
+
+### Configuration
+
+To run the tests against other frontends, set the `FRONTEND_URL`, `TEST_USER_USERNAME` and `TEST_USER_PASSWORD` environment variables to configure the URL and test user credentials
```
nabramow, line 29: Can you give an example of when we'd use this? If we're running against a mobile frontend, for example? I'm trying to think what other frontends we'd use. Like against prod?

app/e2e/readme.md
```diff
@@ -0,0 +1,27 @@
...
+# install Playwright
+yarn install
+# install Playwright browser dependencies
+yarn deps:install
+# install Playwright browsers
+yarn browsers:install
+# run tests
+yarn test
+```
+
```
nabramow, line 23: I did these steps but when my tests ran, I got a bunch of server errors:

  3) [firefox] › test/auth.test.ts:114:9 › Auth › Login › Valid credentials 

    Error: page.goto: NS_ERROR_CONNECTION_REFUSED

  6) [webkit] › test/auth.test.ts:119:9 › Auth › Login › Invalid password 

    Error: page.goto: Could not connect to the server.
    Call log:
      - navigating to "http://localhost:3000/login", waiting until "load"
      
      
Could there be a step missing?

Do we need to run the backend locally maybe? If so I'd definitely add that as a step here.

nabramow:
Could we add a script like `test:e2e` to the package.json in app/web too?

app/web/features/auth/signup/SignupFormContent.tsx
fbalda: I 100% agree, it just happened to be this way already. [I added a suggestion to the tooling PR](https://github.com/Couchers-org/couchers/pull/6665#issuecomment-3216829900), for how to better deal with env variables. Not sure if changing this should be part of this PR, but I'm willing to do so

app/web/platform/usePersistedState.ts
fbalda: Not sure, but I think the mismatch could cause all sorts of bugs in theory. So fingers crossed 😄

app/e2e/readme.md
fbalda: Yeah sorry, that is just a typo

app/e2e/readme.md
fbalda: I think we might want to run tests against staging and/or prod as part of our CD/CI. Also running against staging/prod *might* come in handy, too. E.g. when adding new e2e tests, so you don't have to blow up your RAM by running all our memory hungry tools/services locally 🙈

app/e2e/readme.md
fbalda: Ah yeah, the default setup depends on the local backend. Will add to readme 👍

fbalda:
> test:e2e

You mean like `"test:e2e": "cd ../e2e && yarn test:e2e"`? I feel like that would make it a bit confusing tbh.
I originally planned on moving the yarn root to `/app` and then make `/web`, `/e2e` and `/native` workspaces. That would make `yarn test:e2e` global (as scripts containing ':' are global by default in yarn). Also that would remove some redundancy. Workspaces could also make implementing the native app easier, as we could just reference the web app from the native build tool workspace.
We would have to refactor some of our build scripts though, but at least in the long run it is probably worth it

aapeliv (commented):
This looks good. Please don't rename the env files though.

You can try setting up a gitlab pipeline that runs these automatically.

PR #6705, opened 2025-08-20
Introduced new Communities search implementation

### LINKED ISSUE #5699: Create API to search for communities by keyword
We need an API to search for existing community pages. This API will be used by users on platform to search for existing community pages and by moderators to quickly find communities.

We currently have `communities.Communities.ListCommunities` and `communities.Communities.GetCommunity` based upon the `community_id` but we don't have a way to search by keyword.

The primary use case will be users searching for communities based upon keywords (like "USA" or "United States"), but another may be moderators who search by the `community_id`. The search should query the various community `name` values in the database.

In a future iteration we will want to enable advanced searching, for example to filter results based upon parent (eg countries vs regions vs cities vs neighborhoods).

andrei-kolesnik-california:
This PR introduces a new Communities search implementation with three modes:
- ID search (strict # syntax): supports one or many IDs, arbitrary separators allowed; whitespace after # tolerated. Preserves input order, deduplicates, paginates over the explicit ID list.
- Keyword search (case-insensitive substring): ILIKE %query% over official community names, stable sorting (name ASC, id ASC) and standard pagination.
- Fuzzy fallback (pg_trgm): triggers only when keyword returns zero results and len(query) >= 3. Uses similarity(unaccent(name), unaccent(query)) > threshold with a configurable threshold.

PR closes #5699.

Manual checks (via gRPC client or your CLI wrappers)
1. ID mode (strict #)
- Single: query="#<existing_id>" ⇒ returns that community.
- Multiple w/ separators: query="{pound_sign}1, {pound_sign} 2 : {pound_sign} 3" ⇒ returns [1,2,3] in that order.
- Invalid: query="#", query="#abc" ⇒ INVALID_ARGUMENT: invalid_id_format.
- Pagination: with many IDs, verify page_size and next_page_token.
2. Keyword mode
- query="country 1" (lowercase) ⇒ returns the same set as "Country 1".
- Verify order is name ASC, id ASC.
- Pagination works across pages with stable ordering.
3. Fuzzy fallback
- Use a typo that is not a substring (so keyword = 0) "Coutri 1".
- Expect the correct community (e.g., “Country 1”) present in results.
- Try a query shorter than 3 chars (e.g., "zz") ⇒ empty results, no fuzzy.

Automated tests
- Run backend tests


aapeliv:
I don't understand the three different query types? If I understand correctly the first one is retrieval by ID? Why not a function that does that instead of complicated query parsing logic?

For the search by name; we should default to some good, simple default. In user search we use the trigram index for fuzzy search, I think something like what we do there is fine. Or we could straight up search by substring.

andrei-kolesnik-california:
> I don't understand the three different query types? If I understand correctly the first one is retrieval by ID? Why not a function that does that instead of complicated query parsing logic?
> 
> For the search by name; we should default to some good, simple default. In user search we use the trigram index for fuzzy search, I think something like what we do there is fine. Or we could straight up search by substring.

Yes, the first mode is indeed retrieval by ID, we put it into SearchCommunities so frontend has a single endpoint for all cases instead of using multiple RPCs. If you think it’s better otherwise, I can redo, but to me a single universal lightweight endpoint looks more convenient. The split between substring and fuzzy was intentional: substring search is cheap for the DB and usually enough, while fuzzy with trigrams is much heavier. By checking substring first, we offload the DB and only fall back to fuzzy when needed. That way we get both performance and typo-tolerance without running the expensive query every time.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -144,6 +145,86 @@ def ListCommunities(self, request, context, session):
             next_page_token=encrypt_page_token(str(offset + page_size)) if len(nodes) > page_size else None,
         )
 
+    def SearchCommunities(self, request, context, session):
+        raw_q = (request.query or "").strip()
```
aapeliv: `request.query` is always guaranteed to be a string, no need for `or ""`

app/backend/src/couchers/servicers/communities.py
```diff
@@ -144,6 +145,86 @@ def ListCommunities(self, request, context, session):
...
+        if not raw_q:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, "query_is_empty")
```
aapeliv: Errors need to be descriptive errors in `errors.py`

app/proto/communities.proto
```diff
@@ -268,3 +271,14 @@ message JoinCommunityReq {
 message LeaveCommunityReq {
   int64 community_id = 1;
 }
+
+message SearchCommunitiesReq {
+  string query = 1;
+  int32 page_size = 2;
```
aapeliv: Make page_size a `uint32` so no need to check for negativity then.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -144,6 +145,86 @@ def ListCommunities(self, request, context, session):
...
+        if len(raw_q) < 3:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, "query_too_short")
+
+        req_size = request.page_size or MAX_PAGINATION_LENGTH
+        page_size = max(1, min(MAX_PAGINATION_LENGTH, req_size))
```
aapeliv: See comment on page_size being a `uint`, then there's no need to do the maxing here.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -144,6 +145,86 @@ def ListCommunities(self, request, context, session):
...
+
+        # page_token → offset with validation
+        try:
+            offset = int(decrypt_page_token(request.page_token)) if request.page_token else 0
+            if offset < 0:
+                raise ValueError
+        except Exception:
+            context.abort(grpc.StatusCode.INVALID_ARGUMENT, "invalid_page_token")
```
aapeliv: Please don't wrap in exception, there's no need to catch the case of an invalid decrypt token. If that happens it's good that it throws an ugly error and sends it to us in sentry -- it shouldn't happen normally, it means either some user is doing something dodgy (good for us to know), or we have a programming error (also really important for us to know).

app/backend/src/couchers/servicers/communities.py
```diff
@@ -144,6 +145,86 @@ def ListCommunities(self, request, context, session):
...
+                    )
+                )
+                .order_by(Cluster.name.asc(), Node.id.asc())
+                .limit(page_size + 1)
+                .offset(offset)
+            )
+            rows = session.execute(q3).scalars().all()
+
+        if not rows:
+            return communities_pb2.SearchCommunitiesRes(communities=[])
```
aapeliv: No need to handle this separately

aapeliv (changes requested):
Regarding the algorithm:

* I believe we want `word_similarity(unaccent(search query), unaccent(clusters.name))`. Note that the order of arguments matters (we search for the first argument in the second argument).
* Your case number (3) is already handled in the the word similarity search.
* I don't think we need both similarity and word similarity? Word similarity should capture it all.
* You will want to add a trigram index on `unaccent(clusters.name)`, not sure what the right incranation of commands is, [see docs](https://www.postgresql.org/docs/current/pgtrgm.html)
* If you use the `<%` operator, it will use the default threshold.


So in short: I think it's fine to just retrieve the `page_size` best hits based on `word_similarity`, that should capture it all.

app/backend/src/couchers/servicers/communities.py
andrei-kolesnik-california: Agree. Thanks!

app/backend/src/couchers/servicers/communities.py
andrei-kolesnik-california: Good point, I’ll move these to descriptive constants in errors.py.

app/proto/communities.proto
andrei-kolesnik-california: Makes sense, I’ll change page_size to uint32. Thank you!

app/backend/src/couchers/servicers/communities.py
andrei-kolesnik-california: I'll correct it in accordance to new proto.

app/backend/src/couchers/servicers/communities.py
andrei-kolesnik-california: I'll remove this block entirely and just return the top page_size communities, as you commented.

app/backend/src/couchers/servicers/communities.py
andrei-kolesnik-california: Agree.

andrei-kolesnik-california: Regarding the algorithm:
I simplified the implementation to use only word_similarity.
However, I can’t use the <% operator because with the default threshold (0.6 in Postgres) the query "Coutri 1" doesn’t match "Country 1".
To handle this case, I introduced a constant FUZZY_SIMILARITY_THRESHOLD = 0.35 instead of relying on the database default.

app/proto/communities.proto
```diff
@@ -268,3 +271,13 @@ message JoinCommunityReq {
...
 }
+
+message SearchCommunitiesReq {
+  string query = 1;
+  uint32 page_size = 2;
+}
+
+message SearchCommunitiesRes {
+  repeated Community communities = 1;
+  string next_page_token = 2;
```
aapeliv: You are not supporting pagination here (which is fine), so you should remove this.

aapeliv (changes requested):
One tiny cleanup comment.

You can also add a trigram index to the right table, that would speed up the DB search.

app/proto/communities.proto
andrei-kolesnik-california: Totally agree. Thank you.

app/backend/src/couchers/models.py
```diff
@@ -1876,9 +1880,61 @@ def is_leaf(self) -> bool:
...
             postgresql_where=is_official_cluster,
         ),
+        # trigram index on unaccented name, only for official clusters
+        Index(
+            "idx_clusters_name_unaccented_trgm",
+            name_unaccented,
+            postgresql_using="gin",
+            postgresql_ops={"name_unaccented": "gin_trgm_ops"},
+            postgresql_where=is_official_cluster,
+        ),
```
aapeliv: You do not need to add a new column in the table for this (also triggers are in general a last-resort solution due to how messy they are), you can create an index on a generated column. So I think you just have to replace `name_unaccented` with `func.lower(func.unaccent(name))` or similar.

Also no need to restrict to only official clusters.

app/backend/src/couchers/models.py
```diff
@@ -1876,9 +1880,61 @@ def is_leaf(self) -> bool:
...
+event.listen(
+    Cluster.__table__,
+    "after_create",
+    DDL("CREATE EXTENSION IF NOT EXISTS unaccent;"),
+)
+event.listen(
+    Cluster.__table__,
+    "after_create",
+    DDL("CREATE EXTENSION IF NOT EXISTS pg_trgm;"),
+)
```
aapeliv: We already have these extensions enabled, no need to worry about them.

aapeliv (changes requested):

aapeliv:
Continued in #7077

PR #6723, opened 2025-08-24
Upgrade Sentry

### LINKED ISSUE #6686: Upgrade Sentry
We're already getting warnings about the current version having incorrect peer dependencies as other packages are more up-to-date. We should catch sentry up.

We use these packages from sentry:

    "@sentry/nextjs": "^8.47.0",
    "@sentry/react": "^8.51.0",
    "@sentry/cli": "^2.50.0",
    "sentry-testkit": "^6.2.2",

Sentry migration guide v8 to v9:
https://docs.sentry.io/platforms/javascript/guides/react/migration/v8-to-v9/

Sentry migration guide v9 to v10:
https://docs.sentry.io/platforms/javascript/guides/react/migration/v9-to-v10/

nabramow:
Closes #6686 

Following the manual setup for Sentry here:
https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/

As after upgrading looks like we need to initialize it via these specific files rather than in `_app.ts`.

That config seems to be the only thing that changed that we use related to upgrading @sentry/react and @sentry/next.

I also took out @sentry/cli as it wasn't being imported anywhere.

### DIFF
--- a/app/web/next-env.d.ts
+++ b/app/web/next-env.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/pages/api-reference/config/typescript for more information.
--- a/app/web/package.json
+++ b/app/web/package.json
@@ -31,8 +31,8 @@
     "@mui/system": "^7.3.1",
     "@mui/utils": "^7.3.1",
     "@mui/x-date-pickers": "^8.10.0",
-    "@sentry/nextjs": "^8.47.0",
-    "@sentry/react": "^8.51.0",
+    "@sentry/nextjs": "^10.5.0",
+    "@sentry/react": "^10.5.0",
     "@stripe/stripe-js": "^7.8.0",
     "@tanstack/query-sync-storage-persister": "^5.83.0",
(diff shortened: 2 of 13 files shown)


app/web/next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
```
nabramow, line 3: I did not upgrade this, it got changed when I ran `yarn build` so I think it's okay.

app/web/package.json
```diff
@@ -104,6 +103,7 @@
     "sentry-testkit": "^6.2.2",
     "ts-node": "^10.9.2",
     "typescript": "^5.8.3",
+    "webpack": "^5.101.3",
```
nabramow, line 106: due to a peer dependency error, added this back (we used to have it in here before).

nabramow:
Hey @aapeliv if you wouldn't mind just looking at the Dockerfile part as you added the release bit originally. I wasn't sure if you meant to override the sentry release when adding it to the sourcemaps or were just adding this additionally.

If overriding, they want is to put it in an `SENTRY_RELEASE` env var now, rather than just a value in `Sentry.init`. I assumed that's what you wanted, but if not let me know.

aapeliv (approved):
LGTM LETS GOOO

PR #6724, opened 2025-08-24
Don't redirect on profile save

### LINKED ISSUE #6668: Don't redirect after EditProfile save
The problem right now is if I edit my profile on the first tab, then save, it redirects to my profile page. If I want to edit the "My Home" section as well, I have to go back again.

Acceptance Criteria
- After EditProfile save, it should stay on the same tab, for both the "About Me" and the "My Home" tabs
- The save bar should disappear again after save
- The user should get some kind of feedback that it saved. Maybe a success toast in the center that says "Profile changes saved!"

nabramow:
Closes #6668 

The problem right now is if I edit my profile on the first tab, then save, it redirects to my profile page. If I want to edit the "My Home" section as well, I have to go back again.

This PR removes the redirect and shows a snackbar toast with a success message instead.

**TO TEST:**
- Go to EditProfile, change something and save.
- Should show success toast, stay on page, and save bar should disappear
- Now go to "My Home" tab, change something and save
- Should show success toast, stay on page and save bar should disappear


jesseallhands (changes requested):
A couple of things I found:

1. On desktop (haven't tried mobile yet): When editing my profile and make a change, then save, then decide to keep editing it further and save again, I don't get the snackbar again. 

2. I'm also not sure if a snackbar is the right thing to use on desktop. Snackbars work well on mobile devices because of the small screen, but if you are on a higher-resolution screen, it's easy to miss the snackbar. I think an auto-dismissing modal "save successful!" would be better.

Not sure we need to do 2 in this PR, but 1 should probably be addressed.

nabramow:
Hey @jesseallhands I've made the #1 changes.

I don't think an auto-dismissing modal is an existing thing. When you say that, you basically just mean a snackbar with a different design (i.e. bigger and more centered)? The concept of a modal is that it has buttons to confirm or cancel, no?

PR #6740, opened 2025-08-28
EditProfile - Fix conditional checkboxes and add tests to cover

nabramow:
More EditProfile fixes. @aapeliv noticed some weirdness with the conditional input field values. This is a fix for that that clears them as well when checkbox is unchecked. Also adds tests for these cases.

This ended up being a bit of a refactor of EditHostingPreferences to break out the form data into a separate component. 

Why? 

- That removes the need for the giant useEffect to handle the initialLoad before the api query finishes and then pre-filling the existing data.
- Means we no longer have to pass defaultValues to all the checkboxes
- Separates the logic for clearing the conditional text inputs without there being a ton of useEffects. Rather let the form handle the defaultValues
- Save button still only loads when something is changed


app/web/features/profile/edit/EditHostingPreference.tsx
```diff
@@ -1,723 +1,53 @@
...
-export default function HostingPreferenceForm() {
-  const { t } = useTranslation([GLOBAL, PROFILE]);
-
-  const {
-    updateHostingPreferences,
-    reset: resetUpdate,
-    isPending: updateIsLoading,
-    isError: updateError,
-  } = useUpdateHostingPreferences();
+export default function EditHostingPreference() {
```
nabramow, line 15: Separating concerns here. EditHostingPreference loads the api data. EditHostingPreferenceForm handles the form state.

app/web/features/profile/locales/de.json
```diff
@@ -60,7 +60,7 @@
     "general": "Allgemein",
     "has_housemates": "Hat Mitbewohner",
     "host_drinking": "Trinkt zu Hause",
-    "host_kids": "Hat Kinder",
+    "has_kids": "Hat Kinder",
```
nabramow, line 63: I changed these everywhere because it was really confusing me that "Hosts people with kids" was "acceptKids" and "Has kids" was "host_kids".

nabramow:
I need to refactor EditProfile in a similar way, but didn't want to explode this PR and I need a break 😂. Also that one works okay.

PR #6773, opened 2025-09-03
Add map region search overrides and improve zoom

nabramow:
The problem. For some countries openstreetmaps/nominatim returns the "administrative" region, which can be overly large. In the case of "France", for example, it returns all French territories which is almost spanning the entire map. See also [nominatim github issue here](https://github.com/osm-search/Nominatim/issues/1420).

You can replicate this directly on openstreetmaps here:
- Go here https://nominatim.openstreetmap.org/ui/search.html
- Search "France" and see what gets highlighted

This means it's a data issue and not a problem with our code.

To fix this, I've added overrides for bboxes of affected regions.

I've also added some custom zooms for larger regions that were zooming in too close: United States, Russia, Mexico, China, etc. It was hardcoded to zoom level 10 before.

- Go to map search, search "France"
- Look through user results, should all be in France and map zoomed to France
- Search "United States", should be users all in United States and map zoomed nicely to United States
- Search Russia, India, China, Brazil, etc. The zoom should be a good amount, not too zoomed in.

**Web frontend checklist**
- [ x ] All tests pass

**Other**


jesseallhands (changes requested):
I'm still having the problem when I search for Russia. When searching for Russia, I get a bunch of results that aren't in Russia. It may seem like the problem is solved for other countries too (I wasn't able to replicate in other countries) but I wonder if that's just cause we have too few data points.

vas2525:
@nabramow @jesseallhands please, take a look at how it works on my end for France, Russia, Brazil

https://github.com/user-attachments/assets/687f2782-6239-46a3-b554-e8d7181e6863

jesseallhands:
@vas2525 it's not showing you in Brasil because your profile is empty. You gotta click on "show all profiles" to see yours. This is cause empty profiles are hidden by default (see button in the sidebar).

vas2525:
> @vas2525 it's not showing you in Brasil because your profile is empty. You gotta click on "show all profiles" to see yours. This is cause empty profiles are hidden by default (see button in the sidebar).

Right, I've filled out my profile and noticed that when Brazil is selected, it shows 4 hosts/users when in reality only 3 are displayed. This seems to be a recurring problem, as the same issue is happening with other countries.

Country-Specific Bugs
France:
Dashboard Search: When selected from the dashboard, the map incorrectly focuses on Russia, though the correct French profiles are displayed if you click "Show all profiles."

Search Bar: The map and profiles initially appear correctly. However, clicking "Show all profiles" incorrectly includes profiles from other countries, specifically X, Spain, and Belgium.

Russia:
Dashboard Search: The map is focused incorrectly and shows a profile from Berlin. Clicking "Show all profiles" only shows this same incorrect profile.
Search Bar: The map focuses correctly, but it displays profiles from various countries. Clicking "Show all profiles" continues to show users from different countries, along with one correct profile from Russia (Paula from Barnaul).

China:
Dashboard Search: The map is focused correctly, but no users are displayed. Clicking "Show all profiles" also shows no users.
Search Bar: The map focuses correctly, but no users are displayed. Clicking "Show all profiles" incorrectly shows profiles from India and Russia.

Australia:
 Dashboard - all good
 Search bar - zoom in to close
 

https://github.com/user-attachments/assets/941ad795-eea0-4d68-ad6f-80884d2bc1c3

fbalda:
Why are the dashboard and "search" searches different anyway? Why would we want the logic twice and with different outcomes (e.g. zoom levels/bounding boxes different between dashboard and search)? I might be missing something. If it isn't intended, I'd probably unify the search logic and roll that into this PR as well

vas2525:
@fbalda, I’m not sure yet, and that’s what I’ve managed to find out so far.

nabramow:
@fbalda It just passes the data to the regular search.

@vas2525 This is supposed to be a quick fix. The focus question is "is this better than what we have on prod now?". If not, I rather just close this and leave it as is as it's not a priority. We'd need to totally re-write a bunch of the logic to get it working perfectly and region search is not a major use case.

I rather just disable region searching and force users to use the "search this area" function if this is so buggy to be honest. It's more of a feature we have by accident than on purpose.

fbalda:
> @fbalda It just passes the data to the regular search.
> 
> @vas2525 This is supposed to be a quick fix. The focus question is "is this better than what we have on prod now?". If not, I rather just close this and leave it as is as it's not a priority. We'd need to totally re-write a bunch of the logic to get it working perfectly and region search is not a major use case.
> 
> I rather just disable region searching and force users to use the "search this area" function if this is so buggy to be honest. It's more of a feature we have by accident than on purpose.

Sorry, I should have been more specific: 
I think the search code uses a lot of duplicated code, e.g. your fix applies to the `FloatingSearchControls`, which uses `LocationAutocompleteOutlined` which shares no business logic with the `LocationAutocomplete` that the dashboard search uses, leading to inconsistent behavior.

But I agree that this might be a bigger task (especially to make sure there are no regressions), and maybe not the highest prio. Your changes look good as a band-aid, so I'm fine with merging this, but I think we should make a ticket for this, prio tbd

nabramow:
> Sorry, I should have been more specific: I think the search code uses a lot of duplicated code, e.g. your fix applies to the `FloatingSearchControls`, which uses `LocationAutocompleteOutlined` which shares no business logic with the `LocationAutocomplete` that the dashboard search uses, leading to inconsistent behavior.

So the zoom logic I believe is only relevant to the maps search, so that's okay I think.

The hooks logic will impact the `useGeocodeQuery` hook, and that hook is used multiple places so I believe it should cover all of them. That hook wraps around the logic of turning a user-typed place name and returning with geo-data via the 3rd party open street maps API.

nabramow:
@jesseallhands @vas2525 So looking into this, the issue with Russia is if you draw a box fully around it, it's so big it gets a chunk of the US as well due to the part close to Alaska. So I'm not sure I can cleanly fix this.

Again, I think the better solution is disabling region search.

fbalda:
> > Sorry, I should have been more specific: I think the search code uses a lot of duplicated code, e.g. your fix applies to the `FloatingSearchControls`, which uses `LocationAutocompleteOutlined` which shares no business logic with the `LocationAutocomplete` that the dashboard search uses, leading to inconsistent behavior.
> 
> So the zoom logic I believe is only relevant to the maps search, so that's okay I think.
> 
> The hooks logic will impact the `useGeocodeQuery` hook, and that hook is used multiple places so I believe it should cover all of them. That hook wraps around the logic of turning a user-typed place name and returning with geo-data via the 3rd party open street maps API.

Looked through the code in detail and you are correct that your changes *should* affect both searches the same, but in `HeroSearch.tsx` line 51, we shuffle the bounding box values returned from the search. I assume there was once are reason for that, but it now causes the bounding box to be wrong when searching from the dashboard. Removing the shuffle fixes the difference in search between dashboard and map. Unfortunately I can't suggest changes to unchanged files. It's a simple fix, but I can also push it, if you prefer. 

I think this caused the misunderstandings, as we were looking at different bugs 😄

Edit: Could also open a new PR for that of course

nabramow:
> Looked through the code in detail and you are correct that your changes _should_ affect both searches the same, but in `HeroSearch.tsx` line 51, we shuffle the bounding box values returned from the search. I assume there was once are reason for that, but it now causes the bounding box to be wrong when searching from the dashboard. Removing the shuffle fixes the difference in search between dashboard and map. Unfortunately I can't suggest changes to unchanged files. It's a simple fix, but I can also push it, if you prefer.
> 
> I think this caused the misunderstandings, as we were looking at different bugs 😄
> 
> Edit: Could also open a new PR for that of course

Ahhh yes I can fix that. It was a bug fix at some point but looks like the API sorted themselves out since then!

nabramow:
I'm going to close this as I want to do a PR to turn off region search instead

PR #6813, opened 2025-09-12
Block new chat message notifications when previous is unread

### LINKED ISSUE #6234: Group message notifications within a certain time
When someone messages, it sends a notification for each message. When you next log in, you have a ton of notifications.

Perhaps there's a way to group them within a certain time? Or only send a notification for the first message, unless the last one is marked `is_seen`?

---

[Aapeli:]

This needs to be done carefully, and we need to check how other places do it.

In the backend, the thinking is that notifications for the same topic+key should be "grouped", e.g. in iOS you can have notifs for the same convo be stacked, and the idea would be that the same topic+key (e.g. a given conversation) would be stacked like this.

andrei-kolesnik-california:
This PR introduces a guard in the notification system to avoid generating multiple chat:message notifications for the same group chat when a user already has an unread notification.
Added a helper function user_has_unseen_notifications(session, user_id, key) to check for existing unseen notifs in a chat.
Updated notification generation logic to respect this guard.
Adjusted tests (test_no_more_notifications_if_previous_message_is_unread, test_chat_notifications) to align with the new behavior.
This reduces redundant notifications and ensures users don’t get spammed with multiple alerts for the same conversation.

This PR closes #6234.

How to test this PR:
- Create a group chat between two or more users.
- Send multiple messages from one user without the recipient reading notifications.
- Verify that only one notification is created per recipient until they read it.
- Mark notifications as seen, then send another message → a new notification should be created.
- Run the full test suite to confirm adjustments (pytest).


aapeliv (commented):
I am not sure if this is the right approach: to just not notify. I feel it should be later in the flow: that the notification is generated, but the notification does not get delivered? At least there should be a time limit there.

aapeliv:
Closing this as discussed.

PR #6814, opened 2025-09-13
Improve reference messaging

### LINKED ISSUE #3530: Common issue: leaving friend references after stay

nabramow:
Closes #3530 

- Changes hardcoded friend reference button to "Write Friend Reference" instead of "Write Reference" since it is always a friend reference button anyway
- Adds a dynamic "Write Host Reference/Write Surfer Reference" button. This only shows if there are available host request references for this current logged in user for the profile they are on. It redirects to the proper tab on the messages page.
- Adds a backend "HasGivenHostRequestReference" that returns a boolean if user has already given a reference for the user
- Adds backend tests
- Adds "You already wrote a reference for this user and this stay" alert if user tries to write a reference for the same stay again
- Adds an error if user tries to write a friend reference for a person they are not friends with
- Adds error if user tries to write referenec after 14 days
- Adds frontend tests
- Breaks up logic in `LeaveReferencePage` a bit so it's more readable

Screenshot 2025-09-13 at 8 14 03 PM

Screenshot 2025-09-13 at 8 22 51 PM

**TO TEST:**

- Run backend locally
- Have two accounts open, one in an incognito window is easier so you don’t have to log in and out a lot
- Send a host request from one user to the other and confirm
- Send and accept a friend request
- Go into DB to host_requests table and change the from and to dates to the past so you can write a reference
- Write a reference for one of the users
- Go to http://localhost:1080/#/email/ and click on the write reference link while logged into the account for the user who already wrote the reference
- You should get the “You already wrote a reference” message and click the button to go to that user’s references
- Now go to the other window for the user who didn’t write the reference yet
- Go to http://localhost:1080/#/email/ and click the write reference link
- It should take you to the page to leave a reference with the “Did you stay with?” question
- Go to that user's profile and go to the "References" tab
- You should see a link now to the left of the write a reference button "You have a pending reference for this user"


aapeliv (commented):
Backend looks good to me

fbalda:
I followed the steps, but I don't get an email with a "Write reference" link after writing the first reference, am I missing something?

jesseallhands:
Do we have an error message for the link when a user clicks the link in the email to leave a reference after the time window has passed? I believe this happens quite common (user wants to leave a reference, but too much time has passed).
Or should this be dealt with in a different issue/PR?

nabramow:
> I followed the steps, but I don't get an email with a "Write reference" link after writing the first reference, am I missing something?

Let me think - so you need to have the backend running locally. Then go to `.env.localdev`, copy the `NEXT_PUBLIC_API` value and replace it temporarily in `.env.development`. That should have it use the local api info instead of stage. That might be the issue why nothing is in the mail dev link.

fbalda:
> > I followed the steps, but I don't get an email with a "Write reference" link after writing the first reference, am I missing something?
> 
> Let me think - so you need to have the backend running locally. Then go to `.env.localdev`, copy the `NEXT_PUBLIC_API` value and replace it temporarily in `.env.development`. That should have it use the local api info instead of stage. That might be the issue why nothing is in the mail dev link.

Mhm, I don't get it. I was running everything locally and using the local API, otherwise I couldn't have signed in with the local accounts (Aapeli and one I signed up through maildev), or is there something else I'm missing?

PR #6804, opened 2025-09-09
Add recipient city to rate limit emails; Improve displaying for lists

spreeni:
- Add recipient city to rate limit emails
- Improve displaying for chat participant lists in rate limit emails

closes #6510 

**Backend checklist**
- [x] Formatted my code by running `make format` in `app/backend`
- [ ] Added tests for any new code or added a regression test if fixing a bug
- [x] All tests pass
- [x] Run the backend locally and it works
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Other**
*Untick the following if you'd prefer that maintainers don't push commits/merge your branch.*
- [x] A maintainer can push commits to my branch
- [x] A maintainer can merge my PR (you can also merge after approval)

PR #6826, opened 2025-09-15
Backend/bugfix/report emails have strings rendered as lists

### LINKED ISSUE #6824: Report emails have strings rendered as lists
Here:

https://github.com/Couchers-org/couchers/blob/c9c96ec66873c2d3006331489031e0bb25bdf0a5/app/backend/templates/system/rate_limit_violation_report.md?plain=1#L27

if `value` is a string, it is `iterable`, so it will be printed as `['s','t','r','i','n','g']`, instead of as expected.

spreeni:
When introducing better rendering of lists in [my last PR](https://github.com/Couchers-org/couchers/pull/6804), this affected strings as well. I thought I tested this, but accidentally tested it on a list of one entry, so there it looked like it didn't split the string.

closes #6824 

**Backend checklist**
- [x] All tests pass

**Other**


aapeliv (dismissed):
I'm not sure what's going on here, since the old one was indeed merged, but this is not on top  of that one. Either way it looks good.

spreeni:
@aapeliv  Ah, there was an issue with the commit so now it's up-to-date and showing correctly. Have a look again and let's merge it :)

aapeliv (approved):

PR #6890, opened 2025-10-04
Fix build warnings and improve build time

nabramow:
* Takes out the markdown pages from the SSR pre-load as it was making the ssr log excessive times in red during build
* app/web/pages/dashboard.tsx: Load only GLOBAL, DASHBOARD, NOTIFICATIONS in getStaticProps.
* app/web/features/communities/events/EventCard.tsx: Add optional attendeesCountFormatter prop; fall back to communities:attendees_count so don't need to load would COMMUNITIES namespace in /dashboard (fixes build warning)

**Warnings/errors addressed**
* Fixed large page data warning by trimming SSR namespaces:
`"Warning: data for page "/dashboard" is 149 kB which exceeds the threshold of 128 kB, this amount of data can reduce performance."`
* Changed imports from `next-i18next` to import from `i18n` since `i18n` already imports from `next-i18next` (fixes build warning)


app/web/features/communities/events/EventCard.tsx
```diff
@@ -78,9 +78,15 @@ export const EVENT_CARD_TEST_ID = "event-card";
...
+  // Optional formatter to override attendees count text (used by dashboard to avoid loading COMMUNITIES)
+  attendeesCountFormatter?: (count: number) => string;
 }
 
-export default function EventCard({ event, className }: EventCardProps) {
+export default function EventCard({
+  event,
+  className,
+  attendeesCountFormatter,
+}: EventCardProps) {
```
nabramow, line 89: This translation was the only one from the `COMMUNITIES` namespace loading on the dashboard. Dashboard was giving a warning on build about having too many strings and too big a size, so I made this translation dynamic so we could move a copy to the dashboard translations and remove the COMMUNITIES namespace from the translation SSR preloading.

app/web/i18n/index.ts
```diff
@@ -1,4 +1,4 @@
-import resources from "i18n/resources";
+import type resources from "i18n/resources";
```
nabramow, line 1: This was causing a webpack string serialization warning for being too many strings. This resources file loads every single translation json file. Just exporting the type here rather than all the strings fixes the webpack warning.

app/web/pages/[...slug].tsx
```diff
@@ -6,21 +6,16 @@ import { AUTH, GLOBAL, NOTIFICATIONS } from "i18n/namespaces";
...
-    slug,
-    frontmatter: md.attributes,
-    content: md.html,
-  };
+  return { slug, frontmatter: md.attributes, content: md.html };
 }
 
 export const getStaticPaths: GetStaticPaths = () => ({
-  paths: getAllMarkdownPathsWithLocales(),
+  paths: [],
```
Comment (nabramow) [inline app/web/pages/[...slug].tsx:18]: We talked about this a few months ago on Slack, finally doing it. Especially as we don't translate the markdown files yet, loading a billion copies of them for different locales was showing red scary numbers at build time.

aapeliv (changes requested):
This is broken for me. Steps to repro:

1. Go to vercel link
2. Click on Mission and Values
3. Get following error page

image

nabramow:
> This is broken for me. Steps to repro:
> 
> 1. Go to vercel link
> 2. Click on Mission and Values
> 3. Get following error page

@aapeliv This should be fixed now!

PR #6956, opened 2025-10-19
Fix: Handle notFound in getStaticProps

nabramow:
Fixing a bug introduced in #6890 .

After PR #6890, the catch-all route `[...slug].tsx` was intercepting internal Next.js/webpack requests during development (e.g., `.webpack.hot-update.json`, `.css.map`), causing build errors.

Solution in this PR:
- Return proper 404s for non-existent markdown files

### DIFF
--- a/app/web/features/markdown/MarkdownPage.tsx
+++ b/app/web/features/markdown/MarkdownPage.tsx
@@ -174,7 +174,7 @@ export default function MarkdownPage({
           {crumbs.map((crumb, index) => {
             const isLast = index === crumbs.length - 1;
             return isLast ? (
-              <Typography key={crumb.key} color="textPrimary">
+              <Typography key={`${crumb.key}-${index}`} color="textPrimary">
                 {crumb.value}
               </Typography>
             ) : (
--- a/app/web/pages/[...slug].tsx
+++ b/app/web/pages/[...slug].tsx
@@ -14,21 +14,42 @@ async function getMarkdownPageBySlug(
   return { slug, frontmatter: md.attributes, content: md.html };
 }
 
-export const getStaticPaths: GetStaticPaths = () => ({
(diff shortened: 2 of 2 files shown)


aapeliv:
> Add whitelist regex validation (`/^[a-zA-Z0-9_-]+$/`) to reject invalid slugs

Could you clarify what this does exactly?

nabramow:
> > Add whitelist regex validation (`/^[a-zA-Z0-9_-]+$/`) to reject invalid slugs
> 
> Could you clarify what this does exactly?

@aapeliv Sure, it blocks bad paths like:

.webpack.hot-update.json (contains .)
_next/static/... (starts with _)
.. or . (navigation tricks)
Any special characters that shouldn't be in URLs

a.k.a. we don' want webpack dev files, source maps, next.js internal files, or other internal dev files to get caught, but we do want the markdown files to get through.

Can't just search `.md` because the slugs look like this: `slug = ["blog", "2021", "03", "25", "couch-surfing-online-events"]`.

aapeliv:
So we can't have underscores in markdown generated pages anymore?

This seems a bit hacky to me. Could we not check if the file exists and then only serve it in that case?

nabramow:
> So we can't have underscores in markdown generated pages anymore?
> 
> This seems a bit hacky to me. Could we not check if the file exists and then only serve it in that case?

I read a bit about this a bit more and seems it's only a quirk for the dev environment , i.e. webpack and how next works in dev. So I took out the slug filtering and am just letting the internal dev stuff land in the catch and return `notFound` now.

That seems to work fine. 

Also added a duplicate key fix to `MarkdownPage` as I was seeing a browser warning when clicking on the blog page.

aapeliv (approved):
lgtm

PR #7005, opened 2025-10-29
Add json linting

### LINKED ISSUE #6834: Upgrade next lint
Getting this error when we run `yarn lint:fix` on the frontend:

`next lint` is deprecated and will be removed in Next.js 16.
For new projects, use create-next-app to choose your preferred linter.
For existing projects, migrate to the ESLint CLI:
npx @next/codemod@canary next-lint-to-eslint-cli .

We should run that codemod and check it fixes it.

nabramow:
We keep having issues with the translation json files getting messed up and double nested, etc.

We noticed json is not being linted so it doesn't fail the lint test and lets us merge it. This PR adds json linting.

We had to switch from next lint to regular eslint in the linting script for this as next lint doesn't support it. Should be fine though as next lint seems to be simply a more opinionated wrapper around eslint. We still have the next rules in the plugins.


nabramow:
@aapeliv Before I pull in the main branch - see proof that it's failing for messed up json.

PR #7015, opened 2025-10-29
Migrate to uv for backend dependency management

WouldYouKindly:
This PR migrates dependency management from `pip` to `uv` which is quickly becoming the community's standard. Benefits:

- Speed of dependency resolution and installation (it's very very fast)
- No need to manually manage and activate virtual environments, just use `uv sync / add / remove` to manage dependencies and `uv run ...` to run python modules. The `.venv` is still there if one wants to do things the old-fashioned way :)
- No need to manually manage your python installation: just specify python version in pyproject.toml, and uv will install the required version. So no pyenv etc.


app/backend/src/couchers/crypto.py
```diff
@@ -149,7 +148,7 @@ def aead_generate_key():
     return random_bytes(_aead_key_len)
 
 
-def aead_encrypt(key: bytes, secret_data: bytes, plaintext_data: bytes = b"", nonce: Optional[bytes] = None) -> bytes:
+def aead_encrypt(key: bytes, secret_data: bytes, plaintext_data: bytes = b"", nonce: bytes | None = None):
```
aapeliv: Seems you accidentally dropped the output type

app/backend/Dockerfile
```diff
@@ -23,11 +23,16 @@ COPY --from=build /usr/local/bin/entr /usr/local/bin/entr
 
 WORKDIR /app
 
-COPY requirements.txt /app
-RUN pip install --no-cache-dir -r requirements.txt
+# Make sure uv is in PATH
+ENV PATH="/root/.local/bin:$PATH"
+
+COPY pyproject.toml uv.lock /app/
+RUN wget -qO- https://astral.sh/uv/install.sh | sh && uv sync --frozen --no-install-project
```
aapeliv: Let’s pin the uv version as well, here and elsewhere

aapeliv (commented):
Need to update dependabot for uv and make sure it works. Also it’s unclear to me where we now store the actual human-defined (not pinned) versions, like we used to do in `requirements.in` files?

app/backend/Dockerfile
WouldYouKindly: done

app/backend/src/couchers/crypto.py
WouldYouKindly: The type was incorrect, actually. I fixed it now to be `tuple[bytes, bytes]`

WouldYouKindly:
> Need to update dependabot for uv and make sure it works. Also it’s unclear to me where we now store the actual human-defined (not pinned) versions, like we used to do in requirements.in files?

@aapeliv I've updated the dependabot config according to https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependabot. Unfortunately I couldn't find a way to test it without merging the PR.
w.r.t. non-pinned versions: normally these would be in `pyproject.toml`, but for now I put pinned versions there to not change too much in a single PR. I plan to remove the pins in a separate PR.

app/backend/src/couchers/crypto.py
aapeliv: I don't see the output type, did you push the commit?

aapeliv:
Otherwise LGTM

app/backend/src/couchers/crypto.py
WouldYouKindly: @aapeliv I didn't! Pushed it now.

aapeliv (approved):
LGTM!

PR #7053, opened 2025-10-31
Speed up tests by truncating tables instead of dropping them

WouldYouKindly:
This cuts running time from 20 to 7 minutes.

### DIFF
--- a/app/backend/src/tests/conftest.py
+++ b/app/backend/src/tests/conftest.py
@@ -3,6 +3,21 @@ from tempfile import TemporaryDirectory
 
 prometheus_multiproc_dir = TemporaryDirectory()
 
+environ["PROMETHEUS_MULTIPROC_DIR"] = prometheus_multiproc_dir.name
 
-def pytest_sessionstart(session):
-    environ["PROMETHEUS_MULTIPROC_DIR"] = prometheus_multiproc_dir.name
+from tests.test_fixtures import (  # noqa
+    account_session,
+    auth_api_session,
+    db,
+    create_database,
+    email_fields,
+    fast_passwords,
+    generate_user,
+    mock_notification_email,
+    process_jobs,
+    public_session,
+    push_collector,
+    real_account_session,
+    requests_session,
+    testconfig,
+)
--- a/app/backend/src/tests/test_account.py
+++ b/app/backend/src/tests/test_account.py
@@ -1047,8 +1047,11 @@ def test_reminders(db):
 
 
 def test_volunteer_stuff(db):
+    # taken from couchers/app/backend/resources/badges.json
+    board_member_id = 8347
+
     # with password
-    user, token = generate_user(name="Von Tester", username="tester", city="Amsterdam")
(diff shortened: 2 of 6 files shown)


app/backend/src/tests/conftest.py
```diff
@@ -3,6 +3,21 @@
...
+    fast_passwords,
+    generate_user,
+    mock_notification_email,
+    process_jobs,
+    public_session,
+    push_collector,
+    real_account_session,
+    requests_session,
+    testconfig,
+)
```
WouldYouKindly, line 23: This is a bit more convenient than importing it in every test file.

app/backend/src/tests/test_account.py
```diff
@@ -1047,8 +1047,11 @@ def test_reminders(db):
 
 
 def test_volunteer_stuff(db):
+    # taken from couchers/app/backend/resources/badges.json
+    board_member_id = 8347
```
WouldYouKindly, line 1051: Before we dropped the entire database, so the new user's id was always `1`. Now that we don't, we have to be explicit.

app/backend/src/tests/test_fixtures.py
```diff
@@ -99,24 +99,14 @@
...
-                "DROP EXTENSION IF EXISTS postgis CASCADE;"
-                "CREATE SCHEMA public;"
-                "CREATE SCHEMA logging;"
-                "CREATE EXTENSION postgis;"
-                "CREATE EXTENSION pg_trgm;"
-                "CREATE EXTENSION btree_gist;"
-            )
-        )
+        for table in Base.metadata.tables.values():
+            if table.name in ("regions", "languages", "timezone_areas"):
```
WouldYouKindly, line 106: Not recreatating static data shaves a couple of minutes from a full test run.

app/backend/src/tests/test_fixtures.py
```diff
@@ -99,24 +99,14 @@
 )
 
 
-def drop_all():
+def truncate_all_tables():
     """drop everything currently in the database"""
```
aapeliv, line 103: docstring is stale now

app/backend/src/tests/test_db.py
```diff
@@ -128,7 +135,7 @@ def strip_leading_whitespace(lines):
 @pytest.mark.skipif(not run_migration_test(), reason="Migration test disabled")
 def test_migrations(testconfig):
```
aapeliv: I am thinking we would want this to run on a fully clean database since it's testing the migrations. What do you think?

app/backend/src/tests/test_fixtures.py
WouldYouKindly: I'll address this in a follow-up PR, if you don't mind!

app/backend/src/tests/test_db.py
WouldYouKindly: Yeah, I've updated the code, please take a look.

I also tried to create a separate database for this test, but it produces a small diff - looks like the two ways to apply migrations are not 100% the same. But I feel that it's a topic for a separate PR.

aapeliv (approved):
Lgtm

PR #7077, opened 2025-11-01
Implement community search

### LINKED ISSUE #5699: Create API to search for communities by keyword

aapeliv:

Implement community search, picking up from Andrei. Closes #5699.


nabramow:
I think you need to re-generate the protos so it stops failing the web build?

app/backend/src/tests/test_fixtures.py
```diff
@@ -133,7 +133,7 @@ def create_schema_from_models():
     """
 
     # create the slugify function
-    functions = Path(__file__).parent / "slugify.sql"
+    functions = Path(__file__).parent / "sql_functions.sql"
```
WouldYouKindly, line 126: comment is outdated now

WouldYouKindly (dismissed):

aapeliv:
> I think you need to re-generate the protos so it stops failing the web build?

CI does it automatically. I think this time it was just a transient failure.

app/backend/src/tests/test_fixtures.py
aapeliv: thanks for picking that up

Issue #6403, opened 2025-07-12
Get app navigation working


nabramow:
Get the logged in and logged out nav bar flow working.

Add key mobile buttons at the bottom and see if we like that.

PR #7080, opened 2025-11-01
Send native app message when user logs out

nabramow:

Part of #6403 

Since the authentication is handled in the web app and we're using WebEmbed, we need a way to tell the mobile app when the user logged out for conditional rendering, blocking certain routes, etc.

[Based on these docs](https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#communicating-between-js-and-native), this seems to be the way to do it. We'll send a message to the web app saying "LOGOUT" and then I'll receive in in the mobile app (in a separate PR).

This needs to merge and deploy first before I have access to test in the mobile code.

Not much to test yet.


aapeliv (approved):
lgtm

PR #7087, opened 2025-11-02
Improve SV flow

### LINKED ISSUE #4317: Where is the NFC chip on my passport?
# Picture page

* Finland (2022)
* Germany (2017+): according to website

# Back page

* US (2024)
* Germany (2016)
* Mexico (2023)

# Other

* Australia (2017): middle page (marked with symbol + "chip name")

### LINKED ISSUE #7030: Make strong verification process clearer
- When user clicks "Start strong verification" button on account settings page, redirect to a page that prominitly lists the instructions instead
- The button to start the process with Iris should be at the bottom.
- When they click, open Iris in a new tab so user still has instructions handy
- Make a step and make it bold and obvious "You need to press your phone against the page with the chip. In some passports this may be different (e.g. in US Passport it's on the last page with passport open).

nabramow:
Adds instructions to clarify the strong verification flow and direct links to download the Iris ID app.

Closes #7030
Closes #4317 

Screenshot 2025-11-04 at 12 34 58 AM


aapeliv:
I think this is great!

A few comments:
1. On iOS at least, you don't actually need to download Iris ID: you can do it through an "App clip" that doesn't install the app fully on the phone, and doesn't require going to the app store.
2. On the NFC chip part: I think firstly we have very low coverage on which passports have it where. Second, I'm thinking about how people would use this index; maybe we should have an input where people can type in their country name? Or organize the list by country name alphabetically so they can scan down?

nabramow:
> I think this is great!
> 
> A few comments:
> 
> 1. On iOS at least, you don't actually need to download Iris ID: you can do it through an "App clip" that doesn't install the app fully on the phone, and doesn't require going to the app store.
> 2. On the NFC chip part: I think firstly we have very low coverage on which passports have it where. Second, I'm thinking about how people would use this index; maybe we should have an input where people can type in their country name? Or organize the list by country name alphabetically so they can scan down?

1. I'll add a little comment for apple that it's option and they can just click "Open" instead.

2. I was trying to keep this simple. Rather than make it more complex maybe I can make it more understated  as just text like "For example, US and UK passport have it on the first page, German has it on the last page, etc. So you might need to try different pages of your passport until it works".

nabramow:
Okay @aapeliv I made some changes!

vas2525:
@nabramow 
NFC Chip locations:
Brazil(2025) - back cover
Indonesia(2025) - back cover
Russia(2019) -  photo page

PR #7094, opened 2025-11-03
Improve backend tests

WouldYouKindly:


app/backend/src/tests/test_db.py
```diff
@@ -133,28 +131,17 @@ def strip_leading_whitespace(lines):
...
     This test will only run successfully if you have `pg_dump` installed and everything set up, which only happens if the
     test is being run within Gitlab CI where we do all that setup. So we disable it unless explicitly marked to run.
 
     Compares the database schema built up from migrations, with the
     schema built by models.py. Both scenarios are started from an
     empty database, and dumped with pg_dump. Any unexplainable
-    differences in the output are reported in unified diff format and
-    fails the test.
+    differences in the output are reported in a unified diff format and
+    fail the test.
```
aapeliv, line 144: (nit: "unified diff format" is the name of a format)

app/backend/src/tests/test_fixtures.py
```diff
@@ -210,39 +197,100 @@ def drop_database() -> None:
...
             )
         )
 
 
-def recreate_database():
+@contextmanager
+def autocommit_engine(url: str):
+    engine = create_engine(
+        url,
+        isolation_level="AUTOCOMMIT",
```
aapeliv, line 217: Why do you want an autocommit engine? Could you add a docstring here please?

If I understand correctly it's so that you can run things like `create database`.

app/backend/src/tests/test_fixtures.py
```diff
@@ -210,39 +197,100 @@ def drop_database() -> None:
...
+@pytest.fixture(scope="session")
+def postgres_engine() -> Generator[Engine]:
+    """
+    SQLAlchemy engine connected to "postgres" database.
+    """
+    dsn = config["DATABASE_CONNECTION_STRING"]
+    if not dsn.endswith("/testdb"):
+        raise RuntimeError(f"DATABASE_CONNECTION_STRING must point to /testdb, but was {dsn}")
+
+    postgres_dsn = re.sub(r"/testdb$", "/postgres", dsn)
```
aapeliv, line 232: I'm a bit allergic to regex in this kind of spot, but it's fine.

app/backend/src/tests/test_public.py
```diff
@@ -42,28 +42,29 @@ def test_GetPublicMapLayer(db):
         http_body = public.GetPublicUsers(empty_pb2.Empty())
         assert http_body.content_type == "application/json"
         data = json.loads(http_body.data)
-        print(data)
+        # Sort to ensure a deterministic order
```
aapeliv, line 45: I recall this having originally been in a fixed/non-random order. Do you know why it wasn't deterministic now?

app/backend/Makefile
```diff
@@ -32,11 +32,22 @@ upgrade:
...
 	uv run alembic upgrade +1
 
+setup-db:
+	docker compose -f ../docker-compose.test.yml up -d --wait --no-deps postgres_tests
+	# Creating postgis extension takes several seconds, and we don't want to do it on
+	# every test run.
+	docker compose -f ../docker-compose.test.yml exec postgres_tests psql -U postgres -c \
+		"CREATE EXTENSION IF NOT EXISTS postgis; \
+		 CREATE EXTENSION IF NOT EXISTS pg_trgm; \
+		 CREATE EXTENSION IF NOT EXISTS btree_gist"
```
aapeliv, line 42: Could we keep this in the python test harness somehow?

aapeliv (commented):
These are mostly questions.

If you want to go one step further, you could write a very brief doc explaining how the testing database is set up (template copying, autocommit stuff, what the command line args do), but up to you!

app/backend/src/tests/test_db.py
WouldYouKindly: ah! changed it back

app/backend/src/tests/test_fixtures.py
WouldYouKindly: Exactly that. Added a comment

app/backend/src/tests/test_public.py
WouldYouKindly: I don't to be honest. I kind of assume it's because the database is new every time, but I have no idea of what is happening exactly

app/backend/Makefile
WouldYouKindly: It is still in the fixtures for CI etc. I replicated it here as an optimisation: when running a single test, it's a bit annoying to wait 1-2 seconds for the extensions to be created.

WouldYouKindly:
> These are mostly questions.
> 
> If you want to go one step further, you could write a very brief doc explaining how the testing database is set up (template copying, autocommit stuff, what the command line args do), but up to you!

I added a comment to `template_db` fixture - hopefully should be enough for now?

aapeliv (approved):
LGTM

PR #7104, opened 2025-11-05
Split models info multiple files

WouldYouKindly:
The `models.py` file is too big for Claude, so I decided to split it. I import all models into `__init__.py`, so the rest of the code is still importing `from couchers.models import ...`. 

I also added `.pyi` generation for the protos, so that the types could be used by type checkers and IDEs.


aapeliv (approved):
LGTM if these are identical.

Is it possible to have this in git commit ignore file so that one can still easily trace model changes? If not, no worries

WouldYouKindly:
> LGTM if these are identical.
> 
> Is it possible to have this in git commit ignore file so that one can still easily trace model changes? If not, no worries

@aapeliv these should be identical, yes! I tried to be as careful as I could 😅 

Will add the commit to ignore revs file in the next PR

PR #7112, opened 2025-11-07
Fix not passing config to check_config()

WouldYouKindly:


app/backend/src/couchers/config.py
```diff
@@ -104,32 +104,28 @@
 ]
 
 
-def check_config(config: dict[str, Any]) -> None:
+def check_config(cfg: dict[str, Any]) -> None:
```
WouldYouKindly, line 107: renamed to not clash

aapeliv (approved):
lgtm

PR #7117, opened 2025-11-07
Add mypy and fix some errors

WouldYouKindly:
Adds a command `make mypy` to backend Makefile that runs mypy for a small subset of files. Also, these files are now type-annotated (with a small number of type: ignores.) At some point I will add mypy check to the CI. 

There are no code changes, only type annotations.

Also adds `make protos` that generates Protobufs.


app/backend/pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "Jinja2==3.1.6",
     "luhn==0.2.0",
     "markdown-it-py==4.0.0",
+    "mypy>=1.18.2",
```
aapeliv, line 49: Can we put it in dev reqs, I think that's a `uv` feature?

aapeliv (dismissed):
check comment

app/backend/pyproject.toml
WouldYouKindly: done

aapeliv (approved):
lgtm

PR #7139, opened 2025-11-08
Add type annotations for SQLAlchemy models

WouldYouKindly:
Adds type annotations for models, and converts them to mypy-compatible syntax, e.g.,

```python
# was 
id = Column(BigInteger, primary_key=True)

# now
id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
```

### DIFF
--- a/app/backend/src/couchers/models/auth.py
+++ b/app/backend/src/couchers/models/auth.py
@@ -1,10 +1,16 @@
-from sqlalchemy import Boolean, Column, DateTime, ForeignKey, Index, Integer, String, func, text
+from datetime import datetime
+from typing import TYPE_CHECKING, Any
+
+from sqlalchemy import Boolean, DateTime, ForeignKey, Index, Integer, String, func, text
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import Mapped, mapped_column, relationship
 from sqlalchemy.sql import expression
 
 from couchers.models.base import Base
 from couchers.utils import now
+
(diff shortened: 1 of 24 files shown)


app/backend/src/couchers/models/auth.py
```diff
@@ -23,44 +29,46 @@ class UserSession(Base):
...
     # an api key is put in the authorization header (e.g. "authorization: Bearer <token>")
     # a session cookie is set in the "couchers-sesh" cookie (e.g. "cookie: couchers-sesh=<token>")
     # when a session is created, it's fixed as one or the other for security reasons
     # for api keys to be useful, they should be long-lived and have a long expiry
-    is_api_key = Column(Boolean, nullable=False, server_default=expression.false())
+    is_api_key: Mapped[bool] = mapped_column(Boolean, server_default=expression.false())
 
     # whether it's a long-lived or short-lived session
-    long_lived = Column(Boolean, nullable=False)
+    long_lived: Mapped[bool] = mapped_column(Boolean)
```
aapeliv, line 44: The internet tells me you don't even need the right hand side here when it knows the type?

aapeliv (approved):

app/backend/src/couchers/models/auth.py
WouldYouKindly: 🤔 I'll take a look!

WouldYouKindly: You're right! I'd keep the full form though - there are very few cases when one can omit the right hand side, and it may be a bit confusing. Also, if you try to use such column in an Index later, it will give you an error :(

PR #7141, opened 2025-11-08
Use an ARM-based image to generate protos on mac

WouldYouKindly:
On my machine it's ~20 times faster, and takes around 1s now.

### DIFF
--- a/app/backend/Makefile
+++ b/app/backend/Makefile
@@ -4,12 +4,15 @@
 UV_VERSION=0.9.6
 ROOT_DIR := $(dir $(realpath $(lastword $(MAKEFILE_LIST))))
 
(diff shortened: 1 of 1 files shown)


aapeliv (changes requested):
Could you instead modify our container?


https://github.com/Couchers-org/grpc

WouldYouKindly:
> Could you instead modify our container?
> 
> https://github.com/Couchers-org/grpc

@aapeliv I've tried that, and it's not trivial unfortunately. Maybe I'll get around to that at some point, but for now I'd just like to have a way to quickly generate protos on a mac.

aapeliv:
I don't want to have this kind of platform difference (it'll cause a nightmare to debug with devs on different platforms). Let's then switch both to the other container?

Also, you can install the tools locally and run the script outside of docker if you wish, which is my preferred way.

PR #7143, opened 2025-11-09
Attempt to fix mobile notif typing

### LINKED ISSUE #7105: Typing problem on mobile devices - can't see what you're typing
Subject: Home Screen app typing problem 
Description:
I have the Couchers Home Screen app. I got a message notification, opened it, and began to respond 

Results:
I should be able to see the words as I'm typing them, but instead I could not see the words. I am using iPhone13 mini, and I believe the home screen app open using Safari. Note: this only happens when I open a message from a notification 

Backend version: develop-88a27622
Frontend version: develop-ba3add22
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.0.1 Mobile/15E148 Safari/604.1
Screen resolution: 375x610
Page: https://couchers.org/search#10/40.7127/-74.006
User: [@bknicholas](https://couchers.org/user/bknicholas) (12838)

nabramow:
Trying to fix #7105

Kinda going in blind here, it's not really possible to test until it's on stage as notification urls redirect to stage away from local. Also I don't have an Android. So we'll need to deploy and then see.


jesseallhands (changes requested):
Tried testing using messages. When I land on the message page instead of being taken to the latest message (and where I can type the reply) I land in the middle of the message thread. See screenshot below.

Also, the send button isn't working for me. I tap it and nothing happens. 

Screenshot_20251109_130303_Firefox


I'm using Firefox Mobile on android 16, Firefox version 145.0 (Build #2016124103)

nabramow:
Hey @jesse please read the description.On Nov 9, 2025, at 12:10 PM, Jesse ***@***.***> wrote:﻿
@jesseallhands requested changes on this pull request.

Tried testing using messages. When I land on the message page instead of being taken to the latest message (and where I can type the reply) I land in the middle of the message thread. See screenshot below.
Also, the send button isn't working for me. I tap it and nothing happens.
Screenshot_20251109_130303_Firefox.jpg (view on web)
I'm using Firefox Mobile on android 16, Firefox version 145.0 (Build #2016124103)

jesseallhands:
On Firefox and chrome the primary problem in this issue seems to be resolved.

The problem of landing on messages mid-thread when you navigate to a message from the messages page is also resolved on Firefox but on chrome that issues persists. Not sure if that problem was introduced in this issue or if it should be resolved in a different issue but it is problematic from a ui/ux perspective. Screenshot below of where I landed on chrome after clicking a message on the messages page:

Screenshot_20251109_134406_Chrome

nabramow:
@jesseallhands So you can't really test this fix here as I wrote in the description. This is because the notifs take you to a different environment than the changes are in, so it's only once it's merged to stage, etc. that the environments match.

So I guess with this I just want to make sure messages are not worse on Android. I imagine most of the issues you raised here are already in prod/next, but please let me know if not as then a bug would be introduced by merging this.

I guess most important thing is anything different/worse than on prod/next. If not we can merge and then see if it fixes it.

nabramow:
Closing as I think we should just do a big bug fix search before the app releases as it will be easier to test there.

PR #7146, opened 2025-11-09
More type annotations

WouldYouKindly:


app/backend/src/couchers/jobs/handlers.py
```diff
@@ -146,7 +146,7 @@
 refresh_materialized_views_rapid.SCHEDULE = timedelta(seconds=30)
 
 
-def send_email(payload):
+def send_email(payload: "jobs_pb2.SendEmailPayload") -> None:
```
aapeliv: Why not import the real type?

app/backend/src/couchers/helpers/geoip.py
```diff
@@ -10,31 +10,38 @@
...
     try:
         with geoip2.database.Reader(config["GEOLITE2_ASN_MMDB_FILE_LOCATION"]) as reader:
             response = reader.asn(ip_address)
-            return response.autonomous_system_number, response.autonomous_system_organization, response.network
+            asn_number = response.autonomous_system_number
+            asn_org = response.autonomous_system_organization
+            network_str = str(response.network)
+            if asn_number is not None and asn_org is not None:
+                return asn_number, asn_org, network_str
+            return None
```
aapeliv: Why change the logic?

app/backend/src/couchers/migrations/env.py
```diff
@@ -1,49 +1,54 @@
...
+config: Config = context.config
 
 config.set_main_option("sqlalchemy.url", couchers_config["DATABASE_CONNECTION_STRING"])
 
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if config.get_main_option("dont_mess_up_logging", "dont_care") == "dont_care":
+    if not config.config_file_name:
+        raise RuntimeError(config.config_file_name)
```
aapeliv, line 23: I’m not sure what this does exactly?

app/backend/src/couchers/notifications/render.py
```diff
@@ -926,3 +934,5 @@ def render_notification(user, notification) -> RenderedNotification:
         )
     else:
         raise NotImplementedError(f"Unknown topic-action: {notification.topic}:{notification.action}")
```
aapeliv: Let's just move this out of the `else` and to the end of the function, then we don't need to worry about those other `else`s to catch unhandled ones per topic

app/backend/src/couchers/rate_limits/definitions.py
```diff
@@ -22,24 +23,25 @@
 from couchers.utils import now
 
 if TYPE_CHECKING:
```
aapeliv: now this if can go?

app/backend/src/couchers/rate_limits/definitions.py
```diff
@@ -22,24 +23,25 @@
...
 
 
 RATE_LIMIT_INTERVAL = timedelta(hours=24)
 RATE_LIMIT_INTERVAL_STRING = "24 hours"
 
 
-def _get_user_host_requests_in_past_time_interval(session, user_id) -> list[dict]:
-    return (
-        session.execute(
+def _get_user_host_requests_in_past_time_interval(session: Session, user_id: int) -> list[dict[str, Any]]:
```
aapeliv: I have a feeling the type was wrong here? I don't think iterating through and applying `dict` does the same thing.

It would be my preference if the function body here could be as before (it's also much cleaner in my opinion). I think the return type should then be a sqllachemy query or similar (you can actually even remove the `session.execute()` and `mappings().all()` to just have the query logic.

aapeliv (commented):
I think adding typing is great, thanks for doing this.

Is there a problem with returning with just the `return` keyword? I see you changed that in a few places.

app/backend/src/couchers/jobs/handlers.py
WouldYouKindly: Claude... fixed.

app/backend/src/couchers/helpers/geoip.py
WouldYouKindly: I now changed it to raise an exception if the values are None. The calling code assumes that the fields are not None, so we should throw an error here.

app/backend/src/couchers/migrations/env.py
WouldYouKindly: If for some reason file name is not set, we'll raise an error. It's a sanity check and also makes mypy happy.

app/backend/src/couchers/notifications/render.py
WouldYouKindly: done

app/backend/src/couchers/rate_limits/definitions.py
WouldYouKindly: Done, thanks.

app/backend/src/couchers/rate_limits/definitions.py
WouldYouKindly: Reverted the change, and annotated with the appropriate `Sequence[RowMapping]`. Don't want to change the logic here for now.

WouldYouKindly:
> I think adding typing is great, thanks for doing this.
> 
> Is there a problem with returning with just the `return` keyword? I see you changed that in a few places.

So it turns out that PEP8 recommends `return None` if there are other return statements with a value in the function: https://peps.python.org/pep-0008/

```
Be consistent in return statements. Either all return statements in a function should return an expression, 
or none of them should. If any return statement returns an expression, any return statements where no value 
is returned should explicitly state this as return None, and an explicit return statement should be present 
at the end of the function (if reachable)
```

That's why mypy complains about it in strict mode. If you have a strong opinion about it we can disable this specific rule though.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -844,7 +844,7 @@ def update_badge(badge_id: str, members: list[int]):
 update_badges.SCHEDULE = timedelta(minutes=15)
 
 
-def finalize_strong_verification(payload):
+def finalize_strong_verification(payload: "jobs_pb2.FinalizeStrongVerificationPayload") -> None:
```
aapeliv, line 847: Same as earlier here

aapeliv (approved):
lgtm

PR #7164, opened 2025-11-10
More type annotations

WouldYouKindly:
I'm close to being done :) The only ones left are interceptors and servicers.

### DIFF
--- a/app/backend/src/couchers/db.py
+++ b/app/backend/src/couchers/db.py
@@ -6,6 +6,7 @@ from collections.abc import Generator, Sequence
 from contextlib import contextmanager
 from os import getpid
 from threading import get_ident
+from typing import cast
 
 from alembic import command
 from alembic.config import Config
@@ -18,6 +19,7 @@ from sqlalchemy.sql import and_, func, literal, or_
 
 from couchers.config import config
 from couchers.constants import SERVER_THREADS, WORKER_THREADS
+from couchers.context import CouchersContext
 from couchers.models import (
     Cluster,
     ClusterRole,
@@ -73,7 +75,7 @@ def session_scope() -> Generator[Session]:
                         filename_line = f"{frame.filename}:{frame.lineno}"
                     except Exception as e:
                         filename_line = "{unknown file}"
-                    backend_pid = session.execute(text("SELECT pg_backend_pid();")).scalar()
+                    backend_pid = session.execute(text("SELECT pg_backend_pid();")).scalar_one()
                     logger.debug(f"SScope: got {backend_pid=} at {filename_line}")
                     rollspan.set_attribute("db.backend_pid", backend_pid)
                     rollspan.set_attribute("db.filename_line", filename_line)
@@ -109,7 +111,7 @@ def worker_repeatable_read_session_scope() -> Generator[Session]:
                         filename_line = f"{frame.filename}:{frame.lineno}"
                     except Exception as e:
                         filename_line = "{unknown file}"
-                    backend_pid = session.execute(text("SELECT pg_backend_pid();")).scalar()
+                    backend_pid = session.execute(text("SELECT pg_backend_pid();")).scalar_one()
                     logger.debug(f"SScope (worker): got {backend_pid=} at {filename_line}")
                     rollspan.set_attribute("db.backend_pid", backend_pid)
                     rollspan.set_attribute("db.filename_line", filename_line)
@@ -134,7 +136,7 @@ def db_post_fork() -> None:
     _get_base_engine().dispose(close=False)
 
 
-def are_friends(session: Session, context, other_user: int) -> bool:
+def are_friends(session: Session, context: CouchersContext, other_user: int) -> bool:
     return (
         session.execute(
             select(FriendRelationship)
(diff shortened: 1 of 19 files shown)


app/backend/src/couchers/db.py
```diff
@@ -201,66 +203,59 @@ def get_node_parents_recursively(session: Session, node_id: int) -> Sequence[Row
...
-                .where(ClusterSubscription.role == ClusterRole.admin)
-                .where(ClusterSubscription.user_id == user_id)
-                .join(Cluster, Cluster.id == ClusterSubscription.cluster_id)
-                .where(Cluster.is_official_cluster)
-                .where(Cluster.parent_node_id == subquery.c.id)
-            ).exists()
-        )
-    ).scalar_one()
-
-    return _can_moderate_any_cluster(
```
aapeliv, line 236: Lol i didn't realize there was an extra return here. I guess it came from some prior optimization.

app/backend/src/couchers/interceptors.py
```diff
@@ -67,7 +72,9 @@ def _try_get_and_update_user_details(token, is_api_key, ip_address, user_agent):
         if not result:
             return None
         else:
-            user, user_session, user_activity = result
+            user: User = result[0]
+            user_session: UserSession = result[1]
+            user_activity: UserActivity = result[2]
```
aapeliv: Is there not a nicer way to do this?

app/backend/src/couchers/materialized_views.py
```diff
@@ -182,8 +191,17 @@ def make_lite_users_selectable(create=False):
 class LiteUser(Base):
     __table__ = lite_users
 
+    # to allow type annotations without affecting SQLAlchemy
+    __allow_unmapped__ = True
 
-def make_clustered_users_selectable(create=False):
+    # A subset enough to make mypy happy. Taken from "make_lite_users_selectable".
```
aapeliv, line 198: Could you add the whole set of columns if you're going this direction, please? Should be as defined above

app/backend/src/couchers/materialized_views.py
```diff
@@ -304,7 +322,7 @@ class UserResponseRate(Base):
     __table__ = user_response_rates
 
 
-def refresh_materialized_views(payload: empty_pb2.Empty):
+def refresh_materialized_views(payload: empty_pb2.Empty) -> None:
```
aapeliv, line 334: What's the diff between `-> None` and `-> NoReturn`?

app/backend/src/couchers/tasks.py
```diff
@@ -55,14 +63,20 @@ def send_signup_email(session, flow):
...
-def send_email_changed_confirmation_to_new_email(session, user):
+def send_email_changed_confirmation_to_new_email(session: Session, user: User) -> None:
     """
-    Send an email to user's new email address requesting confirmation of email change
+    Send an email to the user's new email address requesting confirmation of email change
     """
     logger.info(
-        f"Sending email changed (confirmation) email to {user=}'s new email address, (old email: {user.email}, new email: {user.new_email=})"
+        f"Sending email changed (confirmation) email to {user=}'s new email address, "
+        f"(old email: {user.email}, new email: {user.new_email=})"
```
aapeliv, line 72: What happened here?

app/backend/src/couchers/materialized_views.py
WouldYouKindly: NoReturn means that the function will always raise an exception (e.g., abort())

app/backend/src/couchers/tasks.py
WouldYouKindly: Technically a user can have new_email or new_email_token empty, so mypy complained, and I had to add sanity checks.

app/backend/src/couchers/interceptors.py
WouldYouKindly: There is! Fixed.

app/backend/src/couchers/materialized_views.py
WouldYouKindly: done

app/backend/src/couchers/materialized_views.py
aapeliv: ok good to know!

app/backend/src/couchers/tasks.py
aapeliv: no but i don't see a difference here except the string got cut?

WouldYouKindly: yeah I just split it into two lines, because it didn't fit on my screen :D

app/backend/src/couchers/interceptors.py
```diff
@@ -68,6 +74,9 @@ def _try_get_and_update_user_details(token, is_api_key, ip_address, user_agent):
             return None
         else:
             user, user_session, user_activity = result
+            # user: User = result[0]
+            # user_session: UserSession = result[1]
+            # user_activity: UserActivity | None = result[2]
```
aapeliv: Is this comment meant to be here?

WouldYouKindly: Nope. Removed it, thanks!

aapeliv (approved):

PR #7167, opened 2025-11-10
Make sure our backend apps receive signals

WouldYouKindly:
Make sure that our python process is not PID 1 inside the container, so that signals are properly handled.

See https://hynek.me/articles/docker-signals

Also updates uv to the latest minor version.


aapeliv (approved):
lgtm, thanks for fixing!

PR #7179, opened 2025-11-12
Add some type annotations to test caching in CI

WouldYouKindly:


app/backend/src/couchers/tracing.py
```diff
@@ -19,8 +19,8 @@ def setup_tracing() -> None:
...
         SQLAlchemyInstrumentor().instrument(engine=_get_base_engine(), enable_commenter=True, commenter_options={})
 
-        trace.set_tracer_provider(TracerProvider(resource=Resource(attributes={"service.name": "backend"})))
-
-        trace.get_tracer_provider().add_span_processor(
+        provider = TracerProvider(resource=Resource(attributes={"service.name": "backend"}))
+        provider.add_span_processor(
             BatchSpanProcessor(OTLPSpanExporter(endpoint=config["OPENTELEMETRY_ENDPOINT"], insecure=True))
         )
+        trace.set_tracer_provider(provider)
```
WouldYouKindly, line 26: get_tracer_provider returns an incorrect type, so we use the instance we created to avoid errors

PR #7181, opened 2025-11-12
Differentiate merch from donations in backend

nabramow:
- Added migration to add `invoice_type` column to the `invoices` table
- Created `InvoiceType` enum (do we want an enum or a string?). Enums are annoying to change but safer since it's records concerning money?
- Adds check in webhook logic to determine if `payment_succeeded` is from donation or merch order
- Updates `invoice_type` accordingly

**Context**:

In our backend we call `stripe.checkout.Session.create` to create a stripe checkout session for a donation initiation. That seems to correlate with the [checkout session object in the stripe docs ](https://docs.stripe.com/api/checkout/sessions/object?api-version=2025-10-29.preview)(using the deprecated events object oops) which has a `metadata` object you can add key/values to.

For the merch orders, I checked what data is in the event by logging into stripe, looking in webhooks and checking the event that was sent during our tests. The `site_url` seemed to be the most reliable so I went with that but it's not totally ideal, I'm not sure if we could add metadata to that somehow?

**Test**:
- You can test the migration here, but I think we need to merge it to stage to be able to actually test with stripe using the test credit card data. We should do that before merging to prod to make sure the donations and merch do actually get separated.


app/backend/src/couchers/servicers/donations.py
```diff
@@ -129,28 +130,50 @@ def Webhook(self, request, context, session):
...
+            elif metadata.get("site_url") == "https://shop.couchershq.org":
+                # This is from WooCommerce merch shop
+                invoice_type = InvoiceType.merch
+            else:
+                # Unknown payment source - ignore it
+                logger.warning(f"Stripe payment intent {payment_intent_id} has unknown metadata, ignoring")
+                return httpbody_pb2.HttpBody(
+                    content_type="application/json",
+                    data=json.dumps({"success": True, "ignored": True}).encode("ascii"),
+                )
```
WouldYouKindly: Don't we still want to save it to the database though? Or is it enough for us to track in Stripe's UI? Maybe we should send an error to Sentry, because no one looks at the warning logs :)

nabramow: Hmm yeah good point I should throw an error here and sentry makes sense!

app/backend/src/couchers/servicers/donations.py
```diff
@@ -129,28 +131,51 @@ def Webhook(self, request, context, session):
...
+                # This is from WooCommerce merch shop
+                invoice_type = InvoiceType.merch
+            else:
+                # Unknown payment source - this should never happen
+                sentry_sdk.set_tag("stripe_payment_intent_id", payment_intent_id)
+                sentry_sdk.set_tag("stripe_customer_id", customer_id)
+                sentry_sdk.set_context("stripe_metadata", metadata)
+                error_msg = f"Unable to determine invoice_type for Stripe payment intent {payment_intent_id}. Expected metadata.type='donation' or metadata.site_url='https://shop.couchershq.org', but got: {metadata}"
+                logger.error(error_msg)
+                raise ValueError(error_msg)
```
WouldYouKindly: Can we move it into a function?

```python
def infer_invoice_type(data_object: dict[str, Any]) -> InvoiceType:
   ...
```

Alternatively, the function can accept `metadata`. Then we'd have to set sentry tags outside, unconditionnaly (shouldn't be a problem, since event will be sent only in case of an error).

app/backend/src/couchers/servicers/donations.py
```diff
@@ -129,28 +131,51 @@ def Webhook(self, request, context, session):
...
-            user.has_donated = True
+            payment_intent_id = data_object["payment_intent"]
+
+            # Get invoice type from charge metadata
+            metadata = data_object.get("metadata", {})
+
+            # Check if this is from Couchers (type=donation) or WooCommerce (has site_url=shop)
+            if metadata.get("type") == "donation":
+                invoice_type = InvoiceType.donation
+            elif metadata.get("site_url") == "https://shop.couchershq.org":
```
aapeliv: Let's put this URL in an environment variable

app/backend/src/couchers/models/donations.py
```diff
@@ -16,6 +16,11 @@ class DonationType(enum.Enum):
     recurring = enum.auto()
 
 
+class InvoiceType(enum.Enum):
+    donation = enum.auto()
+    merch = enum.auto()
```
aapeliv: Let's call these `on_platform` and `external_shop`

aapeliv (changes requested):

app/backend/src/couchers/servicers/donations.py
```diff
@@ -129,28 +131,51 @@ def Webhook(self, request, context, session):
...
+                sentry_sdk.set_tag("stripe_payment_intent_id", payment_intent_id)
+                sentry_sdk.set_tag("stripe_customer_id", customer_id)
+                sentry_sdk.set_context("stripe_metadata", metadata)
+                error_msg = f"Unable to determine invoice_type for Stripe payment intent {payment_intent_id}. Expected metadata.type='donation' or metadata.site_url='https://shop.couchershq.org', but got: {metadata}"
+                logger.error(error_msg)
+                raise ValueError(error_msg)
+
+            # Only mark as donated if it's a donation (not merch)
+            if invoice_type == InvoiceType.donation:
+                user.has_donated = True
```
aapeliv: Let's also mark as donor with merch

aapeliv (approved):

app/backend/src/couchers/models/donations.py
nabramow: @aapeliv Hmm but I'm thinking about the actual price for the merch no? I know we will have the additional donations via merch, but that's not this.

PR #7185, opened 2025-11-13
Put args at the end of the web Dockerfiles so as to not bust the build cache

WouldYouKindly:
We change these arguments with every commit. When ARG is changed, every command after it has to be re-executed, because Docker assumes that its result may depend on the value of ARG. As far as I understand though, these arguments do not affect the build - only the runtime. By putting them at the end, we allow Docker to re-use the cached layers from previous builds.

### DIFF
--- a/app/web/Dockerfile
+++ b/app/web/Dockerfile
@@ -6,6 +6,19 @@ WORKDIR /app
 # disable next.js telemetry
 ENV NEXT_TELEMETRY_DISABLED=1
 
+COPY package.json yarn.lock ./
+# https://github.com/yarnpkg/yarn/issues/8242
+RUN yarn config set network-timeout 300000 && yarn install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


app/web/Dockerfile
```diff
@@ -6,6 +6,19 @@ WORKDIR /app
...
+
+COPY . .
+
+RUN cp .env.$environment env && \
+    rm .env.* && \
+    mv env .env.local
+
+RUN yarn build
+
+# These are at the end so as to not invalidate Docker build cache
```
WouldYouKindly, line 21: This looks a bit weird, but all I did was to move args/envs to the end, and merge a couple of RUNs

nabramow (approved):
I cannot comment so much on best practices here, but I can say it builds fine and the app runs so nothing is broken! ;)

PR #7187, opened 2025-11-13
Shrink backend image by not including uv

WouldYouKindly:
- Don't include uv into the final image
- Only include specific files and directories when doing `COPY` in the Dockerfile, to avoid unnecessary rebuilds when e.g., a Makefile or a Readme is changed


aapeliv (approved):
lgtm

PR #7188, opened 2025-11-13
Speed up web build

WouldYouKindly:
- Save web images locally to avoid pulling them a second later when we do `docker create`
- Don't add unnecessary files to the web image


aapeliv (approved):
lgtm

PR #7189, opened 2025-11-13
Make media image smaller

WouldYouKindly:


app/generate_protos.sh
```diff
@@ -6,7 +6,7 @@ rm -rf proto/gen
 mkdir -p proto/gen/python/proto
 mkdir -p proto/gen/ts/proto
 mkdir -p backend/src/proto/
-mkdir -p media/src/proto/
+mkdir -p media/src/media/proto/
```
WouldYouKindly, line 9: Put protos inside media package so that we can do `from media.proto import ...`

aapeliv (approved):
lgtm

PR #7192, opened 2025-11-14
Type-annotate interceptors.py

WouldYouKindly:
There are so many type ignores because we pass Session as a third argument, which does not conform to generated method signatures. Maybe we can find a better way in the future.


aapeliv (dismissed):
not clear to me why args to `_store_log` need the ignore?

WouldYouKindly:
> not clear to me why args to `_store_log` need the ignore?

Good point. Removed all ignores (but had to do some casts)

aapeliv (approved):

PR #7201, opened 2025-11-15
UMS

aapeliv:


app/backend/src/couchers/models/moderation.py
```diff
@@ -0,0 +1,189 @@
...
+    __tablename__ = "moderation_log"
+
+    id: Mapped[int] = mapped_column(
+        BigInteger, moderation_seq, primary_key=True, server_default=moderation_seq.next_value()
+    )
+    moderation_state_id: Mapped[int] = mapped_column(ForeignKey("moderation_states.id"), index=True)
+
+    time: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now())
+    action: Mapped[ModerationAction] = mapped_column(Enum(ModerationAction))
+    moderator_user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
```
WouldYouKindly, line 163: moderator_id?

app/backend/src/couchers/models/moderation.py
```diff
@@ -0,0 +1,189 @@
...
+    History of moderation actions
+
+    This table provides a complete audit trail of all moderation actions taken,
+    including who performed the action and what changed.
+    """
+
+    __tablename__ = "moderation_log"
+
+    id: Mapped[int] = mapped_column(
+        BigInteger, moderation_seq, primary_key=True, server_default=moderation_seq.next_value()
```
WouldYouKindly, line 157: why do we need to have the same sequence for all tables?

aapeliv: we don't need to. but it's nice. in these kinds of systems you end up with a lot of IDs flying around of different types, it can get very messy very quick; especially if IDs are often created in parallel (so that they are similar). this allows us to set up a system later which just looks up which table it belongs to and lets us just jump over there. and there won't be any confusion because a given ID will surely point to at most one moderatable object

app/backend/src/couchers/models/moderation.py
aapeliv: We use `_user_id` suffic pretty commonly for user id columns. I'm not sure it's completely universal.

PR #7202, opened 2025-11-15
Hotfix to get docker working over weekend

nabramow:
This is not a permanent fix, but it unblocks things until @WouldYouKindly has a chance to fix the docker stuff.


aapeliv (approved):
full trust

PR #7205, opened 2025-11-16
Type-annotate a bunch of servicers

WouldYouKindly:


app/backend/src/couchers/servicers/api.py
```diff
@@ -40,13 +43,14 @@
...
 from couchers.sql import couchers_select as select
-from couchers.sql import is_valid_user_id, is_valid_username
 from couchers.utils import (
     Duration_from_timedelta,
     Timestamp_from_datetime,
     create_coordinate,
     get_coordinates,
     is_valid_name,
+    is_valid_user_id,
+    is_valid_username,
```
WouldYouKindly, line 53: Were imported from the wrong location

app/backend/src/couchers/servicers/api.py
```diff
@@ -697,15 +715,17 @@ def SendFriendRequest(self, request, context, session):
             session,
             user_id=friend_relationship.to_user_id,
             topic_action="friend_request:create",
-            key=friend_relationship.from_user_id,
+            key=str(friend_relationship.from_user_id),
```
WouldYouKindly, line 718: There's currently an implicit conversion there, but we have to be explicit for mypy's happiness

app/backend/src/couchers/servicers/pages.py
```diff
@@ -52,9 +53,9 @@ def _can_moderate_page(session: Session, page: Page, user_id: int) -> bool:
     return can_moderate_node(session, user_id, page.parent_node_id)
 
 
-def page_to_pb(session: Session, page: Page, context) -> pages_pb2.Page:
-    first_version = page.versions[0]
-    current_version = page.versions[-1]
+def page_to_pb(session: Session, page: Page, context: CouchersContext) -> pages_pb2.Page:
+    first_version = page.versions[0]  # type: ignore[attr-defined]
```
WouldYouKindly, line 57: `backref` breaks type annotations, I plan to convert to `back_populates` in a future PR

app/backend/src/couchers/servicers/public.py
```diff
@@ -90,16 +110,7 @@ def _get_volunteers(session):
...
-    def format_volunteer(volunteer, lite_user):
-        if volunteer.link_type:
-            link_type = volunteer.link_type
-            link_text = volunteer.link_text
-            link_url = volunteer.link_url
-        else:
-            link_type = "couchers"
-            link_text = f"@{lite_user.username}"
-            link_url = urls.user_link(username=lite_user.username)
-
```
WouldYouKindly, line 102: unused

app/backend/src/couchers/models/events.py
```diff
@@ -166,19 +166,21 @@ def coordinates(self) -> tuple[float, float] | None:
...
-        return self.during.upper
+    def end_time(self) -> datetime:
+        return not_none(self.during.upper)
 
-    @end_time.expression
-    def end_time(cls) -> Any:  # noqa: ARG003,D102
+    @end_time.inplace.expression
+    @classmethod
+    def _end_time(cls) -> ColumnElement[datetime]:
         return func.upper(cls.during)
```
aapeliv, line 184: what's going on here?

app/backend/src/couchers/servicers/api.py
```diff
@@ -351,12 +361,12 @@ def UpdateProfile(self, request, context, session):
         if request.hosting_status != api_pb2.HOSTING_STATUS_UNSPECIFIED:
             if user.do_not_email and request.hosting_status != api_pb2.HOSTING_STATUS_CANT_HOST:
                 context.abort_with_error_code(grpc.StatusCode.FAILED_PRECONDITION, "do_not_email_cannot_host")
-            user.hosting_status = hostingstatus2sql[request.hosting_status]
+            user.hosting_status = hostingstatus2sql[request.hosting_status]  # type: ignore[assignment]
```
aapeliv, line 364: what's this one for?

aapeliv (commented):
what's the benefit of type annotating all RPCs? do we need to be manually updating those? what happens when we change the proto definition? does it get flagged by mypy or we need to now update more things to build a new endpoint?

app/backend/src/couchers/models/events.py
WouldYouKindly: It's a way to make hybrid_property intelligible to mypy (on an instance-level it returns a datetime, on a class level - a Column). See https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html#defining-expression-behavior-distinct-from-attribute-behavior

app/backend/src/couchers/servicers/api.py
WouldYouKindly: mypy cannot understand the types here for some reason. I tried to make it work, but after spending some time, decided it wasn't worth it for now. I think this is because protobuf's enum are not proper python enums...

WouldYouKindly:
> what's the benefit of type annotating all RPCs? do we need to be manually updating those? what happens when we change the proto definition? does it get flagged by mypy or we need to now update more things to build a new endpoint?

@aapeliv good question. The main benefit is that we have most of our business logic inside servicers, so having types there will help catch type errors and also with autocompletion. 

Now, the types of the servicers methods are not checked, because `protobuf` does not generate type definitions for the interfaces yet. So in theory nothing is stopping you from specifying the wrong type. But you still have to import the type from generated protobuf code, so the chance to make a mistake is not huge. And then you probably modify logic more often than you write a new handler, and having types helps with that a lot.

(Also, `protoc` will hopefully at some point generate types for everything!)

> do we need to be manually updating those? what happens when we change the proto definition?

Unless you change the types name, you don't need to manually update anything. But then you'll get an import error, or a type error if you e.g., delete a field that is used inside a servicer, or change the field's type.

aapeliv:
> you probably modify logic more often than you write a new handler

That's generally not the case.

Is there a way to generate code to generate the handler types automatically in some other code that doesn't need human-intervention?

I think it's a fair bit of extra effort to add the types. Due to formatting it also takes up a bunch more lines of code. And while LOC is not any meaningful measure of anything, it does mean it's more verbose and less concise and readable.

app/backend/src/couchers/servicers/search.py
```diff
@@ -100,7 +105,7 @@ def _build_tsv(A, B=None, C=None, D=None):
     return tsv
 
 
-def _build_doc(A, B=None, C=None, D=None):
+def _build_doc(A: list[str], B: list[str] | None = None, C: list[str] | None = None, D: list[str] | None = None) -> str:
```
aapeliv: These are not lists of strings, I believe

WouldYouKindly:
> Is there a way to generate code to generate the handler types automatically in some other code that doesn't need human-intervention?

The best we can do is to add annotations to the generated interfaces. Then you would be able to go to the interface method from your IDE and copy the types. Not sure how easy that is tbh, but might give it a try. 

> I think it's a fair bit of extra effort to add the types. Due to formatting it also takes up a bunch more lines of code. And while LOC is not any meaningful measure of anything, it does mean it's more verbose and less concise and readable.

I agree it's more verbose! But it's true of type annotations everywhere. It's a trade-off we have to make - personally I feel types are worth it almost always.

Let's take a random method as an example and see how much effort it takes: [ListEventOrganizers](https://github.com/Couchers-org/couchers/blob/91ca3becdea088b6bca4b9ba807aa7dbe0bc1a23/app/backend/src/couchers/servicers/events.py#L873).

`context` and `session` are the same across all servicers - so it's not different from any other function you'd want to annotate. You can copy these from the neighbouring methods too. 

`request` is a bit trickier. You have 2 options here. 
- Annotate it with `Any` if you're feeling lazy.  This way you still get the benefits of typing `context` and `session` (autocomplete etc). 
- Type the name of the file you're in (`events`), followed by `_pb2`. Then start typing the method name, and most likely you'll get an autocomplete suggestion. Then you get autocomplete for all request's fields, and you'll get an error if a field's type changes. 

Since we have most of our business logic in servicers, if we don't type them, we'll lost a lot of type safety. Take this [line](https://github.com/Couchers-org/couchers/blob/91ca3becdea088b6bca4b9ba807aa7dbe0bc1a23/app/backend/src/couchers/servicers/events.py#L876).  What does it matter if `_get_event_and_occurrence_one_or_none` is typed, if we call it with an untyped `request.event_id` that bypasses type checking?

WouldYouKindly:
@aapeliv I found a grpc plugin that generates types for servicers! (https://github.com/nipunn1313/mypy-protobuf). So after adding a new protobuf and running `generate_proto.sh` you'll have a new abstract class/method with type annotations, and you'll be able to copy those.

aapeliv:
let me know when this is ready for a re-review, or if it's no longer needed given the mypy change!

WouldYouKindly:
Closing in favor of https://github.com/Couchers-org/couchers/pull/7253

PR #7213, opened 2025-11-16
Implement ghost users

### LINKED ISSUE #4297: Ghost profile issues

### LINKED ISSUE #7079: Common issue: content disappears...
...when the author is banned/deleted.

aapeliv:

Closes #4297, closes #7079.

Continues from #6633.

Makes it so that instead of endless spinners, it shows a "ghost user" profile.

**Backend checklist**
- [ ] All tests pass

**Web frontend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -7,6 +7,7 @@ from sqlalchemy.sql import and_, delete, distinct, func, intersect, or_, union
 
 from couchers import urls
 from couchers.config import config
+from couchers.constants import GHOST_USER_DISPLAY_NAME, GHOST_USERNAME
 from couchers.crypto import b64encode, generate_hash_signature, random_hex
 from couchers.helpers.strong_verification import get_strong_verification_fields
 from couchers.materialized_views import LiteUser, UserResponseRate
@@ -39,6 +40,7 @@ from couchers.proto import api_pb2, api_pb2_grpc, media_pb2, notification_data_p
 from couchers.rate_limits.check import process_rate_limits_and_check_abort
 from couchers.rate_limits.definitions import RATE_LIMIT_INTERVAL_STRING
 from couchers.resources import get_badge_dict, language_is_allowed, region_is_allowed
+from couchers.servicers.blocking import is_not_visible
 from couchers.sql import couchers_select as select
 from couchers.sql import is_valid_user_id, is_valid_username
 from couchers.utils import (
@@ -217,9 +219,7 @@ class API(api_pb2_grpc.APIServicer):
         )
 
     def GetUser(self, request, context, session):
-        user = session.execute(
-            select(User).where_users_visible(context).where_username_or_id(request.user)
-        ).scalar_one_or_none()
+        user = session.execute(select(User).where_username_or_id(request.user)).scalar_one_or_none()
 
         if not user:
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "user_not_found")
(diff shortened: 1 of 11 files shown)


nabramow:
This causes an error on the frontend that totally takes down the web app. To replicate, run app locally, block one of your connections that you have a message with (I blocked the local Itsi account), then click through to go to the now blocked profile and I get this error screen:

```
## Error Type
Runtime RangeError

## Error Message
Invalid time zone specified: 


    at Date.toLocaleString (<anonymous>:null:null)
    at RemainingAboutLabels (features/profile/view/userLabels.tsx:274:23)

## Code Frame
  272 |       <LabelAndText
  273 |         label={t("profile:heading.local_time")}
> 274 |         text={dayjs().tz(user.timezone).format("LT")}
      |                       ^
  275 |       />
  276 |     </>
  277 |   );
```

When I unblock the user and reload the profile, it works again without this error.

aapeliv:
@nabramow fixed. frontend loads now and can click around after banning Itsi

nabramow (approved):
Okay this works well enough now! I have some thoughts of how we want to display things in future iterations, for example if someone just has a ton of references from "Ghost", it throws an error when they click "My Home", "References" tabs, etc, but I think this is already an improvement.

I think as a next step might be good to hide a lot of the data on the profile of a ghost, so don't show friend request button, don't show tabs, don't show profile sections and only show the message about the ghost profile. We could do some conditional rendering on the frontend for that.

aapeliv:
> I think as a next step might be good to hide a lot of the data on the profile of a ghost, so don't show friend request button, don't show tabs, don't show profile sections and only show the message about the ghost profile. We could do some conditional rendering on the frontend for that.

We should probably have an `is_ghost` attribute in the User proto so that we can hide most stuff and give the user a hint that this is actual a legit thing.

app/backend/src/couchers/servicers/api.py
```diff
@@ -904,9 +910,25 @@ def get_num_references(session, user_ids):
     )
 
 
-def user_model_to_pb(db_user, session, context):
+def user_model_to_pb(db_user, session, context, *, is_admin_see_ghosts=False, is_get_user_return_ghosts=False):
```
WouldYouKindly, line 913: Not sure about these names. Maybe `see_ghosts` and `return_ghosts`?

app/backend/src/couchers/servicers/api.py
```diff
@@ -904,9 +910,25 @@ def get_num_references(session, user_ids):
...
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
+
+    if not is_admin_see_ghosts and is_not_visible(session, context.user_id, db_user.id):
+        # User is not visible (deleted, banned, or blocked)
+        if is_get_user_return_ghosts:
```
WouldYouKindly, line 919: Why would we want to not return a ghost user at all?

aapeliv: This gets called in bg workers. If a user deletes their account and we were about to notify them or similar, it's good to know of such issues by throwing. It's not necessary to throw but I think it can catch some ugly bugs at some point!

PR #7229, opened 2025-11-18
Fix passing env vars to yarn build

WouldYouKindly:
NEXT_ env variables are used by Next.js during the build. They are embedded into the frontend bundle.

This is mostly fine, since if frontend is changed, and we trigger a build job, we probably want to rebiuld :) The only case when we'd like to avoid a rebuild is a build for `develop` when we start all builds, even if stuff hasn't changed. Ideally we'd use a cached image there. 

Thoughts: maybe we don't really need these variables on the web site? I guess we can have a separate static file with these values - it would be easy to put that into `.next/static` as a separate RUN stage in Dockerfile. The only issue is SENTRY_RELEASE, but OTOH we should proxy requests to Sentry because of the adblockers - and then we can inject sentry release on the backend.

### DIFF
--- a/app/web/Dockerfile
+++ b/app/web/Dockerfile
@@ -16,9 +16,7 @@ RUN cp .env.$environment env && \
     rm .env.* && \
     mv env .env.local
 
-RUN yarn build
-
-# These are at the end so as to not invalidate Docker build cache
+# These are used during yarn build by Next.js
 ARG version
 ENV NEXT_PUBLIC_VERSION=$version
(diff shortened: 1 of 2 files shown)


WouldYouKindly:
@aapeliv is there a reason we build everything after a merge to develop, even if there were no changes? For frontend this forces the clients to download the new bundle

aapeliv (approved):
lgtm, thanks for fixing!

PR #7235, opened 2025-11-19
Add a post-processing script for python grpc type stubs

WouldYouKindly:
I found a plugin for protoc that can generate type annotations for servicers classes. However, the code it generates leads to enourmous memory consumption when running mypy over it. This PR adds a post-processing script that removes the problematic parts from the generated `.pyi` files. Without these problematic parts `mypy` works just fine.

This PR is dependant on https://github.com/Couchers-org/grpc/pull/4


WouldYouKindly:
@aapeliv it works now!

aapeliv (approved):

PR #7238, opened 2025-11-21
Speed up tests

WouldYouKindly:
A bunch of small speedups for tests:

- Don't add regions and languages to users unless needed (actually needed by 2 tests)
- Move class creating out of frequently called functions (the speedup is tiny, but why not have it)
- Remove superfluous commits (session.commit() just before session_scope closes)
- Don't pull data unnecessarily from the db in tests (`update(User).where(...).values(is_deleted=True)` instead of `user = select(User); user.is_deleted = True`

This all saves around 5s on my machine.

### DIFF
--- a/app/backend/src/couchers/servicers/blocking.py
+++ b/app/backend/src/couchers/servicers/blocking.py
@@ -48,7 +48,7 @@ class Blocking(blocking_pb2_grpc.BlockingServicer):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "cant_block_self")
 
         if session.execute(
-            select(UserBlock)
+            select(UserBlock.id)
             .where(UserBlock.blocking_user_id == context.user_id)
             .where(UserBlock.blocked_user_id == blockee.id)
         ).scalar_one_or_none():
--- a/app/backend/src/couchers/servicers/blocking.py
+++ b/app/backend/src/couchers/servicers/blocking.py
@@ -1,5 +1,6 @@
 import grpc
 from google.protobuf import empty_pb2
+from sqlalchemy import exists
 from sqlalchemy.sql import not_, or_, union
 
 from couchers import urls
@@ -48,10 +49,12 @@ class Blocking(blocking_pb2_grpc.BlockingServicer):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "cant_block_self")
 
         if session.execute(
-            select(UserBlock.id)
-            .where(UserBlock.blocking_user_id == context.user_id)
-            .where(UserBlock.blocked_user_id == blockee.id)
-        ).scalar_one_or_none():
+            select(
+                exists()
+                .where(UserBlock.blocking_user_id == context.user_id)
+                .where(UserBlock.blocked_user_id == blockee.id)
+            )
+        ).scalar_one():
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "user_already_blocked")
         else:
             user_block = UserBlock(
--- a/app/backend/src/tests/test_fixtures.py
+++ b/app/backend/src/tests/test_fixtures.py
@@ -376,8 +376,6 @@ def generate_user(
             session.add(LanguageAbility(user_id=user.id, language_code=lang, fluency=fluency))
 
         # this expires the user, so now it's "dirty"
-        session.commit()
-
         token, _ = create_session(_MockCouchersContext(), session, user, False, set_cookie=False)
 
         # deleted user aborts session creation, hence this follows and necessitates a second commit
--- a/app/backend/src/tests/test_fixtures.py
+++ b/app/backend/src/tests/test_fixtures.py
@@ -1,12 +1,12 @@
 import os
 import re
-from collections.abc import Generator
+from collections.abc import Generator, Sequence
 from concurrent import futures
 from contextlib import contextmanager
 from dataclasses import dataclass
 from datetime import date, timedelta
 from pathlib import Path
-from typing import Sequence, Any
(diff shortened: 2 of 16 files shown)


app/backend/src/couchers/servicers/blocking.py
```diff
@@ -48,7 +48,7 @@ def BlockUser(self, request, context, session):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "cant_block_self")
 
         if session.execute(
-            select(UserBlock)
+            select(UserBlock.id)
```
aapeliv: You could even do `exists` here to make it faster

app/backend/src/tests/test_fixtures.py
```diff
@@ -438,8 +440,9 @@ def headers(self):
 
 
 def get_user_id_and_token(session: Session, username: str) -> tuple[int, str]:
-    user_id = session.execute(select(User).where(User.username == username)).scalar_one().id
-    token = session.execute(select(UserSession).where(UserSession.user_id == user_id)).scalar_one().token
+    user_id, token = session.execute(
+        select(User.id, UserSession.token).select_from(User).join(UserSession).where(User.username == username)
+    ).one()
     return user_id, token
```
aapeliv, line 445: now no need to assign them separately here?

app/backend/src/tests/test_fixtures.py
```diff
@@ -527,32 +524,28 @@ def auth_api_session(
...
+
 @contextmanager
 def api_session(token):
     """
     Create an API for testing, uses the token for auth
     """
-    channel = FakeChannel(token)
-    api_pb2_grpc.add_APIServicer_to_server(API(), channel)
-    yield api_pb2_grpc.APIStub(channel)
+    stub = _SESSION_CACHE.get((api_pb2_grpc.APIStub, token))
```
aapeliv: What's the point of this? is it an expensive operation?

aapeliv: and anyway, why not just use cachetools?

app/backend/src/tests/test_references.py
```diff
@@ -116,14 +118,6 @@ def create_host_request_by_date(
...
     )
 
-    message = Message(
-        time=from_date + timedelta(seconds=2),
-        conversation_id=conversation.id,
-        author_id=surfer_user_id,
-        text="Hi, I'm requesting to be hosted.",
-        message_type=MessageType.text,
-    )
-
```
aapeliv, line 126: Why did you delete this?

app/backend/src/tests/test_fixtures.py
WouldYouKindly: Sort of. Without desctructuring, the return type is `Row[tuple[int, str]]`, not just tuple. To keep things simple, I'd keep the assignment.

app/backend/src/tests/test_fixtures.py
WouldYouKindly: 1. It shaves 20 seconds off of test run time. Creating a channel makes several db requests, so it's fairly slow.
2. I was surprised to learn that neither functools, nor cachetools support decorating context managers. We could go with something like

```python
@contextmanager
def gis_session(token):
    @cached(cache=LRU_CACHE)
    def _(tkn: str):
        channel = FakeChannel(tkn)
        gis_pb2_grpc.add_GISServicer_to_server(GIS(), channel)
        stub = gis_pb2_grpc.GISStub(channel)
        _SESSION_CACHE[(gis_pb2_grpc.GISStub, tkn)] = stub
    yield _(token)
```

but I think, the current approach is not much worse. It's ugly, but simple.

app/backend/src/tests/test_references.py
WouldYouKindly: It's not added to the session, and is unused. Unless I'm missing something, it cannot possibly have any effect on the test?

app/backend/src/couchers/servicers/blocking.py
WouldYouKindly: Done! :)

app/backend/src/tests/test_fixtures.py
aapeliv: > Creating a channel makes several db requests, so it's fairly slow

I don't see that in the code? I think `FakeChannel.unary_unary` is called when the RPC is actually called? If not, then we should likely move the light perm check into `fake_handler`?

app/backend/src/tests/test_references.py
aapeliv: Well yes but it should still be there, it's a part of what's created when a host req is created.

Maybe it's not tested, but someday it might be...

aapeliv: (In other words, it's supposed to be an invariant that every host req has a message)

app/backend/src/tests/test_fixtures.py
WouldYouKindly: No, it's called in APIStub.__init__. We should probably move it, yeah. Then caching would make no sense, since every request will check permissions. Which raises another question - do we want to test the permission logic in every test? It's quite expensive in terms of test runtime, but it doesn't really give us much confidence. Something to think about... in the meantime, I'll move the check and remove the caching.

WouldYouKindly: @aapeliv removed the caching and moved permission check into `fake_handler`.

app/backend/src/tests/test_references.py
WouldYouKindly: Got you, brought it back and added a comment for the future me.

aapeliv (approved):
lgtm

PR #7253, opened 2025-11-24
Annotate all servicers' methods

WouldYouKindly:
Annotates the parameters and the return types of servicers' methods. This was done by automatically copying types from .pyi files generated by `protoc`.

I don't include the script that did the copying, because it's not really generic.


aapeliv (changes requested):
As discussed, please wait until we merge the editor PR.

Additionally, could you please:

1. Separate the big change that adds types to every servicer RPC into a single commit with just that change.
2. Add that commit into the git rev parse ignored commits.

aapeliv:
Close in favour of  #7503?

PR #7257, opened 2025-11-25
Validate host request length on the backend

### LINKED ISSUE #5952: Min character count for host request (backend)
Require at least 50 characters for a request

tristanlabelle:
Mirrors the frontend minimum host request length check on the backend.

Closes #5952

Added a pytest case for the new condition. This required updating existing tests whose initial request length was now invalid.

N/A

**Backend checklist**
- [ ] All tests pass

### DIFF
--- a/app/backend/src/couchers/constants.py
+++ b/app/backend/src/couchers/constants.py
@@ -63,6 +63,9 @@ ACTIVENESS_PROBE_EXPIRY_TIME = timedelta(days=4)
 HOST_REQUEST_MAX_REMINDERS = 1
 HOST_REQUEST_REMINDER_INTERVAL = timedelta(days=2)
 
+# Note: Javascript's string.length is in utf16 code units, Python's len(str) is in utf8 code units.
+HOST_REQUEST_MIN_LENGTH_UTF16 = 250  # Must match frontend
+
 ANTIBOT_FREQ = timedelta(hours=48)
 
 EVENT_REMINDER_TIMEDELTA = timedelta(hours=24)
--- a/app/backend/src/couchers/servicers/requests.py
+++ b/app/backend/src/couchers/servicers/requests.py
@@ -162,6 +162,14 @@ def _possibly_observe_first_response_time(session, host_request, user_id, respon
         )
 
 
+def _is_host_request_long_enough(text: str) -> bool:
+    # Python's len(str) does not match Javascript's string.length.
+    # e.g. len("é") == 2 but "é".length == 1.
+    # To match the frontend's validation, measure the string in utf16 code units.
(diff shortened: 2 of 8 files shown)


app/backend/src/couchers/constants.py
```diff
@@ -63,6 +63,9 @@
 HOST_REQUEST_MAX_REMINDERS = 1
 HOST_REQUEST_REMINDER_INTERVAL = timedelta(days=2)
 
+# Note: Javascript's string.length is in utf16 code units, Python's len(str) is in utf8 code units.
+HOST_REQUEST_MIN_LENGTH_UTF16 = 250  # Must match frontend
```
tristanlabelle, line 67: The ticket #5952 says 50 characters, but the frontend validates 250. Which should be used? And what definition of a character? The most i18n friendly is a composed Unicode character, though 250 Chinese characters makes for a very long message.

app/backend/src/couchers/servicers/requests.py
```diff
@@ -201,6 +202,15 @@ def CreateHostRequest(self, request, context, session):
         if to_date - from_date > timedelta(days=365):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "date_to_after_one_year")
 
+        # Check minimum length
+        text_length_utf16 = len(request.text.encode("utf-16")) / 2 - 1  # number of code units minus BOM
```
aapeliv: It's not clear to me what this is doing and why this is required vs computing `len` of the string itself?

Afaik all strings in python3 are utf8 encoded and this is computing some kind of sensible length?

app/backend/src/couchers/constants.py
aapeliv: Yes that is a good point. We may wish to do something smart here, do you have a proposal?

tristanlabelle: Short of having different per-language limits, counting grapheme clusters minus whitespace matches the user perception of a character and is slightly less punitive for CJK languages. I have most of the implementation here: https://github.com/Couchers-org/couchers/pull/7289, but I'd merge it separately since it'll involve teaching React about this definition of length.

app/backend/src/couchers/servicers/requests.py
tristanlabelle: I've extracted a function with more comments. Python's utf8 length will not match javascript's utf16 length. In many cases Javascript's length is closer to a human-perceived character so I'm implementing that in Python.

tristanlabelle:
@aapeliv This is finally ready for review. I'll then follow-up with the more human-friendly character counting in https://github.com/Couchers-org/couchers/pull/7289

aapeliv (approved):
lgtm

PR #7268, opened 2025-11-27
Add a hint to pull the latest image to 'make protos'

WouldYouKindly:


aapeliv (commented):
it might be easier to just add `--pull` to the docker command?

aapeliv (approved):

WouldYouKindly:
> it might be easier to just add `--pull` to the docker command?

idk. It's easier, but it adds 1s of latency to the command (and that's on a good connection.) Considering how often we update the image, I'd rather avoid the latency.

PR #7284, opened 2025-11-28
[WIP] ignore

aapeliv:


nabramow:
@aapeliv When I try to allow push notifications I am getting this error:

```
 ERROR  Failed to register push notification subscription: [TypeError: req.setPlatform is not a function (it is undefined)]
```

IMG_C7C9C4755D1B-1

nabramow:
Also this, when I send the test push notification it's not creating a row in the delivery_attempts table, and I see this in background_jobs:

Screenshot 2025-11-29 at 12 17 10 PM

Claude suggests: "The issue was that _send_expo was trying to access payload.topic_action, payload.title, etc. directly, but those fields don't exist on the proto. The data is JSON-encoded in payload.data (same as for web push)."

nabramow:
I pushed some changes that fixes the errors above. Now the push notifications work!

IMG_D289BF4BB696-1

Feel free to make adjustments though if you don't like these changes!

I also made it so the Test Notifications button only shows in dev environment.

PR #7285, opened 2025-11-28
Ignore stuff that should be ignored in bcow

aapeliv:
It's time to stop cheating on our coverage metrics.

Works now: https://ae066722--bcov.preview.couchershq.org/

### DIFF
--- a/app/backend/Dockerfile
+++ b/app/backend/Dockerfile
@@ -48,8 +48,8 @@ RUN --mount=type=cache,target=/root/.cache \
         --locked \
         --no-editable
 
-# We don't need them in the final image (uv.lock is more than 300Kb!)
-RUN rm /app/pyproject.toml /app/uv.lock
+# We don't need uv.lock in the final image (it's more than 300Kb!)
+RUN rm /app/uv.lock
 
 FROM base
 
--- a/app/backend/pyproject.toml
+++ b/app/backend/pyproject.toml
@@ -111,7 +111,7 @@ module = ["http_ece", "py_vapid", "luhn", "sqlalchemy_utils.*"]
 ignore_missing_imports = true
 
 [tool.coverage.run]
-omit = ["src/couchers/proto/**", "src/couchers/migrations/versions/**", "src/tests/**", "src/dummy_data.py"]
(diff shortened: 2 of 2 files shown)


app/backend/Dockerfile
```diff
@@ -48,8 +48,8 @@ RUN --mount=type=cache,target=/root/.cache \
         --locked \
         --no-editable
 
-# We don't need them in the final image (uv.lock is more than 300Kb!)
-RUN rm /app/pyproject.toml /app/uv.lock
+# We don't need uv.lock in the final image (it's more than 300Kb!)
+RUN rm /app/uv.lock
```
WouldYouKindly, line 52: Do we need it because its contains coverage settings?

app/backend/pyproject.toml
```diff
@@ -111,7 +111,7 @@ module = ["http_ece", "py_vapid", "luhn", "sqlalchemy_utils.*"]
 ignore_missing_imports = true
 
 [tool.coverage.run]
-omit = ["src/couchers/proto/*", "src/couchers/migrations/*", "src/dummy_data.py"]
+omit = ["src/couchers/proto/**", "src/couchers/migrations/versions/**", "src/tests/**", "src/dummy_data.py"]
```
WouldYouKindly: Here's an argument for measuring coverate of our tests from coverage.py author: https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html

app/backend/Dockerfile
aapeliv: yes

app/backend/pyproject.toml
aapeliv: I think right now no one is really looking at the coverage. I don't look at it partly because it's too noisy.

I'm down to do this either way, not strongly opinionated. But we do need to exclude protos and migrations just make it messy.

WouldYouKindly: Let's keep the tests, and omit protos and migrations then?

aapeliv: Sounds good.

WouldYouKindly (approved):

PR #7313, opened 2025-11-30
Try to implement dark theme

aapeliv:

todo:

1. some chrome on map page is not the right color, in particular the overlay widgets are white-on-white
2. the button to expand the map page sidebar is white-on-white
3. hosting status is black-on-black on the map page
4. message response pane in messages is white bg even on dark mode
5. "View my profile" and some other buttons are black-on-black

**Web frontend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -1,4 +1,10 @@
-import { Box, Container, GlobalStyles, useMediaQuery } from "@mui/material";
+import {
+  Box,
+  Container,
+  GlobalStyles,
+  useMediaQuery,
+  useTheme,
+} from "@mui/material";
 import { styled } from "@mui/material/styles";
 import CenteredSpinner from "components/CenteredSpinner/CenteredSpinner";
 import CookieBanner from "components/CookieBanner";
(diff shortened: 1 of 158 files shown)


nabramow:
Hey @aapeliv - first off - let's do this in chunks so we can more easily test none of the style got messed up. Especially since AI is doing it and I already see it's "improving" some unrelated things haha.

What about if we do one major page at a time? So do the events page, communities page, dashboard, map, etc in one PR each. That way it's easier to test and see if anything is awry.

Before that though - it looks like MUI has functionality built in for this that we should try first before this manual way:

https://mui.com/material-ui/customization/dark-mode/

If that doesn't work, I think we should have a streamlined way for checking if it's light or darkMode as well. React Context is a good use case for this. See this blog post for an example:

https://letsreact.org/how-to-implement-dark-light-mode-in-react-and-improve-your-user-experience


Besides that, some themes I noticed AI getting wrong:

import `theme` from theme instead of using `useTheme` hook, that's our pattern on the frontend.

For the naming, if we're adding a new color for darkMode let's set it in the theme and reference that instead of doing a new variable in every component.

aapeliv:
@claude undo all the useTheme stuff and just fix it up in theme.ts

PR #7374, opened 2025-12-06
[slop] Add photo galleries frontend

aapeliv:
@nabramow here's what CC gave me. Feel free to take/leave/do whatever.


nabramow:
I think we can probably try to use MUI's ImageList for this rather than doing totally manually:

https://mui.com/material-ui/react-image-list/

I can take this over and give it a go though.

PR #7382, opened 2025-12-06
Finish Next migration, upgrade i18next and do MUI integration

### LINKED ISSUE #5499: Finish upgrade Next to v15
Blocked by #5381 (React v19 upgrade). 

Migration Guide to v15:
https://nextjs.org/docs/app/building-your-application/upgrading/version-15

Looks like we accidentally upgraded to v15 in #6203 . It's not broken so I will leave for now but we should go through that migration guide after react v19 upgrade and make sure we got everything.

nabramow:
Closes #5499 

Finishes the next.js v15 migration. Can't do v16 yet as we need to upgrade to eslint v9 and flat config first.

Also handles the MUI integration with the pages router and fixes the "Show map" button that had text fading into the background.

...and upgraded i18next while I was in there as no changes were needed.

Doing this now as it should help with implementing dark mode which would be nice for the app.


aapeliv (approved):
full trust

PR #7387, opened 2025-12-07
Update mobile readme physical device instructions

nabramow:
On a zoom with @aapeliv getting the local build on his device set up and realized this step needed clarification.


aapeliv (dismissed):
we could also mention that they need to join the couchers apple developer group to do this

Issue #7368, opened 2025-12-06
Backend translations: make emails translatable


aapeliv:
Right now emails are hardcoded in `.mjml` files here: https://github.com/Couchers-org/couchers/tree/develop/app/backend/templates/v2

We should update these so that we can translate them.

I think the way to do it is to take each individual string and replace it with a string key, e.g.

https://github.com/Couchers-org/couchers/blob/bbcf1d7307008d5be9f401294528c8b0cfaab821/app/backend/templates/v2/account_deletion_complete.mjml#L4

replace `You have successfully deleted your account from Couchers.org.` with something like `{{ email_account_deletion_complete.success_message }}`, and extracting the actual string into `en.json`.

This will require some thinking for parts that interpolate strings. That bit will require being careful with string injection (the emails don't necessarily do this correctly everywhere at the moment).

(I also suggest employing an LLM.)

tristanlabelle:
My plan:

Leverage jinja2 parameterized filters, to support parameterized translations like: `{{ "greeting_msg"|t(name=user.name) }}`, by defining a function `t` that looks up the `greeting_msg` string and substitutes `name` using the jinja template argument `user`. Use the jinja context to pass in the user's locale and retrieve it in the translation function implementation for lookup. Make the translation function escape its output.

tristanlabelle:
Weblate component created with all email strings as of today 🎉

PR #7418, opened 2025-12-09
Basic email translation infrastructure

tristanlabelle:
Implements translating emails via a new jinja filter "v2translate" taking a string id and optional substitution arguments. This PR only migrates the body of the donation_received email as a proof of concept and to align on the design.

Design decisions:
1. `{{ "string_id"|v2translate(name=user.name) }}` syntax (for a string identified with `string_id` with one `{{name}}` substitution)
2. All substitutions get html-escaped
3. Strings can include basic html tags like \<a> or \<b> (otherwise we have to break it and concatenate, which is not i18n friendly)
4. html and plaintext emails share strings. In the case of plaintext emails, any html tags are stripped or converted into equivalents.

See #7368

- Added a unit test for the `v2translate` jinja filter.
- The modified email was already validated by a test: `uv run -m pytest src/tests/test_email.py`

N/A, usual Python testing setup.

**Backend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/i18n/locales/en.json
+++ b/app/backend/src/couchers/i18n/locales/en.json
@@ -259,16 +259,16 @@
     "no_money_title": "Keep money out of it",
     "no_money_guideline": "Keep things non-transactional. Don't exchange money while staying or hosting. Couchers.org is about learning, experiences, and getting to know people."
   },
-  "notification_donation_received": {
-    "title": "Thank you for your donation to Couchers.org!",
-    "greeting": "Dear {{name}},",
-    "thanks_donation": "Thank you so much for your donation of ${{amount}} to Couchers.org.",
-    "donation_purpose": "Your contribution will go towards building and sustaining the Couchers.org platform and community, and is vital for our goal of a completely free and non-profit generation of couch surfing.",
-    "invoice_receipt_info": "You can download an invoice and receipt for the donation here:",
-    "download_invoice": "Download invoice",
-    "questions_contact": "If you have any questions about your donation, please email us at <a href=\"mailto:donations@couchers.org\">donations@couchers.org</a>.",
-    "generosity_helps": "Your generosity will help deliver the platform for everyone.",
+  "notifications": {
+    "donation_received_title": "Thank you for your donation to Couchers.org!",
+    "dear_name": "Dear {{name}},",
+    "donation_received_thanks_amount": "Thank you so much for your donation of ${{amount}} to Couchers.org.",
+    "donation_received_purpose": "Your contribution will go towards building and sustaining the Couchers.org platform and community, and is vital for our goal of a completely free and non-profit generation of couch surfing.",
+    "donation_received_invoice_receipt_info": "You can download an invoice and receipt for the donation here:",
+    "donation_received_download_invoice": "Download invoice",
+    "donation_received_questions_contact": "If you have any questions about your donation, please email us at <a href=\"mailto:donations@couchers.org\">donations@couchers.org</a>.",
+    "donation_received_generosity_helps": "Your generosity will help deliver the platform for everyone.",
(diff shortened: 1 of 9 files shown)


app/backend/src/couchers/i18n/locales/en.json
```diff
@@ -258,5 +258,17 @@
     "be_safe_guideline": "Read profiles, references, and resources. Report any inappropriate content or behavior.",
     "no_money_title": "Keep money out of it",
     "no_money_guideline": "Keep things non-transactional. Don't exchange money while staying or hosting. Couchers.org is about learning, experiences, and getting to know people."
+  },
+  "notification_donation_received": {
```
tristanlabelle: I'm leaning towards having a single "notifications" component, and using "donation_received_" prefixes on strings, since some strings will be shared between notifications. Thoughts?

app/backend/src/couchers/notifications/background.py
```diff
@@ -45,6 +45,8 @@ def _send_email_notification(session: Session, user: User, notification: Notific
     template_args = {
         "user": user,
         "time": notification.created,
+        # Lookup strings for this notification
+        "_component": f"notification_{rendered.email_template_name}",
```
tristanlabelle: I might remove this and have all notifications look up strings from a single "notifications" component

app/backend/templates/v2/generated_html/donation_received.html
tristanlabelle, line 1: Self-blocking until I test the html version too

app/backend/src/couchers/templates/v2.py
```diff
@@ -85,6 +88,52 @@ def v2markdown(value: str) -> str:
...
 
 
+@pass_context
+def v2translate(context: Context, key: str, **kwargs: Any) -> str:
+    """
+    Jinja2 filter to translate a string key with substitutions.
+
+    Usage in template:
+        {{ "greeting_key"|v2translate(name=user.name) }}
+    """
```
kalvdans, line 106: Can you add a unit test for this function? It is quite complex and the unit test will also help translators to see how to format the translation string.

tristanlabelle: Is there a way to mock the global translations dictionary? If I'm writing a unit test I'd like to not depend on adding a real translated string.

tristanlabelle: I ended up patching `get_raw_translation_string`. See the latest.

app/backend/src/couchers/i18n/locales/en.json
tristanlabelle: Implemented in latest

app/backend/src/couchers/notifications/background.py
tristanlabelle: Implemented in latest

app/backend/src/couchers/templates/v2.py
kalvdans: Looks very good! LGTM!

kalvdans (dismissed):

app/backend/templates/v2/generated_html/donation_received.html
tristanlabelle: Oh actually that's already covered with a test:

```
assert "Thank you so much for your donation of $20 to Couchers.org." in email.html
```

tristanlabelle:
@kalvdans Could you reapprove? I changed the html expansion of newlines to `<br>` since that's the form preferred in the mjml rendering.

kalvdans (approved):
> @kalvdans Could you reapprove? I changed the html expansion of newlines to `<br>` since that's the form preferred in the mjml rendering.

I don't fully understand how we use mjml in couchers, but looks good to me! Email clients don't need proper xhtml.

PR #7429, opened 2025-12-12
Use i18next pluralization for x_minutes_ago-like strings

### LINKED ISSUE #7430: "x minutes ago"-like strings don't support pluralization
Reported by @vas2525 via Slack. Applies to "x minutes ago", "x days ago" etc.

tristanlabelle:
These strings were using manually separated singular/plural strings, which is unexpected for translators (weblate only shows one field) and does not work for all languages (e.g. some have a notion of dual plurality).

To have plural forms appear in weblate, the i18next format requires using `count` as the variable name. The English string should also expose a singular/plural and the code should pass in an integer template parameter, not a string (as does `.toFixed()`)

The code is duplicated in the native folder, so I fixed it in both places, but consolidating would be good.

Will require relocalizing those strings.

Closes #7430

I have a follow-up refactoring here: https://github.com/Couchers-org/couchers/pull/7436

Run the frontend and look at some search results

image

N/A

**Web frontend checklist**
- [x] All tests pass

**Other**


app/native/resources/locales/en.json
```diff
@@ -138,19 +138,19 @@
...
-    "one_hour_ago": "1 hour ago",
-    "x_hours_ago": "{{date}} hours ago",
-    "one_day_ago": "1 day ago",
-    "x_days_ago": "{{date}} days ago",
-    "one_week_ago": "1 week ago",
-    "x_weeks_ago": "{{date}} weeks ago",
-    "one_month_ago": "1 month ago",
-    "x_months_ago": "{{date}} months ago",
-    "one_year_ago": "1 year ago",
-    "x_years_ago": "{{date}} years ago"
```
nabramow, line 153: FYI this is a dead file but can't hurt to update it I guess. In the future not worth your time to do this one! We will delete I assume once the app is officially out and we're sure we don't need the old code.

nabramow (approved):
LGTM!

app/native/resources/locales/en.json
tristanlabelle: All of native/resources/locales is dead or something more broad/specific?

Issue #7370, opened 2025-12-06
Default language loads as English — detect and match browser/system language


c-kreuz:
Currently, when loading the Couchers site, the default language is set to English regardless of the user’s browser or system settings. This creates a mismatch for non-English speakers and requires manual adjustment to switch to their preferred language.

- [ ] Retrieve the browser/system language using standard APIs.
- [ ] Map the detected language to available translations in Couchers.
- [ ] Use internal fallback routes if the detected language is unsupported.
- [ ] Ensure this detection runs only on initial load, respecting any manual language changes made by the user afterward.
- [ ] Consider caching the user’s language preference once selected, so it persists across sessions.
- [ ] Ensure compatibility with both desktop and mobile browsers.

PR #7431, opened 2025-12-12
Fix(web): Improve browser language detection and fallback logic

rohitmane-dev:
This PR addresses #7370 by improving how we parse and match the browser's Accept-Language header.

Previously, the logic would prematurely strip region codes (e.g., converting pt-BR to pt) before checking for support. This meant that even if we explicitly supported pt-BR, the system might incorrectly fallback to the base language or English.

Changes:

Updated getBrowserLocale in middleware.ts with a more robust matching strategy:

- Exact match: Checks if the full locale (e.g., pt-BR) is supported.
- Base match: Checks if the base language (e.g.,  es from es-MX) is supported.
- Fuzzy match: Falls back to the first supported dialect (e.g., zh -> zh-Hans).

- Added comprehensive unit tests in app/web/test/middleware_language_detection.test.ts
 to verify these scenarios and prevent regressions.

Verification: I added a new test suite to cover various locale combinations. You can run it with: yarn test app/web/test/middleware_language_detection.test.ts


rohitmane-dev:
i have made a similar PR with updated branch name

PR #7432, opened 2025-12-12
Fix(web): Improve browser language detection logic and region handling

### LINKED ISSUE #7370: Default language loads as English — detect and match browser/system language

rohitmane-dev:
This PR addresses #7370 by improving how we parse and match the browser's Accept-Language header.

Previously, the logic would prematurely strip region codes (e.g., converting pt-BR to pt) before checking for support. This meant that even if we explicitly supported pt-BR, the system might incorrectly fallback to the base language or English.

Changes:

Updated getBrowserLocale in middleware.ts with a more robust matching strategy:
- Exact match: Checks if the full locale (e.g., pt-BR) is supported.
- Base match: Checks if the base language (e.g., es from es-MX) is supported.
- Fuzzy match: Falls back to the first supported dialect (e.g., zh -> zh-Hans).
- Added comprehensive unit tests in app/web/test/middleware_language_detection.test.ts to verify these scenarios and prevent regressions.
Verification: I added a new test suite to cover various locale combinations. You can run it with: yarn test app/web/test/middleware_language_detection.test.ts

Closes #7370


aapeliv:
Hi, thanks but we are not going to merge this PR. I don't understand what exactly you are trying to do here as it does not address the ticket directly. Additionally it appears this is AI-generated. A combination of confusion around the task + being AI-generated is a combo for a lot of wasted time from both sides!

rohitmane-dev:
> Hi, thanks but we are not going to merge this PR. I don't understand what exactly you are trying to do here as it does not address the ticket directly. Additionally it appears this is AI-generated. A combination of confusion around the task + being AI-generated is a combo for a lot of wasted time from both sides!

Hi, thanks for taking the time to review this.

Sorry for the confusion and for the time this caused — that wasn’t my intention. You’re right that the PR didn’t clearly address the ticket, and I should have made sure I fully understood the issue before opening it. I did use AI while working on the changes, but I take responsibility for submitting something that wasn’t well aligned with the task.

I appreciate the feedback and will treat this as a learning experience. I’ll make sure to better understand the problem and expectations before contributing again.

Thanks again for the review and for maintaining the project.

Issue #7420, opened 2025-12-10
Make email footers localizable


tristanlabelle:
Currently `background.py` appends to the email a section that allows unsubscribing using string concatenation. These strings should be localizable, and ideally be lifted to template files so they work the same way as the body of the email

PR #7457, opened 2025-12-15
Migrate email footer to template file

tristanlabelle:
Migrates the email footer from in-code string concatenation to the templating system. This will allow the footer to be localized the same way as email contents.

Next up I'll encapsulate the email templating logic so it can be reused between its two uses and tested more easily.

Relates to #7420

Run a local client/server and get any email to be sent.

After:
image

image

N/A

**Backend checklist**
- [ ] All tests pass

**Other**


app/backend/src/couchers/templates/v2.py
```diff
@@ -167,15 +167,14 @@ def send_simple_pretty_email(
     """
     template_args[CONTEXT_YEAR_KEY] = now().year
     template_args[CONTEXT_TIMEZONE_DISPLAY_KEY] = get_tz_as_text("Etc/UTC")
-
-    plain_unsub_section = "\n\n---\n\nThis is a security email, you cannot unsubscribe from it."
-    html_unsub_section = "This is a security email, you cannot unsubscribe from it."
+    template_args["_email_is_critical"] = True  # Results in no unsubscribe footer.
```
kalvdans: Nit: Please don't use leading underscore for template arguments, it affects readability and makes us believe we are using an internal variable in the email templates.

app/backend/src/couchers/notifications/background.py
```diff
@@ -52,44 +52,27 @@ def _send_email_notification(session: Session, user: User, notification: Notific
     template_args = {
         "user": user,
         "time": notification.created,
+        "_email_is_critical": rendered.is_critical,
+        "_manage_notifications_link": urls.notification_settings_link(),
+        "_notification_topic_action": rendered.email_topic_action_unsubscribe_text,
+        "_notification_topic_action_link": generate_unsub_topic_action(notification),
+        "_notification_topic_key": rendered.email_topic_key_unsubscribe_text,
+        "_notification_topic_key_link": generate_unsub_topic_key(notification),
+        "_do_not_email_link": generate_do_not_email(user),
```
kalvdans: Remove leading underscore for the template arguments.

kalvdans (dismissed):
looks good, hopefully your next PR will get rid of the duplication.

app/backend/src/couchers/templates/v2.py
tristanlabelle: The underscore-prefixed arguments support the email boilerplate (footer stuff), and can be handled generically by the email infrastructure. My rationale is to avoid clashes with email-specific variables, for example:

https://github.com/Couchers-org/couchers/blob/237c1a6f9e1351fc50869bf4894ca6af957c9086/app/backend/src/couchers/notifications/render.py#L63-L69

From the perspective of the `host_request__plain` template, the underscored variables could be considered internal.

kalvdans: I don't see how footer variables would ever clash with email variables, but if they do we could prefix them with `footer_` and `email_` respectively. Not a big deal though, I'm fine with underscore :)

tristanlabelle: I like the footer_ prefix, fixed in latest

app/backend/src/couchers/notifications/background.py
tristanlabelle: Fixed in latest

tristanlabelle:
@kalvdans could you reapprove? I changed the footer_ prefix and fixed the test

kalvdans (approved):

PR #7473, opened 2025-12-18
Fix and clarify moderator note strings

tristanlabelle:
Translators asked to fix typo in acknowledge. I also updated the text because "mod note" was confusing translators. Since there were some bad translations already, I rotated the string key to force relocalization.

**Backend checklist**
- [ ] All tests pass

**Other**


c-kreuz (dismissed):

c-kreuz:
Lgtm thanks for fixing this

tristanlabelle:
@c-kreuz , I had to iterate because of a failing test (still referencing a bad string), so I updated other strings that used "mod". I'll need re-approval.

nabramow:
@tristanlabelle I think actually we should take the mod component out of weblate as it doesn't need to be translated. It's only Jesse using it right now, so we only need English. So we don't want to waste peoples energy. Best plan of action here is just remove that component from weblate as it got added by accident.

aapeliv:
> @tristanlabelle I think actually we should take the mod component out of weblate as it doesn't need to be translated. It's only Jesse using it right now, so we only need English. So we don't want to waste peoples energy. Best plan of action here is just remove that component from weblate as it got added by accident.

There is a bit that is user-facing; but most of the mod stuff is not.

tristanlabelle:
@nabramow / @aapeliv , I've scoped down this PR to just non-moderator facing UI, and I am un-localizing the mod component in https://github.com/Couchers-org/couchers/pull/7489.

tristanlabelle:
@c-kreuz I'll need a final approval now that I've removed the mod component changes as per @nabramow 's feedback.

aapeliv (approved):

Issue #7465, opened 2025-12-17
The backend i18next implementation does not support nested keys


tristanlabelle:
The backend has a custom i18next implementation which only supports one level of key nesting. Email localization would benefit from the extra context that nesting would provide.

There also isn't pluralization support, so we should investigate libraries.

PR #7475, opened 2025-12-19
More complete i18next Python implementation

tristanlabelle:
The backend uses a simple i18next implementation, because there are no good libraries available for it. This PR adds a new implementation that supports nested keys, pluralization, and is more testable.

The new implementation is only used by tests in this PR, I will switch the backend over in a subsequent PR.

The existing implementation pre-baked languages by filling in missing strings from fallback languages, the new implementation chains lookups when a string is not found.

I think the only functional change is that the new implementation fallbacks if a string is found whose substitutions cannot be fulfilled, whereas it would previously leave the substitution, e.g. `{{name}}`.

First step towards #7465

Added plenty of unit tests.

N/A

**Backend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/i18n/i18next.py
+++ b/app/backend/src/couchers/i18n/i18next.py
@@ -1,0 +1,167 @@
+"""
+Implements localizing strings stored in the i18next json format.
+"""
+
+import re
+from dataclasses import dataclass, field
+
+from couchers.i18n.plurals import PluralRule
+
+PLURALIZABLE_VARIABLE_NAME = "count"
+"""Special variable name for which i18next supports pluralization forms."""
+
+
+@dataclass
+class I18Next:
+    """Retrieves translated strings from their keys based on the i18next format."""
+
+    languages_by_code: "dict[str, Language]" = field(default_factory=dict)
+    fallback_language: "Language | None" = None
+    """The language used to look up strings in unsupported languages."""
+
+    def add_language(self, code: str, plural_rule: PluralRule) -> "Language":
+        language = Language(code, plural_rule)
+        self.languages_by_code[code] = language
+        return language
+
+    def find_string(self, key: str, language_code: str, substitutions: dict | None = None) -> "String | None":
+        """Find the string that will be localized, applying fallbacks and variant selection."""
+        language = self.languages_by_code.get(language_code) or self.fallback_language
+        while True:
+            if language is None:
+                raise LocalizationError(language_code, key)
+
+            string = language.find_string(key, substitutions)
+            if string is None or not string.template.can_render(substitutions):
+                if language.fallback is None and language != self.fallback_language:
+                    language = self.fallback_language
+                else:
+                    language = language.fallback
+                continue
+
+            return string
+
+    def localize(self, string_key: str, language_code: str, substitutions: dict | None = None) -> str:
+        if string := self.find_string(string_key, language_code, substitutions):
+            return string.render(substitutions)
+        else:
+            raise LocalizationError(language_code, string_key)
+
+
+@dataclass
+class Language:
+    """A set of translated strings for a language."""
+
+    code: str
+    """The language code, e.g. 'en'"""
+    plural_rule: PluralRule
+    """The rule for plurals in this language."""
+    strings_by_key: "dict[str, String]" = field(default_factory=dict)
+    fallback: "Language | None" = None
+
+    def load_json_dict(self, json_dict: dict):
+        def add_strings(json_dict: dict, key_prefix: str | None):
+            for k, v in json_dict.items():
+                full_key = f"{key_prefix}.{k}" if key_prefix else k
+                if isinstance(v, str):
+                    self.add_string(full_key, v)
+                elif isinstance(v, dict):
+                    add_strings(v, key_prefix=full_key)
+                else:
+                    raise ValueError(f"Unexpected value type in language JSON: {type(v)}")
+
+        add_strings(json_dict, key_prefix=None)
+
+    def add_string(self, key: str, value: str):
+        self.strings_by_key[key] = String(key, StringTemplate.parse(value))
+
+    def find_string(self, key: str, substitutions: dict | None = None) -> "String | None":
+        # if we have a numerical "count" substitution,
+        # i18next will first search for a key with a suffix
+        # based on the plural category suggested by the count
+        # according to the current language's rules.
+        if substitutions:
+            if count := substitutions.get(PLURALIZABLE_VARIABLE_NAME):
+                if isinstance(count, int):
+                    plural_category = self.plural_rule(count)
+                    plural_key = key + "_" + plural_category.value
+                    if string := self.strings_by_key.get(plural_key):
+                        return string
+        return self.strings_by_key.get(key)
+
+    def localize(self, string_key: str, substitutions: dict | None = None) -> str:
+        string = self.find_string(string_key, substitutions)
+        if string is None:
+            raise Exception()
+        return string.render(substitutions)
+
+
+@dataclass
+class String:
+    """An i18next string key + template pair."""
+
+    key: str
+    template: "StringTemplate"
+
+    def render(self, substitutions: dict | None) -> str:
+        return self.template.render(substitutions)
+
+
+@dataclass
+class StringTemplate:
+    """A string value which may contain variable placeholders."""
+
+    segments: "list[StringSegment]"
+
+    def can_render(self, substitutions: dict | None) -> bool:
+        for segment in self.segments:
+            if segment.is_variable:
+                if not substitutions or substitutions.get(segment.text) is None:
+                    return False
+        return True
+
+    def render(self, substitutions: dict | None) -> str:
+        # Early out for simple strings without variables
+        if len(self.segments) == 1 and not self.segments[0].is_variable:
+            return self.segments[0].text
+
+        result = ""
+        for segment in self.segments:
+            if segment.is_variable:
+                if segment.text in substitutions:
+                    result += str(substitutions[segment.text])
+                else:
+                    raise ValueError(f"Missing substitution for variable '{segment.text}'")
+            else:
+                result += segment.text
+        return result
+
+    @staticmethod
+    def parse(value: str) -> "list[StringSegment]":
+        last_index = 0
+        segments: list[StringSegment] = []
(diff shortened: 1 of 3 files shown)


kalvdans:
I'm afraid I can't review this until it is complete; the tricky part I guess is how to download Unicode CLDR and bundle it in the installation. Aren't there any existing Python language support for Unicode CLDR?

tristanlabelle:
@kalvdans I consider the implementation to be complete, outside of adding a few more plural rules for other languages. We don't need to bundle Unicode CLDR as there is no lookup required, the only thing i18next requires is classifying cardinal integers into plural categories (so we look up the correct string). This is a small snippet of code per language, not something we can read from Unicode CLDR.

I avoided migrating the existing backend translation code to this implementation in this PR, to keep the review scoped. Is that migration what you would like to see in this PR?

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -0,0 +1,81 @@
+"""
+Implements Unicode CLDR pluralization rules, used by i18next.
+See https://cldr.unicode.org/index/cldr-spec/plural-rules
+"""
+
+from enum import Enum
+from typing import Callable  # noqa: UP035
```
WouldYouKindly: Why the noqa? There should be a way to avoid it

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -0,0 +1,81 @@
...
+    def en(count: int) -> PluralCategory:
+        count = abs(count)
+        if count == 1:
+            return PluralCategory.ONE  # 1 apple
+        return PluralCategory.OTHER  # 2 apples
+
+    @staticmethod
+    def es(count: int) -> PluralCategory:
+        count = abs(count)
+        if count == 1 or count == 1:
```
WouldYouKindly: ?

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -0,0 +1,81 @@
...
+            return PluralCategory.MANY  # 1000000 de manzanas
+
+        return PluralCategory.OTHER  # 2 manzanas
+
+    @staticmethod
+    def fr(count: int) -> PluralCategory:
+        count = abs(count)
+        if count == 0 or count == 1:
+            return PluralCategory.ONE  # 0 pomme, 1 pomme
+        if count > 0 and count % 1_000_000 == 0:
```
WouldYouKindly: Count cannot be less than 1 here

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+
+PLURALIZABLE_VARIABLE_NAME = "count"
+"""Special variable name for which i18next supports pluralization forms."""
+
+
+@dataclass
+class I18Next:
+    """Retrieves translated strings from their keys based on the i18next format."""
+
+    languages_by_code: "dict[str, Language]" = field(default_factory=dict)
```
WouldYouKindly, line 19: I don't think you need to quote the types

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+    fallback_language: "Language | None" = None
+    """The language used to look up strings in unsupported languages."""
+
+    def add_language(self, code: str, plural_rule: PluralRule) -> "Language":
+        language = Language(code, plural_rule)
+        self.languages_by_code[code] = language
+        return language
+
+    def find_string(self, key: str, language_code: str, substitutions: dict | None = None) -> "String | None":
```
WouldYouKindly: Please specify dict's contents

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+        """Find the string that will be localized, applying fallbacks and variant selection."""
+        language = self.languages_by_code.get(language_code) or self.fallback_language
```
WouldYouKindly:
```suggestion
        language = self.languages_by_code.get(language_code, self.fallback_language)
```

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@dataclass
+class String:
+    """An i18next string key + template pair."""
+
+    key: str
+    template: "StringTemplate"
+
+    def render(self, substitutions: dict | None) -> str:
```
WouldYouKindly: Please specify dict's contents

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+        return self.strings_by_key.get(key)
+
+    def localize(self, string_key: str, substitutions: dict | None = None) -> str:
+        string = self.find_string(string_key, substitutions)
+        if string is None:
+            raise Exception()
+        return string.render(substitutions)
+
+
+@dataclass
```
WouldYouKindly:
```suggestion
@dataclass(frozen=True, slots=True)
```

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+                if not substitutions or substitutions.get(segment.text) is None:
+                    return False
+        return True
+
+    def render(self, substitutions: dict | None) -> str:
+        # Early out for simple strings without variables
+        if len(self.segments) == 1 and not self.segments[0].is_variable:
+            return self.segments[0].text
+
+        result = ""
```
WouldYouKindly: It's probably better to collect parts in a list and then join them at the end

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+        for segment in self.segments:
+            if segment.is_variable:
+                if not substitutions or substitutions.get(segment.text) is None:
+                    return False
+        return True
+
+    def render(self, substitutions: dict | None) -> str:
+        # Early out for simple strings without variables
+        if len(self.segments) == 1 and not self.segments[0].is_variable:
+            return self.segments[0].text
```
WouldYouKindly: I would remove this bit tbh. I think it increases complexity and doesn't give any performance benefits

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+                    plural_category = self.plural_rule(count)
+                    plural_key = key + "_" + plural_category.value
+                    if string := self.strings_by_key.get(plural_key):
+                        return string
+        return self.strings_by_key.get(key)
+
+    def localize(self, string_key: str, substitutions: dict | None = None) -> str:
+        string = self.find_string(string_key, substitutions)
+        if string is None:
+            raise Exception()
```
WouldYouKindly: Probably a more descriptive error is needed. Maybe include string_key and substitutions?

kalvdans:
> We don't need to bundle Unicode CLDR as there is no lookup required

Thanks for clarifiying. Please remove all references to Unicode CLDR, since it is confusing if we don't intend to bundle it.

> I avoided migrating the existing backend translation code to this implementation in this PR, to keep the review scoped. 

That's fine, it's good to have it separately reviewed. I don't have any more remarks except the ones WouldYouKindly gave :)

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -0,0 +1,167 @@
...
+Implements localizing strings stored in the i18next json format.
+"""
+
+import re
+from dataclasses import dataclass, field
+
+from couchers.i18n.plurals import PluralRule
+
+PLURALIZABLE_VARIABLE_NAME = "count"
+"""Special variable name for which i18next supports pluralization forms."""
```
aapeliv, line 12: I think you can have these comments above the line as single-line comments (#)

app/backend/src/couchers/i18n/i18next.py
tristanlabelle: You do if a symbol's definition appears further down in the file

Image

I tend to order type declarations from high-level to low-level since that makes the file most easily understood from top to bottom, so it often requires quoting types.

app/backend/src/couchers/i18n/i18next.py
tristanlabelle: Unfortunately the Python convention is trailing docstrings and that is what Python tools expect. E.g. hover with comment above:

Image

Hover with docstring below:

Image

app/backend/src/couchers/i18n/i18next.py
tristanlabelle: Oops yes, switched to LocalizationError

app/backend/src/couchers/i18n/i18next.py
tristanlabelle: Ah I didn't know about slots!

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: Ruff suggests importing from collections.abc instead, which I thought was a bug since it doesn't have to do with collections but it turns out the typing import is deprecated. Updated

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: Leftover from negative handling

tristanlabelle:
@WouldYouKindly @kalvdans @aapeliv All comments addressed.

@kalvdans 
> Please remove all references to Unicode CLDR, since it is confusing if we don't intend to bundle it.

I disagree since I'm specifically implementing the plural rules on https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html , and the Unicode CLDR spec is what came up with the zero/one/two/few/many/other categorization so it serves as documentation. My comments specifically reference the Unicode CLDR plural rules, but let me know how you'd update them if they're ambiguous.

kalvdans (approved):
>  My comments specifically reference the Unicode CLDR plural rules, but let me know how you'd update them if they're ambiguous.

Thanks for the explanation! It was mostly the file header that caught my eye, but reading it again it makes sense :)

I don't like getattr() but that's minor thing. Could use a dict mapping from language string to cardinal categorization function instead.

tristanlabelle:
@kalvdans I'll replace the getattr in the follow-up PR that integrates this implementation.

PR #7478, opened 2025-12-19
Update mobile and frontend job ads

nabramow:
Made a new frontend dev job add to be more mid-level/senior-focused, as we really need one more person before we can mentor more juniors.

Created a less senior mobile developer ad. We only had a leader one, this seems more accurate to where we're at now and maybe less intimidating.


aapeliv (approved):
lgtm

PR #7483, opened 2025-12-20
fix(web): add showGhosts prop to filter ghost accounts

### LINKED ISSUE #7476: Ghost Account listed on Badges page
Subject: Ghost Account listed on Badges page
Description:
Go on swagster badges page

Results:
Ghost account shows up in listing and should not be visible

Backend version: develop-75d28819
Frontend version: develop-75d28819
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36
Screen resolution: 1920x911
Page: https://couchers.org/de/badges/swagster
User: [@chrisk](https://couchers.org/user/chrisk) (34148)

aapeliv:
Fixes #7476

Adds a `showGhosts` boolean prop (default `false`) to the `UsersList` component that filters out users with `isGhost=true`. Uses a type predicate to properly narrow the TypeScript type after filtering, fixing the CI type error.

---

### DIFF
--- a/app/web/components/UsersList.tsx
+++ b/app/web/components/UsersList.tsx
@@ -27,6 +27,7 @@ interface UsersListProps {
   getUserMenuItems?: (
     user: LiteUser.AsObject,
   ) => EllipsisMenuItem[] | undefined;
+  showGhosts?: boolean;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


nabramow:
Why don't we just filter out `is_ghost` for the badges queries on the backend? Frontend only makes sense to me if there is a case where we would want to show ghosts in the badges and I can't think of why we ever would.

PR #7487, opened 2025-12-20
Fix browser language detection issues

### LINKED ISSUE #7171: Not saving language
Subject: Not saving language
Description:
If I'm logged out, then changing the display language does not set the NEXT_LOCALE cookie; so the backend always responds in English!

Results:

Backend version: develop-b498d02a
Frontend version: develop-b498d02a-next
User Agent: Mozilla/5.0 (X11; Linux x86_64; rv:145.0) Gecko/20100101 Firefox/145.0
Screen resolution: 1504x1323
Page: https://next.couchershq.org/landing
User: <not logged in>

### LINKED ISSUE #7370: Default language loads as English — detect and match browser/system language

nabramow:
This should address two issues:

- If no NEXT_LOCALE cookie set, user the browser locale - Closes #7370 
- User should be able to set the language when logged out - Closes #7171 

Note we'll need to test this on stage or deploy to a branch where the backend and frontend envs match to properly test.

How to test:

- Set your browser language to something we support, in Brave it's at brave://settings/languages then put the new language AT THE TOP.
- Log out of couchers and clear all cookies
- Go to base route, it should change to the browser language
- Change the language via the picker, it should change to what you expect
- Change to English, it should work
- Clear NEXT_LOCALE cookie again
- Try a few different pages with no locale in url, i.e. /login, /signup, they should redirect to browser language (clear cookie each time)
- Change to a language different than the one your backend cookie is set to
- Log in
- It should change to your backend cookie language after login

### DIFF
--- a/app/web/features/translate/LanguagePickerSelect.test.tsx
+++ b/app/web/features/translate/LanguagePickerSelect.test.tsx
@@ -73,6 +73,13 @@ describe("LanguagePickerSelect", () => {
   });
 
   it("calls changeLanguage and re-routes with locale on selection", async () => {
+    // Mock document.cookie
+    const cookieSetter = jest.fn();
+    Object.defineProperty(document, "cookie", {
+      set: cookieSetter,
+      configurable: true,
+    });
+
     render(<LanguagePickerSelect />, { wrapper });
 
     expect(mockRouter).toEqual(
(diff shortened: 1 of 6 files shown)


aapeliv (approved):
lgtm

PR #7503, opened 2025-12-22
Add type annotations to servicers

WouldYouKindly:
- Add a script that copies type annotations from .pyi to .py GRPC files
- Use it to add type annotations to all servicers
- Take advantage of the latest mypy-protobuf and remove our workaround for OOM crash from postprocess_grpc_stubs.py
- Fix some mypy errors


aapeliv (dismissed):
lgtm if tests pass. could you also add 055ef4610f37155d31aa98f8047c329fab250d8f to the ignored revs

app/backend/src/couchers/sql.py
```diff
@@ -9,16 +9,11 @@
...
-    from typing import Protocol
-
     from couchers.materialized_views import LiteUser
 
     type _UserLike = type[User | LiteUser | SignupFlow]
     type _User = type[User | LiteUser]
-
-    class ModeratedContent(Protocol):
-        moderation_state_id: InstrumentedAttribute[int]
-        __moderation_author_column__: str
```
WouldYouKindly, line 21: This doesn't work with mypy unfortunately. I think it sees these as instance properties... I couldn't figure out how to fix this, and had to resort to enumeration all moderated types

app/backend/src/couchers/notifications/expo_api.py
```diff
@@ -90,7 +93,10 @@ def get_expo_push_receipts(ticket_ids: list[str]) -> dict[str, Any]:
             timeout=10,
         )
         response.raise_for_status()
-        return response.json().get("data", {})
+        result = response.json().get("data", {})
+        if not isinstance(result, dict):
+            raise ValueError(f"Expected a dict, but got {result}")
```
WouldYouKindly, line 98: Probably better to fail here than somewhere down the line

aapeliv (dismissed):

aapeliv (approved):

PR #7508, opened 2025-12-23
Fix alembic migrations

aapeliv:

When running migrations, alembic will generate stuff like `geom` columns, and it was recently also generating something for moderation lists. This fixes it up so that on a clean and up to date database, it generates an empty migration with no extra cruft.

**Backend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/models/rest.py
+++ b/app/backend/src/couchers/models/rest.py
@@ -501,8 +501,6 @@ class ModerationUserListMember(Base):
     user_id: Mapped[int] = mapped_column(ForeignKey("users.id"), primary_key=True)
     moderation_list_id: Mapped[int] = mapped_column(ForeignKey("moderation_user_lists.id"), primary_key=True)
 
-    __table_args__ = (UniqueConstraint("user_id", "moderation_list_id"),)
-
 
 class AntiBotLog(Base):
     __tablename__ = "antibot_logs"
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/models/rest.py
```diff
@@ -501,8 +501,6 @@ class ModerationUserListMember(Base):
     user_id: Mapped[int] = mapped_column(ForeignKey("users.id"), primary_key=True)
     moderation_list_id: Mapped[int] = mapped_column(ForeignKey("moderation_user_lists.id"), primary_key=True)
 
-    __table_args__ = (UniqueConstraint("user_id", "moderation_list_id"),)
```
tristanlabelle, line 504: I'm out of my depth as to why removing the uniqueness constraint impacts migrations and why not having the constraint is fine.

tristanlabelle (approved):

kalvdans (approved):
First I thought it is strange to create objects of old database contents, but I assume this is only filtering on the new database scheme: https://alembic.sqlalchemy.org/en/latest/api/runtime.html#alembic.runtime.environment.EnvironmentContext.configure.params.include_object

Ok for me besides commenting on the removed constraint.

app/backend/src/couchers/models/rest.py
aapeliv: The double primary keys imply this uniqueness constraint, so this is redundant.

PR #7511, opened 2025-12-23
Migrate pluralizable strings to i18next v4 format

### LINKED ISSUE #7453: Weblate and the frontend disagree on the i18next version
See code (v4 compatibility):

https://github.com/Couchers-org/couchers/blob/87f963f09f885e42f87c44db8f5e716efcb955b1/app/web/next-i18next.config.js#L22

but Weblate (v3):

Image

So Weblate interprets and updates the json files as v3. It looks like the library is backwards compatible so the app works still work that way, but we should update to v4 across the board to avoid issues, which involves updating pluralized keys.

tristanlabelle:
As per #7453, we mix various i18next plural formats across our stacks. This standardizes on the latest, which uses unicode CLDR suffixes for plurals. See https://www.i18next.com/misc/json-format . Note that some languages use one/other (EN/DE/NL), while others use one/many/other (FR/ES/PT) or one/few/other (RU), or just a single string (ZH).

After merging this, I will have to update Weblate to tell it to use the v4 format for those components. There might also be merge conflicts (hopefully unlikely) that I will resolve.

Fixes #7453

Switch languages and check event attendees count for correct pluralization.

image

N/A

**Web frontend checklist**
- [ ] All tests pass

**Other**


tristanlabelle:
@nabramow ping!

aapeliv (dismissed):
lgtm if spot tested locally and on vercel and it works

tristanlabelle:
@aapeliv  Rebased to fix new Catalan string. This actually fixes pluralization currently broken:

image
image

Spot tested in dashboard, profile.

aapeliv (approved):
lgtm. i thought we had "members" on the dashboard, do you know when this broke?

tristanlabelle:
@aapeliv , I also remember this working so I'm surprised about it too. It's like the mismatch between Weblate saving i18next v3 files and the frontend expecting v4 files finally kicked in. I think this is the right fix anyways though.

PR #7512, opened 2025-12-23
Fix most of the type errors

WouldYouKindly:
The only type errors left are caused by hybrid_property and CouchersSelect - I'll address these in separate PRs


app/backend/src/couchers/servicers/communities.py
```diff
@@ -375,6 +377,8 @@ def ListGuides(
     def ListEvents(
         self, request: communities_pb2.ListEventsReq, context: CouchersContext, session: Session
     ) -> communities_pb2.ListEventsRes:
+        from couchers.servicers.events import event_to_pb
```
aapeliv: why was this added? don't see the need for it

app/backend/src/couchers/servicers/editor.py
```diff
@@ -60,11 +61,11 @@ def volunteer_to_pb(session, volunteer: Volunteer) -> editor_pb2.Volunteer:
     )
 
 
-def generate_new_blog_post_notifications(payload: jobs_pb2.GenerateNewBlogPostNotificationsPayload):
+def generate_new_blog_post_notifications(payload: jobs_pb2.GenerateNewBlogPostNotificationsPayload) -> None:
     with session_scope() as session:
         all_users_ids = session.execute(select(User.id).where(User.is_visible)).scalars().all()
         for user_id in all_users_ids:
-            context = make_background_user_context(user_id=user_id)
+            make_background_user_context(user_id=user_id)
```
aapeliv: If this is not used, it can go?

app/backend/src/couchers/servicers/search.py
```diff
@@ -523,16 +569,17 @@ def _user_search_inner(request, context, session):
...
         .order_by(User.recommendation_score.desc())
         .limit(page_size + 1)
     )
     res = session.execute(statement).all()
+    users: list[int]
     if res:
-        users, rec_scores = zip(*res)
+        users, rec_scores = zip(*res)  # type: ignore[assignment]
     else:
         users = []
```
aapeliv: You can just put the initialization outside the if then and no need for else

.git-blame-ignore-revs
```diff
@@ -12,5 +12,6 @@ ed6cdae66d80d46c74b8f29d14c488831d5d6e5c
 81c0f8bf5494e8c92dce72e4baf98a2e8bf5906b
 # Split models.py into multiple files
 2b88cfc3d827781444f3f8f79b1bcc16d72fc486
-# Add types to grpc servicers
+# Add types
 055ef4610f37155d31aa98f8047c329fab250d8f
+10ffc6710ca1c5e59144765c21d01df13681ad30
```
aapeliv: this is the wrong commit now? I think you don't need to include this. and please only include stuff that contains entirely mechanical work (e.g. type annotations/formatting/etc), but if you in the same commit also do any even slightly real changes, let's try to keep it out of here

app/backend/src/couchers/servicers/communities.py
WouldYouKindly: Claude was fixing circular impors. This is indeed not needed - removed

app/backend/src/couchers/servicers/editor.py
WouldYouKindly: I thought it had side effects like checking user's permissions, but it doesn't. Removed!

app/backend/src/couchers/servicers/search.py
WouldYouKindly: True, done

.git-blame-ignore-revs
WouldYouKindly: Very good point. Removed the commit.

aapeliv (approved):

app/backend/src/couchers/servicers/auth.py
```diff
@@ -342,7 +348,7 @@ def SignupFlow(
             user.profile_gallery_id = profile_gallery.id
 
             if flow.filled_feedback:
-                form = ContributorForm(
+                form_ = ContributorForm(
```
kalvdans, line 351: Why underscore here?

kalvdans (approved):
Great job annotating types!

app/backend/src/couchers/servicers/auth.py
WouldYouKindly: @kalvdans `form` is defined above, and has a different type: `form = request.feedback`

PR #7513, opened 2025-12-23
Extract Select methods to functions to preserve type information

WouldYouKindly:
Problem: we lose type information when using CouchersSelect:

```python

user = session.execute(
     select(User).where_username_or_email(request.user).where(~User.is_deleted)
).scalar_one_or_none()

reveal_type(user)
# Revealed type is "Any | None"
```

Proposed solution: instead of subclassing Select, use functions to generate where conditions:

```python
from sqlalchemy import select

user = session.execute(
     select(User)
    .where(username_or_email(request.user))
    .where(~User.is_deleted)
).scalar_one_or_none()

reveal_type(user)
# Revealed type is "couchers.models.users.User | None"
```

IMO, this looks as good as the current solution. 

One caveat is that it doesn't work as nice for cases when we need to join with another table. There, we have to pass the query to a function:

```python
query = select(User).where(~User.is_deleted)
query = where_moderated_content_visible(query, context, table=...)
user = session.execute(query).scalar_one_or_none()

reveal_type(user)
# Revealed type is "couchers.models.users.User | None"
```

I think having correct types for queries is very important. Right now all calls to `session.execute` return `Any` which allows to access any fields, leading to unsafe code. The above is the best solution I was able to find - I quite like the ergonomics of subclassing Select, but unfortutely there's no way to keep type safety with that approach.

This PR only extracts code into separate functions. If/when we merge it, I will update the callers (with a lot of help from Claude 😅 )


app/backend/src/couchers/sql.py
```diff
@@ -26,207 +26,256 @@ class CouchersSelect(Select[Any]):
...
-        TODO: if you use this with a non-null column, check what's going on
-        """
-        hr_visible = exists(
-            couchers_select(HostRequest)
-            .where(HostRequest.moderation_state_id == column)
-            .where_moderated_content_visible(context, HostRequest)
+        return self.where(moderation_state_column_visible(context, column))
+
+
+def username_or_email(value: str, table: "_UserLike" = User):
```
tristanlabelle: Can the result of these be typed? It looks like `ExpressionRole` might serve this purpose.

tristanlabelle (dismissed):
Makes sense, yay for more typing.

aapeliv (commented):
I don't understand what the benefit of this change is?

WouldYouKindly:
> I don't understand what the benefit of this change is?

@aapeliv (I assume you've read the PR description, but if you haven't, please check it.)

Right now, result of all calls to `session.execute()` are typed as `Any`. This means that we get no type errors for accessing missing properties:

```python
user = session.execute(
     select(User).where_username_or_email(request.user).where(~User.is_deleted)
).scalar_one_or_none()

user.thingamajig + 42  # no type error! 
```

It also makes refactoring much harder, because your IDE cannot easily find all callsites.

This PR fixes that.

aapeliv:
Well this does not actually make the change? Or am I missing something?

WouldYouKindly:
> Well this does not actually make the change? Or am I missing something?

This PR just addes the functions that can later be used to replace CouchersSelect. I wanted to see if you're fine with that before doing the rest of the work.

app/backend/src/couchers/sql.py
WouldYouKindly: Of course! Thanks for spotting, done

aapeliv:
Yes happy to have this merged

WouldYouKindly:
@tristanlabelle @aapeliv ready for review! I removed CouchersSelect and switched to functions

tristanlabelle (approved):
The loss of the fluent syntax is a little unfortunate. If there was a way to extend Select[T] with the additional methods in a way that mypy understands, that'd be great, but even if that's not possible I think the typing benefits are greater than the slight readability hit and we should move forward. Thanks!

WouldYouKindly:
@aapeliv merging on Tristan's authority and since you previously said you're fine with the idea. Please feel free to add comments - will fix in a follow-up PR!

PR #7522, opened 2025-12-25
Update backend types to support ints in string substitutions

tristanlabelle:
String substitution dictionaries were previously typed as `dict[str, str]`, but we need to support int values for pluralization.

Once I integrate i18next.py in the backend, I'll probably define a `SubstitutionDict` type that we can reuse, but for now I'm keeping that code on an island of its own.

N/A, non-executable code

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/i18n/i18n.py
+++ b/app/backend/src/couchers/i18n/i18n.py
@@ -96,7 +96,7 @@ def get_translations() -> dict[str, dict[str, dict[str, str]]]:
 
 
 def get_raw_translation_string(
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/i18n/i18n.py
tristanlabelle, line 1: @WouldYouKindly , I'm happy to wait after https://github.com/Couchers-org/couchers/pull/7512 merges to rebase/merge this.

kalvdans (approved):

WouldYouKindly (approved):

PR #7525, opened 2025-12-25
Fix remaining type errors

WouldYouKindly:
This fixes all remaining type errors - `make mypy` now passes. 

Most significant changes:
- change hybrid_property's to use sqlalchemy 2.0 style, understood by mypy
- Migrate from `backref` to `back_populates` for relationship. `backref` is considered legacy and confuses mypy, as it sees no attribute on the referenced model. I only changed `backref`s that were causing errors for now.
- Removed assignment to `Event.last_changed`, since the table has no such column, and it had no effect.

### DIFF
--- a/app/backend/src/couchers/i18n/i18n.py
+++ b/app/backend/src/couchers/i18n/i18n.py
@@ -1,4 +1,5 @@
 import json
+from collections.abc import Mapping
 from functools import lru_cache
 from pathlib import Path
 
@@ -96,7 +97,7 @@ def get_translations() -> dict[str, dict[str, dict[str, str]]]:
 
 
 def get_raw_translation_string(
-    lang: str | None, component: str, string_name: str, *, substitutions: dict[str, str | int] | None = None
+    lang: str | None, component: str, string_name: str, *, substitutions: Mapping[str, str | int] | None = None
 ) -> str:
     """
     Retrieves a translated string from the all_langs_all_strings dictionary
--- a/app/backend/src/couchers/i18n/plurals.py
+++ b/app/backend/src/couchers/i18n/plurals.py
@@ -37,7 +37,7 @@ class PluralRules:
         lang_family = lang if separator_index == -1 else lang[0:separator_index]
 
         # Resolve one of the methods below
-        return getattr(PluralRules, lang_family, default=None)
+        return getattr(PluralRules, lang_family, None)
 
     @staticmethod
     def de(count: int) -> PluralCategory:
--- a/app/backend/src/couchers/models/events.py
+++ b/app/backend/src/couchers/models/events.py
@@ -1,6 +1,6 @@
 import enum
 from datetime import datetime
-from typing import cast
+from typing import TYPE_CHECKING, cast
 
 from geoalchemy2 import Geometry
 from psycopg2.extras import DateTimeTZRange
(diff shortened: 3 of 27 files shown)


app/backend/src/couchers/i18n/i18n.py
```diff
@@ -96,7 +97,7 @@ def get_translations() -> dict[str, dict[str, dict[str, str]]]:
 
 
 def get_raw_translation_string(
-    lang: str | None, component: str, string_name: str, *, substitutions: dict[str, str | int] | None = None
+    lang: str | None, component: str, string_name: str, *, substitutions: Mapping[str, str | int] | None = None
```
WouldYouKindly, line 100: mypy complained about covariancy

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -37,7 +37,7 @@ def for_language(lang: str) -> PluralRule | None:
         lang_family = lang if separator_index == -1 else lang[0:separator_index]
 
         # Resolve one of the methods below
-        return getattr(PluralRules, lang_family, default=None)
+        return getattr(PluralRules, lang_family, None)
```
WouldYouKindly, line 40: getattr doesn't allow keyword argument, this was actually failing at runtime

app/backend/src/couchers/servicers/admin.py
```diff
@@ -545,7 +545,7 @@ def ListUserIds(
         user_ids = (
             session.execute(
                 select(User.id)
-                .where(or_(User.id <= next_user_id, next_user_id == 0))
+                .where(or_(User.id <= next_user_id, to_bool(next_user_id == 0)))
```
WouldYouKindly, line 548: mypy complained about passing python booleans to `where`. We have to pass `sqlalchemy.true/false` instead (probably a good thing, since it will prevent us from accidentally passing booleans instead of comparing columns)

app/backend/src/couchers/servicers/donations.py
```diff
@@ -24,7 +24,7 @@ def _create_stripe_customer(session: Session, user: User) -> None:
     customer = stripe.Customer.create(
         email=user.email,
         # metadata allows us to store arbitrary metadata for ourselves
-        metadata={"user_id": user.id},
+        metadata={"user_id": user.id},  # type: ignore[dict-item]
```
WouldYouKindly, line 27: @aapeliv stripe expects it to be dict[str, str], but it seems that we don't really use it anymore?

app/backend/src/couchers/servicers/events.py
```diff
@@ -607,7 +607,6 @@ def UpdateEvent(
         if request.HasField("title"):
             notify_updated.append("title")
             event.title = request.title.value
-            event.last_edited = now()
```
WouldYouKindly, line 610: Event doesn't have such a column.

app/backend/src/couchers/servicers/donations.py
aapeliv: We need the info in stripe just to have it there

app/backend/src/couchers/servicers/events.py
aapeliv: O.o

WouldYouKindly: EventOccurence does, and we already update it there

app/backend/src/couchers/models/events.py
```diff
@@ -142,6 +146,10 @@ class EventOccurrence(Base):
     )
 
     photo = relationship("Upload")
+    attendances = relationship("EventOccurrenceAttendee", back_populates="occurrence", lazy="dynamic")
```
tristanlabelle, line 149: Does this also need a `: Mapped["..."]` annotation?

app/backend/src/couchers/utils.py
```diff
@@ -411,11 +411,11 @@ def get_tz_as_text(tz_name: str) -> str:
...
-    return SimpleNamespace(user_id=0)
-
-
 def not_none[T](x: T | None) -> T:
     if x is None:
         raise ValueError("Expected a value but got None")
     return x
+
+
+def to_bool(value: bool) -> ColumnElement[bool]:
```
tristanlabelle, line 420: In a broadly named file like `utils.py`, it's not obvious that `to_bool` has to do with sql. Maybe move to `couchers.sql` or rename to `to_bool_column`?

tristanlabelle (approved):
Looks good!

app/backend/src/couchers/utils.py
WouldYouKindly: Good point, will move in a later PR.

app/backend/src/couchers/models/events.py
WouldYouKindly: It does :(

PR #7526, opened 2025-12-25
Mypy in CI

WouldYouKindly:


app/backend/src/couchers/templates/v2.py
```diff
@@ -96,6 +96,17 @@ def v2markdown(value: str) -> str:
     return md.render(value)  # type: ignore[no-any-return]
 
 
+def replace_tag(match: re.Match[str]) -> str:
```
tristanlabelle, line 99: Does mypy have issues with local functions?

app/backend/src/couchers/i18n/i18n.py
```diff
@@ -96,7 +97,7 @@ def get_translations() -> dict[str, dict[str, dict[str, str]]]:
 
 
 def get_raw_translation_string(
-    lang: str | None, component: str, string_name: str, *, substitutions: dict[str, str | int] | None = None
+    lang: str | None, component: str, string_name: str, *, substitutions: Mapping[str, str | int] | None = None
```
tristanlabelle, line 100: If taking Mapping is a best practice (because the dict is only read?), does mypy enforce it or where can a dev learn about it?

app/.gitlab-ci.yml
```diff
@@ -383,6 +383,25 @@ test:backend-format:
...
+test:backend-mypy:
+  needs: ["build:backend"]
+  stage: test
+  image: $BACKEND_TAG
+  inherit:
+    # no docker login
+    default: false
+  script:
+    - cd app/backend
+    - mypy src/couchers
```
tristanlabelle, line 395: Is this mypy necessarily the same as would be selected by `uv run -m mypy`? (pinnable in pyproject.toml)

tristanlabelle (commented):
I'm excited for this. I believe it really helped quality at my previous job. Happy to review fully when ready.

app/backend/src/couchers/templates/v2.py
WouldYouKindly: @tristanlabelle probably better to leave the comments at https://github.com/Couchers-org/couchers/pull/7525 (if you have any more). This will be rebased once that one is merged

WouldYouKindly: But to answer your question: mypy is fine with local functions, but performance-wise, it's better to avoid those. Usually it doesn't matter, but if the templating function is called many times during a single template rendering, it can lead to a couple of ms lost. It's tiny, but why lose performance if you can avoid it without sacrificing anything

app/backend/src/couchers/i18n/i18n.py
WouldYouKindly: This was prompted by mypy error, yes. Try changing it back to dict and run mypy over it - the resulting error message will give you further hints. Feel free to ping me in DM if needed!

app/.gitlab-ci.yml
WouldYouKindly: There's no `uv` in the image when we run this :(

WouldYouKindly:
@aapeliv @tristanlabelle ready for review!

app/backend/src/couchers/i18n/i18next.py
```diff
@@ -114,7 +114,7 @@ def render(self, substitutions: dict[str, str | int] | None) -> str:
 class StringTemplate:
     """A string value which may contain variable placeholders."""
 
-    segments: "list[StringSegment]"
+    segments: list["StringSegment"]
```
WouldYouKindly, line 117: SQLAlchemy requires this form on relationships, so I changed it everywhere for consistency

e.g., `users: Mapped["list[User]"] = relationship(...)` fails at runtime - it treats the column as `Mapped["User"]` for some reason.

But `users: Mapped[list["User"]] = relationship(...)` (`list` is unquoted) works.

app/backend/src/couchers/models/background_jobs.py
```diff
@@ -52,7 +52,7 @@ class BackgroundJob(Base):
     payload: Mapped[bytes] = mapped_column(Binary)
 
     # if the job failed, we write that info here
-    failure_info: Mapped[str | None] = mapped_column(String, nullable=True)
```
tristanlabelle, line 55: Confirming, this is not needed because of https://docs.sqlalchemy.org/en/21/orm/mapping_api.html#sqlalchemy.orm.mapped_column ?

> If omitted, the nullability is derived from the type annotation based on whether or not typing.Optional (or its equivalent) is present.

app/backend/src/couchers/models/clusters.py
```diff
@@ -68,32 +73,36 @@ class Cluster(Base):
...
+    events: Mapped["Event"] = relationship(
+        "Event", backref="clusters", secondary="cluster_event_associations", viewonly=True
+    )
+    discussions: Mapped["Discussion"] = relationship(
         "Discussion", backref="clusters", secondary="cluster_discussion_associations", viewonly=True
     )
 
     # includes also admins
-    members = relationship(
+    members: DynamicMapped["User"] = relationship(
```
tristanlabelle, line 105: I thought the quoting was only needed if the declaration of User happens further below in the file, but here User is clearly in scope when we get to declarations so I didn't expect needing the `if TYPE_CHECKING` trick, or is it because we have circular references between files?

tristanlabelle (approved):
Looks good!

app/backend/src/couchers/models/background_jobs.py
WouldYouKindly: yes, exactly

app/backend/src/couchers/models/clusters.py
WouldYouKindly: Yes, TYPE_CHECKING is to avoid a circular depdencecy. I'm not 100% sure that there would be one, but it was easier to add TYPE_CHECKING than to check :) 

And since User won't be there at runtime, it has to be a string.

PR #7530, opened 2025-12-26
Profile photo galleries

### LINKED ISSUE #4091: [pictures] Being able to upload multiple images
first do this issue!
https://github.com/Couchers-org/couchers/issues/3887

**As a user:**
I want to upload multiple images for 'my home' section, so the surfers can see the pictures of the place beforehand.  

**current situation:**
it's only possible to upload one image (and its the profile image).  

**tasks**:  
- Implement a **component** which shows a list of pictures, then allow to upload more picture and display all the pictures uploaded
- It's only a component **later** will be implemented on the page

**inspiration**

Bildschirmfoto 2023-10-22 um 20 27 41

nabramow:
Closes #4091 

- Adds backend migration to sync galleries for existing users and make galleries source of truth for if they have uploaded at least one photo
- Updates materialized view
- Adds gallery_user_id to the protos so frontend can access it
- Adds drag and drop upload implementation to EditProfilePage
- Adds gallery and PhotoLightbox to frontend ProfilePage

**TO TEST:**
- Upload and delete photos then view them on your profile on web and mobile view
- Ideally also log into an account without strong verification - should only be allowed 1 photo (or you can go delete your strong verification for your user on SQL)


app/backend/src/couchers/models/users.py
```diff
@@ -399,11 +399,25 @@ class User(Base):
 
     @hybrid_property
     def has_completed_profile(self) -> bool:
-        return self.avatar_key is not None and self.about_me is not None and len(self.about_me) >= 150
+        # Check if user has at least one photo in their profile gallery
+        has_photo = self.profile_gallery is not None and len(self.profile_gallery.photos) > 0
```
aapeliv: This will fetch the full profile gallery, need to replace with an `exists` or similar

app/backend/src/couchers/models/users.py
```diff
@@ -399,11 +399,25 @@ class User(Base):
...
+        return has_photo and self.about_me is not None and len(self.about_me) >= 150
 
     @has_completed_profile.expression
     def has_completed_profile(cls):
-        return (cls.avatar_key != None) & (func.character_length(cls.about_me) >= 150)
+        # Check if user has at least one photo in their profile gallery
+        from couchers.models import PhotoGalleryItem
```
aapeliv: No inline imports please

app/backend/src/couchers/models/users.py
```diff
@@ -399,11 +399,25 @@ class User(Base):
...
+
+        return (
+            (cls.profile_gallery_id != None)
+            & (
+                sa_select(func.count(PhotoGalleryItem.id))
```
aapeliv: Replace with `exists`

app/backend/src/couchers/materialized_views.py
```diff
@@ -148,6 +149,15 @@ def make_lite_users_selectable(create: bool = False) -> Select[Any]:
         .subquery(name="sv_subquery")
     )
 
+    # Subquery to get the first photo from each user's profile gallery
+    first_gallery_photo_subquery = sa_select(
+        PhotoGalleryItem.gallery_id,
+        PhotoGalleryItem.upload_key,
+        func.row_number()
+        .over(partition_by=PhotoGalleryItem.gallery_id, order_by=PhotoGalleryItem.position)
+        .label("rn"),
```
aapeliv: I think you can select the min position upload key and that will be cleaner?

aapeliv (changes requested):
Commented on backend. You'll probably want to get rid of `avatar_key` stuff from the user model if you migrate fully to galleries.

app/backend/src/couchers/models/users.py
nabramow: @aapeliv This is to avoid a circular import:

* User has profile_gallery_id → PhotoGallery
* PhotoGallery has owner_user_id → User

This creates a circular dependency at the module level, which Python can't resolve at import time.

Would you rather me move all gallery-checking logic to a separate service layer? We don't have a galleries service yet.

app/backend/src/couchers/models/users.py
```diff
@@ -405,12 +404,26 @@ class User(Base):
...
+        # Check if user has at least one photo in their profile gallery
+        if self.profile_gallery_id is None:
+            has_photo = False
+        else:
+            session = object_session(self)
+            has_photo = session.query(
+                sa_select(1).where(PhotoGalleryItem.gallery_id == self.profile_gallery_id).exists()
+            ).scalar()
+
+        return has_photo and self.about_me is not None and len(self.about_me) >= 150
```
WouldYouKindly: IMO this is getting too complex for a model method. Should we move it into a function (e.g., in `couchers.sql`)?

app/backend/src/couchers/models/users.py
```diff
@@ -405,12 +404,26 @@ class User(Base):
 
     @hybrid_property
     def has_completed_profile(self) -> bool:
-        return self.avatar_key is not None and self.about_me is not None and len(self.about_me) >= 150
+        # Check if user has at least one photo in their profile gallery
+        if self.profile_gallery_id is None:
+            has_photo = False
+        else:
+            session = object_session(self)
+            has_photo = session.query(
```
WouldYouKindly: session.query is a legacy form; use session.execute, please

app/backend/src/couchers/materialized_views.py
```diff
@@ -148,6 +149,18 @@ def make_lite_users_selectable(create: bool = False) -> Select[Any]:
...
+    first_gallery_photo_subquery = (
+        sa_select(
+            PhotoGalleryItem.gallery_id,
+            PhotoGalleryItem.upload_key,
+        )
+        .distinct(PhotoGalleryItem.gallery_id)
+        .order_by(PhotoGalleryItem.gallery_id, PhotoGalleryItem.position)
+        .subquery(name="first_photo")
+    )
+
```
WouldYouKindly, line 161: This looks almost exactly like the query above - can we use the same query in both places?

app/backend/src/couchers/servicers/auth.py
```diff
@@ -707,9 +708,19 @@ def GetInviteCodeInfo(
...
 
+        # Get first photo from profile gallery for avatar
+        first_gallery_photo = None
+        if user.profile_gallery_id:
+            first_gallery_photo = session.execute(
+                select(PhotoGalleryItem)
+                .where(PhotoGalleryItem.gallery_id == user.profile_gallery_id)
+                .order_by(PhotoGalleryItem.position)
+                .limit(1)
+            ).scalar_one_or_none()
```
WouldYouKindly: Should we move it into a function to not repeat ourselves?

app/backend/src/couchers/migrations/versions/6862ecf6494d_backfill_missing_profile_galleries.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        INSERT INTO photo_gallery_items (gallery_id, upload_key, position, caption)
+        SELECT
+            u.profile_gallery_id,
+            u.avatar_key,
+            0.0,  -- First position
+            NULL  -- No caption initially
+        FROM users u
+        WHERE u.avatar_key IS NOT NULL
+        AND u.profile_gallery_id IS NOT NULL
```
WouldYouKindly: By this point it can't be null, right?

app/backend/src/couchers/migrations/versions/6862ecf6494d_backfill_missing_profile_galleries.py
```diff
@@ -0,0 +1,248 @@
...
+        AND NOT EXISTS (
+            -- Don't add if it's already in the gallery
+            SELECT 1
+            FROM photo_gallery_items pgi
+            WHERE pgi.gallery_id = u.profile_gallery_id
+            AND pgi.upload_key = u.avatar_key
+        )
```
WouldYouKindly: We can use on conflict do nothing, since we have a unique index on "gallery_id", "upload_key"

app/backend/src/couchers/models/users.py
WouldYouKindly: Another option is to add a boolean column and calculate this at insert/update time. That way we'll avoid a query to photo galleries every time we fetch a user info! And it would also simplify lite_users

app/backend/src/couchers/migrations/versions/6862ecf6494d_backfill_missing_profile_galleries.py
nabramow: Ah yeah that's true as  they get one even if they don't have photos yet.

app/backend/src/couchers/servicers/auth.py
nabramow: Yes, good catch!

app/backend/src/couchers/models/users.py
nabramow: I'm going to give the second one an attempt!

app/web/features/profile/locales/en.json
```diff
@@ -310,5 +311,30 @@
   "hosting_preferences_success_message": "Hosting preferences saved!",
   "couldnt_read_file": "Couldn't read file",
   "select_an_image": "Select an image",
-  "names_profile_photo": "{{name}}'s profile photo"
+  "names_profile_photo": "{{name}}'s profile photo",
+  "gallery": {
+    "photo": "Photo",
```
tristanlabelle: Hello! Short strings are the most difficult to translate because they don't provide much context as to how they are used in the UI. A good rule of thumb is: the shorter the string, the longer the key should be. It's very useful to translators that you're using a "gallery" grouping, and adding some suffixes as to what the role of the strings in the UI (like the _error below), or decoupling the key from the text to describe the role in the UI helps even more. See guidance in:

https://github.com/Couchers-org/couchers/blob/develop/docs/localization.md

The bottom line is: it's worth being a little verbose and specific when naming keys because looping back after a translator reports they don't have enough context is much more inefficient.

app/web/features/profile/locales/en.json
```diff
@@ -31,6 +31,7 @@
     "travel_section": "Travelled to",
     "lived_section": "Lived in",
     "map_section": "Travel map",
+    "photos_section": "Photos",
```
tristanlabelle, line 34: If we're introducing the term Photos in this PR, any thoughts about "photos" vs "pictures"? I'm used to seeing the term "pictures" on websites I use. "Profile Photos" doesn't sound as familiar as "Profile Pictures".

app/web/features/profile/locales/en.json
nabramow: I've updated the translation key, and also updaterd the .cursorrules file to make sure AI knows the rules as well. Feel free to check that out in the files to make sure you agree with the rules I added.

app/web/features/profile/locales/en.json
nabramow: I always think of photos as camera photos and pictures can be illustrations.

tristanlabelle: I checked Couchsurfing, AirBnB and Booking and they all use "Photos". Carry on!

app/web/.cursorrules
```diff
@@ -48,6 +48,29 @@ To run the app locally in development, use `yarn start` (not Docker)
...
+  - `_error`: Error message
+  - `_tooltip`: Tooltip text
+- ✅ **DO** use hierarchical grouping for related strings
+  - Example: `communities.events.start_date_label`
+- ✅ **DO** use descriptive placeholder names
+  - Good: `{{fullName}}`, `{{date}}`, `{{time}}`
+  - Bad: `{{x}}`, `{{val}}`
+- ❌ **AVOID** single words or very short keys like `photo`, `update`, `delete`
+- ❌ **AVOID** keys that replicate the English text
+- ❌ **AVOID** generic strings used in multiple contexts (translators need context!)
```
tristanlabelle: Good call on telling AI about localization! All the info here is just as valuable for humans though, so can you instead update `localization.md` so that we only point the AI to that file? Otherwise this copy will certainly get out of sync. I'm also happy to help do this as a separate PR since it's unrelated. Let me know.

app/web/features/profile/locales/en.json
```diff
@@ -312,5 +313,30 @@
...
+    "reorder_success": "Photos reordered successfully!",
+    "upload_success": "Photo uploaded successfully!",
+    "profile_photos_title": "Profile Photos",
+    "profile_photos_description": "Add photos of yourself to help others get to know you. Drag and drop to reorder.",
+    "verification_required_for_more_photos": "You've reached your photo limit. Complete strong verification to add more photos.",
+    "get_verified": "Get verified",
+    "photo_viewer_aria": "Photo viewer",
+    "previous_photo": "Previous photo",
+    "next_photo": "Next photo",
+    "close_viewer": "Close photo viewer"
```
tristanlabelle, line 340: Here's an attempt at more descriptive labels, but please double-check as I'm not sure what's a header or a label and what's a button or a link (though distinguishing those probably doesn't matter for translation). I renamed "empty_gallery_header" but not sure if this is for a section or the entire gallery. There's also both "aria" and "a11y". Do you think "aria" is clearer? I can update the guidance.

```suggestion
    "photo_item_a11y": "Photo",
    "primary_header": "Primary",
    "add_photo_button": "Add photo",
    "uploading_progress_message": "Uploading...",
    "remove_photo_button": "Remove photo",
    "select_photo_button": "Select a photo",
    "photo_pagination": "{{current}} of {{total}}",
    "empty_gallery_header": "Add your first photo",
    "empty_gallery_description": "Click here or the button above to upload a photo",
    "no_photos": "No photos yet",
    "add_photo_error": "Failed to add photo. Please try again.",
    "remove_photo_error": "Failed to remove photo. Please try again.",
    "move_photo_error": "Failed to reorder photos. Please try again.",
    "reorder_success": "Photos reordered successfully!",
    "upload_success": "Photo uploaded successfully!",
    "profile_photos_header": "Profile Photos",
    "profile_photos_description": "Add photos of yourself to help others get to know you. Drag and drop to reorder.",
    "verification_required_for_more_photos": "You've reached your photo limit. Complete strong verification to add more photos.",
    "get_verified_button": "Get verified",
    "photo_viewer_aria": "Photo viewer",
    "previous_photo_button": "Previous photo",
    "next_photo_button": "Next photo",
    "close_viewer_button": "Close photo viewer"
```

app/web/features/profile/locales/en.json
tristanlabelle: Thanks, and good call on the AI instructions -- I followed up on that separately.

I meant my comments for all the newly added strings. See my new comment with a suggestion and let me know. This is a great test run for the new best practices and we can update them based on our convo here.

app/web/features/profile/locales/en.json
nabramow: We use a11y all across the codebase. Let's drop this for now as I think there will be a lot of code changes so the labels might change. Why don't you do a final review after it's code approved?

app/web/.cursorrules
nabramow: Yes let's do it in a separate PR or make a ticket to do it later. We also want to add it to CLAUDE.md.

tristanlabelle: Filed https://github.com/Couchers-org/couchers/issues/7663. I'll take this on this week unless you do before.

app/web/features/profile/locales/en.json
tristanlabelle: You're welcome to let me know if you'd like me to check a future iteration. Hopefully the examples above help.

app/backend/src/couchers/migrations/versions/6862ecf6494d_backfill_missing_profile_galleries.py
```diff
@@ -0,0 +1,243 @@
...
+            FROM users
+            WHERE profile_gallery_id IS NULL
+            RETURNING id, owner_user_id
+        )
+        UPDATE users
+        SET profile_gallery_id = new_galleries.id
+        FROM new_galleries
+        WHERE users.id = new_galleries.owner_user_id
+        """
+    )
```
aapeliv: This migration should have been done in a prior PR. Indeed prod shows no such users without profie gallery id

app/backend/src/couchers/migrations/versions/7e65965e02d0_add_has_completed_profile_column.py
```diff
@@ -0,0 +1,149 @@
...
+        ON sv_subquery.id = users.id;
+
+        CREATE INDEX idx_lite_users_geom ON lite_users USING gist (geom);
+        CREATE UNIQUE INDEX uq_lite_users_id ON lite_users(id);
+        CREATE UNIQUE INDEX uq_lite_users_username ON lite_users(username);
+        CREATE INDEX ix_lite_users_id_visible ON lite_users USING hash (id) WHERE is_visible;
+        CREATE INDEX ix_lite_users_username_visible ON lite_users USING hash (username) WHERE is_visible;
+        """
+    )
+    op.drop_column("users", "has_completed_profile")
```
aapeliv: please merge the two migrations into one

app/backend/src/couchers/models/uploads.py
```diff
@@ -113,3 +114,45 @@ class PhotoGalleryItem(Base, kw_only=True):
...
+    """
+    Returns a subquery that selects the first photo (by position) from each photo gallery.
+
+    The subquery has columns: gallery_id, upload_key
+
+    Usage:
+        first_photo = get_first_gallery_photo_subquery()
+        query = select(User).outerjoin(first_photo, first_photo.c.gallery_id == User.profile_gallery_id)
+    """
+    from sqlalchemy import select
```
aapeliv: please move the import to the top

app/backend/src/couchers/models/users.py
```diff
@@ -159,10 +161,15 @@ class User(Base, kw_only=True):
     pronouns: Mapped[str | None] = mapped_column(String, default=None)
     birthdate: Mapped[date] = mapped_column(Date)  # in the timezone of birthplace
 
-    avatar_key: Mapped[str | None] = mapped_column(ForeignKey("uploads.key"), default=None)
-
     # Profile photo gallery for this user (photos about themselves)
-    profile_gallery_id: Mapped[int | None] = mapped_column(ForeignKey("photo_galleries.id"), default=None)
+    # The first photo in the gallery (by position) is used as the avatar
+    profile_gallery_id: Mapped[int | None] = mapped_column(
+        ForeignKey("photo_galleries.id"), nullable=True, default=None
```
aapeliv: I don't think we need to add the nullability, comes from the fact that it's an `int|None`

app/backend/src/couchers/models/users.py
```diff
@@ -579,3 +578,50 @@ class RegionLived(Base, kw_only=True):
     id: Mapped[int] = mapped_column(BigInteger, primary_key=True, init=False)
     user_id: Mapped[int] = mapped_column(ForeignKey("users.id"), index=True)
     region_code: Mapped[str] = mapped_column(ForeignKey("regions.code", deferrable=True))
+
+
+# Event handlers to keep User.has_completed_profile in sync
```
aapeliv: This is not the way to go, it's too hairy. No backend/postgres-side triggers! We need to come up with a better way of doing it.

aapeliv (changes requested):
let's figure out what we're doing with profile completion before proceeding!

app/backend/src/couchers/models/users.py
nabramow: Since you have stronger opinions on how the backend should be done, can you please give clear, actionable guidance on what you expect instead?

I'm getting conflicting comments here.

It doesn't make sense for me to spend more hours on this trying to guess.

app/backend/src/couchers/migrations/versions/6862ecf6494d_backfill_missing_profile_galleries.py
nabramow: Okay I'm going to try and break this up to switch to profile_gallery_id to be used with our existing interface to break this up into chunks.

nabramow:
Converting this to a draft as I will break it up into smaller, iterative PRs so it's less of a beast. That way we can debate one small part at a time and hopefully get more movement.

Leaving it open for now as I might still want to copy over some code.

nabramow:
Okay @aapeliv backend is working well after testing your changes! I pushed some changes to the frontend image drag and drop style and now I'm happy with the frontend too.

aapeliv (approved):
lgtm

PR #7531, opened 2025-12-27
Switch to new i18next implementation

### LINKED ISSUE #7465: The backend i18next implementation does not support nested keys

tristanlabelle:
- Switches to the i18next implementation for https://github.com/Couchers-org/couchers/pull/7475 which supports nested keys and pluralization.
- Removes the "component" argument in favor of '.'-delimited composite keys.
- Rename `get_raw_translation_string` to `localize_string` since "raw" seems to imply that substitutions aren't applied, and it standardizes terminology on "localize" instead of "translation" (both are fine, just for consistency)
- Removes most `get_raw_translation_string` tests since those are now covered at a more unit level in `test_i18next.py`, keeping only one test that validates the strings are loaded correctly.
- Update email template translation to specify the full key.

This is a prerequisite for translating emails, which will make use of nested string keys.

Fixes #7465

Validate any translated string coming from the backend, for example badge descriptions or running test_email.py.

image

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -133,7 +133,7 @@ class CouchersContext:
         else:
             context = cast(grpc.ServicerContext, self._grpc_context)
             # Get the translated error message using the user's language preference
-            error_message = self.get_localized_string(f"errors.{error_message_id}", substitutions)
+            error_message = self.get_localized_string(f"errors.{error_message_id}", substitutions=substitutions)
             context.abort(status_code, error_message)
 
     def set_cookies(self, cookies: list[str]) -> None:
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -2,7 +2,7 @@ from typing import NoReturn, cast
 
 import grpc
 
-from couchers.i18n.i18n import get_raw_translation_string
+from couchers.i18n.i18n import localize_string
 
 
 class NonInteractiveContextException(Exception):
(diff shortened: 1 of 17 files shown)


WouldYouKindly:
lgtm, but needs review by someone who actually knows this part

app/backend/src/couchers/helpers/badges.py
kalvdans, line 11: Please document what `badge_id` is, since it sounds very strange to use an badge ID string as a translation key as there might be a lot of badges around.

app/backend/src/couchers/i18n/i18n.py
```diff
@@ -3,132 +3,62 @@
...
-
-    for lang in all_languages:
-        if lang == "en":
-            continue  # English is already complete
+    # Apply other fallbacks
+    for from_lang_code, to_lang_code in LANGUAGE_FALLBACKS.items():
+        from_lang = i18next.languages_by_code.get(from_lang_code)
+        to_lang = i18next.languages_by_code.get(to_lang_code)
+        if from_lang is not None and to_lang is not None:
+            from_lang.fallback = to_lang.fallback
```
kalvdans: I think there's a bug here, languages should form a linked list, not just short-circuit to the last fallback. If so, please add a unit test that catches this bug so it won't be re-introduced in the future.

```suggestion
            from_lang.fallback = to_lang
```

app/backend/src/couchers/context.py
```diff
@@ -138,7 +133,7 @@ def abort_with_error_code(
         else:
             context = cast(grpc.ServicerContext, self._grpc_context)
             # Get the translated error message using the user's language preference
-            error_message = self.get_localized_string("errors", error_message_id, substitutions=substitutions)
+            error_message = self.get_localized_string(f"errors.{error_message_id}", substitutions=substitutions)
```
kalvdans, line 136: Maybe let the caller add the `errors.` prefix to the message id. (could be a separate PR if we should do this)

app/backend/src/couchers/i18n/i18n.py
tristanlabelle: Good catch. I added a failing test, then fixed the implementation.

app/backend/src/couchers/helpers/badges.py
tristanlabelle: This is existing code. The badge id is a string like `board_member`. I think this code needs a strongly-typed Badge object more than a documentation comment, but that change would be unrelated to my PR.

app/backend/src/couchers/context.py
tristanlabelle: The current design is that all strings are under `errors.`. I'm indifferent to this approach vs another but I think such a change should be its own PR.

tristanlabelle:
@kalvdans I addressed the bug you found and replied to other comments.

kalvdans (approved):

app/backend/src/couchers/helpers/badges.py
tristanlabelle: Followed up in https://github.com/Couchers-org/couchers/pull/7545

PR #7533, opened 2025-12-27
Implement remaining i18n plural rules on the backend

tristanlabelle:
Adds plural rules for remaining supported languages, for string pluralization support, based on [Unicode rules](https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html). See also #7475

Compare with reference implementations on https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/i18n/plurals.py
+++ b/app/backend/src/couchers/i18n/plurals.py
@@ -30,14 +30,26 @@ class PluralRules:
     Implements Unicode CLDR language rules for known languages.
     See https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html
     """
+    _by_lang_family: dict[str, PluralRule] = {} # Populated below
 
     @staticmethod
     def for_language(lang: str) -> PluralRule | None:
         separator_index = lang.find("-")
         lang_family = lang if separator_index == -1 else lang[0:separator_index]
+        return PluralRules._by_lang_family.get(lang_family)
 
-        # Resolve one of the methods below
-        return getattr(PluralRules, lang_family, None)
+    @staticmethod
+    def ca(count: int) -> PluralCategory:
+        return PluralRules.es(count)  # Same as ES
+
+    @staticmethod
+    def cs(count: int) -> PluralCategory:
+        count = abs(count)
+        if count == 1:
+            return PluralCategory.ONE
+        if count >= 2 and count <= 4:
+            return PluralCategory.FEW
+        return PluralCategory.OTHER
 
     @staticmethod
     def de(count: int) -> PluralCategory:
@@ -57,7 +69,6 @@ class PluralRules:
             return PluralCategory.ONE  # 1 manzana
         if count > 0 and count % 1_000_000 == 0:
             return PluralCategory.MANY  # 1000000 de manzanas
-
         return PluralCategory.OTHER  # 2 manzanas
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


app/backend/src/couchers/i18n/plurals.py
```diff
@@ -70,6 +82,47 @@ def fr(count: int) -> PluralCategory:
...
+    def he(count: int) -> PluralCategory:
+        count = abs(count)
+        if count == 1:
+            return PluralCategory.ONE
+        if count == 2:
+            return PluralCategory.TWO
+        return PluralCategory.OTHER
+
+    @staticmethod
+    def hi(count: int) -> PluralCategory:
```
WouldYouKindly, line 101: It wouldn't hurt to name the language in the docstring

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -84,3 +137,26 @@ def ru(count: int) -> PluralCategory:
...
+    @staticmethod
+    def zh(count: int) -> PluralCategory:
+        return PluralCategory.OTHER
+
+
+# Populate PluralRules lookup table
+for method_name in dir(PluralRules):
+    if len(method_name) == 2:
+        method: PluralRule = getattr(PluralRules, method_name)
+        PluralRules._by_lang_family[method_name] = method
```
WouldYouKindly: Can we avoid executing code and just define it statically? LLM can do the initial work for you, and adding a new language is not very tiresome.

aapeliv (commented):
Do we now require backend code changes to add a new language? This does not seem ideal.

app/backend/src/couchers/i18n/plurals.py
```diff
@@ -70,6 +82,47 @@ def fr(count: int) -> PluralCategory:
...
+    def hi(count: int) -> PluralCategory:
+        count = abs(count)
+        if count <= 1:
+            return PluralCategory.ONE
+        return PluralCategory.OTHER
+
+    @staticmethod
+    def hu(count: int) -> PluralCategory:
+        return PluralRules.en(count)  # Same as EN
+
```
kalvdans, line 112: Use a dict that maps from language string (or language family string) to plural categorization function, instead of lots of small two-letter functions.

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: I'm concerned about keeping two lists in sync (method declarations and switch statement or dictionary declaration). I think my previous `getattr` approach was the clearest way to avoid this problem.

tristanlabelle:
> Do we now require backend code changes to add a new language? This does not seem ideal.

@aapeliv Yes but before we didn't support pluralization at all. The alternative is to use the Python babel library and embed a copy of the [Unicode CLDR pluralization rules](https://github.com/unicode-org/cldr-json/blob/48.0.0/cldr-json/cldr-core/supplemental/plurals.json). I think the concern is minimal given that each function takes a minute to implement.

app/backend/src/couchers/i18n/plurals.py
WouldYouKindly: To ensure the two are synced, you can have a test that iterates over what should be, and makes sure it's present in the dictionary. Not a huge deal I guess; I just don't like executing code in global scope.

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: I don't think two-letter functions should be a concern given the context -- they correspond to well-known language codes.

A dict that maps from language string to function forces me to use lambda one-liners which is going to get ugly for Russian and the more complex languages, and won't allow reusing the implementation through delegation. I think my original `getattr` approach was the cleanest solution.

```py
plural_rules = {
    # TODO: Also handle negatives
    "fr": lambda count:
        PluralCategory.ONE if count == 0 or count == 1 else PluralCategory.OTHER,
    "en": lambda count:
        PluralCategory.ONE if count == 1 else PluralCategory.OTHER,
    ...
}
```

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: Done in latest

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: Reverted to getattr approach

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: Reverted to getattr approach. If you think there's a cleaner and more maintainable approach, an illustrating snippet of code would help.

app/backend/src/couchers/i18n/plurals.py
tristanlabelle: > I just don't like executing code in global scope.

I agree, so reverted to getattr.

app/backend/src/couchers/i18n/plurals.py
kalvdans: I wasn't necessarily thinking about lambda functions, but more like this:
```
plural_rules = {
    "en": english_like_plural_rules,
    "hu": english_like_plural_rules,
    ...
}
```

That way, adding a new language wouldn't mean adding new code, just adding an entry to this dict, and the function names don't have any hidden secondary meaning.

tristanlabelle: Unicode doesn't define broad "English-like plural rules" as a concept, each language's rules are independent. I took a shortcut by reusing some functions when I noticed it was the same, but if I had to choose between formalizing broad plural rules (and choosing to call them English-like vs German-like) vs repeating independent code for each language, I'd go for the latter as it's more in the spirit of Unicode. We should generally assume that adding a new language will require adding a new function with its own logic.

If avoiding `getattr` is very important to you, I can rename the functions to the language's name and have a dictionary mapping e.g. `"pt": PluralRules.portuguese_cardinal`, keeping the one function per rule.

kalvdans: Thanks for the elaboration, I'm fine with `getattr`. It could always be changed later :)

kalvdans (approved):

PR #7536, opened 2025-12-28
Refactor bg job registration to be type-safe

WouldYouKindly:
Create a static dict will background jobs: `name: (function, payload type, schedule)` instead of registering them dynamically and adding attributes to functions.

While the current approach certainly works, it confuses mypy and might be non-obvious to new developers.

### DIFF
--- a/app/backend/src/couchers/jobs/definitions.py
+++ b/app/backend/src/couchers/jobs/definitions.py
@@ -53,64 +53,67 @@ class Job:
     schedule: timedelta | None = None
 
 
-# Job registry - maps job names to job definitions
-JOBS: dict[str, Job] = {
-    "handle_notification": Job(handle_notification, jobs_pb2.HandleNotificationPayload),
-    "send_raw_push_notification_v2": Job(send_raw_push_notification_v2, jobs_pb2.SendRawPushNotificationPayloadV2),
-    "handle_email_digests": Job(handle_email_digests, schedule=timedelta(minutes=15)),
-    "generate_message_notifications": Job(
+# Job registry - first create a list of all jobs
+_JOBS_LIST = [
+    Job(handle_notification, jobs_pb2.HandleNotificationPayload),
+    Job(send_raw_push_notification_v2, jobs_pb2.SendRawPushNotificationPayloadV2),
+    Job(handle_email_digests, schedule=timedelta(minutes=15)),
+    Job(
         generate_message_notifications,
         jobs_pb2.GenerateMessageNotificationsPayload,
     ),
-    "generate_reply_notifications": Job(generate_reply_notifications, jobs_pb2.GenerateReplyNotificationsPayload),
-    "generate_create_discussion_notifications": Job(
+    Job(generate_reply_notifications, jobs_pb2.GenerateReplyNotificationsPayload),
+    Job(
         generate_create_discussion_notifications,
         jobs_pb2.GenerateCreateDiscussionNotificationsPayload,
     ),
-    "generate_event_create_notifications": Job(
+    Job(
         generate_event_create_notifications,
         jobs_pb2.GenerateEventCreateNotificationsPayload,
     ),
-    "generate_event_update_notifications": Job(
+    Job(
         generate_event_update_notifications,
         jobs_pb2.GenerateEventUpdateNotificationsPayload,
     ),
-    "generate_event_cancel_notifications": Job(
+    Job(
         generate_event_cancel_notifications,
         jobs_pb2.GenerateEventCancelNotificationsPayload,
     ),
-    "generate_event_delete_notifications": Job(
+    Job(
         generate_event_delete_notifications,
         jobs_pb2.GenerateEventDeleteNotificationsPayload,
     ),
-    "generate_new_blog_post_notifications": Job(
+    Job(
         generate_new_blog_post_notifications,
         jobs_pb2.GenerateNewBlogPostNotificationsPayload,
     ),
-    "refresh_materialized_views": Job(refresh_materialized_views, schedule=timedelta(minutes=5)),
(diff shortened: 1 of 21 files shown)


aapeliv (changes requested):
Conceptual commentary: could you move the jobs definition to a new file please. And while you are at it, what do you think about refactoring the handlers to break them into multiple files. Now there is one huge file with everything in it, I think it would be better to have them split thematically. It has grown too big.

WouldYouKindly:
@aapeliv done! I also made job defintions into a dataclass with defaults to reduce boilerplate a bit.

w.r.t. splitting: not too sure how to best approach it. We already have many job definitions outside of `jobs` package (for example., `couchers/app/backend/src/couchers/notifications/background.py`) Should we move these to `jobs`? Or should we rather move stuff from `jobs` to related packages?

For current content of `jobs.handlers` I guess we could split it into `notifications` and `misc` - not sure if I see more clear-cut areas

(And anyway, I'd like to do that in a separate PR, since this one is about types.)

app/backend/src/couchers/jobs/definitions.py
```diff
@@ -0,0 +1,116 @@
...
+class Job:
+    """Definition of a background job."""
+
+    handler: Callable[..., Any]
+    payload_type: Any = empty_pb2.Empty
+    schedule: timedelta | None = None
+
+
+# Job registry - maps job names to job definitions
+JOBS: dict[str, Job] = {
```
tristanlabelle: We could define a custom annotation to attach information to the method in a statically typed way, which would avoid having to maintain this dictionary, check this out:

[job_annotation.py](https://github.com/user-attachments/files/24368079/job_annotation.py)

It achieves this API shape:

```py
@job
def send_email(payload: int) -> None:
    pass

@job(schedule=timedelta(hours=24))
def purge_login_tokens(payload: bool) -> None:
    pass
```

WouldYouKindly: This is neat, but personally I'm not a huge fan of dynamic registration. The static dictionary allows us to see all jobs in the same place, and it's stupidly simple.

tristanlabelle: I rather favor keeping the related information together and avoiding duplication for easier maintenance, so I see the current code shape that uses custom attributes achieving this better than the separate lookup dictionary.

WouldYouKindly: @tristanlabelle well, this is a matter of opinion: either way it doesn't make *too* much difference in terms of maintainability. But mypy doesn't understand the current setup, so in order to avoid ignoring the whole class of errors, we should do something. We already use a dict as a registry; this PR only changes how the dict is populated.

tristanlabelle: The explicit list of handlers also introduces new classes of errors where a handler might not be listed, might be listed under a mismatching name or might be listed with the incorrect payload (because the two files need to be kept in sync). I'd be happy to sign off on the other improvements in this PR but I think we can fix the typing issue without requiring the duplication that this dictionary introduces. I'd be curious to hear others' takes too.

WouldYouKindly: Please take a look at the latest commit

aapeliv:
I endorse this PR in that I think these are good changes, it's good to clean this up and make it mypy compatible. I'll let you both hash out the details (don't want to intervene on detailed code review). On my quick look the overall implementation strategy seems solid.

app/backend/src/couchers/jobs/definitions.py
```diff
@@ -0,0 +1,119 @@
...
+)
+from couchers.servicers.threads import generate_reply_notifications
+
+
+@dataclass(frozen=True, slots=True)
+class Job:
+    """Definition of a background job."""
+
+    handler: Callable[..., Any]
+    payload_type: Any = empty_pb2.Empty
```
tristanlabelle: We can deduce the payload type from the handler to eliminate the possibility of inconsistency and simplify the list.

```py
    @property
    def payload_type(self) -> type:
        return get_type_hints(self.handler)["payload"] # Or find the first param's name if hard-coding "payload" is an issue
```

app/backend/src/couchers/jobs/definitions.py
```diff
@@ -0,0 +1,119 @@
...
+    Job(
+        send_postal_verification_postcard,
+        jobs_pb2.SendPostalVerificationPostcardPayload,
+    ),
+    Job(check_database_consistency, schedule=timedelta(hours=24)),
+    Job(auto_approve_moderation_queue, schedule=timedelta(seconds=15)),
+]
+
+# Map job names to job definitions
+JOBS: dict[str, Job] = {job.handler.__name__: job for job in _JOBS_LIST}
```
tristanlabelle: Consider exposing

```py
class Job:
    @property
    def name(self) -> str:
        return self.handler.__name__

tristanlabelle (dismissed):
Thanks for iterating, I think the lack of string constants is an improvement.

app/backend/src/couchers/jobs/definitions.py
tristanlabelle: Thanks for iterating!

aapeliv (dismissed):
Lgtm from my side

app/backend/src/couchers/jobs/definitions.py
WouldYouKindly: done, thanks.

app/backend/src/couchers/jobs/definitions.py
WouldYouKindly: Yeah that's very cool. I initially didn't like it because it looks like too much magic, but seems to work! Done.

tristanlabelle (approved):
Sweet!

PR #7537, opened 2025-12-28
Fix running backend tests in Docker

WouldYouKindly:
1. Suggest running tests locally as a preferred approach.
2. Suggest running tests in Docker via `make docker-test` command. This command spins up a database container and then runs tests with `docker run ...` command, instead of running them continuously as part of docker compose. IMO this is a more natural way to run tests, especially if you're a frontend developer wanting to make a small change.

I tried running tests via currently recommended docker compose way and found it very confusing. The output is mingled between three containers and it's difficult to understand what's going on.


WouldYouKindly:
@aapeliv @nabramow please take a look. I changed the recommended way to run tests in Docker - see if you're opposed to it :)

aapeliv (approved):

PR #7539, opened 2025-12-28
Fix warnings and ignore warnings we can't fix

WouldYouKindly:
- Move pytest configuration into pyproject.toml
- Update geoalchemy2 (0.18 -> 0.18.1)
- Fix sqlalchemy warnings
- Ignore warnings originating from geoalchemy2
- Ignore warnings related to use of object outside of session (they seem harmless)
- Fail tests if a new warning is encountered

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -170,7 +170,7 @@ def _sanitized_bytes(proto: Message | None) -> bytes | None:
                 submessage = getattr(message, name)
                 if not submessage:
                     continue
-                if descriptor.label == descriptor.LABEL_REPEATED:
+                if descriptor.is_repeated:
                     for msg in submessage:
                         _sanitize_message(msg)
                 else:
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -7,7 +7,7 @@ from os import getpid
 from threading import get_ident
 from time import perf_counter_ns
 from traceback import format_exception
-from typing import Any, Never, NoReturn, cast
+from typing import Any, NoReturn, cast
 
 import grpc
 import sentry_sdk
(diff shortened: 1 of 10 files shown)


app/backend/src/couchers/interceptors.py
```diff
@@ -170,7 +170,7 @@ def _sanitize_message(message: Message) -> None:
                 submessage = getattr(message, name)
                 if not submessage:
                     continue
-                if descriptor.label == descriptor.LABEL_REPEATED:
+                if descriptor.is_repeated:
```
aapeliv, line 172: Is this currently tested? Please confirm or add a test to verify this works. We have had issues with this code section in the past, just want to double check.

app/backend/src/couchers/materialized_views.py
```diff
@@ -173,25 +179,27 @@ def make_lite_users_selectable(create: bool = False) -> Select[Any]:
...
-            postgresql_where=lite_users_selectable_create.c.is_visible,
+            postgresql_where=lite_users_subquery.c.is_visible,
         ),
         Index(
             "ix_lite_users_username_visible",
-            lite_users_selectable_create.c.username,
+            lite_users_subquery.c.username,
             postgresql_using="hash",
-            postgresql_where=lite_users_selectable_create.c.is_visible,
+            postgresql_where=lite_users_subquery.c.is_visible,
```
aapeliv, line 202: Thanks for fixing these errros

app/backend/src/couchers/interceptors.py
WouldYouKindly: Good point. I added tests for this, and it works for repeated messages. 

Unfortunately, tests don't cover repeated messages with sensitive fields, as we don't have such protos. I considered adding test protos, but it's too much hassle I think. Correct me if I'm wrong.

aapeliv (approved):
lgtm

PR #7546, opened 2025-12-29
Remove unused en-CORP locale.

tristanlabelle:
As far as I could tell, we only had this locale because of one test, `test_i18n.py`'s `test_fallbacks`, which was removed in #7531 since #7475 added a new i18next implementation that included fallback tests without requiring supporting json files. Apparently "en-CORP" is not a valid locale name either, extensions should be under en-x-.

As I merge this, I will delete the corresponding Weblate language.

N/A - This was never a usable language

N/A

**Backend checklist**
(no code changes)
- [ ] All tests pass

**Web frontend checklist**
(no logic changes)
- [ ] All tests pass

**Other**


aapeliv:
This is like the equivalent of pirate english, etc. Point is that those who need to mess around with translations but do not speak other languages fluently enough can test things and make sure stuff is working.

tristanlabelle:
@aapeliv If it's only used for local testing, one can just as well modify another language regardless of how well they speak it?

If you think it's still useful I'll drop the PR, I don't have a strong opinion.

I'm thinking of implementing a pseudolocale for finding unlocalizable strings though. Maybe that'll serve a similar purpose.

aapeliv:
> @aapeliv If it's only used for local testing, one can just as well modify another language regardless of how well they speak it?

It's not for local testing, it's more like end-to-end testing. E.g. when weblate gets screwed, it's helpful to modify something to make sure it gets picked up.

I'm not that strongly opinionated. This should not really show up on the platform, but it is useful often when in particular weblate has issues and we need to make sure stuff works end to end.

tristanlabelle:
@aapeliv Sounds good, I'll drop this.

PR #7558, opened 2025-12-30
Make email translation conditional

tristanlabelle:
Until all email strings are exposed to translation, don't use the translations in prod.

Also removes dead `CONTEXT_TRANSLATION_COMPONENT_KEY` constant.

Is this kind of env var our preferred way to do this? Can/should I target prod vs next?

This only affects the donations email, which I have made translatable. It's covered by a test.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**


kalvdans (approved):
looks fine code-wise!

app/backend/src/couchers/notifications/background.py
```diff
@@ -50,6 +49,11 @@
 
 def _send_email_notification(session: Session, user: User, notification: Notification) -> None:
     rendered = render_notification(user, notification)
+
+    email_lang = "en"
+    if config.get("ENABLE_NOTIFICATION_TRANSLATIONS", False):
```
aapeliv, line 54: @tristanlabelle I know it's not well documented, but I think elsewhere we normally access the value directly, not via `get`. Also to keep it consistent, we have a default system in the config setup, though I prefer not setting defaults and just requiring all config constants to be set explicitly.

Changing this to access directly will also require updating test setup.

PR #7559, opened 2025-12-30
Extract push notification content to its own type

tristanlabelle:
Adds a `PushNotificationContent` type with everything defining the content of the notification (maps to Android and iOS concepts).

This is a first step towards splitting off email and push notification rendering, since they are used by non-overlapping code paths and shouldn't share strings (push notifications should be much more succint).

I added `render_push_notification` to demonstrate the code shape I'm going towards. Currently it still delegates to `render_notification`. I'll extract the implementation from the latter in a follow-up PR.

This impacted tests which expected a flat argument list to the `_push_to_user` function, so I refactored the tests to store a strongly-typed representation of the pushed notifications.

No functional changes, just run tests.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/notifications/push.py
+++ b/app/backend/src/couchers/notifications/push.py
@@ -43,8 +43,8 @@ def push_to_subscription(
     key: str | None = None,
     ttl: int = 0,
 ) -> None:
-    title = config["NOTIFICATION_PREFIX"] + content[: PushNotificationSubscription.MAX_TITLE_LENGTH]
-    body = content.body[: PushNotificationSubscription.MAX_BODY_LENGTH]
+    title = config["NOTIFICATION_PREFIX"] + content.title[: PushNotificationContent.MAX_TITLE_LENGTH]
+    body = content.body[: PushNotificationContent.MAX_BODY_LENGTH]
     icon_url = content.icon_url or urls.icon_url()
     action_url = content.action_url or ""
     queue_job(
--- a/app/backend/src/couchers/notifications/push.py
+++ b/app/backend/src/couchers/notifications/push.py
@@ -12,9 +12,11 @@ from couchers.models import PushNotificationSubscription
 from couchers.notifications.send_raw_push_notification import send_raw_push_notification_v2
 from couchers.proto.internal import jobs_pb2
 
+
 @dataclass(frozen=True, slots=True)
 class PushNotificationContent:
     """Defines the user-visible content of a push notification."""
+
     # Android reference: https://developer.android.com/develop/ui/views/notifications
     # iOS reference: https://developer.apple.com/documentation/usernotifications/unnotificationcontent
 
@@ -24,11 +26,11 @@ class PushNotificationContent:
     title: str
     """A localized title for the notification, this should be a very short string (2-4 words)."""
     body: str
-    """The main text of the notification."""
-    icon_url: str | None = None
-    """A URL to the icon to show in the notification. If None, will use the default app icon."""
+    """The localized text of the notification body."""
     action_url: str | None = None
     """The URL to open when the notification is clicked. If None, will open the app's main URL."""
+    icon_url: str | None = None
+    """A URL to the icon to show in the notification. If None, will use the default app icon."""
 
 
 def push_to_subscription(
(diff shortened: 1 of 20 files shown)


kalvdans (approved):
lgtm!

app/backend/src/couchers/notifications/render.py
```diff
@@ -995,3 +996,16 @@ def get_localized_string(key: str, *, substitutions: dict[str, str | int] | None
         )
 
     raise NotImplementedError(f"Unknown topic-action: {notification.topic}:{notification.action}")
+
+
+def render_push_notification(user: User, notification: Notification) -> PushNotificationContent:
+    email_notification = render_notification(user, notification)
+    if email_notification.push_title is None:
```
WouldYouKindly, line 1003: This should always be a string, and the original comparison was `if not notification.push_title` - so it covered empty strings. Please check

app/backend/src/couchers/notifications/push.py
```diff
@@ -10,29 +13,50 @@
...
 
 
+@dataclass(frozen=True, slots=True)
+class PushNotificationContent:
+    """Defines the user-visible content of a push notification."""
+
+    # Android reference: https://developer.android.com/develop/ui/views/notifications
+    # iOS reference: https://developer.apple.com/documentation/usernotifications/unnotificationcontent
+
+    MAX_TITLE_LENGTH: ClassVar[int] = 500
```
WouldYouKindly, line 23: 500 is a bit more than a 2-4 words - should we make it lower?

app/backend/src/couchers/notifications/push.py
```diff
@@ -10,29 +13,50 @@
...
     topic_action: str,
+    content: PushNotificationContent,
     key: str | None = None,
-    title: str = "",
-    body: str = "",
-    icon: str | None = None,
-    url: str | None = None,
     ttl: int = 0,
 ) -> None:
+    title = config["NOTIFICATION_PREFIX"] + content.title[: PushNotificationContent.MAX_TITLE_LENGTH]
```
WouldYouKindly, line 46: What do you think about raising an error here? Such a long title seems more like a bug. Or maybe send an event to sentry, but still proceed with sending the notification

app/backend/src/couchers/notifications/push.py
tristanlabelle: I think we should let the OS trim as it sees fit, especially since trimming on utf8 code units is problematic, but I preserved the constant from before in case it is required by some API. @aapeliv might have more context from the original PR https://github.com/Couchers-org/couchers/commit/9411e09fc25a5d297deda6080bad42bd293e4d02 . Otherwise I won't change this bit.

app/backend/src/couchers/notifications/render.py
tristanlabelle: Good call. I inspected the code above and neither None nor the empty string case will happen, so this is dead code. I'm going to fast-follow with a PR that extracts the push messages to this function where I'll also remove this condition, as to save us one approval round on this PR.

app/backend/src/couchers/notifications/push.py
tristanlabelle: I don't like having an arbitrary limit, whether for trimming or raising an error, because then it's hard to know why that number was chosen. If Aapeli says we don't need trimming in the other thread, I'll remove the threshold. Otherwise I'd leave it as is since my PR has a different purpose.

(I'm merging but feel free to continue the convo since I'll be touching this code again very soon)

app/backend/src/couchers/notifications/push.py
```diff
@@ -10,29 +13,50 @@
...
+    MAX_BODY_LENGTH: ClassVar[int] = 2000
+
+    title: str
+    """A localized title for the notification, this should be a very short string (2-4 words)."""
+    body: str
+    """The localized text of the notification body."""
+    action_url: str | None = None
+    """The URL to open when the notification is clicked. If None, will open the app's main URL."""
+    icon_url: str | None = None
+    """A URL to the icon to show in the notification. If None, will use the default app icon."""
```
aapeliv, line 33: @tristanlabelle I think here they should be normal comments above the field. Docstrings don't work on fields like this (like they do with functions/classes) as far as I understand it, and this is very counterintuitive to read.

tristanlabelle: (We discussed offline about this being what the VSCode LSP recognizes. Pending conclusion)

WouldYouKindly: Pycharm also recognises this. But it also recognises the more traditional description in a class docstring:

Screenshot 2026-01-01 at 09 50 55

tristanlabelle: Right, but those are painful to write and maintain. I found that there's a PEP that specifies attribute docstrings for tooling: https://peps.python.org/pep-0257/

> String literals occurring elsewhere in Python code may also act as documentation. They are not recognized by the Python bytecode compiler and are not accessible as runtime object attributes (i.e. not assigned to __doc__), but two types of extra docstrings may be extracted by software tools:
String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.
String literals occurring immediately after another docstring are called “additional docstrings”.

WouldYouKindly: Personally I agree that mixing attribute definition and comments makes it more difficult to read.

> those are painful to write and maintain

Maybe; but I think it's more imporant for the code to be easy to read then easy to write :)

PR #7563, opened 2025-12-31
Add tests for some public endpoints

WouldYouKindly:
- Add missing tests for public endpoints
- Fix returning users with visibility=full

### DIFF
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -98,6 +98,9 @@ class CouchersContext:
     def is_logged_in(self) -> bool:
         return self.__logged_in
 
+    def is_logged_out(self) -> bool:
+        return not self.__logged_in
+
     def get_localized_string(self, key: str, *, substitutions: dict[str, str | int] | None = None) -> str:
         """
         Get a localized string using the user's language preference.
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/servicers/api.py
```diff
@@ -954,7 +954,8 @@ def user_model_to_pb(
     # note that this function should work also for banned/deleted users as it's called from Admin.GetUser
     # note that this function is sometimes called by a logged out user, in which case context comes from make_logged_out_context
 
-    if not is_admin_see_ghosts and is_not_visible(session, context.user_id, db_user.id):
+    viewer_user_id = context.user_id if context.is_logged_in() else None
```
WouldYouKindly, line 957: user_id requires the user to be logged in

app/backend/src/couchers/context.py
```diff
@@ -240,7 +243,7 @@ def make_background_user_context(user_id: int) -> CouchersContext:
 
 def make_logged_out_context() -> CouchersContext:
     return CouchersContext(
-        user_id=0,
+        user_id=None,
```
WouldYouKindly, line 246: If user_id=0, CouchersContext considers the user to be logged in

aapeliv (approved):
lgtm

PR #7571, opened 2026-01-01
Split email and push notification rendering logic

tristanlabelle:
Building upon #7559, splits the logic for email and push notification rendering. The combined logic currently somewhat benefits from sharing strings, but it is because our notification strings are designed to be email titles. We are far from the iOS/Android best practices (which I've summarized in a comment). I will rewrite all push notification strings in a subsequent PR before extracting them for localization.

I chose to have one function per notification kind because it allows strongly typing the `data` variable and getting mypy validation. In a few places I shared the implementation between two types of notifications, but I suspect this will go away when I rewrite the strings.

I also renamed `RenderedNotification` to `RenderedEmailNotification` and changed property names since this type is now email-specific. I didn't rename `render.py` for diffability in this PR, but might change it to `render_email.py` later.

The size of the changes means this PR has some risk, though there are no functional changes and test coverage is pretty good.

There is no functionality change, but you can trigger any notification to validate (e.g. receive a message).

N/A

**Backend checklist**
- [x] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/notifications/render_push.py
+++ b/app/backend/src/couchers/notifications/render_push.py
@@ -577,7 +577,7 @@ def _verification_sv_fail(data: notification_data_pb2.VerificationSVFail) -> Pus
 def render_push_notification(user: User, notification: Notification) -> PushNotificationContent:
     data: empty_pb2.Empty = notification.topic_action.data_type.FromString(notification.data)  # type: ignore[attr-defined]
     # Keep topics sorted (action can follow logical ordering)
-    match notification.topic_action:
+    match notification.topic_action.display:
         case "account_deletion:start":
             return _account_deletion_start(data)
         case "account_deletion:complete":
@@ -685,4 +685,4 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
         case "verification:sv_fail":
             return _verification_sv_fail(data)
         case _:
-            raise NotImplementedError(f"Unknown topic-action: {notification.topic}:{notification.action}")
+            raise NotImplementedError(f"Unknown topic-action: {notification.topic_action.display}")
--- a/app/backend/src/couchers/notifications/render_push.py
+++ b/app/backend/src/couchers/notifications/render_push.py
@@ -29,152 +29,152 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
     # Any-typed so it implicitly converts when calling the methods below
     data: Any = notification.topic_action.data_type.FromString(notification.data)  # type: ignore[attr-defined]
     match notification.topic_action:
+        # Using a match statement enable mypy's exhaustiveness checking.
+        # Every case is has its own function so that they can declare different types for "data",
+        # as mypy wouldn't allow that in a single function.
         # Keep topics sorted (actions can follow logical ordering)
         case NotificationTopicAction.account_deletion__start:
-            return _account_deletion_start(data)
+            return _account_deletion__start(data)
         case NotificationTopicAction.account_deletion__complete:
-            return _account_deletion_complete(data)
+            return _account_deletion__complete(data)
         case NotificationTopicAction.account_deletion__recovered:
-            return _account_deletion_recovered(data)
+            return _account_deletion__recovered(data)
         case NotificationTopicAction.activeness__probe:
-            return _activeness_probe(data)
+            return _activeness__probe(data)
         case NotificationTopicAction.api_key__create:
-            return _api_key_create(data)
+            return _api_key__create(data)
         case NotificationTopicAction.badge__add:
-            return _badge_add(data)
+            return _badge__add(data)
         case NotificationTopicAction.badge__remove:
-            return _badge_remove(data)
+            return _badge__remove(data)
         case NotificationTopicAction.birthdate__change:
-            return _birthdate_change(data, user)
+            return _birthdate__change(data, user)
         case NotificationTopicAction.chat__message:
-            return _chat_message(data)
+            return _chat__message(data)
         case NotificationTopicAction.chat__missed_messages:
-            return _chat_missed_messages(data)
+            return _chat__missed_messages(data)
         case NotificationTopicAction.donation__received:
-            return _donation_received(data)
+            return _donation__received(data)
         case NotificationTopicAction.discussion__create:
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/notifications/render_push.py
```diff
@@ -0,0 +1,689 @@
...
+#   - Describe the event, e.g. "Payment Successful"
+#   - <= 30 chars (Android), most important info in first 20 chars
+#   - Title-style capitalization, no ending punctuation
+# Body:
+#   - <= 80 chars (Android), first 40 visible when collapsed
+#   - Sentence-style capitalization with punctuation
+
+
+def render_push_notification(user: User, notification: Notification) -> PushNotificationContent:
+    data: Any = notification.topic_action.data_type.FromString(notification.data)  # type: ignore[attr-defined]
```
WouldYouKindly, line 28: I think if you type `data_type: google.protobuf.message.Message`, you can avoid type: ignore (any Any annotation as well)

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -0,0 +1,689 @@
...
+    # Keep topics sorted (action can follow logical ordering)
+    match notification.topic_action.display:
```
WouldYouKindly: I wonder if it's possible to match on `notification.topic_action` itself? Then you could have mypy ensure that all enum values are covered by the match statement

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -0,0 +1,689 @@
...
+        case "account_deletion:start":
+            return _account_deletion_start(data)
```
aapeliv: What's the rationale for having all these helper functions? This feels like a lot of boilerplate, can we get rid of this whole match somehow? Or inline most of the functions, etc

tristanlabelle: I was a little bit forced into having all these functions because I want `data` to be correctly typed for each notification type, and mypy assumes a variable's type is that of its first assignment, so I can't declare differently typed `data` variables even in different if blocks. However now that I have the functions I think they're kind of nice to read! I agree the switch duplication is unfortunate though.

I'd be happy to replace the switch with building up the function name and looking it up dynamically using getattr on the module, but I've had pushback against that approach before.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: This Any is by design as it will cast implicitly in each of the calls below, whereas `Message` wouldn't. Any better ideas? (no pun intended)

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I can try. I didn't know mypy did exhaustiveness analysis

tristanlabelle: Sweet, it works! I validated the exhaustiveness analysis. See latest.

When I get a chance I might refactor the email path to do the same.

app/backend/src/couchers/notifications/render_push.py
WouldYouKindly: This could've probably be addressed with a generic class, but enums cannot be generic :( oh well

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Given @WouldYouKindly 's trick of switching on the enum to enable mypy exhaustiveness checking, I want to keep both the big match and the functions, to get all static typing benefits. I removed the comments and aligned the function names with the enum names to make it as nonredundant as possible, and added a comment explaining why this code shape.

tristanlabelle:
@aapeliv / @WouldYouKindly , this is ready for your blessing

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -0,0 +1,661 @@
...
+            return _thread__reply(data)
+        case NotificationTopicAction.verification__sv_success:
+            return _verification__sv_success(data)
+        case NotificationTopicAction.verification__sv_fail:
+            return _verification__sv_fail(data)
+        case _:
+            # Enables mypy's exhaustiveness checking for the cases above.
+            assert_never(notification.topic_action)
+
+    raise NotImplementedError(f"Unknown topic-action: {notification.topic}:{notification.action}")
```
WouldYouKindly: This is alredy covered by assert_never, or am I missing something?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -0,0 +1,661 @@
...
+
+
+def _account_deletion__start(data: empty_pb2.Empty) -> PushNotificationContent:
```
WouldYouKindly: nit: We can actually avoid specifying data here, since we're passing it manually

WouldYouKindly: btw it should actually accept AccountDeletionStart protobuf according to NotificationTopicAction enum

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: You're right. I wasn't sure what the runtime behavior of assert_never was but it turns out it does a runtime failure so this is unreachable. I wanted to make sure we handle the "other" case somehow since Python is so dynamic that you could define a different enumerant at runtime etc. I'll remove it.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I added the empty types for consistency at the call site, I don't really mind either way. I'll put the correct type though since there's one.

tristanlabelle: Changed empties to more precise type or removed them.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Removed in latest

WouldYouKindly (dismissed):

WouldYouKindly (approved):

WouldYouKindly (approved):

kalvdans (approved):

PR #7575, opened 2026-01-01
Try MappedAsDataclass for db models

WouldYouKindly:
Right now we can write `Language(code=42, name=datetime(1))` and mypy won't complain. This PR fixes that by using sqlachemy integration with dataclasses.

It (mostly) only affects type checking - we still can create models and do `session.add(...)`, set up relationship etc. We have to modify the model definitions though:

- primary keys generated by the database need `init=False` on a column definition (`id: Mapped[int] = mapped_column(BigInteger, primary_key=True, init=False)`
- columns with server defaults need additional `default=None` for mypy to understand that these don't have to be supplied 
- relationships need `default=None` or `default_factory=list` so that we don't have to supply them
- It's probably better to always pass ids than relationships objects (`Upload(user_id=user.id)` instead of `Upload(user=user)`). Otherwise we'll have to add `default=None` to all foreign keys as well
- A workaround is needed until sqlalchemy 2.1 to allow the above to work properly (luckily there's an "official" workaround available)

This PR is a show case to make sure we're aligned on the approach

### DIFF
--- a/app/backend/src/couchers/models/static.py
+++ b/app/backend/src/couchers/models/static.py
@@ -1,28 +1,28 @@
 from geoalchemy2 import Geometry
 from sqlalchemy import BigInteger, Index, String
-from sqlalchemy.orm import Mapped, mapped_column
+from sqlalchemy.orm import Mapped, MappedAsDataclass, mapped_column
 
 from couchers.models.base import Base, Geom
 
 
-class Language(Base):
+class Language(MappedAsDataclass, Base, kw_only=True):
     """
     Table of allowed languages (a subset of ISO639-3)
     """
 
     __tablename__ = "languages"
 
-    # ISO639-3 language code, in lowercase, e.g. fin, eng
+    # ISO639-3 language code, in lowercase, e.g., fin, eng
     code: Mapped[str] = mapped_column(String(3), primary_key=True)
 
     # the english name
     name: Mapped[str] = mapped_column(String, unique=True)
 
 
-class TimezoneArea(Base):
+class TimezoneArea(MappedAsDataclass, Base, kw_only=True):
     __tablename__ = "timezone_areas"
 
-    id: Mapped[int] = mapped_column(BigInteger, primary_key=True)
+    id: Mapped[int] = mapped_column(BigInteger, primary_key=True, init=False)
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/models/static.py
```diff
@@ -1,29 +1,29 @@
 from geoalchemy2 import Geometry
 from sqlalchemy import BigInteger, Index, String
-from sqlalchemy.orm import Mapped, mapped_column
+from sqlalchemy.orm import Mapped, MappedAsDataclass, mapped_column
 
 from couchers.models.base import Base, Geom
 
 
-class Language(Base):
+class Language(MappedAsDataclass, Base, kw_only=True):
```
aapeliv, line 8: Can we subclass our own Base somehow so we don't need to specify this for every model?

aapeliv (commented):
I'm not a huge fan of the extra code noise, but I think for mypy it can be worth it. Especially because `default` sounds like something meaningful, e.g. it is kind of confusing to have `server_default` and `default` both defined. Are we sure these do what's expected? It is good for all datetimes to be defined by the db so that they are mostly chronological even if the python/db times are off.

Q: Should we just move to SQLAlchemy 2.1 if some of this is fixed there? If it's far enough in being stable enough, I think that's fine.

app/backend/src/couchers/models/static.py
WouldYouKindly: Yes, we can and we will! But doing that creates like 300 mypy errors, so I'll be moving gradually. Also `kw_only` will have to be specified on every model unfortunately :(

WouldYouKindly:
> Q: Should we just move to SQLAlchemy 2.1 if some of this is fixed there? If it's far enough in being stable enough, I think that's fine.

It's not released yet. Should be easy to upgrade when it's out though

> Especially because default sounds like something meaningful, e.g. it is kind of confusing to have server_default and default both defined. Are we sure these do what's expected?

The behaviour is correct: it sets the column to `None`, which gets translated to `null` in the DB, and then the DB substitutes it with it's own default. 

The other option is to specify `init=False` which will prevent us from setting the field when creating an instance. I think I'll go with this one actually

PR #7576, opened 2026-01-01
Add ability to duplicate events

### LINKED ISSUE #6484: Create duplicate event feature
So event hosts don't have to re-fill everything out every time.

* Event host should have additional "Duplicate" option up at the top right
* When they click, it opens create event page with all the data filled in except the date and times
* Probably we can find a way to pass optional data in the props to the CreateEventPage, or via the routing somehow

nabramow:
Closes #6484 

Adds the functionality for event creators to duplicate an existing event (including photo).

I also cleaned up the event buttons a bit: Took away the red and made it outlined (to match the rest of the app negative actions). I changed edit event and made duplicate event icon buttons instead of words to be more tidy and less cluttered.

**TO TEST:**
- You can only duplicate an event if you are the original creator. So create an event if needed, add a photo, then duplicate it.


app/web/features/communities/events/CreateEventPage.tsx
```diff
@@ -29,7 +31,7 @@ const StyledBackButton = styled(HeaderButton)(() => ({
 }));
 
 export default function CreateEventPage() {
-  const { t } = useTranslation([GLOBAL, COMMUNITIES]);
+  const { t } = useTranslation([GLOBAL, COMMUNITIES, PROFILE]);
```
nabramow, line 34: This was somewhat unrelated but fixing a missing translation string I saw in the browser when testing.

app/backend/src/couchers/servicers/events.py
```diff
@@ -440,6 +440,23 @@ def CreateEvent(
         ):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "photo_not_found")
 
+        # Handle photo duplication from source event
+        photo_key_to_use = request.photo_key if request.photo_key != "" else None
+        if request.duplicate_from_event_id > 0:
```
WouldYouKindly: Two points:
- Let's compare with 0, since it cannot be negative. It will be more obvious that we're handling protobuf's (stupid) defaults
- Let's also skip this if photo_key is explicitly provided

app/backend/src/couchers/servicers/events.py
```diff
@@ -440,6 +440,23 @@ def CreateEvent(
...
+            ).scalar_one_or_none()
+
+            if not source_occurrence:
+                context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "source_event_not_found")
+
+            # If no new photo provided, copy from source if user is the creator
+            if not photo_key_to_use and source_occurrence.photo_key:
+                source_event = source_occurrence.event
+                if source_event.creator_user_id == context.user_id:
+                    photo_key_to_use = source_occurrence.photo_key
```
WouldYouKindly: Should we raise an error if creator is different or there's no photo?

app/backend/src/tests/test_events.py
```diff
@@ -2448,3 +2448,154 @@ def test_update_event_should_notify_queues_job():
...
         assert len(jobs) == job_length_before_update + 1
+
+
+def test_DuplicateEvent(db):
+    # Test cases:
+    # 1. Event creator can duplicate their event
+    # 2. Photo is copied from original event when no new photo provided
+    # 3. New photo overrides original when provided
+    # 4. Non-creator cannot duplicate event (permission check on photo)
+    # 5. Duplicated event has new dates but same content
```
WouldYouKindly: Case 5 is untested (and is not implemented in code)

app/backend/src/tests/test_events.py
```diff
@@ -2448,3 +2448,154 @@ def test_update_event_should_notify_queues_job():
...
+    with session_scope() as session:
+        c_id = create_community(session, 0, 2, "Community", [user1], [], None).id
+
+    start_time = now() + timedelta(hours=2)
+    end_time = start_time + timedelta(hours=3)
+
+    # Create original event with photo
+    with events_session(token1) as api:
+        # First create a mock photo upload
+        from couchers.models import Upload
```
WouldYouKindly: please move to the top

app/backend/src/tests/test_events.py
```diff
@@ -2448,3 +2448,154 @@ def test_update_event_should_notify_queues_job():
...
+                timezone="UTC",
+                duplicate_from_event_id=original_event_id,
+            )
+        )
+
+        assert duplicated_with_new_photo.photo_url
+        assert duplicated_with_new_photo.photo_url != original_event.photo_url  # Different photo
+
+    # Test 3: Non-creator trying to duplicate doesn't get photo copied
+    # (photo_key from original event is only accessible to creator)
```
WouldYouKindly: Nothing is stopping the user from saving the photo to their computer and uploading it again. Why do we want to prevent copying the photo?

aapeliv (commented):
If I understand correctly, you are just copying the photo key ID for the new event. Why not just do this on the frontend?

nabramow:
Okay @aapeliv after our chats I removed the backend since it's supposed to work with multiple events using the same photo_key.

app/web/features/communities/events/CreateEventPage.tsx
aapeliv: thanks!

app/web/features/communities/events/CreateEventPage.tsx
```diff
@@ -41,6 +43,31 @@ export default function CreateEventPage() {
...
+
+  // Extract photo key from URL if duplicating and no new photo provided
+  // URL format: https://media.couchers.org/img/full/{key}.jpg
+  const extractPhotoKeyFromUrl = (
+    photoUrl: string | undefined,
+  ): string | undefined => {
+    if (!photoUrl) return undefined;
+    const match = photoUrl.match(/\/([^/]+)\.jpg$/);
+    return match ? match[1] : undefined;
+  };
```
aapeliv: This isn't going to work. The "key" in the photo is the filename, which is different from the upload key...

aapeliv (changes requested):
Let's discuss how to work around this key/filename issue

app/web/features/communities/events/CreateEventPage.tsx
nabramow: I did test it and it does work...

PR #7578, opened 2026-01-01
Include events from communities on dashboard

### LINKED ISSUE #5400: Bug: Dashboard events doesn't show events for users' communities

nabramow:
Closes #5400 

On dashboard now:

* Users should see both events they are attending and events from their communities
* Dashboard pagination changed to page pagination (this was already implemented, we're just using it)
* Changed initial page size to 4 instead of 2


aapeliv (dismissed):
lgtm code-wise!

nabramow:
@aapeliv I think my only question then is for @jesseallhands if we need to change what the backend is returning to only be at the city level. I think probably yes?

jesseallhands (changes requested):
I tried testing this with my profile based in Iceland, which currently has no events, so I joined NYC which has a couple events currently. After joining NYC I went back to the dashboard and refreshed (tried multiple times) but the dashboard still shows no events. So then I changed my profile location to New York city, saved, then went back to the dashboard and refreshed and I still see no events on the dashboard.

nabramow:
> I tried testing this with my profile based in Iceland, which currently has no events, so I joined NYC which has a couple events currently. After joining NYC I went back to the dashboard and refreshed (tried multiple times) but the dashboard still shows no events. So then I changed my profile location to New York city, saved, then went back to the dashboard and refreshed and I still see no events on the dashboard.

Okay I thiiink I fixed this now! At least when I leave and join the NYC community I see the events disappear and reappear now!

Issue #6528, opened 2025-07-23
Implement archiving messages on the frontend


nabramow:
The backend for archiving was done in #5683 and #7681 .

We need to implement this in the frontend.

My thoughts on how to do it for now:

- Add a Archived tab
- Use the EllipsesMenu to add a dropdown that has an "archive" option with MUI "Archive" icon to left and text "Archive". It should show on the bottom right corner of each discussion.
- On select, it adds it to the archive folder
- In archive folder, all items should have same style menu with with "Unarchive" and use MUI "Unarchive" icon
- In the individual message, add "Unarchive" to menu at top right in ellipsis menu

PR #7579, opened 2026-01-01
Step 1 to prep for new message tabs with archive (suggestion)

nabramow:
Part of #6528 

Goals:

- we want to introduce ability to archive messages and host requests
- we want to introduce an "all", "requests" and "archived" tabs to the messages screen so need to query for that eventually
- group chats also need ability to be archived (right now only host req do on backend)
- host requests have `is_host_archived` and `is_surfer_archived` hardcoded into HostRequest, doesn't follow existing pattern we have e.g. `GroupChatSubscription `

My suggestion:
- make `conversation_subscriptions` table with per-user state (step 1, in this PR)
- add to this table when new group chats or host requests are created (step 1, in this PR)
- replace `group_chat_subscriptions` with `conversation_subscriptions` and migrate existing convos there
- migrate existing host requests so all have `conversation_subscriptions` rows
- frontend can then easily query for the different views, all with `conversation_subscriptions` with different filters

Let me know if y'all are happy with this plan!

**TO TEST:**
- Run backend locally, should run without error
- Change `NEXT_PUBLIC_API_URL` in web `.env.development` to the value from `.env.localdev`
- in `app/web`, yarn install then yarn start the frontend
- Create a group chat
- Check the `conversation_subscriptions` table it should have 1 row for each user in the chat
- Existing chats should still work fine

### DIFF
--- a/app/backend/src/couchers/migrations/versions/a0d6686e2ee4_add_conversation_subscriptions_and_type.py
+++ b/app/backend/src/couchers/migrations/versions/a0d6686e2ee4_add_conversation_subscriptions_and_type.py
@@ -1,0 +1,79 @@
+"""add_conversation_subscriptions_and_type
+
+Revision ID: a0d6686e2ee4
+Revises: f8b4ef6e3819
+Create Date: 2026-01-01 19:00:13.067150
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "a0d6686e2ee4"
+down_revision = "f8b4ef6e3819"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # 1. Create ConversationType enum
+    conversationtype = postgresql.ENUM("group_chat", "host_request", name="conversationtype")
+    conversationtype.create(op.get_bind())
+
+    # 2. Add type column to conversations table (nullable first)
+    op.add_column("conversations", sa.Column("type", conversationtype, nullable=True))
+
+    # 3. Populate conversation types for existing data
+    op.execute("""
+        UPDATE conversations SET type = 'group_chat'
+        WHERE id IN (SELECT id FROM group_chats)
+    """)
+    op.execute("""
+        UPDATE conversations SET type = 'host_request'
+        WHERE id IN (SELECT id FROM host_requests)
+    """)
+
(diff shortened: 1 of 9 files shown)


aapeliv (changes requested):
I don't entirely see where you're going with this strategy. I don't support adding another subscription-type table just for adding archivedeness (it is too denormalized). If you mean to replace `GroupChatSubscription` with this, the difficulty will be in making that migration as it's tightly coupled (`GroupChatSubscription` exists because you can leave and re-enter a group chat, all this logic is moot for host requests, so this will be a pain to deal with). So you would need to do that in one PR, otherwise we'll be in half-migrated state that I feel will be hard to get out of. In other words I don't think doing a partial PR here is good.

I suggest instead that we hack this into the backend for the time being. Host reqs already have archivedeness, we can easily add it to group chats as well, and then deal with it at the API layer. We can also hack the all tab by just querying both and returning a merged list (on backend).

The messaging system is due for a full refactor. We should rethink it given our size and migrate everything over. But that will be a sizeable project in and of itself.

nabramow:
> I don't entirely see where you're going with this strategy. I don't support adding another subscription-type table just for adding archivedeness (it is too denormalized). If you mean to replace `GroupChatSubscription` with this, the difficulty will be in making that migration as it's tightly coupled (`GroupChatSubscription` exists because you can leave and re-enter a group chat, all this logic is moot for host requests, so this will be a pain to deal with). So you would need to do that in one PR, otherwise we'll be in half-migrated state that I feel will be hard to get out of. In other words I don't think doing a partial PR here is good.
> 
> I suggest instead that we hack this into the backend for the time being. Host reqs already have archivedeness, we can easily add it to group chats as well, and then deal with it at the API layer. We can also hack the all tab by just querying both and returning a merged list (on backend).
> 
> The messaging system is due for a full refactor. We should rethink it given our size and migrate everything over. But that will be a sizeable project in and of itself.

Yeah I mean the idea was to set up the backend for that refactor. I went to do the tabs today and it was kinda nasty logic on the frontend, so I was hoping to make it cleaner by handling on the backend. So it wouldn't be just for handling archivedness, it would be prep for querying the new tabs in messages to be "all", "requests", "archived" like in the design.

I know "left" and "joined" and "muted" are related to chats only but I figured those could be null for host requests?

For now I was trying to both do the archived and get set up to change the tabs.

I can definitely hack it instead if you prefer but I do have time now so...

PR #7582, opened 2026-01-02
Fix prometheus startup on 3.14 by only executing the code in the main process

WouldYouKindly:


app/backend/src/app.py
```diff
@@ -33,30 +33,6 @@
 )
 logger = logging.getLogger(__name__)
 
-if config["SENTRY_ENABLED"]:
```
tristanlabelle, line 36: Will moving this block prevent the media server and background process from reporting to sentry and Prometheus?

tristanlabelle (approved):
I'm a little muddy about the different modes in which this code runs but no red flag with the changes, and approving in case time sensitive.

app/backend/src/app.py
WouldYouKindly: It probably will. I've merged to fix the staging, but will fix as soon as I can

PR #7587, opened 2026-01-03
Rewrite push notifications following guidelines

### LINKED ISSUE #7589: Rewrite push notification strings to follow iOS/Android guidelines.
The current titles and bodies were written for emails, they are too long and redundant.

tristanlabelle:
Our push notifications piggybacked on email notifications, resulting in text that did not follow Android and iOS guidelines for notifications, including being too long and redundant between titles and bodies. Before exposing these strings for translation, I rewrote them to more closely follow guidelines.

It turns out we need to separate Android and iOS strings. We often have three pieces of info that fits within iOS' title/subtitle/body, whereas two of them need to be combined in Android's title. The casing convention is also different. Note that this PR does not implement separate notifications on iOS and Android, it just prepares the strings but discards the less general iOS strings. Filed #7617.

I wrote a little style guide for future notifications.

I will fix any test failures from new strings once we've agreed on them.

Fixes #7589

Trigger the notification you want to see.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**
- [ ] A maintainer can push commits to my branch
- [ ] A maintainer can merge my PR (you can also merge after approval)

### DIFF
--- a/app/backend/src/couchers/notifications/render_push.py
+++ b/app/backend/src/couchers/notifications/render_push.py
@@ -3,24 +3,19 @@ Renders a Notification model into a localized push notification.
 """
 
 import logging
+from datetime import date
 from typing import Any, assert_never
 
 from couchers import urls
 from couchers.i18n.localize import format_phone_number, localize_date_from_iso, localize_datetime_for_user
 from couchers.models import Notification, NotificationTopicAction, User
 from couchers.notifications.push import PushNotificationContent
-from couchers.proto import api_pb2, events_pb2, notification_data_pb2
+from couchers.proto import api_pb2, notification_data_pb2
+from couchers.utils import to_aware_datetime
 
 logger = logging.getLogger(__name__)
 
-# Best practices for push notification strings (Android/iOS lowest common denominator):
-# Title:
-#   - Describe the event, e.g. "Payment Successful"
-#   - <= 30 chars (Android), most important info in first 20 chars
-#   - Title-style capitalization, no ending punctuation
-# Body:
-#   - <= 80 chars (Android), first 40 visible when collapsed
-#   - Sentence-style capitalization with punctuation
+# See PushNotificationContent's documentation for notification writing guidelines.
 
 
 def render_push_notification(user: User, notification: Notification) -> PushNotificationContent:
@@ -50,7 +45,7 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
         case NotificationTopicAction.chat__message:
             return _chat__message(data)
         case NotificationTopicAction.chat__missed_messages:
-            return _chat__missed_messages()
+            return _chat__missed_messages(data)
         case NotificationTopicAction.donation__received:
             return _donation__received(data)
         case NotificationTopicAction.discussion__create:
@@ -66,17 +61,17 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
         case NotificationTopicAction.event__create_approved:
             return _event__create_approved(data, user)
         case NotificationTopicAction.event__update:
-            return _event__update(data, user)
+            return _event__update(data)
         case NotificationTopicAction.event__invite_organizer:
-            return _event__invite_organizer(data, user)
+            return _event__invite_organizer(data)
         case NotificationTopicAction.event__comment:
-            return _event__comment(data, user)
+            return _event__comment(data)
         case NotificationTopicAction.event__reminder:
             return _event__reminder(data, user)
         case NotificationTopicAction.event__cancel:
-            return _event__cancel(data, user)
+            return _event__cancel(data)
         case NotificationTopicAction.event__delete:
-            return _event__delete(data, user)
+            return _event__delete(data)
         case NotificationTopicAction.friend_request__create:
             return _friend_request__create(data)
         case NotificationTopicAction.friend_request__accept:
@@ -94,13 +89,13 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
         case NotificationTopicAction.host_request__reminder:
             return _host_request__reminder(data)
         case NotificationTopicAction.host_request__accept:
-            return _host_request__accept(data)
+            return _host_request__accept(data, user)
         case NotificationTopicAction.host_request__reject:
-            return _host_request__reject(data)
+            return _host_request__reject(data, user)
         case NotificationTopicAction.host_request__cancel:
-            return _host_request__cancel(data)
+            return _host_request__cancel(data, user)
         case NotificationTopicAction.host_request__confirm:
-            return _host_request__confirm(data)
+            return _host_request__confirm(data, user)
         case NotificationTopicAction.modnote__create:
             return _modnote__create()
         case NotificationTopicAction.onboarding__reminder:
@@ -148,52 +143,58 @@ def _avatar_url_or_default(user: api_pb2.User) -> str:
 
 def _account_deletion__start(data: notification_data_pb2.AccountDeletionStart) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Account deletion initiated",
-        body="Someone initiated the deletion of your Couchers.org account. To delete your account, please follow the link in the email we sent you.",
+        title="Account deletion requested",
+        ios_title="Account Deletion Requested",
+        body="Use the link we emailed you to confirm.",
     )
 
 
 def _account_deletion__complete(data: notification_data_pb2.AccountDeletionComplete) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Your Couchers.org account has been deleted",
-        body=f"You can still undo this by following the link we emailed to you within {data.undelete_days} days.",
+        title="Account deleted",
+        ios_title="Acccount Deleted",
+        body=f"You can restore it within {data.undelete_days} days using the link we emailed you.",
     )
 
 
 def _account_deletion__recovered() -> PushNotificationContent:
     return PushNotificationContent(
-        title="Your Couchers.org account has been recovered!",
-        body="We have recovered your Couchers.org account as per your request! Welcome back!",
+        title="Account restored",
+        ios_title="Account Restored",
+        body="Welcome back!",
     )
 
 
 def _activeness__probe(data: notification_data_pb2.ActivenessProbe) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Are you still open to hosting on Couchers.org?",
-        body="Please log in to confirm your hosting status.",
+        title="Still open to hosting?",
+        ios_title="Still Open to Hosting?",
+        body="Log in to confirm your hosting status.",
     )
 
 
 def _api_key__create(data: notification_data_pb2.ApiKeyCreate) -> PushNotificationContent:
     return PushNotificationContent(
-        title="An API key was created for your account",
+        title="API key created",
+        ios_title="API Key Created",
         body="Details were sent to you via email.",
-        action_url=urls.app_link(),
     )
 
 
 def _badge__add(data: notification_data_pb2.BadgeAdd) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f"The {data.badge_name} badge was added to your profile",
-        body="Check out your profile to see the new badge!",
+        title=f"New profile badge: {data.badge_name}",
+        ios_title="New Profile Badge",
+        body=f"The {data.badge_name} badge was added to your profile.",
         action_url=urls.profile_link(),
     )
 
 
 def _badge__remove(data: notification_data_pb2.BadgeRemove) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f"The {data.badge_name} badge was removed from your profile",
-        body="You can see all your badges on your profile.",
+        title="Profile badge removed",
+        ios_title="Profile Badge Removed",
+        body=f"The {data.badge_name} badge was removed from your profile.",
         action_url=urls.profile_link(),
     )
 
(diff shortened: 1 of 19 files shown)


app/backend/src/couchers/notifications/render_push.py
```diff
@@ -148,60 +182,52 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
...
-        body="We have recovered your Couchers.org account as per your request! Welcome back!",
+        body="Your account was restored. Welcome back!",
     )
 
 
 def _activeness__probe(data: notification_data_pb2.ActivenessProbe) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Are you still open to hosting on Couchers.org?",
-        body="Please log in to confirm your hosting status.",
+        title="Still Hosting?",
```
aapeliv: Still open to hosting?

I think we want to keep it a bit more iffy so we don’t push people too hard into questioning it?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -243,239 +267,221 @@ def _discussion__create(data: notification_data_pb2.DiscussionCreate) -> PushNot
...
-
 def _event__create_any(data: notification_data_pb2.EventCreate, user: User) -> PushNotificationContent:
-    time_display = _get_event_time_display(data.event, user)
+    date_display = v2date(data.event.start_time)
+    time_display = v2date(data.event.start_time)
     return PushNotificationContent(
-        title=f'{data.inviting_user.name} created an event called "{data.event.title}"',
-        body=f"{time_display}\nCreated by {data.inviting_user.name}\n\n{data.event.content}",
+        title="New Event",
+        body=f"{data.inviting_user.name} created {data.event.title} on {date_display} at {time_display}.",
```
aapeliv: Do we want to highlight the event name a bit more? Also understand if it’s against the best practices though

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -243,239 +267,221 @@ def _discussion__create(data: notification_data_pb2.DiscussionCreate) -> PushNot
...
 
 
-def _event__invite_organizer(data: notification_data_pb2.EventInviteOrganizer, user: User) -> PushNotificationContent:
-    time_display = _get_event_time_display(data.event, user)
+def _event__invite_organizer(data: notification_data_pb2.EventInviteOrganizer) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f'{data.inviting_user.name} invited you to co-organize "{data.event.title}"',
-        body=f"{time_display}\nInvited to co-organize by {data.inviting_user.name}\n\n{data.event.content}",
+        title=data.event.title,
+        body=f"{data.inviting_user.name} invited you to co-organize the event.",
```
aapeliv: “… added you as a co-organizer”?

Currently there’s no accept/reject for this

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -243,239 +267,221 @@ def _discussion__create(data: notification_data_pb2.DiscussionCreate) -> PushNot
...
         icon_url=v2avatar(data.user),
     )
 
 
 def _host_request__missed_messages(data: notification_data_pb2.HostRequestMissedMessages) -> PushNotificationContent:
-    their_your = "their" if data.am_host else "your"
     return PushNotificationContent(
-        title=f"{data.user.name} sent you message(s) in {their_your} host request",
-        body="Check the app for more info.",
+        body=f"You have new unseen messages from {data.user.name}.",
```
aapeliv, line 433: Should title here be the name and content something about n missed messages?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -243,239 +267,221 @@ def _discussion__create(data: notification_data_pb2.DiscussionCreate) -> PushNot
...
         action_url=urls.host_request(host_request_id=data.host_request.host_request_id),
         icon_url=v2avatar(data.user),
     )
 
 
 def _host_request__reminder(data: notification_data_pb2.HostRequestReminder) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f"You have a pending host request from {data.surfer.name}!",
-        body="Please respond to the request!",
+        body=f"You have a pending host request from {data.surfer.name}.",
```
aapeliv: We should make this a bit more aggressive. Maybe again use the user’s name as a title (since it relates to the convo?), and body being some kind of action text. Maybe “Accept or decline the request”. Actually i really don’t know what to put here. It might work to appeal to the person “Aapeli is still waiting for your response, please accept or decline the request”

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -484,39 +490,34 @@ def _onboarding__reminder(data: empty_pb2.Empty, key: str, user: User) -> PushNo
...
-        title="Phone verification started",
         body=f"You started phone number verification with the number {v2phone(data.phone)}.",
         action_url=urls.feature_preview_link(),
     )
 
 
 def _phone_number__verify(data: notification_data_pb2.PhoneNumberVerify) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Phone successfully verified",
         body=f"Your phone was successfully verified as {v2phone(data.phone)} on Couchers.org.",
```
aapeliv: Remove couchers.org?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -526,38 +527,38 @@ def _postal_verification__postcard_sent(
...
+        body="You've completed postal verification. Your address is now confirmed.",
         action_url=urls.account_settings_link(),
     )
 
 
 def _postal_verification__failed(data: notification_data_pb2.PostalVerificationFailed) -> PushNotificationContent:
+    title: str | None = "Postal Verification Failed"
     if data.reason == notification_data_pb2.POSTAL_VERIFICATION_FAIL_REASON_CODE_EXPIRED:
-        reason_message = "Your verification code has expired. Codes are valid for 90 days after the postcard is sent. You can start a new verification attempt."
+        body = "Your verification code has expired. Codes are valid for 90 days after the postcard is sent. You can start a new verification attempt."
```
aapeliv: Should we keep the last line shorter? “You can try again”?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -631,31 +632,30 @@ def _thread__reply(data: notification_data_pb2.ThreadReply) -> PushNotificationC
...
         icon_url=v2avatar(data.author),
         action_url=view_link,
     )
 
 
 def _verification__sv_success(data: empty_pb2.Empty) -> PushNotificationContent:
     return PushNotificationContent(
-        title="Strong Verification succeeded",
-        body="You have been verified with Strong Verification! You will now see a tick next to your name on the platform.",
+        body="You have been verified with Strong Verification. You will now see a tick next to your name on the platform.",
```
aapeliv: Not sure we need the second sentence

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -631,31 +632,30 @@ def _thread__reply(data: notification_data_pb2.ThreadReply) -> PushNotificationC
...
         action_url=urls.account_settings_link(),
     )
 
 
 def _verification__sv_fail(data: notification_data_pb2.VerificationSVFail) -> PushNotificationContent:
     if data.reason == notification_data_pb2.SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER:
-        reason_message = "The date of birth or gender on your profile does not match the date of birth or sex on your passport. Please contact the support team to update your date of birth or gender, or if your passport sex does not match your gender identity."
+        reason_message = "The date of birth or sex on your passport does not match your profile. Contact the support team for help."
```
aapeliv: On profile we have gender (identity), on passports they have sex (something about biology). It’s a bit tricky here since it’s quite a verbose message.

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -631,31 +632,30 @@ def _thread__reply(data: notification_data_pb2.ThreadReply) -> PushNotificationC
...
     elif data.reason == notification_data_pb2.SV_FAIL_REASON_NOT_A_PASSPORT:
-        reason_message = "You tried to verify with a document that is not a passport. You can only use a passport for Strong Verification."
+        reason_message = "You used a document other than a passport. You can only use a passport for Strong Verification."
```
aapeliv: This should no longer be a thing, so we could remove this branch. We now tell Our SV provider to only accept passports.

app/backend/src/couchers/notifications/render_push.py
aapeliv: If it’s not too long technically, I would actually just keep the original

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I thought of that too but that n is not currently part of the payload.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I like it

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I didn't like it either, happy to remove it.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Some of those best practices are being defined in this PR, so we can change it. This is what I followed:

```py
# - Is the notification an update on an existing entity that has a name and that the user is aware of? (e.g. an event)
#   - Title: Entity name
```

I used "New Event" because the user doesn't have context on it yet. I'm not sure this is a good approach though. Your suggestion would be:

```py
    return PushNotificationContent(
        title=data.event.title
        body=f"{data.inviting_user.name} created this event on {date_display} at {time_display}.",
```

This does make it more consistent with follow-up notifications about the event.

If we change it that way, we'd logically also change the new discussion one:

```py
def _discussion__create(data: notification_data_pb2.DiscussionCreate) -> PushNotificationContent:
    return PushNotificationContent(
        title=data.discussion.owner_title,
        body=f"{data.author.name} started a discussion: {data.discussion.title}",
```

Here the title is the community or group and the discussion name is in the body. We could instead flip them:

```py
        title=data.discussion.title,
        body=f"{data.author.name} started the discussion in {data.discussion.owner_title}.",
```

Or even treat it the same as a comment and not provide the context:

```py
        title=data.discussion.title,
        body=f"{data.author.name}: {data.discussion.content}.",
```

Curious for your thoughts?

tristanlabelle: But for example "New blog post" might ambiguous if it directly used the blog title and content. It might not be clear it is a blog post.

tristanlabelle: Ok I figured out that we should really use iOS' title and subtitle both, and have longer titles on Android. I've updated all messages and the guidance accordingly see latest.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I kept the first sentence and changed the second to "Contact our support team for help."

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I think this code shouldn't know about the fact that the value is not used anymore. If we refactor to remove the enumerant, then sure.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I shortened to "You can request a new postcard."

tristanlabelle:
@aapeliv This is finally ready for re-review

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -144,334 +139,380 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
...
         icon_url=v2avatar(data.user),
     )
 
 
 def _host_request__reminder(data: notification_data_pb2.HostRequestReminder) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f"You have a pending host request from {data.surfer.name}!",
-        body="Please respond to the request!",
+        title=f"Pending request by {data.surfer.name}",
+        ios_title="Host Request Pending",
```
aapeliv, line 442: Missing subtitle?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -144,334 +139,380 @@ def render_push_notification(user: User, notification: Notification) -> PushNoti
...
 
-def _host_request__confirm(data: notification_data_pb2.HostRequestConfirm) -> PushNotificationContent:
+def _host_request__confirm(data: notification_data_pb2.HostRequestConfirm, user: User) -> PushNotificationContent:
+    date = v2date(data.host_request.from_date, user)
     return PushNotificationContent(
-        title=f"{data.surfer.name} confirmed their host request",
-        body="Check the app for more info.",
+        title=f"{data.surfer.name} confirmed their stay",
+        ios_title=data.surfer.name,
+        ios_subtitle="Stay Confirmed",
```
aapeliv: Host Request Confirmed?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -522,37 +568,42 @@ def _postal_verification__postcard_sent(
...
+        ios_title="Postal Verification Failed",
+        body=body,
         action_url=urls.account_settings_link(),
     )
 
 
 def _reference__receive_friend(data: notification_data_pb2.ReferenceReceiveFriend) -> PushNotificationContent:
     return PushNotificationContent(
-        title=f"You've received a friend reference from {data.from_user.name}!",
+        title=f"New reference from {data.from_user.name}",
```
aapeliv: friend reference?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -598,8 +649,9 @@ def _reference__reminder(data: notification_data_pb2.ReferenceReminder, referenc
         host_request_id=str(data.host_request_id),
     )
     return PushNotificationContent(
-        title=f"You have {data.days_left} days to write a reference for {data.other_user.name}!",
-        body="It's a nice gesture to write references and helps us build a community together! References will become visible 2 weeks after the stay, or when you've both written a reference for each other, whichever happens first.",
+        title="Write your reference",
+        ios_title="Write Your Reference",
```
aapeliv, line 656: should there be a subtitle thaht's the user's name? and on android have the name in titel?

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -616,42 +668,49 @@ def _reference__reminder_hosted(data: notification_data_pb2.ReferenceReminder) -
...
         icon_url=v2avatar(data.author),
         action_url=view_link,
     )
 
 
 def _verification__sv_success() -> PushNotificationContent:
     return PushNotificationContent(
-        title="Strong Verification succeeded",
-        body="You have been verified with Strong Verification! You will now see a tick next to your name on the platform.",
+        title="Strong verification completed",
```
aapeliv: should capitalize verification here, the thing is called "Strong Verification"

app/backend/src/couchers/notifications/render_push.py
```diff
@@ -616,42 +668,49 @@ def _reference__reminder_hosted(data: notification_data_pb2.ReferenceReminder) -
...
+            "You used a document other than a passport. You can only use a passport for Strong Verification."
+        )
     elif data.reason == notification_data_pb2.SV_FAIL_REASON_DUPLICATE:
-        reason_message = "You tried to verify with a passport that has already been used for verification. Please use another passport."
+        reason_message = "You used a passport that has already been used for verification. Please use another passport."
     else:
         raise Exception("Shouldn't get here")
     return PushNotificationContent(
-        title="Strong Verification failed",
+        title="Strong verification failed",
```
aapeliv: ditto

app/backend/src/couchers/notifications/render_push.py
aapeliv: is the old version too long for the body? we do wish to communicate to people that there is an option to have an exception if the gender and sex correspondence does not match?

aapeliv (changes requested):

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Added it so I can merge, but I think the kind will be clear from context (did I just host them?), and even if it's not you'll see it when following the notification, so I was going for the shorter text.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Updated so I can merge and kinda unrelated to this PR, but I have to pause and think when I see "host request confirmed". i.e. they confirmed that I approved that they requested me to host them, oof!

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: I added it since I don't have too strong an opinion, but my thinking was that it would be immediately redundant with the start of the body, whereas Android has more space to surface info up front.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: It definitely exceeds recommendations but maybe it's fine in practice? Reverted so we can merge.

app/backend/src/couchers/notifications/render_push.py
tristanlabelle: Added.

tristanlabelle:
@aapeliv , I took all your suggestions for expediency since it seems we're close enough now that we can follow-up separately as needed. I'd appreciate your thoughts on the threads though. There was also a bit of a gnarly rebase but I double-checked and my merge seems all right.

app/backend/src/couchers/notifications/render_push.py
aapeliv: I think it was just a consistency question: the other had it in the host request changes series

app/backend/src/couchers/notifications/render_push.py
aapeliv: Yes just felt it was inconsistent between android/ios

aapeliv (approved):
lgtm. most of last round of comments were about consistency, rather than style/preference! feel free to proceed as you wish; either way this is orders of magnitude better than the prior

tristanlabelle:
@aapeliv Thanks for the review. I'm happy with where we landed and more than ready to take off the copyrighter hat haha. I'll turn to extracting the strings for localization.

PR #7592, opened 2026-01-03
Let relationship targets be inferred from type annotations

WouldYouKindly:
- Let relationship targets be inferred from type annotations
- Update slqalchemy 2.0.43 -> 2.0.45
- Remove unnecessary workaround for User.pending_activeness_probe


tristanlabelle (approved):
I'm all for this!

PR #7593, opened 2026-01-03
Always flush changes when adding a new model to a session

WouldYouKindly:
This prevents cases when we created an instance, added it to a session, but haven't flushed it, and then try to use it's databased-generated properties (like id)

## Why we need it 

This is a part of migration to `MappedAsDataclass` for sqlalchemy models.  `MappedAsDataclass`  gives us a type-safe `__init__`; we can no longer forget to pass a required field. But this also means that for relationships we have to pass both foreign key column and the related object, e.g.:

```python
class PhotoGalleryItem(Base):
    __tablename__ = "photo_gallery_items"
    
    gallery_id: Mapped[int] = mapped_column(ForeignKey("photo_galleries.id"), index=True)
    gallery: Mapped[PhotoGallery] = relationship(back_populates="photos")

PhotoGalleryItem(gallery=gallery)  # type error, gallery_id not passed
PhotoGalleryItem(gallery_id=1)  # type error, gallery not passed
```

This is annoying, and I propose we standardise on always passing only the foreign key column, since it's always available. Then it will look like this (note the `init=False` for `gallery`)

```python
class PhotoGalleryItem(Base):
    __tablename__ = "photo_gallery_items"
    
    gallery_id: Mapped[int] = mapped_column(ForeignKey("photo_galleries.id"), index=True)
    gallery: Mapped[PhotoGallery] = relationship(init=False, back_populates="photos")

PhotoGalleryItem(gallery=gallery)  # type error, no such argument
PhotoGalleryItem(gallery_id=1)  # good!
```

The only problem with this approach is that we have to remember to flush the gallery object. Consider:

```python
gallery = PhotoGallery(...)
session.add(gallery)
item = PhotoGalleryItem(gallery_id=gallery.id)  # whooops, gallery.id is None!
```

This is where this PR comes in: by using our own `add` instead of `session.add` we can never forget to flush. 

Ideally, I think that we should use our own wrapper around `Session` everywhere:

```python
class DB:
    _session: Session
    def add(self, obj):
        self._session.add(obj)
        self._session.flush()
        return obj
    
     # Provided so that we don't have to directly access session
     def execute(self, stmt):
         return self._session.execute(stmt)
         
def AddUser(
    self, request: AddUserReq, context: CouchersContext, db: DB
) -> empty_pb2.Empty:
    
    user = db.add(User(...))
```

### DIFF
--- a/app/backend/src/couchers/db.py
+++ b/app/backend/src/couchers/db.py
@@ -20,6 +20,7 @@ from couchers.config import config
 from couchers.constants import SERVER_THREADS, WORKER_THREADS
 from couchers.context import CouchersContext
 from couchers.models import (
+    Base,
     Cluster,
     ClusterRole,
     ClusterSubscription,
@@ -267,3 +268,8 @@ def timezone_at_coordinate(session: Session, geom: WKBElement) -> str | None:
         select(TimezoneArea.tzid).where(func.ST_Contains(TimezoneArea.geom, geom))
     ).scalar_one_or_none()
     return tzid
+
+
+def add(session: Session, obj: Base) -> None:
+    session.add(obj)
+    session.flush()
--- a/app/backend/src/couchers/helpers/badges.py
+++ b/app/backend/src/couchers/helpers/badges.py
@@ -2,6 +2,7 @@ from sqlalchemy.orm import Session
 from sqlalchemy.sql import delete
 
 from couchers.context import make_background_user_context
+from couchers.db import add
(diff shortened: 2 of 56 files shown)


WouldYouKindly:
Discussed it with Aapeli, and decided it's not worth the fuss. If one forgets to flush, it will be caught by tests most of the time.

Also, if we some day start using repositories to create objects, this can be solved on repository level.

PR #7598, opened 2026-01-03
Add docker compose override for local backend + update readme

tristanlabelle:
Running the backend locally currently requires me to locally edit all of:

- `.env.development`: Overwrite with `.env.localdev`
- `backend.dev.env`: Allow the backend to talk to other containerized services
  - `DATABASE_CONNECTION_STRING`: Replace `postgres:6545` with `localhost:6545`
  - `SMTP_HOST`: Replace `maildev` with `localhost`
  - `OPENTELEMETRY_ENDPOINT`: Replace `jaeger` with `localhost`
- `docker-compose.yml`: Expose containerized services to the backend
  - `proxy`: Add `extra_hosts: { backend: 172.17.0.1 }`
  - `media`: Add `extra_hosts: { backend: 172.17.0.1 }`
  - `jaeger`: Add `"4317:4317"` to `ports`

With this override file, I don't have to edit `docker-compose.yml` and can just run `docker compose --file docker-compose.local-backend.yml up --build`. I also updated the readme with the instructions.

Note: `host.docker.internal` is not available in my setup (seems like it's not by default) whereas `172.17.0.1` works, but I added a mechanism to override the backend host as necessary.

`docker compose --file docker-compose.local-backend.yml up --build`

Still need `.env.development` and `backend.dev.env` edits.

**Backend checklist**
N/A
- [ ] All tests pass

**Other**


app/docker-compose.local-backend.yml
```diff
@@ -0,0 +1,34 @@
...
+      file: docker-compose.yml
+      service: maildev
+  media:
+    extends:
+      file: docker-compose.yml
+      service: media
+    # Tell it to use the local backend
+    depends_on: !reset []
+    extra_hosts:
+      backend: ${BACKEND_HOST:-172.17.0.1}
```
aapeliv, line 23: I'm not sure if this IP address is universal. I think it depends on docker setup. At least on some OSs I have used `host.docker.internal` in the past

app/docker-compose.local-backend.yml
```diff
@@ -0,0 +1,34 @@
+# Runs all services except the backend, configuring them to talk to a backend running outside of docker.
+# Use the BACKEND_HOST variable to override the address of the backend as visible from within containers.
+# Run with: docker compose --file docker-compose.local-backend.yml up --build
+services:
+  jaeger:
+    extends:
```
WouldYouKindly, line 6: Wow, didn't know about that!!

app/docker-compose.local-backend.yml
WouldYouKindly: yeah I think host.docker.internal is only for linux

WouldYouKindly:
@tristanlabelle let's add it to backend readme?

Also it would be grand to have a make command that runs docker compose and then runs `python src/app.py` (if you're feeling generous :D)

app/docker-compose.local-backend.yml
tristanlabelle: WSL2 is supposed to be linux but host.docker.internal does not resolve for me. This is why I added a variable for overriding as necessary.

app/docker-compose.local-backend.yml
tristanlabelle: Thanks to Gemini. This is not super well designed imo. There's also an include syntax for the whole base file but then you can't override stuff, or you can specify a base file and it's overrides on the command line. This was the least annoying form.

tristanlabelle:
@aapeliv + @WouldYouKindly , I significantly updated the backend readme with instructions and need a new approval.

WouldYouKindly (approved):

PR #7599, opened 2026-01-04
Run mypy on tests

WouldYouKindly:
Run mypy on `src/tests`. The main benefit is safer refactoring: e.g., you change something, run mypy and see type errors in tests, instead of waiting them to run (or worse, fail in CI). I disabled the most annoying errors (so that type annotating all tests functions is NOT requried), so this doesn't add much work to writing tests.

### DIFF
--- a/app/backend/pyproject.toml
+++ b/app/backend/pyproject.toml
@@ -130,6 +130,18 @@ module = ["http_ece", "py_vapid", "luhn", "sqlalchemy_utils.*", "statsig_python_
 # These libraries don't have type stubs or py.typed markers
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["tests.*"]
+# Test files don't require strict type checking
+disallow_untyped_defs = false
+disallow_untyped_calls = false
+disallow_incomplete_defs = false
+
+[[tool.mypy.overrides]]
+module = ["tests.fixtures.sessions"]
+# Mock gRPC infrastructure intentionally violates type contracts
+disable_error_code = ["arg-type"]
+
 
 # Coverage
 [tool.coverage.run]
--- a/app/backend/src/tests/fixtures/misc.py
+++ b/app/backend/src/tests/fixtures/misc.py
@@ -1,6 +1,10 @@
+from collections.abc import Generator
 from contextlib import contextmanager
 from dataclasses import dataclass
-from unittest.mock import patch
+from typing import Any
+from unittest.mock import Mock, patch
+
+from sqlalchemy.orm import Session
 
 from couchers.jobs.worker import process_job
 from couchers.models import User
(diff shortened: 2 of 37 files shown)


app/backend/src/tests/fixtures/misc.py
```diff
@@ -80,18 +84,18 @@ def get_for_user(
...
         for kwarg in kwargs:
             assert hasattr(push, kwarg), f"Push notification {user_id=}, {ix=} missing field '{kwarg}'"
             assert getattr(push, kwarg) == kwargs[kwarg], (
                 f"Push notification {user_id=}, {ix=} mismatch in field '{kwarg}', expected '{kwargs[kwarg]}' but got '{getattr(push, kwarg)}'"
             )
 
-    def assert_user_has_single_matching(self, user_id, **kwargs):
+    def assert_user_has_single_matching(self, user_id: int, **kwargs: Any) -> None:
         self.assert_user_has_count(user_id, 1)
         self.assert_user_push_matches_fields(user_id, ix=0, **kwargs)
```
tristanlabelle, line 100: Oops I was supposed to have deleted those. I'll open a PR, or feel free to do it here. https://github.com/Couchers-org/couchers/pull/7613

app/backend/src/tests/test_auth.py
```diff
@@ -254,7 +255,7 @@ def _quick_signup() -> int:
     sesh, uid = get_session_cookie_tokens(metadata_interceptor)
     assert sesh == token
 
-    return res.auth_res.user_id
+    return cast(int, res.auth_res.user_id)
```
tristanlabelle, line 258: Any reason for the cast?

Image

app/backend/pyproject.toml
```diff
@@ -130,6 +130,18 @@ module = ["http_ece", "py_vapid", "luhn", "sqlalchemy_utils.*", "statsig_python_
 # These libraries don't have type stubs or py.typed markers
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+module = ["tests.*"]
+# Test files don't require strict type checking
+disallow_untyped_defs = false
+disallow_untyped_calls = false
+disallow_incomplete_defs = false
```
tristanlabelle, line 138: This is fine for scoping this PR but I'd be in favor of eventually removing them and fully typing our tests like the rest of the codebase.

tristanlabelle (approved):

app/backend/src/tests/test_auth.py
WouldYouKindly: Not sure, but mypy complains about it.

app/backend/src/tests/fixtures/misc.py
WouldYouKindly: sorry!

app/backend/pyproject.toml
WouldYouKindly: Thing thing is, you'd have to annotate all test functions with `-> None`, and also all fixtures (including `db: None` in almost all tests). So I'm -1 on removing these.

We should fully type conftest and fixtures though that would have much more bang for the buck.

PR #7615, opened 2026-01-05
Start using MappedAsDataclass for sqlalchemy models

WouldYouKindly:
- Add MappedAsDataclass to Base, and add `init=False` for all models so that they don't have a fully typed init. This is to avoid too much disruption
- Remove `init=False` for a subset of models. This enables typed init for them. 
- Fix the resulting errors for that subset of models

- Run mypy on app.py and dummy_data.py

## Overall approach to typing sqlalchemy models
- We disallow passing a related object when creating a model. No `Gallery(user=user)`, only `Gallery(user_id=user.id)`. 
- We disallow passing attributes that are created by the DB (`id`, `created` etc). If one needs to override them, one has to first create a model, and then set an attribute: `user = User(...); user.id = 42; session.add(user)`. 
- If a field is nullable, it has to have a `default=None` on a `mapped_column` so that mypy knows we don't have to pass it.

### DIFF
--- a/app/backend/Makefile
+++ b/app/backend/Makefile
@@ -26,7 +26,7 @@ format:
 fmt: format
 
 mypy:
-	uv run mypy src/couchers src/tests
+	uv run mypy src
 
 # Downgrade the local database to the previous migration. Good if you ran a migration on a branch and will run the backend on another branch after.
 # Usage:
--- a/app/backend/src/couchers/models/users.py
+++ b/app/backend/src/couchers/models/users.py
@@ -95,7 +95,7 @@ class ProfilePublicVisibility(enum.Enum):
     full = enum.auto()
 
 
-class User(Base, init=False, kw_only=True):
+class User(Base, kw_only=True):
     """
     Basic user and profile details
     """
@@ -109,9 +109,9 @@ class User(Base, init=False, kw_only=True):
     # stored in libsodium hash format, can be null for email login
     hashed_password: Mapped[bytes] = mapped_column(Binary)
     # phone number in E.164 format with leading +, for example "+46701740605"
-    phone: Mapped[str | None] = mapped_column(String, nullable=True, server_default=expression.null())
+    phone: Mapped[str | None] = mapped_column(String, default=None, server_default=expression.null())
     # language preference -- defaults to empty string
-    ui_language_preference: Mapped[str | None] = mapped_column(String, nullable=True, server_default="")
+    ui_language_preference: Mapped[str | None] = mapped_column(String, default=None, server_default="")
 
     # timezones should always be UTC
     ## location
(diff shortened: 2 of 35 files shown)


WouldYouKindly:
@aapeliv @tristanlabelle ready to recieve your blessings

app/backend/src/couchers/utils.py
```diff
@@ -415,3 +418,10 @@ def not_none[T](x: T | None) -> T:
     if x is None:
         raise ValueError("Expected a value but got None")
     return x
+
+
+def is_geom(x: Geom | None) -> Geom:
+    """not_none does not work with unions."""
```
tristanlabelle, line 424: How does `not_none` not work if the body is the same? 🤔

app/backend/src/couchers/models/users.py
```diff
@@ -95,160 +95,166 @@ class ProfilePublicVisibility(enum.Enum):
...
     parking_details: Mapped[ParkingDetails | None] = mapped_column(
-        Enum(ParkingDetails), nullable=True
+        Enum(ParkingDetails), default=None
     )  # CommonMark without images
-    camping_ok: Mapped[bool | None] = mapped_column(Boolean, nullable=True)
+    camping_ok: Mapped[bool | None] = mapped_column(Boolean, default=None)
 
     accepted_tos: Mapped[int] = mapped_column(Integer, default=0)
-    accepted_community_guidelines: Mapped[int] = mapped_column(Integer, server_default="0")
+    accepted_community_guidelines: Mapped[int] = mapped_column(Integer, server_default="0", init=False)
```
tristanlabelle, line 231: Why does this int require init=False?

app/backend/src/couchers/models/users.py
```diff
@@ -95,160 +95,166 @@ class ProfilePublicVisibility(enum.Enum):
...
+    additional_information: Mapped[str | None] = mapped_column(String, default=None)  # CommonMark without images
 
-    is_banned: Mapped[bool] = mapped_column(Boolean, server_default=expression.false())
-    is_deleted: Mapped[bool] = mapped_column(Boolean, server_default=expression.false())
-    is_superuser: Mapped[bool] = mapped_column(Boolean, server_default=expression.false())
-    is_editor: Mapped[bool] = mapped_column(Boolean, server_default=expression.false())
+    is_banned: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
+    is_deleted: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
+    is_superuser: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
+    is_editor: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
```
tristanlabelle, line 188: Why do these bools require init=False?

tristanlabelle (approved):
If I understand well, this is building towards strongly-typed model object initializers, which I'm all for. Will you be progressively removing the class-level init=False while migrating the code afterwards?

app/backend/src/couchers/utils.py
WouldYouKindly: mypy doesn't understand it that the resulting type is `Geom`. It works at runtime obviously :))

app/backend/src/couchers/models/users.py
WouldYouKindly: Because we don't pass this to init and rely on `server_default`. In general, if you have a column with `server_default`, you have three options:

1. Pass a value explicitly (defeats the purpose of a default)
2. `default=None` which renders `NULL`, which gets replacted with the database-level default. This allows you to pass a different value to `__init__` if you wish, while not requiring you to pass it every time.
3. `init=False` which disallows passing it to `__init__`. You can still change the attribute later, when you already created an instance (but before saving to the DB)

I chose the latter, because having both `default` and `server_default` looks weird.

app/backend/src/couchers/models/users.py
WouldYouKindly: see my response above

WouldYouKindly:
> If I understand well, this is building towards strongly-typed model object initializers, which I'm all for. Will you be progressively removing the class-level init=False while migrating the code afterwards?

Yes! In fact, I've already prepared a PR for the rest of the models :)

WouldYouKindly:
@aapeliv don't want to merge this without your approval just in case you have a strong opinion here

app/backend/src/couchers/models/users.py
tristanlabelle: Got it, makes sense. When do we prefer having the database dictate the default vs the application? I don't have a sense for best practices there.

WouldYouKindly: Well, this codebase tends to put all the defaults it can into the database. I don't think it matters too much, since we only have one app using the database.

aapeliv: I am a fan of keeping defaults for columns in the DB unless need requires otherwise. E.g. sometimes someone messes around on PSQL directly (e.g. locally/on stage, hopefully not on prod), so it helps to have everything correctly there.

app/backend/src/couchers/servicers/auth.py
```diff
@@ -324,22 +325,23 @@ def SignupFlow(
...
+                geom_radius=not_none(flow.geom_radius),
+                accepted_tos=not_none(flow.accepted_tos),
                 last_onboarding_email_sent=func.now(),
-                opt_out_of_newsletter=flow.opt_out_of_newsletter,
                 invite_code_id=flow.invite_code_id,
             )
 
+            user.accepted_community_guidelines = flow.accepted_community_guidelines
+            user.onboarding_emails_sent = 1
+            user.opt_out_of_newsletter = not_none(flow.opt_out_of_newsletter)
```
aapeliv, line 343: Why did you pull these out?

app/backend/src/tests/fixtures/db.py
```diff
@@ -158,6 +157,46 @@ def autocommit_engine(url: str):
     engine.dispose()
 
 
+def make_user(**kwargs: Any) -> User:
```
aapeliv, line 160: I would prefer we don't extract this out, I would prefer as small a surface here as possible: it makes life easier to not have people start to use these arbitrarily, etc. If it's only the models tests, maybe you can just `generate_user` and the modify its last active etc after?

app/backend/Makefile
```diff
@@ -26,7 +26,7 @@ format:
 fmt: format
 
 mypy:
-	uv run mypy src/couchers src/tests
+	uv run mypy src
```
aapeliv, line 29: 🎉 🎉 🎉

aapeliv (approved):
lgtm, optional comments from my side

app/backend/src/couchers/servicers/auth.py
WouldYouKindly: This is a downside of using MappedAsDataclass. These fields have server defaults, but mypy doesn't understand that it makes them optional: it shouts at you if you don't pass it. To avoid this, I added `init=False` to these fields. Now we cannot pass them, and have to set them on the created instance.

app/backend/src/tests/fixtures/db.py
WouldYouKindly: I extracted this to separate model creation from DB interations, for the tests that do not need a database.

> it makes life easier to not have people start to use these arbitrarily, etc

Not sure what you mean here. What's the harm of having a factory function?

PR #7627, opened 2026-01-07
Backend: Strongly type NotificationTopicAction references

tristanlabelle:
Useful when looking where NotificationTopicActions are used, and to avoid typos.

Run existing tests

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -99,6 +99,7 @@ from couchers.models import (
     UserBadge,
     Volunteer,
 )
+from couchers.models.notifications import NotificationTopicAction
 from couchers.notifications.expo_api import get_expo_push_receipts
 from couchers.notifications.notify import notify
 from couchers.postal.postcard_service import send_postcard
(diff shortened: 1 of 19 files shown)


app/backend/src/couchers/notifications/render.py
```diff
@@ -138,7 +138,7 @@ def get_localized_string(key: str, *, substitutions: dict[str, str | int] | None
                 },
                 topic_action_unsubscribe_text="Pending host request reminders",
             )
-    elif notification.topic_action.display == "password:change":
+    elif notification.topic_action == NotificationTopicAction.password__change:
```
WouldYouKindly, line 141: Can this be a match statement instead?

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -1137,7 +1145,7 @@ def send_event_reminders(payload: empty_pb2.Empty) -> None:
                 notify(
                     session,
                     user_id=user.id,
-                    topic_action="event:reminder",
+                    topic_action=NotificationTopicAction.event__reminder.display,
```
WouldYouKindly: Can we pass `NotificationTopicAction.event__reminder` and extract display inside `notify` ?

app/backend/src/couchers/notifications/render.py
tristanlabelle: Yes, it's on my list (what made me think of doing this), but it's a larger refactoring since right now we sometimes have shared code for different actions of the same topic. Stay tuned!

app/backend/src/couchers/jobs/handlers.py
tristanlabelle: Good call! Updated in latest

WouldYouKindly (approved):

PR #7628, opened 2026-01-07
Mapped as dataclass 2

WouldYouKindly:
Continuation of https://github.com/Couchers-org/couchers/pull/7615. This PR adds a type-safe init to all sqlalchemy models (and finishes type annotating our codebase!)

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -100,7 +100,7 @@ def _try_get_and_update_user_details(
         if not result:
             return None
         else:
-            user, user_session, user_activity = result
+            user, user_session, user_activity = result._tuple()
 
             # update user last active time if it's been a while
             if now() - user.last_active > timedelta(minutes=5):
@@ -187,7 +187,7 @@ def _sanitized_bytes(proto: Message | None) -> bytes | None:
 def _store_log(
     *,
     method: str,
-    status_code: grpc.StatusCode | None,
+    status_code: str | None,
     duration: float,
     user_id: int | None,
     is_api_key: bool,
@@ -358,7 +358,13 @@ class CouchersMiddlewareInterceptor(grpc.ServerInterceptor):
                 except Exception as e:
                     finished = perf_counter_ns()
                     duration = (finished - start) / 1e6  # ms
-                    code = getattr(couchers_context._grpc_context.code(), "name", None)  # type: ignore[union-attr]
+
+                    if couchers_context._grpc_context:
+                        context_code = couchers_context._grpc_context.code()  # type: ignore[attr-defined]
+                        code = getattr(context_code, "name", None)
+                    else:
+                        code = None
+
                     traceback = "".join(format_exception(type(e), e, e.__traceback__))
                     _store_log(
                         method=method,
--- a/app/backend/src/couchers/models/host_requests.py
+++ b/app/backend/src/couchers/models/host_requests.py
@@ -31,7 +31,7 @@ class HostRequestQuality(enum.Enum):
     low_quality = enum.auto()
 
 
-class HostRequest(Base, init=False, kw_only=True):
+class HostRequest(Base, kw_only=True):
     """
     A request to stay with a host
     """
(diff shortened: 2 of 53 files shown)


app/backend/src/couchers/models/host_requests.py
```diff
@@ -72,17 +72,17 @@ class HostRequest(Base, init=False, kw_only=True):
     surfer_last_seen_message_id: Mapped[int] = mapped_column(BigInteger, default=0)
 
     # number of reference reminders sent out
-    host_sent_reference_reminders: Mapped[int] = mapped_column(BigInteger, server_default=text("0"))
-    surfer_sent_reference_reminders: Mapped[int] = mapped_column(BigInteger, server_default=text("0"))
-    host_sent_request_reminders: Mapped[int] = mapped_column(BigInteger, server_default=text("0"))
+    host_sent_reference_reminders: Mapped[int] = mapped_column(BigInteger, server_default=text("0"), init=False)
```
tristanlabelle, line 75: Noob question: Does the code specify the server default because our ORM means that the code dictates the database schema? Otherwise I'd have expected the code not to need to know about the value of the server default.

app/backend/src/couchers/servicers/events.py
```diff
@@ -441,17 +443,22 @@ def CreateEvent(
         ):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "photo_not_found")
 
+        thread = Thread()
```
tristanlabelle, line 446: For my understanding, remind me the technical reason why we couldn't support `Event(thread=Thread())`?

app/backend/src/couchers/servicers/pages.py
```diff
@@ -120,17 +120,23 @@ def CreatePlace(
         ):
             context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "photo_not_found")
 
+        parent_node = get_parent_node_at_location(session, geom)
+        if not parent_node:
+            context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "location_not_in_any_community")
+        thread = Thread()
+        session.add(thread)
+        session.flush()
```
tristanlabelle, line 128: Could we have a helper like `def add_to_session(session, T) -> T` that returns the added and flushed Thread, so we can keep it inline as `thread_id=add_to_session(session, Thread()).id`?

app/backend/src/couchers/interceptors.py
```diff
@@ -187,7 +187,7 @@ def _sanitize_message(message: Message) -> None:
 def _store_log(
     *,
     method: str,
-    status_code: grpc.StatusCode | None,
+    status_code: str | None,
```
tristanlabelle, line 190: Why couldn't we keep the more specific type?

app/backend/src/tests/fixtures/db.py
```diff
@@ -357,3 +358,10 @@ def add_users_to_new_moderation_list(users: list[User]) -> int:
 
 def run_migration_test():
     return os.environ.get("RUN_MIGRATION_TEST", "false").lower() == "true"
+
+
+def make_volunteer(started_volunteering: date, show_on_team_page: bool = True, **kwargs: Any) -> Volunteer:
```
tristanlabelle, line 363: I think the one-liner isn't worth losing the typing we get from calling the initializer directly.

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -1066,15 +1066,14 @@ def test_send_host_request_reminders(db, moderator):
...
-                text="Looking forward to hosting you!",
-                message_type=MessageType.text,
-            )
+        msg = Message(
+            conversation_id=hr7,
+            author_id=user14.id,
+            text="Looking forward to hosting you!",
+            message_type=MessageType.text,
         )
+        msg.time = now()
```
tristanlabelle, line 1075: Any way we could promote these to `Message.set_time(now) -> Message` so it can be kept as a single expression but also support assigning optional properties?

tristanlabelle (approved):
Looks good, though I think we'd benefit from a helper "add_and_flush(session, T) -> T" function that lets us keep one-liners instead of the variable declaration, add and flush triple that add boilerplate.

app/backend/src/couchers/models/host_requests.py
WouldYouKindly: Yeah these are at the same time table definitions and data objects. All alembic migrations are actually generated from these models.

app/backend/src/couchers/servicers/events.py
WouldYouKindly: If you allow both `Event(thread=thread)` and `Event(thread_id=thread.id)`, mypy expects you to pass both. So to avoid doing that all the time, I had to choose which one is going to be required. I chose the foreign key column, since it's always available (unlike the related object).

app/backend/src/couchers/servicers/pages.py
WouldYouKindly: Yeah I actually had a PR that did exactly that. But it's kinda ugly, and I think we'd benefit from a more radical approach, like having repositories that abstact the DB operations.

app/backend/src/couchers/interceptors.py
WouldYouKindly: The code is passing a string, so I changed the type to reflect the reality (the type check was failing because the calling code is passing Any :((( )

app/backend/src/tests/fixtures/db.py
WouldYouKindly: This is mainly to not update every callsite with with setting an attribute... I agree that's unfortunate. I have some ideas to try to address this

app/backend/src/tests/test_bg_jobs.py
WouldYouKindly: I guess! I'll play with it and see what I can come up with

PR #7629, opened 2026-01-07
Backend: Fix custom email previews

tristanlabelle:
https://github.com/Couchers-org/couchers/blob/b2ff9e41ecb541d3e06aeb409c280c17867f548f/app/backend/src/couchers/notifications/render.py#L596 painstakingly defines preview text for our emails, but the string was never templated into the email.

image

Missing title:

```html
<head>
  <title></title>
  <!--[if !mso]><!-->
```

Missing preview:

```html
<body style=3D"word-spacing:normal;background-color:#eeeeee;">
  <div style=3D"display:none;font-size:1px;color:#ffffff;line-height:1px;ma=
x-height:0px;max-width:0px;opacity:0;overflow:hidden;"></div>
```

After fix:

image

image

Also renames those `subject`/`preview` placeholders with a `header_` prefix since they are/will be filled by a different abstraction level than the email-specific body placeholders, and it makes them consistent with the `footer_` placeholders.

Send a message to get an email notification, then check the preview.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/notifications/background.py
+++ b/app/backend/src/couchers/notifications/background.py
@@ -56,6 +56,8 @@ def _send_email_notification(session: Session, user: User, notification: Notific
         email_lang = user.ui_language_preference or "en"
 
     template_args = {
+        "header_subject": rendered.subject,
+        "header_preview": rendered.preview,
         "user": user,
         "time": notification.created,
         "footer_email_is_critical": rendered.is_critical,
--- a/app/backend/templates/v2/generated_html/account_deletion_complete.html
+++ b/app/backend/templates/v2/generated_html/account_deletion_complete.html
@@ -2,7 +2,7 @@
 <html lang="und" dir="auto" xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
 
 <head>
-  <title>{{ subject }}</title>
+  <title>{{ header_subject }}</title>
   <!--[if !mso]><!-->
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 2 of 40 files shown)


app/backend/src/couchers/notifications/background.py
```diff
@@ -56,6 +56,8 @@ def _send_email_notification(session: Session, user: User, notification: Notific
         email_lang = user.ui_language_preference or "en"
 
     template_args = {
+        "header_subject": rendered.subject,
+        "header_preview": rendered.preview,
```
tristanlabelle, line 60: Self-blocking until tested (internet too slow right now)

app/backend/templates/v2/generated_html/account_deletion_complete.html
```diff
@@ -2,7 +2,7 @@
 <html lang="und" dir="auto" xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office">
 
 <head>
-  <title>{{ subject }}</title>
+  <title>{{ header_subject }}</title>
```
WouldYouKindly, line 5: do we have to rename these two everywhere? It's not required to fix the bug, is it?

tristanlabelle: It's not required but it mirrors the `footer_` placeholders and leaves undecorated placeholder names for the email bodies.

kalvdans (approved):

app/backend/src/couchers/notifications/background.py
tristanlabelle: Tested, screenshots in description

PR #7630, opened 2026-01-07
Refactor permission checks in interceptors

WouldYouKindly:
- Extract checking auth level and permissions into functions
- Reuse these functions in FakeChannel instead of relying on their copy-pasted versions
- Unit test these functions (the benefit is that the tests don't require a database since they're testing pure functions)

Note that there's no change in behaviour, and all the existing tests still pass.

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -1,7 +1,7 @@
 import logging
 from collections.abc import Callable
 from copy import deepcopy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from os import getpid
 from threading import get_ident
@@ -12,6 +12,7 @@ from typing import Any, NoReturn, cast, overload
 import grpc
 import sentry_sdk
 from google.protobuf.descriptor import ServiceDescriptor
+from google.protobuf.descriptor_pool import DescriptorPool
 from google.protobuf.message import Message
 from opentelemetry import trace
 from sqlalchemy import Function, select
@@ -32,6 +33,7 @@ from couchers.descriptor_pool import get_descriptor_pool
 from couchers.metrics import observe_in_servicer_duration_histogram
 from couchers.models import APICall, User, UserActivity, UserSession
 from couchers.proto import annotations_pb2
+from couchers.proto.annotations_pb2 import AuthLevel
 from couchers.utils import (
     create_lang_cookie,
     create_session_cookies,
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/interceptors.py
```diff
@@ -419,6 +364,109 @@ def function_without_couchers_stuff(req: Message, grpc_context: grpc.ServicerCon
...
+    user_agent = headers.get("user-agent")
+
+    ui_lang = parse_ui_lang_cookie(headers)
+    user_id = parse_user_id_cookie(headers)
+
+    return CouchersHeaders(
+        token=token,
+        is_api_key=is_api_key,
+        ip_address=ip_address if isinstance(ip_address, str) else None,
+        user_agent=user_agent if isinstance(user_agent, str) else None,
```
kalvdans, line 401:
```suggestion
        ip_address=ip_address,
        user_agent=user_agent,
```

Nit: They can't be `bytes` if the header doesn't end in '-bin', from https://grpc.github.io/grpc/python/glossary.html#term-metadata :
A key-value pair included in the HTTP header. It is a 2-tuple where the first entry is the key and the second is the value, i.e. (key, value). The metadata key is an ASCII str, and must be a valid HTTP header name. The metadata value can be either a valid HTTP ASCII str, or bytes. If bytes are provided, the key must end with ‘-bin’, i.e. ('binary-metadata-bin', b'\\x00\\xFF')

app/backend/src/couchers/interceptors.py
```diff
@@ -419,6 +364,109 @@ def function_without_couchers_stuff(req: Message, grpc_context: grpc.ServicerCon
...
+        user_id=user_id,
+    )
+
+
+class BadHeaders(Exception):
+    pass
+
+
+class AbortError(Exception):
+    def __init__(self, msg: str, code: grpc.StatusCode) -> None:
```
kalvdans, line 412: Nit: No return value type hints should be done on the `__init__()` method.

kalvdans (approved):
Great refactoring and thanks for adding all those unit tests!

app/backend/src/couchers/interceptors.py
WouldYouKindly: I'll remove it in a follow-up, thanks.

app/backend/src/couchers/interceptors.py
WouldYouKindly: This is good to know! I'd rather keep the defensive check though - who knows if grpc is actually enforcing it at runtime? And whether it can change/break. (we'd also have to type: ignore the line which is even more annoying than an isinstance check)

PR #7638, opened 2026-01-08
Backend: Decouple i18n functions from email templating

tristanlabelle:
Different parts of our backend need to localize dates and times, and currently they need to use email templating functions to do that. This PR splits the responsibility.

1. Adds `localize_date`, `localize_time`, etc. functions to i18n
2. Update non-email templating uses of `v2date`, `v2time`, etc. to use new i18n functions
3. Update `v2date`, `v2time`, etc. to take their locale and timezone from the email formatting context rather than an argument being passed down.

This is working towards further encapsulating email formatting and the v2*** functions.

Run existing tests.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/i18n/i18n.py
+++ b/app/backend/src/couchers/i18n/i18n.py
@@ -5,8 +5,8 @@ from functools import lru_cache
 from pathlib import Path
 from zoneinfo import ZoneInfo
 
-from google.protobuf.timestamp_pb2 import Timestamp
 import phonenumbers
+from google.protobuf.timestamp_pb2 import Timestamp
 
 from couchers.i18n.constants import LANGUAGE_FALLBACKS
 from couchers.i18n.i18next import I18Next
--- a/app/backend/src/couchers/i18n/i18n.py
+++ b/app/backend/src/couchers/i18n/i18n.py
@@ -6,10 +6,12 @@ from pathlib import Path
 from zoneinfo import ZoneInfo
 
 from google.protobuf.timestamp_pb2 import Timestamp
+import phonenumbers
 
 from couchers.i18n.constants import LANGUAGE_FALLBACKS
 from couchers.i18n.i18next import I18Next
 from couchers.i18n.plurals import PluralRules
+from couchers.models.users import User
 from couchers.utils import to_aware_datetime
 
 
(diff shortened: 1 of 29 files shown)


app/backend/src/couchers/i18n/i18n.py
```diff
@@ -65,3 +72,58 @@ def localize_string(lang: str | None, key: str, *, substitutions: Mapping[str, s
         The translated string with substitutions applied
     """
     return get_i18next().localize(key, lang or "en", substitutions)
+
+
+def localize_date(date: datetime.date, locale: str) -> str:
```
WouldYouKindly: So far the convention has been to `from datetime import date` - let's keep it that way?

app/backend/src/couchers/i18n/i18n.py
```diff
@@ -65,3 +72,58 @@ def localize_string(lang: str | None, key: str, *, substitutions: Mapping[str, s
...
+        The localized date and time string.
+    """
+    if isinstance(datetime, Timestamp):
+        datetime = to_aware_datetime(datetime)
+
+    # A timezone-unaware datetime is almost certainly a bug, so we don't support it.
+    assert datetime.tzinfo is not None, "Cannot localize a timezone-unaware datetime."
+
+    if timezone is not None:
+        datetime = datetime.astimezone(timezone)
```
WouldYouKindly: this shadows the imported module/function

app/backend/src/couchers/i18n/i18n.py
```diff
@@ -65,3 +72,58 @@ def localize_string(lang: str | None, key: str, *, substitutions: Mapping[str, s
...
+
+    localized_date = localize_date(datetime.date(), locale)
+    localized_time = localize_time(datetime.time(), locale)
+
+    # TODO(#7590): Account for locale
+    return f"{localized_date} at {localized_time}"
```
WouldYouKindly, line 119: So this doesn't actually localise at the moment?

app/backend/src/couchers/templates/v2.py
```diff
@@ -36,9 +35,27 @@
 
 # Special context values expected by v2 filters
 CONTEXT_YEAR_KEY = "_year"
-CONTEXT_TIMEZONE_DISPLAY_KEY = "_timezone_display"
-CONTEXT_TRANSLATION_LANGUAGE_KEY = "_lang"
-CONTEXT_PLAINTEXT_KEY = "_plain"
+
+
+@dataclass(slots=True, kw_only=True)
```
WouldYouKindly: frozen?

app/backend/src/couchers/templates/v2.py
```diff
@@ -36,9 +35,27 @@
...
+    """The timezone to use when formatting times."""
+
+    locale: str
+    """The locale to use when localizing strings or formatting times."""
+
+    plaintext: bool
+    """If true, strips html tags from localized strings."""
+
+    @staticmethod
+    def get(jinja2_context: Context) -> FilterContext:
```
WouldYouKindly: `get` sounds like it can return `None`. Maybe `from_jinja`?

app/backend/src/couchers/i18n/i18n.py
tristanlabelle: Right, we didn't before and still don't, but the code shape is right for when we implement it. We'll actually need to pull in the Unicode CLDR to correctly localize. The comment points to the GitHub issue.

app/backend/src/couchers/i18n/i18n.py
tristanlabelle: I used `import datetime` so I could name my parameter `date` without clashing, and then I named another parameter `datetime` lol. Fixed in latest. I'm not a fan of Python's approach to namespacing.

app/backend/src/couchers/i18n/i18n.py
tristanlabelle: Changed to value since I now import datetime

app/backend/src/couchers/templates/v2.py
tristanlabelle: It can't return None, users are v2*** functions which can assume that by contract this object will be available.

tristanlabelle: Oh, sorry misread you. I usually think "find" can return None, whereas "get" doesn't. I can use from_jinja

app/backend/src/couchers/templates/v2.py
tristanlabelle: I actually modify it for plaintext then html but I don't really have to, will change

tristanlabelle:
@WouldYouKindly All feedback addressed.

app/backend/src/couchers/templates/v2.py
WouldYouKindly: @tristanlabelle you can use `dataclasses.replace` to get a copy with an updated field

WouldYouKindly (approved):

app/backend/src/couchers/templates/v2.py
tristanlabelle: Oh good to know for next time!

PR #7649, opened 2026-01-10
Build postgres when running locally

WouldYouKindly:
When running locally, build postgis image instead of usign the one from dockerhub. The latter does not have an ARM version, and x86 version is much slower on macs.


tristanlabelle (approved):
Services were previously ordered alphabetically

WouldYouKindly:
> Services were previously ordered alphabetically

I put postgres at the botton to clearly separate it from the ones that just extend the base dockerfile. Feel free to re-order them if you have a strong opinion though

PR #7651, opened 2026-01-10
Add nginx config to proxy Sentry requests through our server

WouldYouKindly:
This is needed to avoid ad-blockers cancelling requests to sentry. Described in their docs here https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-tunnel-option

We'll need to configure sentry in web too


aapeliv (approved):
lgtm

PR #7652, opened 2026-01-11
Add tests for authentication

WouldYouKindly:
Add a bunch of tests for authentication logic. This covers all logic in `_try_get_and_update_user_details` including `where` conditions in the sql query (the only untested thing is splitting `UserActivity` by user-agent, because it's proven to be tricky)

### DIFF
--- a/app/backend/src/tests/test_interceptors.py
+++ b/app/backend/src/tests/test_interceptors.py
@@ -1,5 +1,8 @@
+from collections.abc import Callable, Generator
 from concurrent import futures
 from contextlib import contextmanager
+from datetime import timedelta
+from typing import Any
 from unittest.mock import Mock
 
 import grpc
@@ -7,7 +10,7 @@ import pytest
 from google.protobuf import empty_pb2
 from google.protobuf.descriptor import ServiceDescriptor
 from google.protobuf.descriptor_pool import DescriptorPool
-from sqlalchemy import select
+from sqlalchemy import select, update
 
 from couchers.constants import (
     MISSING_AUTH_LEVEL_ERROR_MESSAGE,
@@ -15,6 +18,7 @@ from couchers.constants import (
 )
 from couchers.crypto import random_hex
 from couchers.db import session_scope
+from couchers.descriptor_pool import get_descriptor_pool
 from couchers.interceptors import (
     AbortError,
     BadHeaders,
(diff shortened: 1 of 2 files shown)


app/backend/src/tests/test_interceptors.py
```diff
@@ -403,6 +417,96 @@ def test_auth_interceptor(db):
...
+    with interceptor_dummy_api(**rpc_def, creds=grpc.local_channel_credentials()) as call_rpc:
+        call_rpc(empty_pb2.Empty(), metadata=(cookie_auth(token), ("x-couchers-real-ip", "1.1.1.1")))
+
+    with session_scope() as session:
+        api_calls = session.execute(
+            select(UserActivity.api_calls).where(UserActivity.ip_address == "1.1.1.1")
+        ).scalar_one()
+        assert api_calls == 1
+
+    # Check that activity is split in time bins.
```
tristanlabelle, line 494: Where are the time bins in this test?

app/backend/src/tests/test_interceptors.py
```diff
@@ -403,6 +417,96 @@ def test_auth_interceptor(db):
...
+
+    # Check that last_active is updated if it wasn't updated in a while.
+    with interceptor_dummy_api(**rpc_def, creds=grpc.local_channel_credentials()) as call_rpc:
+        call_rpc(empty_pb2.Empty(), metadata=(cookie_auth(token),))
+
+    with session_scope() as session:
+        last_active = session.execute(select(User.last_active).where(User.id == user.id)).scalar_one()
+        assert last_active > now() - timedelta(seconds=1)
+
+    # Check that last_active is untouched.
```
tristanlabelle: Why is it not updated this time? (comment could say)

tristanlabelle (dismissed):

app/backend/src/tests/test_interceptors.py
WouldYouKindly: updated the comment to make it clear

app/backend/src/tests/test_interceptors.py
WouldYouKindly: Updated the comment, thanks.

tristanlabelle (approved):
Thanks for iterating!

aapeliv (approved):
lgtm, unclear what you did in the nonexistent service bit

PR #7654, opened 2026-01-11
Backend: Proof of concept of email templating rev

tristanlabelle:
[Not for merging]

Check the test's `EmailContent(` for an illustration of how we'd be building up emails in code that can be rendered to HTML or plaintext both.

Check `prototype.mjml` for how we'd define our sections prototypes. We could actually pull the header and footer into this file too since we could remove all other mjml files.


nabramow:
@tristanlabelle  This one still in progress? If not should we close? Just trying to get the PRs back under control 😅.

tristanlabelle:
@nabramow Closing this one as I've moved forward with an approach.

PR #7655, opened 2026-01-11
DRAFT Add Justfiles

WouldYouKindly:
[just](https://just.systems/man/en/introduction.html) is a better version of `make`. The main features (for me) are sane syntax and error messages, and the ability to run a recipe from any subdirectory. Which means we can define `protos` recipe in the top-level Justfile, and call it from anywhere :) 

The tool seems to be well-supported and widely used, and old enough to be stable. Since no one has a huge affection for `make`, why not give this a try.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,25 @@
+# Justfile for the backend of the application
+# Run 'just [command]' in app/backend to execute the specified command
+
+# See app/backend/readme.md.
+
+PROTO_IMAGE := "registry.gitlab.com/couchers/grpc"
+DB_PASSWORD := "203d805f4b62c0a1b2f1f6b82d4583dfe563ec1619b83ce22ee414e8376a25e7"
+APP_DIR := justfile_directory() + "/app"
+
+# Generate protocol buffer files
+protos:
+    docker pull -q {{PROTO_IMAGE}}
+    docker run --rm -w /app -v {{APP_DIR}}:/app {{PROTO_IMAGE}} ./generate_protos.sh
+
+# Connect to the local development database
+db:
+    PGPASSWORD={{DB_PASSWORD}} psql -h localhost -p 6545 -U postgres -d postgres
+
+# Run backend locally with all needed dependencies (db etc).
+run-backend:
+    #!/usr/bin/env bash
+    cd app/backend
+    just run-backend
+
+alias rb := run-backend
--- a/Justfile
+++ b/Justfile
@@ -20,6 +20,19 @@ db:
 run-backend:
     #!/usr/bin/env bash
     cd app/backend
-    just run-backend
+    just run
 
 alias rb := run-backend
+
+
+setup-frontend:
+    #!/usr/bin/env bash
+    cd app/web
+    just setup
+
+run-frontend:
+    #!/usr/bin/env bash
+    cd app/web
+    just run
+
+alias rf := run-frontend
--- a/app/backend/Justfile
+++ b/app/backend/Justfile
@@ -1,0 +1,80 @@
+# Justfile for the backend of the application
+# Run 'just [command]' in app/backend to execute the specified command
+
+set dotenv-load := true
+set dotenv-filename := "../backend.dev.env"
+
+DB_PASSWORD := "203d805f4b62c0a1b2f1f6b82d4583dfe563ec1619b83ce22ee414e8376a25e7"
+DB_CONNECTION_STRING := "postgresql://postgres:" + DB_PASSWORD + "@localhost:6545/postgres"
+
+# Running `just` without arguments executes this (qc stands for quality control).
+[default]
+qc: fmt mypy test
+
+# Quick quality check.
+qqc: fmt mypy
+
+# Run backend locally with all needed dependencies (db etc).
+run-backend:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    docker compose -f ../docker-compose.local-backend.yml up -d --build --wait
+    export DATABASE_CONNECTION_STRING={{DB_CONNECTION_STRING}}
+    export SMTP_HOST=localhost
+    export OPENTELEMETRY_ENDPOINT=localhost:4317
+    uv run src/app.py
+
+alias rb := run-backend
+
+# Connect to the local development database
+db:
+    PGPASSWORD={{DB_PASSWORD}} psql -h localhost -p 6545 -U postgres -d postgres
+
+# Downgrade the local database to the previous migration
+downgrade:
+    DATABASE_CONNECTION_STRING={{DB_CONNECTION_STRING}} uv run alembic downgrade -1
+
+# Upgrade the local database to the next migration
+upgrade:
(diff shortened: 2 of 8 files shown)


WouldYouKindly:
@aapeliv @nabramow @tristanlabelle wdyt

aapeliv:
I don't think this is a huge deal, so down for whatever others feel is good: let's not bikeshed this. But here are my strong opinions, weakly held :)

---

I think adding another tool that needs to be installed in order to do dev is a high tax. Right now we require "the standard dev tool bundle" (git/curl/make/etc) and docker (which we really can't do without replacing it with half a dozen other things). The point of `make` in my mind is that even though it sucks, it's installed one every single dev machine that exists and is part of the set of standard tools. If we go away from make then why not just write our own python dev tool (with only stdlib, old python3 compat so it runs also on every recent dev system) which will be clean and powerful and we'll be familiar with?

I believe it's also helpful to have this setup so that you can run everything easily without this tool. So we should have a "minimal dev setup", which one can run with standard tools and minimal deps that allows one to get it up and running and be productive; then have power tools that power users can use to be more productive once they are familiar with the system. In other words create a more shallow onboarding curve.

tristanlabelle:
Replied in the Slack convo!

app/backend/Justfile
```diff
@@ -0,0 +1,85 @@
...
+downgrade:
+    DATABASE_CONNECTION_STRING={{DB_CONNECTION_STRING}} uv run alembic downgrade -1
+
+# Upgrade the local database to the next migration
+upgrade:
+    DATABASE_CONNECTION_STRING={{DB_CONNECTION_STRING}} uv run alembic upgrade +1
+
+# Linters.
+format:
+    uv run ruff check --select I --fix . && uv run ruff check . --fix && uv run ruff format .
```
aapeliv: I think we can actually merge the two checks into one `ruff check --fix .`

Justfile
```diff
@@ -0,0 +1,26 @@
+# Justfile for the backend of the application
+# Run 'just [command]' in app/backend to execute the specified command
+
+# Install `just`: https://just.systems/man/en/pre-built-binaries.html
```
aapeliv: These instructions are not clear enough! This expects deep understanding of where to put the binary.

The package manager versions are a bit easier though; but we need to then help people install this

aapeliv (commented):
As mentioned on slack, i'm happy to go with this.

* We should also add a justfile to the frontend repo to keep the main commands in + top level file that e.g. allows setting everything up in one hit.
* In dev we should install latest uv?
* In general we should be careful about installing stuff on people's machines in the just files. I almost feel we should have those things be explicit instructions for setting up the dev env?
* We should move the top level just file into `app/`, but not super strong on that if it is important for someone to keep at the very top level.

Justfile
WouldYouKindly: Removed this, and instead redirected the reader to backend readme. Backend readme now links to the pre-built packages.

app/backend/Justfile
WouldYouKindly: We can! Done

WouldYouKindly:
@aapeliv 

> We should also add a justfile to the frontend repo to keep the main commands in + top level file that e.g. allows setting everything up in one hit.

done.
 
> In dev we should install latest uv?
> In general we should be careful about installing stuff on people's machines in the just files. I almost feel we should have those things be explicit instructions for setting up the dev env?

Agreed. Removed the command from Justfile, and instead pointed people to the official `uv` installer.

> We should move the top level just file into app/, but not super strong on that if it is important for someone to keep at the very top level.

I kinda like being able to run commands without going to `app` first

app/backend/readme.md
```diff
@@ -1,83 +1,91 @@
 # Couchers.org backend
 
-This is the backend for Couchers.org built in Python. The React web frontend and React Native apps make API calls to it using [gRPC](https://grpc.io/) which are defined via [Protocol Buffers](https://protobuf.dev/). The business logic manipulates a [PostgreSQL](https://www.postgresql.org/) database via [SQLAlchemy](https://www.sqlalchemy.org/). Geospatial features are provided by the [PostGIS extension](https://postgis.net/).
+This is the backend for Couchers.org built in Python. The React web frontend and React Native apps 
+make API calls to it using [gRPC](https://grpc.io/) which are defined via [Protocol Buffers](https://protobuf.dev/). The business
```
aapeliv, line 4: what happened with all the prose here? could you undo it please?

app/web/readme.md
```diff
@@ -4,16 +4,19 @@ This is the react/nextjs web frontend for Couchers.org. We are using Typescript
...
 
-_Readme last updated: 2025/08/23._
+_Readme last updated: 2026/01/13._
 
 ## Quick Start
 
 _These instructions should work directly on Linux and macOS. If you are using Windows, please [install Ubuntu via WSL2](https://documentation.ubuntu.com/wsl/en/latest/guides/install-ubuntu-wsl2/), then follow these instructions inside Ubuntu._
 
+First, install `just`. It's a command runner that we use instead of `make`. 
+[Find a package for your OS here](https://just.systems/man/en/packages.html).
```
aapeliv, line 14: Please put in the homebrew and ubuntu install commands here + link to this page for more.

For frontend, how about putting `just` in for power users? It does not add much for the quick start/simple frontend dev. It's one less step. Also everything we have now is bog standard frontend stuff (node+yarn), `just` is less standard so it will add a hurdle.

@nabramow what do you think?

app/web/readme.md
```diff
@@ -34,25 +37,14 @@ You also need `git`, `curl`, and `tar`, they should be preinstalled or be availa
 Now run the following commands to get up and running:
 
 ```sh
-## Check out the repo and navigate to app/web
-# clone the git repo
+# Clone the git repo
 git clone https://github.com/Couchers-org/couchers.git
```
aapeliv, line 41: You do need to `cd couchers` first though

app/web/readme.md
```diff
@@ -146,7 +138,7 @@ yarn start
 
 ## Sentry error logging
 
-We self-host sentry to capture errors. It can be accessed at https://couchers.sentry.io/.
```
aapeliv, line 149: Thanks for catching. Prior was not true!

aapeliv:
Also could you please rebase as backend readme was updated!

app/backend/readme.md
WouldYouKindly: I just added line breaks to make it more readable. At least in pycharm the lines are not wrapped, so one has to scroll right

app/web/readme.md
WouldYouKindly: tbh I'm still experimenting with just and can't say that I'm full of passion for it. Maybe we can add the same couple of commands for backend to make, create a top level makefile, and use includes

tristanlabelle:
I just learned that Just is only a command runner and not a build system like Make: it doesn't support declaring dependencies between inputs and outputs to allow for incremental builds. In my view, it makes it less useful than Make, despite the syntactic advantages.

WouldYouKindly:
Closing, as I don't think the added complexity is justified.

PR #7656, opened 2026-01-12
Add quick start back in to readme

aapeliv:
Updates backend readme with quick start. See slack convo.

**Other**


app/backend/readme.md
```diff
@@ -1,30 +1,36 @@
...
+We want this file to be readable to a relative noob, so let's make sure it has a quick start section that is super straightforward and simple to run, not requiring much understanding on what's going on. We can then explain later in the readme.
+-->
 
-- **Linux or macOS**. If you are using Windows, please [install Ubuntu via WSL2](https://documentation.ubuntu.com/wsl/en/latest/guides/install-ubuntu-wsl2/), then follow these instructions inside Ubuntu.
-- **Docker engine**. Please refer to the [Docker install documentation](https://docs.docker.com/engine/install/) for how to set it up.
-- **A clone of the Couchers git repo**. `git clone https://github.com/Couchers-org/couchers.git`
+This is the backend for Couchers.org built in Python. The React web frontend and React Native apps make API calls to it using [gRPC](https://grpc.io/) which are defined via [Protocol Buffers](https://protobuf.dev/). The business logic manipulates a [PostgreSQL](https://www.postgresql.org/) database via [SQLAlchemy](https://www.sqlalchemy.org/). Geospatial features are provided by the [PostGIS extension](https://postgis.net/).
 
-### Up-to-date protocol buffers
+*Readme last updated: 2025/01/12.*
```
tristanlabelle, line 9: Version control is more reliable at keeping dates updated. If a user sees that this text doesn't match the last commit to this file, what would they do?

app/backend/readme.md
```diff
@@ -1,30 +1,36 @@
...
 
-All steps assume you have up-to-date API definitions generated using protobuf and grpc. If you make or pull changes to \*.proto files, you'll need to rerun this:
+## Quick Start
```
tristanlabelle, line 11: I think we can achieve both goals:

- Make it easier to run for the first time
- Explain how to efficiently work day-to-day

Running `docker compose up --build` is a good for the former but not a viable iteration environment. I'd suggest just adding a first section for the first run (call it quick start if you like) and keeping the detailed instructions afterwards for when you graduate from your first run and have the basics covered. When I got to running outside of docker, that's when I really had to understand how stuff works.

app/backend/readme.md
```diff
@@ -44,26 +50,26 @@ make docker-test file=test_filename.py
...
 ```sh
+curl -LsSf https://astral.sh/uv/install.sh | sh
 cd app/backend
-make setup
-uv sync --frozen
+uv sync
 ```
 
-In a terminal, run all containerized services except the backend.
+In a terminal, run all containerized services except the backend with config set up for .
```
tristanlabelle, line 65: Sentence not complete?

app/backend/readme.md
```diff
@@ -44,26 +50,26 @@ make docker-test file=test_filename.py
 make teardown-docker-test
 ```
 
-## Running the backend locally
+## Running the backend outside Docker
 
-For quicker iteration, you can run the backend locally and have it talk to the rest of the containerized services.
```
tristanlabelle, line 49: I think it's important to call out somewhere that you run the backend locally because that's how you get an efficient development environment.

tristanlabelle (commented):
See my main comment about achieving both the "easy first run" and the "instructions for efficient dev environment" goals.

app/backend/readme.md
```diff
@@ -44,26 +50,26 @@ make docker-test file=test_filename.py
 make teardown-docker-test
 ```
 
-## Running the backend locally
+## Running the backend outside Docker
```
tristanlabelle, line 53: Good call on the "outside Docker" naming

app/backend/readme.md
aapeliv: The purpose of this was that we would keep track of the last time when the quick start was updated, to make sure that it's kept relatively recent. It requires doing a full run through the process in order to verify it all works. I guess a more descriptive statement would be "Quick Start last updated" or similar?

app/backend/readme.md
aapeliv: I basically kept the detailed instructions underneath, I think?

app/backend/readme.md
aapeliv: What exactly is quicker in iterating outside docker? When you run the backend insdie docker, it will reload on code change: in fact it's quicker than the vanilla instructions out of docker here?

app/backend/readme.md
tristanlabelle: I don't believe in humans manually updating dates or users getting much value from them, but time can prove me wrong!

tristanlabelle (approved):
Ah my confusion here was that I thought you couldn't have a dev loop with the docker setup so it was just a way to get the satisfaction of getting it to run once, but didn't lead into a sustainable environment. My bad.

PR #7664, opened 2026-01-12
Mobile fixes - android push notif deepl linking, notif icon, target blank links

nabramow:
Fixes these reported bugs:

* Opening profile via search opens in web browser instead of app
* Android: Clicking on push notification of hosting request takes user to dashboard in app rather than to messages/requests. Same happens with push notifs for "leave a reference" reminders
* Android: The top status bar notification icons are not visible in the standard app view unless the notification icon is not white. See screenshot below of problem and examples of other apps with the proper top status bar ([CS.com](http://cs.com/) app and Airbnb for reference)


aapeliv (approved):
i'm seeing a suss proportion of tests ;)

nabramow:
> i'm seeing a suss proportion of tests ;)

I can't test Android push notifs without a real device so was spiraling haha.

PR #7670, opened 2026-01-13
Photo galleries Part I: Dual write to galleries and avatar key

nabramow:
Breaking this PR up into smaller bits so we can discuss part by part and maybe the review is less overwhelming ;-).

For this PR I am adding code when a user adds a profile photo that it adds to both galleries and avatar_key.

The overall goal is to replace avatar_key with galleries and allow multiple profile photos.

The next steps:
- Migration to migrate existing avatar_keys into galleries #7671 
- Convert frontend to use profile_gallery_id instead of avatar_key (with existing UI)
- Remove avatar_key entirely
- Do new UI design to allow multiple photos

**TO TEST:**
- Run backend locally `docker compose up --build`
- In `.env.development` replace this value with local version: `NEXT_PUBLIC_API_BASE_URL="http://localhost:8888"
`
- In `web` folder run `yarn` to install deps and then `yarn start`
- Go to localhost:3000 in browser and login with `aapeli` dummy user
- Upload a photo via EditProfile and save
- Check DB. `users` table should have `avatar` key and `profile_gallery_items` table should have an `upload_key` with the same id.
- Replace with a different photo
- `profile_gallery_items` should have a new row added and `users` table should reference the new upload_key as the avatar_key

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -354,7 +354,7 @@ class API(api_pb2_grpc.APIServicer):
                 if user.profile_gallery:
                     # Add the new avatar as the only photo
                     item = PhotoGalleryItem(
-                        gallery_id=user.profile_gallery_id,
+                        gallery_id=user.profile_gallery.id,
                         upload_key=request.avatar_key.value,
                         position=0.0,
                     )
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -32,6 +32,7 @@ from couchers.models import (
     Notification,
     NotificationDeliveryType,
     ParkingDetails,
+    PhotoGalleryItem,
     RateLimitAction,
     Reference,
     RegionLived,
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/servicers/api.py
```diff
@@ -343,8 +344,21 @@ def UpdateProfile(
         if request.HasField("avatar_key"):
             if request.avatar_key.is_null:
                 user.avatar_key = None
+                # Also remove all photos from gallery?
+                if user.profile_gallery:
+                    for photo in user.profile_gallery.photos:
+                        session.delete(photo)
```
WouldYouKindly, line 350: So when user removes their avatar, we delete everything from profile gallery? I assume that's because they can only have one photo in profile gallery? Might be good to add a comment. Or ideally check that there's only one photo before deleting

app/backend/src/couchers/servicers/api.py
```diff
@@ -343,8 +344,21 @@ def UpdateProfile(
...
             else:
                 user.avatar_key = request.avatar_key.value
+                # Also add to gallery at position 0 (dual-write)
+                if user.profile_gallery:
+                    # Add the new avatar as the only photo
+                    item = PhotoGalleryItem(
+                        gallery_id=user.profile_gallery.id,
+                        upload_key=request.avatar_key.value,
+                        position=0.0,
```
WouldYouKindly, line 359: What will happen if they already have a photo at position 0?

app/backend/src/tests/test_api.py
```diff
@@ -627,6 +635,90 @@ def test_update_profile_do_not_email(db):
         assert e.value.details() == "You cannot enable meeting up while you have emails turned off in your settings."
 
 
+def test_update_profile_avatar_dual_write(db):
```
WouldYouKindly, line 638: Would be nice to have a test that updates an already existing avatar

aapeliv:
Closing in favor of #7530

PR #7674, opened 2026-01-13
Clean up navigation and add merch shop

nabramow:
- Adds "Merch Shop" to the right side drop down
- Remove "Dashboard" from main nav as they can click the couch icon, but adds it to right side drop down
- Remove "Messages" from right side drop down as it's already in the main nav


aapeliv (dismissed):
lgtm

PR #7677, opened 2026-01-13
Proxy sentry requests through backend

WouldYouKindly:


app/web/instrumentation-client.ts
```diff
@@ -1,7 +1,16 @@
 import * as Sentry from "@sentry/nextjs";
 
+const baseUrl = process.env.NEXT_PUBLIC_API_BASE_URL.replace("/api", "");
```
aapeliv: This is a bit hacky/error prone, maybe put ina nother ENV var?

aapeliv (changes requested):

app/web/instrumentation-client.ts
WouldYouKindly: Yeah, you're right. Done.

WouldYouKindly: @aapeliv ping :)

aapeliv (approved):
lgtm, please let's check on staging before rolling to prod

PR #7681, opened 2026-01-14
Add archived to group chats

nabramow:
Part of #6528 

We have archived functionality for host requests but not group chats. Adding it here.

### DIFF
--- a/app/backend/src/couchers/models/conversations.py
+++ b/app/backend/src/couchers/models/conversations.py
@@ -83,7 +83,7 @@ class GroupChatSubscription(Base, kw_only=True):
 
     last_seen_message_id: Mapped[int] = mapped_column(BigInteger, default=0)
 
-    is_archived: Mapped[bool] = mapped_column(Boolean, default=False, server_default=expression.false())
+    is_archived: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
 
     # when this chat is muted until, DATETIME_INFINITY for "forever"
     muted_until: Mapped[datetime] = mapped_column(
--- a/app/backend/src/couchers/models/conversations.py
+++ b/app/backend/src/couchers/models/conversations.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any
 from sqlalchemy import BigInteger, Boolean, DateTime, Enum, ForeignKey, String, func
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import DynamicMapped, Mapped, mapped_column, relationship
+from sqlalchemy.sql import expression
 
 from couchers.constants import DATETIME_INFINITY, DATETIME_MINUS_INFINITY
 from couchers.models.base import Base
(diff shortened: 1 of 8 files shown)


app/backend/src/tests/test_conversations.py
```diff
@@ -74,6 +74,36 @@ def test_list_group_chats(db, moderator):
...
         assert res.no_more
 
+    # Test archive filtering: archive group_chat1 for user1
+    with session_scope() as session:
+        subscription = session.execute(
+            select(GroupChatSubscription)
+            .where(GroupChatSubscription.group_chat_id == group_chat1_id)
+            .where(GroupChatSubscription.user_id == user1.id)
+        ).scalar_one()
+        subscription.is_archived = True
```
WouldYouKindly: No need to fetch it from the db:

```python
session.execute(
      update(GroupChatSubscription)
      .where(GroupChatSubscription.group_chat_id == group_chat1_id)
      .where(GroupChatSubscription.user_id == user1.id)
      .values(is_archived=True)
)
```

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -367,15 +367,21 @@ def ListGroupChats(
...
         results = session.execute(
             where_moderated_content_visible(
-                select(t, GroupChat, GroupChatSubscription, Message)
-                .join(Message, Message.id == t.c.message_id)
-                .join(GroupChatSubscription, GroupChatSubscription.id == t.c.group_chat_subscriptions_id)
-                .join(GroupChat, GroupChat.conversation_id == t.c.group_chat_id)
-                .where(or_(t.c.message_id < request.last_message_id, to_bool(request.last_message_id == 0)))
-                .order_by(t.c.message_id.desc())
-                .limit(page_size + 1),
+                query.order_by(t.c.message_id.desc()).limit(page_size + 1),
```
WouldYouKindly: fyi: you can order by and limit when you create the query, it doesn't have to be after `where`

app/backend/src/couchers/models/conversations.py
```diff
@@ -82,6 +83,8 @@ class GroupChatSubscription(Base, kw_only=True):
 
     last_seen_message_id: Mapped[int] = mapped_column(BigInteger, default=0)
 
+    is_archived: Mapped[bool] = mapped_column(Boolean, default=False, server_default=expression.false())
```
aapeliv: I think you want `init=False` instead of `default=False`

app/backend/src/couchers/servicers/conversations.py
aapeliv, line 364: I think it's easier to put here a `.where(or_(not request.HasField("only_archived"),GroupChatSubscription.is_archived == request.only_archived))`?

Either way I think you should do the filtering in this subquery instead of the one that joins on messages.

app/proto/conversations.proto
```diff
@@ -171,6 +172,7 @@ message GetGroupChatReq {
 message ListGroupChatsReq {
   uint64 last_message_id = 1;
   uint32 number = 2;
+  optional bool only_archived = 3;
```
aapeliv, line 179: Please don't annotate with `optional`, use `BoolValue` instead

aapeliv (changes requested):

app/proto/conversations.proto
aapeliv: actually fine whatever

aapeliv (approved):
lgtm

PR #7687, opened 2026-01-16
Simplify PR template

tristanlabelle:
Updates the pull request template to make it lighter weight and more consistent. 

I moved instructions addressing the PR author to HTML comments since reviewers don't need to see them. However, I kept the first line as italics because it calls out if some text needed to be replaced and wasn't. Italics feels more like placeholder text than the bold that is also used for checklist headings.

I moved checklist items that are already guaranteed by CI to tips in HTML comments since the CI results is authoritative over a local run's results.

## Testing
Checked the template in a markdown editor and used it here.

Also test by looking at the result on this branch here: https://github.com/Couchers-org/couchers/blob/tristan/update-pr-template/.github/pull_request_template.md

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@ Reference issues with "closes #1234", or simply "#1234" if not closing.
 *Explain how you tested this PR and give clear steps so the reviewer can replicate.*
 
 <!--
-Include any necessary dev environment adjustments such as editing .env files.
+Include screenshots and any necessary dev environment adjustments such as editing .env files.
 Fill applicable checklists below, or remove those that don't apply.
 -->
 
@@ -25,7 +25,7 @@ Fill applicable checklists below, or remove those that don't apply.
 - [ ] Clicked around my changes running locally and it works
 - [ ] Checked Desktop, Mobile and Tablet screen sizes
 
-## Merging
+## For maintainers
 <!-- Untick the following if you'd prefer that maintainers don't push commits/merge your branch. -->
 - [x] Maintainers can push commits to my branch
 - [x] Maintainers can merge this PR for me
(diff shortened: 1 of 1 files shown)


WouldYouKindly:
This is definitely easier to read than the current version!

nabramow:
This looks good to me! Would like @aapeliv to take a look too though!

.github/pull_request_template.md
```diff
@@ -1,49 +1,35 @@
...
 
-**Other**
-*Untick the following if you'd prefer that maintainers don't push commits/merge your branch.*
-- [x] A maintainer can push commits to my branch
-- [x] A maintainer can merge my PR (you can also merge after approval)
-
+<!--
+Maintainers may push commits to or merge your branch unless you uncomment the following:
+**Maintainers: Do not push commits to my branch.**
+**Maintainers: Do not merge this PR.**
```
aapeliv: I think a comment is not visible enough: let's leave the original there, or you can simplify it to fewer lines/checkboxes if you wish. We added this here because maintainers felt bad doing these things without a heads up.

aapeliv (changes requested):

.github/pull_request_template.md
tristanlabelle: Brought it back with minor tweaks

aapeliv (dismissed):
lgtm;

on testing: we might want to further push the responsibility onto the PR creator to "Explain how you tested this code, include screenshots of..." or similar. So basically reviewere djust checks that the PR creator tested it.

on merging: that heading includes more than just merging.

tristanlabelle:
@aapeliv Added `Include screenshots` in the comment that mentioned including dev environment adjustments. Renamed "Merging" to "For maintainers". I'll need reapproval.

aapeliv (approved):

PR #7688, opened 2026-01-16
Backend: Escape jinja2 placeholders by default

tristanlabelle:
Our use of jinja2 templates requires carefully adding a filter to every placeholder to make sure they get escaped if needed. This is error prone. This PR changes the default to escaping HTML tags, and adds a filter that explicitly marks a placeholder as safely containing HTML tags instead. This requires updating all email templates. It also:

- Renames filters to remove the `v2` prefix.
- Adds default formatting logic such that unmarked date/datetime/time objects get formatted with the context's locale.
- Supports providing the `I18Next` object for translation. (Emails will use a different object, and it improves mocking)
- Adds more unit tests

Plenty of unit tests were added. Email coverage is thin though, so run the backend with maildev and get an email or two to confirm.

image

image

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/email/send.py
+++ b/app/backend/src/couchers/email/send.py
@@ -32,8 +32,7 @@ def send_simple_pretty_email(
         "footer_email_is_critical": True,  # Results in no unsubscribe footer.
     }
 
-    # Not yet localizable
-    html_context = Context(output_html=True, timezone=ZoneInfo("Etc/UTC"), locale="en")
+    html_context = Context(output_html=True, timezone=timezone, locale=locale)
     plaintext_context = replace(html_context, output_html=False)
 
     # Format plaintext template
--- a/app/backend/src/couchers/email/send.py
+++ b/app/backend/src/couchers/email/send.py
@@ -1,3 +1,4 @@
+from dataclasses import replace
 from typing import Any
 from zoneinfo import ZoneInfo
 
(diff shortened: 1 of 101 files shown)


app/backend/src/couchers/email/send.py
```diff
@@ -42,10 +42,10 @@ def send_simple_pretty_email(
         html_tmplt,
         template_args,
         Context(
+            output_html=True,
```
tristanlabelle: Self-blocking TODO: Run locally and get an email to verify.

tristanlabelle: Validated and added screenshot to description.

aapeliv (dismissed):
i think lgtm. i hope that after this work you are doing on emails that we can have a system that generates sample emails in CI so we can audit all emails to make sure they are correct (as we discussed in the weekly meeting). in the meantime, with lack of such a tool; happy to move forward.

tristanlabelle:
@aapeliv rebased, so I'll need reapproval.

> i hope that after this work you are doing on emails that we can have a system that generates sample emails in CI so we can audit all emails to make sure they are correct

Yep, I'm getting close and have a plan. I think my next PR in this area will implement the new email templating system and offline pregeneration mechanism in isolation, such that I can push a before/after diff of emails to validate before we migrate over.

aapeliv (approved):
approved if this is just a rebase, please otherwise highlight the changes as this touches 100+ files

tristanlabelle:
@aapeliv Thanks. This is indeed just a rebase. Should I prefer merges in the future? I'm used to always rebasing on conflicts but can adapt.

Most of the noise from the html files will also go away when I'm done with my planned changes to emails. Currently as soon as I touch the template, I necessarily touch 30-40 files.

PR #7701, opened 2026-01-19
Backend: Localize push notifs enabled notification

tristanlabelle:
Push notifications are now translatable, except that one that says push notifications have been enabled, since it's an ad-hoc notification. Make it translatable.

Note that there are other notifications in this file but they are all test notifications and I don't think they can be seen by an end user.

You'd have to re-register for push notifications against a backend from this PR's branch.

N/A

**Backend checklist**
- [ ] All tests pass

**Other**

### DIFF
--- a/app/backend/src/couchers/notifications/locales/en.json
+++ b/app/backend/src/couchers/notifications/locales/en.json
@@ -1,5 +1,12 @@
 {
   "push": {
+    "adhoc": {
+      "push_enabled": {
+        "title": "Push notifications enabled",
+        "ios_title": "Push Notifications Enabled",
+        "body": "You'll now receive notifications on this device."
+      }
+    },
     "account_deletion__start": {
       "title": "Account deletion requested",
       "ios_title": "Account Deletion Requested",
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/notifications/locales/en.json
```diff
@@ -1,5 +1,12 @@
 {
   "push": {
+    "adhoc": {
```
tristanlabelle: Other strings below use "__" separator. Should I update them to use nesting instead?

app/backend/src/couchers/servicers/notifications.py
```diff
@@ -258,11 +258,9 @@ def RegisterMobilePushNotificationSubscription(
             push_notification_subscription_id=subscription.id,
             user_id=context.user_id,
             topic_action="adhoc:setup",
-            content=PushNotificationContent(
-                title="Push notifications enabled",
-                ios_title="Push Notifications Enabled",
-                body="You'll now receive notifications on this device.",
-            ),
+            content=render_adhoc_push_notification(
+                "push_enabled",
```
tristanlabelle: Mismatches the adhoc topic_action, but it's a better translation key name as it has more context.

aapeliv: Maybe rename the topic action in that case?

tristanlabelle: Good call, there is nothing else depending on that name. Renamed and made it more alike other topic_actions.

tristanlabelle:
@aapeliv Ready for re-review

aapeliv (approved):

PR #7706, opened 2026-01-20
Fix notifs test due to push collector change

aapeliv:


nabramow (approved):

app/backend/src/tests/test_notifications.py
```diff
@@ -1236,22 +1236,22 @@ def test_DebugRedeliverPushNotification_success(db, push_collector: PushCollecto
...
     with notifications_session(token) as notifications:
         notifications.DebugRedeliverPushNotification(
             notifications_pb2.DebugRedeliverPushNotificationReq(notification_id=notification_id)
         )
 
     # Verify a new push was sent
-    assert push_collector.count_for_user(user.id) == existing_push_count + 1
-    push = push_collector.get_for_user(user.id, index=existing_push_count)
+    push = push_collector.pop_for_user(user.id, last=True)
     assert "Active Volunteer" in push.content.title
```
nabramow, line 1255: Unrelated but is this gonna cause issues with translated push notifs?

aapeliv: I think it should default to English, so should be ok? I'm not entirely certain. That said, I think that should be handled in a different PR

PR #7707, opened 2026-01-20
Add dark mode toggle back to nav

nabramow:
As discussed in core meeting. This makes it easier to to find. We'll also leave it in settings too.


nabramow:
Okay @c-kreuz I've been converted to your way of thinking and put the toggle back in the nav in addition to in the settings.

c-kreuz (approved):

PR #7715, opened 2026-01-21
Teach Cursor and Claude about i18n best practices and add examples

### LINKED ISSUE #7663: Teach Claude/Cursor about i18n best practices
See comments in https://github.com/Couchers-org/couchers/pull/7530#discussion_r2680983923

Adding examples to `localization.md` may help

tristanlabelle:
Follow-up to discussion in https://github.com/Couchers-org/couchers/pull/7530#discussion_r2680983923 . This should improve the strings generated by AI.

It looks like we're duplicating instructions for Claude and Cursor?

Fixes #7663

Vibe code some strings.

N/A

**Other**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -63,6 +63,10 @@ make mypy
 - Run `make protos` from backend after changes
 - Internal job payloads in `/app/backend/proto/internal/jobs.proto`
 
+### Localization
+- Do not hard-code English text strings, store them in the appropriate locale files (`features/*/locales/en.json`)
+- When adding strings to an `en.json` file, refer to `/docs/localization.md` for string key and text guidance
+
 ## Testing
 
 ### Backend Tests
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -13,21 +13,27 @@ All user-visible strings should be externalized to `en.json` files as to be loca
 ### Writing strings
 
 - ✅ **DO** use placeholders to inject information known by the code into the string, e.g. a user's name into a greeting.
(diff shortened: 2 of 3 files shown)


docs/localization.md
```diff
@@ -36,9 +42,12 @@ String keys are our main way to communicate context to translators. It's okay to
 
 Placeholders allow inserting values known at runtime into the string. They are always preferable to concatenating strings in code, since they allow for reordering based on the language's grammar.
 
-- ✅ **DO** use descriptive placeholder names that allow translators to anticipate the format of the substituted string, e.g. `{{fullName}}`. It's okay to be verbose!
-- ✅ **DO** use separate placeholders for `{{date}}` and `{{time}}`.
```
tristanlabelle, line 40: I removed this guidance because we eventually want to rely on Unicode CLDR for formatting dates and times together.

CLAUDE.md
```diff
@@ -63,6 +63,10 @@ make mypy
 - Run `make protos` from backend after changes
 - Internal job payloads in `/app/backend/proto/internal/jobs.proto`
 
+### Localization
```
tristanlabelle, line 66: Could we consolidate the two files by having them both point to the same place?

nabramow: My understanding is that they don't follow import links so has to be one file each unfortunately. Unless it's changed in the few weeks since I was at my job as we specifically looked into this.

I think you can use /init to regenerate the CLAUDE.md but that will still be a manual step.

tristanlabelle: Hmm git supports symbolic links, which would probably work to make both files the same. Or once we have autoformatting in CI it'd be easy to have a step that copies one file into the other.

nabramow (approved):
LGTM!

PR #7716, opened 2026-01-21
French: Use non-breaking spaces

tristanlabelle:
Some French punctuation marks are preceded with spaces, but those should be non-breaking spaces (U+00A0) rather than regular spaces (U+0020) so that they don't get broken into two lines. This retroactively fixes all French strings to use non-breaking spaces. See https://en.wikipedia.org/wiki/Non-breaking_space

Technically, the recommendation is to use "narrow non-breaking spaces", which is a different Unicode character (U+202F), but since Weblate doesn't offer a way to input this character, we can't expect translators to honor the rule, so "non-breaking spaces" are good enough.

I'll do any weblate merging if this causes conflicts.

N/A - or use the website in French.

N/A

**Other**


app/backend/src/couchers/i18n/locales/fr.json
tristanlabelle, line 1: Self-blocking pending Dale's OK.

c-kreuz: Out of curiosity, how is this case handled in the future? Does it always need manual intervention?

tristanlabelle: Weblate has a tool that the translators need to use: "NBS"

image

c-kreuz (approved):

oskeitje:
I don't see any issue with the nbsp but the question is how can I input?

tristanlabelle:
@oskeitje Use this button!

image

PR #7740, opened 2026-01-25
Public trips - Rename fields in host request

### LINKED ISSUE #7739: Change naming of host_user_id and surfer_user_id in HostRequest
Since with public trips this will go both directions we want to change the naming to be sender and receiver language.

Do this in a separate PR first so we can make sure we got everything and nothing broke.

nabramow:

Closes #7739 

To prepare for public trips, we will first rename surfer_user_id and host_user_id to recipient and initiator. This is because public trips will allow us going both directions so the surfer initiates normally but host initiates with public trips.

## Testing

- Upgrade and downgrade migrations should work
- Send a couple host requests and go through reference flow. Should still work exactly as before.

**Backend checklist**

## For maintainers


app/backend/src/couchers/migrations/versions/0140_rename_host_request_user_columns_v2.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Helper to rename column only if old name exists
+    def rename_column(old_name, new_name):
+        if old_name in columns:
+            op.alter_column("host_requests", old_name, new_column_name=new_name)
+
+    # Helper to rename index only if old name exists
+    def rename_index(old_name, new_name):
+        if old_name in indexes:
+            op.execute(f"ALTER INDEX {old_name} RENAME TO {new_name}")
```
aapeliv: Let's not do this; just normal renaming per-column please

app/backend/src/tests/test_moderation.py
```diff
@@ -45,15 +45,15 @@ def _(testconfig):
     pass
 
 
-def create_test_host_request_with_moderation(surfer_token, host_user_id):
+def create_test_host_request_with_moderation(surfer_token, recipient_user_id):
```
aapeliv: let's not change this one

app/backend/src/tests/test_references.py
```diff
@@ -38,12 +38,12 @@ def _(testconfig):
 
 def create_host_request(
     session: Session,
-    surfer_user_id: int,
-    host_user_id: int,
+    initiator_user_id: int,
+    recipient_user_id: int,
```
aapeliv: here to keep as is

app/proto/requests.proto
```diff
@@ -50,17 +50,17 @@ service Requests {
 }
 
 message CreateHostRequestReq {
-  int64 host_user_id = 1;
```
aapeliv, line 53: The API surface shouldn't change

aapeliv (changes requested):
some comments. let's just change the column names in the host request table in this PR

app/backend/src/couchers/models/host_requests.py
```diff
@@ -66,33 +66,33 @@ class HostRequest(Base, kw_only=True):
...
+    # reason why the initiator/recipient marked that they didn't meet up
     # if null then they haven't marked it such
-    host_reason_didnt_meetup: Mapped[str | None] = mapped_column(String, default=None)
-    surfer_reason_didnt_meetup: Mapped[str | None] = mapped_column(String, default=None)
+    recipient_reason_didnt_meetup: Mapped[str | None] = mapped_column(String, default=None)
+    initiator_reason_didnt_meetup: Mapped[str | None] = mapped_column(String, default=None)
 
+    initiator: Mapped[User] = relationship(
+        init=False, backref="host_requests_initiated", foreign_keys="HostRequest.initiator_user_id"
+    )
```
aapeliv: Please put this under the public trip id

app/backend/src/couchers/models/host_requests.py
```diff
@@ -38,11 +38,11 @@ class HostRequest(Base, kw_only=True):
     """
```
aapeliv, line 40: let's add a docstring to say initiator = surfer and recipient = host in a normal host req

app/backend/src/tests/test_bg_jobs.py
```diff
@@ -985,84 +985,84 @@ def test_send_host_request_reminders(db, moderator):
         # case 1: pending, future, interval elapsed => notify
         hr1 = create_host_request_by_date(
             session=session,
-            surfer_user_id=user1.id,
```
aapeliv, line 988: Let's revert changes in test

aapeliv (changes requested):

aapeliv (approved):
lgtm

PR #7741, opened 2026-01-25
Go back to truncating the tables in tests

WouldYouKindly:
It's almost 2x faster on my local mac, and about a minute faster in CI (this run took 7:40m as against the usualy 8:40-9:15)

When I implemented the template db approach, I was running it against x86 Postgres on my ARM mac - in this setup template db was a clear win. But after moving to the native ARM build of Postgres, the results are reversed. I guess truncating is more CPU-heavy? 

This also makes it so migration test runs whenever `pg_dump` is installed instead of relying on an env variable.


aapeliv (approved):
lgtm

PR #7743, opened 2026-01-25
Run backend tests in parallel in ci

WouldYouKindly:
Splits backend tests into 4 CI jobs: 1 for the migration test, and 3 for the rest of the tests. The migration test alone takes 2 minutes because of the large timezone-areas.sql file it inserts.

Time goes from 7:40 to ~3 minutes.

This PR also vendors `pytest-split` a pytest plugin to split tests between nodes. The reason is that the PyPI package doesn't support pytest 9, and is in general not maintained - but the code is small, and I don't think it's big deal to vendor it. We can switch to PyPI package if it gets updated.


app/backend/src/tests/pytest_split/algorithms.py
```diff
@@ -0,0 +1,161 @@
+# Vendored from https://github.com/jerry-git/pytest-split
+from __future__ import annotations
```
aapeliv, line 23: could you also mention that it's MIT + year and copyright with the guy's name please?

app/backend/src/tests/.test_durations
```diff
@@ -0,0 +1,709 @@
+{
```
aapeliv, line 1: Are these from the CI machine? Could you include info on how to update this file, please? What happens when we introduce new tests?

app/backend/src/tests/test_db.py
```diff
@@ -158,14 +158,14 @@ def restore_db_after_migration_test(db):
...
-    This test will only run successfully if you have `pg_dump` installed and everything set up, which only happens if the
-    test is being run within Gitlab CI where we do all that setup. So we disable it unless explicitly marked to run.
-
-    Compares the database schema built up from migrations, with the
+    Compares the database schema built up from migrations with the
     schema built by models.py. Both scenarios are started from an
-    empty database, and dumped with pg_dump. Any unexplainable
-    differences in the output are reported in unified diff format and
+    empty database and dumped with pg_dump. Any unexplainable
+    differences in the output are reported in a unified diff format and
```
aapeliv: "unified diff format" is still a thing, so grammatically you shoudl write "reported in unified diff format"

app/.gitlab-ci.yml
```diff
@@ -286,6 +286,59 @@ build:nginx-next:
         - app/.gitlab-ci.yml
 
 test:backend:
+  needs: ["build:backend"]
```
aapeliv, line 289: Could we do the yaml section reuse to reduce repetition?

aapeliv (changes requested):
This is awesome!! Didn't realize you could do this with pytest and gitlab. Thanks for making this happen...

Some small comments for maintainability etc.

app/backend/src/tests/pytest_split/algorithms.py
WouldYouKindly: Done, thanks.

app/backend/src/tests/test_db.py
WouldYouKindly: oh my god, I did it again! Reverted

app/backend/src/tests/.test_durations
WouldYouKindly: There's a comment about it here: https://github.com/Couchers-org/couchers/pull/7743/changes#diff-4673565208f129afedbe14bb2ad42939e8d1463eacc6730435c0fb3e306ea17cR643. Not sure if there's a better place for it?

Most of the time we won't have to do anything, as pytest-split assumes an average run time for the tests it doesn't know about. So unless we add some really long tests, the distribution should be mostly even.

app/.gitlab-ci.yml
WouldYouKindly: I'm kind of reluctant to add more yaml magic for just two cases tbh. And only for 20 lines

aapeliv (approved):
lgtm

Issue #7725, opened 2026-01-22
Backend/i18n: Support mutual language fallbacks


tristanlabelle:
For example: If we don't have a Brazilian Portuguese string, a Brazilian user should see the European Portuguese string if it exists, otherwise English. Same story for the European Portuguese user: if we don't have the European Portuguese string, they should see the Brazilian string if it exists, otherwise English. So the graph of fallbacks isn't a only tree pointing down to English.

Also review the frontend's behavior.

PR #7758, opened 2026-01-26
Frontend/i18n: Allow pt to fallback to pt-BR and es to es-419

tristanlabelle:
European and Brazilian Portuguese are mutually intelligible, so both can fallback to each other in case a string is missing. As per convo with Brazilian translator, this was broken.

We have no locale named `pt-PT`, fix it to `pt`.

Also added mutual fallbacks for Castellano (Spanish from Spain) and Latam Spanish for the same reason.

Note: Backend has a similar issue, tracking as https://github.com/Couchers-org/couchers/issues/7725 . Fix in #7760

## Testing
Before: 
image

After:
image

`pt.json` only has:

```json
{
  "location": "Localização"
}
```

Also verified `pt-BR` to `pt` fallback using the communities pages (where `pt` is ahead).

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/next-i18next.config.js
+++ b/app/web/next-i18next.config.js
@@ -6,8 +6,9 @@ const { allLanguages } = require("./i18n/allLanguages");
 const fallbackLng = {
   default: ["en"],
   "pt-BR": ["pt", "en"],
-  "pt-PT": ["pt-BR", "en"],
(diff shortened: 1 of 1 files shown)


aapeliv (approved):
lgtm

PR #7760, opened 2026-01-27
Backend/i18n: Support mutual locale fallbacks

### LINKED ISSUE #7725: Backend/i18n: Support mutual language fallbacks

tristanlabelle:
Allows `pt-BR` and `pt`, plus `es` and `es-419` to fallback to each other, since the languages are mutually intelligible and `pt-BR`/`es-419` represent a bigger pool of speakers than their undecorated variants (so we're more likely to get them translated first). When first implementing i18next for the backend, I thought fallbacks needed to form a tree with English at the root, so I had regressed this.

Also:
- Renames `constants.py` to `locales.py` (since I'm adding functions)
- Moves `load_i18next` from `localize.py` to `locales.py`, renaming to `load_locales` (keeping `localize.py` only for functions consuming locales)

Backend counterpart to #7758

Closes #7725

## Testing
Added a test for mutual fallbacks.

**Backend checklist**

## For maintainers


app/backend/src/couchers/i18n/locales.py
```diff
@@ -0,0 +1,58 @@
...
+# Locale fallbacks (for those that don't fallback to English).
+# If a string is not found in the requested language, we try the provided one before English
+# Some mutually intelligible language variants fallback to each other.
+_LOCALE_FALLBACKS: dict[str, str] = {
+    "pt-BR": "pt",
+    "pt": "pt-BR",
+    "es-419": "es",
+    "es": "es-419",
+    "fr-CA": "fr",
+}
```
aapeliv, line 19: I would in general like it if this kind of data was separate from the code. E.g. in a JSON file or just a constants file. Likewise with the plural rules but I understand that is way more complicated.

aapeliv (approved):
lgtm, feel free to resolve comment and merge

app/backend/src/couchers/i18n/locales.py
tristanlabelle: If there's a way to share that file between the frontend and backend I agree it's worth it and can follow-up. Is there?

Regarding the plural rules, I believe Unicode CLDR encodes them but it's basically a domain-specific language that needs to have an interpreter. Actually when I pull the CLDR in, I'll use the babel library to consume it and it supports that: https://babel.pocoo.org/en/latest/api/plural.html so I'll be able to remove the manually maintained rules.

PR #7779, opened 2026-01-30
Get profile photos working on mobile

nabramow:
Add native photo functionality for mobile.

You need the whole app setup to test this on mobile, so in terms of testing best to just make sure existing web photo upload works as expected (I already tested myself).


app/mobile/app.config.js
```diff
@@ -61,6 +61,15 @@ export default {
...
     ],
+    [
+      "expo-image-picker",
+      {
+        photosPermission:
+          "Allow Couchers to access your photos to upload to your profile.",
+        cameraPermission:
+          "Allow Couchers to access your camera to take profile photos.",
+      },
+    ],
```
nabramow, line 72: I know we need to localize these eventually but will leave for now as it's more complicated than just adding to existing i18n file since it's a native functionality.

tristanlabelle: Could you file an issue and tag with the "i18n" label? I'll be working through that backlog

PR #7785, opened 2026-01-30
Backend/i18n: Introduce LocalizationContext

tristanlabelle:
Introduces a `LocalizationContext` object that includes the locale and timezone, and makes it the main currency for localization info across the backend.

Context: While working on the backend, I ended up piping those two pieces of information together in several places. I realized that we needed a higher-level object than the locale string. This class would also be the place for a 12/24h-clock preference or any additional regional settings we might expose.

Two pieces of code implemented their own solution to this problem and are now simpler, but have a large diff:

- `couchers.templates.v2.Context` mostly existed to pass the locale+timezone. I refactored it to `Jinja2Template` with an `html` flag and a `render` function taking the `LocalizationContext`, much clearer!
- `couchers.notifications.render_push._Renderer` existed to pass the locale+timezone, I removed the class and passed the `LocalizationContext` to every function instead.

In a subsequent PR, I'll change the `CouchersContext` object to have a `LocalizationContext`.

## Testing
Ran some tests locally, let CI run the rest.

**Backend checklist**

## For maintainers


app/backend/src/couchers/i18n/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    localize_date_from_iso,
+    localize_datetime,
+    localize_time,
+    localize_timezone,
+)
+from couchers.models.users import User
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class LocContext:
```
aapeliv: could we call this something a tad more descriptive? LocalizationContext?

app/backend/src/couchers/i18n/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Specifies regional settings used for localization of strings and date/times.
+    Future settings like 12/24h clock preference would go here as well.
+    """
+
+    EN_UTC: ClassVar[LocContext]
```
aapeliv: it should be `EN_US_UTC` then, no? if this is a full localization context, are we putting not only language but also regional variant in?

app/backend/src/couchers/notifications/background.py
```diff
@@ -30,30 +29,26 @@
...
 from couchers.utils import now
 
 logger = logging.getLogger(__name__)
 
 
 def _send_email_notification(session: Session, user: User, notification: Notification) -> None:
-    rendered = render_email_notification(user, notification)
+    loc_context = LocContext.from_user(user)
+    if not config["ENABLE_NOTIFICATION_TRANSLATIONS"]:
+        loc_context = dataclasses.replace(loc_context, locale="en")
```
aapeliv, line 41: I guess this is fine for now, but I think we shouldn't call it a locale if it's just the language preference? It feels wrong to localize to "english" here?

aapeliv (commented):
this looks overall good

app/backend/src/couchers/notifications/background.py
tristanlabelle: Good thinking. Actually a "naked" language is also considered to be a locale. Locales are meant to be hierarchical, so en-US is a more specific locale than en.

Since we don't have a way for a user to choose a specific locale that would distinguish different types of English, I think it makes sense to say that we only have an "en" locale for all Englishes (if I may say) and no "en-US" locale, even if in practice we use US English. We could refine this if we allow the user to pick more specific English locales later.

app/backend/src/couchers/i18n/__init__.py
tristanlabelle: See my reply on your other comment, "en" is also a locale, considered to be more general than "en-US"

app/backend/src/couchers/i18n/__init__.py
tristanlabelle: Yeah, "loc_context" makes a nicer variable name than class name. I'm open to a better name. I thought LocalizationContext was a little too verbose, L10NContext a little too obscure, and Localizer a little too imperative. .NET calls this CultureInfo which isn't too great either. 

If you have a strong preference I'm happy to follow.

tristanlabelle: Yeah, "loc_context" makes a nicer variable name than class name. I'm open to a better name. I thought LocalizationContext was a little too verbose, L10NContext a little too obscure, and Localizer a little too imperative. .NET calls this CultureInfo which isn't too great either. 

If you have a strong preference I'm happy to follow.

app/backend/src/couchers/i18n/__init__.py
```diff
@@ -0,0 +1,69 @@
+from dataclasses import dataclass
```
kalvdans: I'm against code in `__init__.py` files. Can you place this in a regular module instead?

Rationale: It's easy to introduce circular dependencies, as for example importing `couchers.i18n.i18next` also implicitly imports this file, easily in the long run leading to whole tree being imported when just a single file was needed. The pep8 standards neither recommends nor unrecommends this practice, so it's just my personal experience.

app/backend/src/couchers/i18n/__init__.py
kalvdans: `LocaleAndTimezone` would work for me, as 12/24h clock preference are usually included in the locale concept.

app/backend/templates/v2/generated_html/discussion_create.html
kalvdans, line 236: Why do we have generated html checked in?

app/backend/src/couchers/i18n/__init__.py
tristanlabelle: LocaleAndTimezone is too specific, and hard to make a variable name out of.

.NET's CultureInfo specifies "The information includes the names for the culture, the writing system, the calendar used, the sort order of strings, and formatting for dates and numbers.". We probably won't have most of these but date formats is not completely far fetched.

app/backend/templates/v2/generated_html/discussion_create.html
tristanlabelle: As I understand it, because mjml->html generation is slow and we don't have a build system. This PR is a precursor to my refactoring how we do emails which will eliminate most of this checked-in generated code.

app/backend/src/couchers/i18n/__init__.py
tristanlabelle: Thanks for the explanation, I'll move it to its own file. Python's module system is a little unfortunate, it was nice to expose this at a higher level than the sibling lower level files.

WouldYouKindly: @tristanlabelle you can still import stuff in `__init__.py`, it's a common pattern. See https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/couchers/models/__init__.py

app/backend/src/couchers/i18n/__init__.py
WouldYouKindly: My vote goes to LocalizationContext, it's long, but it's obvious

tristanlabelle: Alright, `LocalizationContext` as the type works for me, I'll just keep `loc_context` for variables for convenience.

app/backend/src/couchers/i18n/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @staticmethod
+    def from_user(user: User) -> LocContext:
+        return LocContext(
+            locale=user.ui_language_preference or "en",
+            timezone=ZoneInfo(user.timezone or "Etc/UTC"),
+        )
+
+
+LocContext.EN_UTC = LocContext(locale="en", timezone=ZoneInfo("Etc/UTC"))
```
WouldYouKindly: I'm a bit confused by this - why do we want it as a class variable?

I would either just make it a constant (`EN_UTC = LocContext(locale="en", timezone=ZoneInfo("Etc/UTC"))`) or move into a classmethod.

app/backend/src/couchers/notifications/background.py
tristanlabelle: I think the reasoning is that even "en-US" is wrong, in that the language will have regional variations within the country. There's no level at which you perfectly capture how people express themselves, thus "en" is also valid.

app/backend/src/couchers/i18n/__init__.py
tristanlabelle: For semantic scoping, the same reason enumerants are class variables of enums and not module-scoped (as in C/C++, which annoys everyone). To me it's standard practice...?

If you'd have pushed a PR with a module-scoped instance as you suggest, I'd have naturally commented: why make it a module-scoped variable when you can scope it to the type it's associated with?

WouldYouKindly: To me it looks suspicious because it modifies the class after it's created. If you want it to be class-scoped, why not just have it as a method? 

`loc_context = LocContext.en_utc()`

tristanlabelle: I'll switch to the method because I don't care that much, but for the convo:

> To me it looks suspicious because it modifies the class after it's created

What's suspicious are Python rules that force that form :P. Consuming code doesn't care that the assignment didn't happen inline.

> If you want it to be class-scoped, why not just have it as a method?

I only needed an instance, not a factory method.

tristanlabelle:
@aapeliv @WouldYouKindly @kalvdans Thanks for all the feedback, I've addressed it in the latest. I also had to rebase. To review the post-rebase delta, focus on the diff from the "PR feedback" commit onward (fa0806d27c664e467d9adfa70a04ebf9a0a929de)

app/backend/src/couchers/notifications/background.py
aapeliv: I guess I'm just thinking ahead a little bit, en-US and en-AU have different time display conventions.

aapeliv (approved):

app/backend/src/couchers/notifications/background.py
tristanlabelle: "en-CA" is different as well! But until the UI allows selecting those, I think using the generic locale makes more sense.

PR #7793, opened 2026-01-31
Optimize web builds

WouldYouKindly:
- Run all web builds on large CI nodes (2-2.5 minutes speedup)
- Merge a couple of RUN commands in Dockerfile.prod to save some bytes
- Clear yarn/next/node_modules cache in Dockerfile, since that cache is never used and makes images bigger. Bigger images -> longer pull times -> longer CI


aapeliv (approved):
lgtm if it works

PR #7794, opened 2026-01-31
Only run protos job when proto files change; otherwise download latest protos from develop

WouldYouKindly:
This should save about 50s on CI runs that don't change protobufs.


aapeliv:
I don't like the idea of not building protos in CI every time. This seems error prone, and the preview domain stuff is mostly for convenience outside CI and viewing CI results, it's not reliable enough I think.

Happy to explore what can be done to speed up this part of the job, but we need to be very careful that it very reliably does things in a reproducible way.

WouldYouKindly:
Closing as this has proven to be trickier than I thought. Race conditions when merging develop and such.

PR #7808, opened 2026-02-01
Backend/i18n: Remove "v2" from template module name

tristanlabelle:
There was probably a v1 some time in the past but by this point it's long gone and I've already removed the "v2" prefix to filter functions. It's time to give this module a new name.

Also cleans up some leftover dead/repeated code from the `LocalizationContext` refactoring that introduced `Jinja2Template`.

I'm not renaming the templates/v2 folder in this change. Should I?

## Testing
mypy, pytest, leaving the rest to CI

**Backend checklist**

## For maintainers


aapeliv (approved):
lgtm

aapeliv:
Yes the first version was plaintext only Jinja templates!

PR #7826, opened 2026-02-04
Frontend: Fix pluralization of chars remaining messages

tristanlabelle:
@vas2525 reported that "Please write at least {{count}} character to unlock messaging and requests." looks like a static string, and shouldn't be pluralizable. It's actually a dynamic string, so I changed the spelling to "more characters" to point it out.

Another similar string used "more characters", but was not pluralizable, so I changed its placeholder to "{{count}}" and changed the key since we don't want to reuse old translations that have incorrect placeholder names.

## Testing
N/A

**Web frontend checklist**

## For maintainers


aapeliv (approved):
lgtm

PR #7835, opened 2026-02-06
Backend: further abstract which notifications can be sent to deleted users

tristanlabelle:
Decouples testing for that with knowing about the account deletion flow, and allows to later expand the set of allowed cases.

## Testing
N/A

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/utils.py
+++ b/app/backend/src/couchers/notifications/utils.py
@@ -4,4 +4,13 @@ enum_from_topic_action: dict[tuple[str, str], NotificationTopicAction] = {
     (item.topic, item.action): item for item in NotificationTopicAction
 }
 
-ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
+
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/notifications/utils.py
```diff
@@ -4,4 +4,13 @@
...
-ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
+
+_DELETED_USER_NOTIFICATIONS = {
+    NotificationTopicAction.account_deletion__start,
+    NotificationTopicAction.account_deletion__complete,
+    NotificationTopicAction.account_deletion__recovered,
+}
+
+
+def can_notify_deleted_user(topic_action: NotificationTopicAction) -> bool:
```
tristanlabelle, line 15: @aapeliv would you have this live in a different file?

aapeliv: I think this is good for now, thanks for following up!

aapeliv (approved):
thanks for following up!

PR #7836, opened 2026-02-06
Backend: Rename NotificationTopicAction.user_editable -> !is_critical

tristanlabelle:
Consolidates user_editable and is_critical into a single concept, using is_critical because it emphasizes the importance of delivering those. I preferred it over is_security because not all critical notifications are security-related.

## Testing
N/A

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/settings.py
+++ b/app/backend/src/couchers/notifications/settings.py
@@ -83,7 +83,7 @@ def set_preference(
     delivery_type: NotificationDeliveryType,
     deliver: bool,
 ) -> None:
-    if not topic_action.user_editable:
+    if topic_action.is_critical:
         raise PreferenceNotUserEditableError()
     current_pref = session.execute(
         select(NotificationPreference)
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/notifications/settings.py
```diff
@@ -364,7 +364,7 @@ def get_user_setting_groups(user_id: int) -> list[notifications_pb2.Notification
                         notifications_pb2.NotificationItem(
                             action=action,
                             description=description,
-                            user_editable=topic_action.user_editable,
+                            user_editable=not topic_action.is_critical,
```
tristanlabelle, line 367: @aapeliv I believe the protobuf field could be renamed without issues since it's not part of the serialized format. Let me know if I you think it's safe/I should.

aapeliv: It's used in `{S,G}etNotificationSettings`, so can't change it without possibly changing frontend, etc.

I think this is fine though since the frontend cares about whether these are user editable or not, it makes sense for that to be the API.

aapeliv (approved):
thanks, lgtm

app/backend/src/couchers/notifications/settings.py
tristanlabelle: I think it'd still be compatible because the protobuf protocol is a binary protocol that relies on the field id numbers, not the field names, but fair enough. I'll keep it as-is.

aapeliv: You could rename it to “not_critical” and the binary serialization would of course be the same. But the frontend builds the types from the same protos so it would complain because it refers to this by name

tristanlabelle: Right, we can change the name on the frontend, but you're right that it wouldn't be binary compatible if we flipped the value.

PR #7841, opened 2026-02-07
Backend/emails: Generalize list-unsubscribe header

tristanlabelle:
The `list-unsubscribe` smtp header was only set for a few notifications. Instead of adding it individually to all notifications, this adds a function to determine if a topic supports topic-key unsubscription, so the list-unsubscribe header logic can be decoupled from the email rendering logic, which is now only about rendering the body and title.

While doing this change I found this possible issue with our unsubscription logic: #7842

## Testing
N/A

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/utils.py
+++ b/app/backend/src/couchers/notifications/utils.py
@@ -6,6 +6,7 @@ enum_from_topic_action: dict[tuple[str, str], NotificationTopicAction] = {
 
 ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
 
+
 def can_unsubscribe_topic_key(topic: str | NotificationTopicAction) -> bool:
     """
     Determines whether a user can unsubscribe from all notification actions
--- a/app/backend/src/couchers/notifications/utils.py
+++ b/app/backend/src/couchers/notifications/utils.py
@@ -5,4 +5,3 @@ enum_from_topic_action: dict[tuple[str, str], NotificationTopicAction] = {
 }
 
 ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
-
--- a/app/backend/src/couchers/notifications/utils.py
+++ b/app/backend/src/couchers/notifications/utils.py
@@ -6,13 +6,3 @@ enum_from_topic_action: dict[tuple[str, str], NotificationTopicAction] = {
 
 ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
 
-
-def can_unsubscribe_topic_key(topic: str | NotificationTopicAction) -> bool:
-    """
-    Determines whether a user can unsubscribe from all notification actions
-    concerning a given topic.
-    """
-    if isinstance(topic, NotificationTopicAction):
-        topic = topic.topic
-    # We currently only support unsubscribing from chat topics
-    return topic == NotificationTopicAction.chat__message.topic
--- a/app/backend/src/couchers/notifications/utils.py
+++ b/app/backend/src/couchers/notifications/utils.py
@@ -5,3 +5,13 @@ enum_from_topic_action: dict[tuple[str, str], NotificationTopicAction] = {
 }
 
 ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
+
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/notifications/utils.py
```diff
@@ -5,3 +5,14 @@
...
+
+def can_unsubscribe_topic_key(topic: str | NotificationTopicAction) -> bool:
+    """
+    Determines whether a user can unsubscribe from all notification actions
+    concerning a given topic.
+    """
+    if isinstance(topic, NotificationTopicAction):
+        topic = topic.topic
+    # We currently only support unsubscribing from chat topics
+    return topic == NotificationTopicAction.chat__message.topic
```
aapeliv: this belongs in the unsub/quick links area of code.

app/backend/src/couchers/notifications/utils.py
```diff
@@ -5,3 +5,14 @@
...
 
 ACCOUNT_DELETION_TOPIC = NotificationTopicAction.account_deletion__start.topic
+
+
+def can_unsubscribe_topic_key(topic: str | NotificationTopicAction) -> bool:
+    """
+    Determines whether a user can unsubscribe from all notification actions
+    concerning a given topic.
+    """
+    if isinstance(topic, NotificationTopicAction):
```
aapeliv: why would this be a str? i don't see it in the code

aapeliv (changes requested):
yeah completely unclear why only some badge notifs had this header... 🙃

app/backend/src/couchers/notifications/utils.py
tristanlabelle: I moved it but I think it makes less sense there. Is there a pattern for where to place functions that support enums from the model? I'm not a super fan of making it a property of the enum either but that's an option.

app/backend/src/couchers/notifications/utils.py
tristanlabelle: It's not used. I prefer coding to minimal inputs and you don't need an action to know whether you can unsubscribe to a topic, so the function only needs a topic to fulfill its contract. I overloaded on NotificationTopicAction for convenience.

app/backend/src/couchers/notifications/utils.py
aapeliv: `respond_quick_link` is the counterpart to that; it's good to have them next to each other, so when you go change it, you don't accidentally change one and not the other.

We can down the line refactor this to work better, so instead of a bunch of manual if-else statements, we get some real way to define "this topic supports unsub from `topic_key`" with the unsub SQL, etc. It would be good to support this for more notif types as well.

I guess we also have `settings.py`. So now there's three places where we define notifs info. It would probably be good to get rid of everything we can from `NotificationTopicAction` and put all the real definition somewhere else in one place.

tristanlabelle: > respond_quick_link is the counterpart to that; it's good to have them next to each other, so when you go change it, you don't accidentally change one and not the other.

However generating quick links and responding to them live in different parts of the stack and have different dependencies. I meant to move `respond_quick_link` to a different file because I tracked it down as part of a circular dependency chain that requires a local import with `# Import here to avoid circular dependency`.

I think Session-manipulating code and more abstract code should live separately.

tristanlabelle:
@aapeliv Published the PR. I'm not sure which of your comments are still blockers, could you point them out?

I think we agree that we could reshuffle a few files' functions and I'd be happy to tackle that but it's outside of the scope of this PR.

aapeliv (approved):
lgtm

app/backend/src/couchers/notifications/utils.py
aapeliv: I think the most important part is to make it easier to maintain correctness. Having these in very separate files makes it so that a human working on it (e.g. adding a new notification type as part of something, so might not have *all* the context on the notifs), is likely to do the correct thing in a way that we will easily catch in code review.

I agree that it's a bit messy now, I think if we wish to have both the solution is to refactor these into a cleaner form that gets us both independence from sessions as well as easy to work on correctly.

tristanlabelle: One other way to achieve this is to use a match statement and list all notifications so mypy will complain if new enumerants are added and not covered in the statement, but that's a little heavy!

PR #7849, opened 2026-02-08
Frontend/profile: Switch pronouns to text input for i18n

### LINKED ISSUE #7438: Pronouns cannot be mapped back to radio buttons on the frontend
Pronouns are stored as string in the backend, e.g. "He / him".

After changing the display language, the frontend tries to map back this string to a radio button, since it corresponds to neither "él" nor "ella", it appears as a custom gender:

Image

Pronouns are inherently non-localizable (some languages won't differentiate at all). The pronoun field should be a an optional textfield only. If left empty, the frontend can decide to either not show pronouns or show pronouns based on the gender.

Pronouns should only be used in the profile as strings cannot adapt to support all pronouns. But note that this may cause users to find that they are not referred to using their chosen pronoun in app strings.

tristanlabelle:
Fixes #7438, see that bug for why a radio button is problematic for internationalization.

## Testing
image
image
image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/edit/EditProfile.tsx
+++ b/app/web/features/profile/edit/EditProfile.tsx
@@ -783,59 +783,11 @@ export default function EditProfileForm() {
               </SectionSubtitle>
 
               <FieldGroup>
-                <Typography variant="h3" gutterBottom>
-                  {t("profile:edit_profile_headings.pronouns")}
-                </Typography>
-                <Controller
-                  control={control}
+                <StyledProfileTextInput
+                  id="pronouns"
+                  {...register("pronouns")}
+                  label={t("profile:edit_profile_headings.pronouns")}
                   defaultValue={user.pronouns}
-                  name="pronouns"
-                  render={({ field }) => {
-                    const other =
-                      field.value === t("profile:pronouns.woman") ||
-                      field.value === t("profile:pronouns.man")
-                        ? ""
-                        : field.value;
-                    return (
-                      <RadioGroupContainer>
-                        <StyledRadioGroup
-                          {...field}
-                          row
-                          aria-label={t(
-                            "profile:edit_profile_headings.pronouns",
-                          )}
-                          name="pronouns"
-                          value={field.value}
-                          onChange={(_, value) => field.onChange(value)}
-                        >
-                          <FormControlLabel
-                            value={t("profile:pronouns.woman")}
-                            control={<Radio />}
-                            label={t("profile:pronouns.woman")}
-                          />
-                          <FormControlLabel
-                            value={t("profile:pronouns.man")}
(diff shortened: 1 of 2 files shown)


aapeliv (commented):
I agree that the current setup is silly as it's untraslateable. As discussed it's tricky.

What I'm worried about now is that frankly I think there's a lot of people who don't really know what "pronouns" are: if it has some default options, then at least lots of (English pseaking) users will kind of know what it's about. Combined with the i18n woes, I would frankly try to come up with some way to de-emphasize this in the UI.

I also think it would make more sense to have this input be next to the name.

I think my recommendation for now would be to move this next to the name field and maybe mark it as "(optional)" or put in some example "e.g. she/her" or similar, so it's not confusing to someone who has little clue what this is about?

tristanlabelle:
@aapeliv Totally agree with your take. I think custom pronouns are quite a US-centric (maybe English-centric) issue, and honestly not really my battle if I have to choose between it and internationalization. I'll give a try at de-emphasizing and adding an example (both good ideas), and we'll see what it looks like.

nabramow:
@tristanlabelle Are you still working on this? If not should we close? Just trying to get the PRs back under control 😅.

tristanlabelle:
@nabramow The issue is still relevant so I'll get back to it eventually. I got stumped by react shenanigans. Happy to let a frontend dev pick it up too.

nabramow:
> @nabramow The issue is still relevant so I'll get back to it eventually. I got stumped by react shenanigans. Happy to let a frontend dev pick it up too.

Feel free to post in #engineering with some context we have a few frontend people hanging around. Maybe @darrenvong or @kevinortiz43 also wants to take a look?

I'm still a little confused from this PR what we landed on specifically so maybe we can write that down or add it to the ticket this is related to?

darrenvong:
From the screenshot, I got the impression that it's already working functionally? What's left to fix? The de-emphasis styling, or is there something else I'm missing?

tristanlabelle:
Oh sorry I got confused with another task. This one doesn't have react problem, I just needed to follow-up on Aapeli's feedback. Carry on!

nabramow:
@tristanlabelle What's up with this one? In a previous comment you said you're "happy to let a frontend dev pick this up". No one has and I don't think this is a priority so I will close unless you have a strong objection.

tristanlabelle:
@nabramow No need to close, I just finally followed up on @aapeliv 's review and this is ready for ✅

nabramow (approved):
This looks fine to me and works well. Not sure if @aapeliv wants to take a look though since he did the initial review.

nabramow:
I'd say go ahead and merge this @tristanlabelle !

PR #7862, opened 2026-02-09
Auto format action

aapeliv:

New and re-implemented! `auto-format.yml` workflow that triggers when you open or push to a PR:

* Formats backend code (runs `make format` via `ruff`) and commits changes
* Formats frontend code (runs `yarn format`) and commits changes
* Auto-generates Alembic migrations when model files change but no migration is included
* Regenerates HTML email templates from MJML sources when templates change

This uses path-based change detection (dorny/paths-filter) to only run relevant jobs, though that's not perfect.


tristanlabelle (approved):
Nice GHA! You avoided the pitfalls I'm aware of.

The jobs are all going to race to push commits, causing them all to reset, but it's better than not having them.

There's a way to extract the github app user id to a reusable action but you need two checkouts to get it right. I can look into that afterwards if you'd like.

nabramow:
Hey @aapeliv can you please add a description to explain more what this is for how this will affect the dev flow? I'd like to have a better idea before I approve. Also good to have it historically in the PR to tell what it does later.

nabramow (approved):
LGTM!

WouldYouKindly:
@aapeliv sorry to chime in so late :) I think generating migrations is a bit dangerous - what if I edit a model file by accident and then fail to notice the generated migration? I'd rather fail the CI in case a diff was detected.

PR #7865, opened 2026-02-09
Implement Slack notifications in backend

aapeliv:

## Testing

I tested locally that this slack function works.

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/servicers/donations.py
```diff
@@ -182,6 +183,15 @@ def Webhook(
                         receipt_url=receipt_url,
                     ),
                 )
+
+                # Recurring donations go through Stripe invoices, one-time don't
+                is_recurring = data_object.get("invoice") is not None
+                donation_type = "recurring" if is_recurring else "one-time"
+                user_link = urls.user_link(username=user.username)
+                send_slack_message(
```
tristanlabelle: What's the consequence of an exception bubbling from here? Should we swallow it / report to sentry since it's just a notification?

app/backend/src/couchers/slack.py
tristanlabelle, line 1: I don't think `helpers`/`utils` provides useful namespacing, so I suggest moving one directory up.

app/backend/src/tests/test_donations.py
```diff
@@ -329,6 +329,66 @@ def test_merch_invoice_flow_nonexistent_user(db, monkeypatch):
...
+    # Initiate a one-time donation
+    with donations_session(token) as donations:
+        with patch("couchers.servicers.donations.stripe") as mock:
+            mock.Customer.create.return_value = type("__MockCustomer", (), {"id": "cus_Pv4uq0gT0rDZWN"})
+            mock.checkout.Session.create.return_value = type("__MockCheckoutSession", (), one_time_STRIPE_SESSION)
+            donations.InitiateDonation(donations_pb2.InitiateDonationReq(amount=100, recurring=False))
+
+    # Fire the charge.succeeded webhook and check Slack message
+    with patch("couchers.servicers.donations.send_slack_message") as mock_slack:
+        fire_stripe_event("evt_3P5El3IfR5z29g5k0TLWlfHq")
```
tristanlabelle, line 355: Could you document the magic base64 strings in this test, or if they're not important change them to something that looks more like a test value like "evt_dummy". Same for the test below.

app/backend/src/tests/test_slack.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_send_slack_message_disabled():
+    with patch("couchers.helpers.slack.requests.post") as mock_post:
+        send_slack_message("test-channel", "Test message")
+        mock_post.assert_not_called()
+
+
+def test_send_slack_message_enabled():
+    config["SLACK_ENABLED"] = True
```
tristanlabelle, line 23: For my info, the config is automatically reset after tests? Is this what the fixture above does?

app/backend.dev.env
```diff
@@ -91,3 +91,8 @@ MODERATION_BOT_USER_ID=1
 
 # Enable development APIs (e.g., SendDevPushNotification)
 ENABLE_DEV_APIS=1
+
+# Slack notifications
+SLACK_ENABLED=0
+SLACK_BOT_TOKEN=xoxb-...
```
tristanlabelle, line 97: Intentional? That doesn't look like a real token.

tristanlabelle (commented):
Cool, just some comments about clarity and the token that looks mistaken, so not approving yet.

app/backend/src/couchers/servicers/donations.py
aapeliv: Yeah i was thinking the same. Probably a good idea to wrap this one in sentry exception catcher

app/backend/src/tests/test_donations.py
aapeliv: These are real stripe (test account) events that I captured at some point. It's just the endless ID soup that stripe lives in

app/backend/src/couchers/slack.py
aapeliv: Sure can move up. Just didn’t want to pollute the top level too much

app/backend.dev.env
aapeliv: This is the form of the slack secret (starts with those chars)

aapeliv:
@claude address the review comments

tristanlabelle (approved):

nabramow:
Ooo nice I'm happy if you merge this and we can see how it goes for a bit, then I'll deactivate the current one in the channel to remove clutter.

app/backend/src/couchers/servicers/donations.py
```diff
@@ -182,6 +184,18 @@ def Webhook(
...
                 )
+
+                # Recurring donations go through Stripe invoices, one-time don't
+                is_recurring = data_object.get("invoice") is not None
+                donation_type = "recurring" if is_recurring else "one-time"
+                user_link = urls.user_link(username=user.username)
+                try:
+                    send_slack_message(
+                        config["SLACK_DONATIONS_CHANNEL"],
+                        f"Donation received: ${amount} ({donation_type}) from <{user_link}|{user.name}>",
```
nabramow, line 195: What about merch purchases without a donation, how will that look? And merch with? Can we differentiate to make those ones clear?

aapeliv: it won't go through this code path

app/backend/src/tests/test_slack.py
aapeliv: That's what `testconfig` does, I believe

PR #7888, opened 2026-02-13
fix: clear undelete fields in admin RecoverDeletedUser

### LINKED ISSUE #7887: fix recover deleted account api
Problem: Currently when on the admin console page I try to call the API org.couchers.admin.Admin.RecoverDeletedUser to recover a deleted account, I get an error and the user is not recovered. 

Solution: it is unclear if this is a problem with the API itself or with the console, but considering attempts to call the API on console returns an error, I suspect the API is the problem. In theory it should work since we give users the opportunity to undo their deletion by providing a link to do so in the deletion confirmation email we send them after they confirm their deletion request.

jesseallhands:
Fixes #7887

## Summary

- Clear `undelete_token` and `undelete_until` when admin recovers a deleted user, fixing a database constraint violation
- Add test covering admin recovery after user-initiated deletion flow

## Root Cause

The admin `RecoverDeletedUser` endpoint set `is_deleted=False` but did not clear `undelete_token` and `undelete_until`, which violated the `undelete_nullity` database CHECK constraint.

## Test Plan

- [ ] Verify `test_RecoverDeletedUser_after_user_initiated_deletion` passes
- [ ] Verify existing `test_DeleteUser` still passes
- [ ] Test admin RecoverDeletedUser on the console against a user-deleted account

### DIFF
--- a/app/backend/src/couchers/servicers/admin.py
+++ b/app/backend/src/couchers/servicers/admin.py
@@ -391,6 +391,8 @@ class Admin(admin_pb2_grpc.AdminServicer):
         if not user:
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "user_not_found")
         user.is_deleted = False
+        user.undelete_token = None
+        user.undelete_until = None
         return _user_to_details(session, user)
 
     def CreateApiKey(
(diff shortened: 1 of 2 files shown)


aapeliv (approved):
lgtm

PR #7909, opened 2026-02-16
disable push notifications for banned useer

Krishnag09:
Added a `disable_push_notifications_for_user`. method to be called to disable notification when user is banned admin action is performed. 

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


aapeliv (changes requested):
i think we should do this for all deleted/banned users, right? so maybe a bg job that updates this?

Krishnag09:
> i think we should do this for all deleted/banned users, right? so maybe a bg job that updates this?

i designed the flow so the method to disable notification is called every time a user is banned by admin.  do we need a bg job in addition?

Krishnag09:
@aapeliv - few points. I think you re saying we add this as a job to app/backend/src/couchers/jobs/definitions.py 

two questions about this
- is this solely to bringing existing users to parity?
-  if not what interval do you want to run and any patterns with existing jobs you want me to follow?

how abt this alternate approach to cover all scenarios
- the notifications pause time is updated to now for `DeleteUser`,`BanUser`  and then unpaused for `UnbanUser`, `RecoverDeletedUser`

tristanlabelle (approved):

nabramow:
Would love @aapeliv eyes on this

nabramow:
Closing as checked with @aapeliv who said push notifs already block on banned users.

Krishnag09:
this is the second time in a row my work has been discarded.

PR #7916, opened 2026-02-16
Clarify signup email confirmation page title

aapeliv:
Changed the signup verification page title from "Almost there, one last thing!" to "One last thing: confirm your email." This makes it immediately clear what the user needs to do, instead of keeping it a mystery.

## Testing
Locale string change only — no functional change.

**Web frontend checklist**

## For maintainers


tristanlabelle (approved):
Is it the last thing? I thought email was part one and entering account info followed.

aapeliv:
Yeah this is the last page. Whenever I watch people sign up, they always pause for a second to figure out what this page is about. So i think we should put the action item front and center in the page h1.

PR #7919, opened 2026-02-17
Fix ReferencesView scroll on load more

### LINKED ISSUE #4397: Clicking on "show more references" returns user to top of list
When reading references on someone's profile, clicking on the "show more references" button takes you back to the top of the references list (both on desktop and mobile). This is annoying if the person has a lot of references and you want to read through a lot of them because every time you have to scroll down to the bottom of the list again. 

When clicking on "show more references", it should not return the user to the top of the list. It should keep them in the same spot in the list.

nabramow:
Closes #4397 

Previously if a user was browsing a profile with many references and clicked "load more" it scrolled back to the top, which was annoying. Now it should stay in place.

## Testing
* Change REFERENCES_PAGE_SIZE to something smaller in app/web/service/references.ts
* Go to a profile with more than that references, like Aapeli, me or Jesse on stage
* Load more and see it doesn't jump all the way to the top.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/view/ReferencesView.tsx
+++ b/app/web/features/profile/view/ReferencesView.tsx
@@ -9,6 +9,7 @@ import { RpcError } from "grpc-web";
 import { useTranslation } from "i18n";
 import { GLOBAL, PROFILE } from "i18n/namespaces";
 import { ListReferencesRes } from "proto/references_pb";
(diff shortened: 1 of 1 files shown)


aapeliv (approved):
thanks 🙏

PR #7923, opened 2026-02-19
Backend/i18n: localize_with_markup

tristanlabelle:
Some of our localizable strings for emails will contain simple markup like `<b>` tags. When dealing with markup, it's important to know what is trusted and what is untrusted markup. Markup in translations is always trusted, but interpolated placeholder values are typically not.

To safely support this distinction, this introduces two modes for localizing strings. The default returns a string, which is inherently untrusted, so it can contain markup because rendering code will be responsible for escaping it. `localize_with_markup` returns a `markupsafe.Markup` object, which is trusted markup, so any untrusted placeholders should be escaped during substitution.

Also renames i18n's `Language` to `Translation` and `language_code` to `locale` to follow the rest of the code.

## Testing
Added a test.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/i18n/i18next.py
+++ b/app/backend/src/couchers/i18n/i18next.py
@@ -165,7 +165,7 @@ class StringTemplate:
             if segment.is_variable:
                 if substitutions is not None and segment.text in substitutions:
                     value = substitutions[segment.text]
-                    if with_markup and not isinstance(value, Markup):
+                    if with_markup and not isinstance(value, Markup) and not isinstance(value, int):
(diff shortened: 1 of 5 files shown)


aapeliv (approved):
i had a quick look and it seem sensible. please confirm this is safe against XSS and other types of injection attacks. user-submitted stuff should be encoded and shown as plaintext only.

tristanlabelle:
@aapeliv Yep this change is meant to help with XSS by clearly identifying what is markup-safe and what isn't using the type system. The tests validate that and it'll be further illustrated by the future email rendering code using this.

PR #7925, opened 2026-02-19
Frontend/i18n: Leverage Intl.RelativeTimeFormat in timeAgo

### LINKED ISSUE #7883: Frontend/i18n: Migrate relative times to Intl.RelativeTimeFormat
Instead of defining our own `in {{count}} minutes` strings, we can use `Intl.RelativeTimeFormat`.

It's built in and "widely available": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

https://github.com/Couchers-org/couchers/blob/79a77d7834f1d35a91e15922a2eab486db783a48/app/web/resources/locales/en.json#L236

tristanlabelle:
Instead of managing our own collection of relative time strings, we can leverage the built-in `Intl.RelativeTimeFormat`, which gives us all major languages for free and can be more precise than ours (e.g. "yesterday", etc.). It also supports relative future times that we could use for events (in 1 day, etc.)

Unfortunately it doesn't fit as well in the `t` translation function pattern, since we need the locale value for the API, so I have to pass around the two values.

Also replaces the `fuzzy` parameter with `minimumUnit`, which is what it expressed.

Fixes #7883

## Testing
English:
image

German (note "vorgestern", the day before yesterday):
image

**Web frontend checklist**

## For maintainers


app/web/.env.development.local2
```diff
@@ -0,0 +1,11 @@
+NEXT_PUBLIC_COUCHERS_ENV=dev
```
aapeliv: accidentally added

aapeliv (changes requested):
approve after removing this file

app/web/.env.development.local2
tristanlabelle: Oopsie! Thanks for catching. Deleted it.

tristanlabelle:
@aapeliv Stray file removed. Ready for your reapproval

aapeliv (approved):

PR #7929, opened 2026-02-19
feat: add Slack notification for merch shop purchases

### LINKED ISSUE #7462: Send slack messages from backend rather than via webhooks
Right now all the stripe alerts on Slack are mixed together as we can't differentiate with the webhooks. 

Let's send the Slack messages ourselves via the backend so we can control the channel and message.

aapeliv:
Closes #7462

Send merch purchase notifications to a dedicated Slack channel (`SLACK_MERCH_CHANNEL`) instead of relying on Stripe webhooks that mix all alerts together. For known users, the message includes a link to their profile; for unknown users, it includes their email.


app/backend.dev.env
```diff
@@ -96,3 +96,4 @@ ENABLE_DEV_APIS=1
 SLACK_ENABLED=0
 SLACK_BOT_TOKEN=xoxb-...
 SLACK_DONATIONS_CHANNEL=donations
+SLACK_MERCH_CHANNEL=merch
```
nabramow, line 99: @aapeliv We don't have a #merch channel yet though! Will you create one?

aapeliv: Yeah it will go to another channel in reality; this is just the env var

PR #7933, opened 2026-02-19
Translations update from Couchers Translate

CouchersBot:
Translations update from [Couchers Translate](https://translate.couchershq.org) for [Couchers/Web app - Auth](https://translate.couchershq.org/projects/couchers/web-app-auth/).

Current translation status:

Weblate translation status


app/backend/src/couchers/i18n/locales/ca.json
tristanlabelle, line 1: Blocking to investigate empty string translations

tristanlabelle (approved):

app/backend/src/couchers/i18n/locales/ca.json
tristanlabelle: Alright I got it to only include some reformatting changes which I'm okay with happening since we'll want them eventually anyways.

PR #7936, opened 2026-02-20
Adjust translation string for profile photo

nabramow:
Per @c-kreuz 's request, making this translation string more clear.


aapeliv (changes requested):
you'll need to change the place where it's used though, right?

nabramow:
> you'll need to change the place where it's used though, right?

lol whoops I was coding before coffee. Fixed.

aapeliv (approved):
you'll need to change the place where it's used though, right?

PR #7937, opened 2026-02-20
Add missing is_archived backend

nabramow:
I was working on the frontend and realized I needed this in the backend. Doing it in a separate PR so the frontend can stay only-frontend.

Part of #6528


aapeliv (dismissed):
lgtm

aapeliv (approved):

PR #7954, opened 2026-02-22
Fix missing translation errors from ImageInput

nabramow:
Fixing the missing translation string errors for ImageInput. This was happening as ImageInput was importing from the PROFILE namespace but it's also uses on EditEventPage which is in the COMMUNITIES namespace.

I moved the translations into GLOBAL since this component is reused in a few places.


aapeliv (dismissed):

aapeliv:
actually maybe wait for @tristanlabelle to comment

app/web/resources/locales/en.json
```diff
@@ -41,6 +41,9 @@
   "confirm": "Confirm",
   "copy": "Copy",
   "copied": "Copied to clipboard",
+  "select_an_image": "Select an image",
+  "click_replace_image": "Click to replace image",
+  "couldnt_read_file": "Couldn't read file",
```
tristanlabelle: Since we're forcing relocalization anyways, we can take advantage to improve the keys (scope within context + specify the type)

```json
  "image_input": {
    "a11y": "Select an image",
    "tooltip": "Click to replace image",
    "read_file_error": "Couldn't read file"
  },
```

tristanlabelle (dismissed):
Fix is fine but I'm a little concerned we have a wider transitivity problem where some piece of code needs to know the set of all string namespaces that will be read.

tristanlabelle (dismissed):

nabramow:
> Fix is fine but I'm a little concerned we have a wider transitivity problem where some piece of code needs to know the set of all string namespaces that will be read.

Yeah we have an issue with the GLOBAL namespace being massive and too much in there for this reasson, but it's not a quick thing to solve so 🤷‍♀️. We need generally to do an audit of the translation files. Move things out of global that don't need to be there etc. etc.

PR #7964, opened 2026-02-23
Backend/emails: Localizable email framework

tristanlabelle:
Implements a framework for building localizable emails in code that can be styled as HTML or plaintext, implementing the "host request received" email as a proof of concept. This is not hooked up to production yet and can be tested using a backend script. It's a pivot from #7654 that used jinja2 with parameterizable imports to template as html or plaintext.

Main ideas:
- Low level: Four style-agnostic email "blocks" that all our emails are built of (paragraph, user banner, quote, action).
  - Single mjml file defining templates for each block, keeping all the styling in mjml
  - Alternative plaintext renderer that uses the same model data (no more duplicated strings) 
- High level: Per-email model class defining all required args and how to map them to blocks and subject/preview lines.

This will scale to other emails by defining other classes `HostRequestReceived`, e.g. about 40-50 LOCs per email kind (compare with the txt/mjml/html files we previously had for each email)

There'll be a few things to iron out like having side-by-side buttons instead of one below the other and localizing more footer strings, but nothing scary there.

Related to #7368

## Testing
Run `uv run dump-emails --outdir <dir>` and look at the resulting txt/html.

Diff with currently generated emails: https://github.com/Couchers-org/couchers/commit/60d3f825ccf8e90534d4d07fce3add1ec6fe591a

Note that this is comparing a template with rendered value, and some of the differences are due to the text in the plaintext/html versions previously differing, whereas the new system uses the same text in both.

**Backend checklist**

## For maintainers


aapeliv (approved):
Based on a quick look, this looks good to me. I would love it if we could eventually do the dummy email even deeper (e.g. real host request and user info objects) to make it more end-to-end.

tristanlabelle:
> I would love it if we could eventually do the dummy email even deeper (e.g. real host request and user info objects) to make it more end-to-end.

@aapeliv If we want to push dummyness at a higher level we'd need to be able to generate dummy Notification model objects, that can then be lowered to the corresponding dummy email and dummy push notifications.

PR #7965, opened 2026-02-23
Backend/notifs/i18n: Remove untranslatable strings from json

tristanlabelle:
Translators commented that strings that only have a placeholders are useless to translate. Fair enough, we can lift them into the code. However there was a reason I did it that way: it made it obvious that all strings were accounted for by keeping them in the json file. Now in the code it's not obvious why sometimes ios_title is specified, and other times not, for example. Is this better?

## Testing
N/A

**Backend checklist**

## For maintainers

- [ ] Maintainers can merge this PR for me

### DIFF
--- a/app/backend/src/couchers/notifications/render_push.py
+++ b/app/backend/src/couchers/notifications/render_push.py
@@ -306,6 +306,7 @@ def _render_discussion__create(
     return _get_content(
         NotificationTopicAction.discussion__create,
         loc_context,
+        ios_subtitle=data.discussion.title,
         substitutions={
             "title": data.discussion.title,
             "user": data.author.name,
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/notifications/render_push.py
tristanlabelle, line 1: Curious for reviewers' take on translator convenience vs code correctness risk of this change.

aapeliv:
We could also just bulk-copy these into every language's translation strings, right?

tristanlabelle:
There's no easy way to bulk-copy this string to other languages, weblate doesn't have an "untranslatable flag", and there's an benefit to keeping those in `en.json`: translators can look at nearby strings to see all the strings that make up a notification (title, subtitle, body), giving them more context. I'll drop this.

PR #7967, opened 2026-02-23
Update mailing address

aapeliv:


tristanlabelle (approved):
We'll soon just need to do it once!

tristanlabelle:
@aapeliv Fixing the test here: https://github.com/Couchers-org/couchers/pull/7969

PR #7974, opened 2026-02-25
Backend/i18n: Localize notification descriptions.

tristanlabelle:
Our backend has an API to describe notifications but it is not localized, so the frontend ignores the descriptions and localizes it itself. This localizes the descriptions on the backend, opening the path to future frontends eventually switching over.

The backend should own this localization, since it is the source of truth for notifications. But also, emails can reuse these descriptions in their footer, for the link that says "do not notify when <topic-action-description>".

I copied the strings from the frontend with two changes: I filled in missing strings (non-editable notifications that the frontend won't show but the API reports), and punctuated each one with a period.

I ignored the category descriptions since the frontend ignores the backend's categories and reorganizes notifications according to its own categories.

## Testing
Added tests

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/settings.py
+++ b/app/backend/src/couchers/notifications/settings.py
@@ -11,7 +11,7 @@ from couchers.models import (
     NotificationPreference,
     NotificationTopicAction,
 )
-from couchers.notifications.locales import get_notifs_i18next
+from couchers.notifications.utils import get_topic_action_description
 from couchers.proto import notifications_pb2
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/notifications/settings.py
```diff
@@ -113,40 +114,40 @@ def set_preference(
                 "host_request",
                 "Host requests",
```
aapeliv, line 115: what about these strings?

aapeliv (approved):
as you probably know these strings are not used in the frontend; i think. also added one comment on some other non-translated strings.

app/backend/src/couchers/notifications/settings.py
tristanlabelle: See the PR description -- the frontend already ignores these categories in favor of its own, so it's unclear clients have any use for the current backend categorization.

PR #7975, opened 2026-02-25
Backend/i18n: Give CouchersContext a LocalizationContext

tristanlabelle:
Instead of `CouchersContext.ui_language_preference`, expose a `LocalizationContext`, which also includes the time zone, so all code can appropriately localize dates. Also avoid having to deal with nulls everywhere: if we don't have a value (e.g. media server), fallback to `LocalizationContext.en_utc()`. Finally, remove `get_localized_string`, letting code use LocalizationContext's methods directly.

## Testing
Ran backend locally, created a new account, navigated around.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -195,10 +195,7 @@ def make_interactive_context(
     )
 
 
-def make_one_off_interactive_user_context(
-    couchers_context: CouchersContext,
-    user_id: int
-) -> CouchersContext:
+def make_one_off_interactive_user_context(couchers_context: CouchersContext, user_id: int) -> CouchersContext:
     return CouchersContext(
         is_interactive=True,
         grpc_context=couchers_context._grpc_context,
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -2,7 +2,7 @@ from typing import NoReturn, cast
 
 import grpc
 
-from couchers.i18n.localize import localize_string
+from couchers.i18n import LocalizationContext
 
 
 class NonInteractiveContextException(Exception):
@@ -66,7 +66,7 @@ class CouchersContext:
         user_id: int | None,
         is_api_key: bool | None,
         token: str | None,
-        ui_language_preference: str | None,
+        localization: LocalizationContext,
         sofa: str | None = None,
     ):
         """Don't ever construct directly, always use the `make_*_context_` functions!"""
@@ -74,7 +74,7 @@ class CouchersContext:
         self._user_id = user_id
         self._is_api_key = is_api_key
         self.__token = token
-        self.__ui_language_preference = ui_language_preference
+        self.__localization = localization
         self._sofa = sofa
         self.__is_interactive = is_interactive
         self.__logged_in = self._user_id is not None
@@ -103,20 +103,6 @@ class CouchersContext:
     def is_logged_out(self) -> bool:
         return not self.__logged_in
 
-    def get_localized_string(self, key: str, *, substitutions: dict[str, str | int] | None = None) -> str:
-        """
-        Get a localized string using the user's language preference.
-        Falls back to the default language if no preference is set.
-
-        Args:
-            key: The key for the specific string
-            substitutions: Dictionary of variable substitutions for the string (optional)
-
-        Returns:
-            The translated string with substitutions applied
-        """
-        return localize_string(self.__ui_language_preference, key, substitutions=substitutions)
-
     def abort(self, status_code: grpc.StatusCode, error_message: str) -> NoReturn:
         """
         Raises an error that's returned to the user
@@ -138,7 +124,7 @@ class CouchersContext:
         else:
             context = cast(grpc.ServicerContext, self._grpc_context)
             # Get the translated error message using the user's language preference
-            error_message = self.get_localized_string(f"errors.{error_message_id}", substitutions=substitutions)
+            error_message = self.localization.localize_string(f"errors.{error_message_id}", substitutions=substitutions)
             context.abort(status_code, error_message)
 
     def set_cookies(self, cookies: list[str]) -> None:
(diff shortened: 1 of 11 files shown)


app/backend/src/couchers/context.py
tristanlabelle, line 1: Pending local testing

app/backend/src/couchers/context.py
```diff
@@ -204,22 +190,19 @@ def make_interactive_context(
         user_id=user_id,
         is_api_key=is_api_key,
         token=token,
-        ui_language_preference=ui_language_preference,
+        localization=localization or LocalizationContext.en_utc(),
```
aapeliv: I don't think we should have defaults in these functions here

app/backend/src/couchers/servicers/public.py
```diff
@@ -193,7 +193,9 @@ def GetPublicUser(
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "user_not_found")
 
         if user.public_visibility == ProfilePublicVisibility.full:
-            return public_pb2.GetPublicUserRes(full_user=user_model_to_pb(user, session, make_logged_out_context()))
+            return public_pb2.GetPublicUserRes(
+                full_user=user_model_to_pb(user, session, make_logged_out_context(localization=context.localization))
```
aapeliv, line 197: Unrelated to this PR, but I think we don't need to make a new context here; the one passed into this RPC should work just fine

app/backend/src/couchers/context.py
```diff
@@ -230,27 +213,27 @@ def make_media_context(grpc_context: grpc.ServicerContext) -> CouchersContext:
...
-def make_background_user_context(user_id: int) -> CouchersContext:
+def make_background_user_context(user_id: int, localization: LocalizationContext | None = None) -> CouchersContext:
     return CouchersContext(
         is_interactive=False,
         user_id=user_id,
         is_api_key=None,
         grpc_context=None,
         token=None,
-        ui_language_preference=None,
+        localization=localization or LocalizationContext.en_utc(),
```
aapeliv, line 227: We should really be picking this up from the user's prefs. Otherwise user will get e.g. notifications/emails in English?

app/backend/src/couchers/interceptors.py
```diff
@@ -128,13 +131,17 @@ def _try_get_and_update_user_details(
 
         session.commit()
 
+        loc_context = LocalizationContext(
+            locale=user.ui_language_preference or DEFAULT_LOCALE, timezone=ZoneInfo(user.timezone or "Etc/UTC")
```
aapeliv: Yeah I think this is the place to set the defaults, and pass in a non-None context to the context making functions.

app/backend/src/couchers/interceptors.py
```diff
@@ -361,8 +373,8 @@ def function_without_couchers_stuff(req: Message, grpc_context: grpc.ServicerCon
                     couchers_context.set_cookies(
                         create_session_cookies(auth_info.token, auth_info.user_id, auth_info.token_expiry)
                     )
-                if auth_info.ui_language_preference and auth_info.ui_language_preference != headers.ui_lang:
-                    couchers_context.set_cookies(create_lang_cookie(auth_info.ui_language_preference))
+                if auth_info.loc_context.locale != headers.ui_lang:
+                    couchers_context.set_cookies(create_lang_cookie(auth_info.loc_context.locale))
```
aapeliv: Now you'll always set the cookie; even if the user hasn't explicitly set the value, I think. So often it would default to DEFAULT_LOCALE I think

aapeliv (changes requested):
There's a couple things this is picking up unrelated to the PR, please feel free to fix in this PR, open a new one with fixes, or create a ticket just pointing out the required changes!

app/backend/src/couchers/interceptors.py
tristanlabelle: Good catch, and I'll change the approach. But what does it mean for the user to explicitly set a language preference value? I don't understand how we distinguish None from "en" in the model and would love if we could get to eliminating None by using the cookie value or falling back to "en" at the time of user creation.

app/backend/src/couchers/context.py
tristanlabelle: I agree, it was to stage the changes. I'll file a follow-up issue.

app/backend/src/couchers/context.py
tristanlabelle: Yes, it's not any worse than before though, mostly to help stage the changes. I'll file a follow-up issue.

tristanlabelle: I made the argument mandatory for other functions but this one is a bit more complex at call sites so I filed https://github.com/Couchers-org/couchers/issues/7980 to follow up.

app/backend/src/couchers/servicers/public.py
tristanlabelle: That sounds sensible but I'd rather you push that PR since you understand the implications better.

app/backend/src/couchers/interceptors.py
tristanlabelle: I did it for most except background for which I filed https://github.com/Couchers-org/couchers/issues/7980

app/backend/src/couchers/interceptors.py
tristanlabelle: Reverted to the previous logic but I'd like us to figure out whether a user having a null locale/timezone has any meaning different from "en"/"Etc/UTC", and if not, ideally remove the nullability.

app/backend/src/couchers/context.py
tristanlabelle: Removed except for background contexts. Filed https://github.com/Couchers-org/couchers/issues/7980

app/backend/src/couchers/context.py
tristanlabelle: Tested, see summary

tristanlabelle:
@aapeliv Thanks for the review. It should be good to go now and I've tested the backend locally.

app/backend/src/couchers/interceptors.py
aapeliv: I think null makes sense; it means "no explicit preference". We could imagine parsing the language/TZ headers from the browser in case of no user-set preference.

aapeliv (approved):
lgtm

Issue #7943, opened 2026-02-21
Frontend/i18n: dayjs-formatted dates are locale-unaware


tristanlabelle:
For example, event dates always use the English US format.

dayjs support localization, but they seem to do it manually instead of using the Unicode CLDR so it would be better to leverage the built-in `Intl.DateTimeFormat` ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat))

The same is true for timezone names, and this functionality is also available through `Intl.DateTimeFormat`.

PR #7985, opened 2026-02-28
Frontend/i18n: Localize event date/times

tristanlabelle:
Implements datetime localization for the frontend (leveraging `Intl.DateTimeFormat`) and applies it to events, including on event date ranges.

There might be other event-related UI showing dates that this misses. There are a lot of other places we are formatting dates using dayjs (English-only) which I will follow up with fixing similarly once this gets approved/merged.

Note: dayjs can support localization, but we have to explicitly import the locales we care about and it's their own implementation, whereas `Intl.DateTimeFormat` should rely on standard Unicode CLDR formats, be available for all common languages, and it supports formatting date ranges. Switching over caused minor formatting differences, hence required test fixes.

Part of #7943

## Testing
Ran a local frontend against NEXT.

Event cards are the same in English after this change, modulo minor formatting changes:

image

Long cards now have a full date range:

image

Event page date ranges are the same:

image

But all of this stuff now works in other languages!
image
image
image

**Backend checklist**
**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/communities/events/EventCard.test.tsx
+++ b/app/web/features/communities/events/EventCard.test.tsx
@@ -49,7 +49,9 @@ describe("Event card", () => {
       screen.getByText(thirdEvent.offlineInformation!.address),
     ).toBeVisible();
     expect(
-      screen.getByText("Tue, Jun 29, 2021 9:00 PM - Wed, Jun 30, 2021 2:00 AM"),
+      screen.getByText(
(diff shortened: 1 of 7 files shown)


aapeliv (approved):
this looks good, i would think about how to present timezones in the UI. maybe we can have the text somewhere or just have timezone in long format versions

tristanlabelle:
@aapeliv Agreed, filed https://github.com/Couchers-org/couchers/issues/7998 to follow up.

PR #8016, opened 2026-03-01
Frontend: Remove RelativeTime plugin from dayjs configuration

tristanlabelle:
We don't use dayjs' RelativeTime plugin (having our own logic in timeAgo.ts) and don't want to use it because it's not localized unless we import a bunch of locale plugins.

From the [docs](https://day.js.org/docs/en/plugin/relative-time), this plugin only exports 4 functions. Find in files didn't find uses of any of them, so this should be safe.

## Testing
N/A - dead code

**Web frontend checklist**

## For maintainers


nabramow (approved):
LGTM!

PR #8018, opened 2026-03-01
fix: show create message modal on all message filter tabs

### LINKED ISSUE #7989: When click message it goes to chat section instead of creating message
Subject: When click message it goes to chat section isntead of creating message
Description:
I have a video for you. It's me paul.

Results:

Backend version: develop-9b4d2045
Frontend version: develop-9b4d2045
User Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.3 Safari/605.1.15
Screen resolution: 1710x983
Page: https://couchers.org/messages/chats?to=heidiiiiii
User: [@paul](https://couchers.org/user/paul) (17)

jesseallhands:
Fixes #7989

The `CreateGroupChat` component was only rendered when `filter === "all"`, but when navigating from a user profile via the Message button, the URL is `/messages/chats?to=username` which sets the filter to `"chats"`. This prevented the `CreateGroupChat` component from mounting and reading the `?to=` query param to auto-open its modal dialog.

### DIFF
--- a/app/web/features/messages/AllMessagesTab.tsx
+++ b/app/web/features/messages/AllMessagesTab.tsx
@@ -256,7 +256,7 @@ export default function AllMessagesTab() {
 
   return (
     <StyledWrapper>
-      {!showArchived && filter === "all" && <StyledCreateGroupChatButton />}
+      {!showArchived && <StyledCreateGroupChatButton />}
       <StyledFilterContainer>
         <NotificationBadge count={unseenAllCount}>
           <Chip
--- a/app/web/features/messages/groupchats/CreateGroupChat.tsx
+++ b/app/web/features/messages/groupchats/CreateGroupChat.tsx
@@ -250,10 +250,13 @@ export default function CreateGroupChat({ className }: { className?: string }) {
             )}
           </DialogActions>
           {createMessageToUsername && (
-            <Typography variant="body2" sx={{ px: 3, pb: 2 }}>
+            <Typography
+              variant="body2"
+              sx={{ px: 3, pb: 2, textAlign: "center" }}
+            >
               <Trans
                 i18nKey="messages:create_chat.hosting_request_hint"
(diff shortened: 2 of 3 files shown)


jesseallhands:
When I test this, it does now correctly show the modal, but when I click the "create" button on the modal, I am not taken to the page where I see the conversation and can begin composing the message. However, it does appear to create the conversation because when I go back to the profile and click message again, then I am taken to the page where I can see the conversation and begin composing the message. I see some changes needed here: 

Required changes:
1. I should be taken to the newly created conversation after I click the "create" button in the modal. 
2. The button text should say "Create chat" instead of just "Create"
3. There should be a "Back" button to the right of the "Create chat" button which takes the user back to the profile of the person whose "Message" button was previously pressed. The "Back" button should have the same style as the "Cancel" button that is used on the request page so the "Create chat" button stands out more.

Optional changes not related to the bug but directly related to this feature:
4. We should include a small text below the buttons saying "Sending a hosting request? Go **Back** and use the **Request** button instead."

We need to make sure that this works well on mobile web, desktop, and the app on both android and iOS, so I will test these after claude makes the updates.

jesseallhands:
@claude work on these changes

jesseallhands:
This seems to be working well now, but on the modal, the text below the buttons is wrong. It should read 

> Go **Back** and use the **Request** button instead.

but instead it reads:

> Sending a hosting request? Go Sending a hosting request? Go and use the Request button instead.

jesseallhands:
@claude fix the aforementioned issue about the text below the buttons on the modal

jesseallhands:
The text is mostly fixed now, but the word "Back" should be bold and the helper text should be center aligned like the buttons are, not left aligned.

jesseallhands:
@claude look at my last comment and make the requested changes

aapeliv (approved):

Issue #7978, opened 2026-02-26
Event start time notification uses wrong timezone 


CouchersBot:
Subject: Event start time notification uses wrong timezone 
Description:
Attend an event in a different time zone.

Results:
When attending an event the notification should show the local time zone and not the time zone set in the profile.

Backend version: develop-ea4c0f27
Frontend version: develop-1e74a765
User Agent: Mozilla/5.0 (Linux; Android 16; Pixel 8 Pro Build/BP4A.260205.001; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/145.0.7632.79 Mobile Safari/537.36
Screen resolution: 395x762
Page: https://couchers.org/dashboard
User: [@chrisk](https://couchers.org/user/chrisk) (34148)

c-kreuz:
https://github.com/user-attachments/assets/77cb4667-711d-400a-b0ac-204f10d4a637

tristanlabelle:
@c-kreuz , Good thinking. Which timezone should it be?

- Your phone's timezone 
- Your profile's timezone (autodetected, not sure if it can be different from the phone's)
- The event location's local time

I'd say it could be confusing if it's not the same as your phone clock's timezone, unless we say ", local time" or something like that.

It'll be trickier to use local times in emails since currently our footer state that all times are in the same time zone.

PR #8019, opened 2026-03-02
Frontend/i18n: Localize dates and times

### LINKED ISSUE #7943: Frontend/i18n: dayjs-formatted dates are locale-unaware

tristanlabelle:
`dayjs` only localized date/times when no format string was provided. Switch everything over to the new `localize***` functions which are correctly localized using the browser's bundled Unicode CLDR.

This lead me to find a timezone issue with events: #8064

I'll follow up separately for `monthFormatter`.

Fixes #7943

## Testing
Ran local frontend against NEXT and compared with NEXT frontend for all changed pages:

- CreateEventPage: Created an event with dates/times
- EditEventPage: Modified an event's dates/times
- HostRequestView: looked at a request in the past in future
- HostRequestUserSummarySection: Looked at a request

image

- HostRequestListItem: Looked at requests in the messages list

image

- DiscussionPage: Looked at a discussion

image

- userLabels: Looked at a person's profile

image

- LoginCard: Reviewed account settings logins

image

- Timezone: Reviewed account settings timezone

image

- InviteCodesPage: Reviewed invite code page

image

ModNoteCard: Didn't test! (don't know how to)

**Backend checklist**

**Web frontend checklist**

## For maintainers


nabramow:
Could you add some tests for different countries date formats for the date.ts functions, @tristanlabelle ? Similar to how we have tests in app/web/components/Datepicker.test.tsx for different timezones and locales.

Since we will be re-using this a lot it would be good to have good test coverage.

app/web/utils/date.ts
```diff
@@ -133,7 +122,7 @@ function timestamp2Date(timestamp: Timestamp.AsObject): Date {
 
 function getDayjsTimezone(date: Dayjs): string | undefined {
   // There is no API to get the value, but the state exists and impacts formatting.
-  return (date as any)?.$x?.timezone; // eslint-disable-line @typescript-eslint/no-explicit-any
+  return (date as any)?.$x?.$timezone; // eslint-disable-line @typescript-eslint/no-explicit-any
```
nabramow: This seems a bit sketchy. Is dayjs not exporting this type, is that the issue?

In those cases I'll usually copy the type from their package directly into my own code and reference it just to be a bit safer.

Also - why get the timezone from Dayjs rather than the browser?

tristanlabelle: Yeah, this part of dayjs' design is not very good. You can call `Dayjs.tz()` to associate a timezone with a dayjs object, and that will impact the way it gets formatted, and maybe the day/hour getters, however you cannot test whether a `Dayjs` object has an associated timezone or not, or retrieve that timezone.

We don't always want to use the user's timezone. It makes more sense to show events' times in their local timezone. The `Dayjs` design is that you'd create the `Dayjs` and then use `.tz()` to give it a timezone for formatting. My code follows that pattern by retrieving the stored `Dayjs` timezone for formatting, but it might be better not use the `Dayjs` timezone module at all and require a timezone argument to our formatting functions. I'll play with that idea and iterate.

(thanks for the typing tip!)

nabramow: Yeah I was wondering if passing the timezone might be better, but obviously that's going to mean drilling it through everywhere so fine with this for now.

That makes sense with the events and local time I think we have a bug with that even: #7978

tristanlabelle: I actually went with requiring the timezone to be passed on each call, with UTC_TIMEZONE and BROWSER_TIMEZONE constants, and removed dayjs' timezone module altogether. I think this will avoid mistakes since it forces call sites to be explicit about how they need the date to be formatted.

tristanlabelle: Actually we still need the timezone module for the datetime/picker tests. I'll look into this separately.

tristanlabelle:
@nabramow Addressed comments, ready for re-review!

aapeliv (commented):
I feel like it would be better to standardize the timezone we use in some central place. I feel like hardcoding `BROWSER_TIMEZONE` everywhere is a bit brittle.

I feel like we should always display dates/times either with the user's timezone; or in the timezone of the object that that time relates to.

I would generally be of the opinion that it would be better to define some clear library functions that define a few different options instead of defining at the callsite how we wish for it to show?

tristanlabelle:
@aapeliv Could you draft what you have in mind? I landed on this API shape to address some of the concerns you raise:

> I feel like it would be better to standardize the timezone we use in some central place. I feel like hardcoding BROWSER_TIMEZONE everywhere is a bit brittle.
> I feel like we should always display dates/times either with the user's timezone; or in the timezone of the object that that time relates to.

"always either x or y" is still a per-callsite decision, and we also have a case where we show a users' local time. If I make the browser timezone the default, it's harder to spot errors. For example, currently events use the browser timezone and this is incorrect, and that highlights that we have a bug where the backend doesn't know the correct event timezone so we can't use it. I think of it as being explicit rather than hardcoding.

> I would generally be of the opinion that it would be better to define some clear library functions that define a few different options instead of defining at the callsite how we wish for it to show?

The current shape is the clearest I could get. Whether you need the day of the week, to abbreviate months, or show seconds is a per-callsite decision, and we have examples of all of those options. I'm all ears if you can think of more abstraction.

app/web/features/messages/requests/HostRequestView.tsx
```diff
@@ -146,9 +146,7 @@ export default function HostRequestView({
   const currentUserId = useAuthContext().authState.userId;
   const isHost = host?.userId === currentUserId;
   const otherUser = isHost ? surfer : host;
-  const isRequestPast = dayjs(hostRequest?.toDate).isBefore(
-    dayjs().format("L"),
-  );
+  const isRequestPast = dayjs(hostRequest?.toDate).isBefore(dayjs(), "day");
```
nabramow, line 149: Since we're still using dayjs here, I'm sure curious why the change?

app/web/features/profile/view/userLabels.tsx
```diff
@@ -270,9 +270,11 @@ export const RemainingAboutLabels = ({ user }: Props) => {
       />
       <LabelAndText
         label={t("profile:heading.local_time")}
-        text={dayjs()
-          .tz(user.timezone || "Etc/UTC")
-          .format("LT")}
+        text={localizeDateTime(dayjs(), {
+          timezone: user.timezone || "Etc/UTC",
```
nabramow, line 274: What's the use case for ETC? UTC makes sense to be for a default but the "Etc" is throwing me.

app/web/utils/date.test.ts
```diff
@@ -1,9 +1,118 @@
-import { isSameOrFutureDate } from "utils/date";
+import { isSameOrFutureDate, localizeDateTime, UTC_TIMEZONE } from "utils/date";
 import dayjs from "utils/dayjs";
 
-const FUTURE = dayjs("02-15-2025");
-const PAST = dayjs("10-05-1991");
-const TODAY = dayjs("03-25-2021");
+const FUTURE = dayjs("2025-02-15");
+const PAST = dayjs("1991-10-05");
+const TODAY = dayjs("2021-03-25");
```
nabramow, line 6: Did we have these in the wrong format?

app/web/utils/date.test.ts
```diff
@@ -1,9 +1,118 @@
...
+    expect(
+      localizeDateTime(dayjs("2000-01-01"), {
+        timezone: UTC_TIMEZONE,
+        locale: "es",
+      }).toLowerCase(),
+    ).toContain(
+      "ene", // enero = january
+    );
+  });
+});
```
nabramow, line 115: Could we add a couple more locale ones, i.e. German and USA for example? Just to cover some bases.

app/web/utils/date.ts
```diff
@@ -10,132 +10,92 @@ const monthFormatter = (locale: string) =>
...
+  includeDate?: boolean;
+  /// If including the date, whether to include the day of week (defaults to false).
+  includeDayOfWeek?: boolean;
+  /// Whether to include the time (defaults to true).
+  includeTime?: boolean;
+  /// If including the time, whether to include seconds (defaults to false).
+  includeSeconds?: boolean;
+  /// Whether to abbreviate days of the week and month names (defaults to false).
+  abbreviate?: boolean;
+}
```
nabramow, line 39: Thanks for the comments this is great!

nabramow (approved):
Just did a round of testing and everything is working great for me! I poked around the events, discussions and host requests and changed the locale a bunch.

I just had a couple tiny questions but I will tentatively approve!

app/web/features/messages/requests/HostRequestView.tsx
tristanlabelle: The new version expresses better the day-granularity comparison. Formatting as a date-only and having `isBefore` parse it back is roundabout and relies on both assuming the same mm/dd ordering.

tristanlabelle: The value of `dayjs` is the immutability of its objects and the helper functions it exposes.

app/web/features/profile/view/userLabels.tsx
tristanlabelle: Timezone formats are `A/B`, generally `Region/City` e.g. `Europe/Amsterdam`. `Etc` is the region for timezones that are not location-specific. Actually `UTC` might exist but if it does it is an alias for `Etc/UTC`. See https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

app/web/utils/date.test.ts
tristanlabelle: I think there's only one unambiguous format: the ISO standard `YYYY-MM-DD`. `##-##-####` is ambiguous between `DD-MM-YYYY` (e.g. Indian format) and `MM-DD-YYYY`, especially when the US MM/DD format uses slashes. I don't know what logic dayjs uses to guess what date `10-05-1991` represents, but it might depend on the machine locale so better not rely on it.

app/web/utils/date.test.ts
tristanlabelle: Sure, will do as follow up.

tristanlabelle:
@aapeliv I'm merging this so I can build on top but I'm still all ears for your alternate design.

app/web/utils/date.test.ts
tristanlabelle: Follow-up: https://github.com/Couchers-org/couchers/pull/8147

PR #8023, opened 2026-03-02
Add mobile to auto-format.yml & don't autoformat dependabot

nabramow:
- Prevent the auto-format from running on dependabot PRs here since it's not really necessary.
- Add mobile folder to the auto-format
- Add expo-doctor check to CI when packages change, as expo is very opinionated about what updates when and together with which other packages


aapeliv (approved):
lgtm

PR #8025, opened 2026-03-02
Combine lint and prettier into yarn format

nabramow:
Long overdue, just combining the linting and prettier into the one `yarn format` command so we don't have to run them separately.


app/web/package.json
```diff
@@ -10,7 +10,7 @@
     "injectsourcemaps": "sentry-cli sourcemaps inject .next/",
     "dev": "next dev",
     "start": "yarn dev",
-    "format": "prettier --write '**/*.{ts,tsx,js,jsx,json,css}'",
+    "format": "prettier --write '**/*.{ts,tsx,js,jsx,json,css}' && eslint . --ext .ts,.tsx,.js,.jsx,.json --fix",
```
tristanlabelle, line 13: FYI: CI will now lint https://github.com/Couchers-org/couchers/blob/948f1a3f7980a577b849e7a5d60571cb948949b4/.github/workflows/auto-format.yml#L119 . But I think this is fine and even desirable.

tristanlabelle (approved):

app/web/package.json
nabramow: @tristanlabelle yes that was the goal! should be doing it anyway!

PR #8038, opened 2026-03-03
Frontend/i18n: Apply best practices to invite codes strings

tristanlabelle:
Translators were asking about pluralization of "completed signups", which lead to fixing a few issues with the invite codes strings:

- Replaced string concatenations with placeholders
- Consolidated duplicate strings (one copy of which was unused but still getting translated)
- Used more descriptive string keys (single words are generally not translatable without context)

This will cause some amount of relocalization, but since the original strings included little context, their translation quality might have been poor.

This will conflict with #8019, so please help review that PR!

## Testing
Ran local frontend against NEXT, created and disabled invite codes.

image

**Web frontend checklist**

## For maintainers


aapeliv (approved):
lgtm

PR #7822, opened 2026-02-03
Add validation to display name input

AndyMcGN:
*Describe briefly what this PR is doing and why.*

Completes this [issue](https://github.com/Couchers-org/couchers/issues/6802) about the lack of validation of the display name during sign up.

## Testing
*Explain how you tested this PR and give clear steps so the reviewer can replicate.*

**Backend checklist**

- [x] Added tests for any new code or added a regression test if fixing a bug
- [x] Run the backend locally and it works
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history

**Web frontend checklist**

- [x] There are no console warnings when running the app
- [x] Added tests where relevant
- [x] Clicked around my changes running locally and it works
- [x] Checked Desktop, Mobile and Tablet screen sizes

## For maintainers

- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

Issue #6802, opened 2025-09-09
[Sign UP]Add validation message to display name field


vas2525:
**Actual result:**
No validation in the Display Name field
There is no validation in the Display Name field during Sign UP and on the "Edit Profile" page

<img width="1561" height="959" alt="Image" src="https://github.com/user-attachments/assets/1dcc31e5-1371-4477-8402-8b531f832b4c" />

<img width="988" height="1037" alt="Image" src="https://github.com/user-attachments/assets/a9af5c39-0bcd-46a9-8c77-201585100829" />

**Expected result:**
The Display Name field should have validation during Sign UP

We should do the following
Verify Name Min Lenght(2chr) 
Verify Name MAX Lenght(100chr) 
Verify disallowed characters (!@#$)
Allow full name with hyphen or apostrophe (eg. Anne-Marie O’Connor)
Allow full name with non-Latin characters (eg. “Иван Иванов” (Cyrillic) or “张伟” (Chinese))
Allow Upper and lower cases

PR #8065, opened 2026-03-07
Add backend validation of name during signup

AndyMcGN:

Completes this [issue](https://github.com/Couchers-org/couchers/issues/6802) about the lack of validation of the display name during sign up. 
Corresponds to [this pr](https://github.com/Couchers-org/couchers/pull/7822) which implements frontend validation

## Testing
gRPC requests to signup with valid and invalid usernames

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -1,4 +1,5 @@
 import http.cookies
+import logging
 import re
 import typing
 from collections.abc import Mapping, Sequence
@@ -27,7 +28,7 @@ from couchers.crypto import (
     encrypt_page_token,
 )
 from couchers.proto.internal import internal_pb2
-import logging
+
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -27,6 +27,8 @@ from couchers.crypto import (
     encrypt_page_token,
 )
 from couchers.proto.internal import internal_pb2
+import logging
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from couchers.models import Geom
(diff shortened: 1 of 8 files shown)


app/backend/src/couchers/utils.py
```diff
@@ -59,9 +62,25 @@ def is_valid_username(field: str) -> bool:
...
+    * 2-100 characters
     """
-    return re.match(r"\S+", field) is not None
+    logger.info("validating name: %r", field)
+
+    trimmed = field.strip()
+    if len(trimmed) > 100 or len(trimmed) < 2:
+        return False
+
+    # python's re has no \p{L} -> use character check so we allow any Unicode letter.
```
aapeliv: let's pull in the `regex` library if we need `\p{L}`; this implementation feels problematic (e.g. `isalpha()` does NOT check for just alphanumeric stuff)

aapeliv (changes requested):

app/backend/src/couchers/utils.py
AndyMcGN: ok fair, changed to use the library

app/backend/src/couchers/utils.py
```diff
@@ -28,6 +29,8 @@
 )
 from couchers.proto.internal import internal_pb2
 
+VALID_NAME_PATTERN = regex.compile(r"^[\p{L}\s'-]+$")
```
aapeliv, line 32: please put this in the constants with the other patterns.

this pattern looks fine to me after discussion with gemini; but it does comes up with a couple cases that I think we could support if we wanted to (numbers and some more punctuation). feel free to update or leave as-is; but important it's replicated same on the frontend.

Image

app/backend/src/couchers/utils.py
```diff
@@ -59,9 +62,18 @@ def is_valid_username(field: str) -> bool:
...
-    Checks if it has at least one non-whitespace character
+    Checks that the name satisfies the same rules as the web frontend:
+
+    * only letters (any Unicode letter), whitespace, apostrophes, and hyphens
+      (frontend: ``^[\\p{L}\\s'-]+$``)
+    * 2-100 characters
     """
-    return re.match(r"\S+", field) is not None
+
+    trimmed = field.strip()
```
aapeliv, line 72: Is it trimmed in the frontend for length matching? You're not allowed to start/end with whitespace, so this doesn't do anything, right?

app/web/utils/validation.ts
```diff
@@ -1,5 +1,5 @@
 // taken from backend
-export const nameValidationPattern = /^[\p{L}\s'-]+$/u;
+export const nameValidationPattern = /^[\p{L}'-]+(\s+[\p{L}'-]+)*$/u;
```
nabramow, line 2: I think we don't need this anymore now @aapeliv ? Since it looks like we're fully handling on the backend.

aapeliv: nicer to have it mirrored on frontend so it's interactive

aapeliv (approved):

PR #8081, opened 2026-03-10
Fix a race condition when handling notifications

WouldYouKindly:
Fixes https://couchers.sentry.io/issues/7247061930. Two workers can sometimes start working on the same notification, both read it and assume it's not sent, and then try to insert delivery attempt record to the DB, resulting in unique constraint violiation.


aapeliv (approved):
you're a wizard

PR #8092, opened 2026-03-11
Frontend/i18n: Use backend's notification descriptions

tristanlabelle:
#7974 localized the notification descriptions coming from the backend. The frontend ignored these strings and did its own localization, probably because the backend did not support localization at that time.

Also fixes backend not accounting for localization context when returning the notification descriptions.

This way we can use the same descriptions in the settings page and in the email unsubscribe footers.

## Testing
Ran local backend + frontend. Shows backend settings strings (with extra full stop, but has been removed in a different PR)

image

**Backend checklist**
- Ran locally

**Web frontend checklist**

## For maintainers


aapeliv (changes requested):
The backend descriptions are not yet translated (they are hardcoded to English: https://github.com/Couchers-org/couchers/blob/9a42f6760e77551bc843db1efbc06490a007448c/app/backend/src/couchers/notifications/settings.py#L330); let's implement that in the backend before we merge this change so we don't regress this part of translations?

tristanlabelle:
@aapeliv Thanks for catching! I added the backend-side localization to this PR. All yours to review.

aapeliv (approved):
lgtm

PR #8107, opened 2026-03-15
Add postcard sending

aapeliv:

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+    qr_extend = 5
+
+    # Generate QR code
+    qr = qrcode.QRCode(box_size=10, border=0)
+    qr.add_data(urls.postal_verification_link(code=verification_code))
+    qr.make(fit=True)
+    qr_img: Image.Image = qr.make_image(fill_color="black", back_color="white").get_image().convert("RGBA")
+
+    # Size and paste the QR code into the extended box area
+    qr_size = min(qr_r - qr_l, qr_b - qr_t) + 2 * qr_extend
```
WouldYouKindly: This could be a constant

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+
+    Overlays a QR code and verification code onto the postcard-back.png template.
+    """
+    # Load template
+    template_bytes = get_postcard_back_template()
+    img = Image.open(io.BytesIO(template_bytes)).convert("RGBA")
+    draw = ImageDraw.Draw(img)
+
+    # QR code box position/size from template, extended to cover border
+    qr_l, qr_t, qr_r, qr_b = 227, 419, 539, 731
```
WouldYouKindly: I'd go for qr_left, qr_top etc. Was a bit hard reading the code below.

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+
+    # Size and paste the QR code into the extended box area
+    qr_size = min(qr_r - qr_l, qr_b - qr_t) + 2 * qr_extend
+    qr_img = qr_img.resize((qr_size, qr_size), Image.Resampling.NEAREST)
+    img.paste(qr_img, (qr_l - qr_extend, qr_t - qr_extend))
+
+    # Verification code box position/size from template
+    code_box_x, code_box_y, code_box_w, code_box_h = 251, 761, 264, 80
+    code_cx = code_box_x + code_box_w // 2
+    code_cy = code_box_y + code_box_h // 2
```
WouldYouKindly: This could be pre-calculated to avoid unnecessary work

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+
+    font = ImageFont.truetype(io.BytesIO(get_postcard_font()), 58)
```
WouldYouKindly: same

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    response = requests.post(
+        f"{API_BASE}/place_order",
+        data={
+            "api_key": config["MYPOSTCARD_API_KEY"],
+            "auth_token": auth_token,
+            "product_code": config["MYPOSTCARD_PRODUCT_CODE"],
+            "image_type": "png",
+            "job_data": json.dumps(job_data),
```
WouldYouKindly, line 114: Are you sure this should be a string, not an object?

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+        f"{API_BASE}/request_orders",
+        data={
+            **_credentials(),
+            "date_from": date_from.strftime("%Y-%m-%d"),
+            "date_to": date_to.strftime("%Y-%m-%d"),
+        },
+        timeout=30,
+    )
+    response.raise_for_status()
+    return response.json()
```
WouldYouKindly: Could we please make a dataclass out of it to not work with dicts?

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def download_pdf(job_id: int) -> bytes:
```
WouldYouKindly, line 192: What is this PDF? is it the postcard with the qr code?

app/backend/src/couchers/postal/my_postcard.py
aapeliv: Yep… the API is wonky!

app/backend/src/couchers/postal/my_postcard.py
aapeliv: Yes, a pdf of the postcard that was sent

WouldYouKindly: Then I don't understand why we have an endpoint to get it? Or is it only for admins?

WouldYouKindly: The users should only be able to see the QR code when the postcard arrives, right?

aapeliv: It’s only for admins

app/backend/src/couchers/postal/my_postcard.py
aapeliv: refactored to return a list of job IDs to get around this! (would have made it a dataclass but it would just have a single member)

app/backend/src/couchers/postal/my_postcard.py
aapeliv: claude tells me (after consulting the source code for Pillow) that it's not a thread-safe thing to cache the `ImageFont` object

app/backend/resources/postcard-back-left.png
tristanlabelle, line 1: "User" should probably be lower case?

This image is going to be hard to edit (and localize), it'd be worth double-checking the copy with whoever usually does that.

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Overlays a QR code and verification code onto the postcard-back.png template.
+    """
+    # Load template
+    template_bytes = get_postcard_back_template()
+    img = Image.open(io.BytesIO(template_bytes)).convert("RGBA")
+    draw = ImageDraw.Draw(img)
+
+    # QR code position: exact coordinates in image are (227, 419, 539, 731), extended by 5px in each direction
+    qr_left, qr_top, qr_size = 222, 414, 322
```
tristanlabelle: Keeping all those constants in a (json) file next to the images would make it less error-prone to edit the image in the future.

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,201 @@
...
+from couchers import urls
+from couchers.config import config
+from couchers.resources import get_postcard_back_template, get_postcard_font, get_postcard_front_image
+
+logger = logging.getLogger(__name__)
+
+API_BASE = "https://www.mypostcard.com/api/v1"
+
+
+def _generate_back_left_side(verification_code: str) -> bytes:
```
tristanlabelle: A `_png` suffix would clarify the `bytes`.

app/backend/resources/postcard-back-left.png
tristanlabelle, line 1: `postcard-back-left.png` would be clearer. I was wondering how the address part worked.

app/backend/src/tests/test_postal_verification.py
```diff
@@ -686,3 +689,22 @@ def test_has_postal_verification_helper(db, monkeypatch):
     with session_scope() as session:
         db_user = session.execute(select(User).where(User.id == user.id)).scalar_one()
         assert has_postal_verification(session, db_user)
+
+
+def test_generate_postcard_images():
+    """
+    Generates sample postcard front and back images for visual inspection.
+
+    Output is written to test_artifacts/ (gitignored) and picked up by CI.
```
tristanlabelle, line 698: If we had CI dictate the artifacts location as a test argument when running tests, we wouldn't need to depend on two unrelated parts of the code agreeing on that location.

app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def send_postcard(
+    recipient_name: str,
+    address_line_1: str,
+    address_line_2: str | None,
+    city: str,
+    state: str | None,
+    postal_code: str | None,
+    country: str,
```
tristanlabelle, line 135: Suggestion: `MyPostcardRecipient` dataclass with `to_json` function. That could be shared with `_place_order` and the tests that mock it. It raises the abstraction level from `dict[str, str]`. It could also be constructed from a `postal_verification_pb2.PostalAddress` + name so we don't need to fall down to individual fields.

app/backend/src/tests/conftest.py
```diff
@@ -186,6 +186,11 @@ def testconfig():
     )
 
     config["ENABLE_POSTAL_VERIFICATION"] = False
+    config["MYPOSTCARD_API_KEY"] = "test-api-key"
+    config["MYPOSTCARD_USERNAME"] = "test-username"
+    config["MYPOSTCARD_PASSWORD"] = "test-password"
+    config["MYPOSTCARD_PRODUCT_CODE"] = "J9GCU"
+    config["MYPOSTCARD_CAMPAIGN_ID"] = "295"
```
tristanlabelle, line 194: It's unclear if those codes/ids have a meaning. A comment would help

app/backend/resources/hack-bold.ttf
tristanlabelle, line 1: Please include a description on PRs!

app/backend/src/couchers/urls.py
```diff
@@ -148,4 +148,4 @@ def invite_code_link(*, code: str) -> str:
 
 
 def postal_verification_link(*, code: str) -> str:
-    return f"{config['BASE_URL']}/verify-postal?code={code}"
+    return f"{config['BASE_URL']}/verify-postal?c={code}"
```
tristanlabelle, line 151: I couldn't correlate this change with other parts of the code. Is there a handler that went from handling `code` to `c`?

tristanlabelle (dismissed):
I filed https://github.com/Couchers-org/couchers/issues/8213 to track an i18n follow up. I recommend adding a MyPostcardRecipient recipient and extracting postcard coords to a json file, otherwise looking forward to this!

app/backend/src/couchers/postal/my_postcard.py
WouldYouKindly: It could just be a module-level constant though? Anyway, not a major issue

app/backend/src/couchers/postal/my_postcard.py
WouldYouKindly: Or `type png = bytes`

app/backend/src/couchers/postal/my_postcard.py
WouldYouKindly: +1, but I don't think we even need to_json. `json.dumps(dataclasses.as_dict(recepient)` will do

app/backend/src/tests/conftest.py
WouldYouKindly: These are just some IDs from mypostcard though?

app/backend/src/couchers/postal/my_postcard.py
tristanlabelle: Not quite because the mypostcard property names are camelCase

app/backend/src/couchers/urls.py
aapeliv: I don't think there's any frontend handler currently.

app/backend/resources/postcard-back-left.png
aapeliv: I have a feeling we're going to end up redoing this anyway. So I'm going to leave it for the time being. Probably makes more sense to programmatically generate everything that gets localized?

app/backend/resources/hack-bold.ttf
aapeliv: sorry, my bad

app/backend/src/tests/conftest.py
aapeliv: yes just some IDs. their API is a tad complicated

app/backend/src/tests/test_postal_verification.py
aapeliv: not entirely sure how that would work. i do not see any real case where we would want to modify the default?

i mean the fact that CI picks it up is not really that relevant to this part of code.

WouldYouKindly (approved):

app/backend/resources/postcard-back-left.png
tristanlabelle: Indeed, sounds good then

PR #8112, opened 2026-03-16
I18n: Pluralize remaining unpluralizable strings

tristanlabelle:
These were the last few placeholder numbers that didn't support pluralization. Two reasons to pluralize them:

- Pluralization isn't just for 1 vs many, in some languages a number ending with 2 like 102 will have a different plural form than a number ending in 6 like 106, similar to English's 102nd vs 106th.
- All strings now provide the right example for adding future strings.

I changed the keys because the placeholder's name changed, so we can't reuse existing translations.

I also changed the "65k+" placeholder in favor of formatting the `65000` number because "65k+" is not localized. I assume we're not too concerned about that code path, which implies a backend bug.

## Testing
image

image

image

**Backend checklist**

**Web frontend checklist**

## For maintainers


aapeliv (approved):
lgtm, thanks for fixing!

PR #8160, opened 2026-03-23
Fix desktop menu design

### LINKED ISSUE #8129: Fix desktop main menu design
When we creates the mobile-only full-screen LoggedInMenu, the desktop version got a bit messed up. We need to:

* left align the text for the desktop menu
* make hover highlight the whole row not just the text
* mobile view menu (isMobile and isNativeEmbed conditionals) should remain the same

nabramow:
Closes #8129 

Desktop menu design got a bit messed up when we changed the mobile menu design. This is fixing it by left-aligning and fixing the hover.

## Testing

- View menu in desktop and make sure it looks okay. Main top left menu.
- View in mobile responsive mode and make sure it still looks okay

Screenshot 2026-03-23 at 15 36 28

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/Navigation/LoggedInMenu.tsx
+++ b/app/web/components/Navigation/LoggedInMenu.tsx
@@ -393,10 +393,10 @@ export default function LoggedInMenu({
             flex: 1,
             display: "flex",
             flexDirection: "column",
-            alignItems: "center",
-            justifyContent: "center",
-            gap: { xs: 1.5, md: 3 },
-            textAlign: "center",
+            alignItems: { xs: "center", md: "stretch" },
+            justifyContent: { xs: "center", md: "flex-start" },
+            gap: { xs: 1.5, md: 0 },
+            textAlign: { xs: "center", md: "left" },
           }}
         >
           {items.map((item) => (


app/web/components/Navigation/LoggedInMenu.tsx
```diff
@@ -393,10 +393,10 @@ export default function LoggedInMenu({
             flex: 1,
             display: "flex",
             flexDirection: "column",
-            alignItems: "center",
-            justifyContent: "center",
-            gap: { xs: 1.5, md: 3 },
-            textAlign: "center",
+            alignItems: { xs: "center", md: "stretch" },
```
tristanlabelle, line 396: For my learning, would it be possible to use more explicit names for breakpoints or have constants that expand to "xs" and "md" so we can say something like: `desktop: "stretch"`?

tristanlabelle (approved):
Thanks! Looks so much better

app/web/components/Navigation/LoggedInMenu.tsx
nabramow: This is what the package wants. It's technically by screen size as well, so if we shrunk mobile real small would be the same, so that would technically be inaccurate. We could switch it to `theme breakpoints.down("md")` but that's more verbose and not so much clearer IMO.

Issue #5227, opened 2024-11-25
Move `isMobile` etc into centralized utils


aapeliv:
We have this occuring a bunch:

```js
const isMobile = useMediaQuery(theme.breakpoints.down("md"));
```

In fact it seems sometimes the breakpoint is `md` and sometimes it's `sm`.

We should move this into its own utility function/file.

**Nicole note:**

- When it's `.down` is should be "md" at it means below medium is Mobile.. When it's `.up` them it would mean above "sm" is Desktop. But yeah we should just choose one to represent mobile.

PR #8165, opened 2026-03-23
add screen breakpoint util

lennartvanlaake:

Implements part of https://github.com/Couchers-org/couchers/issues/5227 - creates the utils for centralized management of breakpoints. As requested, does not yet implement the util anywhere. For implementation of this util, see the branch web/refactor/is-mobile-util-implementation.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


lennartvanlaake:
@nabramow  as requested, a util-only PR version of the ìsMobile`thing. I do think it would be smart to actually do a proper sweep to make sure this util is used everywhere soon, otherwise we get the classic issue of adding yet another way of doing mobile breakpoints to the list instead of creating consistency. I kept renamed the 'old' version of the branch here: [web/refactor/is-mobile-util-implementation](https://github.com/Couchers-org/couchers/tree/web/refactor/is-mobile-util-implementation).

nabramow (approved):
Looks good to me! I agree we should def do a sweep. I just want to make sure everything is tested on mobile web and mobile app to be sure we don't break anything as there's a lot of conditionals that the mobile app depends on.

Do you have the staging app set up on your phone? I can add you to the testers list if you need!

In terms of reviewing it would also be easier to do chunks. For example for me as a reviewer "test the whole app" is super overwhelming, but "test the dashboard and navigation menu" is less.

If you wanna do more manageable chunks I can help review and test the PRs as you go.

lennartvanlaake:
> Looks good to me! I agree we should def do a sweep. I just want to make sure everything is tested on mobile web and mobile app to be sure we don't break anything as there's a lot of conditionals that the mobile app depends on.
> 
> Do you have the staging app set up on your phone? I can add you to the testers list if you need!
> 
> In terms of reviewing it would also be easier to do chunks. For example for me as a reviewer "test the whole app" is super overwhelming, but "test the dashboard and navigation menu" is less.
> 
> If you wanna do more manageable chunks I can help review and test the PRs as you go.


Thanks! For me the bottleneck is not necessarily the testing scope but the decision if we want to change stuff if the current breakpoints are inconsistent. What we could do is that I make a bunch of PR's for each functional part (i.e.: navigation, events, etc) and clearly mark anything that requires a functional decision. Then perhaps you can have a look at it and see what we want to do there.

PR #8172, opened 2026-03-25
Frontend/i18n: Fix "countries" singular.

tristanlabelle:
As reported by a translator.


c-kreuz:
If the displayed count reads 1+, then the correct word is “countries” — plural.
1+ country → sounds wrong
1+ countries → correct, because the “+” implies multiple possibilities

tristanlabelle:
@c-kreuz Fine, I don't mind either way. This is actually a hard-coded constant right now.

PR #8162, opened 2026-03-23
Mobile fixes v1.1.16 - signup confirmation and staging asset links

nabramow:
Closes #8041 

- Add confirm-email route to the allowed non logged out routes for mobile to fix signup confirmation
- Add asset links for staging app so staging will also redirect to app if installed (so we can replicate these kinda things on staging app)
- Add user agent string for bug reports

## For maintainers

- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

PR #8179, opened 2026-03-26
Fix Android deep link handling for email confirmation

aapeliv:
+ password reset + email change

The login screen's WebView always loaded /login, ignoring the incoming deep link URL. Now captures the initial URL via Linking.getInitialURL() and listens for runtime deep links, forwarding auth-related paths (/signup, /confirm-email, /complete-password-reset) with their query parameters to the WebView.

Human: This should fix the signup issues we've been having.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/mobile/app/login.tsx
+++ b/app/mobile/app/login.tsx
@@ -1,6 +1,7 @@
 import { useFocusEffect } from "@react-navigation/native";
+import * as Linking from "expo-linking";
 import { Href, useRouter } from "expo-router";
-import { useCallback, useState } from "react";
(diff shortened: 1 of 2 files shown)


nabramow:
I have a related PR to this already in #8162 I'm finishing up FYI. You can see I assigned myself to the ticket on the board.

Please be careful with mobile tickets as there is a process, we need to increase the mobile version number in the right places, do the release notes, test it, etc. or the build will not be downloadable.

PR #8184, opened 2026-03-28
CI: Single auto-format job

### LINKED ISSUE #7880: PRs: GHA auto-formatting/codegen jobs can race
If we need to format/codegen two things, whichever jobs completes and pushes a commit first could cancel the other ones. Ideally we'd have a single job for them all, but GHA code reuses needs careful handling when checking out the head_ref.

tristanlabelle:
Combines CI auto-formatting steps into one job, since pushing commits will cancel the currently running workflow and currently we're racing between all formatting jobs. Preserves the step skipping logic.

The database migration step cannot be combined because of its service dependency.

Fixes #7880

## Testing
Test PR with backend + mobile formatting changes: #8185
Workflow run: https://github.com/Couchers-org/couchers/actions/runs/23673689020/job/68972483178?pr=8185
Result:

image

## For maintainers

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -14,7 +14,9 @@ permissions:
 
 jobs:
   detect-changes:
+    name: Detect changes
     runs-on: ubuntu-latest
+    if: github.actor != 'dependabot[bot]'
     outputs:
       backend: ${{ steps.filter.outputs.backend }}
       frontend: ${{ steps.filter.outputs.frontend }}
@@ -22,10 +24,9 @@ jobs:
       models: ${{ steps.filter.outputs.models }}
       migrations: ${{ steps.filter.outputs.migrations }}
       templates: ${{ steps.filter.outputs.templates }}
+      any: ${{ steps.filter.outputs.changes != '[]' }}
     steps:
-      - uses: actions/checkout@v4
-
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: |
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-format.yml
```diff
@@ -69,109 +74,89 @@ jobs:
...
+        if: needs.detect-changes.outputs.templates == 'true'
+        run: |
+          git add "app/backend/templates/v2/generated_html/*.html"
+          git diff --staged --quiet || git commit -m "Regenerate HTML templates"
+
+      # Push any commits at once, so we only invalidate the workflow once
+      - name: Push changes
+        run: git push
 
   generate-migration:
```
WouldYouKindly, line 161: This can generate a change and have the same effect as formatting. However, I'd argue that we shouldn't be commiting autogenerated migrations without a proper review!! @aapeliv :)

aapeliv: What do you mean by without a proper review? This is just a helper so you don't have to start by hand; but it works all the same as doing it locally. There is a bot that will check the migration but indeed the dev should update the migration after it has been autogenerated (whether locally or in CI).

This is just here to make it a bit easier to autogenerate, the steps are annoying to do (and require a clean DB) to generate locally.

aapeliv (approved):
lgtm

WouldYouKindly:
@aapeliv maybe you're right... I was just thinking that if you run the auto-generation yourself, you at least look at the resulting migration and sometimes spot an issue. When it's generated in an action, you yourself most likely won't look at it. Maybe the reviewer will, but 2 sets of eyes are better.

aapeliv:
> @aapeliv maybe you're right... I was just thinking that if you run the auto-generation yourself, you at least look at the resulting migration and sometimes spot an issue. When it's generated in an action, you yourself most likely won't look at it. Maybe the reviewer will, but 2 sets of eyes are better.

Yeah I agree. There are two main safeguards for this:

1. We do have a CI job to check migrations are correct
2. There is not that LLM-powered migration reviewer that will try to check if the migration can safely be applied when data exists in the DB

But yeah migrations are one of the most common sources of deployment issues.

.github/workflows/auto-format.yml
tristanlabelle: I'm also not keen on using CI for an intermediate development step which the dev is then expected to pull and follow-up on locally, but this is unrelated to this PR's changes so I'll proceed with merging.

PR #8188, opened 2026-03-28
Allow importing your data from couchsurfing.com

WouldYouKindly:
## How it works

A user presses the import button, upload the archive from couchsufring. We extract the JSON with the data right in the browser (to avoid having to deal with zip bombs etc on the backend), send it to `ImportFromCouchsurfing` endpoint, which validates the JSON and updates the database.

Screenshot 2026-03-28 at 10 15 00

image

### DIFF
--- a/app/backend/src/couchers/couchsurfingcom_import.py
+++ b/app/backend/src/couchers/couchsurfingcom_import.py
@@ -1,0 +1,523 @@
+"""
+Helper for importing Couchsurfing™ data export into a Couchers user profile.
+
+This module provides functionality to update an existing Couchers user profile
+with data exported from Couchsurfing.com.
+
+Couchsurfing™ is a trademark of Couchsurfing International, Inc.
+"""
+
+import json
+import logging
+from dataclasses import dataclass
+from dataclasses import field as dataclass_field
+from typing import Any
+
+import cattrs
+from attrs import define, field
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+from sqlalchemy.orm.attributes import InstrumentedAttribute
+
+from couchers.i18n import LocalizationContext
+from couchers.models import (
+    CouchsurfingComImportAttempt,
+    LanguageAbility,
+    LanguageFluency,
+    SleepingArrangement,
+    SmokingLocation,
+    User,
+)
+from couchers.resources import language_is_allowed
+from couchers.utils import now
+
+# Maximum JSON size in bytes (20MB)
+MAX_JSON_SIZE = 20 * 1024 * 1024
+
+# Couchsurfing.com section titles (English-only, matching the CS website)
+# These are the exact titles users see when filling in their CS profile
+CS_SECTION_MY_FAVORITE_MUSIC_MOVIES_BOOKS = "My Favorite Music, Movies & Books"
+CS_SECTION_TEACH_LEARN_SHARE = "Teach, Learn, Share"
+CS_SECTION_ONE_AMAZING_THING_IVE_DONE = "One Amazing Thing I've Done"
+CS_SECTION_WHY_IM_ON_COUCHERS = "Why I'm on Couchers"
+CS_SECTION_WHAT_I_CAN_SHARE_WITH_HOSTS = "What I Can Share With Hosts"
+CS_SECTION_DESCRIPTION_OF_SLEEPING_ARRANGEMENT = "Description of Sleeping Arrangement"
+CS_SECTION_ROOMMATE_SITUATION = "Roommate Situation"
+CS_SECTION_PUBLIC_TRANSPORTATION_ACCESS = "Public Transportation Access"
+CS_SECTION_WHAT_I_CAN_SHARE_WITH_GUESTS = "What I Can Share With Guests"
+CS_SECTION_ADDITIONAL_INFORMATION = "Additional Information"
+
+
+def _serialize_value(value: Any) -> Any:
+    """Serialize a value for JSON storage in the import log."""
+    if value is None:
+        return None
+    if isinstance(value, (str, int, float, bool)):
+        return value
+    if isinstance(value, (SleepingArrangement, SmokingLocation, LanguageFluency)):
+        return value.name
+    return str(value)
+
+
+logger = logging.getLogger(__name__)
+
+
+# --- Couchsurfing.com Data Schema (attrs classes for cattrs) ---
+
+
+@define
+class CouchsurfingComLanguages:
+    """Languages from Couchsurfing.com profile."""
+
+    fluent: list[str | None] = field(factory=list)
+    learning: list[str | None] = field(factory=list)
+
+
+@define
+class CouchsurfingComProfile:
+    """Profile data from Couchsurfing.com export."""
+
+    first_name: str | None = None
+    last_name: str | None = None
+    gender: str | None = None
+    occupation: str | None = None
+    about_me: str | None = None
+    interests: str | None = None
+    media: str | None = None
+    teach: str | None = None
+    amazing_thing: str | None = None
+    surf_reason: str | None = None
+    offer_hosts: str | None = None
+    languages: CouchsurfingComLanguages = field(factory=CouchsurfingComLanguages)
+    address: str | None = None
+    phone: str | None = None
+    birth_date: str | None = None
+    display_name: str | None = None
+    locale: str | None = None
+    emergency_contact: str | None = None
+    emergency_contact_name: str | None = None
+    emergency_contact_phone: str | None = None
+    emergency_contact_email: str | None = None
+    facebook_friends_number: int | None = None
+
+
+@define
+class CouchsurfingComAvailableDays:
+    """Available days of the week from Couchsurfing.com."""
+
+    mon: bool = True
+    tue: bool = True
+    wed: bool = True
+    thu: bool = True
+    fri: bool = True
+    sat: bool = True
+    sun: bool = True
+
+
+@define
+class CouchsurfingComCouch:
+    """Couch/hosting data from Couchsurfing.com export."""
+
+    id: int | None = None
+    max_guests: int | None = None
+    sleeping_arrangement: str | None = None
+    couch_description: str | None = None
+    offer_guests: str | None = None
+    sleeping_description: str | None = None
+    roommate: str | None = None
+    wheelchair_accessible: bool | None = None
+    public_transport_access: str | None = None
+    can_host_children: bool | None = None
+    smoking_ok: bool | None = None
+    smokes: bool | None = None
+    pets_ok: bool | None = None
+    preferred_gender: str | None = None
+    has_children: bool | None = None
+    has_pets: bool | None = None
+    allow_same_day_requests: bool | None = None
+    allow_multiple_parties: bool | None = None
+    available_days_of_week: CouchsurfingComAvailableDays = field(factory=CouchsurfingComAvailableDays)
+    directions: str | None = None
+
+
+@define
+class CouchsurfingComUserData:
+    """User data section from Couchsurfing.com export."""
+
+    id: int | None = None
+    email: str | None = None
+    username: str | None = None
+    current_sign_in_ip: str | None = None
+    last_sign_in_ip: str | None = None
+    profile: CouchsurfingComProfile = field(factory=CouchsurfingComProfile)
+    couch: CouchsurfingComCouch = field(factory=CouchsurfingComCouch)
+
+
+@define
+class CouchsurfingComFriend:
+    """Friend entry from Couchsurfing.com export."""
+
+    profile: str | None = None
+
+
+@define
+class CouchsurfingComFriends:
+    """Friends section from Couchsurfing.com export."""
+
+    friends: list[CouchsurfingComFriend] = field(factory=list)
+
+
+@define
+class CouchsurfingComExport:
+    """Root schema for Couchsurfing.com data export."""
+
+    user_data: CouchsurfingComUserData = field(factory=CouchsurfingComUserData)
+    friends: CouchsurfingComFriends = field(factory=CouchsurfingComFriends)
+    # We don't parse verifications, messages, etc. as they're not used for profile import
+
+
+def structure_couchsurfingcom_export(data: dict[str, Any]) -> CouchsurfingComExport:
+    return cattrs.structure(data, CouchsurfingComExport)
+
+
+COUCHSURFINGCOM_GENDER_MAP = {
+    "male": "Man",
+    "female": "Woman",
+    "other": "Other",
+}
+
+COUCHSURFINGCOM_SLEEPING_ARRANGEMENT_MAP = {
+    "private_room": SleepingArrangement.private,
+    "public_room": SleepingArrangement.common,
+    "shared_room": SleepingArrangement.shared_room,
+    # We explicitly don't support "shared bed".
+}
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class LanguageUpdate:
+    language_code: str
+    fluency: LanguageFluency
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class CouchsurfingComProfileUpdates:
+    """Computed updates from Couchsurfing.com data, ready to be applied to a user."""
+
+    # Profile fields (field_name -> new_value)
+    field_updates: dict[str, Any]
+
+    # Language abilities to set (replaces existing)
+    language_updates: list[LanguageUpdate]
+
+    # Warnings generated during computation
+    warnings: list[str]
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class CouchsurfingComImportResult:
+    """Result of importing Couchsurfing.com data."""
+
+    success: bool
+    fields_updated: list[str] = dataclass_field(default_factory=list)
+    # Note: warnings are internal/debug strings (not localized), used for logging only
+    warnings: list[str] = dataclass_field(default_factory=list)
+    # errors are user-facing and must be localized
+    errors: list[str] = dataclass_field(default_factory=list)
+    # Old values before the import (for logging/rollback)
+    old_values: dict[str, Any] = dataclass_field(default_factory=dict)
+    # New values after the import (for logging)
+    new_values: dict[str, Any] = dataclass_field(default_factory=dict)
+
+    @classmethod
+    def from_error(cls, e: _ImportError, lctx: LocalizationContext) -> CouchsurfingComImportResult:
+        if isinstance(e, TooLarge):
+            return CouchsurfingComImportResult(
+                success=False,
+                errors=[lctx.localize_string("couchsurfingcom_import.errors.json_too_large")],
+            )
+        elif isinstance(e, InvalidJson):
+            return CouchsurfingComImportResult(
+                success=False,
+                errors=[lctx.localize_string("couchsurfingcom_import.errors.invalid_json")],
+            )
+        elif isinstance(e, ValidationError):
+            return CouchsurfingComImportResult(
+                success=False, errors=[lctx.localize_string("couchsurfingcom_import.errors.invalid_format")]
+            )
+        else:
+            raise RuntimeError("Unknown error") from e
+
+
+def _combine_text_sections(*sections: tuple[str, str | None]) -> str | None:
+    """
+    Combine multiple text sections with headers into one string.
+
+    Each section is a tuple of (header, content).
+    """
+    parts = []
+    for header, content in sections:
+        if content and content.strip():
+            parts.append(f"**{header}**\n\n{content.strip()}")
+    return "\n\n---\n\n".join(parts) if parts else None
+
+
+def compute_profile_updates(
+    couchsurfingcom_data: CouchsurfingComExport,
+) -> CouchsurfingComProfileUpdates:
+    field_updates: dict[InstrumentedAttribute[Any], Any] = {}
+    warnings = []
+    language_updates = []
+    profile = couchsurfingcom_data.user_data.profile
+    couch = couchsurfingcom_data.user_data.couch
+
+    # --- Profile Fields ---
+
+    # Name
+    first_name = profile.first_name or ""
+    last_name = profile.last_name or ""
+    if first_name or last_name:
+        full_name = f"{first_name} {last_name}".strip()
+        field_updates[User.name] = full_name
+
+    # Gender
+    if profile.gender:
+        couchers_gender = COUCHSURFINGCOM_GENDER_MAP.get(profile.gender.lower())
+        if couchers_gender:
+            field_updates[User.gender] = couchers_gender
+        else:
+            warnings.append(f"Unknown gender value: {profile.gender}")
+
+    # Simple profile fields
+    field_updates[User.occupation] = profile.occupation
+    field_updates[User.about_me] = profile.about_me
+    field_updates[User.things_i_like] = profile.interests
+
+    # Additional information (combine multiple CS fields)
+    additional_sections = [
+        (CS_SECTION_MY_FAVORITE_MUSIC_MOVIES_BOOKS, profile.media),
+        (CS_SECTION_TEACH_LEARN_SHARE, profile.teach),
+        (CS_SECTION_ONE_AMAZING_THING_IVE_DONE, profile.amazing_thing),
+        (CS_SECTION_WHY_IM_ON_COUCHERS, profile.surf_reason),
+        (CS_SECTION_WHAT_I_CAN_SHARE_WITH_HOSTS, profile.offer_hosts),
+    ]
+    combined_additional = _combine_text_sections(*additional_sections)
+    field_updates[User.additional_information] = combined_additional
+
+    # --- Couch/Hosting Fields ---
+
+    field_updates[User.max_guests] = couch.max_guests
+
+    # Sleeping arrangement
+    if couch.sleeping_arrangement:
+        sleeping = COUCHSURFINGCOM_SLEEPING_ARRANGEMENT_MAP.get(couch.sleeping_arrangement)
+        if sleeping:
+            field_updates[User.sleeping_arrangement] = sleeping
+        else:
+            warnings.append(f"Unknown sleeping arrangement: {couch.sleeping_arrangement}")
+
+    # Combine both sleeping description fields under one CS section header
+    sleeping_texts = [t for t in (couch.couch_description, couch.sleeping_description) if t and t.strip()]
+    combined_sleeping_content = "\n\n".join(sleeping_texts) if sleeping_texts else None
+    sleeping_sections = [(CS_SECTION_DESCRIPTION_OF_SLEEPING_ARRANGEMENT, combined_sleeping_content)]
+    combined_sleeping = _combine_text_sections(*sleeping_sections)
+    field_updates[User.sleeping_details] = combined_sleeping
+
+    # We can't update "has_housemates", because "roommate" is a text field
+    # and can contain anything, including "I live alone" or whatever.
+    field_updates[User.housemate_details] = couch.roommate
+
+    field_updates[User.area] = couch.public_transport_access
+    field_updates[User.other_host_info] = couch.directions
+    field_updates[User.about_place] = couch.offer_guests
+
+    # Hosting preferences
+    field_updates[User.wheelchair_accessible] = couch.wheelchair_accessible
+    field_updates[User.accepts_kids] = couch.can_host_children
+    field_updates[User.has_kids] = couch.has_children
+    field_updates[User.accepts_pets] = couch.pets_ok
+    field_updates[User.has_pets] = couch.has_pets
+    field_updates[User.last_minute] = couch.allow_same_day_requests
+    field_updates[User.smokes_at_home] = couch.smokes
+
+    if couch.smoking_ok is not None:
+        field_updates[User.smoking_allowed] = SmokingLocation.yes if couch.smoking_ok else SmokingLocation.no
+
+    # --- Language Abilities ---
+    added_languages = set[str]()
+
+    # Add fluent languages
+    for lang_code in profile.languages.fluent:
+        if lang_code and lang_code not in added_languages:
+            if language_is_allowed(lang_code):
+                language_updates.append(LanguageUpdate(language_code=lang_code, fluency=LanguageFluency.fluent))
+                added_languages.add(lang_code)
+            else:
+                warnings.append(f"Unknown language code: {lang_code}")
+
+    # Add learning languages (as beginner)
+    for lang_code in profile.languages.learning:
+        if lang_code and lang_code not in added_languages:
+            if language_is_allowed(lang_code):
+                language_updates.append(LanguageUpdate(language_code=lang_code, fluency=LanguageFluency.beginner))
+                added_languages.add(lang_code)
+            else:
+                warnings.append(f"Unknown language code: {lang_code}")
+
+    return CouchsurfingComProfileUpdates(
+        field_updates={col.name: v for col, v in field_updates.items() if v is not None},
+        warnings=warnings,
+        language_updates=language_updates,
+    )
+
+
+def apply_profile_updates(
+    session: Session,
+    user_id: int,
+    updates: CouchsurfingComProfileUpdates,
+    lctx: LocalizationContext,
+    *,
+    overwrite_existing: bool = False,
+) -> CouchsurfingComImportResult:
+    user = session.execute(select(User).where(User.id == user_id)).scalar_one_or_none()
+    if not user:
+        return CouchsurfingComImportResult(
+            success=False,
+            fields_updated=[],
+            warnings=updates.warnings,
+            errors=[lctx.localize_string("couchsurfingcom_import.errors.user_not_found")],
+        )
+
+    fields_updated: list[str] = []
+    old_values: dict[str, Any] = {}
+    new_values: dict[str, Any] = {}
+
+    # Capture old language abilities before any changes
+    old_languages = [{"code": la.language_code, "fluency": la.fluency.name} for la in user.language_abilities]
+
+    # Apply field updates
+    for field_name, new_value in updates.field_updates.items():
+        current_value = getattr(user, field_name)
+        should_update = overwrite_existing or current_value is None
+        if should_update:
+            old_values[field_name] = _serialize_value(current_value)
+            new_values[field_name] = _serialize_value(new_value)
+            fields_updated.append(field_name)
+            setattr(user, field_name, new_value)
+
+    # Apply language updates
+    has_existing_languages = bool(user.language_abilities)
+    if overwrite_existing or not has_existing_languages:
+        for ability in list(user.language_abilities):
+            session.delete(ability)
+        session.flush()
+
+        new_languages = []
+        for lang_update in updates.language_updates:
+            session.add(
+                LanguageAbility(
+                    user_id=user.id,
+                    language_code=lang_update.language_code,
+                    fluency=lang_update.fluency,
+                )
+            )
+            fluency_name = "fluent" if lang_update.fluency == LanguageFluency.fluent else "beginner"
+            fields_updated.append(f"language:{lang_update.language_code}:{fluency_name}")
+            new_languages.append({"code": lang_update.language_code, "fluency": fluency_name})
+
+        if new_languages:
+            old_values["languages"] = old_languages
+            new_values["languages"] = new_languages
+
+    if fields_updated:
+        user.profile_last_updated = now()
+
+    user.has_imported_from_couchsurfing_com = True
+
+    session.flush()
+
+    return CouchsurfingComImportResult(
+        success=True,
+        fields_updated=fields_updated,
+        warnings=updates.warnings,
+        errors=[],
+        old_values=old_values,
+        new_values=new_values,
+    )
+
+
+class _ImportError(Exception): ...
+
+
+class TooLarge(_ImportError):
+    pass
+
+
+class InvalidJson(_ImportError):
+    pass
+
+
+class ValidationError(_ImportError):
+    pass
+
+
+def parse_couchsurfingcom_data(json_data: str) -> CouchsurfingComExport:
+    if len(json_data) > MAX_JSON_SIZE:
+        raise TooLarge()
+
+    try:
+        data: dict[str, Any] = json.loads(json_data)
+    except json.JSONDecodeError as e:
+        raise InvalidJson() from e
+
+    relevant_data = {key: data[key] for key in ("user_data", "friends") if key in data}
+
+    try:
+        structured = structure_couchsurfingcom_export(relevant_data)
+    except Exception as e:
+        raise ValidationError() from e
+
+    return structured
+
+
+def import_couchsurfingcom_json(
+    session: Session,
+    user_id: int,
+    json_data: str,
+    lctx: LocalizationContext,
+    *,
+    overwrite_existing: bool = False,
+) -> CouchsurfingComImportResult:
+    """
+    Import Couchsurfing.com profile data from a JSON string.
+
+    Args:
+        session: Database session
+        user_id: ID of user to update
+        json_data: JSON string containing Couchsurfing.com export data
+        lctx: Localization context for generating localized errors
+        overwrite_existing: Whether to overwrite existing profile fields
+
+    Returns:
+        CouchsurfingComImportResult with success status, updated fields, warnings, and errors
+    """
+    try:
+        couchsurfingcom_data = parse_couchsurfingcom_data(json_data)
+        updates = compute_profile_updates(couchsurfingcom_data)
+        result = apply_profile_updates(session, user_id, updates, lctx, overwrite_existing=overwrite_existing)
+    except _ImportError as e:
+        result = CouchsurfingComImportResult.from_error(e, lctx)
+
+    session.add(
+        CouchsurfingComImportAttempt(
+            user_id=user_id,
+            success=result.success,
+            raw_json=json_data[:MAX_JSON_SIZE],
+            old_values=result.old_values,
+            new_values=result.new_values,
+            warnings=result.warnings,
+            errors=result.errors,
+        )
+    )
+
+    return result
--- a/app/backend/src/couchers/i18n/locales/en.json
+++ b/app/backend/src/couchers/i18n/locales/en.json
@@ -1,4 +1,12 @@
 {
+  "couchsurfingcom_import": {
+    "errors": {
+      "json_too_large": "The uploaded file is too large. Please ensure it is less than 20MB.",
+      "invalid_json": "The uploaded file is not valid JSON.",
+      "invalid_format": "The uploaded file does not match the expected Couchsurfing™ export format.",
+      "user_not_found": "User not found."
+    }
+  },
   "errors": {
     "account_not_found": "An account with that username or email was not found.",
     "account_suspended": "Your account is suspended.",
--- a/app/backend/src/couchers/models/users.py
+++ b/app/backend/src/couchers/models/users.py
@@ -47,6 +47,7 @@ from couchers.utils import get_coordinates, last_active_coarsen, now
 if TYPE_CHECKING:
     from couchers.models import UserBadge
     from couchers.models.admin import UserAdminTag
+    from couchers.models.couchsurfingcom_import import CouchsurfingComImportAttempt
     from couchers.models.public_trips import PublicTrip
     from couchers.models.rest import InviteCode, ModerationUserList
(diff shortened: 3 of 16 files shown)


WouldYouKindly:
@nabramow please check the frontend code - it was all Claude! Also please tell me what you think about extracting the zip archive in the browser

app/web/.env.development
```diff
@@ -1,11 +1,11 @@
-NEXT_PUBLIC_COUCHERS_ENV=preview
-NEXT_PUBLIC_BASE_URL="https://next.couchershq.org"
-NEXT_PUBLIC_API_BASE_URL="https://next.couchershq.org/api"
-NEXT_PUBLIC_MEDIA_BASE_URL="https://dev-user-media.couchershq.org"
-NEXT_PUBLIC_CONSOLE_BASE_URL="https://next-console.couchershq.org"
+NEXT_PUBLIC_COUCHERS_ENV=dev
```
aapeliv, line 1: please undo this bit

aapeliv (changes requested):
sorry to be a bit annoying, but please update all references to say in code "CouchSurfing.com" or "CouchSurfingCom"; we want to be very clear on this even in code. In user-visible stuff and ideally at the top of cs_import, please use "Couchsurfing™"

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+    couchsurfingcom_data: CouchSurfingComExport,
+    *,
+    overwrite_existing: bool = False,
+) -> CouchSurfingComImportResult:
+    updates = compute_profile_updates(couchsurfingcom_data)
+
+    return apply_profile_updates(session, user_id, updates, overwrite_existing=overwrite_existing)
+
+
+def import_couchsurfingcom_json(
```
tristanlabelle, line 456: I suggest having one file for all the CS.com data structures and parsing, and one file for the importing logic into our DB, since logically those are two layers

app/backend/src/couchers/couchsurfingcom_import.py
tristanlabelle, line 1: I think we can find a better home for this than a "helpers" directory, which doesn't really group code meaningfully.

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+
+@define
+class CouchSurfingComFriends:
+    """Friends section from CouchSurfing.com export."""
+
+    friends: list[CouchSurfingComFriend] = field(factory=list)
+
+
+@define
+class CouchSurfingComExport:
```
tristanlabelle: Nice way to define the schema!

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+    gender: str | None = None
+    occupation: str | None = None
+    about_me: str | None = None
+    interests: str | None = None
+    media: str | None = None
+    teach: str | None = None
+    amazing_thing: str | None = None
+    surf_reason: str | None = None
+    offer_hosts: str | None = None
+    languages: CouchSurfingComLanguages = field(factory=CouchSurfingComLanguages)
```
tristanlabelle: Curious: does the `factory=Xxx` force declaring `Xxx` before this class? The type annotation doesn't since Python 3.14

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
   "fill_in_who_i_am": "1. Fill in your \"Who I am\" section with a few sentences about yourself",
   "upload_photo": "2. Upload a photo (by clicking on the avatar on the edit page)",
   "edit_profile_button_text": "Edit your profile",
+  "import_from_couchsurfingcom_button_text": "Import from Couchsurfing™",
```
tristanlabelle: We can group all of these strings under a "couchsurfingcom_import" node

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
...
+  "import_from_couchsurfingcom_success": "Successfully imported your Couchsurfing™ data!",
+  "import_from_couchsurfingcom_no_fields": "No fields were updated from your Couchsurfing™ data.",
```
tristanlabelle: Could you add an `_error` suffix to these? That convention can help translators and standardizing our text. For example this error string has a final period whereas the others below don't.

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
...
+  "import_from_couchsurfingcom_no_json": "No JSON file found",
+  "import_from_couchsurfingcom_invalid_zip": "Could not read the ZIP file",
+  "import_from_couchsurfingcom_invalid_file": "Please select a .zip or .json file",
+  "import_from_couchsurfingcom_modal_instructions": "Go to couchsurfing.com and export your data. Then upload the archive here.",
```
tristanlabelle: I think we should call out somewhere in the UI that this is only importing profile data and not references since I think that's what most people will be hoping to import.

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+            warnings.append(f"Unknown gender value: {profile.gender}")
+
+    # Simple profile fields
+    field_updates["occupation"] = profile.occupation
+    field_updates["about_me"] = profile.about_me
+    field_updates["things_i_like"] = profile.interests
+
+    # Additional information (combine multiple CS fields)
+    additional_sections = [
+        ("Music, Films, Books", profile.media),
```
tristanlabelle: We'll want to localize these, even though we have no way to know what language the CS profile is written in.

tristanlabelle (commented):
I like the direction, left some i18n and small code organization comments. Happy to help with the final review too.

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: I think it's too much hassle tbh. The code is unlikely to be used anywhere else, and it's 500 lines. I think it's good enough to split parsing/updating the db in separate functions.

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: Moved to the top level.

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: Yeah, it does, since `field(factory=CouchSurfingComLanguages)` is executed at module import time.

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: @tristanlabelle any suggestions? I think it's tricky (e.g., user's language is Russian, but the profile is written in English - if we show these in Russian, it's incorrect.) I would not do it for now if there's no obviously correct solution.

app/web/features/dashboard/locales/en.json
WouldYouKindly: Good idea, done.

app/web/features/dashboard/locales/en.json
WouldYouKindly: done

app/web/features/dashboard/locales/en.json
WouldYouKindly: done

WouldYouKindly:
@tristanlabelle @aapeliv this is ready for a final review!

app/backend/src/couchers/models/users.py
```diff
@@ -319,6 +319,9 @@ class User(Base, kw_only=True):
     # whether mods have marked this user has having to update their location
     needs_to_update_location: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
 
+    # whether the user has imported their profile from CouchSurfing
+    has_imported_from_couchsurfing: Mapped[bool] = mapped_column(Boolean, server_default=expression.false(), init=False)
```
tristanlabelle: Is there any value in referencing the log table entry instead? Or do those get purged? I'm not familiar with how we use tables for logging.

app/proto/account.proto
```diff
@@ -119,6 +119,10 @@ service Account {
   rpc UpdateMyVolunteerInfo(UpdateMyVolunteerInfoReq) returns (GetMyVolunteerInfoRes) {
     // Updates the current user's volunteer status if they are volunteer
   }
+
+  rpc ImportFromCouchsurfing(ImportFromCouchsurfingReq) returns (ImportFromCouchsurfingRes) {
```
tristanlabelle: I anticipate we'd prefer `ImportFromCouchSurfingCom` but leaving that to @aapeliv 's rereview

app/proto/account.proto
```diff
@@ -371,3 +378,21 @@ message UpdateMyVolunteerInfoReq {
...
+  string couchsurfing_json = 1;
+  // Whether to overwrite existing profile fields
+  bool overwrite_existing = 2;
+}
+
+message ImportFromCouchsurfingRes {
+  // Whether the import was successful
+  bool success = 1;
+  // Fields that were updated
+  repeated string fields_updated = 2;
```
tristanlabelle: What's the format of the field names? Is it a stable schema? Does the frontend have a way to localize the result? Or are those localized by the backend?

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
   "fill_in_who_i_am": "1. Fill in your \"Who I am\" section with a few sentences about yourself",
   "upload_photo": "2. Upload a photo (by clicking on the avatar on the edit page)",
   "edit_profile_button_text": "Edit your profile",
+  "couchsurfingcom_import": {
+    "button_text": "Import from Couchsurfing™",
+    "no_json": "No JSON file found",
```
tristanlabelle: `_error` suffix ideally

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
...
+    "invalid_zip": "Could not read the ZIP file",
+    "invalid_file": "Please select a .zip or .json file",
+    "modal_instructions": "To import your data, you have to log in to <1>couchsurfing.com</1>, and press \"Manage Account\" in the right top corner. Then press \"Request Data\".\n\nOnce you receive an email from them, click \"See My Data\" in it - it will take you to couchsurfing.com again. There, press the same \"Manage Account\" in the top right corner, and then \"Download Data\". A download will start; once it's done, press \"Select file\" below, and upload the archive from your computer.\n\nPlease note: we can only import profile information such as your location, house rules, and interests. We cannot import your references, friends, or message history.",
+    "select_file": "Select file"
```
tristanlabelle: This key doesn't provide context on how the string is used. Suggestion: `select_file_button_caption`

(generally the key shouldn't be the same as the text)

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,14 @@
   "fill_in_who_i_am": "1. Fill in your \"Who I am\" section with a few sentences about yourself",
   "upload_photo": "2. Upload a photo (by clicking on the avatar on the edit page)",
   "edit_profile_button_text": "Edit your profile",
+  "couchsurfingcom_import": {
+    "button_text": "Import from Couchsurfing™",
+    "no_json": "No JSON file found",
+    "invalid_zip": "Could not read the ZIP file",
+    "invalid_file": "Please select a .zip or .json file",
+    "modal_instructions": "To import your data, you have to log in to <1>couchsurfing.com</1>, and press \"Manage Account\" in the right top corner. Then press \"Request Data\".\n\nOnce you receive an email from them, click \"See My Data\" in it - it will take you to couchsurfing.com again. There, press the same \"Manage Account\" in the top right corner, and then \"Download Data\". A download will start; once it's done, press \"Select file\" below, and upload the archive from your computer.\n\nPlease note: we can only import profile information such as your location, house rules, and interests. We cannot import your references, friends, or message history.",
```
tristanlabelle: Shouldn't the double new lines be paragraphs (`<p>`) in the dialog? I think we could break it down in a few strings that way.

tristanlabelle: I think the following would be enough. Even though I don't expect CS to change in any way, providing click-by-click instructions seems a little brittle.

> To obtain your Couchsurfing™ data, use the "Request Data" feature from your account settings on <1>couchsurfing.com</1>.
> 
> Please note: we can only import profile information such as your location, house rules, and interests. We cannot import your references, friends, or message history.

app/backend/src/couchers/couchsurfingcom_import.py
tristanlabelle: We use "Couch surfing" two words to distinguish from the brand Couchsurfing. In this case we do want to refer to the Couchsurfing brand so I'd expect `CouchsurfingCom`. Same with all the "CouchSurfing.com" comments which I think should either be `couchsurfing.com` or `Couchsurfing(tm)` since URLs have no capitalization.

@aapeliv , do you mean us to avoid using "Couchsurfing" even when referring to Couchsurfing(tm)?

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+COUCHSURFINGCOM_GENDER_MAP = {
+    "male": "Man",
+    "female": "Woman",
+    "other": "Other",
+}
+
+COUCHSURFINGCOM_SLEEPING_ARRANGEMENT_MAP = {
+    "private_room": SleepingArrangement.private,
+    "public_room": SleepingArrangement.common,
+    "shared_room": SleepingArrangement.shared_room,
```
tristanlabelle, line 193: Add a comment about the shared surface option that we don't map by design.

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+        couchers_gender = COUCHSURFINGCOM_GENDER_MAP.get(profile.gender.lower())
+        if couchers_gender:
+            field_updates["gender"] = couchers_gender
+        else:
+            warnings.append(f"Unknown gender value: {profile.gender}")
+
+    # Simple profile fields
+    field_updates["occupation"] = profile.occupation
+    field_updates["about_me"] = profile.about_me
+    field_updates["things_i_like"] = profile.interests
```
tristanlabelle: What's the source of truth for these field names? Can we reference symbol names that will survive refactorings? Especially if those are returned to the frontend, we need some guarantee of stability and possibly ability to localize.

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,477 @@
...
+        full_name = f"{first_name} {last_name}".strip()
+        field_updates["name"] = full_name
+
+    # Gender
+    if profile.gender:
+        couchers_gender = COUCHSURFINGCOM_GENDER_MAP.get(profile.gender.lower())
+        if couchers_gender:
+            field_updates["gender"] = couchers_gender
+        else:
+            warnings.append(f"Unknown gender value: {profile.gender}")
```
tristanlabelle, line 276: The frontend will have no way to translate those warnings. I think we should return localized messages. If you don't want to tackle that, feel free to file a task and I can follow up, as long as the API shape is right.

app/backend/src/couchers/couchsurfingcom_import.py
tristanlabelle: This is something I can follow-up on. I lean towards translating those to avoid an English-centric experience, but the tricky part is that CS tends to naturally be English-centric. I wonder if it'd be overkill to have a checkbox say "My profile is in English".

tristanlabelle (commented):
I still have a bit more reviewing to do. In general the implementation looks great and my main concern is that our API shape be localizable, especially the response fields below. See my comments for discussion.

```
  // Fields that were updated
  repeated string fields_updated = 2;
  // Any warnings from the import (non-fatal issues)
  repeated string warnings = 3;
  // Any errors encountered (fatal issues)
  repeated string errors = 4;

File: app/web/features/dashboard/locales/en.json (hunk 14 above)
Comment (WouldYouKindly) [reply app/web/features/dashboard/locales/en.json]: Not sure about that. Yeah, they may change the process, and a user will be confused. But OTOH the detailed instructions may help some users who otherwise would give up. IDK. Let's think about it when it breaks? :)

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 15 above)
Comment (WouldYouKindly) [reply app/backend/src/couchers/couchsurfingcom_import.py:193]: Not sure what you mean? Do they have a shared surface option?

File: app/web/features/dashboard/locales/en.json (hunk 14 above)
Comment (WouldYouKindly) [reply app/web/features/dashboard/locales/en.json]: @tristanlabelle by the way, do you have a subscription for couchsurfing.com? Does the data export process look different for you? If so, could you please describe it?

**Comment by WouldYouKindly** [comment]
> I wonder if it'd be overkill to have a checkbox say "My profile is in English".

@tristanlabelle Yeah I'd say it's a bit too much :) I don't think it will be a very common case.

File: app/backend/src/couchers/models/users.py (hunk 9 above)
Comment (WouldYouKindly) [reply app/backend/src/couchers/models/users.py]: Amazing idea. Done.

File: app/proto/account.proto (hunk 10 above)
Comment (WouldYouKindly) [reply app/proto/account.proto]: Thanks, done

File: app/proto/account.proto (hunk 11 above)
Comment (WouldYouKindly) [reply app/proto/account.proto]: Extracted the field names into a enum https://github.com/Couchers-org/couchers/pull/8188/changes#diff-7fb0b7558b5bcdbe8bce003c631de5fc695b1d63feb6e0a1e6e8c7e8c6b281faR35

File: app/web/features/dashboard/locales/en.json (hunk 12 above)
Comment (WouldYouKindly) [reply app/web/features/dashboard/locales/en.json]: Added

File: app/web/features/dashboard/locales/en.json (hunk 13 above)
Comment (WouldYouKindly) [reply app/web/features/dashboard/locales/en.json]: changed

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 3 above)
Comment (WouldYouKindly) [reply app/backend/src/couchers/couchsurfingcom_import.py]: Renamed to `CouchsurfingCom` (with the lowercase `s`) in all places

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 16 above)
Comment (WouldYouKindly) [reply app/backend/src/couchers/couchsurfingcom_import.py]: https://github.com/Couchers-org/couchers/pull/8188/changes#diff-7fb0b7558b5bcdbe8bce003c631de5fc695b1d63feb6e0a1e6e8c7e8c6b281faR35

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 17 above)
Comment (WouldYouKindly) [reply app/backend/src/couchers/couchsurfingcom_import.py:276]: @tristanlabelle What should I do to make these localizable? This is my first real backend feature, sorry for my ignorance

File: app/web/features/dashboard/locales/en.json (hunk 14 above)
Comment (WouldYouKindly) [reply app/web/features/dashboard/locales/en.json]: Good idea, done

File: app/proto/account.proto (hunk 11 above)
Comment (tristanlabelle) [reply app/proto/account.proto]: Cool, that works! Could you just update the comment in the protobuf file to specify that the source of truth is in that enum?

File: app/web/features/dashboard/locales/en.json (hunk 14 above)
Comment (tristanlabelle) [reply app/web/features/dashboard/locales/en.json]: Actually yes, the steps are incorrect according to my version of the website. Did you base them on an app or different version? If so that points towards using more generic instructions.

image

image

image

image

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 15 above)
Comment (tristanlabelle) [reply app/backend/src/couchers/couchsurfingcom_import.py:193]: Yes, there's the creepy option. I remember seeing somewhere around https://couchers.org/plan that we explicitly did not want to support that.

How did you figure out the schema such that you know 3 out of 4 values? 🤔

image

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 16 above)
Comment (tristanlabelle) [reply app/backend/src/couchers/couchsurfingcom_import.py]: Sweet!

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 8 above)
Comment (tristanlabelle) [reply app/backend/src/couchers/couchsurfingcom_import.py]: I asked translators for opinions: https://couchersorg.slack.com/archives/C02NW0X3ZNK/p1775605686972989 . Let's not block on this since it doesn't impact the API and can be changed later.

File: app/backend/src/couchers/couchsurfingcom_import.py (hunk 17 above)
Comment (tristanlabelle) [reply app/backend/src/couchers/couchsurfingcom_import.py:276]: No problem at all! First review https://github.com/Couchers-org/couchers/blob/develop/docs/localization.md for general info about how localization works and best practices for strings.

In your case I suggest we add the strings to a `couchsurfingcom_import.errors` node in https://github.com/Couchers-org/couchers/blob/develop/app/backend/src/couchers/i18n/locales/en.json . Then you need to take in a `LocalizationContext` parameter throughout your methods (typically last param) and use it to look up the strings.

File: app/backend/src/couchers/models/logging.py [hunk 18]
```diff
@@ -115,3 +115,42 @@ class EventLog(Base, kw_only=True):
         Index("ix_logging_event_log_user_id_created", "user_id", "created"),
         {"schema": "logging"},
     )
+
+
+class CouchsurfingImportLog(Base, kw_only=True):
```
tristanlabelle: `CouchsurfingComImportLog` (it seemed important to @aapeliv for legal reasons)

app/backend/src/couchers/models/logging.py
```diff
@@ -115,3 +115,42 @@ class EventLog(Base, kw_only=True):
...
+    old_values: Mapped[dict[str, Any]] = mapped_column(JSONB)
+
+    # new values of fields after the import (field_name -> new_value)
+    new_values: Mapped[dict[str, Any]] = mapped_column(JSONB)
+
+    # warnings generated during import
+    warnings: Mapped[list[str]] = mapped_column(JSONB)
+
+    # errors encountered (if any)
+    errors: Mapped[list[str]] = mapped_column(JSONB)
```
tristanlabelle: Worth calling out that these will be localized by design.

```suggestion
    # warnings generated during import (localized)
    warnings: Mapped[list[str]] = mapped_column(JSONB)

    # errors encountered, if any (localized)
    errors: Mapped[list[str]] = mapped_column(JSONB)
```

app/backend/src/couchers/models/logging.py
```diff
@@ -115,3 +115,42 @@ class EventLog(Base, kw_only=True):
...
+    created: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), init=False)
+
+    # user who performed the import
+    user_id: Mapped[int] = mapped_column(BigInteger)
+
+    # whether the import was successful
+    success: Mapped[bool] = mapped_column(Boolean)
+
+    # old values of fields before the import (field_name -> old_value)
+    old_values: Mapped[dict[str, Any]] = mapped_column(JSONB)
```
tristanlabelle: If we pulled the field enum into `couchers/models`, would we be able to use it here as the key? Not sure if that's allowed. Otherwise specify in the comment that the keys are from the enum we use as a source of truth.

app/proto/account.proto
```diff
@@ -119,6 +119,10 @@ service Account {
   rpc UpdateMyVolunteerInfo(UpdateMyVolunteerInfoReq) returns (GetMyVolunteerInfoRes) {
     // Updates the current user's volunteer status if they are volunteer
   }
+
+  rpc ImportFromCouchsurfing(ImportFromCouchsurfingComReq) returns (ImportFromCouchsurfingComRes) {
```
tristanlabelle: This should also be `ImportFromCouchsurfingCom`

app/proto/account.proto
```diff
@@ -371,3 +378,21 @@ message UpdateMyVolunteerInfoReq {
...
+}
+
+message ImportFromCouchsurfingComRes {
+  // Whether the import was successful
+  bool success = 1;
+  // Fields that were updated
+  repeated string fields_updated = 2;
+  // Any warnings from the import (non-fatal issues)
+  repeated string warnings = 3;
+  // Any errors encountered (fatal issues)
```
tristanlabelle: Our API should specify what's localized.

```suggestion
  // Localized warnings from the import (non-fatal issues)
  repeated string warnings = 3;
  // Localized errors encountered (fatal issues)
```

app/backend/Makefile
tristanlabelle, line 1: Could you extract to a separate PR? For independent PR description, review and revertability.

app/backend/src/tests/test_account.py
```diff
@@ -1235,3 +1235,74 @@ def test_volunteer_stuff(db):
...
+                overwrite_existing=True,
+            )
+        )
+
+        assert res.success
+        assert "name" in res.fields_updated
+        assert "gender" in res.fields_updated
+        assert "occupation" in res.fields_updated
+        assert "about_me" in res.fields_updated
+        assert "max_guests" in res.fields_updated
```
tristanlabelle: Can you use the enum now?

app/backend.dev.env
```diff
@@ -43,7 +43,7 @@ REPORTS_EMAIL_RECIPIENT=reports@couchers.org.invalid
 CONTRIBUTOR_FORM_EMAIL_RECIPIENT=forms@couchers.org.invalid
 MODS_EMAIL_RECIPIENT=mods@couchers.org.invalid
 
-ENABLE_EMAIL=1
+ENABLE_EMAIL=0
```
tristanlabelle: This change seems unrelated to this PR and potentially undesirable?

app/web/service/account.ts
```diff
@@ -219,3 +220,14 @@ export async function listInviteCodes() {
   const res = await client.account.listInviteCodes(new Empty());
   return res.toObject();
 }
+
+export async function importFromCouchsurfing(
```
tristanlabelle: Let's do it all the way: `importFromCouchsurfingCom`

app/backend/src/tests/test_cs_import.py
```diff
@@ -0,0 +1,666 @@
...
+        result = update_user_from_couchsurfingcom_export(
+            session, user.id, couchsurfingcom_export, overwrite_existing=True
+        )
+
+        assert result.success
+        assert "name" in result.fields_updated
+        assert "gender" in result.fields_updated
+        assert "occupation" in result.fields_updated
+        assert "about_me" in result.fields_updated
+        assert "things_i_like" in result.fields_updated
```
tristanlabelle: Can you use the enum now?

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,507 @@
...
+
+    # Simple profile fields
+    field_updates[ImportField.occupation] = profile.occupation
+    field_updates[ImportField.about_me] = profile.about_me
+    field_updates[ImportField.things_i_like] = profile.interests
+
+    # Additional information (combine multiple CS fields)
+    additional_sections = [
+        ("Music, Films, Books", profile.media),
```
tristanlabelle: I think we should use the same title strings as on CS, since that was the title that the user was following when they filled in the text. For example this one is `My Favorite Music, Movies & Books` on the website.

Also I'd extract them to constants at the top of the file to call out that we have those English-only sections.

CS sections:

About Me
Why I'm on Couchsurfing (maybe don't use this one verbatim)
My Interests 
My Favorite Music, Movies & Books
One Amazing Thing I've Done
Teach, Learn, Share
What I Can Share With Hosts
Description of Sleeping Arrangement
Roommate Situation
Public Transportation Access
What I Can Share With Guests
Additional Information

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,507 @@
...
+    fields_updated: list[str] = []
+    old_values: dict[str, Any] = {}
+    new_values: dict[str, Any] = {}
+
+    # Capture old language abilities before any changes
+    old_languages = [{"code": la.language_code, "fluency": la.fluency.name} for la in user.language_abilities]
+
+    # Apply field updates
+    for field_name, new_value in updates.field_updates.items():
+        current_value = getattr(user, field_name)
```
tristanlabelle, line 386: If we can do `getattr` here, then `ImportField` is not the source of truth but rather the property/column name of the User table. Do we want to rely on that as stable tokens in our API? If so, then I think the enum is misleading since it seems like it defines the schema, and we should be using something like `User.sleeping_details.__name__` or whatever version of that works.

tristanlabelle (commented):
I really appreciate your work and seeing this take shape. I have another few concerns about the stability/source of truth for profile update fields. If that part of the design is too tricky, we could also simply not include it in the response object for this v1 and follow up afterwards to add it.

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: @tristanlabelle claude. TBH this will need some testing, because I only tested it on my export, which I cannot even modify, since I have no access to my account.

tristanlabelle: Yeah that sounds highly prone to hallucinations given that I can't find a documentation of their schema online.

app/backend/src/couchers/servicers/account.py
```diff
@@ -175,6 +177,14 @@ def GetAccountInfo(
             user.last_donated is None or user.last_donated < DONATION_DRIVE_START
         )
 
+        has_imported_from_couchsurfing = session.execute(
+            select(
+                exists()
+                .where(CouchsurfingComImportLog.user_id == context.user_id)
+                .where(CouchsurfingComImportLog.success == True)
+            )
+        ).scalar_one()
```
aapeliv: if you put it in the logging schema, then don't depend on it for functionality.

aapeliv (commented):
My main question with this is what the plan is for maintenance, error/bug reporting, and testing. We don't have a trove of these export .zips, so we can't just run a bunch through to check how they work. What if they change their .zip format?

This seems naturally quite brittle, so we would probably need an easy way to report bugs for this. I also am not sure if the import log will be particularly useful? I think we would kind of want to ask people if we can store their .zips somewhere for debugging?

WouldYouKindly:
> My main question with this is what the plan is for maintenance, error/bug reporting, and testing. We don't have a trove of these export .zips, so we can't just run a bunch through to check how they work. What if they change their .zip format?
> 
> This seems naturally quite brittle, so we would probably need an easy way to report bugs for this. I also am not sure if the import log will be particularly useful? I think we would kind of want to ask people if we can store their .zips somewhere for debugging?

@aapeliv I actually wanted to ask the whole team to provide me with their exports in a bit, to have some testing data :)

> What if they change their .zip format?

We'll have to adjust the code, I don't think there's any other solution.

> we would probably need an easy way to report bugs for this

I think Sentry should be enough. I just have to make sure that the tracebacks contains all the interesting info.

>  I also am not sure if the import log will be particularly useful? I think we would kind of want to ask people if we can store their .zips 

Yeah I was thinking of storing the whole JSON files in the log. If we exclude message history and verification data from that, I don't think we even need to ask for permission, since the user kind of consents to giving us the data when they decide to import? Or do you think I'm wrong here?

app/backend/src/couchers/servicers/account.py
WouldYouKindly: Good point. Also, if we want to store failed attempts for debugging, we'll have potentially mulitple log entries for a single user.

app/web/service/account.ts
WouldYouKindly: done

app/backend.dev.env
WouldYouKindly: rolled back. @aapeliv does it make sense to disable email by default on dev? When I need to register a new user on my local machine, I really don't want to give it a real email address :)

app/backend/src/tests/test_account.py
WouldYouKindly: TBH I think it makes more sense to use the strings in tests

app/backend/Makefile
WouldYouKindly: sure, done

app/proto/account.proto
WouldYouKindly: thanks

app/backend/src/couchers/models/logging.py
WouldYouKindly: thanks

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: Right. Done! 

Why I'm on Couchsurfing  -> Why I'm on Couchers

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: @tristanlabelle so I removed warnings and fields_updated from the API response. We now take the user to the profile page after a successful import, so there's no need to show them the list of imported fields. Also, no need for localization! :)

app/backend/src/couchers/models/logging.py
WouldYouKindly: I removed the enum entirely, and instead use fields on the User object. I think it's fine since we're not returning them to the frontend anymore. Added a comment about where the fields come from.

app/backend/src/couchers/i18n/locales/en.json
```diff
@@ -1,4 +1,11 @@
 {
+  "couchsurfingcom_import": {
+    "errors": {
+      "json_too_large": "The uploaded file is too large. Please ensure it is less than 20MB.",
+      "invalid_json": "The uploaded file is not valid JSON.",
+      "invalid_format": "The uploaded file does not match the expected Couchsurfing export format."
```
tristanlabelle:
```suggestion
      "invalid_format": "The uploaded file does not match the expected Couchsurfing™ export format."
```

app/proto/account.proto
```diff
@@ -162,6 +166,9 @@ message GetAccountInfoRes {
   bool is_volunteer = 18;
 
   bool should_show_donation_banner = 20;
+
+  // whether the user has already imported their profile from CouchSurfing
+  bool has_imported_from_couchsurfing = 21;
```
tristanlabelle: Another reference to the brand

```
  // whether the user has already imported their profile from couchsurfing.com
  bool has_imported_from_couchsurfing_com = 21;
```

app/proto/account.proto
```diff
@@ -371,3 +378,17 @@ message UpdateMyVolunteerInfoReq {
   // the URL the link should take you to
   google.protobuf.StringValue link_url = 6;
 }
+
+message ImportFromCouchsurfingComReq {
+  // JSON data from CouchSurfing export
```
tristanlabelle: Either use `couchsurfing.com` or `Couchsurfing™`.

```suggestion
  // JSON data from couchsurfing.com export
```

app/backend/src/tests/test_cs_import.py
tristanlabelle: Or some variant of `User.occupation.__name__` for refactorability?

app/backend/src/tests/test_cs_import.py
tristanlabelle, line 1: Suggestion: Related to my earlier comment about splitting the json parsing and the database importing into separate files, if we split the parsing tests and the database importing tests, we wouldn't need the DB for the first half and could test at a more granular/unit level.

app/web/features/dashboard/DashboardBanners.tsx
```diff
@@ -72,6 +211,48 @@ export default function DashboardBanners() {
           )}
         </>
       )}
+      <Dialog
+        open={isImportModalOpen}
+        onClose={handleCloseModal}
+        aria-labelledby="import-cs-dialog-title"
+      >
+        <DialogTitle id="import-cs-dialog-title" onClose={handleCloseModal}>
+          {t("dashboard:couchsurfingcom_import.button_text")}
```
tristanlabelle: It's generally best to duplicate the string as `dialog_title` to allow for translators to use different translations for each. For example an imperative for the button and an infinitive for the title.

tristanlabelle (dismissed):
Thanks for all the iteration. I'm confident with the API and implementation. My last few comments are minor, feel free to follow up separately / skip.

I personally am not concerned about the feature breaking due to the CS schema changing. For one, CS is on life support, but also this is an intrinsic issue with what we're trying to do so we just have to deal with it if it happens. I think Sentry + the log should provide the info we need for diagnosing.

app/proto/account.proto
WouldYouKindly: thanks!

app/web/features/dashboard/DashboardBanners.tsx
WouldYouKindly: done. Please check if I understood you correctly

app/backend/src/tests/test_cs_import.py
WouldYouKindly: @tristanlabelle I'll think about it! Let's not block this PR though?

WouldYouKindly: I didn't split into multiple files, but I did refaftor the code a bit, so now more tests don't use the database. thanks for the suggestion!

aapeliv:
> image

We should split this dialog into a "3 easy steps" format. This is too much reading to do.

Also could we add some metrics to track how people are using this feature, how many start, complete, etc

WouldYouKindly:
@aapeliv does this look good?

image

app/web/features/dashboard/locales/en.json
```diff
@@ -4,6 +4,27 @@
...
+    "step2_title": "Download the archive",
+    "step2_description": "When the email arrives, open \"See My Data\" and download the archive file.",
+    "step3_title": "Upload your file",
+    "step3_description": "Upload the archive you downloaded from Couchsurfing™.",
+    "open_couchsurfing": "Open couchsurfing.com",
+    "drop_file_here": "Drag & drop file here",
+    "or": "or",
+    "choose_file": "Choose file",
+    "no_file_selected": "No file selected",
+    "info_title": "What we can import",
```
tristanlabelle: It helps translators to have string keys communicate the role of the string in the UI, as you've done with `_title` and `_descriptions`. It especially helps with shorter strings, which are more ambiguous, e.g. Chinese has two different "or"'s depending on whether it's interrogative or two options.

Suggestions:

open_couchsurfing_link
drop_file_caption
upload_method_choice_or_text
choose_file_button
no_file_selected_placeholder

See guidelines in https://github.com/Couchers-org/couchers/blob/develop/docs/localization.md

app/web/features/dashboard/locales/en.json
tristanlabelle: The latest' 3-step process is great!

app/web/features/dashboard/DashboardBanners.tsx
```diff
@@ -1,33 +1,137 @@
...
+    setIsImportModalOpen(true);
+  };
+
+  const handleCloseModal = () => {
+    setIsImportModalOpen(false);
+  };
+
+  /**
+   * Keep only relevant fields from the JSON to reduce payload size.
+   * Only user_data and friends are needed for profile import.
```
tristanlabelle: Why friends? The instructions say

>  We can't import references, friends, or message history.

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,520 @@
...
+    *,
+    overwrite_existing: bool = False,
+) -> CouchsurfingComImportResult:
+    user = session.execute(select(User).where(User.id == user_id)).scalar_one_or_none()
+    if not user:
+        return CouchsurfingComImportResult(
+            success=False,
+            fields_updated=[],
+            warnings=updates.warnings,
+            errors=[f"User with ID {user_id} not found"],
```
tristanlabelle: Other errors are localized

app/backend/src/couchers/couchsurfingcom_import.py
tristanlabelle: Works for me!

app/backend/src/couchers/couchsurfingcom_import.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+@dataclass(frozen=True, slots=True, kw_only=True)
+class CouchsurfingComImportResult:
+    """Result of importing Couchsurfing.com data."""
+
+    success: bool
+    fields_updated: list[str] = dataclass_field(default_factory=list)
+    warnings: list[str] = dataclass_field(default_factory=list)
+    errors: list[str] = dataclass_field(default_factory=list)
```
tristanlabelle, line 226: it's a little error prone that errors are localized but warnings are not. Better call out with a comment or even a longer variable name.

app/backend/src/couchers/models/logging.py
tristanlabelle: Sweet!

tristanlabelle (dismissed):
This is good to merge as far as I'm concerned -- feel free to resolve comments and consider following up as relevant.

app/web/features/dashboard/locales/en.json
WouldYouKindly: done

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: fixed, thanks

app/backend/src/couchers/couchsurfingcom_import.py
WouldYouKindly: added a comment

app/web/features/dashboard/DashboardBanners.tsx
WouldYouKindly: fixed thanks

app/web/.env.development
WouldYouKindly: done

app/docker-compose.local-backend.yml
tristanlabelle, line 1: Are these changes right? I wouldn't expect this file to change.

app/backend/src/couchers/couchsurfingcom_import.py
tristanlabelle, line 1: Did you have a chance to test with a post-redesign CS dump? I can't access CS anymore unfortunately so can't help there.

tristanlabelle (approved):
It's hard to know what changed in the last few pushes but assuming it's mostly rebasing, I was happy with where you landed.

WouldYouKindly:
@tristanlabelle I've splitted the changes into backend/frontend parts. This PR now contains only the backend changes. There are some weird hydration issues with FE, so I'm hoping we can at least merge the backend...

Good idea about the CS redesign - I'll see if I can test it.

nabramow:
@WouldYouKindly I fear the CS update might have thrown a wrench in this. Should we close or what's the status of this?

WouldYouKindly:
@nabramow I think that without references, this actually makes little sense. Probably best to mark this one as #wontfix

tristanlabelle:
It's probably the right call, but still unfortunate :(

nabramow:
> @nabramow I think that without references, this actually makes little sense. Probably best to mark this one as #wontfix

Sorry about that Alexey but it was a valiant effort!

Rohlik:
Why was this closed? It is not clear to me from the comments 🧐. I checked my export from Couchsurfing after their redesign, and the references are still present.

WouldYouKindly:
@Rohlik there's no way to import verify that the imported data have not been tampered with. We don't want to allow importing fake references.

Rohlik:
But what if the user would need to put generated URL of that export? In my case the link is:
```
https://storage.googleapis.com/production-461720-data-exports/data-exports/3EoEjGvhNHmvw7aZSHGj0H7b4it_de/couchsurfing-data-export.zip?GoogleAccessId=production-upload-service%40production-461720.iam.gserviceaccount.com&Expires=1784100751&Signature=MQCJthuMkSJpZkDw0NTc%2FkE88akEPnn6aLL6X2iwYmdZnv5YsKuj6s48p2484aitQXN2bZwdodvWXeqhkPKlhCnciARAqe2ex6avfmP4D0%2BSberUBeK9IElwg6YPeH9QA05SZQ04u8NcVjfnaU9UulsJgpsj5JxF2yZ02HQ09BP<redacted>
```
The URL contains some structure and some identifiers, like `/data-exports/` - folder (prefix) inside the bucket or `couchsurfing-data-export.zip` as name of the archive. And `production-461720` is probably a unique Google project identifier for Couchsurfing.
If we check the URL structure, we can be fairly sure it is legitimate, since we will download it directly from the Couchsurfing infrastructure.

PR #8203, opened 2026-03-30
Backend speedups

aapeliv:
Collection of backend performance optimizations and infrastructure improvements targeting slow queries identified via Jaeger traces and EXPLAIN ANALYZE on production.

### Query optimizations

- **Optimize unseen host request counts in Ping** — Rewrite from `JOIN + COUNT(DISTINCT)` to `EXISTS + COUNT(*)`. The old query forced Postgres to choose a slow GiST spatial index for user lookups (2.3ms/loop × 52 loops). The new query uses a semi join that short-circuits after finding the first unseen message. **124ms → 0.7ms (sent) / 7.3ms (received)**

- **Precompute GeoJSON in lite_users materialized view** — `GIS.GetUsers` was spending ~260ms serializing 70k geometries to GeoJSON on every request. Now the GeoJSON Feature string is precomputed per-row in the materialized view and assembled with `string_agg` at query time

- **Reduce N+1 queries in ListGroupChats** — Batch unseen message counts into a single query instead of one per chat, add eager loading for related objects

- **Add eager loading across servicers** — Eliminate N+1 query patterns in conversations, communities, and other endpoints

- **Optimize send_request_notifications query** — Rewrite from scanning all visible users with correlated EXISTS subqueries to driving the candidate user query from host_requests/messages. **~1.5s → ~230ms** (before index, faster with index)

- **Optimize moderation_state_column_visible** — Replace 4 EXISTS subqueries that each seq-scanned a content table + moderation_states (totaling ~580k rows scanned) with a single direct PK lookup on moderation_states. For visible/unlisted content, no content table access is needed. For the rare shadowed case, uses object_type + object_id for a targeted PK lookup into just the relevant content table

### New indexes

- **`ix_users_visible_with_about_me`** — Partial B-tree index on `users(id)` filtering by visibility + `profile_gallery_id IS NOT NULL` + `about_me >= 150 chars`. Speeds up completed profile count queries (metrics, search, jobs)

- **`ix_moderation_states_type_visibility`** — Composite index on `moderation_states(object_type, visibility)`

- **Partial index on `messages(conversation_id, id)`** for text messages — Supports the rewritten send_request_notifications query pattern

### Infrastructure

- **HTTP/2 between nginx and envoy** — Fix request queueing caused by HTTP/1.1 connection limits

- **Enable `pg_stat_statements`** — Add query performance monitoring extension across all environments. Requires postgres restart in production for `shared_preload_libraries`

## Testing

Query optimizations verified via `EXPLAIN ANALYZE` on production data. No functional changes — all queries return identical results.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/models/moderation.py
+++ b/app/backend/src/couchers/models/moderation.py
@@ -96,6 +96,8 @@ class ModerationState(Base, kw_only=True):
         Index("ix_moderation_states_object", object_type, object_id, unique=True),
         # Covering index for visibility filtering - enables index-only scans in where_moderated_content_visible
         Index("ix_moderation_states_id_visibility", id, visibility),
+        # Fast filtering by object type and visibility
+        Index("ix_moderation_states_type_visibility", object_type, visibility),
     )
 
     def __repr__(self) -> str:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,7 +119,6 @@ uv run --project .claude/tools ci-job-log <job-id> --full
 - PostgreSQL with PostGIS extension
 - Migrations in `/app/backend/src/couchers/migrations/versions/`
 - Migrations use ordinal numbering (`0001_`, `0002_`, ...) and must be linear (no branches). New migrations automatically get the next ordinal as their revision ID via `env.py`
-- When creating migrations manually, always use a real source of randomness for any hex values (e.g. `secrets.token_hex()`) and the real current time for timestamps - never fabricate or hardcode these values
 - Models in `/app/backend/src/couchers/models/`
 
 ## Pull Requests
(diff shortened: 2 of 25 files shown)


app/backend/src/couchers/models/moderation.py
```diff
@@ -96,6 +96,8 @@ class ModerationState(Base, kw_only=True):
         Index("ix_moderation_states_object", object_type, object_id, unique=True),
         # Covering index for visibility filtering - enables index-only scans in where_moderated_content_visible
         Index("ix_moderation_states_id_visibility", id, visibility),
+        # Fast filtering by object type and visibility
+        Index("ix_moderation_states_type_visibility", object_type, visibility),
```
WouldYouKindly, line 100: object_type has very low cardinality - does it still have much of an effect?

WouldYouKindly (approved):

app/backend/src/couchers/models/moderation.py
aapeliv: Yes, this is used in the metrics collection. It’s a bit of a silly index, but it speeds it up significantly

PR #8210, opened 2026-03-31
Fix List-Unsubscribe crash for chat__missed_messages

aapeliv:
Fix crash when a user clicks the List-Unsubscribe link in a `chat__missed_messages` email notification.

`can_unsubscribe_topic_key` matched the entire `"chat"` topic, so `chat__missed_messages` (a summary notification with `key=""`) got a topic_key unsubscribe link. When clicked, `respond_quick_link` tried `int("")` and crashed with `ValueError`.

The fix narrows `can_unsubscribe_topic_key` to match `chat__message` specifically (which has a chat ID as its key), so `chat__missed_messages` correctly falls back to a topic_action unsubscribe link.

Sentry: https://couchers.sentry.io/issues/7335038975/events/latest/

## Testing

- Added a regression test (`test_chat_missed_messages_list_unsubscribe_header`) that creates the notification, extracts the List-Unsubscribe header, asserts it's a topic_action link (not topic_key), and calls the Unsubscribe endpoint
- Verified the test fails with the old logic and passes with the fix

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/quick_links.py
+++ b/app/backend/src/couchers/notifications/quick_links.py
@@ -87,15 +87,13 @@ def generate_quick_decline_link(host_request: requests_pb2.HostRequest) -> str:
     )
 
 
-def can_unsubscribe_topic_key(topic: str | NotificationTopicAction) -> bool:
+def can_unsubscribe_topic_key(topic_action: NotificationTopicAction) -> bool:
     """
-    Determines whether a user can unsubscribe from all notification actions
-    concerning a given topic.
+    Determines whether a user can unsubscribe from a specific topic key
+    (e.g. muting a specific chat).
     """
-    if isinstance(topic, NotificationTopicAction):
-        topic = topic.topic
-    # We currently only support unsubscribing from chat topics
-    return topic == NotificationTopicAction.chat__message.topic
+    # Only chat__message has a meaningful key (the chat ID); chat__missed_messages is a summary with no specific chat
+    return topic_action == NotificationTopicAction.chat__message
 
 
 def respond_quick_link(request: auth_pb2.UnsubscribeReq, context: CouchersContext, session: Session) -> str:
--- a/app/backend/src/tests/test_email.py
+++ b/app/backend/src/tests/test_email.py
@@ -1,5 +1,6 @@
 from datetime import timedelta
 from unittest.mock import patch
+from urllib.parse import parse_qs, urlparse
 
 import pytest
 from sqlalchemy import func, select, update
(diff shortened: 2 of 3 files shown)


app/nginx/sites-enabled/api.conf
tristanlabelle, line 1: The 3 nginx files are unrelated

app/backend/src/couchers/notifications/quick_links.py
tristanlabelle, line 56: It would be more robust to make this throw if `notification.key` is empty, and maybe even if `can_unsubscribe_topic_key` returns false, so we catch the issue upstream.

app/backend/src/tests/test_email.py
```diff
@@ -421,6 +428,54 @@ def test_send_donation_email(db, monkeypatch):
...
+                ),
+            )
+
+    assert mock.call_count == 1
+    e = email_fields(mock)
+
+    assert e.list_unsubscribe_header
+
+    # Extract the List-Unsubscribe URL and call the Unsubscribe endpoint.
+    # Before the fix, this crashed with: ValueError: invalid literal for int() with base 10: ''
```
tristanlabelle, line 464: Is it useful to include how it crashed? The way this would crash could change in the future. The point is that the unsubscribe request was invalid or couldn't be executed.

app/backend/src/tests/test_email.py
```diff
@@ -421,6 +428,54 @@ def test_send_donation_email(db, monkeypatch):
         assert email.source_data == "testing_version/donation_received"
 
 
+def test_chat_missed_messages_list_unsubscribe_header(db):
+    """
+    Regression test: chat__missed_messages has key="" (it's a summary, not tied to a single chat).
+    The List-Unsubscribe header must not generate a topic_key unsubscribe link (which would try
+    int("") and crash), but should fall back to a topic_action unsubscribe link.
```
tristanlabelle, line 435: Minor:
1. It's not super useful to reference `int("")`, we don't know where that code is or whether it would actually crash on an `int()` conversion vs something else.
2. The topic_action unsubscribe link is not a fallback, it's included regardless of whether there is a topic_key unsubscribe link or not

tristanlabelle (commented):
Thanks for the fix. It seems like you picked up unrelated nginx changes, so holding off my sign off. I also suggest catching the error more upstream.

tristanlabelle (approved):
Previous comments still apply.

aapeliv:
> Previous comments still apply.

Sorry, I didn't have time to clean up those comments; but I'll merge this to get the bug fixed.

aapeliv:
@claude address the remaining review comments in this PR

PR #8254, opened 2026-04-05
Fix MyPostcard API crash on empty order date range

aapeliv:
The MyPostcard API returns HTTP 500 with an empty body when there are no orders in the requested date range, instead of returning an empty list. This caused `get_order_ids()` to raise an exception on `response.raise_for_status()`.

This adds a workaround to return an empty list when we receive a 500 with no response body, with a TODO to remove once the upstream API is fixed.

## Testing
- Verified that the fix correctly handles 500 responses with empty body by returning `[]`
- Existing behavior for successful responses and other error codes is unchanged

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/postal/my_postcard.py
+++ b/app/backend/src/couchers/postal/my_postcard.py
@@ -185,6 +185,10 @@ def get_order_ids(date_from: date, date_to: date) -> list[int]:
         },
         timeout=30,
     )
+    # todo: remove this fix when upstream fixes API
+    # MyPostcard API returns 500 with empty body when there are no orders in the date range
+    if response.status_code == 500 and not response.content:
+        return []
     response.raise_for_status()
     return [int(order["job_id"]) for order in response.json()["orders"]]
 


app/backend/src/couchers/postal/my_postcard.py
```diff
@@ -185,6 +185,10 @@ def get_order_ids(date_from: date, date_to: date) -> list[int]:
         },
         timeout=30,
     )
+    # todo: remove this fix when upstream fixes API
+    # MyPostcard API returns 500 with empty body when there are no orders in the date range
+    if response.status_code == 500 and not response.content:
+        return []
```
nabramow, line 191: When you say upstream do you mean the 3rd party API has an issue? If so might be good to post the relevant github issue in a comment or something so we can follow up on it in the future?

WouldYouKindly:
@aapeliv could you also disable the job when run locally please? Dev server log now shows a 401 on startup.

aapeliv:
they actually fixed this

aapeliv:
> @aapeliv could you also disable the job when run locally please? Dev server log now shows a 401 on startup.

should be fixed here: https://github.com/Couchers-org/couchers/pull/8246

PR #8279, opened 2026-04-09
Update ops.md with new deployment instructions

nabramow:
I was trying to deploy but I forgot the link and couldn't find it anywhere in our docs. I also see we still have the old instructions which take the site down when we run them.

So updating this here. @aapeliv could you have a look and take out anything else that's no longer relevant? Maybe also add something about the reboot instance?


docs/ops.md
```diff
@@ -52,15 +52,18 @@ ssh couchers2
...
-cd app/deployment
-./deploy.sh
-```
+To deploy:
+
+- Click around `https://next.couchershq.org/` to make sure nothing obvious was broken by recent merges, especially map search.
+- Go to `https://ops.couchershq.org/`
+- Find the "Deploy" button on the latest commit and click it.
+- Refresh the page, the status at the top should update
+- It will take 20ish minutes
```
aapeliv: It takes about a minute

docs/ops.md
```diff
@@ -52,15 +52,18 @@ ssh couchers2
...
+- Once it's done, click around prod, make sure everything is working.
 
+It's good to only deploy if you have a bit of time or know someone else is online and free in case something breaks.
```
aapeliv: It's essential to be able to be around for another 30 minutes if something breaks.

For deployements that don't update the migration, the immediate rollback plan is to go back to the last good known version. For deployments that do update the migration, it should only be done when someone knowledgeable is around and able to unbrick if something happens.

aapeliv (changes requested):

docs/ops.md
nabramow: @aapeliv Updated!

aapeliv (approved):

PR #8284, opened 2026-04-11
Sort admin GetChats by latest message time

aapeliv:
Sort both host requests and group chats in the admin `GetChats` API by the timestamp of their last message (most recent first), so admins see the most recently active conversations at the top.

Previously host requests were sorted by conversation ID and group chats by join date, which didn't reflect actual activity.

## Testing
- Verified formatting passes (`make format`)
- Verified type checking passes (`make mypy`)
- Existing `test_GetChats` test still passes

**Backend checklist**

## For maintainers


jesseallhands (approved):
Yesssssss! Lgtm

PR #8301, opened 2026-04-12
Update zh_Hans translation

HankAviator:

This commit translates untranslated strings and move lines to align with en.json key order.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


tristanlabelle (dismissed):

HankAviator:
The Auto-format job seems to be failing before checkout with:

`[@octokit/auth-app] appId option is required`

It looks like `.github/workflows/auto-format.yml` expects `vars.APP_ID`, but it is empty/unavailable for this PR run. Since this PR only changes locale JSON, maybe it needs a repo/workflow configuration fix rather than a change in the translation.

tristanlabelle:
@HankAviator , this is definitely an infra issue. I'll ping relevant devs again.

tristanlabelle:
Oh @HankAviator , I'm almost certain this is because your PR is from your fork, so for safety reasons GitHub won't give access to vars/secrets to workflows triggered that way. I'll see if we can give you access, and otherwise we can probably tweak the autoformat to not need the secret if the formatting is no-op and no commit is needed, which would be a workaround.

HankAviator:
Thanks @tristanlabelle  most likely i will only work on the i18n

HankAviator:
Superseded by #8378

PR #8308, opened 2026-04-13
Serve .well-known app associations dynamically per env

aapeliv:
I think it's better to have only one app ID in these manifest files, that way the app/websites don't get confused. In particular with both listed, I think one will take priority, and that will mess up the other one (e.g. prod might take priority, and then it's hard to test stuff on stage; or vice versa which is worse).

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


nabramow:
My understanding is that the filename can only be "apple-app-site-association" or it won't be recognized by Apple and Google Play. I could be wrong though did you see this in the docs somewhere? From their docs here it seems very specific:

https://developer.apple.com/documentation/xcode/supporting-associated-domains

So I'm wondering if there's another way to fix this?

aapeliv:
> My understanding is that the filename can only be "apple-app-site-association" or it won't be recognized by Apple and Google Play. I could be wrong though did you see this in the docs somewhere? From their docs here it seems very specific:
> 
> https://developer.apple.com/documentation/xcode/supporting-associated-domains
> 
> So I'm wondering if there's another way to fix this?

It copies them into the canonical location at build time.

nabramow:
> It copies them into the canonical location at build time.

Ah okay sorry I didn't understand at first what was happening with the Dockerfile.

Would it be possible to switch it over so the prod one has the default name and the staging one gets copied over with the "apple-app-site-association-staging" via the staging Dockerfile instead?

Just a nit but I try to keep the pattern in the app that prod is the default name and staging is the special name.

aapeliv:
> > It copies them into the canonical location at build time.
> 
> Ah okay sorry I didn't understand at first what was happening with the Dockerfile.
> 
> Would it be possible to switch it over so the prod one has the default name and the staging one gets copied over with the "apple-app-site-association-staging" via the staging Dockerfile instead?
> 
> Just a nit but I try to keep the pattern in the app that prod is the default name and staging is the special name.

I think in this case it's safer to keep prod as the override, as otherwise it'll end up in all sorts of places!

nabramow (approved):
LGTM

PR #8309, opened 2026-04-13
Frontend/i18n: Localize date/time picker formats

### LINKED ISSUE #8020: Frontend/i18n: Date/time pickers use US format
Image

tristanlabelle:
Date/time pickers were using the browser-preferred language, not the user-selected one. We can use `Intl.DateTimeFormat` to sniff the format and construct a Material UI format string that matches the locale.

Fixes #8020

## Testing
Added tests and ran locally.

image

24h locales use the corresponding clock picker:

image

**Web frontend checklist**

## For maintainers


nabramow:
Looks like you got some timepicker tests failing @tristanlabelle !

tristanlabelle:
@nabramow I fixed the tests. Please have a look at the `act` that was necessary -- this stuff is new to me!

app/web/components/Datepicker.test.tsx
```diff
@@ -149,9 +150,8 @@ describe("DatePicker", () => {
     expect(date?.format("YYYY-MM-DD")).toBe("2021-03-20");
   });
 
-  it("typing should work in en-GB", async () => {
-    const langMock = jest.spyOn(navigator, "language", "get");
-    langMock.mockReturnValue("en-GB");
+  it("typing should work in de's DD.MM.YYYY format", async () => {
+    i18n.changeLanguage("de");
```
nabramow, line 154: Can we keep the test description string the same here? Since it is actually based on locale that makes more sense IMO.

app/web/components/Datepicker.test.tsx
```diff
@@ -45,9 +45,10 @@ describe("DatePicker", () => {
     jest.setSystemTime(new Date("2021-03-20 00:00"));
   });
 
-  afterEach(() => {
+  afterEach(async () => {
     jest.resetAllMocks();
     jest.clearAllTimers();
+    await act(() => i18n.changeLanguage("en"));
```
nabramow, line 51: Does this definitely need to be async?

This doesn't work?:

```suggestion
  afterEach(() => {
jest.resetAllMocks();
jest.clearAllTimers();
   act(() => i18n.changeLanguage("en"));
```

app/web/components/Datepicker.test.tsx
tristanlabelle: Could you provide that description string? I don't understand what you have in mind.

1. I changed `en-GB` to `de` because `en-GB` is not one of our supported locales
2. I added `DD.MM.YYYY` to the description because the useful part of the test is that the locale has this format which is different from US or other formats.

app/web/components/Datepicker.test.tsx
tristanlabelle: Since https://react.dev/reference/react/act is defined as an async function, calling it without awaiting means that it might not have proceeded to completion by the time `act()` returns.

app/web/components/Datepicker.test.tsx
nabramow: Ah just the old one, fine to change it to just "en" if that's better. So for this one "typing should work in en".

Or if you want "typing should work in en DD.MM.YYYY format" if you prefer.

tristanlabelle: I'm still confused? "en" format is not "DD.MM.YYYY", that's de's. That's why I have both tests:

typing should work in de's DD.MM.YYYY format
typing should work in en's MM/DD/YYYY format

tristanlabelle:
@nabramow My latest commit fixed the clock picker am/pm issue you pointed out in https://github.com/Couchers-org/couchers/issues/8020#issuecomment-4304347780

app/web/components/Datepicker.test.tsx
nabramow: Sorry that was a typo I meant broadly you can put both in the strong but leave the locale.

tristanlabelle: I don't mind whatever change but I still don't get your meaning 😩. "put both in the strong but leave the locale.". What is the strong? Could you add another comment with suggestion snippet to get us unblocked?

nabramow: The most recent change looks good to me now!

nabramow (approved):
Okay both date and time picker are working great for me now!

PR #8316, opened 2026-04-14
Add missing public trips and requests backend

nabramow:
A couple things I realized we need when doing the frontend:

1. public trip protos
2. Switch to `User` from `LiteUser` so I can return gender, numReferences to frontend
3. Prep for the reverse host request after discussion about how it should go visually

See https://github.com/Couchers-org/couchers/blob/develop/docs/public_trips.md for implementation plan.

## Testing

**Backend checklist**

## For maintainers


app/proto/public_trips.proto
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  rpc UpdatePublicTripStatus(UpdatePublicTripStatusReq) returns (google.protobuf.Empty) {
+    // Update trip status (e.g. mark as found host or deactivate)
+  }
+}
+
+enum PublicTripStatus {
+  PUBLIC_TRIP_STATUS_UNSPECIFIED = 0;
+  PUBLIC_TRIP_STATUS_SEARCHING_FOR_HOST = 1;
```
tristanlabelle, line 37: Maybe a more general "open"? A public trip might not be because a user is searching for a host, or they might have found hosts for part of the trip.

tristanlabelle (dismissed):

PR #8336, opened 2026-04-16
Backend: Add missing public_trips servicers

nabramow:
Adding more stuff I forgot to get the backend api working for the frontend.

## Testing

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -0,0 +1,217 @@
...
+            .limit(page_size + 1)
+        )
+        public_trips = session.execute(statement).scalars().all()
+
+        return public_trips_pb2.ListPublicTripsRes(
+            public_trips=[public_trip_to_pb(trip, session, context) for trip in public_trips[:page_size]],
+            next_page_token=str(public_trips[-1].id) if len(public_trips) > page_size else None,
+        )
+
+    def ListMyPublicTrips(
```
tristanlabelle: Should we instead make this function "list a user's public trips", taking a user token? That would allow us to have a list of upcoming trip on people's profiles, like CS.

(feel free to address separately)

app/backend/src/tests/test_public_trips.py
```diff
@@ -0,0 +1,446 @@
...
+            api.UpdatePublicTripStatus(
+                public_trips_pb2.UpdatePublicTripStatusReq(
+                    trip_id=trip_id,
+                    status=public_trips_pb2.PUBLIC_TRIP_STATUS_CLOSED,
+                )
+            )
+        assert e.value.code() == grpc.StatusCode.NOT_FOUND
+
+
+def test_update_public_trip_status_reject_searching(db):
```
tristanlabelle:
```suggestion
def test_update_public_trip_cant_reopen(db):
```

tristanlabelle (dismissed):
Looks good!

app/backend/src/couchers/servicers/public_trips.py
nabramow: Yeah that's a good idea! Just adjusted it as it's a pretty easy win IMO even if we don't use it right away.

PR #8338, opened 2026-04-17
Backend/i18n: Support localizing unsubscribe link descriptions

tristanlabelle:
Emails have unsubscribe links for example "unsubscribe from: new messages". I'm extracting these strings for localization so that they can be reused by the new email system which supports localization that I'm putting in place. After this change, both the new and the old system can be abstracted as ways to produce a (title, plaintext body, html body) tuple from a notification, which will simplify migration.

## Testing
Ran tests and added some.

**Backend checklist**

## For maintainers


app/backend/src/couchers/notifications/locales/en.json
```diff
@@ -84,11 +86,14 @@
     }
   },
   "chat": {
+    "unsubscribe_from_description": "this chat (mute)",
     "message": {
-      "event_description": "Someone sends you a message"
+      "event_description": "Someone sends you a message",
+      "unsubscribe_from_description": "new chat messages"
```
aapeliv, line 92: IIUC you are reusing the same `unsubscribe_from_description` name for both types of unsub. I would differentiate for clarity (for both translators + devs). The two different types can already be a bit complicated.

app/backend/src/couchers/notifications/quick_links.py
```diff
@@ -92,15 +91,6 @@ def generate_quick_decline_link(host_request: requests_pb2.HostRequest) -> str:
     )
 
 
-def can_unsubscribe_topic_key(topic_action: NotificationTopicAction) -> bool:
-    """
-    Determines whether a user can unsubscribe from a specific topic key
-    (e.g. muting a specific chat).
-    """
-    # Only chat__message has a meaningful key (the chat ID); chat__missed_messages is a summary with no specific chat
-    return topic_action == NotificationTopicAction.chat__message
```
aapeliv, line 101: I think last time we discussed keeping this here to colocate it next to where the unsub links are actually handled.

aapeliv (changes requested):
two minor comments, otherwise lgtm

aapeliv:
oh, and the other thing on my mind that i thought i already commented: I don't know if the current setup in the footer is the best way to go. if you would like, you could come up with a better way to describe/have the unsub links in the footer before we started translating them!

app/backend/src/couchers/notifications/quick_links.py
tristanlabelle: Yes. It's now grown to also have `get_topic_action_unsubscribe_text` and `get_topic_key_unsubscribe_text` so I put them together. Do you prefer them all in `quick_links.py`?

tristanlabelle:
Abandoning in favor of https://github.com/Couchers-org/couchers/pull/8415, which sidesteps any unsubscribe from redesign so I can focus on getting emails to be localizable.

PR #8343, opened 2026-04-18
Add a make command to auth-generate migrations

WouldYouKindly:
Adds `make revision` command that doesn't require a locally running database. The command starts a fresh docker container from our Postgis image, applies existing migrations, then runs alembic auto-generation command. At the end, the Postgis container is cleaned up. This way it doesn't interfere with one's local database.

Also adds `ruff` rule to disallow local imports - something claude is guilty of a lot...

### DIFF
--- a/app/backend/Makefile
+++ b/app/backend/Makefile
@@ -31,6 +31,13 @@ fmt: format
 mypy:
 	uv run mypy src
 
+# Generate a new database migration by comparing models to an empty database
+# Usage:
+# make revision                     → generates migration with default message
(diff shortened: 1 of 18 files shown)


aapeliv (approved):
lgtm. i would probably just add all the deps in dev dependencies and not have the separate `uv run --with` busienss

WouldYouKindly:
> lgtm. i would probably just add all the deps in dev dependencies and not have the separate `uv run --with` busienss

TBH I don't want to include docker into dependencies, since we ship dev dependencies to prod as well.

PR #8344, opened 2026-04-18
Backend/editor: allow reinstating former volunteers

aapeliv:
Adds a `reinstate_volunteer` flag to the `UpdateVolunteer` editor RPC. The existing `UpdateVolunteer` uses `HasField()`-style partial updates on wrapped fields, which lets you set or change `stopped_volunteering`, but gives no way to clear it back to NULL. This flag explicitly clears `stopped_volunteering`, making a former volunteer current again. Setting both `reinstate_volunteer=true` and `stopped_volunteering` in the same request is rejected as `INVALID_ARGUMENT`.

## Testing
- New tests in `test_editor.py`:
  - `test_UpdateVolunteer_reinstate` — reinstate a former volunteer, confirm `stopped_volunteering` is cleared in the DB and the returned proto.
  - `test_UpdateVolunteer_reinstate_already_current` — no-op on a volunteer who is already current.
  - `test_UpdateVolunteer_reinstate_conflict_with_stopped` — conflict with `stopped_volunteering` returns `INVALID_ARGUMENT`.
- `uv run pytest src/tests/test_editor.py` — all 27 tests pass.
- `make format` and `make mypy` clean.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/i18n/locales/en.json
+++ b/app/backend/src/couchers/i18n/locales/en.json
@@ -20,6 +20,7 @@
     "cannot_create_discussion": "You cannot create a discussion in this group or community.",
     "cannot_edit_that_notification_preference": "That notification preference is not user editable.",
     "cannot_leave_containing_community": "Your location on your profile is within this community, so you cannot leave it. However, you can adjust your notifications in your account settings.",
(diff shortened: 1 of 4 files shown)


jesseallhands (approved):
Lgtm

PR #8345, opened 2026-04-18
Consolidate community discussions/events flags into one

aapeliv:
Collapses `discussions_enabled` and `events_enabled` on `Cluster` into a single `small_community_features_enabled` flag, and updates the frontend to match. The new flag is also the intended gate for public trips (being added in a parallel PR).

Prod data confirmed the two flags were always toggled together — all 354 region/subregion/locality communities have them either both-on or both-off, so there's no data migration concern: the OR of the two old columns equals the new column's value everywhere.

The proto field number (14) is reused for the new name; `events_enabled` (15) is marked `deprecated` and still populated with the same value for backwards compatibility with any older clients.

Not a one-way door — if we ever want independent toggles again we can re-split.

## Testing
- `make format` and `make mypy` pass
- `pytest src/tests/test_communities.py src/tests/test_discussions.py src/tests/test_events.py` — 65/65 pass

**Backend checklist**
- [x] Added tests for any new code or added a regression test if fixing a bug (existing tests cover the gating behavior; field rename doesn't need new tests)
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (migration will be CI-autogenerated)

**Web frontend checklist**

## For maintainers


nabramow:
@aapeliv There's already some `public_trips` backend logic for the region level. Do you want to change that too? 

For example check `servicers/public_trips.py`, line 71 and its related test.

nabramow (dismissed):
This looks good to me now! Thanks for adding that change!

PR #8354, opened 2026-04-19
Backend: Exclude UMS-hidden notifications from unseen count

aapeliv:
The unseen notification count returned by `Ping` (used to render the notification badge) did not exclude notifications hidden by the user moderation system. As a result, users could see a non-zero badge but find no corresponding notification in `ListNotifications`, which already filters by `moderation_state_column_visible`.

Apply the same `moderation_state_column_visible(context, Notification.moderation_state_id)` filter to the `unseen_notification_count` query so the two endpoints stay in sync.

## Testing
- `make format` passes
- Existing mypy errors on `develop` are unrelated to this change

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/api.py
+++ b/app/backend/src/couchers/servicers/api.py
@@ -53,7 +53,13 @@ from couchers.rate_limits.check import process_rate_limits_and_check_abort
 from couchers.rate_limits.definitions import RATE_LIMIT_HOURS
 from couchers.resources import get_badge_dict, language_is_allowed, region_is_allowed
 from couchers.servicers.blocking import is_not_visible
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/servicers/api.py
```diff
@@ -260,6 +266,7 @@ def Ping(self, request: api_pb2.PingReq, context: CouchersContext, session: Sess
                     get_topic_actions_by_delivery_type(session, user.id, NotificationDeliveryType.push)
                 )
             )
+            .where(moderation_state_column_visible(context, Notification.moderation_state_id))
```
WouldYouKindly, line 269: Would be lovely to have a test for this

WouldYouKindly (dismissed):

WouldYouKindly (approved):

PR #8355, opened 2026-04-19
Migrate to psycopg 3

WouldYouKindly:
psycopg2 is in maintenance mode, and psycopg3 is currently recommended. It also should be slightly faster due to prepared statements, and using libpq instead of sending strings under the hood.

I tested search APIs, seem to work fine.


aapeliv (approved):
lgtm. one question: is there a new/different URI "protocol" for v3? i feel like always registering `postgresql` for v3 is a tad hacky?

WouldYouKindly:
> lgtm. one question: is there a new/different URI "protocol" for v3? i feel like always registering `postgresql` for v3 is a tad hacky?

@aapeliv we could change the connection string everywhere to `postgresql+psycopg`, but I thought it was a bit annoying. Do you think it's a better approach that registering an adapter? Happy to change if so

aapeliv:
> > lgtm. one question: is there a new/different URI "protocol" for v3? i feel like always registering `postgresql` for v3 is a tad hacky?
> 
> @aapeliv we could change the connection string everywhere to `postgresql+psycopg`, but I thought it was a bit annoying. Do you think it's a better approach that registering an adapter? Happy to change if so

I'll leave it to your judgement as to how hacky the adapter is. Happy with either one.

PR #8364, opened 2026-04-19
Backend: Use node's timezone for public trip date validation

aapeliv:
Public trip date validation (`CreatePublicTrip` / `UpdatePublicTrip`) previously used the **creator's** home timezone to decide what "today" meant. This is wrong — a user in NYC can create a public trip in Helsinki, and the correctness of the past-date check should depend on where the trip is, not where the user lives. This also caused a flaky test (`test_update_public_trip_in_past`) that failed during the ~00:00–05:00 UTC window because UTC `today() - 1` could equal NYC's "today".

This PR adds a `timezone` column_property on `Node` (mirroring the pattern on `User`, but wrapping the MULTIPOLYGON geom in `ST_PointOnSurface` so we always get an interior point) and switches the `CreatePublicTrip`/`UpdatePublicTrip` past-date checks to use the node's timezone. `ListPublicTrips*` still uses UTC `today()` — it spans many nodes and ~1 day of slack there is fine.

No migration — the column is computed at query time and `deferred=True` so default `SELECT Node` queries pay zero cost.

## Testing
- Updated `_make_node` test helper to use a Helsinki polygon so `node.timezone` resolves to something non-UTC in tests.
- Added `test_create_public_trip_in_past_uses_node_timezone` — user in NYC (default), node in Helsinki, past `from_date` → rejected with `INVALID_ARGUMENT`.
- Padded `test_update_public_trip_in_past` from `today() - 1` to `today() - 2` to be robust against any UTC/local boundary.
- All 31 tests in `test_public_trips.py` pass; `make format` and `make mypy` clean.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/materialized_views.py
+++ b/app/backend/src/couchers/materialized_views.py
@@ -15,9 +15,9 @@ from sqlalchemy.sql import (
     func,
     literal,
     literal_column,
+    select,
     union_all,
 )
-from sqlalchemy.sql import select as sa_select
 from sqlalchemy.sql.functions import percentile_disc
 from sqlalchemy_utils.view import (
     CreateView,
@@ -78,7 +78,7 @@ def create_materialized_view_with_different_ddl(
 
 
 cluster_subscription_counts_selectable = (
-    sa_select(
+    select(
         ClusterSubscription.cluster_id.label("cluster_id"),
         func.count().label("count"),
     )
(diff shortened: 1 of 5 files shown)


nabramow (approved):
Ah I was wondering about the timezone stuff actually as we have a few "today" utility functions on the backend. Thanks!

PR #8368, opened 2026-04-19
Don't review the same migrations over and over again

WouldYouKindly:
Calculates hash of migrations/ directory and saves it to actions cache. On subsequent runs, skips review if hash hasn't changed.


aapeliv (commented):
lgtm but please remove the dummy migration

aapeliv (approved):
lgtm

.github/workflows/migration-review.yml
```diff
@@ -252,3 +268,14 @@ jobs:
           claude_args: |
             --model opus
             --allowedTools "Read,Glob,Grep,Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr comment:*),Bash(gh api:*),mcp__github_inline_comment__create_inline_comment"
+
+      - name: Mark as reviewed
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: mkdir -p .migration-review-marker && touch .migration-review-marker/done
+
+      - name: Save review marker
```
tristanlabelle, line 276: FYI: You can also use `actions/cache` to do both the restore and save actions at once. I think it basically registers the save as a workflow cleanup job.

PR #8371, opened 2026-04-19
Ban 'bot' and 'verify' as usernames

aapeliv:
Adds `bot` and `verify` to the `BANNED_USERNAME_PHRASES` list so these phrases are no longer allowed in usernames.

## Testing
Covered by the existing banned-username tests — the new entries go through the same validation path as the rest of the list.

**Backend checklist**

## For maintainers


app/backend/src/couchers/constants.py
```diff
@@ -12,6 +12,7 @@
 
 BANNED_USERNAME_PHRASES = [
     "admin",
+    "bot",
```
jesseallhands, line 15:
```suggestion
    "bot",
    "safety",
    "security",
    "secure",
    "trust",
```

jesseallhands (changes requested):
I added some phrases. Maybe we should remove "bot" as that could be part of some legitimate names.

jesseallhands (approved):

PR #8372, opened 2026-04-19
Fix PR skill branch naming guidance

aapeliv:
Drops the username fallback from the PR skill's branch-naming rule and expands the allowed area list (adds `devops` and `docs`, and explicitly permits other areas that match the changed files). The previous wording caused the skill to pick `aapeli/...` for a backend-only change instead of `backend/...`.

## Testing
Re-ran the PR skill on this very change — it picked `docs/fix/...` as expected.

## For maintainers


jesseallhands (approved):
Lgtm

PR #8393, opened 2026-04-24
Backend: Allow reopening closed trips and provide community slug in response

nabramow:
There are two things I'm aiming to implement here:

1. Allow reopening a closed trip as long as it's not in the past.

2. When returning the list of public trips, I want it to include the information needed to create a route to that community's page so I can link it on the frontend without additional API calls for each card. 

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -3,7 +3,7 @@
 
 import grpc
 from sqlalchemy import or_, select
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, selectinload
```
nabramow, line 6: I'm mostly curious if `selectinload` is the right thing to use here as I'm still not 100% on python syntax. My understanding is it uses a `SELECT IN` query instead of a join table which should be more performant or at least return a smaller data set.

app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -199,6 +201,7 @@ def ListPublicTripsByUser(
             statement.where(or_(PublicTrip.id <= next_page_id, to_bool(next_page_id == 0)))
             .order_by(PublicTrip.id.desc())
             .limit(page_size + 1)
+            .options(selectinload(PublicTrip.node).selectinload(Node.official_cluster))
```
aapeliv: could you comma-separate these please?

app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -270,8 +273,12 @@ def UpdatePublicTrip(
 
         if request.HasField("status"):
             new_status = publictripstatus2sql.get(request.status)
-            # Only closing is permitted (can't reopen a closed trip).
-            if new_status != PublicTripStatus.closed:
+            if new_status == PublicTripStatus.searching_for_host:
+                # Reopening is only allowed if the trip is not in the past.
+                today_local = today_in_timezone(public_trip.node.timezone)
+                if public_trip.to_date < today_local:
```
aapeliv: do you want it to be strictly a day earlier? so e.g. it has to close tomorrow? let's use the same logic as in creation.

aapeliv (changes requested):

app/backend/src/couchers/servicers/public_trips.py
nabramow: Done!

app/backend/src/couchers/servicers/public_trips.py
nabramow: Done, if I understood correctly.

PR #8398, opened 2026-04-24
Backend: enforce unique active friend relationship per user pair

aapeliv:
Fixes a bug where two `friend_relationship` rows could coexist in `pending` or `accepted` status between the same pair of users (a race in `SendFriendRequest`, flagged by a pre-existing `TODO` in the model).

Adds a partial unique index on `friend_relationships` over `(LEAST(from_user_id, to_user_id), GREATEST(from_user_id, to_user_id))` limited to `status IN ('pending', 'accepted')`, so the constraint is symmetric across direction and doesn't block historical `rejected`/`cancelled` rows from accumulating.

The migration first cancels older duplicates per pair (keeping the latest by `id`) so the index can be created on existing data without error.

## Testing
* New test `test_friend_relationship_unique_active_pair` in `src/tests/test_model_constraints.py` verifies:
  * second active row in the same direction raises `IntegrityError`
  * second active row in the reverse direction raises `IntegrityError`
  * `cancelled` / `rejected` rows for the same pair remain allowed
  * active rows for different pairs are unaffected
* Ran `make format` and `uv run pytest src/tests/test_model_constraints.py src/tests/test_api.py src/tests/test_references.py src/tests/test_visible_users.py` — all passing.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/migrations/versions/0146_unique_active_friend_relationship.py
+++ b/app/backend/src/couchers/migrations/versions/0146_unique_active_friend_relationship.py
@@ -1,0 +1,46 @@
+"""Unique active friend relationship per user pair
+
+Revision ID: 0146
+Revises: 0145
+Create Date: 2026-04-24 00:00:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0146"
+down_revision = "0145"
+branch_labels = None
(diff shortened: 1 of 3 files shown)


aapeliv:
I'm merging this as there is a user being impacted and having a broken experience

Issue #8312, opened 2026-04-14
Read messages still unread


CouchersBot:
Subject: Read messages still unread
Description:
Just as on the web platform, read messages are still shown as unread, even if they are days old.

Results:
Read messages don't appear as unread.

Backend version: develop-a68026c5
Frontend version: develop-a68026c5
User Agent: Mozilla/5.0 (Linux; Android 11; SM-A415F Build/RP1A.200720.012; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/147.0.7727.55 Mobile Safari/537.36 CouchersNative/1.1.16 (android; build unknown; 3024cebc)
Screen resolution: 411x783
Page: https://couchers.org/messages
User: [@koliver](https://couchers.org/user/koliver) (7723)

PR #8400, opened 2026-04-25
feat: Unread messages filter

### LINKED ISSUE #8300: Unread messages filter
Nicole Note: Add a pill after "All" and "Chats" that says "Unread" for this:

Image

----

Subject: Unread messages toggle
Description:
Tap messages 

Results:
Would be nice to have an unread messages toggle (yeah, it’s a Feature request)

Backend version: develop-7dd9a48b
Frontend version: develop-7dd9a48b
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 CouchersNative/1.1.16 (ios; build unknown; e81d170b)
Screen resolution: 414x765
Page: https://couchers.org/pt-BR/dashboard
User: [@fracktal](https://couchers.org/user/fracktal) (77)

darrenvong:

Adds a new unread messages filter in the messages listing page. Closes #8300

While working on this, I noticed there are subtly inconsistent logic for determining whether a chat is unread, so I've also consolidated the logic of that into a new `hasUnreadMessages` helper and ensure this is used in all places that need it (or at least what Claude can find!)

## Testing

Firstly, have two accounts ready and log into both.

### Unread chats
- Send a message from account 2 to account 1, so account 1 has an unread message in a chat
- In account 1, click on the new "Unread" tab and check that only the chat with the unread message is shown

### Unread host requests
- Send a surf request from account 2 to account 1, so account 1 has an unread "Surfing" request
- In account 1, click on the new "Unread" tab in the messages list, and check that only the unread chat and surf request are listed (unless you read the message already, in which case, only the surf request should show)

### Unread surf requests
- In the host request for account 1, send a message in there
- Go to account 2, click on the new "Unread" tab in the messages list, and check that only the unread surf request is listed

### Mark all as read
Sense check that still works as my refactor touches the logic there

https://github.com/user-attachments/assets/5d510d7c-10ab-4de6-8d29-f232bac19fbd

https://github.com/user-attachments/assets/a1bfc2c7-9f33-4be1-bdeb-55e3c914f18c

**Backend checklist**

**Web frontend checklist**

I'm seeing a WebGL warning locally... but I think it's unlikely to be related to my changes??
Screenshot 2026-04-25 at 02 37 17

## For maintainers


nabramow:
@darrenvong Should this close https://github.com/Couchers-org/couchers/issues/8312 as well do you think?

nabramow:
I am seeing a missing translation key in the browser console FYI:

i18next::translator: missingKey en messages mark_all_read_button_text_unread mark_all_read_button_text_unread

nabramow:
I realized if I respond in an old host request, it shows up in unread but it's dimmed out which was kinda confusing.

We could also make a separate ticket for this but I wonder if we should make those also bold if there's a new message there?

Screenshot 2026-04-26 at 10 44 23

nabramow:
For some reason sometimes it shows the unread with a 2 badge but then the initial pageSize is only 1, even though I am pretty sure it's set to more than that:

Screenshot 2026-04-26 at 10 59 34

When I click it it does show the full 2 unread messages host request and message.

nabramow:
This might be outside the scope but I realized the messages don't update live without me refreshing the page. I don't remember if they used to? Do we need to adjust the queryKey or something via tanstack query so it updates without refreshing?

darrenvong:
> @darrenvong Should this close #8312 as well do you think?

@nabramow Maybe... can you find a way to reliably reproduce this though? Things are switching to unread for me fine so I don't know how to reproduce this without more info, or this could be a bit of a wild goose chase...

> I realized if I respond in an old host request, it shows up in unread but it's dimmed out which was kinda confusing.
> 
> We could also make a separate ticket for this but I wonder if we should make those also bold if there's a new message there?

Do you mean we could keep it appearing under the unread filter, have it be bold but not dimmed?
Edit: I made it look more like this, lmk if that's ok or do you want the dates to be not dimmed too?

Screenshot 2026-04-27 at 02 10 51


> For some reason sometimes it shows the unread with a 2 badge but then the initial pageSize is only 1, even though I am pretty sure it's set to more than that:
> 
> When I click it it does show the full 2 unread messages host request and message.

Not sure I fully understand what you mean here... but I think the 2 comes down to the fact you have 2 unread messages in the chat itself? Claude is telling me we compute the unseen counts differently between chats and host requests so maybe the discrepencies are caused by that too? As otherwise, I can't see anything clever the frontend is trying to do here other than summing what's coming back from the API.

Separately, a bit weird though why the unread host request is not showing up (but somehow I'm not reproducing even in my recording 🤔 )

darrenvong:
> This might be outside the scope but I realized the messages don't update live without me refreshing the page. I don't remember if they used to? Do we need to adjust the queryKey or something via tanstack query so it updates without refreshing?

I think this should be fixed with my latest changes... the effects feel kinda grim though but at the same time, they are the equivalent of using `onSuccess` before that API was deprecated in TanStack Query v4 so can't think of a better way either. I've extracted it into a custom hook for better readability instead

app/web/features/messages/groupchats/GroupChatsTab.tsx
darrenvong, line 1: Great find by Claude on this being dead code and since it's only one file... feels ok to follow Scout's rule and just clean this up as part of this PR (which led to dead translation keys that can then be cleaned up) 🧹

nabramow: Oh nice if it's not being used then definitely!

app/web/routes.ts
```diff
@@ -82,11 +82,12 @@ export function routeToUser(username: string, tab?: UserTab) {
 export const messagesRoute = "/messages";
 export const messageTypeStrings = [
   "all",
+  "unread",
   "chats",
   "hosting",
   "surfing",
   "archived",
-];
+] as const;
```
nabramow, line 90: Nit - is there a better type we could use for this? Does it need a type?

nabramow (dismissed):
Just had one little nit comment about a type, but overall this is working great for me, looks good and is ready to go! Thanks for addressing those bugs as I know they were there before you!

app/web/routes.ts
darrenvong: This is meant to fix the type that was there already - without the const assertion, `MessageType` was just inferred as `string` rather than the stricter list of possible tab filters, which I reckon might have caught the invalid translation key issue I accidentally introduced you mentioned earlier. Probably need to add a helper to enforce exhaustive type checking in these cases though 🤔

darrenvong: Ok you sniped me a bit with this comment and I went ahead and cleaned things up further as the `undefined` prop type on the `MessagesHeader` was bothering me on a re-read 😂 - looks like I may need a re-approve since I pushed some new changes unless you find something else wrong with it

app/web/features/messages/requests/HostRequestListItem.tsx
```diff
@@ -204,7 +204,10 @@ export default function HostRequestListItem({
           sx={{ paddingRight: 5 }}
           disableTypography
           primary={
-            <Typography variant="h2">
+            <Typography
+              variant="h2"
+              sx={isPast && isUnread ? { color: "text.primary" } : undefined}
```
nabramow:
```suggestion
              sx={isPast && isUnread ? { color: "var(--mui-palette-text-primary)" } : undefined}
```

Sorry just noticed this now. This is so it works with dark mode which is new since you were here.

app/web/features/messages/requests/HostRequestListItem.tsx
```diff
@@ -248,7 +251,10 @@ export default function HostRequestListItem({
               </StyledDateAndBadgeContainer>
               <TextBody
                 noWrap
-                sx={{ fontWeight: isUnread ? "bold" : "normal" }}
+                sx={{
+                  fontWeight: isUnread ? "bold" : "normal",
+                  ...(isPast && isUnread ? { color: "text.primary" } : {}),
```
nabramow:
```suggestion
                  ...(isPast && isUnread ? { color: "var(--mui-palette-text-primary)" } : {}),
```

app/web/routes.ts
nabramow: Awesome - thanks for doing this!

nabramow (approved):
LGTM!

PR #8414, opened 2026-04-27
Backend: Public trips visibility & prevent dupe offers

nabramow:
- Adds backend logic to allow a user to select same gender only and only have viewers of the same gender see that public trip
- Adds a check in create host request to provide extra protection for same_gender_only 
- Adds error for case where a user offers to host the same trip more than once (the idea is also that I'll show some flag on the frontend if the user already offered to host and hide the "offer to host" button in that case for clarity

## Testing
- make upgrade - same_gender_only field should appear
- make downgrade, it should go away
- make upgrade, it should be back
- make downgrade just to clean up your branch

**Backend checklist**

## For maintainers


tristanlabelle (dismissed):
Looks good!

tristanlabelle (dismissed):

app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -39,6 +39,17 @@ def _is_description_long_enough(text: str) -> bool:
     return text_length_utf16 >= PUBLIC_TRIP_DESCRIPTION_MIN_LENGTH_UTF16
 
 
+def _same_gender_filter(context: CouchersContext) -> ColumnElement[bool]:
+    # Show the trip if same_gender_only is off, the viewer's gender matches the poster's gender,
+    # or the viewer is a moderator.
+    # Uses scalar subqueries rather than extra joins since where_users_column_visible
+    # already joins User on PublicTrip.user_id.
+    viewer_is_moderator = select(User.is_superuser).where(User.id == context.user_id).scalar_subquery()
```
aapeliv: let's use `can_moderate_node` and similar to stay consistent with communities features permission logic.

aapeliv (changes requested):

app/backend/src/couchers/servicers/public_trips.py
nabramow: I've made this change, but it looks like `can_moderate_node` doesn't include mods, only community moderators not superuser moderators. Do we need to add that somewhere then?

PR #8415, opened 2026-04-27
Backend/emails: Prepare for migration to localizable emails

tristanlabelle:
#7964 added a framework for making localizable emails, but it isn't used yet. This is the next step: refactoring the existing email-building code so that two branches can co-exist: one for old system, one for the new system, so I can migrate on a per-email basis.

I made `RenderedEmailNotification` more generic so it can be produced by both the old and the new system. The old system's logic is preserved but now produces a `CustomTemplatedEmail`. Unsubscribe text for topic actions was extracted as it'll be used by both branches. The new system isn't hooked up yet, that'll be the next PR.

## Testing
Ran backend + frontend locally and managed to receive a fine-looking email.

image

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/rendering.py
+++ b/app/backend/src/couchers/email/rendering.py
@@ -75,6 +75,7 @@ class EmailFooter:
         args["footer_timezone_name"] = self.timezone_name
         args["footer_copyright_year"] = self.copyright_year
 
+        args["footer_email_is_critical"] = self.unsubscribe_info is None
         if unsubscribe_info := self.unsubscribe_info:
             args.update(unsubscribe_info.to_template_args())
 
--- a/app/backend/src/couchers/email/rendering.py
+++ b/app/backend/src/couchers/email/rendering.py
@@ -66,8 +66,19 @@ class ActionBlock(EmailBlock):
 
 @dataclass(kw_only=True)
 class EmailFooter:
+    timezone_name: str
     copyright_year: int = now().year
     unsubscribe_info: UnsubscribeInfo | None
+
+    def to_template_args(self) -> dict[str, Any]:
+        args: dict[str, Any] = {}
+        args["footer_timezone_name"] = self.timezone_name
+        args["footer_copyright_year"] = self.copyright_year
+
+        if unsubscribe_info := self.unsubscribe_info:
+            args.update(unsubscribe_info.to_template_args())
+
+        return args
 
 
 @dataclass(kw_only=True)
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/email/rendering.py
```diff
@@ -77,6 +89,19 @@ class UnsubscribeInfo:
     topic_action_link: UnsubscribeLink
     topic_key_link: UnsubscribeLink | None = None
 
+    def to_template_args(self) -> dict[str, Any]:
+        args: dict[str, Any] = {}
+        args["footer_manage_notifications_link"] = self.manage_notifications_url
+        args["footer_do_not_email_link"] = self.do_not_email_url
+        args["footer_notification_topic_action"] = self.topic_action_link.text
+        args["footer_notification_topic_action_link"] = self.topic_action_link.url
```
aapeliv, line 97: Could just create a dict here directly?

app/backend/src/couchers/notifications/render_email.py
```diff
@@ -758,6 +776,113 @@ def get_list_unsubscribe_header(notification: Notification) -> str | None:
...
+
+
+def get_topic_key_unsubscribe_text(topic_action: NotificationTopicAction) -> str:
+    assert can_unsubscribe_topic_key(topic_action)
+    # Not localized because the design will change so avoid useless work by translators.
+    match topic_action:
+        case NotificationTopicAction.chat__message:
+            return "this chat (mute)"
+        case _:
+            raise AssertionError(f"No topic-key description for {topic_action}")
```
aapeliv, line 860: NotImplementedError? I don't think user code should use AssertionError

aapeliv (approved):
i had a moderate look. i assume you tested etc. few small comments but can also do them in a follow up pr if you want to merge this asap

app/backend/src/couchers/email/rendering.py
tristanlabelle: Right, the original code shape was appending to a dictionary received as an argument. Will follow-up.

app/backend/src/couchers/notifications/render_email.py
tristanlabelle: I think the linter made me use an AssertionError instead of `assert False, "blah"`, but here replacing the assert with ValueError and the AssertionError with NotImplementError would make more sense. I'll follow up.

Issue #8171, opened 2026-03-25
Mobile/i18n: The Google Play store page is not localized


tristanlabelle:
See https://play.google.com/store/apps/details?id=org.couchers.android

HankAviator:
Play store in Chinese does have the introduction translated

tristanlabelle:
@HankAviator , is there anything that specifies that it is machine translated? If you didn't write that translation, I doubt we have!

HankAviator:
@tristanlabelle no it doesn't say about machine translation but it seems so

PR #8423, opened 2026-04-29
Android: Support Play Store listing localization

tristanlabelle:
Adds locales json files for app store listings and a script to push the current values to the Google Play store.

Related to #8171

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


nabramow:
Hey @tristanlabelle just before you get too far. The mobile app is in Javascript/Typescript and I see Python here. If you don't know JS I can try and help out with this but we shouldn't put random python in frontend/mobile stuff.

Also I see you added uv and everything to the mobile app, please make sure to remove that we use npm there.

nabramow:
Also just FYI we do already have some locales on the mobile app here:

app/mobile/i18n/locales

I'm not sure if it's possible to keep it in the 18n folder or if the folder path needs to specifically be as you have it.

tristanlabelle:
@nabramow 

> Hey @tristanlabelle just before you get too far. The mobile app is in Javascript/Typescript and I see Python here. If you don't know JS I can try and help out with this but we shouldn't put random python in frontend/mobile stuff.

This is a script to be run either manually or during deploy, it's not trying to execute Python as part of the mobile app, and it wouldn't make sense to implement it as JS/TS. It's in this directory because it relates to the mobile app but it can go anywhere else, I just need whoever is authoritative in repo layout to tell me where.

> Also just FYI we do already have some locales on the mobile app here: app/mobile/i18n/locales

Those files are providing strings for the app itself. It makes sense for strings for the store page to live elsewhere and have their own Weblate component. They won't need to be deployed with the application.

nabramow:
> This is a script to be run either manually or during deploy, it's not trying to execute Python as part of the mobile app, and it wouldn't make sense to implement it as JS/TS. It's in this directory because it relates to the mobile app but it can go anywhere else, I just need whoever is authoritative in repo layout to tell me where.

Yes but the fact you need to install uv and python in the mobile app to have a single python script is not good. Why doesn't it make sense to implement it as JS/TS?

I feel pretty strongly about this.

Is it a one-time script we will delete after or something we will be running multiple times?

tristanlabelle:
> the fact you need to install uv and python in the mobile app

The mobile app is not affected because it's not app code nor called by JS/TS app code, it's a standalone script that talks to the Google Play Store API. A developer or CI needing to run the script will need to have uv/python installed. We'll be running it either every time we deploy a release to the app stores, or every time new translations for app store pages come in.

I can't quite figure out our misunderstanding.

nabramow:
> > the fact you need to install uv and python in the mobile app
> 
> The mobile app is not affected because it's not app code nor called by JS/TS app code, it's a standalone script that talks to the Google Play Store API. A developer or CI needing to run the script will need to have uv/python installed. We'll be running it either every time we deploy a release to the app stores, or every time new translations for app store pages come in.
> 
> I can't quite figure out our misunderstanding.

Our misunderstanding is that if this is code that is going to live in the mobile and/or web folders, it needs to be in Javascript.

tristanlabelle:
Will reimplement as a typescript script

PR #8425, opened 2026-04-29
Frontend: Remove tag input header text

tristanlabelle:
This header text has several issues, I believe more than make it worth keeping:

1. 'Enter' does not actually enter the text written as a filter. Try it: https://next.couchershq.org/profile/edit/about
2. The two sentences in one string are confusing to localize without context (why I looked into this). I think this specific control doesn't need to remind users that they can contact us.
3. The header only appears when the pop-up pops downwards:

image

image

## Testing
Ran locally.

image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/ProfileTagInput.tsx
+++ b/app/web/features/profile/ProfileTagInput.tsx
@@ -91,17 +91,6 @@ const StyledPopper = styled(Popper)(() => ({
   zIndex: 101,
 }));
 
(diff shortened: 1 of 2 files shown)


nabramow:
I'm going to tag @jesseallhands here as this is more his domain, if he's okay removing it or wants us to replace it somewhere else.

aapeliv (approved):
lgtm

PR #8429, opened 2026-04-30
chore: remove watch mode on test command

darrenvong:

With the existing instruction and `test` command, I noticed that when Claude tries to use it to verify its work hasn't broken existing/new tests it has written, it hangs since `yarn test` runs jest in watch mode.

I figured it's better to remove it and if anyone ever wants to run in watch mode again, appending `--watch` at the end of `yarn test` feels less of a nuisance than getting Claude stuck

## Testing

**Backend checklist**

**Web frontend checklist**
N/A - no actual code changes

## For maintainers


nabramow (approved):
Oh good call. Probably a better solution than be stopping claude, running the tests myself and then telling it what failed haha.

PR #8449, opened 2026-05-04
Hide message content in emails for low-mod_score senders

aapeliv:
Adds a `couchers.mod_score` module that centralises the rules driven by a user's `mod_score`. The first rule: when a sender's `mod_score < 1.5`, their message text is replaced with a placeholder in host request and conversation email notifications — the recipient still gets the email so they know they have a message, but they need to log in to read it.

Applies to: `host_request:create`, `host_request:message`, `chat:message`, and per-message in `chat:missed_messages`.

Also adds a test for the admin `SetModScore` RPC, including the default value of `1.0`.

## Testing

- `uv run pytest src/tests/test_admin.py src/tests/test_requests.py src/tests/test_conversations.py src/tests/test_notifications.py src/tests/test_bg_jobs.py` — all 189 pass
- `make format` and `make mypy` clean

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/i18n/locales/en.json
+++ b/app/backend/src/couchers/i18n/locales/en.json
@@ -294,6 +294,7 @@
     "dear_name": "Dear {{name}},",
     "thank_you": "Thank you!",
     "founders_signature": "Aapeli and Itsi,\nCouchers.org Founders",
+    "message_content_hidden": "Please log in to view this message.",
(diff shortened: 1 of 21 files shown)


aapeliv:
@claude this is completely wrong. We want to update the template, and all user visible strings we add need to be translatable.

PR #8452, opened 2026-05-05
Web/profile: One-shot message form on user profile

### LINKED ISSUE #6267: Streamlined direct message dialog
In #4301 (implemented by @andrei-kolesnik-california in #6211), we introduce a direct message API in the backend.

There should be a new frontend dialog for sending this direct message from the user's profile page, instead of the convoluted multi-step process in place currently.

aapeliv:
Human note:

This has always been a bit of a complicated UX piece so hopefully this helps

Screenshot 2026-05-04 at 22 11 22

Screenshot 2026-05-04 at 22 11 30

Also if there already is a 1:1 chat, it will go to it instead of opening up this pane. Also if profile is incomplete, it'll show the modal as normal.

---

Replaces the click-through to the messages page with an inline form, letting users send a first message to another user from their profile in a single step — mirrors the existing host request form pattern.

Backend already supported this via the `SendDirectMessage` gRPC RPC (creates-or-finds the DM and sends the message in one call), so this PR is web-only — it just wires that RPC into the service layer and builds the UI.

The Message button still navigates to an existing chat if a DM already exists between the two users; only the no-existing-DM case is replaced.

## Testing
- Updated `MessageUserButton.test.tsx` to cover the new "open inline form" path; existing tests pass (`yarn test` — 21 tests across `MessageUserButton`, `UserPage`, `UserOverview`).
- Lint passes (`yarn lint:fix`).
- Manual UI testing in browser: not yet done — would appreciate the reviewer exercising the golden path (Profile → Message → form → send → land in new chat) and the existing-DM path (Message → straight to existing chat).

To test locally:
1. Visit another user's profile while logged in
2. Click "Message"
3. With no prior DM: inline form should open below the tabs (like the host request form). Type a message, hit Send — you should land on the new group chat page.
4. With a prior DM: clicking Message should navigate straight to the existing chat (unchanged behavior).

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/messages/locales/en.json
+++ b/app/web/features/messages/locales/en.json
@@ -83,8 +83,7 @@
     "friends_input_label": "Friends",
     "title_label": "Title",
     "create_button": "Create chat",
-    "back_button": "Back",
-    "hosting_request_hint": "Sending a hosting request? Go <back>Back</back> and use the <request>Request</request> button instead."
+    "back_button": "Back"
   },
   "chat_input": {
     "label": "Message"
--- a/app/web/features/profile/actions/MessageUserButton.tsx
+++ b/app/web/features/profile/actions/MessageUserButton.tsx
@@ -1,15 +1,11 @@
-import { useMutation } from "@tanstack/react-query";
 import Button from "components/Button";
 import ProfileIncompleteDialog from "components/ProfileIncompleteDialog/ProfileIncompleteDialog";
 import useAccountInfo from "features/auth/useAccountInfo";
+import useMessageUser from "features/profile/hooks/useMessageUser";
 import { useTranslation } from "i18n";
 import { PROFILE } from "i18n/namespaces";
-import { useRouter } from "next/router";
 import { User } from "proto/api_pb";
 import { useState } from "react";
-import { service } from "service";
-
-import { routeToGroupChat } from "../../../routes";
 
 export default function MessageUserButton({
   user,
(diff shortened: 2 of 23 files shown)


jesseallhands:
As long as this is only accessible on the user profile (and not on the search page) this should be fine. The reason for this is we want to force users to visit the profile page of the person before they send a message or request. This is very important as one of the primary pain points for hosts is that people who request or message do not read the person's profile before doing so.

app/web/features/profile/actions/MessageUserButton.tsx
```diff
@@ -31,8 +33,8 @@ export default function MessageUserButton({
     },
     onSuccess(data) {
       if (!data) {
-        //no existing thread
-        router.push(routeToCreateMessage(user.username));
+        //no existing thread — open inline form
+        setIsMessaging(true);
       } else {
         //has thread
         router.push(routeToGroupChat(data));
```
nabramow: Since this is a bit long could you export this to a separate function like we usually do if possible? Unless it's gonna be very messy passing the router and state updates.

app/web/features/profile/view/NewMessage.tsx
```diff
@@ -0,0 +1,110 @@
...
+}));
+
+interface NewMessageProps {
+  setIsMessaging: (value: boolean) => void;
+}
+
+interface MessageFormData {
+  text: string;
+}
+
```
nabramow: Nit - move these types above the styling. But also since they are only one item each do we need to declare them separately like this vs inline? On first glance each one is only used once.

app/web/features/profile/view/ProfilePage.tsx
```diff
@@ -49,6 +49,9 @@ export default function ProfilePage({ tab = "about" }: { tab?: UserTab }) {
               setIsRequesting={() => {
                 /* TODO: not needed here*/
               }}
+              setIsMessaging={() => {
+                /* TODO: not needed here*/
+              }}
```
nabramow: Remove this then if not needed? Or allow it being undefined in the types instead if it's conditional?

aapeliv:
@claude address the review comments

jesseallhands (approved):

app/web/features/messages/locales/en.json
```diff
@@ -83,8 +83,7 @@
     "friends_input_label": "Friends",
     "title_label": "Title",
     "create_button": "Create chat",
-    "back_button": "Back",
-    "hosting_request_hint": "Sending a hosting request? Go <back>Back</back> and use the <request>Request</request> button instead."
```
jesseallhands, line 87: The text should read:

`Sending a hosting request? Use the <request>Request</request> button instead.`

Then the link should swap the form from being a message form to being a request form. Make sure the text is retained and not discarded (people will be pissed if they type up a lengthy message and then click the link and it deletes all their text).

jesseallhands (changes requested):

app/web/features/messages/locales/en.json
aapeliv: fixed

PR #8456, opened 2026-05-05
Backend/emails: Port emails to new system (part 1)

tristanlabelle:
Ports some mostly security-related emails to the new email system, which uses a single generic html template, unifies html and plaintext bodies, and supports localization. In `couchers.notifications.render_email`, we first try to render emails using the localizable email system (single generic template). If it's not yet handled, we fallback to the old system which has per-email customized html templates.

No changes in functionality.

Changes:

- Add branch to support new email system in `render_email_notifications`
- Move a few emails to the new email system definitions in `emails.py` and their strings to the `en.json`
- Remove the earlier proof of concept of `HostRequestEmail` (coming back in a future PR)
- Adds `EmailBlocksBuilder` to make it simpler to build email bodies out of blocks.
- Addresses some feedback from previous PR about using dict init syntax.
- Fixes newlines not preserved when rendering plaintext loc strings in html contexts.

## Testing
Ran `uv run dump-emails` and reviewed the outputs. See here: https://github.com/Couchers-org/couchers/commit/8c2a58807203f3befd8aa4897343cda3f461ddb9

Ran locally and received this:

image

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/email/emails.py
```diff
@@ -52,100 +62,141 @@ def dummy_data(cls) -> Self:
...
+
+    new_gender: str
+
+    @property
+    def _localize_key_prefix(self) -> str:
+        return "gender_changed"
+
+    def _build_body(self, builder: EmailBlocksBuilder, loc_context: LocalizationContext) -> None:
+        builder.para("body", {"gender": self.new_gender})
+        builder.security_warning_para()
```
aapeliv: why do you prefix `_build_body` with an underscore?

aapeliv (commented):
could you clarify what the changes are in functionality, if any? or is this just a change in the underlying plumbing?

app/backend/src/couchers/email/emails.py
tristanlabelle: it's meant as a protected method, only invoked by the base class' `get_body_blocks`.

tristanlabelle:
@aapeliv , no changes in functionality. Emails migrated to the new system don't need custom mjml/html/txt templates and source their strings from json, so they'll be localizable.

app/backend/src/couchers/email/emails.py
aapeliv: i see. i always thought of a single underscore as private... still doesn't look right to me :P

aapeliv (dismissed):
lgtm in theory. i am cautious about the huge influx of users and the outsized impact of any issues. please decide if you can wait a bit (maybe a week?), and when you do merge, just double check on staging afterwards that everything is working as intended.

app/backend/src/couchers/email/emails.py
tristanlabelle: Removed underscore prefix in latest.

tristanlabelle:
@aapeliv A week has gone by so I intend on merging this, but in the meanwhile I ported several other emails to the new system so I'll need your reapproval.

app/backend/src/couchers/notifications/render_email.py
```diff
@@ -39,49 +46,114 @@ def render_email_notification(
...
+    # Progressively migrate to the new email templating system in couchers.email.emails,
+    # which supports localization and uses a single generic html template.
+    if email := _get_generic_templated_email(user.name, notification):
+        subject = email.get_subject_line(loc_context)
+        preview = email.get_preview_line(loc_context)
+        body_blocks = email.get_body_blocks(loc_context)
+        body_plaintext = render_plaintext_body(blocks=body_blocks, footer=footer, loc_context=loc_context)
+        body_html = render_html_body(
+            subject=subject, preview=preview, blocks=body_blocks, footer=footer, loc_context=loc_context
+        )
```
aapeliv, line 62: could we have the source data here too; with version + info about which notification it was

aapeliv (dismissed):
lgtm, please add the version and then can merge

app/backend/src/couchers/notifications/render_email.py
tristanlabelle: Added, formatted as an SMTP header.

PR #8457, opened 2026-05-05
Events/i18n: Remove start/end display times from protos

tristanlabelle:
Localization of datetimes should happen on the frontend, which can decide what format to use based on the context (e.g. abbreviated months, weekdays or not). The backend-provided time display value was not localized properly and unused. Better just remove it.

## Testing
This was dead code. Tests should cover regressions.

**Backend checklist**

**Web frontend checklist**

## For maintainers


aapeliv (changes requested):
lgtm but i think you got the proto syntax a bit wrong

tristanlabelle:
@aapeliv Fixed the protobuf syntax error.

aapeliv (approved):
lgtm, thanks for the cleanup!

PR #8471, opened 2026-05-07
Backend/requests: Support .ics attachments

### LINKED ISSUE #8461: Backend/emails: support attachments (for event ical's)
Our backend representation of emails (`SendEmailPayload`) needs to support attachments so we can have add ical files.

tristanlabelle:
Generates and attaches iCalendar .ics files to host request accepted+confirmed emails, which are detected by mail clients like Gmail or manually importable in calendar apps.

Caveats:
- The host receives the .ics file in the host request accepted notification email, even though they haven't confirmed yet, because they don't receive a further email upon confirming that could carry the .ics.
- The surfer receives the .ics file upon the host confirming, which makes sense.
- Upon cancellation, only the party getting cancelled on receives an email with the cancellation .ics, since the cancelling party does not get an email.

Unfortunately the `ics` library pulls in the `arrow` and `tatsu` libraries. Its beefier alternative [icalendar](https://github.com/collective/icalendar) depends on `python-dateutil` and `tzdata`, which we are already pulling in. I can port over if this is a concern. The format is simple enough for our needs that we could even implement it ourselves.

Fixes #8461

## Testing
Run the backend and frontend, and follow these steps:

1. Log in as user A, send a host request to user B, log out
2. Log in as user B, accept the host request, notice in maildev the acceptance email to user A, download the email .eml file, then https://www.encryptomatic.com/viewer/ to extract the attachment, then import it into Gmail from settings, log out
3. Log in as user A, confirm the host request, notice in maildev the confirmation email to user B, validate ics attachment
4. Still as user A, cancel the host request, notice in maildev the confirmation email to user A, validate ics attachment

Also added tests

**Backend checklist**

## For maintainers


aapeliv (commented):
i love this idea, let me know when it's ready for review

app/backend/src/couchers/email/calendar_events.py
```diff
@@ -0,0 +1,100 @@
...
+            "calendar_events.host_requests.title_host", loc_context.locale, {"name": other_name}
+        )
+    else:
+        event.name = get_emails_i18next().localize(
+            "calendar_events.host_requests.title_surfer", loc_context.locale, {"name": other_name}
+        )
+
+    event.begin = host_request.from_date
+    event.end = host_request.to_date
+    event.make_all_day()  # Shifts the end date by one day
```
aapeliv: What does that mean?

app/backend/src/couchers/email/calendar_events.py
```diff
@@ -0,0 +1,100 @@
...
+    calendar = Calendar(creator=f"-//Couchers.org//Couchers//{loc_context.locale.upper()}")
+    if event.status == "CANCELLED":
+        calendar.method = "CANCEL"
+    calendar.events.add(event)
+    return cast(str, calendar.serialize())
+
+
+def ics_to_attachment(ics: str) -> EmailAttachment:
+    return EmailAttachment(
+        filename="host_request.ics",
```
aapeliv: should this not be hardcoded in a function with this generic name?

aapeliv (dismissed):
i think it looks good overall

app/backend/src/couchers/email/calendar_events.py
tristanlabelle: Answered in code:

```py
    # Our to_date is inclusive, iCalendar's DTEND is exclusive (for full-day events)
    # make_all_day will adjust the end date by one day accordingly.
    event.begin = host_request.from_date
    event.end = host_request.to_date
    event.make_all_day()

app/backend/src/couchers/email/calendar_events.py
tristanlabelle: Right, updated in latest

tristanlabelle:
@aapeliv Addressed comments. Ready for ✅ !

aapeliv (approved):

PR #8478, opened 2026-05-07
Backend: Strongly type the config object

tristanlabelle:
Redefines the `config` dict to a strongly-typed dataclass leveraging type annotations and default values, such that references throughout the codebase can be validated by mypy.

## Testing
Leaving to tests - I figure either everything will work or nothing will.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/app.py
+++ b/app/backend/src/app.py
@@ -17,7 +17,7 @@ import sentry_sdk
 from sentry_sdk.integrations import excepthook
 from sqlalchemy.sql import text
 
-from couchers.config import check_config, config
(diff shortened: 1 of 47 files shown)


aapeliv (commented):
should we keep them uppercase?

tristanlabelle:
@aapeliv I don't see a benefit to making a style exception and making the Python properties upper case. For consuming code, it's an implementation detail that the values come from environment variables.

tristanlabelle:
Closing in favor of #8932

PR #8510, opened 2026-05-10
Backend/moderation: UMS coverage for thread comments and replies

aapeliv:
Adds UMS (Unified Moderation System) integration for thread `Comment` and `Reply` content — the first of five mod-team-prioritised content types being onboarded onto UMS.

New comments and replies are created with `ModerationVisibility.shadowed` and an `initial_review` queue item, matching the existing UMS pattern (host requests, event occurrences, etc.). Authors can see their own shadowed content; everyone else sees nothing until a moderator approves. Notifications generated by a new comment or reply are now gated on the new content's own `moderation_state_id`, so they fire only after that content is approved (rather than firing immediately and only being deferred by the parent event's state, as previously).

`total_num_responses` now counts only publicly visible content (`visible` + `unlisted`); authors of shadowed content will see one extra reply in the list than the count suggests — accepted to avoid threading a context through `event_to_pb` / `discussion_to_pb` / `page_to_pb`.

Adds `Moderator.approve_comment`, `approve_reply`, and a packed-id-aware `approve_thread_post` helper so existing tests can keep asserting visibility assumptions after the default-shadowed change.

⚠️ **Migration**: autogen will add the `moderation_state_id` FK columns on `comments` and `replies`, but those columns are NOT NULL and the new enum values can't be added in the same transaction as DML. A hand-written data migration is needed before merging — closely modelled on `0131_implement_ums_for_events.py` (enum rename/recreate + per-row `ModerationState` backfill at `visible` + log entries).

## Testing

- `make format` — clean
- `make mypy` — clean for changed files (one preexisting error in `tests/fixtures/misc.py:31` unrelated to this PR)
- `uv run pytest` — 938 tests pass; the only failure is `test_db.py::test_migrations`, which is the expected schema-vs-migration diff that the autogen migration will resolve

New tests in `test_threads.py` cover:
- ModerationState + queue item creation on comment/reply post
- Shadowed comment/reply visible to author only
- Admin approve flow making content visible to others
- Admin hide flow removing content from author and others
- `total_num_responses` excludes shadowed content

Existing affected tests updated:
- `test_event_threads`, `test_event_comment_notification_has_moderation_state`, `test_event_thread_reply_notification_has_moderation_state` (test_events.py)
- `test_discussion_notifications_regression` (test_discussions.py)

**Backend checklist**
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history — see ⚠️ above; needs hand-written data backfill before merge

## For maintainers


app/backend/src/couchers/servicers/threads.py
```diff
@@ -35,16 +47,29 @@ def unpack_thread_id(thread_id: int) -> tuple[int, int]:
     return divmod(thread_id, 10)
 
 
+_PUBLIC_VISIBILITIES = (ModerationVisibility.visible, ModerationVisibility.unlisted)
```
aapeliv: no, always use helpers

app/backend/src/couchers/migrations/versions/0152_backend_moderation_ums_coverage_for_.py
```diff
@@ -0,0 +1,124 @@
...
+            ORDER BY id
+            RETURNING id, object_id
+        )
+        UPDATE comments
+        SET moderation_state_id = new_states.id
+        FROM new_states
+        WHERE comments.id = new_states.object_id
+    """)
+    op.execute("""
+        INSERT INTO moderation_log (id, moderation_state_id, action, moderator_user_id, new_visibility, reason)
```
aapeliv, line 50: need to use backfill ID range

app/backend/src/couchers/servicers/threads.py
```diff
@@ -103,7 +128,7 @@ def generate_reply_notifications(payload: jobs_pb2.GenerateReplyNotificationsPay
                             event=event_to_pb(session, occurrence, context),
                             author=user_model_to_pb(author_user, session, context),
                         ),
-                        moderation_state_id=occurrence.moderation_state_id,
+                        moderation_state_id=comment.moderation_state_id,
```
aapeliv, line 122: need a mechanism for parent-> child dependency

app/backend/src/tests/fixtures/misc.py
```diff
@@ -174,3 +174,53 @@ def approve_event_occurrence(self, occurrence_id: int, reason: str = "Test appro
...
+            api.ModerateContent(
+                moderation_pb2.ModerateContentReq(
+                    moderation_state_id=state_res.moderation_state.moderation_state_id,
+                    action=moderation_pb2.MODERATION_ACTION_APPROVE,
+                    visibility=moderation_pb2.MODERATION_VISIBILITY_VISIBLE,
+                    reason=reason,
+                )
+            )
+
+    def approve_reply(self, reply_id: int, reason: str = "Test approval") -> None:
```
aapeliv, line 197: need to refactor these

app/backend/src/couchers/migrations/versions/0152_backend_moderation_ums_coverage_for_.py
aapeliv: look at first UMS migration

app/backend/src/tests/fixtures/misc.py
```diff
@@ -174,3 +174,53 @@ def approve_event_occurrence(self, occurrence_id: int, reason: str = "Test appro
...
+                )
+            )
+
+    def approve_thread_post(self, packed_thread_id: int, reason: str = "Test approval") -> None:
+        """Approve whichever of Comment/Reply the packed thread_id refers to.
+
+        packed_thread_id uses the convention from couchers.servicers.threads.pack_thread_id:
+        the trailing digit is depth (1=Comment, 2=Reply); the rest is the database id.
+        """
+        database_id, depth = divmod(packed_thread_id, 10)
```
tristanlabelle: Use `unpack_thread_id` to avoid duplicating the logic. You could also move the comment about how it works to that function's documentation.

app/backend/src/couchers/servicers/threads.py
```diff
@@ -35,24 +45,29 @@ def unpack_thread_id(thread_id: int) -> tuple[int, int]:
     return divmod(thread_id, 10)
 
 
-def total_num_responses(session: Session, database_id: int) -> int:
-    """Return the total number of comments and replies to the thread with
-    database id database_id.
-    """
-    comments = select(func.count()).select_from(Comment).where(Comment.thread_id == database_id)
-    replies = (
+def total_num_responses(session: Session, context: CouchersContext, database_id: int) -> int:
```
tristanlabelle, line 49: The caveat in your PR description could be useful in a comment here.

> total_num_responses now counts only publicly visible content (visible + unlisted); authors of shadowed content will see one extra reply in the list than the count suggests — accepted to avoid threading a context through event_to_pb / discussion_to_pb / page_to_pb.

app/backend/src/couchers/servicers/threads.py
```diff
@@ -35,24 +45,29 @@ def unpack_thread_id(thread_id: int) -> tuple[int, int]:
...
+        .where(Comment.thread_id == database_id),
+        context,
+        Reply,
+        is_list_operation=True,
     )
     return session.execute(comments).scalar_one() + session.execute(replies).scalar_one()
 
 
-def thread_to_pb(session: Session, database_id: int) -> threads_pb2.Thread:
+def thread_to_pb(session: Session, context: CouchersContext, database_id: int) -> threads_pb2.Thread:
```
tristanlabelle, line 68: Thought: We could just pass in the logged in user ID to minimize tying this operation to the whole `CouchersContext`. That's the only information needed by `where_moderated_content_visible`.

tristanlabelle (commented):
Looks good to me, modulo your own todo comments.

app/backend/src/couchers/servicers/threads.py
aapeliv: updated above to use `where_moderated_content_visible` which needs it

app/backend/src/couchers/servicers/threads.py
aapeliv: i think for now it's fine as for all objects, children are unreachable if parent is not visible

PR #8516, opened 2026-05-10
Backend/moderation: UMS coverage for references

aapeliv:
Brings `Reference` under the Unified Moderation System (4th of 5 types per the mod-team rollout — Event was skipped for now).

- New references start `shadowed` with an `initial_review` queue item, matching the established Comment/Reply and Discussion pattern
- `ListReferences` filters through `where_moderated_content_visible` (list mode); notification delivery is gated on the new `moderation_state_id`
- Migration `0151` extends the `moderationobjecttype` enum and backfills `ModerationState` rows for every existing reference (`visibility=hidden` for `is_deleted=True` rows, `visible` otherwise) — using the rename/recreate enum trick and the `< 2_000_000` ID range, matching `0149`/`0150`
- `DeleteReference` is deprecated: aborts `FAILED_PRECONDITION` and points admins at UMS for hiding. `EditReferenceText` is unrelated to visibility moderation and left alone
- `private_text` is intentionally NOT included in `moderation_state_to_pb` — only the public `text` is. `private_text` is moderator-only by design

Stacked on top of `backend/feature/ums-for-discussions` (PR #2 in the UMS rollout).

## Testing
- `make format && make mypy` clean (one pre-existing mypy error in `tests/fixtures/misc.py:31` unrelated to this PR)
- Full backend suite: 946 tests green
- New UMS-specific tests in `test_references.py`: state created on creation, shadowed visibility for non-authors, hidden references hidden from author too, `DeleteReference` errors

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/moderation.py
```diff
@@ -276,14 +218,13 @@ def moderation_state_to_pb(state: ModerationState, session: Session) -> moderati
             select(Discussion.creator_user_id, Discussion.title, Discussion.content).where(Discussion.id == object_id)
         ).one()
         content = f"{title}\n\n{body}"
+    elif object_type == ModerationObjectType.reference:
+        author_user_id, content = session.execute(
+            select(Reference.from_user_id, Reference.text).where(Reference.id == object_id)
+        ).one()
     else:
```
tristanlabelle, line 271: If protoc generates good python, a match statement would enable mypy to do exhaustivity checks. Also, could we use the `__moderation_author_id__` value to make this logic generic?

app/backend/src/tests/test_email.py
```diff
@@ -115,13 +118,21 @@ def test_reference_report_email_not_sent(db):
...
+        session.add(moderation_state)
+        session.flush()
         reference = Reference(
             from_user_id=from_user.id,
             to_user_id=to_user.id,
             reference_type=ReferenceType.friend,
             text="This person was very nice to me.",
             rating=0.9,
             was_appropriate=True,
+            moderation_state_id=moderation_state.id,
```
tristanlabelle, line 135: Missing patching to nonzero

app/backend/src/tests/test_email.py
```diff
@@ -115,13 +118,21 @@ def test_reference_report_email_not_sent(db):
     make_friends(from_user, to_user)
 
     with session_scope() as session:
+        moderation_state = ModerationState(
+            object_type=ModerationObjectType.reference,
+            object_id=0,
+            visibility=ModerationVisibility.visible,
+        )
+        session.add(moderation_state)
+        session.flush()
```
tristanlabelle, line 127: Looking at this pattern it seems quite feasible to implement a helper:

`def add_with_moderation_state(obj: Moderable, session: Session, visibility: ModerationVisibility = ModerationVisibility.visible) -> None`

We can use `from typing import Protocol` to define a structural `Moderable` interface for all types which should have a `moderation_state_id`.

app/backend/src/tests/test_events.py
tristanlabelle, line 3141: Use the unpack function.

app/backend/src/tests/test_references.py
```diff
@@ -428,7 +452,7 @@ def test_ListReference_banned_deleted_users(db):
         assert len(refs_sent) == 1
 
 
-def test_WriteFriendReference(db):
+def test_WriteFriendReference(db, moderator):
```
tristanlabelle, line 455: Better strongly type as in other tests.

app/backend/src/tests/test_threads.py
tristanlabelle, line 159:
```suggestion
def _make_thread_and_comment(token, content="hello") -> tuple[int, int]:
```

app/backend/src/couchers/models/discussions.py
tristanlabelle, line 88: `__moderation_author_column__` seems private to us, whereas `__tablename__` is treated specially by sqlalchemy. I think we should use a different convention.

app/backend/src/tests/fixtures/misc.py
tristanlabelle, line 195: It seems like these methods' bodies are duplicated. Can't you all delegate to a single private function that takes `object_type`, `object_id` and `reason` params?

app/backend/src/tests/test_discussions.py
tristanlabelle, line 346: `moderator` could expose `moderate_object(type, id, action, visibility, reason)` since it does this operation internally itself a lot. Maybe one of action/visibility can also default to be deduced from the other.

tristanlabelle (dismissed):
All looks good. My biggest feedback is that I think we could abstract the pattern of creating a moderation object then the desired moderated object, then post-patching it, and maybe also simplify `create_moderation` similarly.

tristanlabelle:
Seems like this PR has now been rolled into https://github.com/Couchers-org/couchers/pull/8518 ? If so please close to avoid rereviewing for naught.

aapeliv:
> Seems like this PR has now been rolled into #8518 ? If so please close to avoid rereviewing for naught.

Plan is to merge this one, then rebase that PR onto this later

app/backend/src/tests/fixtures/misc.py
aapeliv: yes, will do in a follow-up pr

app/backend/src/tests/test_email.py
aapeliv: what do you mean?

app/backend/src/tests/test_email.py
aapeliv: let me try in a follow up PR

app/backend/src/tests/test_email.py
tristanlabelle: The `ModerationState` above is created with `object_id=0` and not updated with the new `Reference`'s `id` afterwards, as elsewhere.

Maybe it doesn't matter for this test but it's not obvious from the code, and why it'd be better to encapsulate this patching operation.

aapeliv: oh right, good catch

PR #8550, opened 2026-05-12
Web/bugfix/language picker mobile

### LINKED ISSUE #8097: Language Picker Not Avaliable On Mobile Without Logging In
Nicole Note: On mobile web responsive view specifically, the language picker does not show on the landing page. I think the best way to do this is to adjust the conditional rendering of the LanguagePickerSelect to show if isMobile is true and the user is also logged out.

-----

Subject: Language Picker Not Avaliable On Mobile Without Logging In
Description:
While logged out, users are unable to choose their preffered language via the language picker.

Results:

Backend version: develop-463b430e
Frontend version: develop-15d93f6d
User Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36
Screen resolution: 1920x1085
Page: https://couchers.org/tr/landing
User: [@thelosttraveler](https://couchers.org/user/thelosttraveler) (64392)

christiansaavedra:
This PR implements conditional validation for displaying the language selector in the navigation bar. The language picker is shown on desktop and on mobile for unauthenticated users, but hidden on mobile for authenticated users to streamline the interface and reduce clutter. Jest uni tests added.

The solution uses a useMemo hook in the [Navigation.tsx] component to compute shouldShowLanguagePickerSelect boolean based on isMobile and authState.authenticated.

On mobile devices, the selector is only displayed when the user is not authenticated, improving the user experience by avoiding unnecessary UI elements for logged-in users who likely have their language preferences set elsewhere.

Testing
Jest files added. I tested this by running the web app locally using yarn start. I verified the language selector visibility on desktop (always shown), mobile unauthenticated (shown), and mobile authenticated (hidden). Used browser dev tools to simulate mobile viewports and authentication states. No console warnings were observed during testing.

Reference issues
"closes #8097"

**Backend checklist**

**Web frontend checklist**
- [x ] There are no console warnings when running the app
- [x ] Added tests where relevant
- [x ] Checked Desktop, Mobile and Tablet screen sizes

## For maintainers


app/web/components/Navigation/Navigation.test.tsx
```diff
@@ -79,3 +84,69 @@ describe("Navigation", () => {
     expect(screen.queryByLabelText("Donation banner")).not.toBeInTheDocument();
   });
 });
+
+it("renders the language picker on mobile when the user is logged out", () => {
+  const originalMatchMedia = window.matchMedia;
+  window.matchMedia = createMatchMedia(800);
+
```
nabramow, line 91: Just out of curiosity, what does this do? First time I've seen it!

nabramow:
So the theory behind this is sound, the only issue is on Login and Signup page, we end up with double language picker and the navigation bar gets very cluttered:

Screenshot 2026-05-13 at 16 53 56

Maybe the solution is add it to this menu at the bottom when isMobile is true and the user is logged out:

Screenshot 2026-05-13 at 16 55 09

And then remove it from the bottom of the Login and Signup forms?

I am open to other suggestions though!

app/web/components/Navigation/Navigation.test.tsx
christiansaavedra: Hi! 

The 800 value represents a simulated viewport width of 800 pixels. This mocks the media query system to behave as if the browser window is 800px wide, which is typically considered a "mobile" or smaller screen size in MUI responsive design breakpoints. 

After this particular test is run, the  window.matchMedia = originalMatchMedia; resets the matchMedia value to the original one.

christiansaavedra:
> So the theory behind this is sound, the only issue is on Login and Signup page, we end up with double language picker and the navigation bar gets very cluttered:
> 
> Screenshot 2026-05-13 at 16 53 56
> Maybe the solution is add it to this menu at the bottom when isMobile is true and the user is logged out:
> 
> Screenshot 2026-05-13 at 16 55 09
> And then remove it from the bottom of the Login and Signup forms?
> 
> I am open to other suggestions though!

That's a very interesting observation, thanks for bringing it up. 

I totally agree! My input to add is, on mobile, to move the Language Selector to the place you suggested (the bottom of the hamburger menu) but on all pages, instead of conditionally rendering according to the page. This way, we keep it on a definitive place and help users to learn where the language selector is located.

What do you think?

nabramow:
> That's a very interesting observation, thanks for bringing it up.
> 
> I totally agree! My input to add is, on mobile, to move the Language Selector to the place you suggested (the bottom of the hamburger menu) but on all pages, instead of conditionally rendering according to the page. This way, we keep it on a definitive place and help users to learn where the language selector is located.
> 
> What do you think?

When the user is logged in, the LanguagePickerSelect is always in the LoggedInMenu like this:

Screenshot 2026-05-14 at 09 18 02

It's only logged out that's the problem, which is why it was originally manually added to the Login and Signup pages.

That's why I think it makes sense to remove those login and signup ones and instead add it to the bottom of the LoggedOutMenu when they are logged out.

christiansaavedra:
> > That's a very interesting observation, thanks for bringing it up.
> > I totally agree! My input to add is, on mobile, to move the Language Selector to the place you suggested (the bottom of the hamburger menu) but on all pages, instead of conditionally rendering according to the page. This way, we keep it on a definitive place and help users to learn where the language selector is located.
> > What do you think?
> 
> When the user is logged in, the LanguagePickerSelect is always in the LoggedInMenu like this:
> 
> Screenshot 2026-05-14 at 09 18 02
> It's only logged out that's the problem, which is why it was originally manually added to the Login and Signup pages.
> 
> That's why I think it makes sense to remove those login and signup ones and instead add it to the bottom of the LoggedOutMenu when they are logged out.

I like that! I've pushed the change to this PR.

nabramow:
Getting there, but I think some logic needs to be adjusted. It should not show in the NavigationBar when the user is logged in like this:

Screenshot 2026-05-21 at 10 58 19

When logged in, it's already in the LoggedInMenu like this:

Screenshot 2026-05-21 at 10 58 35

We don't want to change any of the logged in logic, just the logged out logic.

nabramow:
Maybe put the select below the options so it matches more the LoggedInMenu style?

Screenshot 2026-05-21 at 11 00 37


Also we probably don't want the menu to automatically close when they select something in case they still want to go to something in the menu.

nabramow:
@christiansaavedra Are you still planning on finishing this? Just checking so it doesn't stay blocked a long time.

christiansaavedra:
> @christiansaavedra Are you still planning on finishing this? Just checking so it doesn't stay blocked a long time.

Hey Nicole, can you confirm me if this is the expected behaviour? Im a bit confused but my understanding is that this should be the expected behavior:

a) Logged in case: 
- Render Language selector inside the hamburger menu, at the bottom. 
- Dont render the Language selector in the Navbar

b) Logged out case:
- Dont render Language Selector inside the hamburger menu
- Render the language page in the navbar

Please let me know, thanks

nabramow:
> > @christiansaavedra Are you still planning on finishing this? Just checking so it doesn't stay blocked a long time.
> 
> Hey Nicole, can you confirm me if this is the expected behaviour? Im a bit confused but my understanding is that this should be the expected behavior:
> 
> a) Logged in case:
> 
> * Render Language selector inside the hamburger menu, at the bottom.
> * Dont render the Language selector in the Navbar
> 
> b) Logged out case:
> 
> * Dont render Language Selector inside the hamburger menu
> * Render the language page in the navbar
> 
> Please let me know, thanks

Here are the four views we want:

**Desktop**
- Logged Out: LanguagePicker is already in the nav bar - all good as is
- LoggedIn: LanguagePicker already in the nav bar - all good as is

**Mobile View**
- Logged Out: Add LanguagePicker to the left side hamburger `LoggedOutMenu`:
Screenshot 2026-06-01 at 18 43 55

This shows on landing page, sign up, login so should cover everything. Also remove LanguagePicker from the bottom of the Signup and Login pages

- LoggedIn - This is already good the way it is (it shows in the right side hamburger `LoggedInMenu`


Let me know if that makes sense!

christiansaavedra:
> > > @christiansaavedra Are you still planning on finishing this? Just checking so it doesn't stay blocked a long time.
> > 
> > 
> > Hey Nicole, can you confirm me if this is the expected behaviour? Im a bit confused but my understanding is that this should be the expected behavior:
> > a) Logged in case:
> > 
> > * Render Language selector inside the hamburger menu, at the bottom.
> > * Dont render the Language selector in the Navbar
> > 
> > b) Logged out case:
> > 
> > * Dont render Language Selector inside the hamburger menu
> > * Render the language page in the navbar
> > 
> > Please let me know, thanks
> 
> Here are the four views we want:
> 
> **Desktop**
> 
> * Logged Out: LanguagePicker is already in the nav bar - all good as is
> * LoggedIn: LanguagePicker already in the nav bar - all good as is
> 
> **Mobile View**
> 
> * Logged Out: Add LanguagePicker to the left side hamburger `LoggedOutMenu`:
> 
> Screenshot 2026-06-01 at 18 43 55
> This shows on landing page, sign up, login so should cover everything. Also remove LanguagePicker from the bottom of the Signup and Login pages
> 
> * LoggedIn - This is already good the way it is (it shows in the right side hamburger `LoggedInMenu`
> 
> Let me know if that makes sense!

Pushed, let me know! Thank you

app/web/components/Navigation/Navigation.tsx
```diff
@@ -329,7 +344,11 @@ export default function Navigation() {
           </ListItem>
         ))}
       </List>
-    </div>
+
+      <Box sx={{ marginX: "auto", marginBottom: theme.spacing(2) }}>
```
nabramow:
```suggestion
      <Box
        sx={{ marginX: "auto", marginBottom: theme.spacing(2) }}
        onClick={(e) => e.stopPropagation()}
      >
        <LanguagePickerSelect />
      </Box>
```

This will stop the drawer from closing when a user selects something.

nabramow:
@christiansaavedra Put in one small suggestion then this is ready to go!

christiansaavedra:
> @christiansaavedra Put in one small suggestion then this is ready to go!

Awesome, I accepted it. LFG!

nabramow (approved):
LGTM now!

PR #8566, opened 2026-05-13
Add edit and delete discussions and comments

### LINKED ISSUE #8565: Add update and delete to discussion posts
We want to allow users to edit and delete discussion topics. For the backend part, we should add that to the existing API.

Keep in mind also what is needed for moderation tooling. @aapeliv is a good one to ask about that if you pick up this ticket.

IMO if a discussion topic or a comment has replies, it should replace the cards with a "this discussion has been deleted" and keep the replies.

Editing something should add a note that it was edited with the time.

nabramow:
Adds the ability to update and edit discussions, comments and replies. A few decisions were made here:

- Store edits in three versions tables, requested by @aapeliv for moderation purposes
- When a discussion/comment is deleted, the replies and comments within it will stay, and only that card will be replaced with one that says "This discussion/comment was deleted".
- For the versions tables I am storing the new and previous content in each entry rather than just the previous. My thinking is that this is better for whatever moderation queries we'll need to make.

Since this PR contains backend code it can't be tested on Vercel.

Closes #8565 

## Testing
- Go to a community and add a discussion
- Add a comment to it
- Add a reply to that comment
- Delete the comment, the reply to that comment should still be there and comment replaced with "This comment has been deleted"
- Add another comment
- Delete the main discussion
- It should be replaced with "This discussion has been deleted" but still show the replies
- Now delete all the comments and replies
- Go back to the main discussions page and the thread should be gone

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/servicers/admin.py
```diff
@@ -886,6 +886,17 @@ def EditDiscussion(
...
 
+    def DeleteDiscussion(
+        self, request: admin_pb2.AdminDeleteDiscussionReq, context: CouchersContext, session: Session
+    ) -> empty_pb2.Empty:
+        discussion = session.execute(
+            select(Discussion).where(Discussion.id == request.discussion_id)
+        ).scalar_one_or_none()
+        if not discussion:
+            context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "discussion_not_found")
+        discussion.deleted = now()
```
tristanlabelle, line 1017: Should we no-op if already deleted?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -471,10 +472,10 @@ def ListDiscussions(
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "community_not_found")
         if not node.official_cluster.small_community_features_enabled:
             context.abort_with_error_code(grpc.StatusCode.FAILED_PRECONDITION, "discussions_not_enabled")
+        has_comments = exists().where((Comment.thread_id == Discussion.thread_id) & (Comment.deleted == None))
```
tristanlabelle: How does this interact with UMS and shadowed comments?

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +159,70 @@ def GetDiscussion(
...
+                context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "missing_discussion_title")
+            discussion.title = new_title
+
+        if request.HasField("content"):
+            new_content = request.content.value.strip()
+            if not new_content:
+                context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "missing_discussion_content")
+            discussion.content = new_content
+
+        discussion.last_edited = now()
```
tristanlabelle, line 233: Should we no-op if none of the fields were actually updated?

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +159,70 @@ def GetDiscussion(
...
+        self, request: discussions_pb2.UpdateDiscussionReq, context: CouchersContext, session: Session
+    ) -> discussions_pb2.Discussion:
+        discussion = session.execute(
+            select(Discussion).where(Discussion.id == request.discussion_id)
+        ).scalar_one_or_none()
+        if not discussion:
+            context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "discussion_not_found")
+        if discussion.deleted is not None:
+            context.abort_with_error_code(grpc.StatusCode.FAILED_PRECONDITION, "discussion_deleted")
+        if context.user_id != discussion.creator_user_id:
```
tristanlabelle, line 197: Should moderators be able to update existing discussions?

app/backend/src/couchers/servicers/threads.py
```diff
@@ -36,15 +37,18 @@ def unpack_thread_id(thread_id: int) -> tuple[int, int]:
...
-    comments = select(func.count()).select_from(Comment).where(Comment.thread_id == database_id)
+    comments = (
+        select(func.count()).select_from(Comment).where(Comment.thread_id == database_id).where(Comment.deleted == None)
+    )
     replies = (
         select(func.count())
         .select_from(Reply)
         .join(Comment, Comment.id == Reply.comment_id)
         .where(Comment.thread_id == database_id)
+        .where(Reply.deleted == None)
```
tristanlabelle: FYI: Might conflict with Aapeli's https://github.com/Couchers-org/couchers/pull/8510

app/backend/src/couchers/servicers/threads.py
```diff
@@ -237,10 +243,15 @@ def GetThread(
...
+                    deleted=r.deleted is not None,
+                    content=r.content if r.deleted is None else "",
+                    author_user_id=r.author_user_id if r.deleted is None else 0,
+                    created_time=Timestamp_from_datetime(r.created) if r.deleted is None else None,
                     num_replies=n,
+                    can_edit=(context.user_id == r.author_user_id) if r.deleted is None else False,
+                    last_edited=Timestamp_from_datetime(r.last_edited)
+                    if (r.last_edited and r.deleted is None)
+                    else None,
                 )
```
tristanlabelle: It could be less error-prone to have two calls to `threads_pb2.Reply` within a single if-else for `r.deleted`. Then it'd be immediately clear that the deleted case only include `thread_id`, `deleted` and `num_replies`.

app/backend/src/couchers/servicers/threads.py
```diff
@@ -320,3 +334,64 @@ def PostReply(
...
+        elif depth == 2:
+            obj = session.execute(select(Reply).where(Reply.id == database_id)).scalar_one_or_none()
+        else:
+            context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "thread_not_found")
+
+        if not obj:
+            context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "thread_not_found")
+        if obj.deleted is not None:
+            context.abort_with_error_code(grpc.StatusCode.FAILED_PRECONDITION, "reply_deleted")
+        if obj.author_user_id != context.user_id:
```
tristanlabelle, line 510: `DeleteDiscussion` had a `can_moderate_node` condition here. Does it not apply?

app/backend/src/tests/test_discussions.py
```diff
@@ -144,6 +151,289 @@ def test_create_and_get_discussion(db, push_collector: PushCollector):
...
+        res = api.UpdateDiscussion(
+            discussions_pb2.UpdateDiscussionReq(
+                discussion_id=discussion_id,
+                title=StringValue(value="Updated title"),
+                content=StringValue(value="Updated content"),
+            )
+        )
+        assert res.title == "Updated title"
+        assert res.content == "Updated content"
+        assert res.can_edit
```
tristanlabelle, line 186: The test would be stronger if calling GetDiscussion to see the changed values vs only checking the response.

tristanlabelle (commented):
I can sign off on the backend changes if you split this PR. It's very large to review the frontend part as well. Left noncritical comments.

aapeliv (changes requested):
let's also wait for ums/mod stuff first

app/backend/src/couchers/servicers/threads.py
nabramow: I think I've handled the merge conflicts now

app/backend/src/couchers/servicers/threads.py
nabramow: I adjusted it, the for loop feels a bit more verbose but claude said this is the more "python" way to do it haha, so someone who is more familiar with python please correct me if I'm wrong.

app/backend/src/couchers/servicers/discussions.py
nabramow: I think delete yes, but editing peoples words seems non-standard IMO.

app/backend/src/couchers/servicers/threads.py
nabramow: Ah so I looked into this and realized `can_moderate` node also includes the community builders (admins). I think actually they shouldn't be able to delete or edit discussions, comments or replies so I will remove it as we only want mods to have that power. Good catch!

app/backend/src/couchers/servicers/threads.py
tristanlabelle: You could extract a `reply_proto_to_model` function to help with the verbosity part.

app/backend/src/couchers/servicers/threads.py
tristanlabelle: Our terminology is confusing if `can_moderate` is broader than `mods`.

app/backend/src/couchers/servicers/communities.py
nabramow: Have changed this now after merging in Aapeli's UMS stuff.

app/backend/src/couchers/models/discussions.py
```diff
@@ -29,6 +30,8 @@ class Discussion(Base, kw_only=True):
     thread_id: Mapped[int] = mapped_column(ForeignKey("threads.id"), unique=True)
     moderation_state_id: Mapped[int] = mapped_column(ForeignKey("moderation_states.id"), index=True)
     created: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), init=False)
+    deleted: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), default=None)
+    last_edited: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), default=None)
```
aapeliv, line 34: non-nullable

app/backend/src/couchers/models/discussions.py
```diff
@@ -96,6 +99,7 @@ class Comment(Base, kw_only=True):
     content: Mapped[str] = mapped_column(String)  # CommonMark without images
     created: Mapped[datetime] = mapped_column(DateTime(timezone=True), server_default=func.now(), init=False)
     deleted: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), default=None)
+    last_edited: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), default=None)
```
aapeliv, line 102: ditto. server default func.now

app/backend/src/couchers/models/discussions.py
```diff
@@ -117,10 +121,84 @@ class Reply(Base, kw_only=True):
...
```
aapeliv, line 124: same

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -47,6 +60,8 @@ def discussion_to_pb(session: Session, discussion: Discussion, context: Couchers
         content=discussion.content,
         thread=thread_to_pb(session, context, discussion.thread_id),
         can_moderate=can_moderate,
+        can_edit=(context.user_id == discussion.creator_user_id),
+        last_edited=Timestamp_from_datetime(discussion.last_edited) if discussion.last_edited else None,
```
aapeliv, line 64: redundant

aapeliv (changes requested):
let's also wait for ums/mod stuff first

PR #8567, opened 2026-05-14
Replace Statsig with GrowthBook for experimentation

aapeliv:
Swaps the backend experimentation framework from Statsig to GrowthBook's sync Python SDK. The public API (`check_gate`) is unchanged.

- `statsig-python-core` → `growthbook>=1.2.0` in `pyproject.toml`
- `experimentation.py` rewritten on the sync `GrowthBook` client; the shared `feature_repo` cache is pre-warmed in `setup_experimentation()`
- `STATSIG_SERVER_SECRET_KEY` / `STATSIG_ENVIRONMENT` config → `GROWTHBOOK_API_HOST` / `GROWTHBOOK_CLIENT_KEY` (also updated in `backend.dev.env` and `conftest.py`)
- Dropped the unused `get_experiment`, `get_dynamic_config`, and `log_event` helpers (the last one also collided with `couchers.event_log.log_event`)
- Dropped the elaborate fork-safety boilerplate; the sync GrowthBook client has no background threads, so the comments in `app.py` / `worker.py` were pared down accordingly
- Smoke-test gate renamed `test_statsig_integration` → `test_growthbook_integration`

## Testing
- `make format` is clean
- `make mypy` reports no new errors (the 17 pre-existing errors are unrelated — proto `attachments`, abstract `Account.DismissReminder`, `ics` stubs)
- `from couchers.experimentation import check_gate, setup_experimentation` imports cleanly with the new `growthbook` dep installed
- Local smoke check still needed: bring the backend up against a real GrowthBook instance and confirm the `test_growthbook_integration` gate flips as expected with `EXPERIMENTATION_ENABLED=1`

**Backend checklist**

## For maintainers


app/backend/src/couchers/experimentation.py
```diff
@@ -67,158 +94,108 @@ def setup_experimentation() -> None:
...
+            _record_exposure(user_id, experiment, result)
+
+        gb = GrowthBook(
+            attributes={"id": str(user_id)},
+            features=features,
+            savedGroups=saved_groups,
+            on_experiment_viewed=on_experiment_viewed,
+        )
+        context._growthbook = gb  # type: ignore[attr-defined]
+    return context._growthbook  # type: ignore[attr-defined, no-any-return]
```
WouldYouKindly: Can we add `_growthbook` as an attribute to CouchersContext, with None as default? Then there will be no need for type ignores

aapeliv: yes, done. thanks

Issue #8490, opened 2026-05-09
Backend/i18n: Support date formatting variants


tristanlabelle:
For example this notification shouldn't include the year.

<img width="589" height="179" alt="Image" src="https://github.com/user-attachments/assets/fdc5d94f-ed33-459d-8224-7e79ed95cd60" />

PR #8574, opened 2026-05-14
Backend/i18n: Support localized datetime formats

### LINKED ISSUE #7885: Backend/i18n: Support specifying short/long date formats
Once we support datetime localization, we also want to decide what format we need depending on context. E.g. for date of birth, specifying the day of the week isn't usually relevant. E.g. sometimes we want a compact numerical format.

tristanlabelle:
We only supported one localized datetime format. This isn't quite enough in practice. For example, when we send a push notification about a host request, saying "May 16, 2026" is not ideal: the year is implicit, and it'd be useful to have the day of the week.

image

This implements multiple datetime formats with an API similar to the frontend's. Python doesn't have Javascript's  convenient `Intl.DateTimeFormat`, but the Babel library exposes everything from the Unicode CLDR that we need to implement the functionality, even if it gets a little a little gnarly. Hopefully I documented it well.

I also changed the defaults a little bit to use long-form month names and omit seconds (both unless requested). This aligns with the frontend and is hopefully less surprising.

I also gave `LocalizationContext` the job of resolving the `babel.Locale`, as to not have to do it on every formatting operation.

I also leveraged the new feature to tweak datetime formats in push notifications for events and host requests. I didn't change those in emails since I'm working on email changes in parallel.

Fixes #7885, #8490

## Testing
Updated and added tests.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -46,8 +46,8 @@ class LocalizationContext:
         self.babel_locale = get_babel_locale(self.locale_list)
 
     def __setattr__(self, name: str, value: Any) -> None:
-        # Freeze after initialization. We can't use @dataclass(frozen=True) because some
-        # of our attributes are derived from others, so we can't use the default initializer.
+        # Freeze after initialization. We can't use @dataclass(frozen=True) because then
+        # we need the default initializer and some of our fields shouldn't be parameters.
         if hasattr(self, "babel_locale"):
             raise FrozenInstanceError(f"Cannot modify attribute {name}.")
         return object.__setattr__(self, name, value)
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -1,4 +1,4 @@
-from dataclasses import FrozenInstanceError, dataclass
+from dataclasses import FrozenInstanceError
 from datetime import UTC, date, datetime, time, tzinfo
 from typing import Any
 from zoneinfo import ZoneInfo
@@ -20,7 +20,6 @@ from couchers.models.users import User
 from couchers.utils import to_timezone
 
 
-@dataclass(init=False, slots=True)
 class LocalizationContext:
     """
     Specifies regional settings used for localization of strings and date/times.
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -1,5 +1,6 @@
-from dataclasses import dataclass
+from dataclasses import FrozenInstanceError, dataclass
 from datetime import UTC, date, datetime, time, tzinfo
+from typing import Any
 from zoneinfo import ZoneInfo
 
 import babel
@@ -16,10 +17,10 @@ from couchers.i18n.localize import (
     localize_timezone,
 )
 from couchers.models.users import User
-from couchers.utils import to_aware_datetime
+from couchers.utils import to_timezone
 
 
-@dataclass(init=False)
+@dataclass(init=False, slots=True)
 class LocalizationContext:
     """
     Specifies regional settings used for localization of strings and date/times.
@@ -45,6 +46,13 @@ class LocalizationContext:
         self.timezone = timezone
         self.babel_locale = get_babel_locale(self.locale_list)
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        # Freeze after initialization. We can't use @dataclass(frozen=True) because some
+        # of our attributes are derived from others, so we can't use the default initializer.
+        if hasattr(self, "babel_locale"):
+            raise FrozenInstanceError(f"Cannot modify attribute {name}.")
+        return setattr(self, name, value)
+
     @property
     def localized_timezone(self) -> str:
         return localize_timezone(self.timezone, self.babel_locale)
@@ -59,7 +67,7 @@ class LocalizationContext:
         self, value: date | datetime, *, abbrev: bool = False, with_year: bool = True, with_day_of_week: bool = False
     ) -> str:
         if isinstance(value, datetime):
-            value = value.astimezone(self.timezone).date()
+            value = to_timezone(value, self.timezone).date()
         return localize_date(
             value, self.babel_locale, abbrev=abbrev, with_year=with_year, with_day_of_week=with_day_of_week
         )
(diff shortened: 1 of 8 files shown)


app/backend/src/couchers/i18n/context.py
```diff
@@ -28,34 +29,86 @@ class LocalizationContext:
...
     timezone: tzinfo
 
+    # The Babel locale used for datetime formatting and other Unicode CLDR usage.
+    babel_locale: babel.Locale
+
+    def __init__(self, locale: str, timezone: tzinfo) -> None:
+        self.locale = locale
+        self.locale_list = [self.locale] + get_locale_fallbacks(self.locale)
+        self.timezone = timezone
+        self.babel_locale = LocalizationContext._get_babel_locale(self.locale_list)
```
WouldYouKindly: I think a slightly better way would be to keep `(frozen=True, slots=True, kw_only=True)` and add a class method that does init. This way we'd still have the benefits of immutability and slots

app/backend/src/couchers/i18n/context.py
```diff
@@ -28,34 +29,86 @@ class LocalizationContext:
...
+
+    @staticmethod
+    def _get_babel_locale(locale_list: list[str]) -> babel.Locale:
+        for locale in locale_list:
+            try:
+                return babel.Locale.parse(locale)
+            except babel.UnknownLocaleError:
+                continue
+        return babel.Locale.parse(locale_list[0])  # Will raise babel.UnknownLocaleError
```
WouldYouKindly: I wonder if it's better to raise a custom exception and include the full locale list?

app/backend/src/tests/test_localize.py
```diff
@@ -1,25 +1,36 @@
...
+
 
 def test_localize_date() -> None:
-    assert localize_date(date(2000, 1, 2), "en-US") == "Jan 2, 2000"
-    assert localize_date(date(2000, 1, 2), "fr-FR") == "2 janv. 2000"
+    assert localize_date(date(2000, 1, 2), babel_en) == "January 2, 2000"
+    assert localize_date(date(2000, 1, 2), babel_es, abbrev=True) == "2 ene 2000"
+    assert localize_date(date(2000, 1, 2), babel_fr, with_day_of_week=True) == "dimanche 2 janvier 2000"
+    assert localize_date(date(2000, 1, 2), babel_zh, with_year=False) == "1月2日"
+    assert localize_date(date(2000, 1, 2), babel_en, with_year=False, with_day_of_week=True) == "Sunday, January 2"
```
WouldYouKindly, line 19: Do you think we also need tests that use LocalizationContext? I think there are some untested bits there now

app/backend/src/couchers/i18n/context.py
tristanlabelle: If only that were possible. If you set `frozen=True`, then the only way you can initialize fields is using `object.__setattr__(self, 'field', value)`. I don't think it's worth it. We also can't freeze programmatically by calling some freeze() function.

app/backend/src/tests/test_localize.py
tristanlabelle: It's really just a bag of parameters that get forwarded to the various methods in localize. I don't think there's much to test in there, but I could move `_get_babel_locale` to `localize.py`.

app/backend/src/couchers/i18n/context.py
kalvdans: Just set all fields at once passing it to the constructor `LocalizationContext(field1=val1, field2=val2)`

app/backend/src/couchers/i18n/context.py
tristanlabelle: Done in latest

app/backend/src/couchers/i18n/context.py
tristanlabelle: Only two of the fields should be parameters, the other two are deduced from them (for caching purposes), which makes it not a dataclass because it's not plain old data.

kalvdans: aha, make functions or properties of them and decorate them with @functools.cache

WouldYouKindly: @tristanlabelle I meant like this

```python
@dataclass(frozen=True, slots=True, kw_only=True)
class LocalizationContext:
    locale: str

    # The locale code and all of its fallbacks.
    locale_list: list[str]

    # The timezone to use when formatting date-times and instants.
    timezone: tzinfo

    # The Babel locale used for datetime formatting and other Unicode CLDR usage.
    babel_locale: babel.Locale

    @classmethod
    def make(cls, locale: str, timezone: tzinfo):
        return LocalizationContext(
            locale=locale,
            locale_list=[locale] + get_locale_fallbacks(self.locale),
            timezone=timezone,
            babel_locale=get_babel_locale(self.locale_list),
        )
```

app/backend/src/tests/test_localize.py
WouldYouKindly: The bits like this are currently untested https://github.com/Couchers-org/couchers/pull/8574/changes#diff-5454b73553b6679e4ad8d10db6178b611a48f9876f9d85d94e7d6c52a59c327aR87

tristanlabelle: Added `to_timezone` and a test.

app/backend/src/couchers/i18n/context.py
tristanlabelle: Dataclasses are meant for plain old data classes that you can initialize however you want. Because `LocalizationContext` has non-parameter fields, all of the defaults don't apply anymore: `init=True, repr=True, eq=True, match_args=True`. We shouldn't have an initializer that allows breaking the invariants.

`@functools.cache` isn't the right semantics because I want eager failure in the constructor and artificially calling the property would be hard to follow. Also applying `@functools.cache` on an instance method results in a cache with a key that includes `self`, so there's a global dictionary that keeps our instances from being garbage collected.

I ended up spending an hour or two on this and in the end removed `@dataclass` and added a poor man's `__setattr__` freezing. Happy to remove it and just roll on with a regular class. It's not worth the effort to twist dataclass into applying here.

aapeliv (dismissed):
i'll trust you on this one

tristanlabelle:
Rebased to merge with the feature flags change.

Issue #8421, opened 2026-04-28
Dashboard 2.0


c-kreuz:
Description
To improve user experience and glanceability, we need to implement four new summary sections on the user dashboard. These components will provide a quick overview of upcoming travels, hosting duties, and active requests.

Functional Requirements
1. Upcoming Trips (Surfing)
Logic: Display the next 3 chronological trips with a status of "Confirmed."

Scope: Include both Public Trips (PT) and direct Host Requests.

Navigation: Add a "Show all/more" button linking to the full Trips page.

2. Upcoming Guests (Hosting)
Logic: Display the next 3 chronological confirmed surfers the user is hosting.

Navigation: Add a "Show all/more" button linking to the Hosting management page.

3. Current Host Requests (Inbox)
Logic: Display up to 3 incoming requests, ordered by the date the request was received (oldest first).

Navigation: Add a "Show all/more" button linking to the Inbox.

4. Open Surf Requests (Outbox)
Logic: Display the next 3 pending chronological requests sent by the user.

Feature Enhancement: Implement a UI highlight (e.g., a warning icon or color change) for any pending request that overlaps with an already confirmed stay.

Navigation: Add a "Show all/more" button linking to Sent Requests.

Technical Considerations
Empty States: Ensure each section has a clean "Empty State" design (e.g., "No upcoming trips. Plan one now!") if no data is present.

Responsive Design: These summaries should stack or adjust based on dashboard layout (Grid vs. List).

Caching: Consider how often these summaries refresh to avoid unnecessary API calls on every dashboard load.

PR #8581, opened 2026-05-14
Add ListMyCommunitiesDiscussions RPC to Discussions service

nabramow:
Returns paginated discussions from all region/subregion communities the user belongs to (world and macroregion communities excluded), sorted newest first.

Part of #8421 

## Testing

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/discussions.py
+++ b/app/backend/src/couchers/servicers/discussions.py
@@ -1,23 +1,14 @@
 import logging
 
 import grpc
-from sqlalchemy import or_, select
+from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from couchers.context import CouchersContext, make_background_user_context
 from couchers.db import can_moderate_node, session_scope
 from couchers.event_log import log_event
 from couchers.jobs.enqueue import queue_job
-from couchers.models import (
-    Cluster,
-    ClusterSubscription,
-    Discussion,
-    ModerationObjectType,
-    Node,
-    NodeType,
-    Thread,
-    User,
-)
+from couchers.models import Cluster, ClusterSubscription, Discussion, ModerationObjectType, Thread, User
 from couchers.models.notifications import NotificationTopicAction
 from couchers.moderation.utils import create_moderation
 from couchers.notifications.notify import notify
@@ -26,10 +17,12 @@ from couchers.proto.internal import jobs_pb2
 from couchers.servicers.api import user_model_to_pb
 from couchers.servicers.blocking import is_not_visible
 from couchers.servicers.threads import thread_to_pb
-from couchers.sql import to_bool, where_moderated_content_visible
+from couchers.sql import where_moderated_content_visible
 from couchers.utils import Timestamp_from_datetime
 
 logger = logging.getLogger(__name__)
+
+MAX_DISCUSSIONS_PAGE_SIZE = 25
 
 
 def discussion_to_pb(session: Session, discussion: Discussion, context: CouchersContext) -> discussions_pb2.Discussion:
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +146,40 @@ def GetDiscussion(
             context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "discussion_not_found")
 
         return discussion_to_pb(session, discussion, context)
+
+    def ListMyCommunitiesDiscussions(
+        self, request: discussions_pb2.ListMyCommunitiesDiscussionsReq, context: CouchersContext, session: Session
+    ) -> discussions_pb2.ListMyCommunitiesDiscussionsRes:
+        MAX_PAGE_SIZE = 25
```
aapeliv: put this at top of file please

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +146,40 @@ def GetDiscussion(
...
+        next_page_id = int(request.page_token) if request.page_token else 0
+
+        my_cluster_ids = (
+            session.execute(
+                select(ClusterSubscription.cluster_id)
+                .join(Cluster, Cluster.id == ClusterSubscription.cluster_id)
+                .join(Node, Node.id == Cluster.parent_node_id)
+                .where(ClusterSubscription.user_id == context.user_id)
+                .where(Cluster.is_official_cluster)
+                .where(Node.node_type > NodeType.macroregion)
```
aapeliv: this check should be `small_community_features_enabled` instead

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +146,40 @@ def GetDiscussion(
...
+        discussions = (
+            session.execute(
+                select(Discussion)
+                .where(Discussion.owner_cluster_id.in_(my_cluster_ids))
+                .where(or_(Discussion.id <= next_page_id, to_bool(next_page_id == 0)))
+                .order_by(Discussion.id.desc())
+                .limit(page_size + 1)
+            )
+            .scalars()
+            .all()
```
aapeliv, line 196: combine the queries into one so we don't get two sql calls please

app/backend/src/couchers/servicers/discussions.py
```diff
@@ -145,3 +146,40 @@ def GetDiscussion(
...
+        )
+
+        discussions = (
+            session.execute(
+                select(Discussion)
+                .where(Discussion.owner_cluster_id.in_(my_cluster_ids))
+                .where(or_(Discussion.id <= next_page_id, to_bool(next_page_id == 0)))
```
tristanlabelle: Would it be simpler to default `next_page_id` to 2**64 if not provided instead of zero? Then no need for the additional condition.

app/proto/discussions.proto
```diff
@@ -49,3 +53,13 @@ message CreateDiscussionReq {
 message GetDiscussionReq {
   int64 discussion_id = 1;
 }
+
+message ListMyCommunitiesDiscussionsReq {
+  uint32 page_size = 1;
```
tristanlabelle, line 58: Suggestion: Document the max in the API

```suggestion
  uint32 page_size = 1; // Maximum: 25
```

tristanlabelle (dismissed):

aapeliv (changes requested):

aapeliv:
let's also wait for mod first

aapeliv (approved):
lgtm

PR #8601, opened 2026-05-15
Devex: Add workflow to assign PR reviewers

tristanlabelle:
This workflow will replace CODEOWNERS and address two issues:

1. CODEOWNERS can only match one rule, so it can't add two sets of reviewers for changes straddling frontend and backend.
2. It's easy to forget to add reviewing teams, which hinders cross-participation in code reviews.

New GitHub workflows are untestable, so this is scoped to `github.event.pull_request.user.login == 'tristanlabelle'`. I will test it after merging, then follow up with a PR that switches us from CODEOWNERS.

## Testing
Will test after merging (currently only targeting PRs of yours truly).

## For maintainers

### DIFF
--- a/.github/autoreviewers.yml
+++ b/.github/autoreviewers.yml
@@ -1,0 +1,18 @@
+# Defines PR reviewers based on modified file paths.
+# The syntax is as the dorny/paths-filter action accepts,
+# but our filter names are a label plus a comma-delimited people/teams to assign.
+backend Couchers-org/backend:
+  - 'app/backend/**'
+  - '!app/backend/**/locales/*.json'
+  - 'app/media/**'
+web Couchers-org/web:
+  - 'app/web/**'
+  - '!app/web/**/locales/*.json'
+mobile Couchers-org/mobile:
+  - 'app/mobile/**'
+  - '!app/mobile/**/locales/*.json'
+i18n tristanlabelle:
+  - 'app/**/locales/*.json'
+  - 'app/**/i18n/**'
+docs aapeliv,nabramow:
+  - 'docs/**'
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -1,5 +1 @@
-# Only the last matching line's owners are added to the PR
-* @aapeliv
```
tristanlabelle, line 2: Self-blocking for @aapeliv 's ok (see PR description)

aapeliv:
this is okay with me, but can we write the rules in an easy file, and have the actions script read it? that way you don't need to figure out what the script does to update codeowners

tristanlabelle:
@aapeliv The easy way would be to name rules based on the team being assigned, but we'll need a new team for docs changes. Otherwise, I'd have to invent a (yml?) format, since the filters are more complex than CODEOWNERS supports, and delegate to a Python script to parse it. Let me know your preference.

tristanlabelle:
And for the first option, I believe the dorny/path-filters can read its filters from a file so that'd get us pretty close to your request.

aapeliv:
Yeah I think you could just come up with a simple syntax, it doesn't need to be complicated. And you can either wrangle the actions script into the required form, or write a python script or similar?

If you feel it's too difficult and you're not up for it, maybe we can merge as is too? I just think this is quite hard to modify in the future?

aapeliv:
Also we can make a team for translation/docs if you wish

tristanlabelle:
@aapeliv I found a pretty clean solution, and scoped the workflow to my PRs so I can test it post merge before switching from CODEOWNERS.

tristanlabelle:
@aapeliv (Ready for review)

aapeliv (approved):
looks great, thank you!

PR #8602, opened 2026-05-15
This PR fixing a UI issue in the log in page for English language

kevinortiz43:
Scenario:

User attempts to log in with the correct username and the wrong password . The warning text would say "Wrong password"

Issue: Someone can figure out if they have the right email since the system will tell them that, then they can figure out the password from there. 

The text

"Wrong password." now updated to say "Wrong Username/Email or Password." I updated the python tests as well

Testing
I tested locally

Image attached 
image

**Web frontend checklist**

## For maintainers

## Testing


app/backend/src/tests/test_account.py
```diff
@@ -282,12 +282,12 @@ def test_ChangePasswordV2_normal_wrong_password(db, fast_passwords):
         with pytest.raises(grpc.RpcError) as e:
             account.ChangePasswordV2(
                 account_pb2.ChangePasswordV2Req(
-                    old_password="wrong password",
+                    old_password="Wrong username/email or password",
```
tristanlabelle: The three API calls in this file should not be changed. They're not referencing the error message, but rather trying to show what happens when you try to change a password but enter the wrong old password or something such.

app/backend/src/couchers/servicers/account.py
tristanlabelle, line 1: The changes here are incorrect. If you're changing your password or email and need to enter your old password as confirmation, you can't have a wrong username, you're already logged in. That points to adding a new string specifically for the auth case instead of modifying the existing one.

tristanlabelle (commented):
You've covered all the right parts of the code, but also a few that shouldn't have been impacted. See comments.

app/backend/src/tests/test_account.py
kevinortiz43: I see, I fixed my error
image

app/backend/src/couchers/servicers/account.py
kevinortiz43: I create a new string specifically for auth case and updated the test_auth py test governing that scenario. Changes pushed.

tristanlabelle (approved):
Looks good! test_account changes are not necessary, but inoffensive.

PR #8607, opened 2026-05-15
fix: hydration error on sign up page

darrenvong:

A simpler proposed alternative solution to what #8541 is attempting to fix!

## Testing

**Backend checklist**

**Web frontend checklist**
- [ ] Checked Desktop, Mobile and Tablet screen sizes - no real UI changes here other than rendering logic

## For maintainers

### DIFF
--- a/app/web/features/auth/signup/Signup.tsx
+++ b/app/web/features/auth/signup/Signup.tsx
@@ -52,6 +52,8 @@ export default function Signup() {
   const error = authState.error;
 
   const [loading, setLoading] = useState(false);
+  const [isMounted, setIsMounted] = useState(false);
+  useEffect(() => setIsMounted(true), []);
 
   const urlToken = stringOrFirstString(router.query.token);
   const inviteCode = stringOrFirstString(router.query.code);
@@ -182,7 +184,7 @@ export default function Signup() {
               )}
             </Box>
           )}
-          {loading ? (
+          {!isMounted || loading ? (
             <CenteredSpinner />
           ) : (
             <SignupFormContent inviteCode={inviteCode || undefined} />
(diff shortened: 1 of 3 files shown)


WouldYouKindly (dismissed):

WouldYouKindly:
Tested locally, and it appears to fix the issue. Thanks!

app/web/features/auth/signup/Signup.tsx
```diff
@@ -182,7 +184,7 @@ export default function Signup() {
               )}
             </Box>
           )}
-          {loading ? (
+          {!isMounted || loading ? (
```
WouldYouKindly: @darrenvong just for my understanding, could you explain what's going on, please? Why do we care if the component is mounted? Is it because inviteCode is not available on the server?

darrenvong: It's mostly a workaround (although an [officially recognised ones](https://nextjs.org/docs/messages/react-hydration-error#solution-1-using-useeffect-to-run-on-the-client-only)) for the fact that `localStorage` is not available on the server side. And in this case, applied closer to the root of the problem than whatever Claude generated in your other PR.

Maybe `isClient` as they suggested in the docs would be a clearer naming as we don't really care about the mounted state of the component tbh, they just happen to represent the same state since the effects only run on client side (and so components can't be mounted until the code is run in the client/browser).

darrenvong: Refactored it to be called `isClient` and into its own hook for clarity (especially as the `isClient` pattern may be a necessary evil in some cases) - thanks for the question!

WouldYouKindly (dismissed):

darrenvong:
Ahh sorry, Claude flagged some good points and I simplified the code, so it busted the approval 😅

WouldYouKindly (approved):

PR #8631, opened 2026-05-16
Devex: Switch from CODEOWNERS to autoreviewers.yml

tristanlabelle:
Follow-up to #8601 that removes `CODEOWNERS` in favor of `autoreviewers.yml` after testing.

To support exclusions (e.g. frontend excluding locales-only changes), I had to switch dorny/path-filters to "every" mode, which requires all globs to pass. Thus we have to split off rules that should match on either of multiple paths.

I also split the workflow in more easily debuggable steps.

## Testing
Plenty of iteration in #8627

## For maintainers

### DIFF
--- a/.github/workflows/add-reviewers.yml
+++ b/.github/workflows/add-reviewers.yml
@@ -30,6 +30,9 @@ jobs:
         id: filter
         with:
           filters: .github/autoreviewers.yml
+          # Require every glob to match for a rule to match,
+          # otherwise we cannot express exclusions.
+          # e.g. 'app/web/**' + '!**/en.json' needs to be AND, not OR
           predicate-quantifier: every
 
       - name: Collect reviewers from matched rules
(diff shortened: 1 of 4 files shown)


.github/workflows/add-reviewers.yml
```diff
@@ -19,37 +19,66 @@ jobs:
...
+      - uses: actions/checkout@v4
+        with:
+          sparse-checkout: .github/autoreviewers.yml
+
+      - name: Match rules from autoreviewers.yml
+        uses: dorny/paths-filter@v4
         id: filter
         with:
           filters: .github/autoreviewers.yml
+          predicate-quantifier: every
```
WouldYouKindly, line 36: A comment with a rationale could be helpful, otherwise lgtm

tristanlabelle: Added in latest

aapeliv (approved):
lgtm; we can also get rid of codeowners

PR #8651, opened 2026-05-17
Backend: Use visibility helpers instead of raw User.is_visible

aapeliv:
Replaces direct `User.is_visible` filtering with the `couchers.sql` visibility helpers in three places, so block visibility is applied consistently (per the CLAUDE.md guidance to never use `User.is_visible` directly in queries).

- **`discussions.py`** — `generate_create_discussion_notifications` folds the raw `cluster.members.where(User.is_visible)` filter and the per-recipient `is_not_visible(...)` loop guard into a single query using `users_visible_to_each_other`.
- **`events.py`** — `get_users_to_notify_for_new_event` does the same in both branches; the now-redundant `is_not_visible` guard in `generate_event_create_notifications` is removed.
- **`get_num_references` (api.py)** — now takes a `context` argument and uses `where_users_column_visible`, so reference counts exclude references authored by users not visible to the viewer. Both callers (`user_model_to_pb`, `UserSearchV2`) updated.

Because `users_visible_to_each_other` requires both User tables joined, the dynamic `cluster.members` relationship was converted to explicit `select(User).join(ClusterSubscription, ...)` queries.

Note: `users_visible_to_each_other` excludes a shadowed creator unconditionally, whereas the old `is_not_visible` had a `creator != recipient` exemption — so a shadowed creator who is a member of their own community no longer gets a notification for their own discussion/event. This is niche and arguably more correct.

## Testing
`make format` and `make mypy` clean for the changed files. Ran the existing backend suites covering the affected code:

```
cd app/backend
uv run pytest src/tests/test_discussions.py src/tests/test_events.py src/tests/test_references.py src/tests/test_search.py
```

All 83 tests pass.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/helpers/references.py
+++ b/app/backend/src/couchers/helpers/references.py
@@ -1,10 +1,38 @@
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from sqlalchemy import select
-from sqlalchemy.orm import aliased
+from sqlalchemy.orm import InstrumentedAttribute, aliased
 from sqlalchemy.sql import Select, exists, func, or_
 
-from couchers.models import HostRequest, Reference, ReferenceType
+from couchers.models import HostRequest, Reference, ReferenceType, User
+from couchers.sql import _shadow_clause
+
+if TYPE_CHECKING:
+    from couchers.context import CouchersContext
+
+
+def where_reference_user_visible[T: tuple[Any, ...]](
+    statement: Select[T], context: CouchersContext, column: InstrumentedAttribute[int]
+) -> Select[T]:
+    """
+    Filters references based on the visibility of the user in the given column (the writer
+    or the subject of the reference).
+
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/helpers/references.py
```diff
@@ -1,10 +1,38 @@
...
-from couchers.models import HostRequest, Reference, ReferenceType
+from couchers.models import HostRequest, Reference, ReferenceType, User
+from couchers.sql import _shadow_clause
+
+if TYPE_CHECKING:
+    from couchers.context import CouchersContext
+
+
+def where_reference_user_visible[T: tuple[Any, ...]](
+    statement: Select[T], context: CouchersContext, column: InstrumentedAttribute[int]
```
tristanlabelle: Suggest: `user_id_column`. You have to read the code to understand the name otherwise.

tristanlabelle (dismissed):

PR #8652, opened 2026-05-17
Discover moderated content models dynamically

aapeliv:
Pure refactor of the Unified Moderation System plumbing, no behavior change.

Previously the set of UMS-governed models was encoded in several hand-maintained places: the seven per-object-type branches inside `moderation_state_column_visible`, the `moderationobjecttype2model` dict in `servicers/moderation.py`, and the `_ModeratedContent` union in `sql.py`. Adding a moderated type meant updating each.

This replaces them with a single source of truth on the models themselves:

- Each moderated model declares `__moderation_object_type__` alongside the existing `__moderation_author_column__`.
- `get_moderated_models()` discovers them by scanning the SQLAlchemy mapper registry and returns `{ModerationObjectType: ModeratedModel}`, where `ModeratedModel` resolves the model, object type, and author column.
- `moderation_state_column_visible` and `servicers/moderation.py` now iterate that registry instead of hardcoded lists.
- The duplicated mutual-block `exists(...)` clause is factored into a `_users_block_each_other` helper.

## Testing
- `make format` and `make mypy` pass (the two remaining mypy errors — `test_calendar_events.py` missing `ics` stubs and `fixtures/misc.py` return type — are pre-existing on `develop`).
- 142 tests pass across `test_moderation`, `test_threads`, `test_notifications`, `test_discussions`.
- To replicate: `cd app/backend && make format && make mypy && uv run pytest`.

**Backend checklist**
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no DB changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/models/moderation.py
+++ b/app/backend/src/couchers/models/moderation.py
@@ -9,9 +9,9 @@ import enum
 from dataclasses import dataclass
 from datetime import datetime
 from functools import cache
-from typing import TYPE_CHECKING, Any, Protocol
+from typing import TYPE_CHECKING, Protocol
 
-from sqlalchemy import BigInteger, DateTime, Enum, ForeignKey, Index, String, func
+from sqlalchemy import BigInteger, ColumnElement, DateTime, Enum, ForeignKey, Index, String, func
 from sqlalchemy.orm import Mapped, mapped_column, relationship
 
 from couchers.models.base import Base, moderation_seq
@@ -203,10 +203,9 @@ class ModeratedModel:
 
     object_type: ModerationObjectType
     model: type[ModeratedContent]
-    # InstrumentedAttributes/Columns resolved from the model; Any avoids descriptor-unwrapping on access
-    author_column: Any
-    object_id_column: Any
-    moderation_state_id_column: Any
+    author_column: ColumnElement[int]
+    object_id_column: ColumnElement[int]
+    moderation_state_id_column: ColumnElement[int]
 
 
 @cache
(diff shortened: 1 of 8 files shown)


app/backend/src/couchers/models/moderation.py
```diff
@@ -6,17 +6,28 @@
...
+    from couchers.models.conversations import GroupChat
+    from couchers.models.discussions import Comment, Discussion, Reply
+    from couchers.models.events import EventOccurrence
+    from couchers.models.host_requests import HostRequest
+    from couchers.models.rest import FriendRelationship
     from couchers.models.users import User
 
+    type ModeratedContentModel = type[
+        HostRequest | GroupChat | FriendRelationship | EventOccurrence | Comment | Reply | Discussion
+    ]
```
tristanlabelle: Have you tried using `typing.Protocol`? My proof-of-concept works great and avoids the union and `getattr(cls, cls.__moderation_author_column__),`.

```py
from enum import Enum, auto
from typing import ClassVar, Protocol

class ModerationObjectType(Enum):
    event_occurrence = auto()

class ModeratedContent(Protocol):
    __moderation_author_column__: ClassVar[str]
    __moderation_object_type__: ClassVar[ModerationObjectType]
    # id: Mapped[int]

class EventOccurrence:
    __moderation_author_column__ = "creator_user_id"
    __moderation_object_type__ = ModerationObjectType.event_occurrence


moderated_content_type: type[ModeratedContent] = EventOccurrence
print(moderated_content_type.__moderation_author_column__)
print(moderated_content_type.__moderation_object_type__)

tristanlabelle (dismissed):
Nice, much better. I think we can still go one step further with `typing.Protocol`.

app/backend/src/couchers/models/moderation.py
aapeliv: awesome, implemented, thanks! typing was a bit inflexible, otherwise worked

PR #8665, opened 2026-05-17
Backend/emails: Support custom SMTP headers for attachments

tristanlabelle:
ics attachments are trickier than I thought, one of the ics field (`method`) also needs to be replicated as an SMTP header on the attachment for correct support by email clients. `EmailAttachment` wasn't flexible enough to represent this, so I switched it over to specifying the `Content-Type` and `Content-Disposition` headers directly, which is as generic as it gets. This should improve calendar event compatibility and behavior in email clients.

There's some extra dance to support unquoted `method=XXX` in the SMTP header (not the default Python behavior) since Gemini said that was preferred for backcompat with old email clients.

## Testing
Ran locally, accepted host requests and downloaded emails from MailDev while iterating until I got this desired form for the attachment:

```
--===============2710363521157735036==
Content-Type: text/calendar; charset="utf-8"; method=PUBLISH
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename="host_request.ics"
MIME-Version: 1.0
```

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/proto/internal/jobs.proto
+++ b/app/backend/proto/internal/jobs.proto
@@ -14,15 +14,14 @@ message SendEmailPayload {
   string list_unsubscribe_header = 7;
   // source data as to where this email came from
   string source_data = 8;
-  repeated EmailAttachment attachments = 9;
+  reserved 9; // Previous "attachments" field, which had mime_type+filename subfields.
+  repeated EmailAttachmentV2 attachments = 10;
 }
 
-message EmailAttachment {
-  reserved 1, 2;
-  reserved "filename", "mime_type";
-  bytes data = 3;
-  string content_disposition = 4; // The Content-Disposition header, including parameters
-  string content_type = 5; // The Content-Type header, including parameters
+message EmailAttachmentV2 {
+  bytes data = 1;
+  string content_disposition = 2; // The Content-Disposition header, including parameters
+  string content_type = 3; // The Content-Type header, including parameters
 }
 
 message HandleNotificationPayload {
--- a/app/backend/proto/internal/jobs.proto
+++ b/app/backend/proto/internal/jobs.proto
@@ -18,9 +18,11 @@ message SendEmailPayload {
 }
 
 message EmailAttachment {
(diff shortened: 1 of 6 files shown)


app/backend/src/couchers/email/smtp.py
```diff
@@ -58,15 +57,19 @@ def send_smtp_email(payload: jobs_pb2.SendEmailPayload) -> Email:
...
-                attachment.data, maintype=mime_maintype, subtype=mime_subtype, filename=attachment.filename
-            )
+            # Versioning: ignore older SendEmailPayload that did not specify headers.
+            if not attachment.HasField("content_disposition") or not attachment.HasField("content_type"):
+                continue
+
+            msg.add_attachment(attachment.data, maintype="application", subtype="octet-stream")
+            attachment_part: MIMEPart = msg.get_payload()[-1]
+            attachment_part["Content-Disposition"] = attachment.content_disposition
+            attachment_part["Content-Type"] = attachment.content_type
```
tristanlabelle: TODO: Test with MailDev

app/backend/proto/internal/jobs.proto
```diff
@@ -18,9 +18,11 @@ message SendEmailPayload {
 }
 
 message EmailAttachment {
```
aapeliv: this is fine, but could also do EmailAttachmentV2 since it's basically a completely different proto

aapeliv (dismissed):

app/backend/proto/internal/jobs.proto
tristanlabelle: Good call. I updated to `EmailAttachmentV2` as you suggest since it removes the need to handle backcompat. In fact by rules of protobuf we could keep the name `EmailAttachment` since I retired the field that was referencing its previous definition. I'll keep the V2 for clarity in this PR but might rename it back in a future one.

tristanlabelle:
Rebased to resolve conflicts (adjacent import lines had changed in `render_email.py`)

PR #8669, opened 2026-05-18
Frontend/i18n: Better app store badge string keys

tristanlabelle:
It's important to provide context on the string keys because in this case the translation shouldn't be what translators come up with from the English string, but rather Google's official translations for those.

Added an explanatory string in weblate.

Filed #8670 to follow up on supporting localized badges.

## Testing
N/A

**Web frontend checklist**

## For maintainers


app/web/resources/locales/en.json
```diff
@@ -6,8 +6,8 @@
   "retract": "Retract",
   "couchers": "Couchers.org",
   "learn_more": "Learn more",
-  "download_on_app_store": "Download on the App Store",
-  "get_it_on_google_play": "Get it on Google Play",
```
tristanlabelle, line 10: Self-TODO: Post merge, re-add Weblate explanation to use official translations

c-kreuz (approved):

PR #8679, opened 2026-05-18
feat: dismissible reminders

### LINKED ISSUE #8430: Add dismiss to dashboard reminders using cookies
- Add a little X to the top right of the yellow `ReminderItem` 
- When the user clicks it, it should store a cookie for that specific reminder for one week
- When we do the initial query, we should check the cookies and only display the reminders that have no been dismissed less than a week ago.

darrenvong:

Closes #8430

I went with a localStorage implementation in the end as we already have an existing `usePersistedStorage` hook, which makes reading/writing values a lot easier than manual cookie manipulation.

The return type on it was bothering me though, as values returned can never be `undefined` unless you call `clearValue` (the 3rd destructured value of the tuple). Or explicitly pass `undefined` as a default value, but in that case, that would be captured by the generic T type. This wrong typing made the consuming code more complex - you either need to write fallbacks that are unnecessary, extra code to check for undefinedness, or lie with assertions... and none of those feel great!

So I went ahead and refactored it into a new hook called `useClearablePersistedStorage`, which has the same implementation before, and `usePersistedStorage` now wraps that except with the `undefined` typing on the value removed. I figured it's fine as part of this since the change is relatively small and only a handful of tests and a single consumer got affected.

## Testing

Have two profiles ready. One should have an incomplete profile, the other should be complete so you can send host requests to the incomplete profile.

Once you sent a host request to the incomplete profile, you should see two reminders in the account's dashboard.

Click on the new X button on either of them to dismiss it.

Refresh the page and check that only the reminder you didn't dismiss is shown.

https://github.com/user-attachments/assets/6474952c-3a03-48c5-ae18-bc4b01d4c4f7

https://github.com/user-attachments/assets/bb6eade1-edbb-4bc8-9614-1bd5159b2ba3

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/dashboard/ReminderCarousel.tsx
+++ b/app/web/features/dashboard/ReminderCarousel.tsx
@@ -7,7 +7,8 @@ import { useQuery } from "@tanstack/react-query";
 import Alert from "components/Alert";
 import IconButton from "components/IconButton";
 import { RpcError } from "grpc-web";
-import { GetRemindersRes } from "proto/account_pb";
+import { usePersistedState } from "platform/usePersistedState";
+import { GetRemindersRes, Reminder } from "proto/account_pb";
 import { useEffect, useRef, useState } from "react";
 import { service } from "service";
 
@@ -18,6 +19,54 @@ import ReminderItem from "./ReminderItem";
 const CARD_WIDTH_DESKTOP = 280;
 const CARD_WIDTH_MOBILE = 240;
 const CARD_GAP = 16;
+
+const ONE_WEEK_MS = 7 * 24 * 60 * 60 * 1000;
+
+interface ReminderWithId {
+  id: string;
+  reminder: Reminder.AsObject;
+}
+
+function getReminderId(reminder: Reminder.AsObject): string | null {
+  if (reminder.respondToHostRequestReminder?.hostRequestId != null) {
+    return `respond_host_request:${reminder.respondToHostRequestReminder.hostRequestId}`;
+  }
+  if (reminder.writeReferenceReminder?.hostRequestId != null) {
+    return `write_reference:${reminder.writeReferenceReminder.hostRequestId}`;
+  }
+  if (reminder.completeProfileReminder) {
+    return "complete_profile";
+  }
+  if (reminder.completeVerificationReminder) {
+    return "complete_verification";
+  }
+  return null;
+}
+
+function pruneStaleEntries(
+  dismissedReminders: Record<string, number>,
+  currentDismissalTime: number,
+): Record<string, number> {
+  const out: Record<string, number> = {};
+  for (const [reminderId, dismissalTime] of Object.entries(
+    dismissedReminders,
+  )) {
+    if (currentDismissalTime - dismissalTime < ONE_WEEK_MS) {
+      out[reminderId] = dismissalTime;
+    }
+  }
+  return out;
+}
+
+function isStillDismissed(
+  dismissedReminders: Record<string, number>,
+  key: string,
+): boolean {
+  const now = Date.now();
+  const ts = dismissedReminders[key];
+  if (ts === undefined) return false;
+  return now - ts < ONE_WEEK_MS;
+}
 
 const StyledContainer = styled(Box)(({ theme }) => ({
   display: "flex",
@@ -64,6 +113,10 @@ export default function ReminderCarousel() {
     queryFn: () => service.account.getReminders(),
   });
 
+  const [dismissedReminders, setDismissedReminders] = usePersistedState<
+    Record<string, number>
+  >("dismissedReminders", {});
+
   const scrollerRef = useRef<HTMLDivElement>(null);
   const isMobile = useMediaQuery(theme.breakpoints.down("md"));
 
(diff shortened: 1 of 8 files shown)


app/web/platform/usePersistedState.ts
```diff
@@ -6,17 +6,23 @@ import {
...
-export function usePersistedState<T>(
+export function useClearablePersistedState<T>(
   key: string,
   defaultValue: T,
   storage: StorageType = "localStorage",
 ): [T | undefined, (value: T) => void, () => void] {
   // in ssr, window doesn't exist, just use default
   const saved =
     typeof window !== "undefined" ? window[storage].getItem(key) : null;
-  const [_state, _setState] = useState<T | undefined>(
```
darrenvong, line 17: Nice flag from Claude that if the JSON in local storage is corrupted for whatever reasons, this could blow up the page (`JSON.parse` failing and throws an error) and so I thought worth fixing, even though I'm not sure if it's happened to users in practice much!

app/web/features/dashboard/ReminderCarousel.tsx
```diff
@@ -19,6 +20,47 @@ const CARD_WIDTH_DESKTOP = 280;
...
+    return `write_reference:${reminder.writeReferenceReminder.hostRequestId}`;
+  }
+  if (reminder.completeProfileReminder) {
+    return "complete_profile";
+  }
+  if (reminder.completeVerificationReminder) {
+    return "complete_verification";
+  }
+  return null;
+}
```
nabramow, line 44: Nit - might be slightly less verbose as a switch statement assuming we will add more reminders but feel free to have a different style preference ;).

app/web/features/dashboard/ReminderCarousel.tsx
```diff
@@ -19,6 +20,47 @@ const CARD_WIDTH_DESKTOP = 280;
...
+
+function pruneStaleEntries(
+  dismissedReminders: Record<string, number>,
+  now: number,
+): Record<string, number> {
+  const out: Record<string, number> = {};
+  for (const [k, ts] of Object.entries(dismissedReminders)) {
```
nabramow: Nit - would be a bit more readable if the variables said what they were instead of `k` and `ts`.

app/web/features/dashboard/ReminderCarousel.tsx
```diff
@@ -58,12 +100,21 @@ const StyledArrow = styled(IconButton)({
   },
 });
 
+interface ReminderWithId {
+  id: string;
+  reminder: Reminder.AsObject;
+}
+
```
nabramow: Nit - could we put the types at the top of the file below the imports?

app/web/features/dashboard/ReminderCarousel.tsx
```diff
@@ -58,12 +100,21 @@ const StyledArrow = styled(IconButton)({
...
 export default function ReminderCarousel() {
   const { data, error } = useQuery<GetRemindersRes.AsObject, RpcError>({
     queryKey: [remindersKey],
     queryFn: () => service.account.getReminders(),
   });
 
+  const [dismissedReminders, setDismissedReminders] = usePersistedState<
+    Record<string, number>
+  >("dismissedReminders", {});
+
```
nabramow, line 119: Does this need to be the useClearableState one now?

app/web/platform/usePersistedState.ts
```diff
@@ -37,6 +43,19 @@ export function usePersistedState<T>(
...
+  storage: StorageType = "localStorage",
+) {
+  const [state, setState] = useClearablePersistedState(
+    key,
+    defaultValue,
+    storage,
+  );
+  return [state as T, setState] as const;
+}
+
```
nabramow, line 62: If there is a use case for both can you add a short comment above each explaining when to use which? Or if we should eventually refactor the other away add a `@TODO` in a comment so we know to deal with it eventually

app/web/features/dashboard/ReminderCarousel.tsx
darrenvong: I'm not sure there's a clean way to switch on it though 🤔  since the proto type gets generated here is a bit weird, it's an object with a bag of properties (that may or may not be there), rather than a single variable you can check against different values on (where a switch would work nicely on!)

app/web/features/dashboard/ReminderCarousel.tsx
darrenvong: No, since even with all reminders cleared, we'd still want to default to an empty object I think rather than `undefined`

app/web/platform/usePersistedState.ts
darrenvong: Yeah since there's only one use case at the moment that needs the clearable one in chat messages once you sent the message (otherwise it gets auto-saved into local storage as a draft I believe), I reckon it's probably better to refactor it away eventually to avoid confusion. Will add a comment - already done some tidies in this PR but didn't want to make the refactors bigger than the feature itself 😂

nabramow (approved):
Okay tested it out with two accounts, looks good to me!! Merge away 🚀!

PR #8683, opened 2026-05-18
Revise onboarding message for user profile completion

c-kreuz:
Updated text in onboarding message for clarity and engagement.


app/backend/templates/v2/onboarding2.txt
tristanlabelle, line 1: The strings exist in the mjml and html too, they all need to be updated. Or leave it to when I port them to the new email system.

aapeliv: fixed

aapeliv (approved):

PR #8693, opened 2026-05-19
Backend: Remove AssertionError in non-test code.

tristanlabelle:
Following up on a past review comment.

## Testing
N/A

**Backend checklist**

## For maintainers


aapeliv (dismissed):

tristanlabelle:
@aapeliv Iterated to differentiate `ValueError` (bad input) from `NotImplementedError` (unimplemented case). Ready for rereview

kalvdans (approved):
Since asserts can be disabled with -O, I agree it's better to explicitly raise an error.

I like the assert syntax, wished it could be used reliably... but here we are.

PR #8696, opened 2026-05-19
Backend: Ensure all datetimes are timezone-aware

tristanlabelle:
We shouldn't use Python naive datetimes, aka non-timezone-aware datetimes, to represent an instant in time, since that depends on the server's system timezone which should be irrelevant. Remove the few instances of them and throw if we encounter them.

## Testing
CI ran tests

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -67,6 +67,9 @@ def is_valid_email(field: str) -> bool:
 
 
 def Timestamp_from_datetime(dt: datetime) -> Timestamp:
+    if dt.tzinfo is None:
+        raise ValueError("Cannot convert a naive datetime to a timestamp.")
+
     pb_ts = Timestamp()
     pb_ts.FromDatetime(dt)
     return pb_ts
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/utils.py
```diff
@@ -67,6 +67,9 @@ def is_valid_email(field: str) -> bool:
 
 
 def Timestamp_from_datetime(dt: datetime) -> Timestamp:
+    if dt.tzinfo is None:
+        raise ValueError("Cannot convert a naive datetime to a timestamp.")
+
```
kalvdans, line 72: Nit: naïve means childish, maybe you meant native?

```suggestion
    if dt.tzinfo is None:
        raise ValueError("Cannot convert a timezone-less datetime to a timestamp.")

```

kalvdans (approved):
+1 for forcing timezone on all datetimes!

app/backend/src/couchers/utils.py
tristanlabelle: Python datetimes are actually childish: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

PR #8697, opened 2026-05-19
Backend/emails: Port more emails to new system

tristanlabelle:
Continuing upon #8456, ports emails for the following notifications to the new email system which supports localization and unifies textual and html emails: `friend_request`, `account_deletion`, `discussion`, `thread`.

Adds support for markdown in quoted email blocks, since user text can contain it.

## Testing
Ran `uv run dump-emails`. Examples:

image

(profile images below is broken because this is fake data, including a fake profile pic url)

image

image

**Backend checklist**

## For maintainers


aapeliv (approved):
could you please check with your dump system that they match (and/or every change is OK) between this and the prod versions? also for the first batch?

tristanlabelle:
@aapeliv I opened all new html's to validate and diffed several htmls+txts with our previous templates. The differences are all minor, expected and ok:

- Some text was inconsistent ("Hello" vs "hi", presence of exclamation marks, "If this wasn't you" vs "If you did not initiate this action")
- The html structure for plaintext `<mj-section>` now looks like two `<mj-section>` with their `<mj-text>` instead of one `<mj-section>` with two `<mj-text>`. Visually it's the same.
- Plaintext bodies show links a little differently (`<button-caption>: <url>`)
- Plaintext bodies show people a little differently (`<name>, <age>, <city>`)
- Emails in plaintext bodies are now between <>
- Fix bogus `</mj-text>` in plaintext email. 

Once I'm done porting, I'll dump all htmls and provide them in a zip on Slack so we have an opportunity to tweak any wording before I flip the weblate switch on and expose them to translation.

PR #8729, opened 2026-05-21
Mobile: relay gRPC-web API calls through native code

aapeliv:
Routes the mobile app's gRPC-web API calls through native code instead of the WebView's own HTTP transport.

A new gRPC-web unary interceptor (`nativeApiRelay.ts`) is wired in as the innermost interceptor in the web client. When the page is running inside the native mobile WebView, it bypasses the built-in transport: it serializes and gRPC-web-frames the request, relays it to native over `postMessage` with a correlation id, then unframes/deserializes the reply and surfaces non-OK `grpc-status` as an `RpcError`. On regular web it's a pass-through. The native side (`apiRelay.ts`) performs the real request via `XMLHttpRequest` (binary via `arraybuffer`) and only ever relays to the configured API host.

The relay activates only when the native app advertises support via the injected `apiRelay` flag, so a web-only deploy can never brick already-installed native builds (they keep using their own transport).

## Testing
- Web: `yarn typecheck` (clean for these files), `yarn test service/nativeApiRelay.test.ts` — 5 tests pass (framing helpers, pass-through, relay round-trip, non-OK status → RpcError)
- Mobile: `tsc --noEmit` clean, `jest tests/service/apiRelay.test.ts` — 3 base64 tests pass
- Lint clean for all changed files on both sides

Not yet done: an end-to-end smoke test of a relayed call against a running backend through a real device WebView. The protocol logic is unit-covered, but the live native fetch path has not been exercised on-device.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/mobile/components/WebEmbed.tsx
+++ b/app/mobile/components/WebEmbed.tsx
@@ -24,6 +24,7 @@ import { useAuthContext } from "@/features/auth/AuthContext";
 import { useImagePicker } from "@/hooks/useImagePicker";
 import { useWebNavigation } from "@/hooks/useWebNavigation";
 import errorGraphic from "@/resources/404graphic.png";
+import { relayApiRequest } from "@/service/apiRelay";
 import client from "@/service/client";
 import { dispatchEscapeRef, lastLoginTimeRef } from "@/state/webViewState";
 import { theme } from "@/theme";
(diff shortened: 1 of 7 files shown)


aapeliv:
Turns out this is not the right thing to do. It will leave a bunch of things broken, because not everything goes through gRPC.

PR #8752, opened 2026-05-22
Backend: Update notification is_seen when viewing message or host request

nabramow:
Right now the notification feed item only gets marked read if they click it via the notification feed. We want the unread badge to also be updated by viewing the messages themselves via the messages section.

## Testing

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/conversations.py
+++ b/app/backend/src/couchers/servicers/conversations.py
@@ -5,7 +5,7 @@ from typing import Any, cast
 
 import grpc
 from google.protobuf import empty_pb2
-from sqlalchemy import select
+from sqlalchemy import select, update
 from sqlalchemy.orm import Session, contains_eager
 from sqlalchemy.sql import func, not_, or_
 
@@ -24,6 +24,7 @@ from couchers.models import (
     Message,
     MessageType,
     ModerationObjectType,
+    Notification,
     RateLimitAction,
     User,
 )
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/servicers/conversations.py
```diff
@@ -613,6 +614,13 @@ def MarkLastSeenGroupChat(
 
         subscription.last_seen_message_id = request.last_seen_message_id
 
+        session.execute(
+            update(Notification)
+            .values(is_seen=True)
+            .where(Notification.user_id == context.user_id)
+            .where(Notification.key == str(request.group_chat_id))
```
aapeliv: you need to also scope it to the right notification type

app/backend/src/couchers/servicers/requests.py
```diff
@@ -939,6 +940,13 @@ def MarkLastSeenHostRequest(
                 context.abort_with_error_code(grpc.StatusCode.FAILED_PRECONDITION, "cant_unsee_messages")
             host_request.recipient_last_seen_message_id = request.last_seen_message_id
 
+        session.execute(
+            update(Notification)
+            .values(is_seen=True)
+            .where(Notification.user_id == context.user_id)
+            .where(Notification.key == str(host_request.conversation_id))
+        )
```
aapeliv, line 956: ditto

aapeliv (changes requested):

aapeliv:
would also favor a nice API for this instead of raw DB updates

app/backend/src/couchers/servicers/conversations.py
nabramow: Oh shit good call

app/backend/src/couchers/servicers/requests.py
nabramow: Both done!

nabramow:
> would also favor a nice API for this instead of raw DB updates

Done! Looks way more readable now thanks for the suggestion!

aapeliv (approved):

PR #8756, opened 2026-05-22
Backend/emails: from_notification factory methods

### LINKED ISSUE #8712: Move protobuf notif to Email to static methods
E.g. `HostRequestStatusChanged.from_notification_pb(data: notification_pb2.xxx) -> Self`.

This will enable more static type validation and keep `render_email.py` into a simple switch statement.

tristanlabelle:
Moves converting the notification data into constructing an email object to static methods of each email class. Having separate methods keeps the match statement cleaner and allows typing each notification data to have mypy validate attribute accesses.

I'll rebase as needed to handle conflicts with my other PRs.

Fixes #8712

## Testing
`uv run dump-emails` still works and emails are still fine.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/emails.py
+++ b/app/backend/src/couchers/email/emails.py
@@ -95,6 +95,13 @@ class AccountDeletionStartedEmail(EmailBase):
         builder.security_warning_para()
 
     @classmethod
+    def from_notification(cls, data: notification_data_pb2.AccountDeletionStart, *, user_name: str) -> Self:
+        return cls(
+            user_name=user_name,
+            deletion_link=urls.delete_account_link(account_deletion_token=data.deletion_token),
+        )
+
+    @classmethod
     def dummy_data(cls) -> AccountDeletionStartedEmail:
         return AccountDeletionStartedEmail(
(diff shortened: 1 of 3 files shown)


WouldYouKindly:
> and allows typing each notification data to have mypy validate attribute accesses.

could you elaborate? I don't see anything mypy related in the PR?

UPD: Ah! You mean the methods accept a typed object, not Any. But you're still passing Any inside, so I don't think it makes the code much safer :)

WouldYouKindly (approved):

tristanlabelle:
@WouldYouKindly , yeah `data = notification.topic_action.data_type.FromString(notification.data)  # type: ignore[attr-defined]` is of type `Any`, so none of the `data.<attr>` attributes were checkable by mypy. Since all the `from_notification` methods are strongly typed, those accesses are now checkable. Indeed there's still an implicit `Any -> T` conversion and I could eliminate it with `notification_pb2.Foo.FromString` calls, but that's more verbose and the attr access checking is where I often had bugs in my recent PRs.

PR #8757, opened 2026-05-23
Backend: Hide moderator badge behind a feature flag

aapeliv:
Hides the moderator profile badge from the UI, gated behind a GrowthBook feature flag so it can be rolled out and rolled back without a deploy.

This adds a generic `hidden` flag to badge definitions. A hidden badge is still awarded to users and stored in `user_badges` (no data churn), but is:
- excluded from the `GetBadges` catalog (so it disappears from the `/badges` page),
- stripped from every user's `badges` list on `GetUser`, public profiles, and admin user-details, and
- not announced via add/remove notifications.

The `moderator` badge is marked `hidden`. Whether the flag is honored is controlled by a new global gate `hide_moderator_badge` (via `check_global_gate`, which evaluates against a stable synthetic identity so it works in anonymous/`AUTH_LEVEL_OPEN` and background-job contexts where there is no user). The gate is **off by default**, so behavior is unchanged in production until the flag is turned on; turning it off again is an instant rollback.

To roll out: create a boolean feature/gate named `hide_moderator_badge` in GrowthBook (default off) and turn it on when ready.

## Testing
- Added `test_hidden_badge_is_awarded_but_not_shown` (gate on, the default in tests): the badge row is created but absent from `GetUser` and no notification fires.
- Added `test_hidden_badge_is_shown_when_rollout_gate_off` (disables experimentation via `monkeypatch`): the badge is shown on the profile and the add is notified — verifying the pre-rollout/rollback path.
- Updated `test_GetBadges` to reflect the moderator badge being excluded from the catalog when the gate is on.
- Ran the badge-touching suites (`test_api`, `test_resources`, `test_bg_jobs`, `test_donations`, `test_public`, `test_admin`) — all pass. `make format` clean; `make mypy` has only 2 pre-existing errors in untouched files.

**Backend checklist**
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no DB changes)

**Web frontend checklist**
- [x] Added tests where relevant (defensive guard only — skips badge ids not in the catalog so a hidden id can't crash the profile)

## For maintainers

### DIFF
--- a/app/backend/resources/badges.json
+++ b/app/backend/resources/badges.json
@@ -18,7 +18,8 @@
     {
       "id": "moderator",
       "color": "#c74f5b",
-      "admin_editable": false
+      "admin_editable": false,
+      "flag": "show_moderator_badge"
     },
     {
       "id": "volunteer",
--- a/app/backend/src/couchers/context.py
+++ b/app/backend/src/couchers/context.py
@@ -3,7 +3,6 @@ from typing import TYPE_CHECKING, NoReturn, cast
 import grpc
 
 from couchers import experimentation
-from couchers.config import config
 from couchers.i18n import LocalizationContext
 
 if TYPE_CHECKING:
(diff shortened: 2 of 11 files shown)


nabramow:
I might be misunderstanding, but it looks like you're using this in the badges update query but not the get query?

PR #8785, opened 2026-05-24
Handle edge case dates different days surfer vs host

### LINKED ISSUE #8741: Great job on the app but Request wouldn’t submit for me
Subject: Great job on the app but Request wouldn’t submit for me
Description:
I was trying to send a Request in the app, wrote it all out and then then submit button wouldn’t work

Results:
I logged in instead in Safari on my iPhone copied in the same previous message with same arrival departure dates and the submit button worked 

Backend version: develop-6e933e2b
Frontend version: develop-6e933e2b
User Agent: Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 CouchersNative/1.1.19 (ios; build unknown; c2a3c885)
Screen resolution: 428x796
Page: https://couchers.org/profile
User: [@db](https://couchers.org/user/db) (10172)

nabramow:
Backend was throwing an error if user picked `today` and that was already past in the host's timezone. This PR handles this case with better messaging on the frontend.

* Frontend: validate arrival date against the host's timezone so past dates are blocked before submission
* Frontend: Move form reset to after success so data is preserved on error
* Tests: cover both timezone directions and the error-preservation case
* Dev: add a UTC+14 dummy user and fake timezone polygon for local testing

Closes #8741 

## Testing
 - Run the backend locally: `docker compose down && rm -rf data/postgres && docker compose up` from app/
- Run the web frontend locally: `cp .env.localdev .env.development` then `yarn start` from app/web/
- Log in as aapeli, go to map search and search keyword user "kiri"
- Try sending a host request with today as the arrival date — the datepicker should block it
- Manually change the date to today - should show error
- Try a future date — the request should go through
- To test error preservation: set your network to offline in devtools, fill out the form and submit — your message should survive the error

## For maintainers


app/backend/src/data/dummy_users.json
```diff
@@ -263,6 +263,35 @@
...
+      "languages": [
+        ["eng", "fluent"]
+      ],
+      "occupation": "Marine Biologist",
+      "about_me": "Living on the world's most ahead-of-time island — UTC+14 means I'm always first to see the new day!",
+      "about_place": "A spare room on the most easterly timezone on Earth.",
+      "regions_visited": ["AUS", "NZL", "USA"],
+      "regions_lived": ["AUS"],
+      "hosting_status": "HOSTING_STATUS_CAN_HOST"
+    },
```
nabramow, line 294: Adding this as I think it's useful enough to permanently have in the local dummy users for testing extreme timezone differences.

WouldYouKindly:
@nabramow just for my understanding: why don't we want to send the requests in the host timezone? And make it clear in host request date picker.

nabramow:
> @nabramow just for my understanding: why don't we want to send the requests in the host timezone? And make it clear in host request date picker.

@WouldYouKindly That is what I am doing! Right now for example if I, a surfer in NYC, sent a host request for `today` to a host in Australia, and Australia it's already tomorrow, the backend will throw an error because today has past.

That is correct, but it was clearing the host request form (so user loses what they wrote) and throwing an error that it's in the past and users were getting confused.

We don't store timezone in the host request data (everything is UTC in the database) so we always need to compute it on the fly from their location. I thought best way to avoid this is to make the validation on the frontend match the backend, with a clearer error message. It will also block the date from the calendar to begin with if `today` has passed for the host, so they can't even choose it.

app/backend/resources/timezone_areas.sql-fake
```diff
@@ -3,4 +3,4 @@ INSERT INTO "public"."timezone_areas" ("geom" , "tzid") VALUES ('0106000020E6100
...
 INSERT INTO "public"."timezone_areas" ("geom" , "tzid") VALUES ('0106000020E61000000100000001030000000100000006000000E7EE7FC8AB4563404427DF43430E3CC05E4783784F9661400CC6EA4031BD3CC0FA8A433E499A6140A802CCFE9D7643C058C4AE924043624041F102AB8EB146C09AC37C83F6156340BEA64EB85F9E43C0E7EE7FC8AB4563404427DF43430E3CC0', 'Australia/Sydney');
-INSERT INTO "public"."timezone_areas" ("geom" , "tzid") VALUES ('0106000020E6100000010000000103000000010000000500000000000000008066C0000000000080564000000000008066C000000000008056C0000000000080664000000000008056C0DA773767C2856640000000000080564000000000008066C00000000000805640', 'Etc/UTC');
+INSERT INTO "public"."timezone_areas" ("geom" , "tzid") VALUES ('0106000020E610000001000000010300000001000000050000000000000000C063C0000000000000F83F00000000009063C0000000000000F83F00000000009063C000000000000004400000000000C063C000000000000004400000000000C063C0000000000000F83F', 'Pacific/Kiritimati');
```
aapeliv, line 6: please don't get rid of UTC!

app/backend/src/data/dummy_users.json
aapeliv: Kiritimati Island is beautiful, would recommend

aapeliv (changes requested):
backend looks good bar one minor issue with the UTC tz

app/web/features/profile/view/NewHostRequest.tsx
```diff
@@ -233,10 +234,16 @@ export default function NewHostRequest({
 
   const onSubmit = handleSubmit((data) => {
     mutate(data);
-    reset();
   });
 
+  const hostToday = user.timezone
+    ? dayjs().tz(user.timezone).startOf("day")
+    : dayjs().startOf("day");
```
tristanlabelle, line 241: We might want to be resilient to the user's machine not knowing the same set of timezone names as the server's machine (try/catch and default to browser tz). Otherwise this page is probably going to error out when loading.

tristanlabelle (commented):
Looks good apart from removing UTC

app/web/features/profile/view/NewHostRequest.tsx
aapeliv: I believe these would be the IANA ones and almost everybody should be standardized on them. The exact rules may differ based on versions, but I think if they use IANA, it should be safe enough to assume the names all exist at least

tristanlabelle: IANA database version discrepancies is what I had in mind, especially because the website I used didn't find Pacific/Kirisomething, so not everything is in sync. I also think we don't validate timezone strings we get from the frontend on the backend yet (on my list).

nabramow:
@aapeliv @tristanlabelle Okay put UTC back and moved the other new one above it. Is UTC polygon supposed to be the whole world?

tristanlabelle:
Hmm good point, I don't know if any polygon makes sense for UTC.

aapeliv:
Yeah it’s supposed to be the whole world in the background. Backend handles it.

PR #8809, opened 2026-05-25
Add runbook for release process

nabramow:
Adds a checklist for all the tasks that need to be completed each release so it doesn't only live in my head.


docs/release-checklist.md
```diff
@@ -0,0 +1,29 @@
...
+
+Here are all the steps that need tp be done for each official Couchers release.
+
+## Before you start
+[] Do a post in the #couchops channel for any operations, translation, merch, etc. accomplishments that should also be included in the release blog post.
+[] Check if there are any recruitment needs that should be shouted out in the "Volunteer Needs at Couchers" section of the blog post.
+[] This could also be a good time to add/remove positions from our volunteer page if they are no longer relevant
+
+# Dev Tasks
+[] Increment version number in web `package.json` and `app/version`. Usually we bump in a minor version for a quarterly release.
```
aapeliv: the package.json version is unused afaik; we could drop it and set it to something dummy so we don't hav eto worry about it

docs/release-checklist.md
```diff
@@ -0,0 +1,29 @@
...
+[] Get the list of completed PRs from Aapeli and separate into categories of features
```
aapeliv: will try to make this a script others can run too

docs/release-checklist.md
```diff
@@ -0,0 +1,29 @@
...
+[] Create markdown file blog post for new release in `app/web/markdown/blog` within the correct year, month and day folders. Either write it in there, or if a non-dev has written the post, paste it in and fix up the markdown formatting.
+[] Add any CouchOps accomplishments that are not included in the PRs to the post. Update the "Volunteer Needs at Couchers" section as needed.
+[] Take screenshots of any new relevant features. Make sure no sensitive user data is in the screenshots and use staging if necessary. Add these to the markdown blog post, look at previous posts for how to do it and center them.
+[] Create a new entry in `app/web/dashboardNews.json` at the top
+[] Add or replace current dashboard news widget in `app/web/features/dashboard/Dashboard.tsx`
+[] Update the the activeness probe with the new release blog post in `app/backend/templates/v2/activeness_probe.txt` and `app/backend/templates/v2/activeness_probe.mjml`. Follow the instructions in `app/backend/templates/v2/readme.md` to regenerate the html templates.
```
aapeliv: templates should now be auto-generated, so just need to verify that CI built them correctly

docs/release-checklist.md
```diff
@@ -0,0 +1,29 @@
...
+[] Merge blog post PR with the above changes and deploy via `ops.couchershq.org`
+[] Send blog post notification
+
+# Non-dev tasks
+[] Copy the markdown from the completed blog post into a new Listmonk newsletter and add the `Dear Hi {{ .Subscriber.FirstName }}` and the signature
```
aapeliv: we should sync removed users with listmonk. it should be done automatically

aapeliv (dismissed):
lgtm, some comments on the process

docs/release-checklist.md
nabramow: I don't think it's so difficult to keep it up though? And for devs that's also the standard place to check the version?

docs/release-checklist.md
nabramow: @aapeliv Can you put in a suggestion to update the text for how this should be done then please?

docs/release-checklist.md
nabramow: @aapeliv I'm not sure what that means, can you please put in a suggestion that makes it clear to an outsider trying to go through this list?

docs/release-checklist.md
tristanlabelle: Since I've merged #8957, this is how we should update the bullet:

```
Update the `activeness_probe.latest_release` string in `app/backend/src/couchers/email/locales/en.json` so it has the most relevant top few recent features.
```

We'll also want to add another one for after publishing the blog:

```
Update `LATEST_RELEASE_BLOG_URL` in `app/backend/src/couchers/constants.py` to the url of the blog post for the latest release.
```

btw the formatting for markdown lists is `- [ ] foo`

nabramow: Thanks @tristanlabelle I've just updated and added these!

docs/release-checklist.md
nabramow: @aapeliv You never answered this, I am going to merge so it's not blocked longer. If you want to add something or make a change please submit a PR after.

aapeliv: sorry, i fixed the issue in another PR, so all done now

PR #8824, opened 2026-05-28
Web: Report missing i18n keys (build throw + Sentry)

aapeliv:
Users sometimes see raw i18n keys instead of translated text — rarely on desktop, often in the mobile WebView. This adds instrumentation to catch it from both ends.

Wires `saveMissing` + a `missingKeyHandler` into `next-i18next.config.js`:

- **Build time (`next build`)** → throws, failing the build with the offending `ns:key`. This guarantees the `en` source locale is complete and every namespace is wired into each page's `serverSideTranslations(...)`. Guarded to `NEXT_PHASE === "phase-production-build"` so it does not throw in dev or live SSR (where one missing key would crash a real page).
- **Browser** → `Sentry.captureMessage` (warning), deduped per `key@pathname`, tagged with namespace/key/language and `native_embed` (reusing the `injectedObjectJson().isNativeEmbed` detection), plus `languages_tried`, `pathname`, and `build_id` in extras.

Because the build throw eliminates the deterministic causes (missing-from-`en`, namespace not wired into a page), the remaining Sentry events should be predominantly the runtime failure mode we're chasing: soft-navigation / language-switch races and stale-WebView-cache-after-deploy. The `native_embed` tag and `build_id` let us confirm whether it's mobile-concentrated and deploy-correlated.

## Testing
- `yarn lint` clean on the changed file.
- Full `yarn build` passes with **0** missing-key throws — confirming there's no pre-existing backlog, so the build throw is safe to land.
- Build-throw path is exercised by static generation; the runtime Sentry path triggers in the browser when a key fails to resolve live.

**Web frontend checklist**
- [x] Added tests where relevant (n/a — config-level instrumentation)

## For maintainers

### DIFF
--- a/app/web/next-i18next.config.js
+++ b/app/web/next-i18next.config.js
@@ -15,6 +15,65 @@ const fallbackLng = {
 
 const debugLogging = process.env.NODE_ENV === "development";
 
+// Dedupe client-side missing-key reports so a key missing on a page doesn't
+// flood Sentry on every render/navigation.
+const reportedMissingKeys = new Set();
+
+// Inlined (can't import the TS util here: this config is required by Node at
+// build time, outside webpack). Mirrors utils/nativeLink.ts#isNativeEmbed.
+function inNativeEmbed() {
+  if (typeof window === "undefined" || !window.ReactNativeWebView) return false;
+  try {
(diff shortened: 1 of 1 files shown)


tristanlabelle (approved):
Good call

PR #8826, opened 2026-05-28
Backend: Fix host request rejected email + add email localization guardrail

### LINKED ISSUE #8825: LocalizationError: Could not localize string host_request_status_changed.rejected_by_host.subject for locale en
Sentry Issue: [BACKEND-MR](https://couchers.sentry.io/issues/7510193978/?referrer=github_integration)

```
LocalizationError: Could not localize string host_request_status_changed.rejected_by_host.subject for locale en
(3 additional frame(s) were not displayed)
...
  File "couchers/notifications/render_email.py", line 54, in render_email_notification
    subject = email.get_subject_line(loc_context)
  File "couchers/email/emails.py", line 889, in get_subject_line
    return self._localize(loc_context, "subject", {"other_name": self.other_user.name})
  File "couchers/email/emails.py", line 74, in _localize
    return get_emails_i18next().localize(key, loc_context.locale, substitutions)
  File "couchers/i18n/i18next.py", line 53, in localize
    if string := self.find_string(string_key, locale, substitutions):
  File "couchers/i18n/i18next.py", line 49, in find_string
    raise LocalizationError(locale, key)

Could not localize string host_request_status_changed.rejected_by_host.subject for locale en
```

@claude fix this

aapeliv:
Adds some tests to safeguard for and closes https://github.com/Couchers-org/couchers/issues/8825 (already hotfixed in 543a20f46d32ae0cffe4750d2b81339667b3db59)

---

Fixes a production `LocalizationError` and adds a guardrail so this class of bug can't recur.

`HostRequestStatusChangedEmail` built its i18next key as `host_request_status_changed.rejected_by_host` for the `REJECTED` status, but `email/locales/en.json` defines `declined_by_host`. Any rejected-host-request notification therefore crashed when rendering its subject line. The same band-aid was already hotfixed on `develop` (#543a20f46); this PR redoes it properly behind a regression test.

How it slipped through: emails build their localization keys at runtime (often dynamically from internal state), and nothing rendered them in a test. Even the existing `dump_emails.py` would have missed it, since `dummy_data()` only covers a single status — the rejected branch was never exercised.

This PR is structured as three commits so CI demonstrates the gap:
1. **Revert the hotfix** — reintroduces the bug.
2. **Add the guardrail test** — renders every email and every variant in English, asserting no missing localization key. `EmailBase.dummy_variants()` lets multi-variant emails enumerate all their branches. CI is **red** here, failing only on the rejected-by-host variant.
3. **Re-fix the key** — `rejected_by_host` → `declined_by_host`. CI goes **green**.

## Testing

`uv run pytest src/tests/test_email_localization.py` — 42 variants pass, and (before the fix) the rejected-by-host variant reproduces the exact production `LocalizationError`.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/emails.py
+++ b/app/backend/src/couchers/email/emails.py
@@ -912,7 +912,7 @@ class HostRequestStatusChangedEmail(EmailBase):
             case conversations_pb2.HOST_REQUEST_STATUS_ACCEPTED:
                 return f"{base_key}.accepted_by_host"
             case conversations_pb2.HOST_REQUEST_STATUS_REJECTED:
-                return f"{base_key}.rejected_by_host"
+                return f"{base_key}.declined_by_host"
             case conversations_pb2.HOST_REQUEST_STATUS_CONFIRMED:
                 return f"{base_key}.confirmed_by_surfer"
             case conversations_pb2.HOST_REQUEST_STATUS_CANCELLED:
--- a/app/backend/src/couchers/email/emails.py
+++ b/app/backend/src/couchers/email/emails.py
@@ -877,7 +877,7 @@ class HostRequestStatusChangedEmail(EmailBase):
             case conversations_pb2.HOST_REQUEST_STATUS_ACCEPTED:
                 return f"{base_key}.accepted_by_host"
             case conversations_pb2.HOST_REQUEST_STATUS_REJECTED:
-                return f"{base_key}.declined_by_host"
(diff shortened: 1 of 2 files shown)


tristanlabelle (approved):
Thanks and nice! It had crossed my mind to do this. I might follow-up with removing the `dummy_data` now, not much point to have both that and the variants.

PR #8827, opened 2026-05-28
Backend/emails: Port userless emails to new email system

tristanlabelle:
"Userless" emails include sign up emails and sending an email to a new unverified email address -- cases where we're not using a "User" object's info to send the email. With the new email infra, we can make them use much of the same logic as regular notification emails. I'll be able to consolidate even more once I'm done porting notification emails.

Related to #7368

## Testing
`uv run dump-emails` plus existing email rendering tests (via `test_instances()`).

Also tested with local backend/frontend since these emails are critical.

image
image

**Backend checklist**

## For maintainers


tristanlabelle:
Depends on #8861

aapeliv (approved):

PR #8842, opened 2026-05-28
Backend: Per-request resource accounting on api_calls

aapeliv:
Adds per-request resource accounting to `logging.api_calls` so we can see *why* an endpoint is expensive, not just that it's slow.

Per RPC, over the handler span, we now capture: `query_count`, `write_query_count`, `db_time_ms`, `cpu_ms`, and `client_platform`. A thread-local accumulator in `couchers/perf.py` is fed by SQLAlchemy cursor-execute listeners (writes detected via `ExecutionContext.is_crud`); the interceptor arms it around the handler so the numbers exclude auth and the log insert.

Also surfaced in Prometheus (method-labelled): `db_time_seconds`, `cpu_seconds`, `query_count`, `write_query_count` histograms, plus a `couchers_api_calls_total` counter labelled by method + platform.

With these you can decompose latency into backend CPU vs Postgres vs residual wait, rank endpoints by DB cost, and spot N+1s.

## Testing
Unit + integration tests in `test_interceptors.py` (column population, write detection, histogram/counter observation). `make format` and `make mypy` clean.

**Backend checklist**
- [x] Added migrations if there are any database changes (autogenerated in CI)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/metrics.py
+++ b/app/backend/src/couchers/metrics.py
@@ -132,14 +132,23 @@ servicer_query_count_histogram: Histogram = Histogram(
     "couchers_servicer_query_count",
     "Number of SQL statements executed per gRPC call",
     labelnames=["method"],
-    buckets=(1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, _INF),
+    buckets=(1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, _INF),
+)
+servicer_write_query_count_histogram: Histogram = Histogram(
+    "couchers_servicer_write_query_count",
+    "Number of INSERT/UPDATE/DELETE statements executed per gRPC call",
+    labelnames=["method"],
+    buckets=(1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, _INF),
 )
 
 
-def observe_in_servicer_perf_histograms(method: str, query_count: int, db_time_ms: float, cpu_ms: float) -> None:
+def observe_in_servicer_perf_histograms(
+    method: str, query_count: int, write_query_count: int, db_time_ms: float, cpu_ms: float
+) -> None:
     servicer_db_time_histogram.labels(method).observe(db_time_ms / 1000)
     servicer_cpu_time_histogram.labels(method).observe(cpu_ms / 1000)
     servicer_query_count_histogram.labels(method).observe(query_count)
+    servicer_write_query_count_histogram.labels(method).observe(write_query_count)
 
 
 # list of gauge names and function to execute to set value to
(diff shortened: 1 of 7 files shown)


aapeliv:
@claude review this PR

app/backend/src/couchers/metrics.py
```diff
@@ -86,6 +86,55 @@ def observe_in_servicer_duration_histogram(
...
+servicer_cpu_time_histogram: Histogram = Histogram(
+    "couchers_servicer_cpu_seconds",
+    "Backend thread CPU time per gRPC call",
+    labelnames=["method"],
+)
+servicer_query_count_histogram: Histogram = Histogram(
+    "couchers_servicer_query_count",
+    "Number of SQL statements executed per gRPC call",
+    labelnames=["method"],
+    buckets=(1, 2, 3, 5, 8, 13, 21, 34, 55, 89, 144, 233, 377, _INF),
```
tristanlabelle, line 136: What's the significance of fibo here? A comment would help.

app/backend/src/couchers/models/logging.py
```diff
@@ -64,6 +65,14 @@ class APICall(Base, kw_only=True):
     # human readable perf report
     perf_report: Mapped[str | None] = mapped_column(String, default=None)
 
+    # per-request resource accounting, covering the handler span (see couchers/perf.py)
+    query_count: Mapped[int | None] = mapped_column(BigInteger, default=None)
+    write_query_count: Mapped[int | None] = mapped_column(BigInteger, default=None)
```
tristanlabelle: I suggest `db_query_count`, `db_write_query_count` throughout.

app/backend/src/couchers/perf.py
```diff
@@ -0,0 +1,72 @@
...
+        return
+    elapsed_ms = (perf_counter_ns() - starts.pop()) / 1e6
+    acc: _PerfAccumulator | None = getattr(_local, "acc", None)
+    if acc is None:
+        # Query on a thread with no active request (background job, metrics scrape, etc.) - don't attribute it.
+        return
+    acc.query_count += 1
+    acc.db_time_ms += elapsed_ms
+    # is_crud is set by SQLAlchemy when it compiled an INSERT/UPDATE/DELETE.
+    if context.is_crud:
```
tristanlabelle: That `is_crud` is not documented: https://docs.sqlalchemy.org/en/20/core/internals.html#sqlalchemy.engine.ExecutionContext . There's only `isinsert` and `isupdate`. Any additional considerations or fallbacks needed?

tristanlabelle (dismissed):
Nice addition!

app/backend/src/couchers/metrics.py
aapeliv: it's just some bucketing, not sure what else to put that's

app/backend/src/couchers/models/logging.py
aapeliv: thanks, updated

app/backend/src/couchers/perf.py
aapeliv: thanks, i updated to use those (+ isdelete, which i hope is equally well supported...)

PR #8861, opened 2026-05-30
Backend: Split quick_links for layering/avoiding circular module refs

tristanlabelle:
`notifications.quick_links.py` included both the data-only link generation code and the database-manipulating servicer implementation. I've run into circular dependencies more than once due to this because we need to generate quick_links in email/notification logic that is lower level than servicers, so if they transitively import the Requests servicers, things blow up.

Fix this by raising the `respond_quick_link` function to the servicer layer, where it logically belongs.

For code organization, I preferred putting it in `servicers.auth_quick_links` than making `servicers.auth` bigger, though this the first example of this splitting up of a servicer's logic in two files.

Example of a circular reference if `quick_links` import servicers, and emails try to import `quick_links` (what I'm preparing in https://github.com/Couchers-org/couchers/pull/8827):

```
src/tests/conftest.py:25: in <module>
    from tests.fixtures.db import (  # noqa: E402
src/tests/fixtures/db.py:40: in <module>
    from couchers.servicers.auth import create_session
src/couchers/servicers/auth.py:44: in <module>
    from couchers.notifications.quick_links import respond_quick_link
src/couchers/notifications/quick_links.py:32: in <module>
    from couchers.servicers.requests import Requests
src/couchers/servicers/requests.py:40: in <module>
    from couchers.rate_limits.check import process_rate_limits_and_check_abort
src/couchers/rate_limits/check.py:9: in <module>
    from couchers.tasks import send_rate_limit_violation_report_email
src/couchers/tasks.py:8: in <module>
    import couchers.email.emails as emails
src/couchers/email/emails.py:22: in <module>
    from couchers.notifications.quick_links import generate_quick_decline_link
E   ImportError: cannot import name 'generate_quick_decline_link' from partially initialized module 'couchers.notifications.quick_links' (most likely due to a circular import) (/app/src/couchers/notifications/quick_links.py)
```

## Testing
N/A - No functional changes.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/notifications/quick_links.py
+++ b/app/backend/src/couchers/notifications/quick_links.py
@@ -31,7 +31,7 @@ def _generate_quick_link(payload: Message) -> str:
     return urls.quick_link(payload=b64encode(msg), sig=b64encode(sig))
 
 
-def decode_quick_link(*, payload: bytes, sig: bytes, context: CouchersContext) -> unsubscribe_pb2.UnsubscribePayload:
+def decode_quick_link(payload: bytes, sig: bytes, context: CouchersContext) -> unsubscribe_pb2.UnsubscribePayload:
     if not verify_hash_signature(message=payload, key=get_secret(UNSUBSCRIBE_KEY_NAME), sig=sig):
         context.abort_with_error_code(grpc.StatusCode.PERMISSION_DENIED, "wrong_signature")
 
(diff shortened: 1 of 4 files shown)


aapeliv (dismissed):
i still personally feel like we should have the generation and parsing next to each other. since these get baked into emails and need to survive for some period of time, and the bugs are nasty; it's good to have them next to each other... i feel.

but up to you!

tristanlabelle:
@aapeliv I agree with your generation+parsing being together concern so in the latest I kept `decode_quick_link` in the original `quick_links.py` so encoding to and from `UnsubscribePayload` are now kept together, so that should work for everyone.

PR #8862, opened 2026-05-30
Backend/emails: Consolidate EmailBase.dummy_data and dummy_variants

tristanlabelle:
#8826 added `dummy_variants` as a means to test all email string lookups. There's no use case for the original `dummy_data`, so consolidate them. I chose the name `test_instances` to highlight that they are meant to provide test coverage. Also updated `dump-emails` to support dumping all instances.

## Testing
Ran `uv run dump-emails`

**Backend checklist**

## For maintainers


aapeliv (approved):
thanks!

PR #8865, opened 2026-05-30
Disable EXPERIMENTATION_PASS_ALL_GATES in dev

aapeliv:
Disables `EXPERIMENTATION_PASS_ALL_GATES` in the dev environment.

With it on, every feature gate returned `True` in dev, which force-enabled flag-gated background jobs that call live external APIs with placeholder credentials — e.g. `check_mypostcard_jobs` hitting MyPostcard and `add_users_to_email_list` hitting Listmonk — producing spurious `401 Unauthorized` errors in the backend logs. With the override off, flags fall through to their in-code defaults (mostly `False`) in dev, so these jobs bail cleanly.

Note: with experimentation also disabled in dev, all flag-gated features are now off locally by default. Flip this back temporarily if you need to exercise a gated feature in dev.

## Testing
- Restarted the backend locally; `check_mypostcard_jobs` no longer reaches the MyPostcard API and the recurring 401 is gone.

**Backend checklist**

## For maintainers


aapeliv:
Superseded by #8866, which removes `EXPERIMENTATION_PASS_ALL_GATES` entirely instead of just flipping the dev value.

PR #8866, opened 2026-05-30
Remove EXPERIMENTATION_PASS_ALL_GATES

aapeliv:
Removes the `EXPERIMENTATION_PASS_ALL_GATES` override entirely (backend + web).

## Why
The override forced **every** boolean feature gate to `True`, including on the global/background evaluation path. In dev this force-enabled flag-gated background jobs that call live external APIs with placeholder credentials — e.g. `check_mypostcard_jobs` (MyPostcard) and `add_users_to_email_list` (Listmonk) — producing recurring `401 Unauthorized` / connection errors in the logs. It was also the implicit test default, so the whole suite ran with all flags on regardless of each flag's real in-code default.

## What
- **Backend**: drop the `EXPERIMENTATION_PASS_ALL_GATES` config var (`config.py`) and the short-circuit in `experimentation._boolean_value` (which became identical to `_feature_value`, so it's folded in). `CouchersContext.get_boolean_value` and `get_global_boolean_value` now go straight through `_feature_value`. Removed from `backend.dev.env`.
- **Web**: remove `shouldPassAllGates()` and its use in `CouchersFlagProvider.resolveBooleanEvaluation`, plus `NEXT_PUBLIC_EXPERIMENTATION_PASS_ALL_GATES` from `.env.development`.
- **Tests**: flags now evaluate to their real value (in-code default when experimentation is disabled, which is the test default). Tests that exercised a flag-gated feature and previously relied on the global default-on now opt in explicitly via the existing `feature_flags` fixture: feature-dedicated files (`postal_verification`, `donations`, `strong_verification`) enable their flag in the module's autouse fixture; mixed files (`requests`, `verification`, `calendar_events`) opt in per-test.

Net effect: in dev, flag-gated features (and their background jobs) are now off by default unless turned on via GrowthBook, instead of being silently force-enabled.

## Testing
- `uv run pytest` — **1058 passed**, 4 deselected. The 4 deselected (`test_create_servers`, `test_job_retry`, two `test_check_native_status_*`) fail identically on clean `develop` — port binds from a locally-running backend and a stale local `bugs_pb2` stub — and are unrelated to this change.
- `make format` clean; `make mypy` shows only pre-existing errors (stale local `bugs_pb2` / missing `ics` stubs), identical on `develop`.

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/backend.dev.env
+++ b/app/backend.dev.env
@@ -78,7 +78,6 @@ ACTIVENESS_PROBES_ENABLED=1
 
 # Experimentation (feature flags)
 EXPERIMENTATION_ENABLED=0
-EXPERIMENTATION_PASS_ALL_GATES=1
 GROWTHBOOK_API_HOST=https://gbapi.couchershq.org
 GROWTHBOOK_CLIENT_KEY=sdk-cGmljeXu1BCzHffE
 GROWTHBOOK_CACHE_PATH=growthbook-features-cache.json
--- a/app/backend/src/couchers/config.py
+++ b/app/backend/src/couchers/config.py
@@ -110,8 +110,6 @@ CONFIG_OPTIONS: CONFIG_T = [
     ("IN_TEST", bool, "0"),
     # Experimentation (feature flags via GrowthBook)
     ("EXPERIMENTATION_ENABLED", bool, "0"),
-    # When enabled, all feature gates return True (useful for development/testing)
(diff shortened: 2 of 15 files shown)


aapeliv:
superseded by https://github.com/Couchers-org/couchers/pull/8948

PR #8867, opened 2026-05-30
Backend: fix user_activity unique-violation race in auth interceptor

aapeliv:
Since #8846 started running multiple API worker processes in parallel, we've been seeing a flood of `UniqueViolation` errors on `ix_user_activity_user_id_period_ip_address_user_agent_sofa` coming from the auth interceptor.

The cause is a read-then-insert race in `_try_get_and_update_user_details`: it outer-joined `UserActivity` to find the row for the current `(user_id, period, ip_address, user_agent, sofa)` tuple, and inserted one if absent. A single API process with the GIL kept the window between the check and the insert tiny, so the race almost never fired. With multiple worker processes serving requests in true parallel, two concurrent calls from the same user/device in the same hourly bin both see "no row" and both `INSERT` → the second violates the unique index.

This replaces the read-modify-insert with an atomic `INSERT ... ON CONFLICT DO UPDATE` upsert, which removes the window entirely and increments `api_calls` (and refreshes `client_platform` when provided) atomically.

Because `ip_address`/`user_agent`/`sofa` are nullable and `ON CONFLICT` relies on the unique index detecting collisions — and PostgreSQL treats NULLs as *distinct* by default, so NULL tuples would never conflict and would insert a fresh row every call — the unique index is also changed to `NULLS NOT DISTINCT`. This matches the old behaviour, where `col == None` rendered as `IS NULL` and so matched/incremented existing NULL rows. Migration `0160` rebuilds the existing index; the test DB picks it up automatically via `create_all`.

Existing production data should already be free of duplicate NULL-tuple rows (the old `IS NULL` logic never created them), so rebuilding the index `NULLS NOT DISTINCT` won't fail on existing duplicates.

## Testing
`uv run pytest src/tests/test_interceptors.py` passes. The existing `test_auth_interceptor` already exercises the NULL-tuple dedup path (an API-key call with no browser headers must increment the same row rather than create a second), which now passes against the upsert + `NULLS NOT DISTINCT` index.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -17,7 +17,8 @@ from google.protobuf.descriptor_pool import DescriptorPool
 from google.protobuf.message import Message
 from opentelemetry import trace
 from sqlalchemy import Function, literal_column, select
-from sqlalchemy.sql import and_, func
+from sqlalchemy.dialects.postgresql import insert as pg_insert
+from sqlalchemy.sql import func
 
 from couchers.constants import (
     CALL_CANCELLED_ERROR_MESSAGE,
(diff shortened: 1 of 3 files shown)


aapeliv:
verified no dupes as per migration review. merging for hotfix

PR #8869, opened 2026-05-30
Web: run the prod Next.js server in cluster mode

aapeliv:
The prod web image ran a single `node server.js` process. That pins one CPU core and means a hung or crashed process takes the whole frontend down. The backend recently moved to a multiprocess model (one container, N API workers); this brings the web tier in line.

Adds a clustered entrypoint (`web/server-cluster.js`) that forks `WEB_WORKER_COUNT` workers sharing port 3000 (the OS load-balances accepts across them), respawns any worker that dies, and forwards `SIGTERM`/`SIGINT` so each worker runs Next's own graceful shutdown. The prod Dockerfile copies the wrapper in and switches `CMD` to it; the prod compose sets `WEB_WORKER_COUNT=4` (tunable per host, no rebuild needed).

Scope notes:
- Dev is untouched — `next dev`/`next start` don't use the standalone `server.js`, so this only affects the prod image.
- nginx is untouched — workers share port 3000, so `proxy_pass http://web:3000` is unchanged.

## Testing
- `prettier` + `eslint` pass on the new file (uses the repo's `// eslint-disable-line` convention for the CommonJS `require`s, matching `next.config.js`).
- Smoke-tested the cluster mechanics with a stub `server.js`: confirmed it forks N workers, respawns a `kill -9`'d worker, and on `SIGTERM` forwards to workers (graceful exit) with the primary exiting 0.
- Not yet run against a real production build/deploy; on deploy the web container logs should show `[cluster] primary … starting 4 worker(s)`.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/Dockerfile.prod
+++ b/app/web/Dockerfile.prod
@@ -50,6 +50,7 @@ COPY --from=builder /app/package.json ./package.json
 
 COPY --from=builder --chown=nextjs:nodejs /app/.next/standalone ./
 COPY --from=builder --chown=nextjs:nodejs /app/.next/static ./.next/static
+COPY --from=builder --chown=nextjs:nodejs /app/server-cluster.js ./
 
 COPY --from=builder /app/.env.local ./
 COPY --from=builder /app/proto ./proto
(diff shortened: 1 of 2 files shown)


aapeliv:
@claude review this

PR #8882, opened 2026-05-31
mobile: declare native client platform to the backend

aapeliv:
Adds a `PlatformInterceptor` to the mobile gRPC client that sends the `x-couchers-client-platform` metadata header (`app_ios` / `app_android`) on every request, mirroring the web client's interceptor.

The backend recently started recording the client platform (`logging.api_calls`, `user_activity`, and the active-users-by-platform metric) by reading this client-declared header. The web frontend already sends `web_desktop` / `web_mobile`, but the native app sent nothing, so its traffic showed up as `unknown`. With this change native API calls are attributed to iOS/Android correctly.

A new `utils/clientPlatform.ts` maps `Platform.OS` to the backend enum values (returning `null` for anything other than ios/android, so the web build never sends a bogus value).

## Testing
- `npm run lint` — clean.
- `npm test tests/service/client.test.ts` — 5/5 passing, including a new `PlatformInterceptor` test asserting the metadata header is set to the native platform value.

**Web frontend checklist**

## For maintainers


nabramow (approved):
LGTM!

PR #8885, opened 2026-05-31
Backend: resolve native OTA manifests from a DB packages table

aapeliv:
Moves native OTA version-selection out of the `native_ota_bundles` GrowthBook flag and into a backend-owned `ota_packages` table, so the backend (not a per-platform flag) decides which signed bundle to offer each build, and so packages can be banned.

**How resolution works**
- The table is keyed on `(platform, runtime_version)` — the Expo fingerprint is the compatibility boundary, and many store builds can legitimately share one fingerprint.
- `GetNativeUpdateManifest` serves the **newest non-banned package by the manifest's `createdAt`** for the requesting fingerprint, then fetches the **signed** manifest bytes from the CDN (`{cdn_root}/{version}/{platform}/manifest`, via the existing `@lru_cache`) and serves them **verbatim**. The backend never rebuilds the manifest, so code-signing is completely untouched.
- A stale-but-fresh store build self-heals to the latest bundle on first load; the device's own commitTime selection policy prevents downgrading a newer embedded build.
- `banned` only stops handing a bundle to **new** check-ins (it can't reclaim devices already on it). Rollback is done on the publish side by re-stamping the good bundle with a newer `createdAt`, which then wins automatically.

**Admin APIs** (`Admin` service): `CreateOTAPackage(platform, version, manifest_json)` — `runtimeVersion`/`createdAt`/`id` are read out of the pasted manifest; `ListOTAPackages` (newest-first, with a computed `live` flag); `BanOTAPackage`/`UnbanOTAPackage`.

**Not included (follow-ups):** wiring CI to call `CreateOTAPackage` after publishing, and removing the now-dead `native_ota_bundles` flag (`native_ota_cdn_root` is still read).

## Testing
- `make format`, `make mypy` (only the pre-existing unrelated `ics` stub error), `make protos` — clean.
- Backend tests: `test_bugs.py` (25 passing) covers per-fingerprint resolution, newest-by-`createdAt` ordering, CDN fetch URL + verbatim bytes, banned exclusion, and all directive paths; `test_admin.py` OTA tests (6 passing) cover create/validation/duplicate-version/list+live/ban+unban; migration roundtrip (`test_db.py::test_migrations`) passes.

**Backend checklist**

## For maintainers


aapeliv:
Superseded by #8903, which combines this with #8899 on a single rebased branch.

PR #8899, opened 2026-05-31
Backend: implement CheckNativeStatus evergreen update decision

aapeliv:
Replaces the `CheckNativeStatus` stub (which always returned `NATIVE_UPDATE_ACTION_NONE`) with a real **evergreen + ban** update decision, factored into a new `couchers.native_updates` module.

**Model.** Two independent clocks:
- **Store window (default 91d)** gates the native binary — always in force, because no OTA can rescue a binary past its window.
- **OTA window (default 28d)** gates the JS bundle running on top — only when the client is running an OTA.

Each clock is `none` below 75% of its window, `warn` from 75%, and `block` at 100%. The two are resolved by severity first, then store precedence (a failing binary can only be fixed via the store). The resulting decision maps onto `NativeUpdateInfo` with `required=True` + `act_by = created_at + window`; the existing mobile client already turns a future deadline into a dismissible warning and a past one into a hard block, so the warn→block transition is automatic.

**GrowthBook = policy only**, never release data: support windows, store URL / message / link text. No release tables, no ban lists here.

**OTA registry seam.** The authoritative OTA publish time (and OTA-side banning / registration / tracking) belongs on the side of `GetNativeUpdateManifest`, which is being built separately. This PR introduces a stubbed `couchers.ota_registry.OtaRegistry` (`get_package(platform, update_id) -> OtaPackage | None`) for `native_updates` to read through; until the real registry lands the stub returns `None` and we fall back to the client-reported timestamps.

See `OTA-plan.md` for the full design write-up and open items (exact `debug_json` wire format, ban identifiers).

## Testing

- New `test_native_updates.py` covers parse tolerance, each clock, the warn/block thresholds, severity + store-precedence resolution, policy windows, presentation, and the registry override.
- Updated `test_check_native_status_*` in `test_bugs.py`; added a coverage case where an expired binary produces a blocking `STORE` response end-to-end.
- `make format`, `make mypy` (only the pre-existing `ics` stub error remains, unrelated), `uv run pytest src/tests/test_native_updates.py src/tests/test_bugs.py` — 40 passed.

**Backend checklist**
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (n/a — no DB changes)

## For maintainers

### DIFF
--- a/OTA-plan.md
+++ b/OTA-plan.md
@@ -1,0 +1,128 @@
+# Native update logic — `CheckNativeStatus` evergreen + ban
+
+Design for the backend logic behind `CheckNativeStatus`. The mobile app pings this
+on cold start / foreground with a diagnostics payload and gets back a
+`NativeUpdateInfo` telling it whether (and how) to update. Today the servicer is a
+stub that always returns `NATIVE_UPDATE_ACTION_NONE`. This replaces it with a real
+**evergreen + ban** decision, factored into a shared module so the OTA manifest
+endpoint can reuse the policy bits.
+
+**Status:** design / not yet implemented.
+
+---
+
+## Principles
+
+- **Evergreen.** A build is supported for a fixed time from when it was created.
+  Past that, the client must update. Keeps everyone on recent code.
+- **Bannable.** A specific bad build can be force-expired immediately, regardless
+  of age.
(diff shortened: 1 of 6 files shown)


aapeliv:
Superseded by #8903, which combines this with #8885 on a single rebased branch.

PR #8910, opened 2026-05-31
Change mobile map search results puller style

nabramow:
There were some comments that people didn't see the arrow to realize you could increase the list height on mobile for the map search results. That weird design was never meant to be permanent anyway so how about this?

Screenshot 2026-05-31 at 23 46 45

## Testing
- Use map search to search something
- Go in mobile responsive view
- Use puller to make it go up and down
- Chewck desktop view is still okay

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/search/MapSearchContent.tsx
+++ b/app/web/features/search/MapSearchContent.tsx
@@ -50,16 +50,12 @@ const SearchResultsContainer = styled("div", {
     height: "100%",
     width: isListOnlyView ? "100%" : `${drawerWidth}px`,
 
-    ...(!isListOnlyView && {
-      [theme.breakpoints.down("md")]: {
-        width: "100%",
-        height: "45%",
-        minHeight: 0,
-        order: 2,
-        boxShadow: "0px -2px 4px rgba(0,0,0,0.1)",
-        overflow: "hidden",
-      },
-    }),
+    [theme.breakpoints.down("md")]: {
+      width: "100%",
+      height: isListOnlyView ? "100%" : "45%",
+      order: 2,
+      transition: "height 0.3s ease",
+    },
   }),
 );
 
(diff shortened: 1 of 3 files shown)


darrenvong:
When tested on my phone, the pulling up and down felt janky, even though it works. I think there was probably a native expectation for it to be moved with my finger as I drag it rather than experiencing a small delay before seeing it update, and also for it to stop where I stopped the dragging.

Not sure how much of a pain it is to implement that on web though 😅

https://github.com/user-attachments/assets/1de722e7-82ab-41c1-b33a-abbe17547ac8

darrenvong:
Ahh actually I see that is the current behaviour on prod too, so it's probably the squashed pill styling that made it misleading for me and to have the expectation described in my last message. Maybe update it to a stretched up/down arrow in the middle, in place of the squashed pill, if you don't want to deal with the more complex touch mechanics? 🤔

c-kreuz:
I'm with Darren here. It looks like a slider but it's actually a button. Maybe do something similar as on desktop with the Chevron in a circle overlaping 50%

jesseallhands (changes requested):
I thought the swipe up expandable drawer idea would be excellent initially, but after testing it I have to agree with Darren. I find it harder to use, and I think it's even less visible that you can expand the drawer. 

I think we should basically keep the arrow button we currently have but make it slightly bigger and have it half overlap the map so it's clear you can expand the drawer upward.

nabramow:
@jesseallhands @c-kreuz @darrenvong So I agree on moving it to the center, but my sticking point here is that then instead of becoming really full screen like now, it has this tiny gap that looks a bit weird IMO. But then I can't think of another way to design it to avoid that thats not also weird.

Maybe I'll have a lil convo with Claude Design and see if it has better ideas.

nabramow:
Closing as this isn't quite right and need another design idea.

PR #8917, opened 2026-06-01
web: promote WithFlags to app root

aapeliv:
Promotes the GrowthBook `<WithFlags>` provider + first-load ready-gate from the two badges pages (`pages/badges/index.tsx`, `pages/badges/[id].tsx`) up into `_app.tsx`, so it wraps the whole app.

This is the planned follow-up to #8901, where `<WithFlags>` was deliberately scoped to a single page to de-risk the initial rollout. With it at the root, every page now sees feature flags and shares the single startup load (15-min cache freshness) / background refresh (5-min interval + on-foreground) cycle. `ProfileSheet`'s `BadgeDetail`, which renders globally outside the badges pages, now gets live flag values instead of its in-code default.

`<WithFlags>` sits inside `<AuthProvider>` (it reads `authState.userId` to set the GrowthBook attribute) but outside `CssBaseline`/theme infra, so the `CenteredSpinner` shown during the first flag load is still themed.

## Testing
- `yarn tsc --noEmit`, `yarn lint`, and `yarn format` all clean locally.
- Behaviour is unchanged from #8901; this only moves where the provider is mounted.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/experimentation/WithFlags.tsx
+++ b/app/web/features/experimentation/WithFlags.tsx
@@ -13,8 +13,7 @@ import {
 } from "./growthbook";
 
 // Provides the GrowthBook SDK and waits for the first flag load before rendering its subtree, so
-// flag-driven UI never flickers. Currently mounted around the only flag-consuming page; promoting it
-// to the app root (in _app.tsx) is the intended next step once this is proven in prod.
+// flag-driven UI never flickers. Mounted at the app root (in _app.tsx) so every page sees flags.
 export default function WithFlags({ children }: { children: ReactNode }) {
   const { authState } = useAuthContext();
 
--- a/app/web/pages/_app.tsx
+++ b/app/web/pages/_app.tsx
@@ -16,6 +16,7 @@ import NativeColorSchemeSync from "components/NativeColorSchemeSync";
 import NativeMobileNavigationHandler from "components/NativeMobileNavigationHandler";
 import { AnalyticsProvider } from "features/analytics";
 import AuthProvider from "features/auth/AuthProvider";
+import WithFlags from "features/experimentation/WithFlags";
 import ProfileSheet from "features/profile/ProfileSheet";
 import { ProfileSheetProvider } from "features/profile/ProfileSheetContext";
 import { ReactQueryClientProvider } from "features/reactQueryClient";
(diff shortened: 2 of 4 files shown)


app/web/features/experimentation/WithFlags.tsx
```diff
@@ -13,8 +13,7 @@ import {
 } from "./growthbook";
 
 // Provides the GrowthBook SDK and waits for the first flag load before rendering its subtree, so
-// flag-driven UI never flickers. Currently mounted around the only flag-consuming page; promoting it
-// to the app root (in _app.tsx) is the intended next step once this is proven in prod.
+// flag-driven UI never flickers. Mounted at the app root (in _app.tsx) so every page sees flags.
 export default function WithFlags({ children }: { children: ReactNode }) {
```
nabramow, line 17: If it's specifically feature flags can we call it "withFeatureFlags" just so it's clear?

app/web/features/experimentation/WithFlags.tsx
nabramow, line 37: Also I've always done these sort of HOC (higher order components) in function case so `withFeatureFlags` and then instead of using it as a component in the return you wrap the whole react component so like

```javascript
function BadgesPage() {
  // ...
}

export default withFeatureFlags(BadgesPage);
```

I've been a backend engineer commercially the last few years tho so it could have changed curious what @darrenvong thinks.

I see that pattern in their legacy docs but no reference to HOCs in the new docs:
https://legacy.reactjs.org/docs/higher-order-components.html

app/web/pages/_app.tsx
```diff
@@ -88,10 +89,12 @@ function MyApp(props: AppWithLayoutProps) {
...
-                    <ProfileSheetProvider>
-                      {getLayout(<Component {...pageProps} />)}
-                      <ProfileSheet />
-                    </ProfileSheetProvider>
+                    <WithFlags>
+                      <ProfileSheetProvider>
+                        {getLayout(<Component {...pageProps} />)}
+                        <ProfileSheet />
+                      </ProfileSheetProvider>
+                    </WithFlags>
```
nabramow, line 97: Another stylistic nit. All the other ones wrapping stuff here are providers , I'm still leaning towards it being a function `withFeatureFlags` that then wraps what we need similar to `getLayout`.

On another note do you specifically need the flags in the profile sheet?

nabramow:
@aapeliv I just looked up how GrowthBook suggests we do this for React to see their docs and they have a react SDK package for it, I'd suggest we just use that:

https://docs.growthbook.io/lib/react

This current pattern seems a bit deprecated and I think that will be more readable.

app/web/features/experimentation/WithFlags.tsx
darrenvong: Yeah I'd propose going with their hooks API instead, as there's not much that can't be achieved with hooks these days that the HOC pattern was designed to do.

aapeliv:
superseded by https://github.com/Couchers-org/couchers/pull/8942

PR #8923, opened 2026-06-01
web: drive global message from a feature flag

aapeliv:
The global message that allows us to show an error to every user is currently based on some static JSON files being served from a domain just for this. There's fancy logic to re-load it regularly and so on.

I think it's cleaner to pull the value from the feature flag system given it's basically all the custom load logic to keep values fresh. This PR implements that.

You store it in the `global_message` feature flag as something like this:

```json
{
  "severity": "info",
  "message": "Testing for PR #8923 (stage)"
}
```


aapeliv:
@nabramow cleaned up PR description, fully human-written, 0-slop!

app/web/components/GlobalMessage.tsx
```diff
@@ -1,49 +1,26 @@
...
-  // last time we queried it
-  const [lastCheck, setLastCheck] = usePersistedState<number | null>("globalmessage.lastcheck", null);
-  // the epoch value of the last message we dismissed
-  const [dismissedEpoch, setDismissedEpoch] = usePersistedState<string | null>("globalmessage.dismissed", null);
-
-  useEffect(() => {
-    const fetchData = async () => {
-      try {
-        const response = await fetch(process.env.NEXT_PUBLIC_GLOBAL_MESSAGE_URL);
+  const data = useFeatureValue<GlobalMessageData | null>("global_message", null);
```
nabramow, line 13: All the other flags are more like `global_message_enabled`, shall we change this to have the same pattern?

nabramow:
@aapeliv I tried to test this by adding `"global_message": true` to `feature-flags.dev.json` but didn't see the global message pop up.

What else do I need to do to add it? I also tried logging into the growth book via the express.couchershq.org link in the docs but it says server not configured when I put my nicole couchers org email.

Regarding this comment in the description, "You store it in the global_message feature flag as something like this". Where is that?

I wonder if we should have a short blurb on ops.md or something about how to put this up as usually we'll be doing it in stressful circumstances.

app/web/components/GlobalMessage.tsx
aapeliv: the feature flag needs to include the content of the global message, so it's more than a boolean

aapeliv:
need to set it to the right JSON structure:

```json
"global_message": { "severity": "info", "message": "Hi!" }
```

Screenshot 2026-09-07 at 14 46 37

aapeliv:
> I wonder if we should have a short blurb on ops.md or something about how to put this up as usually we'll be doing it in stressful circumstances.

i will build it into the ops dashboard

PR #8926, opened 2026-06-01
Backend/emails: Port references emails to new system

tristanlabelle:
Same as my several previous PRs, now for references emails. No special considerations, except that I'm using more string key nesting as I intend to do a pass and do that more generally.

Also these were closed with `Thanks for using Couchers!`, I discarded it in favor of the one used by most other emails, `Best,<br />The Couchers.org Team`.

Since these are the last emails to be ported, I removed the `_get_custom_templated_email` branch completely. Yay! I'll do a bit more cleanup after merging.

## Testing
Ran `uv run dump-emails`, otherwise covered by tests.

image

image

image

**Backend checklist**

## For maintainers


aapeliv (approved):
lgtm

PR #8932, opened 2026-06-01
Backend: Support strongly-typed config vars access

tristanlabelle:
Refactors `config` to store values as statically-typed, mypy-verifiable attributes. E.g. `config.background_worker_count` with static type `int` instead of untyped `config["BACKGROUND_WORKER_COUNT"]`.

To maintain backcompat and ease migration, the new `Config` style also mimics a dictionary, so indexed access using all caps environment variable names still works. I'd remove that after migrating to only accessing through attributes. Undefined config attributes are also still possible and will raise `AttributeError` on access, or `KeyError` if using indexed dictionary-like access.

We can't use a regular @dataclass because our code relies on working with partial environment variables [see here](https://github.com/Couchers-org/couchers/blob/045c86ca65a3e0172d23360690a2f4c69f4c3a21/app/backend/src/tests/conftest.py#L14), which I'd like to see if I can address subsequently.

If we're cool with the design, I'll migrate all accesses to use the attributes directly in a follow-up PR.

## Testing
Added tests.

Ran the server locally. I'm hitting these but they seem unrelated:

`[116010:     264350263812128] 2026-06-01 07:35:53,449: couchers.jobs.worker:96: 401 Client Error: Unauthorized for url: https://www.mypostcard.com/api/v1/request_orders`

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/app.py
+++ b/app/backend/src/app.py
@@ -20,7 +20,7 @@ import sentry_sdk
 from sentry_sdk.integrations import excepthook
 from sqlalchemy.sql import text
 
-from couchers.config import check_config, config
+from couchers.config import config
 from couchers.constants import API_BASE_PORT, API_WORKER_COUNT, GRACEFUL_SHUTDOWN_TIMEOUT, MEDIA_PORT
 from couchers.db import apply_migrations, db_post_fork, session_scope
 from couchers.experimentation import setup_experimentation
@@ -33,7 +33,7 @@ from couchers.supervisor import supervise
 from couchers.tracing import setup_tracing
 from dummy_data import add_dummy_data
 
-check_config(config)
+config.check()
 
 logging.basicConfig(
     format="[%(process)5d:%(thread)20d] %(asctime)s: %(name)s:%(lineno)d: %(message)s", level=logging.INFO
--- a/app/backend/src/couchers/config.py
+++ b/app/backend/src/couchers/config.py
@@ -150,46 +150,6 @@ class Config:
                 continue
             self.__setattr__(attr_name, default_value)
 
-    def __getitem__(self, key: str) -> Any:
-        """Weakly-typed indexer access using env var names for backcompat."""
-        attr_name = key.lower()
-        if Config.__annotations__.get(attr_name) is None:
-            raise KeyError(f"No such config key: {key}.")
-
-        try:
-            return self.__getattribute__(attr_name)
-        except AttributeError:
-            raise KeyError(f"Config key undefined and has no default: {key}.") from None
-
-    def __setitem__(self, key: str, value: Any) -> None:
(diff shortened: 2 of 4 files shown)


WouldYouKindly:
@tristanlabelle I'm 100% for static typing over dictionaries. I have another proposal for the implementation though:

```python
@dataclass
class Config:
   host: str
   port: int

def make_config():
    return Config(
        host=os.environ['HOST'],
        port=int(os.environ.get('PORT', 8000)),
    )
```

This way we'll have a typed class with no magic. This is very dumb code, but I don't think we need anything more complex with type annotations, converters and all. 

If you want something more fancy, there's https://environ-config.readthedocs.io/en/stable/index.html or https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/ (I'm not a fan of pydantic but it's widely used)

tristanlabelle:
@WouldYouKindly My goal is to get to that declaration but I tried it in https://github.com/Couchers-org/couchers/pull/8478 and it's not that simple: we currently rely on being able to run without a full environment, e.g. when running migrations or tests, i.e. some `config["XXX"]` entries can be missing. A dataclass would require initializing everything or making every variable Nonable.

This PR's proposal allows us to get the static typing and the right accessing shape throughout the codebase, I'd then work on the codebase changes required to get to a plain dataclass.

aapeliv (changes requested):
i think this is fine, but i'm just wondering if we could keep them uppercase to make it very clear they are fixed constants

tristanlabelle:
@aapeliv 

> i think this is fine, but i'm just wondering if we could keep them uppercase to make it very clear they are fixed constants

We modify those fixed constants in tests. The Python [style guide](https://peps.python.org/pep-0008/#constants) reserves uppercase for actual constants. I think we can better achieve a stronger fixedness with a frozen dataclass, but there's more work than fits in this PR to get there.

Waiting for your call.

kalvdans:
I think global constants should be uppercase as well, even if they are mocked in unit tests

aapeliv:
I'm very glad to see the config become strongly typed. Let's keep them uppercase, it is clearer that way.

Will you follow up with PRs updating all call sites afterwards one by one?

tristanlabelle:
@aapeliv , Ok I'll switch to upper case. And yes I'll update all current call sites in one go. This kind of changes are PR-race prone since new files can add new config references as I'm trying to update them, so it'll probably take a second pass a week later before removing the dictionary-style interface seems safe.

aapeliv:
Thank you! Sounds good

tristanlabelle:
Rebased and updated to ALL_CAPS naming. Ready for your ✅, @aapeliv

aapeliv (approved):
thanks, lgtm!

PR #8938, opened 2026-06-01
Dashboard 2.0 - Trips, Stays and design tidying

### LINKED ISSUE #8421: Dashboard 2.0

nabramow:
- Adds upcoming trips and upcoming stays section at top
- Adds icons to all the sections
- Adds scroll arrows to the events section instead of "see all"
- Makes dashboard news alerts smaller on mobile
- Fades in the reminder cards at the edges for a nicer looking effect

Closes #8421 

Screenshot 2026-06-01 at 23 55 29

## Testing
- Get some upcoming stays and trips and have a look at them in mobile and desktop view

**Web frontend checklist**

## For maintainers


app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -0,0 +1,558 @@
...
+
+  const dateRange = localizeDateTimeRange(fromDate, toDate, {
+    timezone: UTC_TIMEZONE,
+    locale,
+    includeTime: false,
+    abbreviate: true,
+  });
+
+  const cityName = (city: string | undefined) =>
+    city?.split(",")[0].trim() ?? "—";
```
nabramow: I am not sure if this is gonna be 100% reliable for all the cities but it's so long otherwise I just wanted a way to capture the important bit.

aapeliv:
could we swap it "Chris \ In Brisbane" to highlight the host instead of the city as primary? also the location pin is cute but we should show the user's avatar i think?

app/web/features/dashboard/UpcomingStays.tsx
aapeliv: i think we should not split it on the frontend

c-kreuz:
I'm with Aapeli. Maybe do the location pin in line (like you do with the events)

app/web/features/dashboard/locales/en.json
```diff
@@ -73,6 +73,22 @@
     "location_online_label": "Online",
     "today_label": "Today"
   },
+  "stays": {
+    "upcoming_trips_header": "Your upcoming trips",
+    "upcoming_guests_header": "Upcoming guests",
+    "today_label": "Today",
+    "tomorrow_label": "Tomorrow",
+    "in_n_days_label_one": "In {{count}} day",
+    "in_n_days_label_other": "In {{count}} days",
```
tristanlabelle: FYI [Intl.RelativeTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat) can also provide those translated strings. No big deal if it's only days, but if you needed to show more units it'd be better to rely on the API.

app/web/features/dashboard/locales/en.json
```diff
@@ -73,6 +73,22 @@
...
+    "next_week_label": "Next week",
+    "night_count_one": "{{count}} night",
+    "night_count_other": "{{count}} nights",
+    "with_name": "with {{name}}",
+    "from_city": "from {{city}}",
+    "plan_trip_label": "Plan a trip",
```
tristanlabelle: Unused string? It also looks like more of a link than a label. Differentiating between something actionable and a label helps with choosing the right translation.

tristanlabelle (commented):
Super partial review, it's late over here. Hopefully I get time tomorrow.

jesseallhands (changes requested):
Looks fantastic! I love it overall, just have a few suggestions:

1. "Upcoming trips" and "upcoming guests" and "your upcoming events" should be hidden by default if they are empty OR there should be a toggle at the top of the dashboard to hide empty content. Having all these empty sections will be super annoying to have constantly there for people who only primary use the platform for one thing (eg only hosting, only events, etc).
2. The wording should be consistent, it should say "Your upcoming guests"
3. Not sure if it's already there (I don't have enough cards) but swiping the carosel cards should be possible so you don't have to use the arrows (keeping the arrows is fine though to indicate you can see more).
4. When clicking "become a host" it should jump to the "preferences" section where you can edit your hosting status.

app/web/features/dashboard/locales/en.json
nabramow: Ah yes it is unused, was trying out an idea and forgot to remove thanks!

tristanlabelle:
> could we swap it "Chris \ In Brisbane" to highlight the host instead of the city as primary?

Minor: "In Brisbane" in some rare cases will cause localization issues as the article may depend on the gender of the place. If just "Brisbane" provides enough context, we can use that to avoid issues. E.g. in French you'll have "à Paris", "à Berlin" but "au Havre", "au Liechtenstein".

app/web/features/dashboard/CommunityEvents.tsx
```diff
@@ -20,45 +18,68 @@ import EventListRow, {
...
-        <StyledLink href={`${eventsRoute}#discover`}>
-          {t("dashboard:events.see_all_link")}
-        </StyledLink>
+        <div>
+          <IconButton
+            size="small"
+            onClick={() => setPage((p) => p - 1)}
+            disabled={!hasPrev}
+            color={hasPrev ? "primary" : "default"}
+            aria-label={t("dashboard:discussions.prev_page_label")}
```
tristanlabelle: Minor: I'd expect "label" to be a piece of text that describes a nearby control. Naming it "a11y" could help translators find the string in the page, and it hints that the translated string is not length-constrained.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: @aapeliv How do you want to do it then? As far as I can tell there's no city-specific field on the backend?

tristanlabelle: It definitely won't be reliable if/when we localize the city names.

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -0,0 +1,558 @@
...
+
+function filterUpcoming(
+  requests: HostRequest.AsObject[],
+): HostRequest.AsObject[] {
+  const today = dayjs().startOf("day");
+  return requests
+    .filter(
+      (r) =>
+        ACCEPTED_STATUSES.has(r.status) &&
+        dayjs(r.fromDate).diff(today, "day") >= 0,
```
tristanlabelle: This comparison is not safe, fromDate needs to be interpreted in the host's timezone to be comparable with another point in time.

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -0,0 +1,558 @@
...
+  const isHost = authState.userId === hostRequest.hostUserId;
+  const otherUserId = isHost
+    ? hostRequest.surferUserId
+    : hostRequest.hostUserId;
+  const { data: otherUser, isLoading } = useLiteUser(otherUserId);
+
+  const fromDate = dayjs.tz(hostRequest.fromDate, UTC_TIMEZONE);
+  const toDate = dayjs.tz(hostRequest.toDate, UTC_TIMEZONE);
+  const nights = toDate.diff(fromDate, "day");
+  const daysUntil = fromDate.diff(dayjs().startOf("day"), "day");
```
tristanlabelle: This comparison is not safe, "fromDate" is only an accurate point in time if interpreted in the host's timezone, otherwise we can't compare it with now/today. Otherwise you're correct that parsing and formatting timezones need to match to avoid time-shifting.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: I can just take it out an add an ellipsis for now but looks nicer with just the city

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -0,0 +1,558 @@
...
+            aria-label="Next"
+          >
+            <ArrowForward fontSize="small" />
+          </IconButton>
+        </div>
+      </SectionHeader>
+
+      {isLoading ? (
+        <Track>
+          {[0, 1, 2].map((i) => (
```
tristanlabelle, line 147: Document why 3?

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -0,0 +1,558 @@
...
+}: UpcomingStaysWidgetProps) {
+  const scrollerRef = useRef<HTMLDivElement>(null);
+  const [canScrollLeft, setCanScrollLeft] = useState(false);
+  const [canScrollRight, setCanScrollRight] = useState(false);
+
+  const updateScrollState = () => {
+    const el = scrollerRef.current;
+    if (!el) return;
+    setCanScrollLeft(el.scrollLeft > 4);
+    setCanScrollRight(el.scrollLeft < el.scrollWidth - el.clientWidth - 4);
```
tristanlabelle: Document why 4?

tristanlabelle (dismissed):
There are a few tricky bits around timezones, but otherwise no concerns about the code. I look forward to this!

app/web/features/dashboard/UpcomingStays.tsx
tristanlabelle: Thinking more, it's probably safe that if there is a comma, it should be meaningful to break the string there, but we should have an ellipsis fallback for languages that won't use commas.

c-kreuz:
Regarding @jesseallhands 1. :
I disagree that they should be hidden by default, however, I agree that there should be a toggle to hide these sections. We could have a little chevron next to "Your upcoming Guests" and the others, that would unfold or collapse the section.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: For now I just took it out and added an ellipsis. This is what we do everywhere else but yeah I do think we have a need for separating out the city in many places in the app.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: We'd need to do a query for every card to get the host's timezones since it's not stored on the `host_request` in the DB. So I can change this to utc to neutralize a bit but yes there will still be edge cases on both sides where the cards might display a little longer / not long enough for some users.

I think this is not so risky though. If you have a better idea how to handle it given our current constraints let me know. It would be a bit too hacky if I try to query this information on the frontend as it stands as it really should be a backend responsibility IMO.

tristanlabelle: I'll file a bug and tag `i18n`. We'll probably want te return the timezone in the host request object. I  agree it's not a showstopper.

tristanlabelle: @claude if you can file a bug, knock yourself out on this one.

nabramow: @tristanlabelle Wow this is the best use of claude ever ahaha.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: This is just for the placeholder skeleton that shows while the real cards are loading. Three just seemed like a good number visually.

app/web/features/dashboard/UpcomingStays.tsx
nabramow: The width of the client's screen vs the scroll space can be a bit weird around zero so it was just to offset it a bit. I switched it to `Math.round` instead and added a comment explaining the logic.

The original 4 was just an overly conservative guess at a safe threshold.

nabramow: Okay just FYI I made some changes to the logic on the backend which allowed me to take out a bunch of these frontend sorting hacks.

nabramow:
@aapeliv @tristanlabelle @c-kreuz @jesseallhands Okay this is ready for review again with a few notes:

- Let'd do the toggled in a follow up PR as there are some things to discuss like should we store the last value in cookies and then what happens if they get requests and cookies hides it in the meantime? Don't wanna block that for this and it's a bit scope creep

- I'm making an executive decision for now not to hide empty sections. We can discuss more later if anyone is really unhappy about that but it's not our existing pattern and we do have CTAs there.

Everything else should be addressed.

c-kreuz:
I'm happy with your executive decision.
The rest looks good to me as well and I'm happy to ship as is.

2 notes though:

I feel that "Your upcoming Guests" should come first and "Your upcoming Stays" second. Any thoughts on that?
Reasoning: Someone relying on you is more important than someone expecting you as a guest.

We have everything framed except the discussions. Not sure if they should also be framed to make it consistent.

aapeliv (approved):
lfg

PR #8942, opened 2026-06-02
Web: mount feature flag provider at app root

aapeliv:
Promote the GrowthBook provider from the badges page to `_app.tsx` so flag loading begins at app boot for every page (logged-in and logged-out), and drop the per-page `<WithFlags>` wrappers. Uses the standard `@growthbook/growthbook-react` hooks (`useFeatureIsOn`, `useFeatureValue`, `useGrowthBook`) so the bespoke context-direct fallback in `experimentation.ts` goes away. Renamed `WithFlags` → `FeatureFlagProvider` to match the mobile component and slot in alongside the other root providers. Follow-up to 08cbaf0aba2 and addresses review feedback on #8917.

Deliberately not using `<FeaturesReady>`: instead, `AppRoute`'s existing private-page spinner gate is extended with `useGrowthBook().ready` — same UX as today's auth-loading spinner, no second spinner, and `ready` is the SDK-recommended alternative per the docs.

## Testing

- `yarn lint`, `yarn typecheck`, `yarn format` all clean.
- `yarn test` — 841 passed, 1 skipped (same as develop). `test/hookWrapper.tsx` now wraps tests in a `GrowthBookProvider` to mirror prod.

**Web frontend checklist**

## For maintainers


app/web/experimentation.ts
```diff
@@ -19,23 +18,14 @@ export function shouldPassAllGates(): boolean {
...
 }
 
-function useGrowthBookOrUndefined(): GrowthBook | undefined {
-  return useContext(GrowthBookContext).growthbook as GrowthBook | undefined;
-}
-
 export function useGate(gateName: string): boolean {
-  const growthbook = useGrowthBookOrUndefined();
-  return shouldPassAllGates() || (growthbook?.isOn(gateName) ?? false);
+  const isOn = useFeatureIsOn(gateName);
```
nabramow: Nit - can we make it clear what's on with this variable name?

app/web/features/experimentation/FeatureFlagProvider.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function FeatureFlagProvider({
+  children,
+}: {
+  children: ReactNode;
+}) {
+  const { authState } = useAuthContext();
+
+  useEffect(() => {
+    void loadFeatureFlags();
```
nabramow: This `void` pattern looks a bit sus. It looks like they suggest using `growthbook.init` now.

Could this setup might be from a deprecated older version? See their setup docs here:

https://docs.growthbook.io/lib/react#quick-usage

app/web/components/AppRoute.tsx
```diff
@@ -144,7 +146,7 @@ function AppRoute({
 
   return (
     <ErrorBoundary>
-      {isPrivate && (!isMounted || !isAuthenticated) ? (
+      {isPrivate && (!isMounted || !isAuthenticated || !featuresReady) ? (
```
nabramow: I think instead of making this `featuresReady` variable you can just wrap what's supposed to load in a `<FeaturesReady>` component, see here:

https://docs.growthbook.io/lib/react#waiting-for-features-to-load

That looks slightly cleaner.

app/web/test/hookWrapper.tsx
```diff
@@ -8,6 +10,13 @@ import { theme } from "theme";
 
 import AuthProvider from "../features/auth/AuthProvider";
 
+// In prod the GrowthBook provider sits at the app root (see pages/_app.tsx), so any component using
+// flag hooks expects one. Mirror that here with a fresh, uninitialized SDK; flag hooks will return
+// the caller's default and useGrowthBook().ready stays false, which is what tests want.
```
darrenvong: This comment seems misleading and points to something not great - if `.ready` stays false, wouldn't that cause the tests to fail since nothing would load past the loading state?

app/web/components/AppRoute.tsx
darrenvong: I kept this as the `FeaturesReady` only deals with GrowthBook's loading state, and so we'd still need the other conditions and not sure the extra nesting would make this any cleaner to read?

darrenvong (approved):

PR #8945, opened 2026-06-02
Mobile: platform-capabilities module + committed fingerprint baseline

aapeliv:
Sets up the infrastructure to decouple the OTA `runtimeVersion` from the auto-computed Expo fingerprint, so a JS bundle published over the air can run against any sufficiently-recent store binary and gracefully degrade when a native capability is missing — without forcing every transitive native change to wait for a store rollout.

Two pieces, both intentionally landed before any change to `runtimeVersion`:

**1. New `platform-capabilities` host module** (`app/mobile/modules/platform-capabilities/`)

iOS Swift + Android Kotlin sources hardcode the registry the native build supports:

- `capabilityPlatformVersion` (monotonic int — coarse "I require ≥ N" gate)
- `capabilities: string[]` (named registry — fine-grained per-feature gate)

JS reads both through `service/platformCapabilities.ts`, which also exposes `capabilityAvailable(name)` (a TS set-lookup over the cached list — no extra bridge calls). The same service exports `fingerprint`, today sourced from `Updates.runtimeVersion`; when `runtimeVersion` later flips to a static string, its value will move to a native-stamped one and consumers won't change.

Wired into:
- Sentry `initialScope` (capability platform version as a tag, full registry as a `platform_capabilities` context block)
- The triple-tap debug toast (`Capability platform: vN (M caps)` line)

**2. Committed fingerprint baseline + CI gate**

`app/mobile/.fingerprint-baseline-{ios,android}` hold the blessed per-platform fingerprint. `scripts/check-fingerprint-baseline.sh` compares working tree vs. baseline and fails on mismatch with explicit fix-up instructions (revert, or bump the baseline + capability registry in the same PR). New GitLab job `test:mobile-fingerprint-baseline` runs on every mobile-touching pipeline. `native-build.sh` and `devtool-build.sh` re-run the same check before invoking EAS, so develop refuses to ship an unblessed store binary even if a test job is somehow bypassed. Emergency escape hatch: `ALLOW_FINGERPRINT_BASELINE_MISMATCH=true`.

**Not in this PR (deferred until this module is in a deployed native build):**

- Flipping `runtimeVersion` from `{ policy: "fingerprint" }` to a static `"couchers-1"`
- Updating OTA jobs to publish against the static runtime
- A config plugin to stamp the true fingerprint into the binary for `service/platformCapabilities.ts` to read

## Testing

Locally (in `app/mobile`):
- `npx tsc --noEmit` clean
- `npm run lint` clean
- `npm run format` no-op
- `npm test` — 148 tests pass
- `bash scripts/check-fingerprint-baseline.sh ios` / `android` — both match baseline
- Manual mismatch test: edited a baseline to a wrong value → script exits 1 with the right error message
- `ALLOW_FINGERPRINT_BASELINE_MISMATCH=true` bypasses cleanly
- Confirmed JS-only edits leave the fingerprint unchanged (so the gate doesn't false-alarm on bundle-only diffs)

Reviewers can replicate by changing any line under `app.config.js` or adding a new Expo plugin and re-running `bash scripts/check-fingerprint-baseline.sh ios` — the script will print the new fingerprint to copy into the baseline file.


app/mobile/scripts/devtool-build.sh
```diff
@@ -43,6 +43,8 @@ json_field() { node -e 'let s="";process.stdin.on("data",d=>s+=d).on("end",()=>{
 CURRENT="$(npx expo-updates fingerprint:generate --platform "$PLATFORM" 2>/dev/null | json_field '.hash')"
 echo "current $PLATFORM fingerprint:    $CURRENT"
 
+bash scripts/check-fingerprint-baseline.sh "$PLATFORM"
```
aapeliv, line 46: Use the needs thing in gitlab

app/mobile/scripts/native-build.sh
aapeliv, line 60: Now superseded

app/mobile/service/platformCapabilities.ts
```diff
@@ -0,0 +1,14 @@
...
+  capabilityAvailable,
+  capabilityPlatformVersion,
+} from "platform-capabilities";
+
+export { allCapabilities, capabilityAvailable, capabilityPlatformVersion };
+
+// While runtimeVersion uses the fingerprint policy this equals
+// Updates.runtimeVersion; after the flip to a static runtimeVersion this will
+// read from a native-stamped value.
+export const fingerprint: string = Updates.runtimeVersion ?? "unknown";
```
aapeliv, line 14: What?

app/mobile/service/sentry.ts
```diff
@@ -71,6 +76,10 @@ if (sentryEnabled) {
           isEmbeddedLaunch,
           createdAt,
         },
+        platform_capabilities: {
```
aapeliv, line 79: Sentry and all the version reporting needs actual fingerprint

PR #8948, opened 2026-06-02
Backend: resolve feature flags from a local file in dev

aapeliv:
Replaces the `EXPERIMENTATION_ENABLED` / `EXPERIMENTATION_PASS_ALL_GATES` config pair with a single dev-only knob: `FEATURE_FLAGS_FILE_OVERRIDE_PATH`.

It selects where flags come from:
- **Set** (dev only): flags are resolved entirely from that JSON file — a listed key gives its value, an unlisted key falls through to its in-code default, and GrowthBook is never contacted.
- **Empty**: flags are resolved from GrowthBook (the production path), as before.

Setting the path while `DEV=0` is rejected at config check, so it can never shadow GrowthBook in production. GrowthBook's client key / cache path are required whenever the path is empty.

This lets devs pin flag values locally (and run fully offline) without spinning up or reconfiguring GrowthBook.

### Details
- `FEATURE_FLAGS_FILE_OVERRIDE_PATH` maps flag keys to forced values; the file is read once and cached per path. Restart the backend to pick up changes.
- The dev env (`backend.dev.env`) now points at a checked-in `feature-flags.dev.json`. Flags gating features that need real external credentials (Stripe, Iris ID, MyPostcard, SMS, Listmonk, reCAPTCHA) are off there, since dev only has placeholder keys.
- Removed the `EXPERIMENTATION_*` toggles entirely; `get_global_boolean_value` and the per-request context methods go through one evaluation path.
- File-override evaluations don't emit the feature-flag metric (it's just noise for local dev); GrowthBook evaluations still record source + value.

### Testing
- New `flags` pytest fixture for file-override mode: `flags.set_boolean("test_growthbook_integration", False)` (also `set_string` / `set_integer` / `set_float` / `set_object`). Seeds from shared test defaults so flipping one flag leaves the rest on. The existing `feature_flags` fixture still covers GrowthBook rollouts/experiments.
- Tests in `test_experimentation.py` cover file-override resolution, fall-through to in-code defaults, file loading + non-object rejection, and that setup skips GrowthBook in file-override mode. `test_config.py` asserts the dev-only prod guard.
- `uv run pytest src/tests/test_experimentation.py src/tests/test_config.py` — 29 passed.
- `make format` and `make mypy` — clean (pre-existing unrelated mypy errors from stale proto stubs aside).

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/experimentation.py
+++ b/app/backend/src/couchers/experimentation.py
@@ -133,8 +133,6 @@ def _refresh_loop() -> None:
 @cache
 def _load_local_flags(path_str: str) -> dict[str, Any]:
     """Read and validate the dev-only override file; cached per path."""
-    if not path_str:
-        raise ValueError("FEATURE_FLAGS_FILE_OVERRIDE_PATH is empty")
     loaded = json.loads(Path(path_str).read_text())
     if not isinstance(loaded, dict):
         raise ValueError(f"Feature flag override file {path_str} must contain a JSON object")
--- a/app/backend/src/couchers/experimentation.py
+++ b/app/backend/src/couchers/experimentation.py
@@ -1,9 +1,8 @@
 """
 Feature flag and experimentation framework.
 
-FEATURE_FLAGS_FILE_OVERRIDE_PATH selects the flag source: when set (dev only) flags come from that
-JSON file and GrowthBook is not contacted; when empty they come from GrowthBook. Unknown flags fall
-through to their in-code default either way.
+FEATURE_FLAGS_FILE_OVERRIDE_PATH (dev only) reads flags from a JSON file instead of GrowthBook;
+unknown flags fall through to their in-code default either way.
 
 Two ways to evaluate a flag:
   - Per-user/request: use the CouchersContext methods (context.get_boolean_value, get_string_value,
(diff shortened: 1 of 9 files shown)


WouldYouKindly (dismissed):

app/backend/src/tests/conftest.py
```diff
@@ -224,11 +243,13 @@ def testconfig():
     config["RECAPTHCA_API_KEY"] = "..."
     config["RECAPTHCA_SITE_KEY"] = "..."
 
-    config["EXPERIMENTATION_ENABLED"] = False
-    config["EXPERIMENTATION_PASS_ALL_GATES"] = True
+    # File-override mode; gates forced True via the stubbed loader below. Tests needing GrowthBook use `feature_flags`.
+    config["FEATURE_FLAGS_FILE_OVERRIDE_PATH"] = "feature-flags.dev.json"
```
tristanlabelle, line 245: Can we do this more globally than in `testconfig`, which I think requires opting-in? It would be bad for reproducibility if any test ever got feature flags from GrowthBook, so I feel like the save/recover logic is not helpful.

app/backend/src/couchers/experimentation.py
```diff
@@ -129,22 +130,32 @@ def _refresh_loop() -> None:
         # On a failed fetch, keep last-known-good state and retry next tick; the staleness metric climbs.
 
 
+@cache
+def _load_local_flags(path_str: str) -> dict[str, Any]:
+    """Read and validate the dev-only override file; cached per path."""
+    loaded = json.loads(Path(path_str).read_text())
```
tristanlabelle: In the various places we set the path, we always use a relative path. That means relative to the working directory. I think we need a stronger definition, i.e. this should make the path absolute relative to whichever standard location.

tristanlabelle (dismissed):
Thanks! Much better for local dev.

app/backend/src/tests/conftest.py
aapeliv: i think this is kind of the standard way for us to do this. the growthbook api urls aren't set in tests, so there is no risk of going to the wrong place

app/backend/src/couchers/experimentation.py
aapeliv: fixed

PR #8957, opened 2026-06-02
Backend/emails: Port activeness_probe and new_blog_post

tristanlabelle:
Port those two email kinds to the new email system, just like every other. Nothing too special here, except that I updated the `activeness_probe` latest release message to be slightly more evergreen, which will require an additional checklist entry in #8809

## Testing
`uv run dump-emails` + existing test coverage for all emails.

image

image

**Backend checklist**

## For maintainers


app/backend/src/couchers/constants.py
```diff
@@ -6,6 +6,12 @@
 # community guidelines version
 GUIDELINES_VERSION = 1
 
+BLOG_URL = "https://couchers.org/blog"
+
+# When updating this, also update the "activeness_probe" notification strings in
+# src/couchers/email/locales/en.json
+LATEST_RELEASE_BLOG_URL = "https://couchers.org/blog/2026/05/25/couchers-spring-release"
```
tristanlabelle, line 11: Self-blocking TODO: Update checklist in https://github.com/Couchers-org/couchers/pull/8809 after it merges.

app/backend/src/couchers/email/locales/en.json
```diff
@@ -37,6 +37,16 @@
...
     "redelete_instructions": "If you change your mind, you can delete your account at any time."
   },
+  "activeness_probe": {
+    "subject": "Are you still open to hosting on Couchers.org?",
+    "body": "We noticed you haven't logged in for a while and wanted to check if you're still open to hosting surfers.",
+    "instructions_days_one": "To keep your \"Can Host\" status, please log back in within the next {{count}} day. Otherwise, we'll change your hosting status to \"Can't Host.\"",
+    "instructions_days_other": "To keep your \"Can Host\" status, please log back in within the next {{count}} days. Otherwise, we'll change your hosting status to \"Can't Host.\"",
+    "login_action": "Log back in to respond",
+    "encouragement": "These check-ins help surfers find active hosts more easily and keep response times accurate.",
+    "latest_release": "In case you've missed it, we recently released version {{version}}. Some of the recent updates include new mobile app on iOS and Android, improved dashboard, overhauled messaging and new event features.",
```
tristanlabelle, line 48: Self-blocking TODO: Update checklist in https://github.com/Couchers-org/couchers/pull/8809 after it merges.

nabramow:
@tristanlabelle If "version unknown" only a staging issue? Just wanna make sure we're not gonna see that on prod!

tristanlabelle:
@nabramow This `unknown` is from the [default config](https://github.com/Couchers-org/couchers/blob/e5971c557c6c81670203d53f5b0836da60487eae/app/backend/src/couchers/config.py#L21). The screenshot is from the email-dumping script, which doesn't need a full staging or prod environment setup.

aapeliv (approved):

aapeliv (changes requested):
actually, that `version` string is something like `v1.3.18840`. i think we should have another short versions, `v1.3` for the email?

tristanlabelle:
Rebased, merged with other email changes, and added logic to filter down `v1.3.18840` to `1.3`.

PR #8959, opened 2026-06-02
Backend: Fix only_active and add is_statuses to HostRequestReq

nabramow:
I realized the `only_active` in `HostRequestReq` was `<=` when it should be `>=` while working on the dashboard 2.0. I'm also adding `is_statuses` to allow the frontend to send which statuses it wants returned as I need that too.

## Testing

**Backend checklist**

## For maintainers


nabramow:
@tristanlabelle Realized I don't need all that `from_date` logic on the frontend of #8938 if the `only_active` filter actually works ;-).

PR #8963, opened 2026-06-03
Add host request sort by and use backend sort logic in messages

### LINKED ISSUE #8417: Messages: Only show load more button if there are more messages
The "Load more" button is showing even when there are no more messages. We should conditionally render it only if there are more messages to load.

Image

nabramow:
This PR does several things:

* Adds a sort by option to the `HostRequestReq` and adjusts what's uses as the next page token.
* Swaps this token field out in the frontend (otherwise there would be errors so this PR needs to be combined)
* Switches the `AllMessagesTab` to actually use the backend sorting for hosting/surfing, it was doing frontend logic before and turns out we do have backend logic for this. This should fix up the issues in #8417.

Closes #8417 

## Testing
- `cd app/backend` and `make protos`
- `cd app/web`, then `cp .env.localdev .env.development` then `yarn start`
- `docker compose up --build`
- Go to locahost and log in as aapeli dummy user
- Go to messages, should look the same

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/proto/requests.proto
```diff
@@ -107,8 +107,13 @@ message SetHostRequestArchiveStatusRes {
...
 }
 
+enum HostRequestSortBy {
+  HOST_REQUEST_SORT_BY_LAST_ACTIVE = 0;
+  HOST_REQUEST_SORT_BY_FROM_DATE = 1;
+}
+
 message ListHostRequestsReq {
-  int64 last_request_id = 1;
+  string page_token = 1;
```
aapeliv: you can't change the type of an existing field

app/proto/requests.proto
```diff
@@ -122,11 +127,13 @@ message ListHostRequestsReq {
...
   // If non-empty, only return requests with one of these statuses.
   repeated org.couchers.api.conversations.HostRequestStatus status_in = 7;
+
+  HostRequestSortBy sort_by = 8;
 }
 
 message ListHostRequestsRes {
   repeated HostRequest host_requests = 1;
-  int64 last_request_id = 2;
+  string next_page_token = 2;
```
aapeliv: same

app/backend/src/couchers/servicers/requests.py
```diff
@@ -437,8 +437,26 @@ def ListHostRequests(
...
         ).where(message_2.id == None)
 
-        if request.last_request_id != 0:
-            statement = statement.where(Message.id < request.last_request_id)
+        sort_by_from_date = request.sort_by == requests_pb2.HOST_REQUEST_SORT_BY_FROM_DATE
+
+        if sort_by_from_date:
+            if request.page_token:
+                token_date_str, token_conv_id_str = request.page_token.split(":")
+                token_date = parse_date(token_date_str)
```
aapeliv, line 445: you can just cheat and make the page token something like `yyyymmdd * 100_000_000 + req_id`. it's a hack but it works. just leave a comment

aapeliv (changes requested):

app/proto/requests.proto
nabramow: I reserved the old field instead as when I looked at it that seems the way you're supposed to do it? Let me know if not though!

app/backend/src/couchers/servicers/requests.py
tristanlabelle: A pair of pack/unpack functions would be clearer than a comment.

aapeliv:
@claude double check that the pagination logic is sound

aapeliv (approved):
lgtm.

technically it's best practice to reserve enum value 0 for unspecified in proto; but i think this has gone through so many reviews that we can ignore that :)

PR #8964, opened 2026-06-03
Frontend/profile: Validate names like during signup

### LINKED ISSUE #8933: Unable to edit profile
# Unable to edit profile 
## Description
I want update my profile now that I will use it more, but it won't update, it just say name not supported. No more information than that though, so don't know what the issue is. This is both on the app and website 

## Diagnostics
**Backend version**: `develop-5eb8cc8d`
**Frontend version**: `develop-5eb8cc8d`
**User Agent**: `Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Mobile Safari/537.36`
**Locale**: `en`
**Screen resolution**: 384x718
**Page**: https://couchers.org/dashboard
**User**: [@margunn](https://couchers.org/user/margunn) (2483) / `oj9i8Q3LbNN8`

tristanlabelle:
During signup we validate the name field on the frontend, but we didn't when later editing the profile, however the backend still would do the validation.

There's at least one user (#8933) who somehow had a bad character in their name (trailing space), which stopped them from editing their profile because that would fail validation on the backend. With this change they'll see a local error and can figure out to fix it locally.

Fixes #8933

## Testing
Added a test

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/edit/EditProfile.tsx
+++ b/app/web/features/profile/edit/EditProfile.tsx
@@ -323,7 +323,7 @@ export default function EditProfileForm() {
     getValues,
   } = useForm<EditProfileFormValues>({
     shouldFocusError: true,
-    mode: "onTouched",
+    mode: "onBlur",
     values: initialFormValues,
     resetOptions: { keepDirty: true },
   });
--- a/app/web/features/profile/edit/EditProfile.tsx
+++ b/app/web/features/profile/edit/EditProfile.tsx
@@ -45,6 +45,11 @@ import {
   useUnsavedChangesWarning,
 } from "utils/hooks";
 import { useIsNativeEmbed } from "utils/nativeLink";
+import {
+  nameMaxLength,
+  nameMinLength,
+  nameValidationPattern,
+} from "utils/validation";
 
 import {
   ABOUT_ME_MIN_LENGTH,
@@ -568,13 +573,27 @@ export default function EditProfileForm() {
               <FieldGroup>
                 <StyledProfileTextInput
                   id="name"
-                  {...register("name", { required: true })}
+                  {...register("name", {
+                    required: t("auth:basic_form.name.required_error"),
+                    minLength: {
+                      value: nameMinLength,
+                      message: t("auth:basic_form.name.min_length_error"),
+                    },
+                    maxLength: {
+                      value: nameMaxLength,
+                      message: t("auth:basic_form.name.max_length_error"),
+                    },
+                    pattern: {
+                      message: t(
+                        "auth:basic_form.name.invalid_characters_error",
+                      ),
+                      value: nameValidationPattern,
+                    },
+                  })}
                   label={t("profile:edit_profile_headings.name")}
                   defaultValue={user.name}
-                  error={!!errors.name}
-                  helperText={
-                    errors.name ? t("profile:edit_profile_name_required") : ""
-                  }
+                  error={!!errors?.name?.message}
+                  helperText={errors?.name?.message ?? " "}
                 />
               </FieldGroup>
 
--- a/app/web/features/profile/edit/EditProfile.tsx
+++ b/app/web/features/profile/edit/EditProfile.tsx
@@ -278,6 +278,7 @@ export default function EditProfileForm() {
     useState(false);
   const [showSuccessToast, setShowSuccessToast] = useState(false);
   const galleryEditorRef = useRef<HTMLDivElement>(null);
+  const hasInitialized = useRef(false);
 
   const {
     control,
@@ -297,8 +298,11 @@ export default function EditProfileForm() {
 
   // Reset form with user data when user and data are loaded
   // This allows only showing save bar once something changes
+  // hasInitialized prevents background refetches from overwriting in-progress edits
   useEffect(() => {
     if (user && languages && regions) {
+      if (hasInitialized.current) return;
+      hasInitialized.current = true;
       reset(
         {
           name: user.name,
@@ -584,10 +588,10 @@ export default function EditProfileForm() {
                       message: t("auth:basic_form.name.max_length_error"),
                     },
                     pattern: {
+                      value: nameValidationPattern,
                       message: t(
                         "auth:basic_form.name.invalid_characters_error",
                       ),
-                      value: nameValidationPattern,
                     },
                   })}
                   label={t("profile:edit_profile_headings.name")}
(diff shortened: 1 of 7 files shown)


app/web/features/profile/edit/EditProfile.tsx
nabramow:
```suggestion
    shouldFocusError: true,
    mode: 'onTouched'
```

This will surface the error without having to submit first, which I think it a bit more user friendly.

app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -292,8 +298,11 @@ export default function EditProfileForm() {
 
   // Reset form with user data when user and data are loaded
   // This allows only showing save bar once something changes
+  // hasInitialized prevents background refetches from overwriting in-progress edits
   useEffect(() => {
     if (user && languages && regions) {
+      if (hasInitialized.current) return;
+      hasInitialized.current = true;
```
nabramow: Ooo okay this hack is making me realize my hack from before I was familiar with react-hook-forms shouldn't probably be fixed so we don't need this.

A better way to do this is in the `useForm` is to rely on the `values` and `resetOptions`:

https://react-hook-form.com/docs/useform#values
https://react-hook-form.com/docs/useform/reset

Basically I would replace lines 300-363 with this in `EditProfile.tsx`:

```typescript
  const formValues =
    user && languages && regions
      ? {
          name: user.name,
          pronouns: user.pronouns,
          hometown: user.hometown,
          occupation: user.occupation,
          education: user.education,
          hostingStatus: user.hostingStatus,
          meetupStatus: user.meetupStatus,
          fluentLanguages: user.languageAbilitiesList
            .map((ability) => languages[ability.code] || "")
            .filter(Boolean),
          regionsVisited: user.regionsVisitedList
            .map((region) => regions[region] || "")
            .filter(Boolean),
          regionsLived: user.regionsLivedList
            .map((region) => regions[region] || "")
            .filter(Boolean),
          aboutMe: user.aboutMe,
          thingsILike: user.thingsILike || DEFAULT_HOBBIES_HEADINGS,
          additionalInformation: user.additionalInformation,
          location: {
            city: user.city,
            lat: user.lat,
            lng: user.lng,
            radius: user.radius,
          },
        }
      : undefined;

  const {
    control,
    register,
    handleSubmit,
    setValue,
    reset,
    formState: { errors, isDirty, isSubmitted },
    watch,
    getValues,
  } = useForm<EditProfileFormValues>({
    shouldFocusError: true,
    mode: "onTouched",
    values: formValues,
    resetOptions: { keepDirtyValues: true },
  });
```

And then switch out to use the the same logic for the values in `EditHostingPreferences.tsx` and remove the `hasInitialized` logic.

Not to give you more work but just so we're not building a hack on top of a hack. That should better work with react-hook-form rather than hacking around it.

app/web/utils/validation.ts
```diff
@@ -1,5 +1,7 @@
 // taken from backend
 export const nameValidationPattern = /^[\p{L}'-]+(\s+[\p{L}'-]+)*$/u;
+export const nameMinLength = 2;
+export const nameMaxLength = 100;
```
nabramow, line 4: Just checking but this matches the backend yeah?

app/web/features/profile/edit/EditProfile.tsx
tristanlabelle: I suspected there was a better way to go about it. Thanks for pointing it out!

tristanlabelle: Ok I used the `useForm` `values`-based initialization and that works. I'm less clear about

> switch out to use the the same logic for the values in EditHostingPreferences.tsx

If you mean this kind of field access:

```ts
    const userToFormValues = (user: User.AsObject): HostingPreferenceData => ({
      lastMinute: !!user.lastMinute?.value,
      wheelchairAccessible: !!user.wheelchairAccessible?.value,
      campingOk: !!user.campingOk?.value,
      acceptsKids: !!user.acceptsKids?.value,
```

Then I think it doesn't apply because the proto types are declared differently

```proto
  string username = 2;
  string name = 3;
  string city = 4;
  string hometown = 100;

  google.protobuf.BoolValue has_kids = 109;
  google.protobuf.BoolValue accepts_kids = 26;
  google.protobuf.StringValue kid_details = 110;
  google.protobuf.BoolValue has_housemates = 111;
  google.protobuf.StringValue housemate_details = 112;
```

app/web/features/profile/edit/EditProfile.tsx
tristanlabelle: I switched to `onBlur` since that's what we use everywhere. There is no use of `onTouched`.

app/web/utils/validation.ts
tristanlabelle: Yes. I extracted constants on the backend as well in the last iteration.

tristanlabelle:
@nabramow Ready for re-review. Let me know about the `onTouched` and form values logic.

app/web/features/profile/edit/EditProfile.tsx
nabramow: Ah I just meant since we're fixing the hack in `EditProfile.tsx`, let's just fix it in `EditHostingPreferences.tsx` while we're there since they use the same logic.

Or you're saying that doesn't work?

nabramow:
> @nabramow Ready for re-review. Let me know about the `onTouched` and form values logic.

This looks good to me! Just wanna clarify whether it makes sense to refactor the EditHostingPreferences to use `values` instead of the `useEffect` while you're in there then happy to approve.

app/web/features/profile/edit/EditProfile.tsx
tristanlabelle: I had another look into this and I think the pattern is different in `EditHostingPreferences.tsx` because it shows a spinner until the `user` object can be loaded, at which point it extracts the relevant fields as `HostingPreferenceData` and passes them as `useForm(defaultValues:)`. Since the `HostingPreferenceForm` is only used once we have valid data, it doesn't need to use `useForm(values:)` to lazily initialize the form (`defaultValues` and `values` should be equivalent in this case).

Unless we want to remove the spinner, it seems like `EditHostingPreferences.tsx` is fine?

nabramow (approved):
good to go!

PR #8965, opened 2026-06-03
Mobile: More fixes

nabramow:
- Fix issue with Android back gesture only working on first level
- Move profile incomplete translations to profile namespace which is already loaded everywhere to fix inconsistent translations loading
- Clean up incomplete profile dialog to match current primary and secondary button colors
- Update mobile dev script
- Remove the first message snackbar

## Testing

**Web frontend checklist**

## For maintainers


app/web/features/profile/locales/ja.json
```diff
@@ -1 +1,15 @@
-{}
+{
+  "complete_profile_dialog": {
+    "title": "Almost there: please complete your profile to continue!",
```
tristanlabelle: My Japanese isn't the best, but this doesn't look like it. Careful not to merge English strings here. Same for a few other languages.

nabramow: Ooopsie good catch I just removed all the stuff that got added.

PR #8982, opened 2026-06-04
feat: Make donation page publicly accessible

darrenvong:

Renders a public version of the donation page, so users without an account can still access it and email support to donate something if they don't want to tie it to an account.

## Testing

- Check that `/donate` is now publicly accessible
- When accessing `/donate` without logging in, the donation form is not rendered (since the backend API for this is private)
- For logged in users, the page looks the same as before

Screenshot 2026-06-04 at 23 21 53

**Backend checklist**

**Web frontend checklist**

## For maintainers


aapeliv (approved):
thank you!!

PR #8983, opened 2026-06-04
Web/Bugfix/date-proper-capitalization

### LINKED ISSUE #8164: Frontend/i18n: localized datetimes are not capitalized
Outside of English and German, datetimes for other languages are not capitalized by `Intl.DateTimeFormat`, but in many cases the UI context requires it. We'll probably need a "capitalize the first letter" function.

Image

Image

kevinortiz43:
Standalone dates (event cards, the event page, profile "joined" / local time, the reference pill, the month chip) now have their first letter capitalized so they read like the start of a sentence, while dates interpolated into a larger sentence keep their natural casing.

Adds an opt-in `capitalize` flag to the date helpers in `app/web/utils/date.ts`. When set, only the first Unicode **grapheme cluster** is uppercased (via `Intl.Segmenter` + `toLocaleUpperCase`); non-capitalizable first clusters (digits, CJK, …) are left unchanged. Day and month names are **never** capitalized mid-string, and there are **no per-language special cases**. Call sites that render a standalone date opt in via `capitalize: true`; dates interpolated into translated sentences keep `Intl`'s default (sentence-style) casing.

This replaces the earlier per-word `capitalizeDateString` regex, which wrongly capitalized month/day names and relied on a hardcoded Spanish/Portuguese word list.

Closes #8164

## Testing
Updated `app/web/utils/date.test.ts`: by default day/month names stay lowercase (e.g. Spanish "enero"); with `capitalize: true` only the first letter is uppercased — verified that a leading weekday capitalizes ("Sábado…") while the month stays lowercase mid-string ("…enero…"), and that a digit-first date is unchanged. Also verified manually in English, Spanish and Portuguese-Brazil that standalone dates capitalize the first letter and in-sentence dates do not.

**Web frontend checklist**

## For maintainers

image

image

### DIFF
--- a/app/web/features/communities/events/EventCard.tsx
+++ b/app/web/features/communities/events/EventCard.tsx
@@ -141,6 +141,7 @@ export default function EventCard({ event, className }: EventCardProps) {
       locale,
       includeDayOfWeek: true,
       abbreviate: true,
+      capitalize: true,
     },
   );
 
--- a/app/web/features/communities/events/EventPage.tsx
+++ b/app/web/features/communities/events/EventPage.tsx
@@ -441,6 +441,7 @@ export default function EventPage({
                       timezone: BROWSER_TIMEZONE,
                       locale,
                       includeDayOfWeek: true,
+                      capitalize: true,
                     },
                   )}
                 </Typography>
--- a/app/web/features/communities/events/LongEventCard.tsx
+++ b/app/web/features/communities/events/LongEventCard.tsx
@@ -145,6 +145,7 @@ const LongEventCard = ({
       locale,
       includeDayOfWeek: true,
       includeTime: true,
+      capitalize: true,
     },
   );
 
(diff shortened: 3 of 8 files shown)


tristanlabelle (changes requested):
Thanks for looking into this @kevinortiz43 and sorry my ticket wasn't clear enough, I originally wrote it for an audience of myself. 😅

Most languages do not capitalize days of the week or month names, so we don't want to play with those rules. However, latin alphabet languages do quite universally capitalize the first word of a sentence. The `Intl.DateTimeFormat` doesn't have a way to know whether the date is part of a standalone sentence or not, so it defaults to assuming part of a sentence. For example in Spanish you could say "The event is on {{date}}" and you'd want to interpolate "miércoles 25 de abril" -- "miércoles" not "Miércoles".

However we have some places in the UI where we know the date stands alone and is a full sentence into itself so to speak, so it would be more natural for the first character to be upper cased. This needs to be controlled by the call site because it depends whether the date will be formatted in a larger text or not.

We shouldn't need per-language special cases to implement this. If the date is standalone and its first Unicode grapheme cluster is capitalizable, we capitalize it.

kevinortiz43:
OH! LOL okay yeah honestly that makes more sense, okay can do and fix.

kevinortiz43:
@tristanlabelle update done

tristanlabelle (approved):
Wow great stuff! Nice use of Intl.Segmenter

PR #8985, opened 2026-06-05
fix(web): use read-only LL date format for signup birthdate

kevinortiz43:
Adds an optional `format` prop to the shared `Datepicker` component, falling back to the existing locale-derived format when not set. Applies it on the signup birthdate field with a localized `LL` format and a read-only input so users pick the date via the calendar rather than typing it in an ambiguous format.

## Testing
- Ran `eslint --fix` on both changed files — clean.
- Manual: open signup, birthdate field shows `LL`-formatted localized date and input is read-only; calendar picker still works.

**Web frontend checklist**

## For maintainers


tristanlabelle:
Should this PR be closed in favor of #8986 ?

kevinortiz43:
oh yeah this pull request is wrong, I accidentally used the same branch for a different bug in this one.

PR #8986, opened 2026-06-05
fix(web): use read-only LL date format for signup birthdate

kevinortiz43:
Makes the signup **birthdate** field render as a localized, read-only long date that is set only through the calendar picker.

**What changed**
- Added a `pickerInputOnly` mode to the shared `Datepicker`. In this mode the field is display-only: no text entry, clicking anywhere opens the calendar, and a chosen date shows as a long date with the month name (`LL`).
- **Per-language localization:** the birthdate renders month names and the calendar in the selected app language (e.g. German `8. April 1990`, Portuguese `8 de abril de 1990`) instead of always English. MUI's `AdapterDayjs` localizes from `LocalizationProvider`'s `adapterLocale`, so that is now passed in `_app.tsx`, mapped from `i18n.language`; `utils/dayjs` also keeps the dayjs global locale in sync for non-MUI output.
- **Fixed the empty-field placeholder bug:** MUI's masked single input rendered the localized date-format placeholder (e.g. Portuguese `DD de … de AAAA`) in the empty field after focus/blur. For `pickerInputOnly` the field is now a custom read-only field slot (`ReadOnlyDateField`) that is **blank when empty** and formats the value through the picker adapter (so it respects `adapterLocale`). No mask, ever.

## Testing
- `yarn test` on `components/Datepicker.test.tsx` and `features/auth/signup/AccountForm.test.tsx` — 27 passing, covering: `pickerInputOnly` is read-only with no mask placeholder, opens on click, shows the localized long date; `adapterLocale` drives the format (`8. April 1990`); the signup form sets the birthdate via the calendar and submits the correct value.
- AccountForm tests were reworked from typing the date to selecting it via the calendar (the field is read-only now); the blank-birthdate case has its own setup, and the too-young/too-old typing tests were dropped because the picker's `minDate`/`maxDate` (18–120 years) make those dates unselectable.
- `yarn prettier` + `yarn eslint` + `tsc` on the changed files — clean.
- Manual: signup birthdate field is read-only, opens on click, shows no placeholder when empty, and renders the month name in the selected language.

**Web frontend checklist**

## For maintainers

image

image

### DIFF
--- a/app/web/components/Datepicker.tsx
+++ b/app/web/components/Datepicker.tsx
@@ -1,7 +1,11 @@
-import { InputProps } from "@mui/material";
-import { DatePicker } from "@mui/x-date-pickers";
+import { InputProps, TextField } from "@mui/material";
+import {
+  DatePicker,
+  DatePickerProps,
+  usePickerAdapter,
+  usePickerContext,
+} from "@mui/x-date-pickers";
 import { useTranslation } from "i18n";
-import { useState } from "react";
 import { Control, Controller, UseControllerProps } from "react-hook-form";
 import { getMuiDateFormat } from "utils/date";
 import dayjs, { Dayjs } from "utils/dayjs";
@@ -32,6 +36,59 @@ interface DatepickerProps {
   pickerInputOnly?: boolean;
 }
 
+interface ReadOnlyDateFieldProps {
+  className?: string;
+  id?: string;
+  label?: React.ReactNode;
+  error?: boolean;
+  helperText?: React.ReactNode;
+  variant?: "standard" | "outlined" | "filled";
+  ariaLabel: string;
+  inputProps?: InputProps;
+}
+
+// Read-only field used by `pickerInputOnly` pickers. It renders the selected
+// date as a localized long date (month name) and is blank when no date is set,
+// so there is never an editable section mask or format placeholder. Clicking it
+// opens the calendar, which is the only way to set the value.
+const ReadOnlyDateField = ({
+  className,
+  id,
+  label,
+  error,
+  helperText,
+  variant,
+  ariaLabel,
+  inputProps,
+}: ReadOnlyDateFieldProps) => {
+  const pickerContext = usePickerContext<Dayjs | null>();
+  const adapter = usePickerAdapter();
+  const value = pickerContext.value;
+  // Format through the picker adapter so the long date respects the picker's
+  // adapterLocale (e.g. German "8. April 1990"), not just dayjs's global locale.
+  const display = value ? adapter.formatByString(value, "LL") : "";
+
+  return (
+    <TextField
+      ref={pickerContext.triggerRef}
+      className={className}
+      id={id}
+      label={label}
+      error={error}
+      helperText={helperText}
+      variant={variant}
+      fullWidth
+      value={display}
+      onClick={() => pickerContext.setOpen(true)}
+      slotProps={{
+        inputLabel: { shrink: true },
+        htmlInput: { readOnly: true, "aria-label": ariaLabel },
+        input: inputProps,
+      }}
+    />
+  );
+};
+
 const Datepicker = ({
   className,
   control,
@@ -52,7 +109,32 @@ const Datepicker = ({
   pickerInputOnly = false,
 }: DatepickerProps) => {
   const { t, i18n } = useTranslation();
-  const [open, setOpen] = useState(false);
+  const ariaLabel = t("components.datepicker.change_date");
+  const helperNode = (
+    <span data-testid={`${name}-helper-text`}>{helperText}</span>
+  );
+
+  // `pickerInputOnly` swaps the editable masked field for a read-only field slot
+  // (ReadOnlyDateField) so there is never a format placeholder. ReadOnlyDateField
+  // deliberately doesn't conform to MUI's injected field props (it reads value
+  // and open-state from the picker context), so the slot wiring is cast through
+  // `unknown` — MUI's own custom-field pattern requires this.
+  const pickerOnlyProps = {
+    slots: { field: ReadOnlyDateField },
+    slotProps: {
+      field: {
+        className,
+        id,
+        label,
+        error,
+        helperText: helperNode,
+        variant,
+        ariaLabel,
+        inputProps,
+      },
+    },
+  } as unknown as Partial<DatePickerProps>;
+
   return (
     <Controller
       control={control}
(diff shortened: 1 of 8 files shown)


jesseallhands (changes requested):
Sorry I wasn't sure if this is ready for testing yet or not, but I tested in vercel and here are a few changes needed:

1. When I tested, text input is still possible (see screenshot below). We should disallow any text input at all. Clicking on any part of the input box or calendar icon should open the datepicker.

2. There shouldn't be any prefilled text. Currently the prefilled text is `MMMM DD, YYYY`
Image

3. The spacing of the "please enter a real birthdate" helper text below the input box seems off:

Image

app/web/components/Datepicker.tsx
```diff
@@ -23,6 +23,7 @@ interface DatepickerProps {
   testId?: string;
   variant?: "standard" | "outlined" | "filled";
   inputProps?: InputProps;
+  format?: string;
```
tristanlabelle: I suggest abstracting the Mui format and rather making this a "pickerInputOnly" boolean or something like that.

tristanlabelle:
I'm a little bit concerned about whether we can get MUI to show the localized month names since I don't want to rely on dayjs' localization vs the browser-provided, and I didn't find MUI support for browser-provided localization. I can investigate that separately though.

@claude File a bug labeled `i18n` for the issue of localizing textual dates shown by the mui date picker.

> Note: LL currently renders the month name in English regardless of the selected language, since the dayjs locale isn't switched with the app language. This avoids the numeric DD/MM vs MM/DD ambiguity but is not yet per-language localized.

kevinortiz43:
Remove the ability to type in any numbers, remove the placeholder , pickerInputOnly format for the datepicker.tsx, copy. @tristanlabelle  yeah I don't think we could get the months localized with the current mui picker, I did some research on it.

tristanlabelle:
@kevinortiz43 I think we could extend `@mui/x-date-pickers/AdapterDayjs` to override its date formatting method in the specific case of dates formatted using month names, which we don't need to support parsing.

kevinortiz43:
@tristanlabelle  @jesseallhands I have updated this pull request per the recommendations. Please check it out and let me know. I have included screenshots for the localization of months per language

jesseallhands (changes requested):
Tested on mobile web on firefox and chrome in both English and German and seems to be working great and as expected! Awesome!

I tested on desktop in firefox 151.0.3/windows 11 and encountered some buggy behavior. The birthdate input field is empty to start, and when you click it then it shows the datepicker, but if you click elsewhere (outside the datepicker) to get out of the datepicker, then you see text appear in the birthdate input field. In the case of Portuguese then it looks like below:

Image

It shouldn't show any text in the text input box if you click outside the datepicker box. 

Edit: It looks like this behavior also exists on mobile web.

kevinortiz43:
let me see

kevinortiz43:
@jesseallhands okay try now with the porteguese language

jesseallhands (dismissed):
Tested on mobile web and the bug is fixed, now working exactly as expected! Excellent!

app/web/utils/dayjs.ts
```diff
@@ -13,5 +40,51 @@ dayjs.extend(RelativeTime);
...
+    I18N_TO_DAYJS_LOCALE[language.split("-")[0]] ??
+    "en"
+  );
+}
+
+/// Sets dayjs's global locale from an i18n language code, so localized dayjs
+/// output (relative time, durations) matches the app language.
+export function setDayjsLocale(language: string): void {
+  dayjs.locale(i18nToDayjsLocale(language));
+}
```
tristanlabelle: Did you try using `myDayjs.locale(ln).format("LL")`? It seems preferable to me to specify the locale at the formatting site than rely on global state and keeping dayjs' state in sync with the app.

app/web/utils/dayjs.ts
```diff
@@ -1,3 +1,30 @@
...
+import "dayjs/locale/nl";
+import "dayjs/locale/pl";
+import "dayjs/locale/pt";
+import "dayjs/locale/pt-br";
+import "dayjs/locale/ru";
+import "dayjs/locale/sv";
+import "dayjs/locale/tr";
+import "dayjs/locale/uk";
+import "dayjs/locale/zh-cn";
+import "dayjs/locale/zh-tw";
```
tristanlabelle, line 27: I don't know if we have a way to measure the impact of these additional imports on app size or such. FYI @nabramow . So far I could avoid this by relying on `Intl.DateTimeFormat` but I understand that might not work with the picker.

tristanlabelle (dismissed):
I'll rely on Jesse's testing. It looks perfect for functionality but I'd rather we avoid the global dayjs locale state and wonder what's the impact of importing all locales. I'm not too sure how to evaluate that. However, I think I might be able to implement the custom adapter solution afterwards, such that we wouldn't need the dayjs locales.

app/web/utils/dayjs.ts
tristanlabelle: @claude File an i18n-labeled issue assigned to me on using a custom DateAdapter based on Intl.DateTimeFormat to avoid relying on dayjs locales.

kevinortiz43:
Addressed the global dayjs locale review (commit 8f8e7061b7): removed setDayjsLocale + the _app.tsx useEffect; picker localizes via LocalizationProvider adapterLocale; ResponseRateLabel humanize() now uses .locale(i18nToDayjsLocale(language)) at the call site; dayjs.test.ts rewritten to the call-site pattern. 30 tests pass; verified pt-BR response time 'Normalmente entre 6 minutos e 16 minutos'. Import-size + Intl custom adapter left to #9058.

tristanlabelle (approved):

app/web/utils/dayjs.ts
kevinortiz43: removed setDayjsLocale + the _app.tsx useEffect; picker localizes via LocalizationProvider adapterLocale; ResponseRateLabel humanize() now uses .locale(i18nToDayjsLocale(language))

app/web/utils/dayjs.ts
kevinortiz43: Addressed the global dayjs locale review (commit https://github.com/Couchers-org/couchers/commit/8f8e7061b7dbddd848774a766e26354ae4d6a957): removed setDayjsLocale + the _app.tsx useEffect; picker localizes via LocalizationProvider adapterLocale; ResponseRateLabel humanize() now uses .locale(i18nToDayjsLocale(language)) at the call site; dayjs.test.ts rewritten to the call-site pattern.

PR #8987, opened 2026-06-05
Backend: add user email campaign engine + host my home nudge

aapeliv:
Adds a small registry-driven engine for User-targeted email campaigns, and ships the first campaign on it: a nudge to hosts who completed the onboarding email track but still have an empty "My Home" section.

The engine is designed so future onboarding/lifecycle emails are a thin addition: a predicate function over `User`, an optional per-user filter (for things like GrowthBook windows that vary per user), and a `NotificationTopicAction`. Dedup lives in a new `user_email_campaign_sends` table (one row per `(user_id, campaign_key)`), so we stop accreting `*_emails_sent` columns on `User` for every campaign.

The host my home nudge itself is gated by an integer GrowthBook flag `host_my_home_nudge_days_after_signup` (default `-1` = off). When enabled, the flag's value is the days-since-signup at which the user becomes eligible; the window is 2 months wide. Email-only delivery, uses the new generic templating system.

`send_onboarding_emails` and friends are intentionally left alone for now — easy to port later.

## Testing
- Added `tests/test_user_email_campaigns.py` with 9 cases covering the engine and the host nudge campaign: happy path, dedup across reruns, flag-disabled, signup before/after window, hosting status variants, onboarding-not-completed gate, my-home-already-complete gate.
- `test_email_localization` auto-picks up the new `HostMyHomeNudgeEmail` class and verifies every key resolves in English.
- `make format` and `make mypy` clean (existing mypy errors on `develop` unrelated).

To exercise manually: set `host_my_home_nudge_days_after_signup` in GrowthBook to a small int, create a `can_host` user with `onboarding_emails_sent=2`, joined that many days ago, with `max_guests`/`sleeping_arrangement` blank; the next hourly tick should send.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/locales/en.json
+++ b/app/backend/src/couchers/email/locales/en.json
@@ -162,6 +162,14 @@
       "body": "<b>{{other_name}}</b> cancelled their host request:"
     }
   },
+  "host_my_home_nudge": {
+    "subject": "Get ready to host a couch surfer",
+    "preview": "Travellers are looking for a host in your area &mdash; finish your My Home section to start showing up.",
+    "intro": "You signed up as someone who's open to hosting on Couchers.org, and there are travellers passing through who would love to stay with you.",
+    "details": "The last step before surfers can find you and send you a request is your \"My Home\" section: a bit about your place, how many guests you can take, and your sleeping arrangement.",
+    "cta_action": "Complete your My Home section",
+    "closing": "It only takes a couple of minutes, and once it's done you'll start showing up to people searching to stay in your area."
+  },
   "moderator_note": {
     "subject": "You have received a mod note",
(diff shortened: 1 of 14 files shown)


app/backend/src/couchers/email/locales/en.json
```diff
@@ -162,6 +162,14 @@
       "body": "<b>{{other_name}}</b> cancelled their host request:"
     }
   },
+  "host_my_home_nudge": {
+    "subject": "Get ready to host a couch surfer",
+    "preview": "Travellers are looking for a host in your area &mdash; finish your My Home section to start showing up.",
```
aapeliv, line 167: need to write non-ai text because we are human

PR #8989, opened 2026-06-05
Skip package.json scripts and .gitignore in mobile fingerprint

aapeliv:
Excludes the package.json `scripts` section and `.gitignore` contents from the mobile Expo fingerprint, since neither affects the native binary and so neither should move the `runtimeVersion` (or trigger a store rebuild).

- Swaps the narrow `PackageJsonAndroidAndIosScriptsIfNotContainRun` skip for `PackageJsonScriptsAll` — the former only stripped the `android`/`ios` script keys, so adding any other npm script (e.g. the recent `fingerprints:check`/`fingerprints:write`) silently moved the fingerprint. The latter is a superset, so the narrow skip is dropped.
- Adds `GitIgnore` to stop hashing `.gitignore` file contents.
- Regenerates the pinned `app/mobile/fingerprints` to match.

Note: on the current tree all three variants now hash identically per platform. This is **not** caused by these skips — the previously-committed per-variant values were stale, and the old config produces the same collapse when run against the current tree. `app.config.js` (which sets per-variant scheme/bundleId/name/icons) is unchanged, so the per-variant ExpoConfig differences aren't reaching the fingerprinter on this tree. Cross-variant isolation is enforced by per-variant OTA URL + code-signing cert rather than the runtimeVersion, so this is not a safety regression, but flagging it for awareness.

## Testing
- `npm run fingerprints:write` from `app/mobile` regenerates the committed values; `npm run fingerprints:check` then passes.
- Verified the skip change is the only cause of the hash move by computing the old committed config against the current tree.

## For maintainers

### DIFF
--- a/app/mobile/fingerprint.config.js
+++ b/app/mobile/fingerprint.config.js
@@ -1,15 +1,12 @@
-// @expo/fingerprint hashes the resolved Expo config, including `extra`. Our
-// app.config.js injects the git commit hash into `extra.gitHash`, so without
-// this skip every commit produced a unique runtimeVersion and an OTA update
-// could never match an already-installed build. Nothing under `extra` (router,
-// eas, gitHash) affects the native binary, so excluding it keeps the fingerprint
-// stable across commits while still rebuilding when real native inputs change.
-// The default package.json script skip is preserved. ExpoConfigVersions excludes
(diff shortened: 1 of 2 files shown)


aapeliv:
superseded by https://github.com/Couchers-org/couchers/pull/8988

PR #8991, opened 2026-06-05
Serialize friend and host request notifications from the recipient's perspective

aapeliv:
Friend request and host request notifications serialized the other user (and, for host requests, perspective-dependent fields like last-seen/archived/feedback) from the acting user's own context rather than the recipient's. Since these payloads are rendered for and delivered to the recipient, they must be built from the recipient's context. This aligns these sites with the host request reminder/background-job and reference paths.

## Testing
Existing API, notification, and email tests pass (`test_api.py`, `test_requests.py`, `test_notifications.py`, `test_email.py`).

**Backend checklist**

## For maintainers


tristanlabelle (approved):
Makes sense to me

PR #9002, opened 2026-06-05
Backend/i18n: Add admin translation component for admin/editor errors

aapeliv:
Adds a new English-only `admin` i18next translation component (a fourth component alongside `main`/`notifications`/`email`) and moves all admin/editor-exclusive error strings into it, out of the main locale files.

Admin and editor servicers are internal staff tooling, so their error strings don't need to be translated into all languages — this carves them out into their own component, matching how the `email` component is English-only.

**What changed:**
- New `i18n/admin_locales/en.json` with 36 admin/editor error strings, plus `get_admin_i18next()` and a small component registry in `i18n/locales.py`.
- `abort_with_error_code(..., translation_component="main")` — new optional arg that selects which component holds the error string. Fully backward-compatible; all existing calls keep using `main`.
- Tagged 45 call sites (32 in `admin.py`, 13 in `editor.py`) covering the 36 admin/editor-exclusive codes with `translation_component="admin"`, and removed those keys from all 12 main locale files (including existing translations, consistent with English-only).
- The 8 error codes also raised by user-facing servicers (`user_not_found`, `community_not_found`, `event_not_found`, `badge_not_found`, `discussion_not_found`, `invalid_birthdate`, `postal_verification_attempt_not_found`, `reference_no_text`) stay in `main` and were left untouched.
- Fixed a latent bug: `parent_node_not_found` and `postcard_not_sent` were raised in `editor.py` but missing from every locale file (they'd have thrown `LocalizationError` at runtime) — added English strings for both.

## Testing
- `make format` and `make mypy` clean on all touched files.
- `test_admin`, `test_editor`, `test_i18n_locales`, `test_i18next` — 115 passed. These tests assert on the localized English text, which is unchanged after the move.
- Manually verified: admin keys resolve (incl. non-English locale falling back to English), moved keys are gone from `main`, and shared keys still resolve from `main`.

**Backend checklist**
- [x] Added tests for any new code or added a regression test if fixing a bug (existing admin/editor/i18n tests cover the behavior; no new code paths beyond the component switch)
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no DB changes)

## For maintainers


app/backend/src/couchers/i18n/admin_locales/en.json
```diff
@@ -0,0 +1,40 @@
+{
+  "errors": {
+    "admin_blog_blurb_too_long": "The blog post blurb is too long.",
```
tristanlabelle: Could also do a pass to remove `admin_` prefixes now

tristanlabelle (dismissed):
Not bad

app/backend/src/couchers/context.py
```diff
@@ -120,17 +121,29 @@ def abort(self, status_code: grpc.StatusCode, error_message: str) -> NoReturn:
             context.abort(status_code, error_message)
 
     def abort_with_error_code(
-        self, status_code: grpc.StatusCode, error_message_id: str, *, substitutions: dict[str, str | int] | None = None
+        self,
+        status_code: grpc.StatusCode,
+        error_message_id: str,
+        *,
+        substitutions: dict[str, str | int] | None = None,
+        translation_component: str = "main",
```
tristanlabelle: That's a bit verbose. Ideas:

1. Rename to just `component`
2. Support `error_message_id` of "admin:foo_bar" since it kinds of mimics the frontend i18next namespace syntax, though with implicit "errors." parent key.

PR #9003, opened 2026-06-05
Backend/auth: Cut redundant DB round-trips in auth interceptor

aapeliv:
Screenshot 2026-06-05 at 20 00 14

Reduces the cost of `_try_get_and_update_user_details`, which pyroscope showed at ~28% of backend CPU. Two changes:

1. Build `UserAuthInfo` before `session.commit()` so the default `expire_on_commit` doesn't force a redundant re-SELECT of the user + session rows when constructing the result.
2. Compute `is_jailed` as a column (two correlated `EXISTS` subqueries hitting existing partial indexes) in the main auth query, instead of firing a separate `mod_notes` `COUNT` and an activeness-probe relationship lazy-load on every authenticated request.

The canonical jail definition stays in the model (added `.expression` forms for `jailed_pending_mod_notes`, `jailed_pending_activeness_probe`, and `is_jailed`), so the interceptor and `jail.py` servicer can't drift. Net: ~3 fewer DB round-trips per authenticated request.

## Testing
- `make format` and `make mypy` clean (no new errors in changed files)
- `test_interceptors.py`, `test_jail.py`, `test_auth.py` all pass
- Verified generated SQL uses `ix_mod_notes_unacknowledged` and `ix_activeness_probe_unique_pending_response`

**Backend checklist**
- [x] Added tests for any new code or added a regression test if fixing a bug (covered by existing interceptor/jail/auth tests)
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no schema changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -105,7 +105,9 @@ def _try_get_and_update_user_details(
 
     with session_scope() as session:
         result = session.execute(
-            select(User, UserSession)
+            # compute is_jailed as a column so the two jail checks ride along in this query as EXISTS subqueries,
+            # rather than firing a separate count + relationship lazy-load when we read user.is_jailed below
+            select(User, UserSession, User.is_jailed)
             .select_from(UserSession)
             .join(User, User.id == UserSession.user_id)
             .where(User.is_visible)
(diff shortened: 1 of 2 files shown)


tristanlabelle (approved):
Today I learned about hybrid properties

PR #9005, opened 2026-06-06
Backend/logging: Memoize sensitive-field detection in API call logging

aapeliv:
`_sanitized_bytes` strips fields marked `(sensitive)` before logging every request and response to the `APICall` table. It did this by `deepcopy`-ing the message and recursively walking every field, calling `descriptor.GetOptions().Extensions[...]` per field — on **every** RPC. Profiling (pyroscope) showed this at ~2% of total backend CPU self-time, almost all of it wasted: there are only 6 sensitive fields in the entire API (all password strings in `auth.proto`/`account.proto`), so the vast majority of messages contain nothing to strip.

Sensitivity is a static property of a message type, so this caches the descriptor analysis:

- `_descriptor_has_sensitive(descriptor)` — `@cache`d, cycle-safe reachability walk; whether a type transitively contains any sensitive field.
- `_sanitize_plan(descriptor)` — `@cache`d per-type plan (fields to clear + which subfields are worth recursing into).
- `_sanitized_bytes` — for the ~99% of types with no sensitive field anywhere, now just `SerializeToString()` with **no deepcopy and no walk**. Only password-bearing messages take the copy-and-clear path.

Behavior is unchanged — the same fields are cleared.

## Testing
- `make format` and `make mypy` clean (no errors in `interceptors.py`)
- `test_sanitized_bytes.py` — all 13 existing tests pass
- Strengthened the sensitive-field tests with a byte-level roundtrip assertion: the plaintext secret is present in the unsanitized `SerializeToString()` but absent from the sanitized output (`assert b"<secret>" not in result`), directly verifying the secret never reaches the serialized payload rather than only that the field reads back as its default.

**Backend checklist**
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no schema changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/interceptors.py
+++ b/app/backend/src/couchers/interceptors.py
@@ -3,6 +3,7 @@ from collections.abc import Callable, Mapping
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+from functools import cache
 from os import getpid
 from threading import get_ident
 from time import perf_counter_ns
@@ -12,7 +13,7 @@ from zoneinfo import ZoneInfo
 
 import grpc
 import sentry_sdk
-from google.protobuf.descriptor import ServiceDescriptor
+from google.protobuf.descriptor import Descriptor, ServiceDescriptor
 from google.protobuf.descriptor_pool import DescriptorPool
 from google.protobuf.message import Message
 from opentelemetry import trace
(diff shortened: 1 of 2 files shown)


app/backend/src/couchers/interceptors.py
```diff
@@ -190,35 +191,71 @@ def unauthenticated_handler[T, R](
     return abort_handler(message, status_code)
 
 
+@cache
+def _descriptor_has_sensitive(descriptor: Descriptor) -> bool:
```
tristanlabelle, line 195: It would make sense to recurse since this is cached. Otherwise we could visit the tree, and later several of its subtrees redundantly.

app/backend/src/couchers/interceptors.py
```diff
@@ -190,35 +191,71 @@ def unauthenticated_handler[T, R](
...
+        for f in d.fields:
+            if f.GetOptions().Extensions[annotations_pb2.sensitive]:
+                return True
+            if f.message_type is not None:
+                stack.append(f.message_type)
+    return False
+
+
+@cache
+def _sanitize_plan(descriptor: Descriptor) -> tuple[tuple[str, ...], tuple[tuple[str, bool], ...]]:
```
tristanlabelle: The type got gnarly. A small frozen dataclass with `fields_to_clear` + `fields_to_recurse` would help.

tristanlabelle (dismissed):
Nice little optimization

app/backend/src/couchers/interceptors.py
aapeliv: I think there are self-referential types, like `google.protobuf.Value`

app/backend/src/couchers/interceptors.py
aapeliv: done

PR #9026, opened 2026-06-07
Web: Add empty alt to language flag img to fix lint warning

aapeliv:
Adds an empty `alt=""` to the decorative language flag `` in `LanguagePickerSelect`, fixing a `jsx-a11y/alt-text` lint warning. The flag is purely decorative (it sits beside the language name), so an empty alt is the correct accessible treatment.

## Testing
Ran `yarn lint` — passes with no warnings or errors.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/translate/LanguagePickerSelect.tsx
+++ b/app/web/features/translate/LanguagePickerSelect.tsx
@@ -140,6 +140,7 @@ export default function LanguagePickerSelect({
     return (
       <img
         src={`https://cdn.couchers.org/img/language-icons/${flagCode}.svg`}
+        alt=""
         style={{ width: 25, ...commonStyles }}
       />
     );


app/web/features/translate/LanguagePickerSelect.tsx
```diff
@@ -140,6 +140,7 @@ export default function LanguagePickerSelect({
     return (
       <img
         src={`https://cdn.couchers.org/img/language-icons/${flagCode}.svg`}
+        alt=""
```
nabramow, line 143:
```suggestion
        alt=`flag-{flagCode}`
```

Maybe a real alt is better while you're here?

aapeliv: I think this is already next to something with alt/screen reader label with the code. So the alt would not help much here, and it's just a (disputed) visual-only aid anyway. So I think the correct thing is to hide it if it's not available.

PR #9031, opened 2026-06-07
Add exclude attending and organizing from my communities events

nabramow:
We already have `attending` and `organizing` as boolean params to the `ListMyEvents` query, but when `my_communities=true` it overrides it if the events we're attending or organizing are in our communities.

Therefore I'm adding `exclude_attending` for this, to try and make it as backward compatible as possible.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/servicers/events.py
```diff
@@ -1205,6 +1205,16 @@ def ListMyEvents(
 
         query = query.where(or_(*where_))
 
+        if request.exclude_attending:
```
aapeliv: let's combine this with the include_attending branch, with an error if both are set.

we don't need to query organizing separately, if you organize you should be set to attent too almost always.

i'm also not sure what the sql will render as with the separate select, it would be better to do the outer join and exclude

nabramow: Pushed these changes now!

nabramow:
@aapeliv I've addressed your comments now!

app/backend/src/couchers/servicers/events.py
```diff
@@ -1205,6 +1210,9 @@ def ListMyEvents(
...
+            query = query.where(EventOccurrenceAttendee.user_id == None)
```
aapeliv: combine with above branch

aapeliv (changes requested):

app/backend/src/couchers/servicers/events.py
nabramow: Done

app/backend/src/couchers/servicers/events.py
```diff
@@ -1178,15 +1178,22 @@ def ListMyEvents(
...
                 and_(
                     EventOccurrenceAttendee.occurrence_id == EventOccurrence.id,
                     EventOccurrenceAttendee.user_id == context.user_id,
                 ),
             )
-            where_.append(EventOccurrenceAttendee.user_id != None)
+            if include_attending:
+                where_.append(EventOccurrenceAttendee.user_id != None)
+            elif request.exclude_attending:
+                query = query.where(EventOccurrenceAttendee.user_id == None)
```
aapeliv: Please use the same style with where_ as elsewhere

aapeliv (approved):

PR #9034, opened 2026-06-07
Add offers_count and option to sort asc and desc to public trips

nabramow:
- Adding option to sort public trips by ascending or descending (we want asc on dashboard widget and desc on my public trips page)
- Add logic to show number of offers if the user is looking at their own public trips. I think this is helpful and I also think I want to add some view where people can click through and see offers per public trip somewhere via the "my public trips" page (then clicking an offer will open that individual messages thread). Design for that TBD keeping this backend for now.

THE GOAL (ignore the nitpicky details of this design it's just a rough prototype via Claude Design I will fix the design inconsistencies):

Screenshot 2026-06-07 at 22 45 18

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -40,6 +41,18 @@ def _is_description_long_enough(text: str) -> bool:
...
+def _parse_page_token(page_token: str) -> tuple[date | None, int | None]:
+    """Parse a page token into (from_date, trip_id). Returns (None, None) for first page."""
+    if not page_token:
+        return None, None
+    if ":" in page_token:
+        date_str, id_str = page_token.rsplit(":", 1)
+        return date.fromisoformat(date_str), int(id_str)
+    # Legacy format: id-only string
+    legacy_id = int(page_token)
+    return None, legacy_id if legacy_id else None
```
nabramow: I thiiiink I can probably remove this and pretend it never happened since it's not on prod yet

app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -238,17 +260,47 @@ def ListPublicTripsByUser(
...
+            else:
+                statement = statement.where(
+                    or_(
+                        PublicTrip.from_date < cursor_date,
+                        and_(PublicTrip.from_date == cursor_date, PublicTrip.id < cursor_id),
+                    )
+                )
+        elif cursor_id is not None:
+            # Legacy id-only cursor
+            statement = statement.where(PublicTrip.id <= cursor_id)
```
nabramow: probably can remove this

app/proto/public_trips.proto
```diff
@@ -90,6 +93,14 @@ message ListPublicTripsByUserReq {
...
   uint32 page_size = 2;
   string page_token = 3;
+
+  // If non-empty, only trips with these statuses are returned.
+  // Ignored when viewing another user's trips (always filtered to active+upcoming).
+  repeated PublicTripStatus statuses_in = 4;
+
+  // If true, trips are sorted ascending by start date (soonest first).
+  // Default (false) sorts descending (latest start date first).
+  bool ascending = 5;
```
nabramow, line 103: I'm filtering this on the frontend right now but it should be backend logic

aapeliv (dismissed):

PR #9036, opened 2026-06-07
Backend/emails: Simplify templating code

tristanlabelle:
With the last email kind having been ported to the new email system, we can clean up a few things:

1. Remove the now unused jinja2 templating formatters for all kinds of non-text values. A lesson from the new email system is that it's better to the formatting in code and only pass down string/Markup arguments to the jinja2 template. This is because there are many ways one could want to format a date, for example, for which we need our rich python support.
2. We now have a single `.mjml` file, so we can merge the template header and footer in a single file.

## Testing
Ran `_build.sh` and `uv run dump-emails`.

**Backend checklist**

## For maintainers


aapeliv (approved):
lgtm

PR #9040, opened 2026-06-08
Backend/i18n: Support localizing userless emails

### LINKED ISSUE #7734: Backend/emails: Make security emails translatable
Currently there are two of those going through `send_simple_pretty_email`, but we lose the locale information even though we probably have it upstream.

- `send_signup_email`
- `send_email_changed_confirmation_to_new_email` (we definitely have the user's locale)

tristanlabelle:
So-called "userless emails" are when we're not sending to the info stored in a `User` object. They too can be localized since we do have the localization context via the couchers context (which also provides feature flags).

Fixes #7734

## Testing
Left to CI

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/queuing.py
+++ b/app/backend/src/couchers/email/queuing.py
@@ -5,6 +5,7 @@ import yaml
 from sqlalchemy.orm.session import Session
 
 from couchers.config import config
+from couchers.context import CouchersContext
 from couchers.email.emails import EmailBase
 from couchers.email.rendering import EmailFooter, render_html_body, render_plaintext_body
 from couchers.i18n import LocalizationContext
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/email/queuing.py
```diff
@@ -37,15 +38,18 @@ def queue_email(session: Session, payload: jobs_pb2.SendEmailPayload) -> None:
...
     """
     This is a simplified version of couchers.notifications.background._send_email_notification
 
     It's for the few security emails where we don't have a user to email but send directly to an email address.
     """
 
-    # Not yet localizable
-    loc_context = LocalizationContext.en_utc()
+    loc_context = context.localization
+    if not context.get_boolean_value("notification_translations_enabled", default=False):
```
aapeliv, line 51: just FYI since the user is logged out; this will not be targeted. whatever rule is the last catch all rule in growthbook will take precedence (not code default tho)

aapeliv (approved):
lgtm

app/backend/src/couchers/email/queuing.py
tristanlabelle: Right, we don't have a user but still want to ask growthbook for the value so it can be controlled remotely. Would you have bypassed the `CouchersContext` and used the `experimentation` module directly in this case?

I'll carry on with merging but can revisit as needed.

PR #9041, opened 2026-06-08
i18n: zh-hans and zh-hant translation and layout update

HankAviator:

1. Update wrong wording: 
    * host was wrongly translated as "tv show host" or a "computer host"
    * Couchers was wrongly translated as a chair
    * other small fixes
2. Remove the space between CJK and non CJK chars

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/web/pages/main.css
tristanlabelle, line 1: @HankAviator , I took the liberty of pushing this to your branch so we transfer the spacing responsibility from the strings to the browser in one go.

Before (this string didn't have a space):

Image

After:

Image

tristanlabelle (dismissed):
Approving but let's get @nabramow 's ok on the CSS.

app/web/pages/main.css
tristanlabelle: @nabramow Is this the right place to put this CSS?

nabramow: No, this will effect all the CSS in the app within an html tag so I don't think we want it here, or do you? Which text are you specifically trying to target?

If it's a specific MUI component it can go in the theme.ts. 

I don't really understand this PR enough to tell you where to put it instead though.

tristanlabelle: The intent is to apply it globally, which applying to the root `<html>` tag achieves. It changes the display of any Chinese/Japenese/Korean adjacent to a latin letter or digit to add a little spacing between the two, basically enabling cross-script kerning. It has no effect if there are no CJK characters and I can't think of any text where we wouldn't want that spacing since it eases readability. Han was working around the lack of this feature by manually adding spaces between Chinese characters and placeholders.

See docs: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-autospace

tristanlabelle: Also FYI @aapeliv since you deleted this `main.css` in https://github.com/Couchers-org/couchers/commit/79cbc1821eb8092da8b2937a83b728037dce18b9

tristanlabelle (approved):

PR #9042, opened 2026-06-08
Frontend/friend reqs: "decline" -> "dismiss"

tristanlabelle:
Following a conversation about whether the befriender would be notified of a "decline" response, it seems like "dismiss" is less ambiguous.

My new string keys are inconsistent, but context is more important than consistency for translators, and at some point I'll lock weblate and do a big rename of our string keys to improve them.

## Testing
image

image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/connections/friends/FriendRequestsReceived.tsx
+++ b/app/web/features/connections/friends/FriendRequestsReceived.tsx
@@ -35,7 +35,7 @@ function RespondToFriendRequestAction({
   return (
     <Stack direction="row" spacing={1}>
       <Button
-        aria-label={t("connections:decline")}
+        aria-label={t("connections:friend_requests_dismiss_button")}
         onClick={() => {
           reset();
           respondToFriendRequest({
@@ -47,7 +47,7 @@ function RespondToFriendRequestAction({
         variant="outlined"
         loading={isLoading}
       >
-        {t("connections:decline")}
+        {t("connections:friend_requests_dismiss_button")}
       </Button>
       <Button
         aria-label={t("connections:accept")}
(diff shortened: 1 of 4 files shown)


nabramow (approved):
LGTM! Thanks for fixing this!

jesseallhands:
When clicking on the "pending" button, are we no longer taking the user to the "my connections" page to let them respond there? This way they can respond to any other open requests there too.

nabramow:
> When clicking on the "pending" button, are we no longer taking the user to the "my connections" page to let them respond there? This way they can respond to any other open requests there too.

Ah yeah I kinda like this idea, have the pending click take them to the my connections page and jump to that specific card via an anchor link maybe?

jesseallhands (changes requested):
Clicking the "pending" button should take the user to the "my connections" page, ideally to the specific card for that user (but also a simple direct to "my connections" is fine too.

tristanlabelle:
> Clicking the "pending" button should take the user to the "my connections" page, ideally to the specific card for that user (but also a simple direct to "my connections" is fine too.

@jesseallhands , I'm not against this but it is unrelated to my PR. If we agree with the "decline" -> "dismiss" string change we should move forward with this PR and we can look into the behavior change separately. You can `@-claude` and ask it to have a follow-up ticket filed when an idea like this pops into a PR.

nabramow (approved):

tristanlabelle:
@claude file the issue @jesseallhands raised about what the pending button should do, mark it as feature, frontend and add it to the engineering board as TODO is you have the permissions.

> Clicking the "pending" button should take the user to the "my connections" page, ideally to the specific card for that user (but also a simple direct to "my connections" is fine too.

jesseallhands (approved):
lgtm

PR #9043, opened 2026-06-08
Render browsable sample emails in CI preview

aapeliv:
Sample: https://cd947569--test-artifacts.preview.couchershq.org/emails/index.html
---
Renders every transactional email (with dummy data) in CI and publishes a browsable HTML index of them to the preview environment, so reviewers can eyeball email changes without spinning anything up.

Rather than a bespoke job, this rides the existing `test_artifacts` → `preview:backend` path (the same mechanism the postcard samples use):

- **`dump_emails.py`**: refactored the dump loop into a reusable `dump_all()` that renders every email variation in **every supported locale** (one subdirectory per locale) and writes a single-page `index.html`: an accordion with one row per email variation (class, variation number, subject) that expands to show the HTML and plaintext renders side by side in iframes (stacked on narrow screens), plus a locale selector that swaps subjects and reloads open previews. The selected locale is reflected in the URL query (`?locale=fr`) for shareable links. Previews load lazily, only when a row is expanded. No standalone env setup needed; it relies on config being present (test harness or loaded deployment env), same as before.
- **`test_dump_emails.py`** (new): renders all variants into `test_artifacts/emails/` under the existing `testconfig` fixture, mirroring `test_generate_postcard_images`.
- **`.gitlab-ci.yml`**: the only change is making the `test_artifacts` copy recursive (`cp -r`) so the `emails/` subdir flows through, plus an echo pointing at the index. The existing `preview:backend` job already uploads `test_artifacts/` to S3, so the index lands at `https://<sha>--test-artifacts.preview.couchershq.org/emails/index.html` (and the `<slug>` variant).

Note: email strings currently only have English translations (`couchers/email/locales/` has just `en.json`), so for now locales differ only in dates and other babel-localized content — e.g. EventCreatedEmail in `fr` shows "mardi 15 juillet". Translations will appear in the picker automatically as they land.

## Testing
`uv run pytest src/tests/test_dump_emails.py` passes and produces 68 variations across 48 email types in 24 locales plus the index; all linked files exist. Verified the rendered index in headless Firefox: collapsed accordion, expanded side-by-side previews, and locale switching via `?locale=` all work. `make format` passes and `make mypy` introduces no new errors.

## For maintainers


app/backend/src/couchers/email/dump_emails_index.html.jinja2
tristanlabelle, line 1: Suggestion: `.html.jinja2` extension to call out it's a template.

tristanlabelle (dismissed):
Nice goodie! Feels a little overkill to generate N * M emails for all variants and locales but if it takes a fraction of a second, sure.

How do we discover this? Maybe add a mention to the emails readme.md?

aapeliv:
> How do we discover this? Maybe add a mention to the emails readme.md?

I'm working on a couchersbot github integration to show previews etc. I'm going to put it there?

app/backend/src/couchers/email/dump_emails_index.html.jinja2
aapeliv: ahahah claude originally did that. but i didn't like it because i think our other jinja templates are pure .html...

Issue #8523, opened 2026-05-10
language preferences with a wrong flag symbol 


CouchersBot:
Subject: language preferences with a wrong flag symbol 
Description:
Hi all, 

today I got an email notice, "Malcolm Williams created a new discussion called Taipei Thursday Social Event Thursday 14th May in Taiwan".   As I clicked "View discussion" to log in, the content was displayed in Simplified Chinese. When I scrolled down the language panel to choose other languages, I found the inappropriate layout of the flags attached with the languages. The language called "ZH-HANT", is displayed with a flag of China. In China, the only official writing system is simplified Chinese. Henceforth it is correct that "ZH-HANS" bears the flag 🇨🇳. When it comes to "ZH-HANT", it looks incorrect and awkward to bear the flag 🇨🇳.

Results:
As far as I know, the writing system of traditional Chinese characters is officially and publicly used in Taiwan, where the flag of the country looks like 🇹🇼. It would ne great that you will verify and fix the problem. When the language is displayed with an appropriate flag symbol 🇹🇼, I believe more and more users would like to choose it as their preferred language, instead of others. Thank you for your time and further revise.

Backend version: develop-f31fba77
Frontend version: develop-f31fba77
User Agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/138.0.0.0 Mobile Safari/537.36
Screen resolution: 360x620
Page: https://couchers.org/translate
User: [@sylviawu](https://couchers.org/user/sylviawu) (13876)

kevinortiz43:
@tristanlabelle so would you like the language symbols to be removed and changed to EN, SP, GE, etc?

tristanlabelle:
@kevinortiz43 , you're welcome to pick it up. Many of the i18n tasks are assigned to me by default but I'm mostly on the backend at the moment. 

We want to minimally:

1. remove flags
2. show the full names of the languages without localization, e.g. English, Français, Español, 中文简体 -- every language name in its own language.

Then we can discuss if we also want to show the names of languages localized to the current language or tweak the design of the language picker. But hopefully those would be separate follow-up changes.

PR #9049, opened 2026-06-09
fix(web): replace language-picker flags with each language's autonym (#8523)

kevinortiz43:
Removes country flags from the language selector and shows each language by its **autonym** (its name in its own language), so options are recognizable regardless of the current UI language. Fixes the original bug where written languages were mapped to one country flag (most notably zh-Hant "Chinese (Traditional)" showing the China flag, same as zh-Hans).

**What changed**
- **No flags** in the navbar/settings picker (`LanguagePickerSelect`) or the `/translate` progress page (`TranslationProgress`).
- Each language renders by its **autonym** via a `nativeName` field on `LANGUAGE_MAP` (e.g. `Deutsch`, `Nederlands`, `中文（繁體）`, `Русский`).
- **Removed the English `name` field** from the `Language` interface and every `LANGUAGE_MAP` entry. It was never read in product code and an English-only constant can't be localized — names now come from a localizable source or `nativeName` (addresses @tristanlabelle's review note).
- **Settings page** now sources the preferred-language name from the localizable `language_names` translations (`t(\`global:language_names.${i18n.language}\`)`) instead of `LANGUAGE_MAP[...].nativeName`, so the sentence localizes with the UI (English → `English`, pt-BR → `Português (Brasil)`).
- **Deleted all flag icons:** removed the last in-repo language flag `CatalanFlagIcon` (+ its export); the other flags were CDN images already dropped from the picker. Stripped the now-unused `flagIconCode` field from `LANGUAGE_MAP` and the `FLAG_CODES` map / `getFlagCode` / `flagCode` wiring from the translation-file generator.
- **Scripts + docs in sync:** `create-translation-files.js` no longer writes the removed `name` field into `constants.ts` (seeds `nativeName` only; English name still added to `language_names`). `docs/localization.md` no longer references "flag code" / the wrong `.ts` extension.

## Testing
- `yarn test` (full web suite) — all `features/translate/` suites pass (autonyms render; zh-Hant/zh-Hans distinct; no flag images). Remaining failures are pre-existing flaky input-timing tests unrelated to this change.
- `yarn format` (eslint + prettier) + `tsc` on changed files — clean.
- Manual (Playwright): picker dropdown lists all autonyms with zero flag images; settings preferred-language sentence localizes correctly (pt-BR → `Português (Brasil)`, English → `English`).

**Web frontend checklist**

## For maintainers

image

image


kevinortiz43:
@tristanlabelle I have switched out the flags to the string of the language. Test it out and let me know.

app/web/features/translate/LanguagePickerSelect.tsx
```diff
@@ -124,27 +122,6 @@ export default function LanguagePickerSelect({
...
-      );
-    }
-
-    return (
-      <img
-        src={`https://cdn.couchers.org/img/language-icons/${flagCode}.svg`}
-        alt=""
-        style={{ width: 25, ...commonStyles }}
-      />
-    );
```
tristanlabelle, line 146: @nabramow I don't think we'll be using these flag images anymore, should we delete them from somewhere? They're not in the repo.

app/web/features/translate/TranslationProgress.tsx
```diff
@@ -9,7 +9,6 @@ import {
   Typography,
   useMediaQuery,
 } from "@mui/material";
-import { CatalanFlagIcon } from "components/Icons";
```
tristanlabelle, line 12: We should be able to delete that file now. I wouldn't keep it speculatively, it stays in the git history if we need it.

app/web/i18n/constants.ts
```diff
@@ -1,5 +1,9 @@
 interface Language {
+  // English name, used in localized sentences (e.g. settings copy).
   name: string;
```
tristanlabelle: @claude File a bug labeled `i18n` and `frontend` that the settings page use non-localized language names, as in:

> Su idioma preferido actualmente es Spanish (Spain).

tristanlabelle (dismissed):
Good stuff, it looks better than I expected too! Double-check that the picker in the settings page looks fine too before merging.

image

app/web/features/translate/TranslationProgress.tsx
nabramow: That package actually since added the flag so we don't have to manually do it too. It's been on my to-do list to switch it out for a bit. So looks like I can cross that off my to-do!

app/web/features/translate/LanguagePickerSelect.tsx
kevinortiz43: I have deleted the flag icons where I found them.

app/web/features/translate/TranslationProgress.tsx
kevinortiz43: I deleted the use of those flag icons, and the catallan flag.tsx since that one was used specifically for that flag.

app/web/i18n/constants.ts
kevinortiz43: Resolved the settings page issue.

app/web/i18n/constants.ts
```diff
@@ -1,6 +1,9 @@
...
```
tristanlabelle: I wonder if we could remove the English names. They can only be useful for diagnostics, if we reference them in product code, we won't be i18n'd. No need to do this in this PR though.

app/web/features/translate/LanguagePickerSettings.tsx
```diff
@@ -19,7 +19,7 @@ export default function LanguagePickerSettings({
   const { t } = useTranslation([GLOBAL]);
   const router = useRouter();
   const { locale } = router;
-  const languageName = LANGUAGE_MAP[locale || "en"]?.name;
+  const languageName = LANGUAGE_MAP[locale || "en"]?.nativeName;
```
tristanlabelle: I'm happy to do this as a follow-up but it'd be more idiomatic to get the locale from the translation module and lookup its name in our translation files.:

```ts
const { t, language: locale } = useTranslation([GLOBAL]);
const languageName = t(`language_names.${locale}`);

**Review by tristanlabelle — DISMISSED** [review]
Looks great. I have two minor comments but feel free to resolve and I can tackle them as follow-up.

**Comment by nabramow** [comment]
@kevinortiz43 @tristanlabelle we have two scripts that add and remove languages that need to be updated as well to remove the flag part and replace it with whatever we're doing now FYI:

scripts/create-translation-files.js
scripts/delete-translation-files.js

And make sure nothing needs to be updated in this readme:

docs/localization.md

File: app/web/i18n/constants.ts (hunk 4 above)
Comment (kevinortiz43) [reply app/web/i18n/constants.ts]: What do you mean by removing English names

Comment (tristanlabelle) [reply app/web/i18n/constants.ts]: Removing the Language.name property. I don't see a use for English-only name constants, unless for diagnostics. In the app we'd always want to get language names from a localizable source (or the nativeName property you added).

Comment (kevinortiz43) [reply app/web/i18n/constants.ts]: ah okay we can remove it, doesn't sound hard.

File: app/web/features/translate/LanguagePickerSettings.tsx (hunk 5 above)
Comment (kevinortiz43) [reply app/web/features/translate/LanguagePickerSettings.tsx]: doesn't look too difficult let me see I can do that

File: app/web/scripts/create-translation-files.js [hunk 6]
```diff
@@ -575,7 +575,6 @@ function addToConstants(languageCode, languageName) {
     const indent = "  ";
     const newEntry = [
       `${indent}"${languageCode}": {`,
-      `${indent}  name: "${languageName}",`,
       // Placeholder autonym — a maintainer should replace this with the
       // language's name in its own language (shown in the language picker).
       `${indent}  nativeName: "${languageName}",`,
```
tristanlabelle, line 580: It might be easy to miss that the `nativeName` is incorrect. We could generate:

```
`${indent}  nativeName: "${languageName}", // FIXME: Update to autonym (i.e. the language name in the language itself)`
```

Comments are generally not valid json so it'll be obvious that something is wrong.

Or you can try populating the correct value with:

```js
nativeName = new Intl.DisplayNames([locale], { type: 'language' }).of(locale)
nativeName = nativeName.charAt(0).toUpperCase() + nativeName.slice(1);
```

docs/localization.md
```diff
@@ -90,6 +90,7 @@ yarn delete-translation-files sv  # Removes Swedish
 This script will:
 - Delete all translation files for the language
 - Remove the language from `allLanguages.js` and `constants.ts`
+- Remove the English name from `language_names` in `resources/locales/en.json`
```
tristanlabelle, line 93: Since your PR removes that field, this is moot.

tristanlabelle (approved):
Thanks. Feel free to merge anytime and do any follow-ups in a different PR. I don't want to keep inflating the scope of this one!

app/web/scripts/create-translation-files.js
kevinortiz43: Copy

docs/localization.md
kevinortiz43: Will do in new PR

app/web/features/translate/LanguagePickerSettings.tsx
kevinortiz43: Done

aapeliv:
this PR broke the language picker in top right when logged in, and the one in settings. i'll be reverting this

kevinortiz43:
Was this in mobile? I tested it in desktop and the vercel deployment for web and it worked okay.

kevinortiz43:
Reverified with Aapeli and Tristan. This merge is good, remerging back in.

PR #9055, opened 2026-06-09
Fix unbounded width on messages pages, centralize standard page width in PageContainer

aapeliv:
Tested all the mentioned pages. All others look the same as before except messaging pages (fixed) and password reset (fixed).

---
The messages routes (`/messages`, `/messages/chats/<id>`, `/messages/request/<id>`) use the `full-width` layout variant but never re-constrained their content, so on wide desktop screens they stretched across the whole viewport with no max-width, making them hard to use.

This PR fixes that and makes the underlying pattern principled:

- Adds `components/PageContainer.tsx` — a thin wrapper around MUI `Container` that bakes in the platform-standard page width (`lg`), exported alongside a `STANDARD_PAGE_MAX_WIDTH` constant which `AppRoute`'s `standard` variant now also uses. The standard width is defined in exactly one place.
- Wraps the messages page content in `PageContainer` (with `disableGutters`, since the views carry their own edge padding — mobile/tablet appearance is unchanged). The page keeps the `full-width` variant because the `standard` variant's bottom padding would break the chat view's `100dvh`-based height math.
- Migrates all other hardcoded `<Container maxWidth="lg">` usages to `PageContainer`: Dashboard, Team, Signup, LandingPage, HistoryTimeline, WhatIsCouchSurfingPage. Raw `Container` now signals a deliberately non-standard width (e.g. the landing page's full-bleed band).
- Switches `terms`, `password-reset`, and `complete-password-reset` from the `full-screen` variant to the standard layout — they have no full-bleed content, so the variant did nothing useful. Their now-redundant hand-rolled width/padding wrappers (`TOS`, `ResetPassword`, `CompleteResetPassword`) are removed in favor of the layout's container.

Per review, the `full-screen` variant is now merged into `full-width` — the two have been behaviorally identical since nav-height handling moved into the flex layout chain, so the alias was vestigial. Its remaining users (landing, login, signup, invite) now pass `full-width`.

## Testing

- `npx tsc --noEmit`, `eslint`, and `prettier` pass on all changed files
- Ran test suites for `features/auth/password`, `components/AppRoute`, `features/dashboard`, `features/team` — 56 tests pass
- Still to do before marking ready: visual check on desktop/tablet/mobile of `/messages/chats/<id>`, `/messages/request/<id>`, the messages list tabs, `/terms`, `/password-reset`, and `/complete-password-reset` (the last three had minor padding geometry changes)

**Web frontend checklist**
- [x] Added tests where relevant (no new behavior; existing suites cover the touched components)

## For maintainers

### DIFF
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -5,6 +5,7 @@ import CenteredSpinner from "components/CenteredSpinner/CenteredSpinner";
 import CookieBanner from "components/CookieBanner";
 import ErrorBoundary from "components/ErrorBoundary";
 import Footer from "components/Footer";
+import { STANDARD_PAGE_MAX_WIDTH } from "components/PageContainer";
 import { useAuthContext } from "features/auth/AuthProvider";
 import { PushNotificationBanner } from "features/notifications/PushNotificationBanner";
 import { useRouter } from "next/router";
@@ -174,7 +175,7 @@ function AppRoute({
                 variant === "full-width" ||
                 variant === "no-overflow"
                   ? false
-                  : "lg"
+                  : STANDARD_PAGE_MAX_WIDTH
               }
             >
               {children}
--- a/app/web/components/AppRoute.tsx
+++ b/app/web/components/AppRoute.tsx
@@ -19,7 +19,7 @@ import Navigation from "./Navigation";
 interface AppRouteProps {
   isPrivate: boolean;
   noFooter?: boolean;
-  variant?: "standard" | "full-screen" | "full-width" | "no-overflow";
+  variant?: "standard" | "full-width" | "no-overflow";
   bottomMargin?: string;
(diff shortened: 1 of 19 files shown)


nabramow:
I'm going to tag @darrenvong as a reviewer here as he was working on this super recently.

app/web/components/AppRoute.tsx
```diff
@@ -174,7 +175,7 @@ function AppRoute({
                 variant === "full-width" ||
```
darrenvong: Since you're in the area refactoring these... these two variants look identical to me, so I reckon it might be worth cleaning that up too and only keep `full-width`

aapeliv: done

darrenvong (approved):
Oops sorry for the delay for this, LGTM!

PR #9064, opened 2026-06-10
Backend/emails: Consolidate render_email

### LINKED ISSUE #7419: Consolidate duplicated email templating logic
`background.py`'s `_send_email_notification` and `v2.py`'s `send_simple_pretty_email` duplicate email the email templating logic. To avoid having to keep the two in sync, we should consolidate them.

tristanlabelle:
We had 3 different pieces of code doing the same sequence of calls to render an `EmailBase` into subject+body strings. This PR extracts it into `render_email`.

Also adjusts how the email system is split between files:

- `blocks.py`: Data model for blocks-based emails, including `EmailBase` base class.
- `emails.py`: All concrete email definitions implementing `EmailBase`.
- `rendering.py`: Rendering logic for block-based emails.

Fixes #7419

## Testing
Left to existing CI coverage.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/blocks.py
+++ b/app/backend/src/couchers/email/blocks.py
@@ -1,0 +1,248 @@
+"""
+Data model for emails built out of well-known blocks,
+that can be rendered HTML and plaintext for any locale.
+"""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Self
+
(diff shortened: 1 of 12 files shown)


aapeliv (approved):
lgtm and clicked around on the generated emails and they seem non-broken :)

PR #9069, opened 2026-06-10
Backend/emails: Make footers mostly localizable

tristanlabelle:
... except for the `Turn off emails for: <topic-action>` part which is inherently tricky to localize because of string concatenation and might need a new design. I'll address it separately.

Related to #7420

## Testing
Ran `uv run dump-emails`:

image

```

---

Manage notification preferences: <https://example.com/manage-notifications>
Turn off emails for topic-action: <https://example.com/unsubscribe>

Do not email me (disables hosting): <https://example.com/do-not-email>
```

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/locales/en.json
+++ b/app/backend/src/couchers/email/locales/en.json
@@ -12,7 +12,7 @@
       "donate_link": "Donate",
       "volunteer_link": "Volunteer",
       "blog_link": "Blog",
-      "nonprofit_note": "<i>Couchers.org is a project of Couchers, Inc. a U.S. 501(c)(3) non-profit, tax-exempt organization.</i>",
+      "nonprofit_note": "Couchers.org is a project of Couchers, Inc. a U.S. 501(c)(3) non-profit, tax-exempt organization.",
       "security_email_note": "This is a security email, you cannot unsubscribe from it.",
       "notification_settings_link": "Manage notification preferences",
       "do_not_email_link": "Do not email me (disables hosting)",
--- a/app/backend/src/couchers/email/locales/en.json
+++ b/app/backend/src/couchers/email/locales/en.json
@@ -4,7 +4,20 @@
     "closing_line": "Best,<br>The Couchers.org Team",
     "founders_signature": "Aapeli and Itsi,<br>Couchers.org Founders",
     "security_warning_contact_support": "If you did not initiate this action, please contact us by emailing <a href=\"mailto:support@couchers.org\">support@couchers.org</a> so we can sort this out as soon as possible!",
-    "do_not_reply_request": "<b>Do not reply to this email.</b> Use one of the buttons above to reply to this request."
+    "do_not_reply_request": "<b>Do not reply to this email.</b> Use one of the buttons above to reply to this request.",
+    "footer": {
+      "received_because": "You're receiving this email because you signed up for Couchers.org.",
+      "contact_support": "You can reach our support team at <a href=\"mailto:support@couchers.org\">support@couchers.org</a>.",
+      "timezone_note": "All times are in {{ timezone }}, based on your profile location.",
+      "donate_link": "Donate",
+      "volunteer_link": "Volunteer",
+      "blog_link": "Blog",
+      "nonprofit_note": "<i>Couchers.org is a project of Couchers, Inc. a U.S. 501(c)(3) non-profit, tax-exempt organization.</i>",
+      "security_email_note": "This is a security email, you cannot unsubscribe from it.",
+      "notification_settings_link": "Manage notification preferences",
+      "do_not_email_link": "Do not email me (disables hosting)",
+      "happy_couch_icon_caption": "The happy Couchers couch hopes you have a great rest of the day."
+    }
   },
   "calendar_events": {
     "title_cancelled": "Cancelled: {{ title }}",
--- a/app/backend/src/couchers/email/rendering.py
+++ b/app/backend/src/couchers/email/rendering.py
@@ -156,6 +156,7 @@ def _get_footer_template_args(footer: EmailFooter, loc_context: LocalizationCont
     }
 
     if unsubscribe_info := footer.unsubscribe_info:
+        # TODO(#7420): Localize "Turn off emails for: " text, avoiding string concatenations.
         args.update(
             {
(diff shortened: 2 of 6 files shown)


app/backend/src/couchers/email/locales/en.json
```diff
@@ -4,7 +4,20 @@
     "closing_line": "Best,<br>The Couchers.org Team",
     "founders_signature": "Aapeli and Itsi,<br>Couchers.org Founders",
     "security_warning_contact_support": "If you did not initiate this action, please contact us by emailing <a href=\"mailto:support@couchers.org\">support@couchers.org</a> so we can sort this out as soon as possible!",
-    "do_not_reply_request": "<b>Do not reply to this email.</b> Use one of the buttons above to reply to this request."
+    "do_not_reply_request": "<b>Do not reply to this email.</b> Use one of the buttons above to reply to this request.",
+    "footer": {
+      "received_because": "You're receiving this email because you signed up for Couchers.org.",
+      "contact_support": "You can reach our support team at <a href=\"mailto:support@couchers.org\">support@couchers.org</a>.",
```
aapeliv, line 10: is it possible to interpolate the email address and link instead?

app/backend/src/couchers/email/locales/en.json
```diff
@@ -4,7 +4,20 @@
...
+      "timezone_note": "All times are in {{ timezone }}, based on your profile location.",
+      "donate_link": "Donate",
+      "volunteer_link": "Volunteer",
+      "blog_link": "Blog",
+      "nonprofit_note": "<i>Couchers.org is a project of Couchers, Inc. a U.S. 501(c)(3) non-profit, tax-exempt organization.</i>",
```
aapeliv: should the markup be a part of this string since the entire string is italics?

app/backend/src/couchers/email/locales/en.json
```diff
@@ -4,7 +4,20 @@
...
+      "security_email_note": "This is a security email, you cannot unsubscribe from it.",
+      "notification_settings_link": "Manage notification preferences",
+      "do_not_email_link": "Do not email me (disables hosting)",
+      "happy_couch_icon_caption": "The happy Couchers couch hopes you have a great rest of the day."
```
aapeliv, line 19: :)

app/backend/src/couchers/email/rendering.py
```diff
@@ -116,6 +120,47 @@ def _to_plaintext(text: str | Markup) -> str:
...
+        "donate_link": localize(".donate_link"),
+        "volunteer_link": localize(".volunteer_link"),
+        "blog_link": localize(".blog_link"),
+        "nonprofit_note": localize(".nonprofit_note"),
+        "is_critical": footer.unsubscribe_info is None,
+    }
+
+    if unsubscribe_info := footer.unsubscribe_info:
+        # TODO(#7420): Localize "Turn off emails for: " text, avoiding string concatenations.
+        args.update(
```
aapeliv, line 160: the .update looks ugly to me; but this is really not important :P

app/backend/templates/v2/generated_html/blocks.html
```diff
@@ -378,18 +378,21 @@
...
                               </tr>
                               <tr>
                                 <td align="center" style="font-size:0px;padding:10px 25px;word-break:break-word;">
-                                  <div style="font-family:Ubuntu, Helvetica, Arial, sans-serif;font-size:10px;line-height:1;text-align:center;color:#333333;"><a href="https://couchers.org/donate" style="color: #00a398;">Donate</a> | <a href="https://github.com/Couchers-org/couchers" style="color: #00a398;">GitHub</a> | <a href="https://couchers.org/volunteer" style="color: #00a398;">Volunteer</a> | <a href="https://couchers.org/blog" style="color: #00a398;">Blog</a></div>
+                                  <div style="font-family:Ubuntu, Helvetica, Arial, sans-serif;font-size:10px;line-height:1;text-align:center;color:#333333;"><a href="https://couchers.org/donate" style="color: #00a398;">{{ donate_link }}</a> | <a href="https://github.com/Couchers-org/couchers" style="color: #00a398;">GitHub</a> | <a href="https://couchers.org/volunteer" style="color: #00a398;">{{ volunteer_link }}</a> | <a href="https://couchers.org/blog" style="color: #00a398;">{{ blog_link }}</a></div>
```
aapeliv, line 389: would we want "GitHub" to be a localizable string? is there a language where having it isolated like this makes no sense and should be prefixed or postfixed with something?

aapeliv (dismissed):
none are blockers but please feel free to fix any of the comments

app/backend/src/couchers/email/locales/en.json
tristanlabelle: Yes, and there are a few other instances where we can do this so @claude , please file a child issue of the localizable emails issue about auto-linking URLs in email localizable strings. Label "i18n" and "backend" and assign to me.

app/backend/src/couchers/email/locales/en.json
tristanlabelle: I left it in because the CJK/Arabic translations won't want italics, it's very unnatural on their scripts.

I was also thinking of switching those to `<strong>`/`<em>`, but even then I'm not sure how we conditionally make those style differently if we have CJK text, especially given email CSS limitations. I won't block on this because it's one of the things I plan to revisit in https://github.com/Couchers-org/couchers/issues/7368 .

@HankAviator Do you have experience or recommendations about how to handle text emphasis and especially italics with CJK scripts?

app/backend/templates/v2/generated_html/blocks.html
tristanlabelle: I can't think of a scenario, so I'll wait for a translator to yell at me and fix it then.

app/backend/src/couchers/email/locales/en.json
HankAviator: It looks not natural but people accept it as a format today (given underlining is even less used on modern web). It's ok to keep the consistency for CJK chars. I haven't heard issues introduced by italics in CJK so far @tristanlabelle

tristanlabelle: Good to know! I'll take the italics out of the string then.

aapeliv (approved):

PR #9093, opened 2026-06-13
Backend: add per-step signup funnel counters

aapeliv:
Adds Prometheus counters for each step of the signup flow, so the funnel between signup initiation and completion is observable for realtime-ish monitoring (current volume is ~100-300 signups/day).

We already expose `couchers_signup_initiations_total` and `couchers_signup_completions_total`, from which the headline completion rate is computable downstream as `sum(rate(completions[24h])) / rate(initiations[24h])` (the <20min completion time makes cohort lag negligible at a 24h window). The new counters fill in the intermediate steps so a broken deploy points at *which* step died rather than just nudging the blended rate:

- `couchers_signup_account_filled_total`
- `couchers_signup_email_verified_total`
- `couchers_signup_guidelines_accepted_total`
- `couchers_signup_motivations_filled_total`

They are unlabeled (matching `signup_initiations_total`) for clean ratios, and each fires exactly once the first time a flow passes its gate. The account/email/motivations steps are already guarded against re-entry; guidelines acceptance is not, so its increment is gated on the `accepted_community_guidelines < GUIDELINES_VERSION` transition.

## Testing
- Added `test_signup_funnel_counters`, which runs an incremental signup and asserts each counter fires exactly once — including that re-accepting the community guidelines does not double-count.
- `make format` and `make mypy` clean for the touched files (pre-existing mypy errors in unrelated test files remain).
- All 41 tests in `test_auth.py` pass.

No database changes / migrations.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/metrics.py
+++ b/app/backend/src/couchers/metrics.py
@@ -622,6 +622,25 @@ signup_completions_counter: Counter = Counter(
     "Number of completed signups",
     labelnames=["gender"],
 )
+# Per-step signup funnel counters. Each fires once, the first time a signup flow satisfies the given gate, so
+# that step_total/initiations_total gives the fraction of signups that reached that step. Unlabeled to match
+# signup_initiations_counter for clean ratios.
+signup_account_filled_counter: Counter = Counter(
+    "couchers_signup_account_filled_total",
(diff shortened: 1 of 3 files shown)


aapeliv:
@claude check this pr is correct

PR #9096, opened 2026-06-13
Backend/emails: Move image embedding to templating code

### LINKED ISSUE #8863: Backend/emails: Specify email inline images in SendEmailPayload
Currently `send_email` is coupled with the email templates and fixes up image files references in the html. We can move that logic into the email templating code by expressing this as `repeated SendEmailPayload.html_related: EmailAttachment`, and adding `content_id` and `data_path` to `EmailAttachment`.

tristanlabelle:
Fixes #8863

Moves the responsibility for embedding images in emails from `send_smtp_email` (emailing job) to `render_email` (templating code), by extending `SendEmailPayload` to support specifying `html_related_parts` and extending `EmailPart` to support a `Content-ID` header and referencing data from a path (the image file).

With this change, `send_smtp_email` merely reconstructs the SMTP message and sends it. However for backcompat I kept the a code path that does the embedding in `send_smtp_email` as fallback until we've deployed long enough to drain the emailing job queue from jobs that didn't do the embedding.

Also renames `EmailAttachmentV2` to `EmailPart` since it can naturally represent both attachments and related image parts.

## Testing
Ran backend + frontend locally and verified with an email that images still work:

image

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/proto/internal/jobs.proto
+++ b/app/backend/proto/internal/jobs.proto
@@ -15,13 +15,17 @@ message SendEmailPayload {
   // source data as to where this email came from
   string source_data = 8;
   reserved 9; // Previous "attachments" field, which had mime_type+filename subfields.
-  repeated EmailAttachmentV2 attachments = 10;
+  repeated EmailPart attachments = 10;
+  repeated EmailPart html_related_parts = 11; // MIME parts which are multipart/related to the HTML body.
 }
 
-message EmailAttachmentV2 {
+// A MIME part of an SMPT email, usable for attached files or inline images.
+message EmailPart {
   bytes data = 1;
   string content_disposition = 2; // The Content-Disposition header, including parameters
   string content_type = 3; // The Content-Type header, including parameters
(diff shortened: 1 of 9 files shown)


app/backend/proto/internal/jobs.proto
```diff
@@ -15,13 +15,17 @@ message SendEmailPayload {
   // source data as to where this email came from
   string source_data = 8;
   reserved 9; // Previous "attachments" field, which had mime_type+filename subfields.
-  repeated EmailAttachmentV2 attachments = 10;
+  repeated EmailPart attachments = 10;
```
aapeliv, line 18: i guess this is wire compatible, so can reuse 10?

aapeliv (approved):
lgtm

app/backend/proto/internal/jobs.proto
tristanlabelle: Yes, protobuf message names are not part of the wire format, and other changes are only adding fields.

PR #9098, opened 2026-06-13
Backend/emails/i18n: Localize EventUpdatedEmail's items

### LINKED ISSUE #8875: Backend/i18n: Support localizing changed fields in event updated email
For example: "time, location" should be localized. We need to support localizing fields and formatting localized lists.

tristanlabelle:
The "event updated" email includes a list of modified items, but those were not localized. This PR updates the protos to reference items as enums, implements localization for each item, and for lists in general.

Kept a backcompat path to handle queued `EventUpdated` notifications, and filed #9117 to remove it once the queue drains.

Fixes #8875

## Testing
Ran `uv run dump-emails` and added a test for `localize_list`.

image

**Backend checklist**

## For maintainers


tristanlabelle:
@aapeliv Now includes migrating protos to using enums for event updated items. Ready for review.

app/backend/proto/internal/jobs.proto
```diff
@@ -76,7 +76,11 @@ message GenerateEventCreateNotificationsPayload {
 message GenerateEventUpdateNotificationsPayload {
   int64 updating_user_id = 1;
   int64 occurrence_id = 2;
-  repeated string updated_items = 3;
+  // TODO(#9117): Remove once unused. Was not i18n-friendly. Use updated_enum_items.
+  repeated string updated_str_items = 3 [deprecated = true];
+  // EventUpdateItem is from the shared protos and cannot be referenced here.
+  // int32 has the same on-wire encoding.
+  repeated int32 updated_enum_items = 4;
```
aapeliv, line 83: how about just renaming the field to `update_items_identifiers`, then translating on the receiving side based on "start time" -> key?

i know it's vaguely ugly to use "start time" as an identifier instead of an enum, but also much simpler than this int32/enum stuff?

app/proto/notification_data.proto
```diff
@@ -133,10 +133,20 @@ message EventCreate {
   }
 }
 
+enum EventUpdateItem {
+  EVENT_UPDATE_ITEM_TITLE = 0;
```
aapeliv: should have an unspecified as 0

app/backend/proto/internal/jobs.proto
tristanlabelle: The source of truth for these strings is somewhere in a backend servicer, and it flows through `GenerateEventUpdateNotificationsPayload` >`EventUpdate` > to be read by some distant email code. Having an enum makes for a stronger API and it's move obvious how to evolve when we later add e.g. `EVENT_UPDATE_ITEM_MAX_PARTICIPANTS`.

I'll be able to remove the `updated_str_items` path soon enough, which will make things simple again.

(even with strings, I'd have decoupled the localization string key from the item value, so it wouldn't have saved that match statement)

aapeliv: sure. i still feel there is a cost to changes to notification_data.proto

aapeliv (approved):

app/backend/proto/internal/jobs.proto
tristanlabelle: > i still feel there is a cost to changes to notification_data.proto

I'm willing to suffer the dual migration PR cost (deprecation + later removal) for the cause :P

PR #9104, opened 2026-06-13
Attach Sentry session replay + per-user links to bug reports

aapeliv:
Bug reports currently carry three free-text fields plus minor diagnostics (user agent, URL, version, screen size), which is often not enough to reproduce a reported issue. This PR enriches the GitHub issue a report creates with links into Sentry, reusing the Sentry project the web/mobile apps already report to.

What it does:
- **Session replay (web)**: enables `Sentry.replayIntegration()` in **buffer mode** — `replaysSessionSampleRate: 0` (no continuous recording), `replaysOnErrorSampleRate: 1.0`, with `maskAllText` / `maskAllInputs` / `blockAllMedia`. We keep a rolling in-memory buffer and only upload on an error or when a bug report is filed. `service/bugs.ts` force-flushes the replay on report and sends its id in a new proto field; the backend turns it into a `**Session replay**` link in the issue.
- **Per-user link**: when the reporter is logged in, the issue gets a `**Sentry (this user)**` link to their recent frontend errors. Closes a gap where `Sentry.setUser` wasn't re-applied on page reload from persisted auth, so returning sessions stay attributed.
- The new `sentry_replay_id` (tag 9 on `ReportBugReq`) is validated against `^[0-9a-f]{32}$` server-side before interpolation, so a malformed client value can't be injected into the issue markdown.

Notes for reviewers:
- Buffer mode captures roughly the last ~60s before flush (the lead-up to the report), not the whole session.
- The links are assembled in `couchers/sentry.py`, since `urls.py` is for our own URLs. The dashboard host is inline and the project id comes from `SENTRY_FRONTEND_PROJECT_ID`, matching how we already handle third-party endpoints (`api.github.com` inline + `BUG_TOOL_GITHUB_REPO` in config).
- The flush on submit is capped at 2s so a slow upload can't stall the report; the upload continues in the background and the replay id is attached either way.

## Testing
- Backend: full `pytest src` — 1320 passing. Updated the logged-in issue-body assertion and added cases for a valid replay link and for a malformed replay id being omitted.
- Web: full `yarn test` — 137 suites / 956 tests passing; `yarn typecheck` clean. `AuthProvider` now calls `Sentry.setUser`, so three suites that mock `platform/sentry` with only `captureException` needed `setUser` added to the mock.
- `make format` / `make mypy` (clean) and prettier run on changed files.
- Not yet exercised end-to-end against a live Sentry project — reviewer should confirm the replay flush + link resolves in staging.

**Backend checklist**

**Web frontend checklist**

## For maintainers


jesseallhands (approved):
lgtm

PR #9122, opened 2026-06-15
Backend/emails: Use user content as the preview line

tristanlabelle:
We previously inconsistently either reused or paraphrased the email subject as a preview line. I think we should use it to surface the most important information from the email body not redundant with the subject, which is any quoted user content such as the message content or event description. If we don't have that, don't include a preview line (Gmail will show the beginning of the email).

Since those can contain markdown, but preview lines wouldn't display them, adds a `markdown_to_plaintext` function and refactors our markdown conversion to its own file, with new tests. This allowed me to find a discrepancy with respect to how newlines are interpreted vs the frontend, which I'm aso fixing (see `breaks` option and `newline` rule).

## Testing
Added markdown/html-related tests. Preview lines themselves are difficult to test without the Gmail client (MailDev) doesn't surface them.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/emails.py
+++ b/app/backend/src/couchers/email/emails.py
@@ -271,6 +271,9 @@ class ChatMessageReceivedEmail(EmailBase):
             loc_context, ".subject", {"author": self.author.name, "group": self.group_chat_title or ""}
         )
 
+    def get_preview_line(self, loc_context: LocalizationContext) -> str | None:
+        return self.text
+
     def get_body_blocks(self, loc_context: LocalizationContext) -> list[EmailBlock]:
         builder = self._body_builder(loc_context)
         builder.para(".body", {"author": self.author.name, "group": self.group_chat_title or ""})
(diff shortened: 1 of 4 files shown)


aapeliv (commented):
what about when the user puts in something naughty? i'm not opposed but we gotta be careful here.

how about instead using the subtitle similar to what we do on ios?

tristanlabelle:
@aapeliv 

> what about when the user puts in something naughty? i'm not opposed but we gotta be careful here.

Consider a worst case scenario: A & B are in a couple, A sees that B received a notification email for a host request message saying "let's sleep together" within its first few chars. Would B think of blaming Couchers? If A had access to B's email, they could have opened it and seen the message, if A saw an email push notification on B's locked phone, B could have disabled previews when the phone is locked, if A was looking at the Gmail client over B's shoulder, I think that relationship has other issues :P. Basically I think it's outside of Couchers' problem space.

> how about instead using the subtitle similar to what we do on ios?

Emails don't have a subtitle. The email client picks up the first text from the inner HTML, which for us will by default be "Hi <name>" and then paraphrasing the subject (not very useful). The preview line is injected at the top of the html body with styling that makes it invisible, so the client picks it up and displays it as to the right end of these:

image

aapeliv:
> > what about when the user puts in something naughty? i'm not opposed but we gotta be careful here.
> 
> Consider a worst case scenario: A & B are in a couple, A sees that B received a notification email for a host request message saying "let's sleep together" within its first few chars. Would B think of blaming Couchers? If A had access to B's email, they could have opened it and seen the message, if A saw an email push notification on B's locked phone, B could have disabled previews when the phone is locked, if A was looking at the Gmail client over B's shoulder, I think that relationship has other issues :P. Basically I think it's outside of Couchers' problem space.

I was thinking more along the lines of scammy stuff where it looks like the message is system-generated because it's in the preview line.

Either way, I do not think this is any worse than our current practice of putting much user generated content in the subject line. I had a think about this and I see this as no more risky.

aapeliv (approved):

PR #9127, opened 2026-06-16
Add feature flag docs

aapeliv:


nabramow:
@aapeliv This is a great start! One thing that would be super helpful is to add examples of the flags in use for frontend and backend to the doc. Maybe we could copy/paste some existing code blocks where the different types of flags are being used?

nabramow:
When do I use `useFeatureValue` and `useFeatureIsOn` and what's the difference between them?

nabramow:
Can you add some info for how to make a feature across frontend and backend visible on dev and staging but not prod?

nabramow:
Maybe some info about how to use this in the dev flow, i.e. do we need to manually switch the feature flags on when we code to see the feature while it's in progress?

docs/feature-flags.md
```diff
@@ -0,0 +1,48 @@
...
+* **Description**: Give it a meaningful description in Markdown.
+* **Can be toggled per-user?** tick this if you always use this via `context` in the backend, and if it is possible enable this feature per-user. E.g. some background jobs cannot be toggled per-user as they either run or don't run, or some backend sampling logic that is not associated to users. On the other hand, even if it doesn't make much sense to show a feature per-user, but it is possible, you should tick this option. It tracks possibility, not whether it makes sense.
+
+![Screenshot showing how to add a feature](feature-flag-setup.png)
+
+## Local development
+
+The local backend does not pull the feature flags from the server, instead it resolves them from the override file in `feature-flags.dev.json`. You should set a value here that makes sense for devs developing locally (e.g. enable experimental features), and you can use it for local debug/testing temporary overrides (but don't commit this).
+
+TODO: there is currently no such mechanism in the frontend; you can just manually modify the place where you check the flags.
```
nabramow, line 36: Let's move this right to the top so it's obvious no frontend for now?

docs/feature-flags.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Setting the value of a feature flag to a non-default value
+
+To set the value of a feature flag to a non-default value for everyone for a given environment, follow this process:
```
nabramow, line 40: It's not super clear to me what this means and when we would want to do it?

aapeliv:
combining into https://github.com/Couchers-org/couchers/pull/9504

PR #9133, opened 2026-06-17
Events: Remove support for online events

### LINKED ISSUE #9004: Remove support for online events
As discussed and unanimously agreed during the 2026-06-05 dev meeting. Online events don't fit in local communities, since anyone can join them, and it also makes no sense to put them all in the global community (there would be too many of them to be useful), they're also seeing little usage.

This will remove complexity when fixing up event timezones.

Since there are several online events in the database and probably some still scheduled, we should only remove support for creating new online events for now, they should still be viewable.

tristanlabelle:
As decided during dev meetings, since online events doesn't fit will in our communities structure and are underused for the complexity they bring.

On the backend: 

- Remove API support for creating/updating/scheduling online events.
- Database can still represent legacy online events.
- Query APIs will now return existing legacy online events as offline, with address = link at coords (0, 0).
- Remove API support for filtering on online/offline events (since they're all surfaced as offline).
- Keep online event protobuf representation for already queued notifications (or do we care? it only allows showing the link in emails)
- Update tests

On the frontend:

- Remove display and editing UI for online events.
- Remove search filtering of online/offline events.
- Update tests.

Closes #9004

## Testing
Ran backend + frontend locally. Viewed community events, created an event, updated an event.

image
image
image

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/proto/events.proto
```diff
@@ -168,7 +168,8 @@ message CreateEventReq {
   // from media server
   string photo_key = 3;
   oneof mode {
-    OnlineEventInformation online_information = 4;
+    // Server will now reject virtual events
+    OnlineEventInformation online_information = 4 [deprecated = true];
```
aapeliv: just remove + reserve

aapeliv (commented):
i'm not sure what we should do about existing online events

app/proto/events.proto
tristanlabelle: That's what I first did but then I didn't have a way to detect on the backend that a client was attempting to create an online event to show an appropriate error because I couldn't refer to `online_information`. If I remove it I can only show a generic "missing location" error. Or do we assume that we always have matching frontend-backend pair, outside of devs using a local frontend?

tristanlabelle:
> i'm not sure what we should do about existing online events

I'm kicking the can a bit by preventing them from being created/scheduled. At some point they'll mostly/all be past events and we can consider the next stage of deprecation.

app/proto/events.proto
tristanlabelle: From dev meeting, we decided to remove as much as possible, and surface legacy online events as being offline.

app/backend/src/couchers/email/locales/en.json
tristanlabelle, line 1: Self-blocking pending local testing.

tristanlabelle:
@aapeliv , ready for review. I've now fully removed online events from the API.

aapeliv (dismissed):
lgtm: i would go a step further and just do a migration to kill online events from the DB

app/web/features/communities/events/EventCard.test.tsx
```diff
@@ -7,7 +7,7 @@ import EventCard from "./EventCard";
 
 const { t } = i18n;
 
-const [firstEvent, secondEvent, thirdEvent, cancelledEvent] = events;
+const [firstEvent, _, thirdEvent, cancelledEvent] = events;
```
nabramow: We should just get rid of the online event in the fixtures so we don't need the underscore here. That's just dummy data anyway.

app/web/features/communities/events/EventPage.test.tsx
```diff
@@ -25,7 +25,7 @@ const { t } = i18n;
 
 jest.mock("components/MarkdownInput");
 
-const [firstEvent, secondEvent, thirdEvent] = events;
+const [firstEvent, _, thirdEvent] = events;
```
nabramow: Same here let's just remove that fixture data

app/web/.eslintrc.json
```diff
@@ -54,8 +54,8 @@
     "unused-imports/no-unused-vars": [
       "warn",
       {
-        "argsIgnorePattern": "theme",
-        "varsIgnorePattern": "classes|useStyles"
+        "argsIgnorePattern": "_|theme",
+        "varsIgnorePattern": "_|classes|useStyles"
```
nabramow: This looks sus, why are you changing this?

app/web/test/fixtures/events.json
```diff
@@ -58,8 +58,10 @@
     "canEdit": false,
     "canModerate": false,
     "isNext": false,
-    "onlineInformation": {
-      "link": "http://zoom.link"
+    "offlineInformation": {
+      "address": "Amsterdam",
+      "lat": 1,
+      "lng": 2
```
nabramow: I'm gonna assume AI did something weird here? I don't think we changed the data shape to have "offlineInformation"? Probably we just want to remove this?

nabramow:
What will we do for a past event, for example, if someone goes directly to the url for it and it used to be an online event? It's an edge case but still could happen I guess.

app/web/test/fixtures/events.json
tristanlabelle: Nothing weird, other entries in this file all have `offlineInformation`, this turns the event into an offline event.

app/web/features/communities/events/EventPage.test.tsx
tristanlabelle: The problem is that we have pagination tests based on having 3 events: The first page shows the first two, and the second page shows the third one. Using 1 item-sized pages seems like a degenerate case, and we can't use the fourth event because it's canceled, so I think we still want 3 events.

tristanlabelle:
> What will we do for a past event, for example, if someone goes directly to the url for it and it used to be an online event? It's an edge case but still could happen I guess.

See PR description:

> Query APIs will now return existing legacy online events as offline, with address = link at coords (0, 0).

app/web/.eslintrc.json
tristanlabelle: Because no test now references the second event, but we still need it for pagination so we have three events. List deconstruction needs a name.

Let me see if I can swap the second and third events instead.

tristanlabelle:
@nabramow Ok in the latest I've swapped the second and third event so I can pattern match only the first two, and we still have three events for pagination tests. I don't need the `_` name binding anymore.

app/web/test/fixtures/events.json
```diff
@@ -36,12 +36,12 @@
     "timezone": "Europe/Amsterdam"
   },
   {
-    "attendanceState": 2,
-    "content": "It's the beginning of spring! Join us for a chat about we plan to plant and get tips on how to keep your plants healthy.",
-    "creatorUserId": 3,
-    "eventId": 2,
-    "goingCount": 16,
-    "maybeCount": 0,
+    "attendanceState": 0,
```
tristanlabelle, line 39: This event and the next were swapped, and the online event was made offline.

PR #9125, opened 2026-06-16
Revert PR #9049 (language picker flag removal)

aapeliv:
Reverts #9049, which broke the language picker in the top-right header when logged in, as well as the language picker in settings.

This reverts merge commit `6c4620478` (keeping `develop` as the mainline), restoring the previous language picker behavior and the flag icons it had removed.

## Testing
- Reverted the merge cleanly; working tree builds from the prior known-good state.
- Reviewer should verify locally: while logged in, open the language picker in the top-right header and confirm it renders and switches languages correctly, and check the language picker in Settings.

**Web frontend checklist**
- [ ] There are no console warnings when running the app
- [ ] Added tests where relevant
- [ ] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

## For maintainers

- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

_This PR was created with the Couchers PR skill._

PR #9137, opened 2026-06-18
Revert "Revert PR #9049 (language picker flag removal)"

kevinortiz43:
Reverts Couchers-org/couchers#9125


kevinortiz43:
Reverified with Aapeli and Tristan, new feature did not cause an issue with the language dropdown in production. Requesting the reversion to be reverted

aapeliv (approved):

PR #9141, opened 2026-06-19
web: Add native app install upsell for Android and iOS

### LINKED ISSUE #9142: Look into how to nudge Android app download
It works to nudge apple users from the web browser to download the app. We're not sure why it's not working on Android.

- Look into how this should work on Android and implement and test it

aapeliv:
Adds native-app install upsell signals on both platforms so the site nudges visitors toward the native apps, complementing the existing deep-link associations (`apple-app-site-association` / `assetlinks.json`).

**Android** — adds `related_applications` and `prefer_related_applications: true` to the web app manifest so Chrome prefers the native Android app over the installable PWA. Following the existing staging→prod swap convention (used for `robots`, `assetlinks`, `apple-app-site-association`):
- `public/manifest.json` (default) uses the staging package `org.couchers.staging.android`.
- `public/manifest-prod.json` (new) uses the prod package `org.couchers.android` plus an `itunes` entry, copied over `manifest.json` in `Dockerfile.prod`.

**iOS** — adds the `apple-itunes-app` meta tag in `_document.tsx` so Safari shows the App Store Smart App Banner. Gated to `NEXT_PUBLIC_COUCHERS_ENV === "prod"` since only the prod app (app-id `6623776751`) is listed on the App Store.

Behavioural notes:
- `prefer_related_applications: true` suppresses Chrome's PWA install prompt in favour of the native app. Chrome also no longer shows automatic native-app install banners, so the Android side is a signal rather than a visible banner.
- The iOS `apple-itunes-app` tag *does* produce a prominent banner at the top of Safari on every page — so it's the more visible of the two changes.

## Testing
- Validated both manifest JSON files parse correctly; linted `_document.tsx`.
- Verified the `Dockerfile.prod` swap line mirrors the existing prod overrides.
- Reviewer can confirm: `/manifest.json` shows the staging package on staging and the prod package on a prod build; the Smart App Banner appears in Safari only on the prod environment.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/pages/_document.tsx
+++ b/app/web/pages/_document.tsx
@@ -24,6 +24,10 @@ export default function MyDocument(
 
         <link rel="preconnect" href="https://cdn.couchers.org" />
         <meta name="theme-color" content={theme.palette.primary.main} />
+        {/* iOS Smart App Banner; only the prod app is listed on the App Store */}
+        {process.env.NEXT_PUBLIC_COUCHERS_ENV === "prod" && (
+          <meta name="apple-itunes-app" content="app-id=6623776751" />
+        )}
         <link rel="manifest" href="/manifest.json" />
         <link rel="apple-touch-icon" href="/logo512.png" />
         <link
--- a/app/web/public/manifest-prod.json
+++ b/app/web/public/manifest-prod.json
@@ -26,8 +26,8 @@
   "related_applications": [
     {
       "platform": "play",
-      "id": "org.couchers.staging.android",
-      "url": "https://play.google.com/store/apps/details?id=org.couchers.staging.android"
+      "id": "org.couchers.android",
+      "url": "https://play.google.com/store/apps/details?id=org.couchers.android"
     }
   ]
 }
--- a/app/web/Dockerfile.prod
+++ b/app/web/Dockerfile.prod
@@ -17,6 +17,7 @@ RUN cp .env.$environment env && \
     mv env .env.local && \
     if [ "$environment" = "production" ]; then \
         cp public/robots-prod.txt public/robots.txt && \
+        cp public/manifest-prod.json public/manifest.json && \
         cp public/.well-known/apple-app-site-association-prod public/.well-known/apple-app-site-association && \
         cp public/.well-known/assetlinks-prod.json public/.well-known/assetlinks.json; \
     fi


nabramow:
Does this cover #9142 ?

aapeliv:
> Does this cover #9142 ?

i think so!

app/web/public/manifest.json
```diff
@@ -21,5 +21,13 @@
...
   "display": "standalone",
   "theme_color": "#00a398",
-  "background_color": "#ffffff"
+  "background_color": "#ffffff",
+  "prefer_related_applications": true,
+  "related_applications": [
+    {
+      "platform": "play",
+      "id": "org.couchers.staging.android",
+      "url": "https://play.google.com/store/apps/details?id=org.couchers.staging.android"
```
nabramow, line 30: I don't think we want the staging one here.

app/web/public/manifest-prod.json
```diff
@@ -0,0 +1,37 @@
...
+  "background_color": "#ffffff",
+  "prefer_related_applications": true,
+  "related_applications": [
+    {
+      "platform": "play",
+      "id": "org.couchers.android",
+      "url": "https://play.google.com/store/apps/details?id=org.couchers.android"
+    },
+    {
+      "platform": "itunes",
```
nabramow: iTunes?

app/web/public/manifest.json
aapeliv: we replace this with the prod one for prod. having prod defaults is kinda not good as it can slip through

app/web/public/manifest-prod.json
aapeliv: thanks; it's an old platform but it seems apple doesn't do anything with it anymore! removed.

app/web/pages/_document.tsx
```diff
@@ -24,6 +24,10 @@ export default function MyDocument(
 
         <link rel="preconnect" href="https://cdn.couchers.org" />
         <meta name="theme-color" content={theme.palette.primary.main} />
+        {/* iOS Smart App Banner; only the prod app is listed on the App Store */}
+        {process.env.NEXT_PUBLIC_COUCHERS_ENV === "prod" && (
```
nabramow, line 28: Do we not want to use `NODE_ENV` here? I sort of remember some of the others being a bit wonky due to our dev setup.

app/web/pages/_document.tsx
nabramow, line 18:
```suggestion
  props: DocumentProps & DocumentHeadTagsProps & { appArgument?: string },
```

see comment below

app/web/pages/_document.tsx
```diff
@@ -24,6 +24,10 @@ export default function MyDocument(
...
+          <meta name="apple-itunes-app" content="app-id=6623776751" />
+        )}
```
nabramow, line 30:
```suggestion
        {/* iOS Smart App Banner; only the prod app is listed on the App Store.
            app-argument passes the current URL so tapping the banner opens that page in the app
            (our apple-app-site-association claims all paths) rather than the app's home screen. */}
        {process.env.NEXT_PUBLIC_COUCHERS_ENV === "prod" && (
          <meta
            name="apple-itunes-app"
            content={`app-id=6623776751${
              props.appArgument ? `, app-argument=${props.appArgument}` : ""
            }`}
          />
        )}
```

app/web/pages/_document.tsx
nabramow, line 61:
```suggestion
  return {
    ...finalProps,
    appArgument: `${process.env.NEXT_PUBLIC_BASE_URL}${ctx.asPath ?? ""}`,
  };
```

nabramow:
So I looked into this, and what I read it that iOS has this built in OPEN vs GET app banner (only on mobile safari browser), but Android (and not safari) doesn't.

For Android, `manifest.json` declares `related_applications` + `prefer_related_applications: true`. This does not produce a banner, Chrome has no manifest-driven native-app promotion. Its only real value is enabling install detection. So Android and Chrome-iOS users currently see nothing, and closing that gap needs a custom component.

So I think we'd needs to build a custom banner for android, then do logic to make sure the iOS and Android one don't appear at the same time, and get it to not be fussy with webviews.

I guess your choice here is make this an iOS only PR or do the whole song and dance for Android.

Feel free to check me though this is just what I found when I looked up how to do it

aapeliv:
@nabramow i will close this for now, it's bigger than i thought and i don't have expertise here

PR #9146, opened 2026-06-19
Backend: Add reverse lookup for upload uses

aapeliv:
Adds a system to answer "for a given upload, here are all its uses". `get_upload_uses(session, key)` in `couchers/upload_uses.py` is a reusable helper that derives every place an upload is referenced — directly from the existing foreign keys to `uploads.key`, so there's no new table to keep in sync.

It returns a list (an upload can be used in several places at once) of typed uses:
- `profile_gallery_photo` / `profile_gallery_photo_avatar` (the first-by-position gallery item is the avatar)
- `event` (the occurrence id, as exposed by the API)
- `page`

Each use carries `is_current`: gallery items are always current; an event occurrence is current unless deleted; a page version is current only if it's the page's latest version. Historical/superseded references are still returned, just flagged `is_current=False`.

These uses are now surfaced in the admin `ListUserUploads` endpoint (new `UploadUse`/`UploadUseType` in `admin.proto`).

To keep the reverse lookup complete, a drift-guard test introspects SQLAlchemy metadata for every foreign key targeting `uploads.key` and asserts each is handled by the helper, and sync-reminder comments were added near the `Upload` model and the media servicer.

## Testing
- New `test_upload_uses.py` covers gallery (avatar vs non-avatar), event (incl. deleted → not current), page (incl. superseded version → not current), an upload used in multiple places, unused uploads, and the FK drift guard.
- Added `test_ListUserUploads_uses` asserting uses surface through the admin API.
- `make format` clean; `make mypy` clean apart from a pre-existing, unrelated `ics` stub error in `test_calendar_events.py`; relevant test suites pass.

**Backend checklist**
- [ ] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no database changes)

## For maintainers


jesseallhands (dismissed):
Lgtm

PR #9147, opened 2026-06-19
Add GetDashboardV2 RPC to aggregate dashboard content

aapeliv:
Adds a single `GetDashboardV2` RPC (on a new `Dashboard` service) that returns all the content the dashboard loads today in one call, instead of the frontend fanning out to ~6 separate RPCs (reminders, surfing & hosting host requests, my events, community events, and community discussions).

The win is removing the per-request round-trip and interceptor/write overhead paid 6× today (auth, `last_active`/`UserActivity` writes, connection-pool checkout), not reducing query count — the same queries run, now in one transaction.

The query logic for each section is extracted into module-level functions (`get_reminders`, `list_host_requests`, `list_my_events`, `list_my_communities_discussions`) that **both** the existing per-section RPCs and the new aggregator call, so there is no duplicated query/visibility/moderation logic. The existing servicer methods now delegate to these.

This is **backend only** and a **draft** — the web frontend wiring to consume the new RPC is intentionally not included yet.

Note: generated proto stubs are gitignored and regenerated by CI from `dashboard.proto`.

## Testing
- Added `app/backend/src/tests/test_dashboard.py`: asserts `GetDashboardV2` is byte-identical to calling the 6 individual RPCs with the same parameters the web frontend uses, plus role-based bucketing (surfer vs host) and the empty-state.
- `make mypy` and `make format` pass.
- The dashboard test suite passed locally. The local testing DB went down before I could re-run the full `requests`/`account`/`events`/`discussions` suites to confirm the extraction is behavior-preserving — relying on CI for that; the extraction is a pure delegation so existing behavior should be unchanged.

**Backend checklist**
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (n/a — no DB changes)

## For maintainers


nabramow:
Shall we call this `getDashboard` not `getDashboardV2` since we don't have a v1 for this?

app/backend/src/couchers/servicers/dashboard.py
```diff
@@ -0,0 +1,74 @@
...
+                    only_active=True,
+                    status_in=[
+                        conversations_pb2.HOST_REQUEST_STATUS_ACCEPTED,
+                        conversations_pb2.HOST_REQUEST_STATUS_CONFIRMED,
+                    ],
+                    sort_by=requests_pb2.HOST_REQUEST_SORT_BY_FROM_DATE,
+                ),
+                context,
+                session,
+            ),
```
nabramow, line 53: These two blocks are identical except `only_sent` vs `only_received`. You could extract a small helper like `_upcoming_host_requests_req(surfing: bool)` so the parameters can't drift apart between the two?

nabramow:
@aapeliv I pushed some changes to add the `exclude_attending` logic I merged since you started this PR (and added a test).

I also went ahead and tested that the backend runs fine locally. Code looks good to me, just that one nit comment.

I'd love to get this merged so I can do #9286 on top of this logic instead of what we have now (as it would be cleaner IMO).

PR #9158, opened 2026-06-21
Backend/i18n: Externalize fallbacks languages from I18Next

tristanlabelle:
Instead of baking the locale fallbacks into the `I18Next` object, pass a locales list when localizing strings.

This allows implementing all-or-nothing localization, say for emails, as follows (pseudocode):

```py
  try:
    return localize_emails(locales=["fr-CA", "fr"]) # No fallback to English, will raise an error.
  except LocalizationError:
    return localize_emails(locales=["en"])
```

It also splits the responsibility of knowing our fallback chain from the implementation of the I18Next format.

## Testing
Updated tests.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/blocks.py
+++ b/app/backend/src/couchers/email/blocks.py
@@ -51,7 +51,7 @@ class EmailBase(ABC):
         standard_closing: bool = True,
         security_warning: bool = False,
     ) -> EmailBlocksBuilder:
-        builder = EmailBlocksBuilder(locale=loc_context.locale, string_key_base=self.string_key_base)
+        builder = EmailBlocksBuilder(locales=loc_context.locale_list, string_key_base=self.string_key_base)
         if standard_greeting:
             builder.para("generic.greeting_line", {"name": self.user_name})
         if standard_closing:
@@ -78,7 +78,7 @@ class EmailBase(ABC):
         self, loc_context: LocalizationContext, key: str, substitutions: SubstitutionDict | None = None
     ) -> str:
         key = full_string_key(key, relative_base=self.string_key_base)
-        return get_emails_i18next().localize(key, loc_context.locale, substitutions)
+        return loc_context.localize_string(key, i18next=get_emails_i18next(), substitutions=substitutions)
 
 
 @dataclass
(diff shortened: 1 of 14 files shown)


aapeliv (commented):
I feel like this is not quite the right system: now you have different areas of the code worrying about locale lists, which to me shouldn't be the concern of application code?

I feel like what you want here is an "all-or-nothing localization block", where you guarante that every translation there is localized to a given language or fully falls back. We may want a similar one for e.g. notifications (so you don't get a partially translated notification)? Similarly you could imagine you'd want it in an RPC. Though that seems difficult to achieve in cases where you can't fall back and retry with a try-except? It seems it may require some kind of lazy evaluation in most cases?

If you feel this is the best way, I'm also happy with this. But I feel like a cleaner abstraction would be even a limited version of this kind of "localization context/block" that raises if there is a missing translation, but does not contain the local fallback chain, just a boolean basically that it's in an all-or-nothing section, or similar. Then at least we don't need to embed the locale fallback chain in application code.

tristanlabelle:
> I feel like this is not quite the right system: now you have different areas of the code worrying about locale lists, which to me shouldn't be the concern of application code?

I convinced myself that having low-level loc APIs take a locale list is the most flexible approach, and that's probably why all the javascript `Intl.Xxx` APIs are designed that way. Even if you could know ahead of time or lazy evaluate all loc string lookups to know if some will fail, you'd need locale lists, because you wouldn't want to try `pt` and otherwise `pt-BR`, and otherwise `en`, you'd want to try `["pt", "pt-BR"]` (mutually intelligible fallbacks) and otherwise just `"en"`.

But also currently the date/language/region lookups do not support fallbacks and I'm worried a language could get broken because of this without us realizing. There's no `I18Next` for those so I'd want to also take a locale list, and in that case I want the decision of the fallback chain to above both `I18Next` and the date/language/region functions, i.e. in `LocalizationContext`.

For transational localization, the API would look something like `LocalizationContext.localize(callback: Function[list[Locale]])`, calling the callback a second time if the first time throws. I could also encapsulate the list of locales in a `LocaleChain` object and take that in low-level functions if the list is too ugly.

aapeliv (dismissed):
thanks for engaging with it. i think your solution here is clean enough, and i'm sure it will evolve over time anyway!

PR #9160, opened 2026-06-22
Frontend/search: Show strong verif badges

tristanlabelle:
We forgot to show the strong verification badge in search results, where it's quite useful.

## Testing
Since NEXT doesn't have strong verified users, I injected `user.hasStrongVerification = true` for testing:

image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/view/UserOverview.tsx
+++ b/app/web/features/profile/view/UserOverview.tsx
@@ -126,11 +126,9 @@ export default function UserOverview({
       )}
 
       <StyledWrapper>
-        <StyledIntro variant="h1">
-          <span>
-            {user.name}
-            {user.hasStrongVerification && <StrongVerificationBadge />}
-          </span>
+        <StyledIntro variant="h1" gap={0.5}>
+          {user.name}
+          {user.hasStrongVerification && <StrongVerificationBadge />}
         </StyledIntro>
         <StyledLink
           href={routeToUser(user.username)}
--- a/app/web/features/profile/view/UserOverview.tsx
+++ b/app/web/features/profile/view/UserOverview.tsx
@@ -129,7 +129,7 @@ export default function UserOverview({
         <StyledIntro variant="h1">
           <span>
             {user.name}
-            {user.hasStrongVerification ? <StrongVerificationBadge /> : null}
+            {user.hasStrongVerification && <StrongVerificationBadge />}
           </span>
         </StyledIntro>
         <StyledLink
--- a/app/web/features/search/SeachResultUserCard.tsx
+++ b/app/web/features/search/SeachResultUserCard.tsx
@@ -3,6 +3,7 @@ import { FlexboxProps, useMediaQuery } from "@mui/system";
 import Avatar from "components/Avatar";
 import { OpenInNewIcon } from "components/Icons";
 import ProfileLink from "components/ProfileLink/ProfileLink";
+import StrongVerificationBadge from "components/StrongVerificationBadge";
 import StyledLink from "components/StyledLink";
 import { useImpressionRef, useLogEvent } from "features/analytics/hooks";
 import { useSearchAnalytics } from "features/analytics/searchAnalyticsContext";
@@ -291,6 +292,10 @@ const SearchResultUserCard = ({
                   }}
                 >
                   {user.name}
+                  {"hasStrongVerification" in user &&
+                  user.hasStrongVerification ? (
+                    <StrongVerificationBadge />
+                  ) : null}
                 </Typography>
               </ProfileLink>
             </FlexRow>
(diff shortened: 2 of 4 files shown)


app/web/features/search/SeachResultUserCard.tsx
```diff
@@ -291,6 +292,10 @@ const SearchResultUserCard = ({
                   }}
                 >
                   {user.name}
+                  {"hasStrongVerification" in user &&
+                  user.hasStrongVerification ? (
+                    <StrongVerificationBadge />
+                  ) : null}
```
nabramow:
```suggestion
                  {user.hasStrongVerification && <StrongVerificationBadge />}
```

The "hasStrongVerification" in user check is redundant. If the key doesn't exist, user.hasStrongVerification is just undefined, which is falsy, so && short-circuits anyway. 

The ? (...) : null is also unnecessary. When the left side of && is falsy, it evaluates to false, which React doesn't render. So the explicit null branch adds nothing.

tristanlabelle: Ah, I copied that dance from elsewhere, but now I see it had a special typing situation going on. Fixed in latest.

nabramow:
@tristanlabelle I think we need to align it with the name vertically so they are centered? The badge seems higher than the text.

Screenshot 2026-06-22 at 16 42 09



btw both Aapeli and I have strong verification badges on next if you query in NYC or London!

tristanlabelle:
The way I have it looks exactly the same as on the profile page (verified by overlapping the screenshots). Does it look off to you on the profile page as well? If so it might be a separate change.

nabramow:
> The way I have it looks exactly the same as on the profile page (verified by overlapping the screenshots). Does it look off to you on the profile page as well? If so it might be a separate change.

Yeah same on the profile page. It's more obvious in this image where you can see it falls in the top of the box: 

Screenshot 2026-06-22 at 20 55 57

app/web/features/profile/view/UserOverview.tsx
```diff
@@ -129,7 +129,7 @@ export default function UserOverview({
         <StyledIntro variant="h1">
           <span>
             {user.name}
-            {user.hasStrongVerification ? <StrongVerificationBadge /> : null}
+            {user.hasStrongVerification && <StrongVerificationBadge />}
```
avreally: We should avoid unsafe "&&" operator for conditional rendering as it might cause unintentionally rendered values. The ternary was actually correct, was it not working?

app/web/features/search/SeachResultUserCard.tsx
```diff
@@ -291,6 +292,7 @@ const SearchResultUserCard = ({
                   }}
                 >
                   {user.name}
+                  {user.hasStrongVerification && <StrongVerificationBadge />}
```
avreally: And the same here

app/web/features/profile/view/UserOverview.tsx
nabramow: @avreally I think we only have to worry about that for example if it's a `.length` value that might evaluate to zero. This should just evaluate to undefined or false.

avreally: @nabramow there can be other situations as well, so it's a good pattern to use ternary as it's more reliable. But that's just my opinion 🙂

tristanlabelle: I switched to `&&` at @nabramow 's [suggestion](https://github.com/Couchers-org/couchers/pull/9160#discussion_r3450704754). I checked the docs and React documents both the ternary and `&&` patterns as legit: https://react.dev/learn/conditional-rendering#logical-and-operator- .

app/web/features/search/SeachResultUserCard.tsx
tristanlabelle: See https://github.com/Couchers-org/couchers/pull/9160#discussion_r3450704754 and https://github.com/Couchers-org/couchers/pull/9160#discussion_r3462104231

tristanlabelle:
@nabramow It looks worse with `display: inline-block; vertical-align: middle;`. I think it makes sense that the icon sits on the text baseline as it currently does.

image

`font-size: 90%` makes it more the size of the text, if you like that better.

image

But I'd like a ✅ for the fix I'm implemented in all cases.

nabramow:
> @nabramow It looks worse with `display: inline-block; vertical-align: middle;`. I think it makes sense that the icon sits on the text baseline as it currently does.
> 
> image
> `font-size: 90%` makes it more the size of the text, if you like that better.
> 
> image
> But I'd like a ✅ for the fix I'm implemented in all cases.

Such is frontend work, I'm afraid. Would you like me to push the change to your branch? Generally you fix this by wrapping both in a box and centering them.

tristanlabelle:
> Would you like me to push the change to your branch?

I would rather we merge this and let you follow up with the visual tweak, but otherwise go ahead.

nabramow (approved):
Put in some changes to do the style the more proper react way. If you're cool with it go ahead and merge, I tested all the components with changes to check they look okay.

PR #9162, opened 2026-06-22
Backend: exclude hidden references from the reference count

### LINKED ISSUE #2374: A user's reference count doesn't always match with the number of those that are visible
Subject: A user's reference count doesn't always match with the number of those that are visible
Description:
Visit the profile (see exact link in report for example at time of report), where they presumably have hosted/surfed recently and someone has left them a reference, but they haven't done so yet.

Results:
The number of references the user has should have excluded ones that haven't come up in the profile yet (or at least that's my assumption of the problem)

Backend version: develop-0615fc3a
Frontend version: develop-0615fc3a
User Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:95.0) Gecko/20100101 Firefox/95.0
Page: https://app.couchers.org/user/chel/references
User: test (132)

### LINKED ISSUE #4946: Reference doesn't appear
Subject: Reference doesn't appear
Description:
Nothing

Results:
I got a notification about my host giving me a reference. I opened the references tab on my profile. The reference is indicated in the reference count, but the reference itself is missing.

Backend version: develop-06099492
Frontend version: develop-06099492
User Agent: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/129.0.0.0 Mobile Safari/537.36
Page: https://couchers.org/profile/references
User: bubatov (45225)

dymbe:
image

The reference count on a user's profile (User.num_references) counted all references received, but ListReferences hides host/surf references until the recipient writes their reciprocal reference or the 2-week writing window closes. The count therefore included references the list hides, so a profile could show a count of 1 even though the references list is empty. This is confusing.

The issue has previously been reported in #2374 and #4946, in the latter it was described as working as intended, but I would challenge this assertion.

Solution: extract the list's visibility rule into a shared reference_publicly_visible() helper and apply it in both ListReferences and get_num_references so the two can't drift. The list refactor is behaviour-equivalent.

Closes #2374, closes #4946

## Testing

- Added a backend regression test (`test_num_references_matches_visible_list`)
  asserting the count equals the visible list both while a reciprocal reference is
  hidden (0/0) and after it's written (1/1). Verified it **fails on develop**
  (count 1, list 0) and passes with this change.
- `make format` + `make mypy` clean; `test_references.py` (16) and `test_api.py` (46)
  pass locally.
- Manually verified in a local stack: a profile with a hidden reciprocal reference
  shows a matching badge/list with this change, vs an inflated badge on develop.

**Backend checklist**
- [x] No database changes, so no migrations needed

## For maintainers


app/backend/src/couchers/sql.py
```diff
@@ -256,3 +259,37 @@ def _relevant_user_blocks(user_id: int) -> Select[tuple[int]]:
 
 def to_bool(value: bool) -> ColumnElement[bool]:
     return true() if value else false()
+
+
+def reference_publicly_visible() -> ColumnElement[bool]:
```
aapeliv: this is very weird to me how this works; given that it's not taking any parameters? also the name should be more descriptive. you can move it to references.py and import into api.py

app/backend/src/tests/test_references.py
```diff
@@ -411,6 +417,41 @@ def test_ListPagination(db):
         assert not res.next_page_token
 
 
+def test_num_references_matches_visible_list(db):
```
aapeliv, line 429: please add a few asserts into existing tests to make sure they work as expected

aapeliv (changes requested):

nabramow:
@dymbe are you still working on this? Trying to clean up if not!

dymbe:
@nabramow Just got back from vacation, will look into this shortly

app/backend/src/couchers/sql.py
dymbe: Yeah you're right, it's a bit odd. It works because the EXISTS subqueries auto-correlates against the outer query's FROM when the expression is added with `.where()`. But I changed it to the `where_*` helper style. I could not move it to `servicers/references.py` by the way, since this would create an import cycle with `servicers/api.py`. I created a `helpers/references.py` and put it there instead.

app/backend/src/tests/test_references.py
dymbe: I added some new tests, and it seemed to work fine except for when I added it to `test_ListReference_banned_deleted_users`. Seems like there is a similar mismatch for deleted users completely unrelated to this PR, i.e. the list view shows references from deleted users, but the count does not include them. I verified this happens on develop too. I included a commented out assert [here](https://github.com/Couchers-org/couchers/pull/9162/changes#diff-716069b75adae19d1f092853357d732d567e5dd5bdef62efeb58146875c8a99aR513) so you can see what I'm talking about. If you want I'll remove the commented out lines before merging though. I can also create an issue for it and start working on that separately

aapeliv (approved):
this looks great to me now, thank you!

app/backend/src/tests/test_references.py
dymbe: Created a PR for this: #9392

PR #9163, opened 2026-06-22
Frontend/i18n: Localize list separators

tristanlabelle:
Not all languages use the western comma. I'm not using `Intl.ListFormat` directly because it will always add "and" / "or", which we don't want in our current contexts.

## Testing
View a profile with regions visited:

image
image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/messages/groupchats/GroupChatListItem.tsx
+++ b/app/web/features/messages/groupchats/GroupChatListItem.tsx
@@ -67,7 +67,10 @@ export default function GroupChatListItem({
   className,
   isArchived = false,
 }: GroupChatListItemProps) {
-  const { t } = useTranslation(MESSAGES);
+  const {
+    t,
+    i18n: { language: locale },
+  } = useTranslation(MESSAGES);
   const currentUserId = useAuthContext().authState.userId!;
   const latestMessageAuthorId = groupChat.latestMessage?.authorUserId;
 
@@ -93,6 +96,7 @@ export default function GroupChatListItem({
     groupChatMembersQuery,
     currentUserId,
     t,
+    locale,
   );
   //text is the control message text or message text
   let text = "";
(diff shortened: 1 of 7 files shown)


nabramow:
Wow new fun fact for me, I did not know this!

Just checking scope, this is only the regions visited in the profile yeah? Will this eventually apply to everywhere we have commas like in the dashboard widgets, map search results card, etc? Or it's just where we have lists of things?

Screenshot 2026-06-24 at 10 43 34

HankAviator (commented):
Quite detailed-oriented commit. The usage of the enumeration comma has a rule of thumb: there aren't many items (less than or equal to 5 items usually), each item should be short (about 4 characters or less). Considering the country list can be very long, this punctuation will make user feel "busy" or "messy", in that case, a comma is better.

This commit itself is fine but using comma will be much simpler @tristanlabelle

tristanlabelle:
@nabramow It's specifically for lists of things. I'm not sure what the exact boundary is but Chinese uses regular commas between clauses of a sentence, but the enumeration back-comma for lists (especially of nouns).

But we have the expert here :) @HankAviator , Interesting! What happens with longer lists of nouns? What about when we'll have localized the region names as well? Which one would you prefer seeing?

1. 加拿大，墨西哥，美國
2. 加拿大、墨西哥、美國

(assuming those lists are arbitrarily long, as some people have visited 50 countries)

HankAviator:
`，` is completely fine, people rarely use `、` nowadays online (except for grammar police like me but even I use it so rarely as it can't be well understood internationally). In your example both are fine, 2nd looks slightly better to me, but I don't feel it's worthwhile to make the change just for Chinese (not sure about JK though). @tristanlabelle

tristanlabelle:
Ok! I'll drop this change then. Thanks for being such a permissive grammar police :)

PR #9164, opened 2026-06-22
Add exclude_attending to event search query

nabramow:
I didn't realize when I did the other that this is a separate query (since this one supports more searching params). Adding it here too.

## Testing

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/search.py
```diff
@@ -787,6 +787,10 @@ def _user_to_search_user(user_id: int) -> search_pb2.SearchUser:
     def EventSearch(
         self, request: search_pb2.EventSearchReq, context: CouchersContext, session: Session
     ) -> search_pb2.EventSearchRes:
+        if request.attending and request.exclude_attending:
+            context.abort_with_error_code(
+                grpc.StatusCode.INVALID_ARGUMENT, "cannot_combine_attending_and_exclude_attending"
+            )
```
nabramow, line 793: The other one had `include_attending` which was a bit different but I think it still doesn't make sense to allow both at the same time?

app/backend/src/couchers/servicers/search.py
```diff
@@ -849,7 +872,11 @@ def EventSearch(
                 )
                 where_.append(Event.parent_node_id.in_(my_communities))
 
-            statement = statement.where(or_(*where_))
+            if where_:
+                statement = statement.where(or_(*where_))
+
```
nabramow, line 877: an empty OR means "no inclusion filter" (return everything), not "return nothing"?

WouldYouKindly (approved):

Issue #8424, opened 2026-04-29
i18n: Spoken languages are not translatable


tristanlabelle:
For example:

<img width="1156" height="299" alt="Image" src="https://github.com/user-attachments/assets/ce6422a9-ffaa-4c53-83de-ca56a53ee6cf" />

They come from a database that contains English strings:

```py
@functools.cache
def get_language_dict() -> dict[str, str]:
    """
    Get a list of allowed languages as a dictionary of {code: name}.
    """
    with session_scope() as session:
        return {language.code: language.name for language in session.execute(select(Language)).scalars().all()}
```

tristanlabelle:
The frontend has localized mapping from language codes to language names which we can use if the language codes match (2 vs 3 char codes?)

PR #9169, opened 2026-06-23
Frontend: Split key/labels in ProfileTagInput

tristanlabelle:
Refactors `ProfileTagInput` to take key:label pairs as the list of options, for example `{ en: "English", fr: "French" }`. We still display the labels, but only expose the keys to consuming code since those are what the business logic needs. This way consumers don't need to map and reverse-map display strings.

This is related to and motivated by #8424: it will then make it trivial to translate spoken languages by providing a dictionary like `{ en: "Anglais", fr: "Français" }` since we have translation keys for (some) language codes.

No functional changes.

## Testing
Edited profile using the Vercel link on this PR.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/ProfileTagInput.tsx
+++ b/app/web/features/profile/ProfileTagInput.tsx
@@ -155,7 +155,10 @@ export default function ProfileTagInput({
 }: ProfileTagInputProps) {
   const { t } = useTranslation(PROFILE);
 
-  // If any value doesn't map to an option, add a fake option for it.
+  // In case some value doesn't map to an option, add a fake option for it.
+  // For example if value is [en, xx] and options is { en: "English", fr: "French" },
+  // then we extend to { en: "English", fr: "French", xx: "xx" },
+  // so the all values can be displayed.
   const effectiveOptions: Record<string, string> = {
     ...Object.fromEntries(value.map((k) => [k, k])),
     ...options,
--- a/app/web/features/profile/ProfileTagInput.tsx
+++ b/app/web/features/profile/ProfileTagInput.tsx
@@ -19,13 +19,18 @@ import React, { useRef, useState } from "react";
 import { ControllerRenderProps } from "react-hook-form";
 import { theme } from "theme";
 
+interface ProfileTagOption {
+  key: string;
+  label: string;
+}
+
 interface ProfileTagInputProps {
   onChange: (_: unknown, value: string[]) => void;
   value: string[];
-  options: string[];
+  // Possible options keys and their labels
+  options: Record<string, string>;
   label: string;
   id: string;
-  allowCsv?: boolean;
   className?: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   inputFieldProps?: ControllerRenderProps<any, string>;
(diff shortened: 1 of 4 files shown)


app/web/features/profile/edit/EditProfile.tsx
tristanlabelle, line 1: Self-blocking to test once https://next.couchershq.org is back online

app/web/features/profile/hooks/useRegions.ts
```diff
@@ -3,27 +3,17 @@ import { regionsKey } from "features/queryKeys";
...
-          },
-          {
-            regions: {} as { [code: string]: string },
-            regionsLookup: {} as { [name: string]: string },
-          },
+      service.resources
+        .getRegions()
+        .then((result) =>
+          Object.fromEntries(
+            result.regionsList.map(({ alpha3, name }) => [alpha3, name]),
```
nabramow, line 13: Out of curiosity, what is alpha3?

app/web/features/profile/ProfileTagInput.tsx
```diff
@@ -150,6 +155,12 @@ export default function ProfileTagInput({
 }: ProfileTagInputProps) {
   const { t } = useTranslation(PROFILE);
 
+  // If any value doesn't map to an option, add a fake option for it.
+  const effectiveOptions: Record<string, string> = {
+    ...Object.fromEntries(value.map((k) => [k, k])),
+    ...options,
+  };
+
```
nabramow: Don't we want it to default to english?

nabramow:
This seems okay to me, at least I tested and everything still works as before. I assume the actual ability to localize these will come in a follow up PR?

Do we also need to handle the dropdown where people choose the languages?:

Screenshot 2026-06-24 at 10 36 09

app/web/features/profile/hooks/useRegions.ts
tristanlabelle: A 3-letter country code. For example, from the backend `regions.json`:

```json
  {
    "name": "Afghanistan",
    "wikidata": "Q889",
    "alpha2": "AF",
    "alpha3": "AFG"
  },
```

app/web/features/profile/ProfileTagInput.tsx
tristanlabelle: This case is if the spoken language list includes a value that is not part of the options (a bug edge case). For example, if the backend said a user speaks "en", "fr", "xx". Then we extend the options to `en: English, fr: French, xx: xx`. This means we'll still display the "xx" language entry even if we have no name for it. (I also updated the comment with this).

It's independent from localization. It simplifies previous logic meant to handle this case.

tristanlabelle:
@nabramow , yes, my next PR will bake localization in `useRegions`/`useLanguages` so it returns localized labels where available. The dropdown is already handled in this PR (See the `<StyledAutocompleteOption>` part).

nabramow (approved):
LGTM pulled the branch and nothing looks different at least

app/web/features/profile/edit/EditProfile.tsx
tristanlabelle: Now tested.

PR #9173, opened 2026-06-24
Backend/i18n: Localize region & spoken language names

### LINKED ISSUE #8424: i18n: Spoken languages are not translatable

tristanlabelle:
Updates the `GetLanguages` and `GetRegions` APIs to return names localized in the current user's locale, at least where the Unicode CLDR defines those translations, which should hold for all regions and most common languages.

In both cases, our database stores 3-char codes, but the Unicode CLDR lookup is by 2-char codes. I had to add a mapping from 3-char codes to 2-char codes (where it exists) as to be able to leverage the CLDR without changing our database tables. In the case of languages, I updated `languages.json` to add 2-char codes for languages that have one.

In the end this fix is independent from #9169, but that frontend change is still an improvement.

Fixes #8424

## Testing
Ran local backend + frontend, added tests.

image

image

**Backend checklist**

## For maintainers


nabramow:
@tristanlabelle Looks like some web tests are failing!

tristanlabelle:
@nabramow , the failures were a fluke / unrelated flakiness. I rebased and tests passed. Ready for ✅ 

> FAIL features/publicTrips/PublicTripsSection.test.tsx (26.293 s)
  ● PublicTripsSection › create › submits valid form data to createPublicTrip and closes the dialog
    thrown: "Exceeded timeout of 10000 ms for a test.

app/backend/src/couchers/i18n/localize.py
```diff
@@ -30,6 +31,44 @@ def localize_string(lang: str | None, key: str, *, substitutions: Mapping[str, s
...
+        # If the part3 code (3-character) has a corresponding part1 code (2-character),
+        # the latter is what the CLDR recognizes.
+        code = get_language_codes_iso639_3_to_1().get(code, code)
+    try:
+        name = babel.Locale.parse(code).get_language_name(locale)
+        if name is None:
+            return None
+        if standalone:
+            # The result won't be embedded in a larger sentence. Capitalize the first letter if applicable.
+            name = name[:1].upper() + name[1:]
```
aapeliv: is this a universal convention? or is there a function to capitalize for standalone presentation?

aapeliv (commented):
Babel claims to support RFC3066, which claims to support also the three character codes https://datatracker.ietf.org/doc/html/rfc3066.html#section-2.2

This appears to work for me:

```
$ uv run --with babel python
Python 3.14.2 (main, Dec  5 2025, 21:11:58) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import babel
>>> babel.Locale.parse("fin").get_language_name("swe")
'finska'
>>>
```

do we then not need modification to the languages source of truth file?

tristanlabelle:
@aapeliv Thanks for catching the support for 3-letter language codes. I'll remove the extra complexity. We do need the 2-letter code for regions though.

app/backend/src/couchers/i18n/localize.py
tristanlabelle: I asked Gemini and it says pretty close but to use `.title()` instead of `.upper()` to handle weird edge cases.

str has `.capitalize()`, but it lowercases the rest of the string so we'd end up with "English (us)".

I'm comfortable to go with this (`.title()`) and react if a translator/user complains.

tristanlabelle:
@aapeliv 2-letter language codes are out. Ready for ✅

app/backend/src/couchers/i18n/localize.py
```diff
@@ -35,6 +36,43 @@ def localize_list(items: Sequence[str], locale: babel.Locale) -> str:
...
+        The localized name, or None if no localized name is available.
+    """
+    try:
+        name = babel.Locale.parse(code).get_language_name(locale)
+        if name is None:
+            return None
+        if standalone:
+            # The Unicode CLDR returns a casing that allows embedding in a larger sentence, e.g. "español".
+            # If we're displaying the language name on its own, capitalize its first letter if applicable.
+            # An LLM prompt revealed that this holds for all major languages.
```
aapeliv, line 58: thanks for leaving the comment. it appears teh CLDR data contains info on this kind of thing but babel does not expose it. i'm happy with this solution for us right now!

app/backend/src/couchers/servicers/resources.py
```diff
@@ -48,7 +48,11 @@ def GetRegions(
     ) -> resources_pb2.GetRegionsRes:
         return resources_pb2.GetRegionsRes(
             regions=[
-                resources_pb2.Region(alpha3=alpha3, name=name) for alpha3, name in sorted(get_region_dict().items())
+                resources_pb2.Region(
+                    alpha3=alpha3,
+                    name=context.localization.try_localize_region_name_from_iso3166(alpha3) or english_name,
+                )
+                for alpha3, english_name in sorted(get_region_dict().items())
```
aapeliv, line 55: Should we re-sort these in the locale of the viewer? can leave to follow up, etc. but just a thought

app/backend/src/couchers/resources.py
```diff
@@ -36,14 +37,21 @@ def get_icon(name: str) -> str:
...
     """
     with session_scope() as session:
         return {region.code: region.name for region in session.execute(select(Region)).scalars().all()}
 
 
+@functools.cache
+def get_region_code_iso3166_alpha3_to_alpha2() -> Mapping[str, str]:
+    with open(resources_folder / "regions.json", "r") as f:
+        json_regions = json.load(f)
+    return {region["alpha3"]: region["alpha2"] for region in json_regions if "alpha2" in region}
```
aapeliv, line 57: Currently there is a bit of an inconsistency issue here: `regions.json` is copied into the database "manually" in migrations whenever it is updated. This now loads it directly at runtime, so we can potentially have a mismatch with what's in the DB and what's read here.

at minimum, let's add a note here in the docstring. sorry for all the pre-existing footguns :/

aapeliv (commented):
sorry one more thing on the backend side!

app/backend/src/couchers/servicers/resources.py
tristanlabelle: Hmm, we're sorting by the ascii keys so it's language-agnostic. I'd leave sorting display strings to the frontend.

app/backend/src/couchers/i18n/localize.py
tristanlabelle: Oh yeah you're right. The CLDR capitalization stuff is fascinating but probably overkill to get perfectly right.

https://sites.google.com/unicode.org/cldr/translation/translation-guide-general/capitalization

app/backend/src/couchers/resources.py
tristanlabelle: You're right it can be confusing. I'll add a comment. I consider the mismatch is fine since we're already looking up the string in a CLDR table not necessarily in sync with DB entries. It's a best effort that should work 99% of the time and otherwise we still have the English fallback.

tristanlabelle: Added:

```py
    """
    Gets a best-effort mapping table from ISO-3166 alpha3 to alpha2 codes.
    The alpha3 keys of get_region_dict() are not guaranteed to be present in this table.
    Valid region codes, this mapping, and localized strings are all versioned separately (database / repo / CLDR).
    """
```

tristanlabelle:
@aapeliv , should be good now!

aapeliv (approved):
Lgtm!

PR #9175, opened 2026-06-24
fix: Show group avatar icon for group chats in list view

darrenvong:

Again, I don't think there is an official issue for this other than from the feedback @c-kreuz gave in Slack: https://couchersorg.slack.com/archives/C014KP1K6RJ/p1780315992620689

But before the fix, group chats were showing the picture of the last person who sent a message, which is a bit misleading since their face shouldn't be used as the display picture of the group. We should probably implement a group display picture feature separately if we want to display a group chat picture.

For now, we display a group icon as the fix.

## Testing

- Check that for group chats, a group avatar icon is now correctly displayed in the chat list view
- Check that for DMs, the other user's picture is displayed as the avatar as before in the chat list view
  - For users with no profile picture, their initials should be shown instead as before

https://github.com/user-attachments/assets/e5954f8c-b776-43c9-9755-e3c33d354d6a

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/messages/groupchats/GroupChatListItem.test.tsx
+++ b/app/web/features/messages/groupchats/GroupChatListItem.test.tsx
@@ -1,0 +1,60 @@
+import { render, screen } from "@testing-library/react";
+import GroupChatListItem from "features/messages/groupchats/GroupChatListItem";
+import { GroupChat } from "proto/conversations_pb";
+import { service } from "service";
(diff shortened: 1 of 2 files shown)


nabramow (approved):
Tested and this looks good to me! Good catch I never even realized we had this logic!

PR #9178, opened 2026-06-25
Include feature-flags.dev.json in backend Docker build context

aapeliv:
The backend `.dockerignore` ignores everything and only allow-lists specific paths. `feature-flags.dev.json` (added in 892f073c1, "Switch to growthbook feature flags") was not in the allow-list, so `COPY . /app` in the Dockerfile silently excluded it from the build context. Because `backend.dev.env` sets `FEATURE_FLAGS_FILE_OVERRIDE_PATH=feature-flags.dev.json`, the backend crashed on startup with `FileNotFoundError: .../feature-flags.dev.json` — and rebuilding with `docker compose up --build` didn't help, since the file was never in the build context to begin with.

This adds `!feature-flags.dev.json` to the allow-list so it gets copied into the image.

## Testing
- `docker compose up --build` and confirm the backend starts without the `FileNotFoundError` from `setup_experimentation()`.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/.dockerignore
+++ b/app/backend/.dockerignore
@@ -9,6 +9,7 @@
 !templates
 !alembic.ini
 !pyproject.toml
+!feature-flags.dev.json
 
 # It's not included into the final image, but is still needed here
 # for "--mount=type=bind"


nabramow (approved):
LGTM!

PR #9179, opened 2026-06-25
Backend/i18n: Make exclude_attending error string less technical.

tristanlabelle:
Reported by @c-kreuz . Those backend servicer errors should be user-readable even if the conditions are from a bug in our code.

## Testing
N/A, string only change

**Backend checklist**

## For maintainers


c-kreuz (approved):

nabramow (approved):
Oh good call thanks!

PR #9182, opened 2026-06-25
Backend/SV: Differentiate failures due to birthdate vs gender mismatch

### LINKED ISSUE #9174: Strong Verification failure message can't tell users whether birthdate or gender mismatched
### Problem

When a Strong Verification attempt fails because the passport data doesn't match the user's profile, we show a single combined message that mentions **both** date of birth **and** gender:

> Your Strong Verification attempt has failed because the date of birth or gender on your profile does not match the date of birth or sex on your passport. Please contact the support team to update your date of birth or gender, or if your passport sex does not match your gender identity.

Because this message can't tell the user *which* field actually mismatched, some users contact support confused about why their verification failed (raised in the email copy review in #9150, on `app/backend/src/couchers/email/locales/en.json`).

### Why it's a catch-all today

In `app/backend/src/couchers/jobs/handlers.py` we only check the combined predicate `has_strong_verification(user)` and, on failure, emit a single reason:

https://github.com/Couchers-org/couchers/blob/cdeff2550e168956ef78368befee9d5842203e17/app/backend/src/couchers/jobs/handlers.py#L1013-L1022

```python
else:
    notify(
        session,
        user_id=verification_attempt.user_id,
        topic_action=NotificationTopicAction.verification__sv_fail,
        key="",
        data=notification_data_pb2.VerificationSVFail(
            reason=notification_data_pb2.SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER
        ),
    )
```

`has_strong_verification` is defined as the AND of the birthdate and gender checks, so by the time we reach the `else` we've thrown away which one failed:

```python
# couchers/models/verification.py
def has_strong_verification(self, user):
    return self.is_valid & self._raw_birthdate_match(user) & self._raw_gender_match(user)
```

### Good news

The model already exposes the granular checks `matches_birthdate(user)` and `matches_gender(user)` (in `couchers/models/verification.py`), so we *can* distinguish the cause without new data:

```python
def matches_birthdate(self, user): return self.is_valid & self._raw_birthdate_match(user)
def matches_gender(self, user):    return self.is_valid & self._raw_gender_match(user)
```

### Proposed fix

- In the failure branch, evaluate `matches_birthdate` / `matches_gender` separately and pick a more specific `SV_FAIL_REASON_*` (e.g. wrong birthdate vs. wrong gender; possibly keep the combined reason only for the both-wrong case).
- Add the corresponding reason enum value(s) to the `VerificationSVFail` proto.
- Split the email/notification copy into separate, targeted messages (the email copy review noted this should be a follow-up, not part of #9150).

### Notes / open questions

- @aapeliv has the most context on the Strong Verification flow and whether there are edge cases (e.g. `passport_sex == unspecified`, the `has_passport_sex_gender_exception` escape hatch) that should still fall back to the combined message.
- This affects more than just the email — push notifications and any other surface using `SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER` would need updating too.

_Filed from a review discussion in #9150._

tristanlabelle:
Users were confused from the message saying that their passport date of birth or gender mismatched their profile, so add enumerants that differentiate those cases and update notifications.

To avoid duplicating more test code, I refactored the IRIS callback mocking logic (see changes to existing tests).

Also fixes `GetStrongVerificationAttemptStatus` returning `UNKNOWN` in the duplicate status case.

With help from [Claude Code](https://claude.ai/code)

Closes #9174.

## Testing
Added tests.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/account.py
+++ b/app/backend/src/couchers/servicers/account.py
@@ -516,6 +516,7 @@ class Account(account_pb2_grpc.AccountServicer):
             StrongVerificationAttemptStatus.in_progress_waiting_on_user_in_app: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_IN_PROGRESS_WAITING_ON_USER_IN_APP,
             StrongVerificationAttemptStatus.in_progress_waiting_on_backend: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_IN_PROGRESS_WAITING_ON_BACKEND,
             StrongVerificationAttemptStatus.failed: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_FAILED,
+            StrongVerificationAttemptStatus.duplicate: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_FAILED,
         }
         return account_pb2.GetStrongVerificationAttemptStatusRes(
             status=status_to_pb.get(
--- a/app/proto/notification_data.proto
+++ b/app/proto/notification_data.proto
@@ -188,12 +188,16 @@ message EventReminder {
 
 enum SVFailReason {
   SV_FAIL_REASON_UNKNOWN = 0;
-  // birthdate doesn't match profile or gender
+  // both birthdate and gender don't match the profile
   SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER = 1;
   // not a passport (e.g. used a ID card)
   SV_FAIL_REASON_NOT_A_PASSPORT = 2;
   // this passport is already used by someone else or the same user
   SV_FAIL_REASON_DUPLICATE = 3;
+  // only the birthdate doesn't match the profile
+  SV_FAIL_REASON_WRONG_BIRTHDATE = 4;
+  // only the gender doesn't match the profile
+  SV_FAIL_REASON_WRONG_GENDER = 5;
 }
 
 message VerificationSVFail {
--- a/app/backend/src/couchers/email/emails.py
+++ b/app/backend/src/couchers/email/emails.py
@@ -1930,6 +1930,10 @@ class StrongVerificationFailedEmail(EmailBase):
         match self.reason:
             case notification_data_pb2.SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER:
                 reason_string_key = ".reason_wrong_birthdate_or_gender"
+            case notification_data_pb2.SV_FAIL_REASON_WRONG_BIRTHDATE:
+                reason_string_key = ".reason_wrong_birthdate"
+            case notification_data_pb2.SV_FAIL_REASON_WRONG_GENDER:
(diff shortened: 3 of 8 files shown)


app/backend/src/couchers/servicers/account.py
```diff
@@ -516,6 +516,7 @@ def GetStrongVerificationAttemptStatus(
             StrongVerificationAttemptStatus.in_progress_waiting_on_user_in_app: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_IN_PROGRESS_WAITING_ON_USER_IN_APP,
             StrongVerificationAttemptStatus.in_progress_waiting_on_backend: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_IN_PROGRESS_WAITING_ON_BACKEND,
             StrongVerificationAttemptStatus.failed: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_FAILED,
+            StrongVerificationAttemptStatus.duplicate: account_pb2.STRONG_VERIFICATION_ATTEMPT_STATUS_FAILED,
```
aapeliv, line 519: what happened here?

app/proto/notification_data.proto
```diff
@@ -188,12 +188,16 @@ message EventReminder {
 
 enum SVFailReason {
   SV_FAIL_REASON_UNKNOWN = 0;
-  // birthdate doesn't match profile or gender
+  // both birthdate and gender don't match the profile
   SV_FAIL_REASON_WRONG_BIRTHDATE_OR_GENDER = 1;
```
aapeliv, line 192: then should this be renamed to AND? and also should we update the translation strings?

aapeliv (approved):
fine as-is if you have verified this works; added some comments

app/backend/src/couchers/servicers/account.py
tristanlabelle: The API was returning UNKNOWN as the status in this case. In the test I expected FAILED. It seemed like an oversight?

The semantics are not super clear as to whether it's querying that the verification attempt succeeded (they provided a legit passport), in which case this api should return SUCCESS, or that it resulted in obtaining a strong verification, in this case it makes sense to return FAILED.

app/proto/notification_data.proto
tristanlabelle: That makes sense. I was wondering if to make it into a more generic failure but handling three cases is still manageable.

tristanlabelle:
@aapeliv 

> fine as-is if you have verified this works;

Anything I can do to verify Strong Verification on top of the tests I added/updated?

aapeliv:
> @aapeliv
> 
> > fine as-is if you have verified this works;
> 
> Anything I can do to verify Strong Verification on top of the tests I added/updated?

i guess you can't on dev. once it's on stage you can test there though.

nabramow:
@tristanlabelle what's up with this one are we waiting for something?

PR #9192, opened 2026-06-27
Backend/emails: Take links out of loc files.

tristanlabelle:
Having html links in loc strings introduces risks that translators make mistakes with the syntax, and makes it more difficult to modify the URLs. It's more maintainable to interpolate the links from the code.

## Testing
Checked at generated emails on this PR:

image

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/blocks.py
+++ b/app/backend/src/couchers/email/blocks.py
@@ -13,7 +13,7 @@ from couchers import urls
 from couchers.email.locales import get_emails_i18next
 from couchers.i18n import LocalizationContext
 from couchers.i18n.i18next import SubstitutionDict, full_string_key
-from couchers.markup import html_email_link
+from couchers.markup import html_mailto_link
 from couchers.proto import api_pb2
 from couchers.utils import now
 
@@ -59,7 +59,7 @@ class EmailBase(ABC):
             # Not localizing the suggested subject line to encourage folks to use English if they can.
             builder.para(
                 "generic.security_warning_contact_support",
-                {"email_link": html_email_link("support@couchers.org", subject="Action not initiated by me")},
+                {"email_link": html_mailto_link("support@couchers.org", subject="Action not initiated by me")},
                 epilogue=True,
             )
         if default_closing:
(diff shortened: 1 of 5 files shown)


app/backend/src/couchers/email/blocks.py
```diff
@@ -57,7 +58,12 @@ def _body_builder(
         if standard_closing:
             builder.para("generic.closing_line", epilogue=True)
         if security_warning:
-            builder.para("generic.security_warning_contact_support", epilogue=True)
+            # Not localizing the suggested subject line to encourage folks to use English if they can.
+            builder.para(
+                "generic.security_warning_contact_support",
+                {"email_link": html_email_link("support@couchers.org", subject="Action not initiated by me")},
```
tristanlabelle: @jesseallhands , do we want to localize the support email subject line? Or use it as a hint for people to write in English if they can?

app/backend/src/couchers/email/emails.py
```diff
@@ -1734,17 +1734,29 @@ def get_body_blocks(self, loc_context: LocalizationContext) -> list[EmailBlock]:
...
             builder.para(".share_with_friends")
-            builder.para(".link", {"url": urls.app_link()})
+            builder.para(".link", {"link": html_link(urls.app_link())})
             builder.para(".platform_under_development")
             builder.para(".thanks_for_joining")
-            builder.para(".signature")
+            builder.para(
+                ".signature",
+                {
+                    "email_link": html_email_link("aapeliv@couchers.org"),
```
aapeliv: no `v` here

aapeliv (changes requested):
lgtm otherwise

app/backend/src/couchers/email/blocks.py
jesseallhands: No we don't need to. It's just a hint for support

app/backend/src/couchers/email/emails.py
tristanlabelle: Fixed. Waiting on your approval

tristanlabelle:
Fixed email address and rebased. Merging based on "lgtm otherwise"

Issue #9172, opened 2026-06-24
Backend/i18n: Language list contains invalid and deprecated ISO639-3 codes


tristanlabelle:
Specifically, `_tw` is invented (`nan` is official), `smi` is a part2 code (group of languages), and `ajp` is deprecated.

aapeliv:
Solution is not to remove them. `_tw` is there in addition to `nan` on purpsoe

tristanlabelle:
Further thinking / context: https://github.com/Couchers-org/couchers/pull/9204#issuecomment-4843968078

PR #9204, opened 2026-06-29
Backend: replace deprecated ISO639-3 code ajp with apc

kevinortiz43:
Replaces the deprecated ISO 639-3 code `ajp` ("Arabic (Levantine South)") with `apc` ("Arabic (Levantine North)") in the allowed-languages list, and migrates existing user language abilities off it.

`ajp` was retired in ISO 639-3 Change Request 2022-006 and merged into `apc` (which already exists in the list). Migration `0170` reloads the static `languages` table and remaps `language_abilities` `ajp`→`apc`, with a dedupe `DELETE` before the rename so a user already holding `apc` doesn't violate `UniqueConstraint(user_id, language_code)` (that unique constraint is not deferrable). `downgrade()` is intentionally a no-op (irreversible — matches migration `0027`).

### Scope note
Issue #9172 also flagged `_tw` and `smi`. After fact-checking against current ISO 639:
- **`smi`** is **not** deprecated — it's a valid ISO 639-2/B collective code for the Sami languages (just not an individual 639-3 code). **Kept.**
- **`_tw`** is non-standard but was an intentional entry for "Taiwanese". **Kept.**

So this PR is scoped to the one genuinely-deprecated code. The broader question of whether the list should be strict ISO 639-3 vs a curated practical list is left to the discussion in #9172.

Refs #9172

## Testing

- `make format` (ruff) + `make mypy` clean on the changed files.
- `pytest src/tests/test_resources.py src/tests/test_dummy_data.py` pass.
- `test_languages_resource_drops_deprecated_ajp` loads the real `languages.json` via `copy_resources_to_database` and asserts `ajp` is absent while `apc`, `_tw`, and `smi` are present.
- Migration `0170` verified against real Postgres: seeded `language_abilities` with `ajp` (incl. a user already holding `apc`); after migration `ajp`→`apc`, the duplicate collapses with no unique-constraint violation, and `_tw`/`smi` rows are untouched. (CI's `test:migrations` runs the full alembic apply.)

**Backend checklist**

## For maintainers


aapeliv (dismissed):
let's not do this. these were all added by request (except maybe the Arabic one) and for a reason.

also sami is indeed a langauge with this ISO 639 code: i'm not sure why you are saying it's not?

kevinortiz43:
> let's not do this. these were all added by request (except maybe the Arabic one) and for a reason.
> 
> also sami is indeed a langauge with this ISO 639 code: i'm not sure why you are saying it's not?

Just following this bug issue it said Sami and the other languages were deprecated. https://github.com/Couchers-org/couchers/issues/9172 

Okay so you only want the Arabic one to be updated? @aapeliv

kevinortiz43:
@aapeliv I double checked you were right, I only updated the arabic iso code I did not touch the other iso codes

app/backend/src/tests/test_resources.py
```diff
@@ -57,6 +61,18 @@ def test_GetLanguages(db):
...
+def test_languages_resource_drops_deprecated_ajp(db):
+    # Load the real languages.json (not the hardcoded testing fixture) so a future bad code is caught.
+    # Mirrors test_add_dummy_data's pattern of calling copy_resources_to_database directly.
+    with session_scope() as session:
+        copy_resources_to_database(session)
+        codes = set(session.execute(select(Language.code)).scalars().all())
+    assert "ajp" not in codes  # deprecated; ISO 639-3 CR 2022-006 merged it into apc
+    assert "apc" in codes
+    assert "_tw" in codes  # kept intentionally (non-standard but deliberate)
+    assert "smi" in codes  # kept: valid ISO 639-2 collective code, not deprecated
```
tristanlabelle: Those two are not super related to the current test as currently named.

tristanlabelle (dismissed):
The current form of the PR seems without risk.

tristanlabelle:
I filed the bug when I found that `_tw` was not a real `iso639` code, so that put in doubt the whole file: what's `iso639` and what is custom? Going through the git history, I couldn't find why `_tw` was added. My concern is that we document the data to be `iso639`, and I now have code that consumes it as `iso639` for localization, but there is an invisible outlier.

Without reconsidering `_tw`, I would like to at least split `languages.json` into `languages-iso639-3.json` and `languages-custom.json` to make things obvious if we add languages in the future. Thinking about doing that is when I found that Sami is iso639-2 and not iso639-3. It's unclear why, and currently we have no way to track exceptions and their reasons.

app/backend/src/tests/test_resources.py
kevinortiz43: I removed these tests

tristanlabelle (approved):

PR #9209, opened 2026-06-30
Backend/emails: v2 copy for onboarding email

tristanlabelle:
As per discussions with @aapeliv .

Minor tweaks to prose:
- uploading a profile picture -> uploading a photo (we use the term photo on the profile)
- Our best members are those ~that~who were invited by someone else.

I also more clearly split the code for the first and second onboarding emails since they're independent.

## Testing
`uv run dump-emails`

image

**Backend checklist**

## For maintainers


aapeliv (approved):
copy looks good to me

PR #9211, opened 2026-06-30
Frontend/i18n: Remove inline <Trans> text

### LINKED ISSUE #9008: Frontend/i18n: Investigate removing duplicated English placeholders
Sometimes when we use `<1>foo</1>` we also duplicate the "foo" text in the code which seems unnecessary.

We should also be able to simplify nested placeholders like `<1>{{email}}</1>` to simply `<email />` or so.

tristanlabelle:
Our `<Trans>` tags often included a copy of the string from `en.json`, which unavoidably lead to the two drifting apart as noted in #9008. It turns out we don't have to do that and can simply provide the components to be instantiated by the string without duplicating the text.

Closes #9008

## Testing
Used vercel preview to spot check:

`ProfileIncompleteDialog`

image

`StrongVerificationPage` (note the text is from the more recent `en.json`, the inline text was stale)

image

`PushNotificationSetting`

image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/Navigation/ReportDialog.tsx
+++ b/app/web/components/Navigation/ReportDialog.tsx
@@ -111,7 +111,7 @@ export default function ReportDialog({ open, onClose }: DialogProps) {
                 </StyledLink>
               ),
             }}
-          ></Trans>
+          />
         </Snackbar>
       )}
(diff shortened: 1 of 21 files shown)


nabramow (approved):
Oh great, I always thought this seemed a bit weird! I tested a couple and seems okay to me!

Issue #7998, opened 2026-03-01
Frontend/i18n: Show event timezone


tristanlabelle:
Event times are in local times, but this may not be obvious, so it would help to show the timezone.

tristanlabelle:
Decision from today's meeting: Don't show the timezone.

PR #9218, opened 2026-07-01
Backend+Frontend/Events: Determine events timestamps in the event's timezone

### LINKED ISSUE #8064: Event timezone information is lost, can't display in local time
When creating an event, we convert the start date+time and end date+time that the user enters from their browser timezone to UTC, then save it without specifying a timezone, so the backend defaults to UTC. This means that we lose information about what timezone it was created in and can't show the event in its local timezone (which is probably what makes the most sense).

There is also the issue that my browser's timezone isn't necessarily that of the event I'm creating: I can create an event in a place I'll be traveling to.

I think the correct way to do this will be to add a timezone field to the event editing form, defaulting to the browser's timezone or deducing it from the location, and then have the frontend persist it to the backend.

tristanlabelle:
Details in #8064. Event creation only resulted in valid UTC timestamps if it the event location was in the same timezone as the user's browser. To fix this, the frontend must send the datetime components to the backend, which can convert them into a timestamp only after figuring out the correct timezone from the location.

The backend now stores timezones for events (based on the location) and will include them in responses to the frontend. I backfilled the event timezones by deducing one from their coordinates.

We have no use case for the frontend telling the backend the timezone of the event (not basing it on location), and weren't using that code path, so I removed it from the protos.

When updating an event's location to a new timezone, existing start/end local datetimes and reinterpreted into the new timezone, as if the user had modified and submitted them. So if the event was at midnight in Helsinki, it will still be at midnight after moving to New York, event though that is a different timestamp.

Editing an event now depends on its timezone so I changed test data to use timezones that don't have DST transitions.

After merging this, an existing bug will become more obvious: After creating an event in a foreign timezone from my browser, the time shown for the event will mismatch what I entered. Previously it would appear the same to me, but would be off for anyone in the timezone where the event was happening, since we have a bad datetime in the db. I'll improve this slightly with #7998, but there's nothing to do about the bad data for existing events already in the db.

Includes small fixes: having a single zero coordinate shouldn't error. Event times should be at minute granularity.

Fixes #8064, #9230

## Testing
Ran backend + frontend and created, then updated an event. Also added a test.

Event created in New York (`timezone_areas.sql-fake` has Americas/New_York), while my browser timezone is Montreal (the same):

image
image

Event created in Helsinki (`timezone_areas.sql-fake` has Europe/Helsinki), while my browser timezone is Montreal. The discrepancy is expected and explained above, it means that the event was successfully created at midnight in the Helsinki timezone, and is then shown in my browser timezone.

image
image

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/migrations/versions/0172_add_event_timezone_column.py
```diff
@@ -0,0 +1,24 @@
...
+revision = "0170"
+down_revision = "0169"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("event_occurrences", sa.Column("timezone", sa.String(), nullable=True))
+
+
```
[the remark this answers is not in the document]
tristanlabelle: Switched to non-nullable column with backfilling from geom, and otherwise defaulting to Etc/UTC.

app/backend/src/tests/test_events.py
```diff
@@ -2459,6 +2432,64 @@ def test_event_photo_key(db):
...
+                location=events_pb2.EventLocation(address="New York", lat=40.712776, lng=-74.005974),
+            )
+        )
+
+        # The time of the event should not have changed, so it's still at midnight in Helsinki, not in New York.
+        assert update_res.timezone == "America/New_York"
+        assert update_res.start_time == create_res.start_time
+        assert to_aware_datetime(update_res.start_time).hour != 0
+        assert update_res.start_time.ToDatetime(tzinfo=ZoneInfo("Europe/Helsinki")).hour == 0
+        assert update_res.start_time.ToDatetime(tzinfo=ZoneInfo("America/New_York")).hour != 0
```
tristanlabelle: @aapeliv curious for your thoughts on semantics here. When creating an event, we specify the components e.g. 1970-01-01T00:00, then it's stored as a timestamp on the server.

When updating an event's location, which implicitly updates the timezone, I currently keep the timestamp (so it won't be midnight in the new location). Alternatively we could reinterpret "1970-01-01T00:00" into the new timezone event if it wasn't provided again in the update request.

I think it's ok to have an understanding that the backend stores timestamps and only the create/updatereq deal with datetime components to be timezone-agnostic.

app/backend/src/couchers/migrations/versions/0172_add_event_timezone_column.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        UPDATE event_occurrences
+        SET timezone = (
+            SELECT tzid
+            FROM timezone_areas
+            WHERE ST_Contains(timezone_areas.geom, event_occurrences.geom)
+            LIMIT 1
+        )
+        WHERE event_occurrences.geom IS NOT NULL
+        """
```
WouldYouKindly, line 32: This will lock the table for the duration of the update. I assume geom queries are kind of slow. Should we do a batched update here?

(technically, it will lock the rows as they are updated)

app/backend/src/couchers/servicers/events.py
```diff
@@ -228,6 +230,38 @@ def _get_event_and_occurrence_one_or_none(
     return result._tuple() if result else None
 
 
+def _check_location(location: events_pb2.EventLocation | None, context: CouchersContext) -> tuple[WKBElement, str]:
+    # As protobuf parses a missing value as 0.0, this is not a permitted event coordinate value
+    if not (location and location.address and location.lat and location.lng):
+        context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "missing_event_address_or_location")
+    if location.lat == 0 and location.lng == 0:
```
WouldYouKindly, line 230: Check for negative values to be pedantic?

app/backend/src/couchers/servicers/events.py
```diff
@@ -228,6 +230,38 @@ def _get_event_and_occurrence_one_or_none(
...
+def _check_timezone_at(geom: WKBElement, context: CouchersContext, session: Session) -> ZoneInfo:
+    timezone_id = session.execute(
+        select(TimezoneArea.tzid).where(func.ST_Contains(TimezoneArea.geom, func.ST_PointOnSurface(geom))).limit(1)
+    ).scalar_one_or_none()
+    if not timezone_id:
+        context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "event_timezone_not_found")
+
+    try:
+        return ZoneInfo(timezone_id)
+    except ZoneInfoNotFoundError:
```
WouldYouKindly: This means we have an invalid timezone in our database. I think we should treat it as a bug?

app/backend/src/couchers/migrations/versions/0172_add_event_timezone_column.py
tristanlabelle: I'm a newbe here, how would do I do a batched update?

app/backend/src/tests/test_events.py
tristanlabelle: I decided to change the behavior. If an UpdateEvent call specifies a new location with a new timezone without specifying new times, I now reinterpret the local datetime from the old to the new timezone. This makes sense because the user will have seen the datetime pickers and not have touched them, so they expect the same local time in the new location.

app/backend/src/couchers/servicers/events.py
tristanlabelle: Negative values are legit for both of these. Actually the test above is broken though, one of the coords could be zero, but not both. I'll fix that.

app/backend/src/couchers/servicers/events.py
tristanlabelle: Sure. How would you do it? Log to sentry and return UTC, or let the exception propagate?

WouldYouKindly: I would just let it propagate tbh

app/backend/src/couchers/migrations/versions/0172_add_event_timezone_column.py
aapeliv: this is fine. the geometries are divided into small squares, so there is a fast index lookup. this migration will takes at most seconds

app/backend/src/couchers/utils.py
```diff
@@ -130,6 +130,20 @@ def to_timezone(value: Timestamp | datetime, timezone: tzinfo) -> datetime:
     return value.astimezone(timezone)
 
 
+def is_utc_or_gmt(timezone: str) -> bool:
+    # There are various interchangeable names for the UTC+0 timezone.
+    # Null island is at Etc/UTC with dummy timezone data used for local testing, but Etc/GMT with real timezone data.
+    return timezone in ("Etc/UTC", "UTC", "Etc/GMT", "GMT")
```
aapeliv: i don't see this used outside tests; it's also quite fragile; could we move it somewhere in test code where fragility is less of an issue?

aapeliv:
I had a quick look, added minor comments. Overall looks appropriate. I know in other parts of the code we leave the TZ lookup to be defined at query time: do events warrant different treatment? Consistency is good if there is no reason to change one or the other!

app/backend/src/couchers/utils.py
tristanlabelle: Good call, I'll move it in the tests.

tristanlabelle:
> I had a quick look, added minor comments. Overall looks appropriate. I know in other parts of the code we leave the TZ lookup to be defined at query time: do events warrant different treatment? Consistency is good if there is no reason to change one or the other!

@aapeliv Good convo. I started out implemented as a lazy query but a few things complicate it:

1. We still have some events without a geom (legacy online events)
2. When creating an event, I need the timezone to determine the timestamp. At that time I don't have a DB object yet. Similar problem with updates.
3. If the event timezone change, its timestamps also need to be updated. If timezones are resolved lazily and we update our timezone regions data, an event could move between timezones without a chance for the code to be involved.

To me that summed up to a good reason for storing the timezone.

(also as a hypothetical 4., it allows us to have the frontend dictate the timezone in the future if needed)

app/backend/src/tests/test_events.py
```diff
@@ -2383,7 +2367,9 @@ def test_update_event_should_notify_queues_job():
     api.UpdateEvent(
         events_pb2.UpdateEventReq(
             event_id=event_id,
-            start_time=Timestamp_from_datetime(start + timedelta(hours=4)),
+            start_datetime_iso8601_local=wrappers_pb2.StringValue(
+                value=datetime_to_iso8601_local(start + timedelta(hours=5))
```
tristanlabelle, line 2338: @aapeliv I updated hours from 4 to 5 here because my new code won't cause a notification if the new time is the same as the old time. Should we do this more generally?

app/backend/src/couchers/servicers/events.py
tristanlabelle: Updated in latest to let propagate. I also tested with all values we insert in the DB, they're all supported by ZoneInfo.

tristanlabelle:
@aapeliv / @WouldYouKindly Comments addressed. Ready for review!

tristanlabelle:
Rebased on top of #9236, which makes the frontend part trivial.

aapeliv:
> > I had a quick look, added minor comments. Overall looks appropriate. I know in other parts of the code we leave the TZ lookup to be defined at query time: do events warrant different treatment? Consistency is good if there is no reason to change one or the other!
> 
> @aapeliv Good convo. I started out implemented as a lazy query but a few things complicate it:
> 
>     1. We still have some events without a geom (legacy online events)
> 
>     2. When creating an event, I need the timezone to determine the timestamp. At that time I don't have a DB object yet. Similar problem with updates.
> 
>     3. If the event timezone change, its timestamps also need to be updated. If timezones are resolved lazily and we update our timezone regions data, an event could move between timezones without a chance for the code to be involved.
> 
> 
> To me that summed up to a good reason for storing the timezone.
> 
> (also as a hypothetical 4., it allows us to have the frontend dictate the timezone in the future if needed)

Yeah that makes sense, it sounds like this implementation is the best approach here. Thanks!

PR #9219, opened 2026-07-01
Backend: Create new message thread queries and add public trips

### LINKED ISSUE #8962: Improve backend sorting of messages for current UI
We're doing a lot of funky frontend logic to make the messages filtering work. We should move this logic to the backend so it's cleaner.

Take a look at the messages page to see what we need.

In my head it should be an rpc with various params for the different filters we need

- ListAllMessagesAndRequests queries chats, group chats, and host requests and returns based on last active/most recently message and unread at the top
- Takes a param for type so the hosting, surfing can be separated
- Has a param for archived for the archived filter
- Has a param for unread only
- Ability to choose what to sort by, so on dashboard we can choose only to show host_requests or surf_requests by ascending order `from_date`
- Notifications for the messages, missed_reminders for messages, etc should be marked seen when the user views the message

nabramow:
Closes #8962

- Adds new message thread query that better fits the new UI. Also adds public trips to it.

FYI - I have the frontend in another branch off this one `na/web/messages-pub-trips` and tested this there.

## Testing

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/api.py
```diff
@@ -277,6 +322,9 @@ def Ping(self, request: api_pb2.PingReq, context: CouchersContext, session: Sess
             unseen_message_count=unseen_message_count,
             unseen_sent_host_request_count=unseen_sent_host_request_count,
             unseen_received_host_request_count=unseen_received_host_request_count,
+            unseen_hosting_host_request_count=unseen_hosting_host_request_count,
+            unseen_surfing_host_request_count=unseen_surfing_host_request_count,
+            unseen_public_trip_offer_count=unseen_public_trip_offer_count,
```
aapeliv, line 297: `unseen_sent_host_request_count` should equal `unseen_surfing_host_request_count` pre-public trips, right? is there a reason to have them separate?

app/backend/src/couchers/servicers/api.py
```diff
@@ -229,6 +229,51 @@ def Ping(self, request: api_pb2.PingReq, context: CouchersContext, session: Sess
             )
         ).scalar_one()
 
+        # Role-based unread host-request counts. Unlike sent/received above, these
+        # classify by stay-role, so public-trip offers (role reversal) bucket
+        # correctly: the offering host counts under hosting, the traveller under surfing.
+        def role_based_unseen_host_request_count(party_predicate: ColumnElement[bool]) -> int:
```
aapeliv: i am concerned about the performance of the Ping endpoint, and this will add a good chunk of extra work. we should optimize these queries to instead subquery all requests where the given user is a party and that have unread messages, then use that as a subquery to count by type.

i can do it as a follow up, so doesn't need to block this.

app/backend/src/couchers/servicers/public_trips.py
```diff
@@ -82,6 +82,23 @@ def public_trip_to_pb(
...
+        # Per-status tally of offers, for the trip owner's grouped view.
+        counts_by_status: dict[HostRequestStatus, int] = dict(
+            session.execute(  # type: ignore[arg-type]
+                select(HostRequest.status, func.count())
+                .where(HostRequest.public_trip_id == public_trip.id)
+                .where(HostRequest.status != HostRequestStatus.cancelled)
+                .group_by(HostRequest.status)
+            ).all()
+        )
+        pb.offer_tally.CopyFrom(
```
aapeliv: i'm not sure you need CopyFrom here, i think you can just assign?

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
   bool no_more = 3;
 }
 
+enum MessageThreadFilter {
+  MESSAGE_THREAD_FILTER_ALL = 0;
```
aapeliv: needs an unspecified

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
...
+  HostRequestStatus status = 4;
+  google.protobuf.Timestamp created = 5;
+  // dates as "yyyy-mm-dd"
+  string from_date = 6;
+  string to_date = 7;
+  uint64 last_seen_message_id = 8;
+  Message latest_message = 9;
+  string hosting_city = 10;
+  bool is_archived = 11;
+  // true if this thread is a public-trip offer (public_trip_id is set)
```
aapeliv: also requires current user to be recipient?

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
   bool no_more = 3;
 }
 
+enum MessageThreadFilter {
+  MESSAGE_THREAD_FILTER_ALL = 0;
+  MESSAGE_THREAD_FILTER_UNREAD = 1;
+  MESSAGE_THREAD_FILTER_CHATS = 2; // group chats + DMs only
+  MESSAGE_THREAD_FILTER_HOSTING = 3; // role-based: I am the host of the stay
+  MESSAGE_THREAD_FILTER_SURFING = 4; // role-based: I am the guest/traveller
+  MESSAGE_THREAD_FILTER_PUBLIC_TRIPS = 5; // my incoming public-trip offers (I'm the traveller)
```
aapeliv: why do public trips get separate logic? why doesn't it have filter for host requests?

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
...
+    GroupChat group_chat = 1;
+    HostRequestThread host_request = 2;
+  }
+}
+
+message ListMessageThreadsReq {
+  MessageThreadFilter filter = 1;
+  // orthogonal: when present, restrict to archived (true) or non-archived (false)
+  optional bool only_archived = 2;
+  uint32 number = 3;
```
aapeliv: is this page_size?

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
...
+  // opaque cursor; the Message.id high-water mark
```
aapeliv: it's not opaque if you say what it is :)

app/proto/public_trips.proto
```diff
@@ -67,6 +67,15 @@ message PublicTrip {
...
   // state and link to the existing thread. Not set for the trip owner.
   int64 viewer_host_request_id = 13;
+  // Per-status breakdown of the offers on this trip. Only populated when the
+  // viewer is the trip owner.
+  optional PublicTripOfferTally offer_tally = 14;
+}
+
+message PublicTripOfferTally {
+  int32 pending = 1;
+  int32 accepted = 2; // accepted + confirmed
```
aapeliv: why combine the two together?

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +719,179 @@ def ListGroupChats(
             no_more=len(results) <= page_size,
         )
 
+    def ListMessageThreads(
```
aapeliv, line 763: i will need to look at the changes in this file when it's not past midnight :)

aapeliv (commented):
this is quite a huge PR. i added some initial comments but didn't have the chance to go through the entirety of the new logic in detail yet.

nabramow:
> this is quite a huge PR. i added some initial comments but didn't have the chance to go through the entirety of the new logic in detail yet.

Yeah I wasn't sure how else to break it up as it's basically one big query!

app/backend/src/couchers/servicers/api.py
nabramow: You're right `unseen_sent_host_request_count` should `unseen_surfing_host_request_count` compute the same number today as right now every request's surfer is also its initiator. 

However the code keeps them separate anticipating public trips, where that will stop being true.

I think once public trips is out we could migrate the clients off `sent/received` onto `surfing/hosting`?

app/backend/src/couchers/servicers/api.py
nabramow: makes sense!

app/backend/src/couchers/servicers/public_trips.py
nabramow: yeah looks like I can as long as it's not the whole message

app/proto/conversations.proto
nabramow: Hmm but there is no unspecified in practice or you mean just use ALL as unspecified?

nabramow: I will add it but have it be an invalid argument I think

app/proto/conversations.proto
nabramow: Good catch, this made me realize that the flag is redundant anyway. 

`is_public_trip_offer` just duplicated `public_trip_id`'s presence, and `viewer_is_host` is derivable from `host_user_id == myId`. Since roles are already sent viewer-independently, I dropped both. 

Now we can tell : 
- is it an offer? → `public_trip_id` present
- am I the host? → `host_user_id == me`
- is it an offer to me? → `public_trip_id` present and `host_user_id != me`.

app/proto/conversations.proto
nabramow: The filters mirror the UI tabs, and there's no single "host requests" tab. Requests are shown split by role (Hosting vs Surfing), not as one combined list.

`PUBLIC_TRIPS` is broken out because there will be a dedicated "offers to host me" tab that needs its own paginated list in the UI. The main bug now is pagination doesn't really work as we're hacking it together on the frontend so trying to fix that in this PR.

app/proto/conversations.proto
nabramow: Oh good call yes it's `page_size` I will rename it.

app/backend/src/couchers/servicers/conversations.py
nabramow: It should only be additive if that helps! The idea is merge backend, connect frontend, then remove the old one in three separate steps.

app/proto/conversations.proto
nabramow: ahah fair, I re-wrote the comment to make it more clear in any case

app/proto/public_trips.proto
nabramow: Ah good point I guess better if I handle that on the client side so we have the option to split it later.

app/backend/src/couchers/servicers/requests.py
```diff
@@ -405,6 +405,7 @@ def GetHostRequest(
 
         return host_request_to_pb(host_request, session, context)
 
+    # TODO(remove after FE migrates to ListMessageThreads)
```
tristanlabelle: Can you have this TODO refer to a GitHub issue? (`# TODO(#1234): blah blah blah`)

app/proto/public_trips.proto
tristanlabelle, line 63: Should we deprecate in favor of `offer_tally`?

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
...
+  MESSAGE_THREAD_FILTER_CHATS = 3; // group chats + DMs only
+  MESSAGE_THREAD_FILTER_HOSTING = 4; // role-based: I am the host of the stay
+  MESSAGE_THREAD_FILTER_SURFING = 5; // role-based: I am the guest/traveller
+  MESSAGE_THREAD_FILTER_PUBLIC_TRIPS = 6; // my incoming public-trip offers (I'm the traveller)
+}
+
+// A host request (or public-trip offer) shaped for the unified thread list.
+// surfer_user_id/host_user_id are ROLE-based: for a public-trip offer (role
+// reversal) they are swapped relative to the underlying initiator/recipient.
+message HostRequestThread {
```
tristanlabelle: This is duplicating HostRequest's every field, with subtle changes. We should be referencing it instead. For example now I'd need to add a timezone in two places.

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
...
+  optional bool only_archived = 2;
+  uint32 page_size = 3;
+  // pagination cursor; pass back the next_page_token from the previous response, or empty for the first page
+  string page_token = 4;
+}
+
+message ListMessageThreadsRes {
+  repeated MessageThread threads = 1;
+  string next_page_token = 2;
+  bool no_more = 3;
```
tristanlabelle: Isn't `no_more` redundant with `next_page_token` being empty? I think it'd be clearer to represent it that way.

app/proto/conversations.proto
```diff
@@ -185,6 +197,67 @@ message ListGroupChatsRes {
   bool no_more = 3;
 }
 
+enum MessageThreadFilter {
```
tristanlabelle: Even if the UI has those mutually exclusive filters, would it make sense for the backend to model those as combinable? E.g. support specifying searching for Hosting + Surfing?

app/backend/src/tests/test_message_threads.py
```diff
@@ -0,0 +1,409 @@
...
+from tests.fixtures.db import generate_user
+from tests.fixtures.sessions import conversations_session, real_api_session, requests_session
+
+
+@pytest.fixture(autouse=True)
+def _(testconfig):
+    pass
+
+
+HOST_REQUEST_TEXT = "a" * 300
```
tristanlabelle: You can avoid the magic number by using `valid_request_text` from `test_requests.py` (consider moving it elsewhere for reuse)

app/backend/src/tests/test_message_threads.py
```diff
@@ -0,0 +1,409 @@
...
+                small_community_features_enabled=True,
+            )
+        )
+        session.flush()
+        trip = PublicTrip(
+            user_id=user_id,
+            node_id=node.id,
+            from_date=today() + timedelta(days=5),
+            to_date=today() + timedelta(days=10),
+            description="x" * 200,
```
tristanlabelle: Avoid the 200 magic number, can we reference a constant?

app/backend/src/tests/test_message_threads.py
```diff
@@ -0,0 +1,409 @@
...
+        session.add(
+            Cluster(
+                name="Test community",
+                description="Test",
+                parent_node_id=node.id,
+                is_official_cluster=True,
+                small_community_features_enabled=True,
+            )
+        )
+        session.flush()
```
tristanlabelle: Let's (extract and) reuse `create_community` from `test_communities.py`.

app/backend/src/tests/test_message_threads.py
```diff
@@ -0,0 +1,409 @@
...
+        trip = PublicTrip(
```
tristanlabelle: Would be great to reuse `_create_trip_directly` and just have one place creating DB entries if we can't use the API.

app/backend/src/tests/test_message_threads.py
```diff
@@ -0,0 +1,409 @@
...
+
+def test_list_message_threads_single_cursor_pagination_across_kinds(db, moderator):
+    user1, token1 = generate_user()
+    others = [generate_user() for _ in range(6)]
+
+    expected_ids = set()
+    # interleave creating group chats and host requests so both kinds straddle page boundaries
+    for i, (other, other_token) in enumerate(others):
+        expected_ids.add(_create_group_chat(token1, [other.id], moderator))
+        expected_ids.add(_create_host_request(other_token, user1.id, moderator))
```
tristanlabelle: Are group chat and host request IDs necessarily non-overlapping? I think we should track them separately.

app/backend/src/couchers/servicers/api.py
```diff
@@ -229,6 +229,51 @@ def Ping(self, request: api_pb2.PingReq, context: CouchersContext, session: Sess
...
+                and_(HostRequest.public_trip_id.is_(None), HostRequest.recipient_user_id == me),
+                and_(HostRequest.public_trip_id.isnot(None), HostRequest.initiator_user_id == me),
+            )
+        )
+        unseen_surfing_host_request_count = role_based_unseen_host_request_count(
+            or_(
+                and_(HostRequest.public_trip_id.is_(None), HostRequest.initiator_user_id == me),
+                and_(HostRequest.public_trip_id.isnot(None), HostRequest.recipient_user_id == me),
+            )
+        )
```
tristanlabelle: I think we could make this a single database query for entries where either the initiator or recipient is "me", then do the different types of counting in Python. Maybe we can even cover the 5 counts that way in a single query.

tristanlabelle (commented):
I still need to go through `conversations.py` due to its size but so far I left comments about code duplication and a few API shape thoughts.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+
+    candidate_query = select(
+        visible_group_chats.c.conversation_id.label("conversation_id"),
+        visible_group_chats.c.latest_message_id.label("latest_message_id"),
+        literal("group_chat").label("kind"),
+    )
+    if unread:
+        # restrict to chats with at least one message newer than the user's last-seen
+        candidate_query = candidate_query.join(
+            GroupChatSubscription, GroupChatSubscription.id == visible_group_chats.c.subscription_id
```
aapeliv: you are double-joining here on GroupChatSubscription; you can move this above where you wrap in the `select`, so then you don't need this join. similar to only_archived

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+            and_(HostRequest.public_trip_id.is_(None), HostRequest.recipient_user_id == viewer_id),
+            and_(HostRequest.public_trip_id.isnot(None), HostRequest.initiator_user_id == viewer_id),
+        )
+    if thread_filter == conversations_pb2.MESSAGE_THREAD_FILTER_SURFING:
+        # I am the guest of the stay: normal request I sent, or an offer I received
+        return or_(
+            and_(HostRequest.public_trip_id.is_(None), HostRequest.initiator_user_id == viewer_id),
+            and_(HostRequest.public_trip_id.isnot(None), HostRequest.recipient_user_id == viewer_id),
+        )
+    if thread_filter == conversations_pb2.MESSAGE_THREAD_FILTER_PUBLIC_TRIPS:
```
aapeliv: let's call this MY_PUBLIC_TRIPS, or SURFING_PUBLIC_TRIPS; it is confusing otherwise

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+        mute_info=_mute_info(subscription),
+        can_message=_user_can_message(session, context, group_chat),
+        is_archived=subscription.is_archived,
+    )
+
+
+def _host_request_viewer_last_seen(host_request: HostRequest, user_id: int) -> int:
+    """The viewer's role-specific last-seen message id for a host request."""
+    if host_request.initiator_user_id == user_id:
+        return host_request.initiator_last_seen_message_id
```
aapeliv: please add a similar if for recipient_user_id, and throw if the user_id matches neither; that's safer

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+        only_admins_invite=group_chat.only_admins_invite,
+        is_dm=group_chat.is_dm,
+        created=Timestamp_from_datetime(group_chat.conversation.created),
+        unseen_message_count=unseen_message_count,
+        last_seen_message_id=subscription.last_seen_message_id,
+        latest_message=_message_to_pb(message) if message else None,
+        mute_info=_mute_info(subscription),
+        can_message=_user_can_message(session, context, group_chat),
+        is_archived=subscription.is_archived,
+    )
```
aapeliv: We could take this opportunity to prune stuff that we don't use in the frontend; e.g. the visible members/admins is N queries per group chat.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+            func.max(Message.id).label("message_id"),
+        )
+        .join(Message, Message.conversation_id == GroupChatSubscription.group_chat_id)
+        .where(GroupChatSubscription.user_id == context.user_id)
+        .where(GroupChatSubscription.group_chat_id.in_(group_chat_ids))
+        .where(Message.time >= GroupChatSubscription.joined)
+        .where(or_(Message.time <= GroupChatSubscription.left, GroupChatSubscription.left == None))
+        .group_by(GroupChatSubscription.group_chat_id)
+        .subquery()
+    )
```
aapeliv: I think this logic is duplicated now multiple times? In existing code as well. Given the complexity of this query, we should only construct it once. (Similar with group chats)

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -340,7 +361,284 @@ def _mute_info(subscription: GroupChatSubscription) -> conversations_pb2.MuteInf
...
+    subscription_ids = [row.GroupChatSubscription.id for row in group_chat_rows]
+    unseen_count_by_subscription: dict[int, int] = dict(
+        session.execute(  # type: ignore[arg-type]
+            select(GroupChatSubscription.id, func.count(Message.id))
+            .join(Message, Message.conversation_id == GroupChatSubscription.group_chat_id)
+            .where(GroupChatSubscription.id.in_(subscription_ids))
+            .where(Message.id > GroupChatSubscription.last_seen_message_id)
+            .group_by(GroupChatSubscription.id)
+        ).all()
+    )
```
aapeliv: Do this in the same SQL query so we don't need to roundtrip back to python in between

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +716,185 @@ def ListGroupChats(
...
+
+        if thread_filter == conversations_pb2.MESSAGE_THREAD_FILTER_PUBLIC_TRIPS and not context.get_boolean_value(
+            "public_trips_enabled", False
+        ):
+            return conversations_pb2.ListMessageThreadsRes(threads=[], no_more=True)
+
+        page_size = request.page_size if request.page_size != 0 else DEFAULT_PAGINATION_LENGTH
+        page_size = min(page_size, MAX_PAGE_SIZE)
+        only_archived = request.only_archived if request.HasField("only_archived") else None
+        unread = thread_filter == conversations_pb2.MESSAGE_THREAD_FILTER_UNREAD
```
aapeliv: only_unread?

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +716,185 @@ def ListGroupChats(
...
+            combined_candidates.c.kind,
+        )
+        if request.page_token:
+            page_query = page_query.where(combined_candidates.c.latest_message_id < int(request.page_token))
+        page_query = page_query.order_by(combined_candidates.c.latest_message_id.desc()).limit(page_size + 1)
+        candidate_rows = session.execute(page_query).all()
+
+        page_rows = candidate_rows[:page_size]
+        no_more = len(candidate_rows) <= page_size
+        next_page_token = str(min(row.latest_message_id for row in page_rows)) if (page_rows and not no_more) else ""
```
aapeliv: could we encrypt the page token please? there's a helper function. that way it's truly opaque

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +716,185 @@ def ListGroupChats(
...
+
+        latest_message_id_by_conversation = {row.conversation_id: row.latest_message_id for row in page_rows}
+        group_chat_ids = [row.conversation_id for row in page_rows if row.kind == "group_chat"]
+        host_request_ids = [row.conversation_id for row in page_rows if row.kind == "host_request"]
+
+        # Hydrate each kind in a batch, then re-assemble in the paginated order.
+        group_chats_by_id = _build_group_chats_pb(session, context, group_chat_ids)
+        host_request_threads_by_id = _build_host_request_threads_pb(
+            session, context, host_request_ids, latest_message_id_by_conversation
+        )
```
aapeliv: you are running those queries twice now for each group chat/host request. let's avoid that if we can, since this API will be called a lot. i have done some optimization on the exsting group chats queries in the past given it was taking lots of compute.

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +716,185 @@ def ListGroupChats(
...
+            conversations_pb2.MESSAGE_THREAD_FILTER_UNREAD,
+            conversations_pb2.MESSAGE_THREAD_FILTER_HOSTING,
+            conversations_pb2.MESSAGE_THREAD_FILTER_SURFING,
+            conversations_pb2.MESSAGE_THREAD_FILTER_PUBLIC_TRIPS,
+        )
+
+        if include_chats:
+            latest_message_id_by_conversation = {
+                row.conversation_id: row.latest_message_id
+                for row in session.execute(_group_chat_candidate_query(context, only_archived, unread)).all()
```
aapeliv: it looks like you are using the same logic and same candidate query building; why not unify further?

app/backend/src/couchers/servicers/conversations.py
```diff
@@ -425,6 +716,185 @@ def ListGroupChats(
...
+                    .scalars()
+                    .all()
+                )
+                for host_request in host_requests:
+                    latest_message_id = latest_message_id_by_conversation[host_request.conversation_id]
+                    if host_request.initiator_user_id == context.user_id:
+                        if host_request.initiator_last_seen_message_id < latest_message_id:
+                            host_request.initiator_last_seen_message_id = latest_message_id
+                    elif host_request.recipient_last_seen_message_id < latest_message_id:
+                        host_request.recipient_last_seen_message_id = latest_message_id
```
aapeliv: This will issue queries for every host request separately; we can use a bulk update to reduce the roundtrips

aapeliv (commented):
had a look through the big function and left a lot of comments.

overall, it's quite a big piece. this is touching a high volume surface, so some perf comments as well.

there is a lot of "if public trip, then switch surfer/host", which is something i was hoping we would avoid with the public trips design. i wonder if we are shoehorning it in in a way that is going to be difficult down the line.

app/backend/src/couchers/servicers/conversations.py
nabramow: Ah yes done.

app/backend/src/couchers/servicers/conversations.py
nabramow: done

app/backend/src/couchers/servicers/conversations.py
nabramow: done

app/backend/src/couchers/servicers/conversations.py
nabramow: I'll do MY_PUBLIC_TRIPS

app/backend/src/couchers/servicers/conversations.py
nabramow: will do, good catch

app/proto/conversations.proto
nabramow: yes fair we are using this sporadically in the codebase I'll switch it out and also add to AGENTS.md to prefer next_page_token

app/backend/src/couchers/servicers/api.py
nabramow: @claude can you make a ticket and add to the Engineering project board for aapeli's comment so we can address this in a follow up PR?

app/proto/conversations.proto
nabramow: With the idea that we might combine them on the UI in the future you mean?

tristanlabelle: Possibly, or it might be that it's just as easy/easier to implement it as combinable filters on the backend.

app/backend/src/couchers/servicers/conversations.py
nabramow: I'll de-dupe in the new code. There will be one separate one for the legacy code but we'll remove that once this goes live.

app/backend/src/tests/test_message_threads.py
nabramow: Oh nice, done (as soon as I push my changes)

app/proto/conversations.proto
nabramow: Ah yeah I get what you mean, so basically make it more flexible so we can pass category to show rather than hardcode it. I'll do that, makes it more flexible if we decide different UI in the future.

app/backend/src/tests/test_message_threads.py
nabramow: This is just so it does `x` 200 times so it's not referencing anything specific to remember and doesn't really matter what the number is.

app/backend/src/tests/test_message_threads.py
nabramow: they don't overlap, but I can add a check just in case

app/backend/src/couchers/servicers/api.py
nabramow: I believe this is similar to what @aapeliv commented below about refactoring the ping queries overall in a follow up?

app/proto/public_trips.proto
nabramow: Ooo actually this got duplicated we should use offer count. Removing this.

app/backend/src/tests/test_message_threads.py
tristanlabelle: Then I suggest using a dummy or empty string that won't raise the question!

app/backend/src/couchers/notifications/notify.py
```diff
@@ -82,3 +82,24 @@ def mark_notifications_seen(
...
         .where(Notification.topic_action.in_(topic_actions))
     )
+
+
+def mark_notifications_seen_for_keys(
+    session: Session,
+    *,
+    user_id: int,
+    keys: list[str],
+    topic_actions: list[NotificationTopicAction],
```
aapeliv: This is wrong and dangerous; need to pass in paris of `(topic_actions, keys)` to mark as seen.

app/backend/src/couchers/notifications/notify.py
```diff
@@ -75,10 +76,30 @@ def mark_notifications_seen(
...
+    user_id: int,
+    topic_actions_and_keys: Sequence[tuple[Sequence[NotificationTopicAction], Sequence[str]]],
+) -> None:
+    """
+    Bulk variant of mark_notifications_seen, marking several groups of notifications seen in one update.
+
+    Each group pairs topic actions with the keys those actions are keyed by. Keys are only meaningful
+    within a topic action, so they must not be pooled into one flat list: chat__message is keyed by
+    chat id while chat__missed_messages is a summary keyed with "", and unrelated topics reuse the
+    same key values for entirely different things.
```
aapeliv: remove this

nabramow:
Closing as this was broken into separate PRs in #9460, #9462, #9464, #9472

PR #9229, opened 2026-07-03
Web: sub-community drill-down dropdown on community breadcrumbs

### LINKED ISSUE #5545: Communities Breadcrumbs - add dropdown for more specific communities
If you're in a community page, and there's a more specific community page, we want to add a dropdown allowing the user to select it. Right now they can use breadcrumbs to go less specific, but not more.

For example, for the Europe, Africa and the Middle East we want to show the dropdown with the countries.

For "Germany" community page, the dropdown should show the cities.

This is where it should go:

Image

You should be able to use the same query the browse communities dialog is using.

Acceptance Criteria
- On community page load, a dropdown should show to the right of the breadcrumbs if there are more specific communities
- The dropdown should display the cities in alphabetical order 
- If there are no more specific communities (i.e. already in a city community), the dropdown should not show
- The dropdown should only show the next more specific level. So Global would show "Europe, Africa, and Middle East", "Americas", "Asia-Pacific", Germany would show "Berlin", "Hamburg", "Munich", etc.

kevinortiz43:
Adds a searchable dropdown as the last item inside the community-page breadcrumbs. It lists the current community's direct sub-communities and navigates to the selected one.

Fixes #5545

- New `SubCommunitiesDropdown` — a MUI `Button` + `Menu` with a search field and a filterable item list.
- Reuses the existing `useListSubCommunities` / `listCommunities` query and its pages are drained via `fetchNextPage` so all children load, then the count gates rendering.
- Responsive: on mobile the breadcrumb row stacks above a centered Join/Leave button

## Testing
Verified against online staging (`next.couchershq.org`) with Playwright, logged in:
- Added `SubCommunitiesDropdown.test.tsx` (5 tests). Added a `listCommunities` mock to `DiscussionPage.test.tsx` since it renders the shared sub-header

**Web frontend checklist**

## For maintainers


nabramow:
When I have my language set to a longer language like German, and go in mobile view, the breadcrumbsa re overlapping with the button, we need to come up with a design solution for this:

Screenshot 2026-07-06 at 10 35 13

nabramow:
Two things here:

1. As I type in this search, the width of the menu increases. It should stay the same width the whole time.
2. On mobile view, the placeholder text is cut off initially. Maybe it needs to be something shorter? Just "Location" perhaps?

Screenshot 2026-07-06 at 10 36 56

nabramow:
This also looks a little weird on mobile. Maybe we do conditional design for mobile and make the breadcrumbs the first row and the button the second row. I think there's not enough space for both on the same screen. So can do some kinda 'display: flex, flex-direction: row' for mobile only.

Screenshot 2026-07-06 at 10 39 15

Open to other ideas I just feel something needs to be a bit different here for mobile view.

jesseallhands (commented):
I really like how this is looking! Great job! 

I agree with the points Nicole made. I would suggest that the "leave community"/"join community" button not only appears in a separate row below the breadcrumbs on mobile, but also is a button with a single line of text rather than a button with 2 rows of text since two rows result in it taking up more real estate.

kevinortiz43:
I think when I checked the android version I didn't think to check the other languages. let me fix or see what I can do @nabramow

kevinortiz43:
@nabramow I have fixed the mobile width for German, and other languages. Can you double check.

nabramow:
@kevinortiz43 I fear this still looks a bit awkward on mobile. For example with the "select area" having a gap in space on the left, and the way it's going on 3 lines. What if we:

1) Also make the fontSize slightly smaller on mobile view only so the text fits more
2) Center the "Leave community" button 
3) Do something so that if it wraps to the next line, there's not a space to the left?

Screenshot 2026-07-09 at 10 42 31

Also wondering if @jesseallhands has any better ideas. I think a slightly smaller font will help it wrap a bit though.

kevinortiz43:
@nabramow  I think decreasing the font size should help but I am wondering if for mobile would it be worth calling the countries by their abbreviations. Such as EU for Europe, AFR for Africa, etc.

kevinortiz43:
@nabramow I think the issue is specifically europa africa and middle east is way too long. I tried setting it up in one line but it looks bad. 
image

This is the fix I did.

image

 
image

image

kevinortiz43:
@nabramow  this is the other recommendation we can do for 320px

image

nabramow:
@kevinortiz43 I think the smaller font where it goes 2 rows instead of 3 is okay and left-aligns.

The "Join/Leave Community" button centered looks fine enough, but don't center the tabs, leave that as it is.

kevinortiz43:
@nabramow Hey! I just did a push yesterday, double checking everything this is ready for your review

nabramow (approved):
Okay this looks good to go for me now, merge away! 🚀

PR #9234, opened 2026-07-04
Web/communities: rank community search by visible name, not hidden path (#9129)

### LINKED ISSUE #9129: Frontend/communities: Communities search sort order is unintuitive
Consider searching for `y` on NEXT. You get `Sydney` before `Pyongyang` because what's actually being sorted is `Global Community / Asia-Pacific
 / Australia / Sydney` vs `Global Community / Asia-Pacific / North Korea / Pyongyang`, but that's very not obvious for a user. Any search filtering should be based on text that is displayed to the user.

Image

kevinortiz43:
Community search ranked results by a hidden hierarchical path (parent names), not the visible community name — typing "y" ranked Sydney (…/Australia/Sydney) above Pyongyang (…/North Korea/Pyongyang) because "Australia" < "North Korea". This moves ranking to the server, by visible name (pg_trgm, typo-tolerant).

closes #9129

**Backend (`SearchCommunities`)**
- Ranks by visible name: empty → all name-ordered; 1–2 chars → prefix match; ≥3 chars → fuzzy `word_similarity`. Dropped the old "<3 chars" error.
- Returns a light `CommunitySummary`; retired proto field 1 (`reserved`), data moved to `results = 2`.

**Frontend**
- Single debounced `searchCommunities` query; removed the client-side fetch-all + path sort. `filterOptions={(x) => x}` so MUI keeps the server order.

**Performance**
- Batched the per-row parent lookup (was a recursive-CTE N+1 per result) into one CTE, in `SearchCommunities` and `ListAllCommunities`.

## Testing
- Frontend verified on staging (`next`): `syd` → Sydney, `sydny` (typo) → Sydney, region shown muted. Desktop + mobile + Android emulator.
- `CommunitySearch.test.tsx` passes (incl. #9129 order guard). Backend tests CI-gated (couldn't run locally).

**Backend checklist**
- [ ] Run the backend locally and it works — couldn't run locally (env); CI-gated
- [x] Added migrations if there are any database changes — n/a (uses existing trgm index, migration 0114)

**Web frontend checklist**

## For maintainers


app/web/features/communities/CommunitiesPage/CommunitySearch.tsx
```diff
@@ -1,27 +1,50 @@
...
+  "communityId" | "name" | "slug" | "parentsList"
+>;
+
+// Backend SearchCommunities requires >= 3 chars; below that we browse/filter the
+// full list client-side. 400 ms debounce mirrors the confirmed #8837 decision.
+const MIN_SEARCH_LENGTH = 3;
+const COMMUNITY_SEARCH_DEBOUNCE_MS = 400;
+
+// parentsList is ordered root -> ... -> immediate parent -> self (the node itself
+// is the last entry), so the region context is the second-to-last entry.
```
nabramow: We don't need to leave long comments in like this. As a general rule it's best to make the code readable itself, and only leave a comment for things like trade-offs, design decisions, etc, i.e stuff that's notn immediately obvious

app/web/features/communities/CommunitiesPage/CommunitySearch.tsx
```diff
@@ -30,103 +53,80 @@ const StyledAutocomplete = styled(
...
-        // Finally sort by community name
-        return a.name.localeCompare(b.name, i18n.language);
-      }),
-    [allCommunities, i18n.language],
+      debounce(
+        (value: string) => setDebouncedInput(value),
+        COMMUNITY_SEARCH_DEBOUNCE_MS,
+      ),
+    [],
   );
```
nabramow: Not to throw a rock in this, but generally when I see this much logic for a query on the frontend, it's a sign to me that we need to adjust the logic on the backend to return what we actually need (or make a new query for this purpose). ESPECIALLY as I see two queries here for one component.

Generally the backend should handle business logic, and the frontend should handle presentational logic.

IMO we should do a separate backend PR for the query, then change this frontend PR to consume that change.

kevinortiz43: Done — moved the ranking/business logic to the backend. `SearchCommunities` now serves the dropdown at every query length (empty → all communities name-ordered; 1–2 chars → prefix match on the unaccented trgm index; ≥3 chars → pg_trgm fuzzy) and returns a light `CommunitySummary` payload via a new proto field. The frontend is now a single debounced query that renders the backend order verbatim — the second `listAllCommunities` query and all client-side filtering/sorting are gone.

One deviation from your suggestion: rather than a separate backend PR, I bundled the backend change into this PR. Since the proto is the FE↔BE contract and this dropdown is the only consumer of `SearchCommunities`, shipping both together keeps it atomic (no deploy window where the FE expects a field the BE isn't sending yet). Happy to split if you'd prefer.

app/web/features/communities/CommunitiesPage/CommunitySearch.tsx
kevinortiz43: Done - I removed long comments from this PR

app/web/features/communities/CommunitiesPage/CommunitySearch.tsx
nabramow: Tagging @aapeliv here for backend review.

app/backend/src/couchers/servicers/communities.py
```diff
@@ -173,25 +188,51 @@ def SearchCommunities(
...
+
+        rows = session.execute(query).all()
 
-        return communities_pb2.SearchCommunitiesRes(communities=communities_to_pb(session, rows, context))
+        return communities_pb2.SearchCommunitiesRes(
+            results=[
+                _community_summary_to_pb(session, node, cluster, member_count, user_subscription)
+                for node, cluster, member_count, user_subscription in rows
+            ]
+        )
```
nabramow, line 247: I'm not great with python but I wonder about the performance when I see a for loop like this.

Could you explain real quick and simply what the logic should accomplish here?

kevinortiz43: The database search already ran and handed back a list of matching communities as raw rows. Each row is just loose data: the community node, its cluster info, how many members it has. This loop just reformats each row into the response object the frontend expects for each community it fills in name, slug, member count, whether you're a member, etc. One community in, one clean result out. No extra searching or caching.

kevinortiz43: @nabramow you were right to flag this my earlier "no extra searching" was wrong. `_parents_to_pb` was firing a recursive-CTE query per result row (a per-row N+1), and the same pattern was in `ListAllCommunities`. Fixed: added `get_nodes_parents_recursively`, which fetches all rows' ancestors in one CTE so both endpoints now do a single batched lookup.

kevinortiz43:
@nabramow This is ready for your review

kevinortiz43:
@aapeliv requesting review

aapeliv:
Sorry, it's not entirely clear to me what the issue here is? What's the rationale behind this PR?

kevinortiz43:
@aapeliv the rationale was to solve this bug https://github.com/Couchers-org/couchers/issues/9129 where "Consider searching for y on NEXT. You get Sydney before Pyongyang because what's actually being sorted is Global Community / Asia-Pacific / Australia / Sydney vs Global Community / Asia-Pacific / North Korea / Pyongyang, but that's very not obvious for a user. Any search filtering should be based on text that is displayed to the user."

nabramow:
@kevinortiz43 I'm not a huge python person but I'm not overly convinced by this approach. It seems like a LOT of complicated logic to fix something small. Also that we're hacking around something rather than fixing the core. Can you tell us more why you chose this approach in your own words? 

I think this code might also be replaced soon once @Gwened's location autocomplete work is done (plz correct me if I'm wrong @Gwened), so might not make sense to hack something together before then.

@tristanlabelle created this ticket though so curious his input.

Gwened:
> #9129

If I understand #9129, it doesn't refer to the autocomplete widget (the one helping to enter a world location/address in general). It refers to the ordering in the results (shown as you type) of a database search of the communities.

I can't tell if the current ordering needs fixing, but I don't think the new location works will change it in a near future.

kevinortiz43:
@nabramow so the issue for 9129 the fix moves sorting into the database query itself, ordered by visible name using Postgres fuzzy match, so DB returns ordered results and frontend just displays them. The code in communities.py's SearchCommunities, is a separate N+1 performance fix on parent lookups, not part of the ordering fix itself. N+1 was in calling one parent-lookup query per search result. This was why when you would type Y you would get Sydney before Pyongyang.

nabramow:
> @nabramow so the issue for 9129 the fix moves sorting into the database query itself, ordered by visible name using Postgres fuzzy match, so DB returns ordered results and frontend just displays them. The code in communities.py's SearchCommunities, is a separate N+1 performance fix on parent lookups, not part of the ordering fix itself. N+1 was in calling one parent-lookup query per search result. This was why when you would type Y you would get Sydney before Pyongyang.

Okay thank you that makes more sense. I'll let @aapeliv and @tristanlabelle decide for detailed review since they can python better than me.

kevinortiz43:
Sounds good, I will stand by and listen to any feedback

app/backend/src/couchers/db.py
```diff
@@ -198,15 +198,59 @@ def _get_node_parents_recursive_cte_subquery(node_id: int) -> Subquery:
     ).subquery()
 
 
-def get_node_parents_recursively(session: Session, node_id: int) -> Sequence[Row[tuple[int, int, int, Cluster]]]:
-    subquery = _get_node_parents_recursive_cte_subquery(node_id)
-    return session.execute(
-        select(subquery, Cluster)
+def _get_nodes_parents_recursive_cte_subquery(node_ids: Sequence[int]) -> Subquery:
```
tristanlabelle, line 201: Can `_get_node_parents_recursive_cte_subquery` be implemented in terms of this function now?

app/backend/src/couchers/servicers/communities.py
```diff
@@ -172,25 +199,52 @@ def SearchCommunities(
...
             .where(Cluster.is_official_cluster)
-            .where(word_similarity_score > COMMUNITIES_SEARCH_FUZZY_SIMILARITY_THRESHOLD)
-            .order_by(word_similarity_score.desc(), Cluster.name.asc(), Node.id.asc())
-            .limit(page_size)
         )
 
-        rows = session.execute(query.options(selectinload(Node.official_cluster))).scalars().all()
+        if not raw_query:
+            query = query.order_by(Cluster.name.asc(), Node.id.asc()).limit(page_size)
+        elif len(raw_query) < 3:
```
tristanlabelle, line 222: Switching matching algorithm at 3 chars seems problematic because the results from adding the third char might not be a subset from the results of the first two chars, which would be surprising.

app/backend/src/couchers/db.py
```diff
@@ -198,15 +198,59 @@ def _get_node_parents_recursive_cte_subquery(node_id: int) -> Subquery:
...
+            subquery.c.level,
+            Cluster,
+        )
         .join(Cluster, Cluster.parent_node_id == subquery.c.id)
         .where(Cluster.is_official_cluster)
-        .order_by(subquery.c.level.desc())
+        .order_by(subquery.c.start_node_id, subquery.c.level.desc())
     ).all()
 
+    result: dict[int, list[tuple[int, int, int, Cluster]]] = {node_id: [] for node_id in node_ids}
```
tristanlabelle, line 245: If there's anything we can do about these unreadable nested types, that'd be great. Maybe start with defining a type alias for that tuple?

tristanlabelle (commented):
Ok it looks like the fix is to switch from `listAllCommunities` + client-side filtering to `searchCommunities`, which already orders things in a way that is expected by the users? I think that generally makes sense because server-side filtering scales better. However it's hard to tease that in this PR because of the unrelated changes, and search in the Vercel preview is broken probably due to the protos change.

Summarizing your changes:

1. Batching of community parent recursion
2. 1-2 character queries as prefix matches
3. SearchCommunitiesRes Community -> CommunitySummary
4. Switch from `listAllCommunities` to `searchCommunities` + debouncing input

I'm only unsure about (2) because it means that adding a third char may add results instead of only further filtering to a subset. But mostly we should not be reviewing these changes in a single PR -- it takes a lot of brainpower to realize that they are independent (I think?). Could you split them out in separate PRs?

nabramow:
@kevinortiz43 Are you still working on this or should it be closed? Looks like there are some merge conflicts as well now.

PR #9235, opened 2026-07-04
Frontend: Use iso8601 values with datetime pickers

### LINKED ISSUE #9232: Frontend: Change datetime pickers to deal with ISO8601 strings
It's just too confusing to deal with Date/Dayjs objects when the timezone is undefined.

tristanlabelle:
Switches our `Datepicker`/`Timepicker` components API to use ISO8601 date/time strings (`YYYY-MM-DD` and `HH:mm`) formats, instead of `Dayjs` values.

Rationale: Our datetime pickers are seldom used for picking datetimes in the current timezone:

- Birthdate is timezone-agnostic
- Hosting dates are in the host timezone
- Event dates will soon be in their location's timezone

It's error-prone to represent those as `Dayjs` object because `Dayjs` object will parse the datetime components using the browser timezone and represent it internally as a timestamp that would be incorrect in the real timezone (e.g. for comparison). This has already resulted in bad event time data. By using ISO8601 date/time strings we preserve the timezone-agnosticity of the values entered by the user, and it makes the code call out when we need to turn them into a `Dayjs` for comparison or manipulation.

The underlying MUI `DatePicker`/`TimePicker` still deal in terms of `Dayjs` values, so our wrapper `Datepicker`/`Timepicker` convert internally but otherwise encapsulate this complexity.

It might be valuable to create some `DateOnly` and `TimeOnly` classes that internally use the ISO8601 string as their representation, so the variable names are not as obnoxious.

Fixes #9232

## Testing
Updated tests.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/communities/events/CreateEventPage.tsx
+++ b/app/web/features/communities/events/CreateEventPage.tsx
@@ -16,14 +16,13 @@ import { dashboardRoute, eventsRoute, routeToEvent } from "routes";
 import { service } from "service";
 import type { CreateEventInput } from "service/events";
 import { theme } from "theme";
-import dayjs from "utils/dayjs";
+import { iso8601ToDayjs } from "utils/date";
+import { sendNativeBack, useIsNativeEmbed } from "utils/nativeLink";
 import stringOrFirstString from "utils/stringOrFirstString";
 
-import { sendNativeBack, useIsNativeEmbed } from "../../../utils/nativeLink";
 import { communityEventsBaseKey } from "../../queryKeys";
 import EventForm, { CreateEventVariables } from "./EventForm";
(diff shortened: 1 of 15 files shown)


app/web/features/communities/events/CreateEventPage.tsx
tristanlabelle, line 1: Self-blocking TODO: Test with creating and editing an event.

app/web/features/profile/view/NewHostRequest.tsx
tristanlabelle, line 1: Self-blocking TODO: Test with sending a host request.

tristanlabelle:
Abandoning in favor of #9236

Issue #8113, opened 2026-03-16
Protocol buffer typescript for uint64 is incorrect


tristanlabelle:
The generated typescript for protobuf uint64 fields is incorrect. It's typed as `number`, but at runtime it's a `string`. I think this is because Javascript's number type is a double-precision float and can't accurately represent uint64 values.

```ts
export namespace GetSignupPageInfoRes {
  export type AsObject = {
    lastSignup?: google_protobuf_timestamp_pb.Timestamp.AsObject,
    lastLocation: string,
    userCount: number,
  }
} 
```

PR #9236, opened 2026-07-04
Frontend: Use Temporal PlainDate+PlainTime

tristanlabelle:
Switches most date and time representations to `Temporal.PlainDate/Time` APIs, made available by polyfill, including making our `Datepicker` component use `Temporal.PlainDate` and `Timepicker` use `Temporal.PlainTime` (while internally converting to Dayjs due to MUI requirements).

`Dayjs` objects are problematic because 1. they implicitly use the browser timezone, 2. are not meant to represent date-only or time-only values (timezone agnostic), 3. can represent invalid values. This trickiness lead our events to have incorrect times, and is also inadequate for birthdates (no intrinsic timezone) or host request (host location timezone).

Using `Temporal.PlainDate/Time` objects clearly calls out where we convert plain date/times into a point in time (Dayjs object), scoping down where timezone complexity can get involved (e.g. comparisons with "now").

This also supports #9218 which switches event creation to using iso8601 datetime strings instead of timestamps.

The Temporal polyfill requires some gymnastics to work with jest's module loading (thanks Claude).

There are a few more places we could advantageously replace `Dayjs` with `Temporal` APIs. I might follow-up after this PR.

## Testing
Updated all tests.

Ran locally and signed up with birthdate, created an event with date, updated an event with date, sent a host request with date. All working fine.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/communities/events/EventTimeChanger.test.tsx
+++ b/app/web/features/communities/events/EventTimeChanger.test.tsx
@@ -1,3 +1,4 @@
+import { Temporal } from "@js-temporal/polyfill";
 import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Event } from "proto/events_pb";
@@ -478,10 +479,12 @@ describe("Event time changer", () => {
       expect(onValidSubmit).toHaveBeenCalledTimes(1);
       const submittedData = onValidSubmit.mock.calls[0][0];
 
-      expect(submittedData.startDate.format("YYYY-MM-DD")).toBe("2021-08-05");
-      expect(submittedData.startTime.format("HH:mm")).toBe("14:00");
-      expect(submittedData.endDate.format("YYYY-MM-DD")).toBe("2021-08-06");
-      expect(submittedData.endTime.format("HH:mm")).toBe("15:00");
+      expect(submittedData.startDate).toBe(
+        Temporal.PlainDate.from("2021-08-05"),
+      );
+      expect(submittedData.startTime).toBe(Temporal.PlainTime.from("14:00"));
+      expect(submittedData.endDate).toBe(Temporal.PlainDate.from("2021-08-06"));
+      expect(submittedData.endTime).toBe(Temporal.PlainTime.from("15:00"));
     });
   });
 });
--- a/app/web/features/communities/events/EventTimeChanger.test.tsx
+++ b/app/web/features/communities/events/EventTimeChanger.test.tsx
@@ -125,6 +125,7 @@ describe("Event time changer", () => {
 
   it("should show proper error if startDate is today but startTime is in the past", async () => {
     jest.setSystemTime(new Date("2021-08-01 23:00"));
+    expect(Temporal.Now.plainDateISO().toString() == "2021-08-01");
 
     render(<TestForm />, { wrapper });
 
--- a/app/web/features/communities/events/EventTimeChanger.test.tsx
+++ b/app/web/features/communities/events/EventTimeChanger.test.tsx
@@ -479,12 +479,14 @@ describe("Event time changer", () => {
       expect(onValidSubmit).toHaveBeenCalledTimes(1);
       const submittedData = onValidSubmit.mock.calls[0][0];
 
-      expect(submittedData.startDate).toBe(
+      expect(submittedData.startDate).toEqual(
         Temporal.PlainDate.from("2021-08-05"),
       );
-      expect(submittedData.startTime).toBe(Temporal.PlainTime.from("14:00"));
-      expect(submittedData.endDate).toBe(Temporal.PlainDate.from("2021-08-06"));
-      expect(submittedData.endTime).toBe(Temporal.PlainTime.from("15:00"));
+      expect(submittedData.startTime).toEqual(Temporal.PlainTime.from("14:00"));
+      expect(submittedData.endDate).toEqual(
+        Temporal.PlainDate.from("2021-08-06"),
+      );
+      expect(submittedData.endTime).toEqual(Temporal.PlainTime.from("15:00"));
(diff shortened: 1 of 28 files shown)


app/web/features/communities/events/EventTimeChanger.test.tsx
tristanlabelle, line 1: Self-blocking TODO: Test locally (create+update)

app/web/features/profile/view/NewHostRequest.tsx
tristanlabelle, line 1: Self-blocking TODO: Test locally

nabramow:
@tristanlabelle Looking at this package I'm a little concerned if it's stable?

https://www.npmjs.com/package/@js-temporal/polyfill

Two things stand out to me: 1) It's not at version 1 yet, which means probably lots of breaking changes each update and 2) Last update was a year ago, is it still being maintained? I'd hate to introduce a package and rely on it that's already out of date.

tristanlabelle:
@nabramow , good catch, I found https://www.npmjs.com/package/temporal-polyfill which is more up-to-date. Let me try switching.

app/web/features/communities/events/EventTimeChanger.test.tsx
tristanlabelle: Done

app/web/features/profile/view/NewHostRequest.tsx
tristanlabelle: Done

tristanlabelle:
Switched over and tested locally. Ready for review @nabramow

app/web/jest.config.ts
```diff
@@ -38,7 +39,6 @@ const customJestConfig: Config = {
     // https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object
     "^.+\\.(js|jsx|ts|tsx)$": ["babel-jest", { presets: ["next/babel"] }],
   },
-  transformIgnorePatterns: ["/node_modules/"],
```
nabramow, line 41: Why take this out? Seems unrelated?

app/web/jest.resolver.js
```diff
@@ -0,0 +1,19 @@
...
+    (name) => path === name || path.startsWith(`${name}/`),
+  );
+  if (isEsmOnly) {
+    return options.defaultResolver(path, {
+      ...options,
+      conditions: [...(options.conditions ?? []), "import"],
+    });
+  }
+  return options.defaultResolver(path, options);
+};
```
nabramow, line 19: Wow so this sent me down a whole rabbit hole of "why do we need to support compiling to CommonJS when our frontend is ESM?" and then I realized our frontend protobuf grpc package only compiles to CommonJS, which also means we lose any benefits of Typescript there.

Not your problem here but I added a tech debt ticket as it seems there is a more modern package we could switch to for the frontend protobuf stuff: #9246 . More and more packages are ESM-only (like this temporal-polyfill one), it's the more modern way, so def some tech debt we should handle eventually.

nabramow (dismissed):
Apologies for the delay, I had to really read up on this! Seems we're on the cutting edge of date management since Temporal is new to JS?!

Anyway this all looks good, and I tested a few places and it's working as before.

app/web/jest.config.ts
tristanlabelle: You're right. Claude first had a hacky solution in here that I had it revert but I missed this line. Brought it back.

app/web/jest.resolver.js
tristanlabelle: Yeah I'm just getting introduced to CommonJS vs ESM. Fascinating stuff. I also wanted to investigate changing the protobuf code generator because we get a lot of type annotations wrong: #8113

tristanlabelle:
> Seems we're on the cutting edge of date management since Temporal is new to JS?!

Yes, and it's so great to use a time API that gets the abstractions right. It's a can of worms in most programming languages. Thanks for the review!

app/web/jest.config.ts
tristanlabelle: Hmm removing it causes an error in all tests that depend on Temporal. I asked Claude and it said that line was previously unneeded but inoffensive and now needs removing, so I removed it again. Full explanation follows:

> It wasn't actually functionally necessary — it was harmless boilerplate that happened to become harmful once you added transpilePackages.
> 
> Tracing it back to when jest.config.ts was first created (9b5844e3b, Dec 2024), the custom config had:
> 
> 
> transformIgnorePatterns: ["/node_modules/", "^.+\\.module\\.(css|sass|scss)$"],
> This is a very common copy-paste snippet from generic Next.js+Jest setup guides (from before next/jest reliably supplied good defaults on its own). Looking at next/jest's source, both of those patterns are already things it generates by default:
> 
> The CSS-module pattern ^.+\.module\.(css|sass|scss)$ is always appended regardless of your config.
> The /node_modules/ pattern is exactly what next/jest falls back to when transpilePackages is empty (transpiled ? [...] : ['/node_modules/']).
> So at the time, restating "/node_modules/" in your own config was a no-op — it duplicated a rule that was already in effect, so nothing changed by having it there. It stayed silently redundant for over a year because no dependency needed selective transpilation.
> 
> The moment you added transpilePackages: ["temporal-polyfill", "temporal-utils"], next/jest's generated default pattern changed from a blanket /node_modules/ to a smarter one that excludes those two packages. At that point the old leftover line stopped being a harmless duplicate and became an actively conflicting blanket rule again (since transformIgnorePatterns entries are OR'd, any match — including your redundant literal one — wins).
> 
> So: it was never really "needed," it just used to be inert. Worth calling that out explicitly in a comment (as I proposed) so it doesn't get reintroduced as an "obviously fine, it was there before" cleanup.

PR #9242, opened 2026-07-05
Backend/i18n: Split languages.json to call out custom codes

tristanlabelle:
Gives us a way to track the custom codes vs standard iso638 ones.

Related to #9172

## Testing
N/A - leaving to tests.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/resources.py
+++ b/app/backend/src/couchers/resources.py
@@ -170,9 +170,11 @@ def copy_resources_to_database(session: Session) -> None:
     with open(resources_folder / "regions.json", "r") as f:
         regions = [(region["alpha3"], region["name"]) for region in json.load(f)]
 
-    with open(resources_folder / "languages-iso639.json", "r") as f1:
-        with open(resources_folder / "languages-custom.json", "r") as f2:
-            languages = [(language["alpha3"], language["name"]) for language in (json.load(f1) + json.load(f2))]
+    with (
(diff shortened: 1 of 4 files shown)


app/backend/src/couchers/resources.py
```diff
@@ -170,8 +170,9 @@ def copy_resources_to_database(session: Session) -> None:
     with open(resources_folder / "regions.json", "r") as f:
         regions = [(region["alpha3"], region["name"]) for region in json.load(f)]
 
-    with open(resources_folder / "languages.json", "r") as f:
-        languages = [(language["code"], language["name"]) for language in json.load(f)]
+    with open(resources_folder / "languages-iso639.json", "r") as f1:
+        with open(resources_folder / "languages-custom.json", "r") as f2:
```
aapeliv: you can also comma separate in one with

aapeliv (dismissed):
yeah this is fine if you want them separate. why do you want to call it alpha3 though? :P

app/backend/src/couchers/resources.py
tristanlabelle: Ah, I tried but was missing the parens. Updated in latest.

tristanlabelle:
@aapeliv I used the term `alpha3` because "iso639 code" is still ambiguous. `set3` would be even better if we didn't have a `set2` value in there.

PR #9247, opened 2026-07-08
Frontend/i18n: Use Temporal in localizeDateTime

tristanlabelle:
Updates our `localizeDateTime` function and variants to take `Temporal.PlainDateTime` objects, which means timezones become irrelevant at that layer. This decouples locale formatting from timezone logic and removes many uses of `dayjs`.

For example, we can format a hosting start/end date ("YYYY-MM-DD" string) directly instead of having to interpreting it in an arbitrary timezone and then formatting it by specifying the same timezone again (error prone). E.g. we don't need these comments anymore:

I also updated the test events to have times at minute granularity since that's what the backend allows and that's what gets displayed. Besides `1624934247.732` was not a valid `seconds` value; by protobuf it should be an integer.

```js
// Host request are plain dates (no time),
// just make sure to parse and format them in the same timezone.
```

Next up after this merge I'll look at `timeAgo`.

## Testing
Compared Vercel preview with next. Event times are the same, my timezone and local time are the same, another user's local time are the same, host request dates are the same, current login times are the same.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/profile/view/userLabels.tsx
+++ b/app/web/features/profile/view/userLabels.tsx
@@ -282,9 +282,7 @@ export const RemainingAboutLabels = ({ user }: Props) => {
         text={
           user.joined
             ? localizeYearMonth(
-                timestampToPlainDateTime(user.joined, {
-                  timezone: undefined,
-                }).toPlainDate(),
+                timestampToPlainDateTime(user.joined).toPlainDate(),
                 {
                   locale,
                   abbreviate: true,
@@ -297,9 +295,7 @@ export const RemainingAboutLabels = ({ user }: Props) => {
       <LabelAndText
         label={t("profile:heading.local_time")}
         text={localizeDateTime(
-          Temporal.Now.zonedDateTimeISO(
-            user.timezone || UTC_TIMEZONE,
-          ).toPlainDateTime(),
+          Temporal.Now.plainDateTimeISO(user.timezone || UTC_TIMEZONE),
           {
             locale,
             includeDate: false,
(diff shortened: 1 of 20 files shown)


app/web/features/profile/view/userLabels.tsx
```diff
@@ -279,21 +281,30 @@ export const RemainingAboutLabels = ({ user }: Props) => {
...
         label={t("profile:heading.local_time")}
-        text={localizeDateTime(dayjs(), {
-          timezone: user.timezone || UTC_TIMEZONE,
-          locale,
-          includeDate: false,
-        })}
+        text={localizeDateTime(
+          Temporal.Now.zonedDateTimeISO(
+            user.timezone || UTC_TIMEZONE,
+          ).toPlainDateTime(),
```
tristanlabelle: Can be simplified to `.plainDateTimeISO`

app/web/utils/date.ts
```diff
@@ -255,6 +243,34 @@ export function getMuiTimeFormat(locale: string): string {
...
+  const nanos =
+    BigInt(timestamp.nanos) + BigInt(timestamp.seconds) * 1_000_000_000n;
+  return new Temporal.Instant(nanos);
+}
+
+/// Converts a Temporal Instant to a PlainDateTime in the browser's timezone.
+export function instantToPlainDateTime(
+  instant: Temporal.Instant,
+  // Require callers to explicitly specify the timezone or undefined for clarity.
+  options: { timezone: string | undefined },
```
tristanlabelle: Maybe this is a bit obnoxious.

nabramow (approved):
I clicked around and checked most of the changes components and everything looks the same! Code-wise looks like mostly changing out the function so good to go!

PR #9248, opened 2026-07-08
Frontend: Profile textbox templates to localizable placeholders

### LINKED ISSUE #8558: Frontend/Profile/i18n: Profile fields template text is not translated
See https://github.com/Couchers-org/couchers/blob/315fdef14c3752dd3e261d82a741caf3d3cdfe37/app/web/features/profile/edit/constants.ts

tristanlabelle:
Switches from non-localizable templates with headings to one-line localizable placeholders.

We could just localize the templates but I think as currently implemented they are problematic for users. If you open random profiles, at least half of them have template leftovers in their profile description (pics below). I think using headings is a power user feature so we shouldn't force it by default. Besides our styling for the headings is confusing with the heading of the section. Moreover our template had non-translatable cultural references.

I completely removed the `DEFAULT_ABOUT_HOME_HEADINGS` template because it had a single heading "What I can share with guests", which could be confusing: are you filling in "about my home" or "what I can share with guests"?

Pics from random profiles that include template leftovers today:

image
image
image
image
image

Fixes #8558

## Testing
Vercel preview:

image

image

image

**Web frontend checklist**

## For maintainers


app/web/features/profile/edit/EditProfile.tsx
```diff
@@ -382,9 +378,6 @@ export default function EditProfileForm() {
                 fluency: LanguageAbility.Fluency.FLUENCY_FLUENT,
               })),
             },
-            thingsILike: DEFAULT_HOBBIES_HEADINGS.includes(data.thingsILike)
-              ? ""
-              : data.thingsILike,
```
nabramow, line 387: Shouldn't this still be set to 'data.thingsILike'? Did you mean to remove it all?

app/web/features/profile/edit/EditHostingPreferenceForm.tsx
```diff
@@ -437,7 +430,7 @@ export default function HostingPreferenceForm({
               label={t("profile:home_info_headings.about_home")}
               name="aboutPlace"
               control={control}
-              defaultValue={user.aboutPlace || DEFAULT_ABOUT_HOME_HEADINGS}
+              defaultValue={user.aboutPlace}
```
nabramow, line 433: When I test this one it looks like the placeholder isn't working here when the textbox is empty:

Image

tristanlabelle: From PR description:

> I completely removed the DEFAULT_ABOUT_HOME_HEADINGS template because it had a single heading "What I can share with guests", which could be confusing: are you filling in "about my home" or "what I can share with guests"?

I think a placeholder would make sense if we can come up with 3+ examples of what should go into that box.

app/web/features/profile/edit/EditProfile.tsx
tristanlabelle: It's already interpolated in by `...restData` a few lines above. This was a special case for `thingsILike` that we don't need anymore. I verified editing this field still works.

nabramow (approved):
Looks okay to me and tested fine after discussions in comments. Would be good to come up with a new placeholder for the about my home but I got two hours of sleep so my brain is empty now 😅.

PR #9251, opened 2026-07-09
Frontend/i18n: implement timeAgo in terms of Temporal

tristanlabelle:
Following up to #9236 and #9247, this migrates `timeAgo`, now `localizeRelativeTime`, in terms of the `Temporal` API. I preserved all functionality, improved support for times in the future ("in x minutes"), added formatting options, and added more tests. This further reduces our dependency on `dayjs`, which is more error-prone.

## Testing
Compared the following relative times between Vercel preview and next -- nothing changed: Dashboard discussions, chat messages, host requests, communities discussions, landing last signup, current logins, notifications, profile last active.

Added lots of tests.

**Web frontend checklist**

## For maintainers


app/web/features/communities/discussions/Comment.tsx
```diff
@@ -218,10 +217,9 @@ export default function Comment({
     setIsEditing(false);
   };
 
-  const replyDate = comment.createdTime
-    ? timestamp2Date(comment.createdTime)
-    : undefined;
-  const postedTime = replyDate ? timeAgo({ since: replyDate, t, locale }) : "";
+  const postedTime = comment.createdTime
+    ? localizeRelativeTime(comment.createdTime, locale)
+    : "";
```
nabramow, line 222: Do we not want this replyDate logic anymore? I notice we're doing more than just switching it out here.

app/web/features/communities/discussions/DiscussionCard.tsx
```diff
@@ -78,10 +77,9 @@ export default function DiscussionCard({
     discussion.deleted ? undefined : discussion.creatorUserId,
   );
 
-  const date = discussion.created
-    ? timestamp2Date(discussion.created)
-    : undefined;
-  const postedTime = date ? timeAgo({ since: date, t, locale }) : null;
+  const postedTime = discussion.created
+    ? localizeRelativeTime(discussion.created, locale)
+    : null;
```
nabramow, line 82: Same here we don't need this date logic anymore?

app/web/utils/date.ts
```diff
@@ -273,16 +274,131 @@ export function timestampToPlainDateTime(
...
 }
 
-function timestamp2Date(timestamp: Timestamp.AsObject): Date {
-  return new Date(Math.floor(timestamp.seconds * 1e3 + timestamp.nanos / 1e6));
+export interface LocalizeRelativeTimeOptions {
+  style?: Intl.RelativeTimeFormatStyle;
+  /// We override the default to "auto"
+  numeric?: Intl.RelativeTimeFormatNumeric;
+  /// If true, capitalize if the script supports it.
+  standalone?: boolean;
```
nabramow: I wonder if we should name this something more obvious to what it does? capitalizeSomething, for example.

app/web/utils/date.ts
```diff
@@ -273,16 +274,131 @@ export function timestampToPlainDateTime(
...
 }
 
-/// Localizes a number of days as a relative time string (e.g. "today", "tomorrow", "in 3 days").
-export function localizeRelativeDays(days: number, locale: string): string {
-  return new Intl.RelativeTimeFormat(locale, { numeric: "auto" }).format(
-    days,
-    "day",
-  );
+// Creating Intl objects every time is slow, so cache them.
+const relativeTimeFormatCache = new Map<string, Intl.RelativeTimeFormat>();
```
nabramow, line 286: Is this cached per user? Just making sure we don't accidentally store a ton in memory.

app/web/features/communities/discussions/Comment.tsx
tristanlabelle: There is no functional change: `replyDate` was an intermediate variable for initializing `postedTime`. I now do it in one statement.

app/web/features/communities/discussions/DiscussionCard.tsx
tristanlabelle: Same, no functional change, just simplifying initialization.

app/web/utils/date.ts
tristanlabelle: Switched to `capitalize`, same as `localizeDateTime`.

(standalone is Unicode terminology, but I agree more technical/less obvious)

app/web/utils/date.ts
tristanlabelle: This is either running client-side, where it probably won't exceed a couple entries (lifetime = how long you stay on the page), or if doing server-side rendering could be at most multiplied by the number of languages, so it's not per-user nor unbounded (Claude estimates max ~144 entries)

tristanlabelle (commented):
@nabramow addressed all comments

nabramow:
I'll approve this but just making a note that this response time one seems kinda confusing. Not sure if related to this though!

Screenshot 2026-07-15 at 09 41 34

nabramow (approved):
Code looks good, clicking around the app and everything looks okay.

tristanlabelle:
@nabramow , the response time part is unrelated. We're not using `timeAgo()` there but rather `dayjs`' `.humanize()`. It's on the back of my mind to improve it later.

https://github.com/Couchers-org/couchers/blob/58426e8416497eb838d3a48aa76294d55a630cf9/app/web/features/profile/locales/en.json#L275

Issue #9246, opened 2026-07-08
Migrate proto codegen from grpc-web to Connect-ES / Protobuf-ES (ESM-native)


nabramow:
## Summary

Our web frontend (`app/web`) generates gRPC clients with **protoc-gen-grpc-web**, which emits **CommonJS** — 64 `proto/*_pb.js` + `*_grpc_web_pb.js` files. This keeps the app as a CommonJS package (no `"type": "module"`) and forces build/test shims every time we adopt an ESM-only dependency.

This surfaced in #9236 (Temporal adoption), which required two workarounds purely for ESM/CJS interop:
- `transpilePackages` in `next.config.js`
- a custom `jest.resolver.js` (to request the `import` export condition)

## Why now

ESM-only packages are increasingly common (e.g. `temporal-polyfill`). The official `grpc-web` generator has no real ESM support — grpc/grpc-web#535 has been open for years — so we can't fix this by bumping a version or flipping a flag. The ecosystem's modern, ESM-native answer is Buf's **Connect-ES + Protobuf-ES**.

## Proposed direction

- Replace `protoc-gen-grpc-web` with **`protoc-gen-es`** (+ `protoc-gen-connect-es`).
- Use **`@connectrpc/connect-web`**'s `createGrpcWebTransport` — this speaks the **same gRPC-web wire protocol**, so **no backend / Envoy changes needed**.
- Result: native ESM generated code (tree-shakeable, TS-first), removes all 64 CommonJS proto files, and unblocks moving `app/web` to `"type": "module"`.

## Scope / cost (not a version bump — a migration)

- ~319 `proto/*` import sites across the app, plus the `service/*.ts` layer.
- Message-construction API changes: `const r = new Foo(); r.setBar(x)` (google-protobuf) → Protobuf-ES field style. Sweeping, mostly mechanical.
- **Recommend a spike first:** migrate ONE service end-to-end to size the full effort before committing.

## Notes

- No backend change required (gRPC-web transport retained).
- Separately/independently: the Temporal polyfill shims (`transpilePackages`, `jest.resolver.js`) can be deleted once native Temporal reaches Baseline (~mid-2026). Smaller, standalone cleanup.

## References

- PR #9236 (Temporal adoption — where this surfaced)
- grpc/grpc-web#535 (ESM support, long-open)
- https://github.com/bufbuild/protobuf-es
- https://github.com/connectrpc/connect-es

tristanlabelle:
Related to #8113

PR #9252, opened 2026-07-09
chore: Update Jest to 30

### LINKED ISSUE #6978: Upgrade jest to v30
https://jestjs.io/docs/upgrading-to-jest30

darrenvong:
Replaced deprecated test assertions alias that are no longer supported in Jest 30

Closes #6978

This was luckily a lot less painful than expected. The major relevant changes:
- Replacing deprecated matcher alias as detailed in the [migration guide](https://jestjs.io/docs/upgrading-to-jest30#removal-of-alias-matcher-functions). The relevant ones for us seem to be:
  - `expect(fn).toBeCalled()` → `expect(fn).toHaveBeenCalled()`
  - `expect(fn).toBeCalledTimes(n)` → `expect(fn).toHaveBeenCalledTimes(n)`
  - `expect(fn).toBeCalledWith(arg)` → `expect(fn).toHaveBeenCalledWith(arg)`
- Replacing direct `window.location` manipulation via `Object.defineProperty` with the web-standard API for mutating URL (`window.history.pushState`) since this is no longer supported past JSDOM 20. Bumping `jest-environment-jsdom` to the latest version has bumped the underlying JSDOM to v26.

## Testing

**Backend checklist**

**Web frontend checklist**
N/A - this is purely a test package upgrade change, with relevant tweaks due to breaking changes introduced from upgrading!

## For maintainers


app/web/jest.config.ts
```diff
@@ -1,5 +1,5 @@
 import type { Config } from "jest";
-import nextJest from "next/jest";
+import nextJest from "next/jest.js";
```
darrenvong, line 2: According to Claude:
> Jest 30 loads the TS config through ESM, and ESM resolution doesn't auto-complete the `.js` that `next/jest` relied on under CommonJS

Seems legit as the error itself when running the test suite locally suggested the same fix:
```
Error: Jest: Failed to parse the TypeScript config file .../jest.config.ts
  Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/next/jest' imported from .../jest.config.ts
Did you mean to import "next/jest.js"?
```

nabramow: Another argument for us to do #9246 I think, then we don't have to worry about CommonJS anymore.

nabramow (approved):
This all looks pretty straight-forward! Thanks for doing this! Tests passing, no type errors and website runs fine.

app/web/jest.config.ts
darrenvong: Yeah 100%, I remember coming across more modern implementation of the gRPC transpiler which seems to have more accurate types as well. Definitely sounds like a spike task!

Issue #8684, opened 2026-05-18
Make "username" and "display name" more clear during signup to prevent confusion


jesseallhands:
Support has received multiple complaints in the past from people claiming they did not know their username would be shown publicly (and they either request to change it or delete their account and create a new one). 

The current helper text of the "Username" input box during signup says "This is your unique login ID, not your display name", which is confusing because it makes it seem like this would not be publicly visible and only used for login. This helper text also is no longer visible if you click in the box and then click somewhere else.

Current situation:
<img width="1036" height="814" alt="Image" src="https://github.com/user-attachments/assets/bc8fd1a9-f7b4-4734-ad01-c4320cab795b" />

<img width="1038" height="1008" alt="Image" src="https://github.com/user-attachments/assets/db1b95c4-d3ef-49d0-8bab-6262ae0d7425" />

To fix the recurring confusion some users have, we should put the username and display name text input boxes right by each other (username on top, display name directly below). The labels could be improved by providing examples:

* `Username`
`A unique ID displayed on your profile, eg. "exampleusername"`
and 
* `Name`
`Your real name displayed on your profile, eg. "John" or "John Smith"`

Since display name is already populated from the first signup page where the person enters their name and email, the "Name" field would be pre-populated with whatever they put there.

PR #9254, opened 2026-07-09
Update en.json to address issue 4145 and 8684

### LINKED ISSUE #4145: Add helper text to "username" text field during signup
Some users have been confused about what is meant by "username" and have contacted support or reported a bug because of that confusion (some think this is the name actually displayed on their account).

Could we add a helper text below the "username" text field to help differentiate between "username" and "display name"? Or allow users to provide their username and their display on the same page so they can see there is a difference?

This issue was also mentioned in Couchers-org/couchers#2409

nc225:
closes #4145 and #8684 

## Testing
Checked the text for errors

**Web frontend checklist**

### DIFF
--- a/app/web/features/auth/locales/en.json
+++ b/app/web/features/auth/locales/en.json
@@ -69,7 +69,7 @@
     "header": "Your account details",
     "username": {
       "field_label": "Username",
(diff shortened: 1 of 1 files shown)


nc225:
closing this PR to revise + resubmit

PR #9261, opened 2026-07-10
clarify that username is public

### LINKED ISSUE #4145: Add helper text to "username" text field during signup

nc225:
clarifying that username is public and different from profile name
closes https://github.com/Couchers-org/couchers/issues/4145 and https://github.com/Couchers-org/couchers/issues/8684
## Testing
checked text
**Web frontend checklist**

## For maintainers


nabramow:
@nc225 Is the profile name helper text supposed to show here?

Screenshot 2026-07-14 at 09 49 56

nc225:
@nabramow No, I believe that box is for the user's profile/display name. It should show on the next page when creating username

nabramow:
> @nabramow No, I believe that box is for the user's profile/display name. It should show on the next page when creating username

The username one is fine. I'm talking about this change:

```
"helper_text": "This is your profile name."
```

Where should that be visible?

nc225:
@nabramow Hmm good point. I'm not sure where that ones comes in, maybe in Edit Profile? In any case, I had included that part since I had thought the username/name were on the same screen, but since they're not how about we drop the `name` helper_text and keep the `username` helper_text. Shall I submit a new PR?

nabramow:
> @nabramow Hmm good point. I'm not sure where that ones comes in, maybe in Edit Profile? In any case, I had included that part since I had thought the username/name were on the same screen, but since they're not how about we drop the `name` helper_text and keep the `username` helper_text. Shall I submit a new PR?

You don't need to submit a new PR, I'd just remove that line from this one if we're not sure where it should go!

nabramow (approved):
Looks good to me now!

PR #9264, opened 2026-07-11
fix(profile): sort localized tags using Chinese collation

### LINKED ISSUE #9228: Profile/i18n: Region/language list sorting is off
Confirmed by @HankAviator :

Yeah it looks quite weird. The most common two sorting methods are

1. Based on Chinese Pinyin
2. or based on English translation

It's definitely not method 1 to me

We're possibly sorting on the country/language code instead of the display string.

Image

HankAviator:

Fixes #9228 

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/web/features/profile/ProfileTagInput.tsx
```diff
@@ -36,6 +36,18 @@ interface ProfileTagInputProps {
   inputFieldProps?: ControllerRenderProps<any, string>;
 }
 
+export function getLabelSortLocales(language: string): string[] {
+  if (language === "zh-Hans" || language.startsWith("zh-Hans-")) {
+    return ["zh-Hans-u-co-pinyin", "zh-CN-u-co-pinyin", "zh-u-co-pinyin"];
```
tristanlabelle: Nice find! Do we need this kind of sorting anywhere else in the app? It might make sense to extract this to `i18n/sorting.ts`, If so where we could further abstract to `getLocalizedListComparer` or something and return the `Intl.Collator.compare` function directly.

tristanlabelle: Let's leave a comment in those two branches, for non-Chinese-aware devs

app/web/features/profile/ProfileTagInput.tsx
```diff
@@ -36,6 +36,18 @@ interface ProfileTagInputProps {
...
+  }
+
+  if (language === "zh-Hant" || language.startsWith("zh-Hant-")) {
+    return ["zh-Hant-u-co-stroke", "zh-TW-u-co-stroke", "zh-u-co-stroke"];
```
tristanlabelle: I'm wondering if all machines will support this sorting order (e.g. on Windows you might have to install a language pack). To make sure it doesn't blow up, let's keep `language` as the last fallback, since that'll probably work.

app/web/features/profile/ProfileTagInput.test.ts
```diff
@@ -0,0 +1,50 @@
+import { getLabelSortLocales } from "./ProfileTagInput";
+
+describe("getLabelSortLocales", () => {
+  it("uses pinyin collation for zh-Hans", () => {
+    const locales = getLabelSortLocales("zh-Hans");
+
+    expect(locales).toEqual([
+      "zh-Hans-u-co-pinyin",
+      "zh-CN-u-co-pinyin",
+      "zh-u-co-pinyin",
```
tristanlabelle: You could create the `Intl.Collator` and expect() that its `.resolvedOptions().locale` has `-pinyin` instead. I'm wondering if that'll be the case on all machines if they don't have Chinese support installed though, so maybe it's safer to only do the comparison assertion if we get a -pinyin locale.

tristanlabelle (commented):
Great find, thanks! A few comments before we merge, or let me know if you want to hand it off to me.

HankAviator:
Pushed a new commit to address all comments, thanks @tristanlabelle !

app/web/i18n/sorting.test.ts
```diff
@@ -0,0 +1,37 @@
+import { getLocalizedListComparer } from "./sorting";
+
+describe("getLocalizedListComparer", () => {
+  it("uses pinyin collation for zh-Hans when supported", () => {
+    const resolvedLocale = new Intl.Collator(
+      "zh-Hans-u-co-pinyin",
+    ).resolvedOptions().locale;
```
tristanlabelle: There's an implicit dependency between the constant here and the one in `getLocalizedListComparer`. We can avoid that by also exporting something like `getLocaleCollator` from `sorting.ts` and return the `Intl.Collator` object, then use `.resolvedOptions().locale` and `.compare` on that object.

Then `getLocalizedListComparer` is simply `getLocaleCollator(locale).compare` so up to you if you want to keep it. I suggested it to have a user-friendlier name because most of us have no idea what a collator is.

tristanlabelle (dismissed):
Thanks!

tristanlabelle (approved):

app/web/i18n/sorting.test.ts
HankAviator: Refactored to address this, added documented `getLocaleCollator()` and retained the friendly `getLocalizedListComparer()` API in sorting.ts

Issue #9262, opened 2026-07-10
Update Communities Section on Dashboard


c-kreuz:
The aim is to align the community card design with other sections of the dashboard, to make it more obvious, that the member is part of several communities. See "Your upcoming stays" for reference.

In addition this text should also be updated to (add "and join"):
You have been added to all communities based on your location. Feel free to [browse and join communities](https://next.couchershq.org/communities) in other locations as well.

This text should be removed:
"Don't see your community? [Get it started!](https://forms.monday.com/forms/d7b6f1bd47a092e23b63f6ef9db1594b)"

<img width="375" height="357" alt="Image" src="https://github.com/user-attachments/assets/bdc295bb-3eb8-422a-905c-2781096a1dff" />

<img width="377" height="240" alt="Image" src="https://github.com/user-attachments/assets/28df88aa-a181-4cf2-9a5d-7a3a73afe177" />

Issue #9221, opened 2026-07-02
reorder communities on dashboard and communities page


jesseallhands:
**Problem** 

Right now in the "Your communities" section on dashboard and communities page, the Global, Continental and Country communities are shown first followed by city communities. This is leading to people overlooking their local communities since many people don't realize you can swipe to scroll on mobile to show more communities.

**Solution** 
Since city communities are the ones people use the most, they should be shown first. If the user has joined multiple cities or countries, their "home" city (associated with their profile) should be shown first followed by other cities in alphabetical order. Same with countries (home first, rest in alphabetical order).

I believe the order should be changed on the backend not the frontend but that should be clarified with Aapeli and Nicole first.

kevinortiz43:
"I propose we sort by "region type" smal -> large, then community id asc (latest ones first)"

1. Scroll to see cards in the community section
2. Too easy to request communities - maybe disable this request or button?

nabramow:
Closing as the actual re-ordering was done in. Any design changes we want can be considered as a separate issue.

@kevinortiz43 I will make a separate ticket and assign to you for the design stuff.

PR #9270, opened 2026-07-12
Communities: order "Your communities" city-first and align section design

kevinortiz43:
Addresses #9262.

**#9262 — card design**
- Matched the community card design to "Your upcoming stays".
- Added a member-count badge to the header (reused from Upcoming Stays) so it's obvious how many communities you're in.
- Removed the "Don't see your community? Get it started!" prompt and added "and join" to the browse text (en.json only).
- Community cards are now fixed 200px.

before

after
image

## Testing
- Verified on desktop + mobile locally.

**Web frontend checklist**
- [ ] Added tests where relevant — no new FE test (styling + copy + shared-component extraction)
- [x] Checked Desktop and Mobile

## For maintainers


app/backend/src/couchers/servicers/communities.py
```diff
@@ -570,18 +570,31 @@ def ListUserCommunities(
...
+
+        user_geom = select(User.geom).where(User.id == user_id).scalar_subquery()
+        type_rank = case(
+            (Node.node_type == NodeType.locality, 0),
+            (Node.node_type == NodeType.sublocality, 1),
+            (Node.node_type == NodeType.subregion, 2),
+            (Node.node_type == NodeType.region, 3),
+            (Node.node_type == NodeType.macroregion, 4),
+            (Node.node_type == NodeType.world, 5),
+            else_=6,
```
aapeliv: the enums are ordered, you don't need to cast to a rank

aapeliv (dismissed):

app/backend/src/couchers/servicers/communities.py
kevinortiz43: For this story they wanted the "local community" or their city to be show first rather than the global community, that's is why its ranked like this. https://github.com/Couchers-org/couchers/issues/9221

aapeliv: Screenshot 2026-07-19 at 16 15 37

I'm saying that the enum is already ordinal (ordered). So you can simply order by NodeType desc, or similar. You don't need to hardcode numbers here (which is also harder to maintain)

app/backend/src/couchers/servicers/communities.py
```diff
@@ -570,18 +570,31 @@ def ListUserCommunities(
...
+        )
+        is_home = func.ST_Contains(Node.geom, user_geom)
```
aapeliv: let's get rid of this special behavior, it makes sorting difficult, and it doesn't fit the communities logical design

app/backend/src/couchers/servicers/communities.py
```diff
@@ -570,18 +570,31 @@ def ListUserCommunities(
...
                 select(Node)
                 .join(Cluster, Cluster.parent_node_id == Node.id)
                 .join(ClusterSubscription, ClusterSubscription.cluster_id == Cluster.id)
                 .where(ClusterSubscription.user_id == user_id)
                 .where(Cluster.is_official_cluster)
-                .where(Node.id >= next_node_id)
-                .order_by(Node.id)
+                .order_by(type_rank, is_home.desc(), Cluster.name.asc(), Node.id.asc())
                 .limit(page_size + 1)
+                .offset(offset)
```
aapeliv: offset+limit requires postgres to render the entire nodes list; we can instead store the pagination logic in the page token (NodeType ordinal + Node id) which allows us to filter in SQL

nabramow:
Just popping in, I don't think we're supposed to edit non-english translation files in PRs. Am I remembering that right @tristanlabelle ? It causes merge conflicts in Weblate and I think weblate will handle removals itself when the english is removed?

app/backend/src/couchers/servicers/communities.py
kevinortiz43: As we have talked I used node_type.desc() to sort the communities instead

app/backend/src/couchers/servicers/communities.py
kevinortiz43: I removed this code as per our conversation

app/backend/src/couchers/servicers/communities.py
kevinortiz43: Filtering first in SQL.

kevinortiz43:
@nabramow added the non en translations back in.

tristanlabelle:
> Just popping in, I don't think we're supposed to edit non-english translation files in PRs. Am I remembering that right @tristanlabelle ? It causes merge conflicts in Weblate and I think weblate will handle removals itself when the english is removed?

Correct. It's better to avoid modifying translation files unless there's a strong reason. Rationale here: https://github.com/Couchers-org/couchers/blob/develop/docs/weblate.md#avoiding-conflicts . Weblate will cleanup removals automatically indeed. I see the latest changes to this PR only has `en.json` changes 👍

kevinortiz43:
Yeah I removed the non english ones because it is no longer used but I added them back to avoid any issues.

app/backend/src/couchers/servicers/communities.py
kevinortiz43: No longer needed with PR 9301

app/backend/src/couchers/servicers/communities.py
kevinortiz43: No longer needed with PR 9301

app/backend/src/couchers/servicers/communities.py
kevinortiz43: No longer needed with PR 9301

jesseallhands (dismissed):
The order seems to be working well but I noticed a couple things:

1. Tap to swipe on mobile doesn't work like with upcoming guests etc.
2. The gradient fade on the side to show you can swipe isn't present for communities. This is what I mean by the fade:

Image

kevinortiz43:
Ah I see mobile issue, okay will fix it up

kevinortiz43:
@nabramow this is ready for you review

kevinortiz43:
I fixed the android scrolling issue

kevinortiz43 (commented):
I fixed the inability to scroll communities cards in the dashboard.

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -103,21 +104,7 @@ function UpcomingStaysWidget({
...
-                background: "var(--mui-palette-grey-50)",
-                borderRadius: 999,
-                px: 1.125,
-                lineHeight: "20px",
-                display: "inline-block",
-              }}
-            >
-              {requests.length}
-            </Box>
+            <SectionCountBadge count={requests.length} />
```
nabramow: I think it would make sense to use our existing `Badge` component for this, for example that's what we use to show number of references on the references tab in profile.

nabramow:
This is working well for me! I only wonder if we should make the width of the community cards a little smaller? Curious what you think @jesseallhands ? It seems they mostly don't take up all the space. But I get the thought was to match the other cards?

Screenshot 2026-07-24 at 09 06 52

kevinortiz43:
Yeah I was trying to make it consistent like the other cards but I can make them smaller as well. I will wait till Jesse gives feedback.

app/web/features/dashboard/UpcomingStays.tsx
kevinortiz43: Okay I can use the existing grey pill

kevinortiz43:
@nabramow regarding the card size should I just make it consistent like the other cards in the dashboard or just shrink them to look better?

kevinortiz43:
I think that is the last thing for this task

nabramow:
> @nabramow regarding the card size should I just make it consistent like the other cards in the dashboard or just shrink them to look better?

Since no one has responded I say shrink them a bit in width! Just check one of the longer ones like the Europe one to make sure it looks okay. If people don't like it we can always adjust a bit later!

app/web/features/dashboard/UpcomingStays.tsx
```diff
@@ -103,21 +104,7 @@ function UpcomingStaysWidget({
...
-                background: "var(--mui-palette-grey-50)",
-                borderRadius: 999,
-                px: 1.125,
-                lineHeight: "20px",
-                display: "inline-block",
-              }}
-            >
-              {requests.length}
-            </Box>
+            <CountBadge count={requests.length} />
```
nabramow: Let's remove the CountBadge component please, also from the other sections. In general good to keep a PR on just the specific topic and not creep in other things in other areas. Was this requested by someone?

app/web/features/profile/view/UserCard.tsx
```diff
@@ -35,9 +36,7 @@ export const sectionLabels = (
         }}
       >
         {t("profile:heading.references")}
-        {!!user?.numReferences && (
-          <StyledNumReferences>{user?.numReferences}</StyledNumReferences>
-        )}
+        {!!user?.numReferences && <CountBadge count={user.numReferences} />}
```
nabramow: Let's put this back? I'm a bit confused about why this change? Did I miss something in that Slack thread?

nabramow:
The communities component itself looks good and works well! Just some comments about removing extra UI changes that weren't requested as far as I can see? Please correct me if I am wrong!

app/web/features/dashboard/UpcomingStays.tsx
kevinortiz43: @nabramow this was part of the original tasking https://github.com/Couchers-org/couchers/issues/9262 "The aim is to align the community card design with other sections of the dashboard, to make it more obvious, that the member is part of several communities. See "Your upcoming stays" for reference." to add the count badge. Would you like me not to add the count badge anymore?

app/web/features/profile/view/UserCard.tsx
kevinortiz43: Hey sorry uh a bit confused the tasking https://github.com/Couchers-org/couchers/issues/9262 asked me to add the count badge, if you want I can revert it easily.

kevinortiz43: Ah I think it was a task created while you were travelling maybe that is why there is confusion

nabramow: yes two things here:

1) Let's not start to combine multiple tickets into one, it gets chaotic for everyone. Stick to the scope of the one ticket you're working on. If a non-dev starts pushing to add things outside of the scope, it's generally better to push back on that (this will happen a lot at work!). This ensures things get done iteratively, rather than stuff added and added and not getting done, and probably will also make you want to pull your hair out less! ;-)
2) If it's not in the TODO column (or at least Coming Next), it means I might not have even seen it yet, and anyone could have added it to the board. It's unsorted, not agreed upon and does not mean we will for sure do it. Once it's prioritized (in TODO or Coming Next columns), then it's safe to do.

jesseallhands:
> @nabramow regarding the card size should I just make it consistent like the other cards in the dashboard or just shrink them to look better?

This sounds good to me! I would almost want two rows of cards then since we are basically saving that much room. My only worry is what it might look like with really long community names like "Las Palmas de Gran Canaria"

app/web/features/profile/view/UserCard.tsx
kevinortiz43: @nabramow no problem I will take out the count badge and all of these instances form this PR and then make that code a separate PR. In the future I will make sure the task has been agreed upon first.

nabramow:
> > @nabramow regarding the card size should I just make it consistent like the other cards in the dashboard or just shrink them to look better?
> 
> This sounds good to me! I would almost want two rows of cards then since we are basically saving that much room. My only worry is what it might look like with really long community names like "Las Palmas de Gran Canaria"

Let's keep one row for now as on mobile that takes up a lot of real estate.

kevinortiz43:
@nabramow okay!

nabramow (approved):
@kevinortiz43 pushed a couple changes since we already went back-and-forth quite a bit. I removed the forced pagination and changed it to the existing pattern we use for the other widgets, and I removed the changes to `UserCard` since it's not related to this PR.

This is good to go now let me know if you have any questions!

PR #9282, opened 2026-07-15
Frontend: More dayjs -> Temporal migration

tristanlabelle:
- Migrate misc code still using dayjs to Temporal.
- Abstract dayjs' "humanize" api behind `localizeDuration` for consistency and so we can switch to browser APIs in the future.
- Extract spreading a duration's units into approximate date units and back (`approxTimeDuration` + `approxDateDuration`)
- Add more tests

There are still some public trip-related code using dayjs but I don't know how to test that so I didn't touch it.

## Testing
Compared vercel preview and next: host requests, last active, local time, upcoming trips+stays on dashboard. Saw no difference.

**Web frontend checklist**

## For maintainers


app/web/utils/date.ts
```diff
@@ -274,6 +275,93 @@ export function timestampToPlainDateTime(
   return instantToPlainDateTime(timestampToInstant(timestamp), timezone);
 }
 
+const APPROX_DAYS_PER_YEAR = 365;
+const APPROX_DAYS_PER_MONTH = 30;
+
+/// Converts a duration which might have date units,
+/// to an approximate equivalent which only has time units.
```
nabramow, line 282: Could you put an example in the comment maybe I'm still a bit unsure the use case upon reading this. Is this like "2 hours ago"?

app/web/utils/date.ts
```diff
@@ -364,41 +462,31 @@ export function localizeRelativeTime(
...
-    if (!isPositive) duration = duration.negated();
-  }
-  return localizeDurationLargestUnit(duration, locale, options);
+  const duration = instant.since(options?.relativeTo ?? Temporal.Now.instant());
+  return localizeTimeOffset(duration, locale, options);
 }
 
-export { numNights };
+/// Localizes a duration (amount of time), expressing it in the largest possible unit,
+/// e.g. "4 days" for a duration of 4 days and 3 minutes. (no "in " prefix)
```
nabramow, line 472: This is a great comment example!

app/web/features/messages/groupchats/MuteDialog.tsx
```diff
@@ -20,7 +20,8 @@ import { useTranslation } from "i18n";
 import { GLOBAL, MESSAGES } from "i18n/namespaces";
 import React, { useState } from "react";
 import { service } from "service";
-import dayjs from "utils/dayjs";
+import { Temporal } from "temporal-polyfill";
+import { approxTimeDuration } from "utils/date";
```
nabramow, line 24: Not me just learning we had a mute feature 😅

nabramow (dismissed):
Just one nit with the comment, otherwise everything seems to work as before!

app/web/utils/date.ts
tristanlabelle: Fair, I added this:

/// E.g. if the user asks to snooze something for 1 month (non-specific duration),
/// we need to convert that to an amount of time by approximating hours/month.

app/web/features/messages/groupchats/MuteDialog.tsx
tristanlabelle: Same here!

PR #9301, opened 2026-07-19
Backend: order ListUserCommunities by node type, most specific first

### LINKED ISSUE #9221: reorder communities on dashboard and communities page

aapeliv:
`ListUserCommunities` was ordered by `Node.id` asc, which generally meant the World community first, then continents, countries, and so on — the least useful order for the user. This changes the ordering to `(node_type desc, Node.id asc)` so the most specific communities (sublocality → … → world) come first, and stores the resulting double cursor `(node_type, node_id)` in an encrypted page token, following the existing `encrypt_page_token`/`decrypt_page_token` pattern used by `ListCommunities`.

The Postgres `nodetype` enum sort order matches the Python ordinals (migration 0136 inserted `macroregion` with `AFTER 'world'`), so `node_type DESC` in SQL sorts sublocality first and world last.

## Testing
- Updated the expected order in `test_ListUserCommunities`, `test_ListOtherUserCommunities`, and `test_enforce_community_memberships_for_user`.
- Added pagination coverage to `test_ListUserCommunities`: pages through with `page_size=2`, passing `next_page_token` forward across node-type boundaries to exercise both legs of the cursor condition, and asserts the final page returns no token.
- All 25 tests in `test_communities.py` pass; `make format` and `make mypy` are clean.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/communities.py
+++ b/app/backend/src/couchers/servicers/communities.py
@@ -6,7 +6,7 @@ import grpc
 from google.protobuf import empty_pb2
 from sqlalchemy import select
 from sqlalchemy.orm import Session, selectinload
-from sqlalchemy.sql import delete, func, or_
+from sqlalchemy.sql import and_, delete, func, or_
 
 from couchers.constants import COMMUNITIES_SEARCH_FUZZY_SIMILARITY_THRESHOLD
 from couchers.context import CouchersContext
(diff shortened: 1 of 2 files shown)


nabramow (approved):
Code looks good and I pulled and tested and can confirm the smaller communities appear first now!

Tagging @kevinortiz43 as an FYI as I believe this will close the backend part of his ticket #9270

PR #9302, opened 2026-07-19
Backend: Cache localized GetLanguages/GetRegions responses per locale

aapeliv:
Since #9173, `Resources/GetLanguages` localizes all ~300 language names on every request. Each name goes through `babel.Locale.parse`, and ~120 of our ISO639-3 codes aren't known to babel at all — for those, babel does an uncached filesystem stat and raises `UnknownLocaleError` each time. That adds up to ~38ms of pure CPU per request (measured warm), on an endpoint that used to be a cached-dict lookup — which is what caused the backend CPU usage spike starting 2026-06-28.

The responses are fully static per locale, so this caches the built response protos in an `lru_cache` keyed on the babel locale (bounded by the set of supported UI locales). Per-request cost returns to effectively zero; the first request per locale after startup pays the localization cost once. `GetRegions` gets the same treatment (it was cheap, but it's the same shape).

Sharing one proto message across requests is safe here: nothing mutates responses after the servicer returns them — the tracing interceptor's sensitive-field sanitization deepcopies first — and there's a comment in the servicer noting this caveat.

Also removes the now-unused per-code `try_localize_*` wrappers from `LocalizationContext`, since the servicer now calls the module-level functions from `couchers/i18n/localize.py` directly (those remain and are still tested).

## Testing

- Benchmarked the old path warm: ~38ms CPU per `GetLanguages` call (~24ms of that in the exception path for the 122 codes babel doesn't know); the new path is an `lru_cache` hit after the first call per locale.
- `uv run pytest src/tests/test_resources.py src/tests/test_localize.py` — 13 passed. `test_resources.py` already covers localized language/region names for non-English locales, which exercises the cached path.
- `make format` and `make mypy` pass.

**Backend checklist**
- [x] Added tests for any new code or added a regression test if fixing a bug (covered by existing localization tests in `test_resources.py`/`test_localize.py`)
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no database changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/resources.py
+++ b/app/backend/src/couchers/servicers/resources.py
@@ -1,17 +1,14 @@
+import functools
 import logging
 
+import babel
 from google.protobuf import empty_pb2
 from sqlalchemy.orm import Session
 
 from couchers.context import CouchersContext
+from couchers.i18n.localize import get_localized_language_names, get_localized_region_names
 from couchers.proto import resources_pb2, resources_pb2_grpc
-from couchers.resources import (
-    get_badge_dict,
-    get_icon,
-    get_language_dict,
-    get_region_dict,
-    get_terms_of_service,
-)
+from couchers.resources import get_badge_dict, get_icon, get_terms_of_service
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/servicers/resources.py
```diff
@@ -23,6 +20,36 @@
 ]
 
 
+# These responses are static per locale, and localizing ~300 names costs tens of ms of CPU (mostly
+# babel.Locale.parse), so cache the built protos. Sharing one message across requests is safe as
+# long as nothing mutates responses after return (the tracing interceptor deepcopies before
+# sanitizing).
+@functools.lru_cache
+def _get_regions_res(locale: babel.Locale) -> resources_pb2.GetRegionsRes:
```
tristanlabelle, line 28: I'd have expected that `babel.Locale.parse` might return a new instance every time. If you see the caching working, all good.

tristanlabelle (approved):
Thanks!

PR #9306, opened 2026-07-22
Backend: Split shared conversation types into a leaf proto file

nabramow:
**It's a lot of files but mostly all import path changes!!! Purely mechanical: no fields, semantics, or behavior change.**

*Moves `Message` (+ its `MessageContent*` submessages) and `HostRequestStatus` from `conversations.proto` into a new `messages.proto`, keeping the same proto package. All type names and the wire format are completely unchanged.

**Why**: Preparation for public trips. `requests.proto` imports `conversations.proto` for exactly these types, so `conversations.proto` can't reference anything from `requests.proto` back (protobuf forbids circular imports). Review feedback on #9219 from @tristanlabelle asked the new unified message-threads API to reference the existing `HostRequest` message instead of duplicating its fields. This split is what makes that possible.

* Follows protobuf's guidance that widely used types belong in their own dependency-free files: https://protobuf.dev/best-practices/dos-donts/, see "Do Define Message Types in Separate Files"
* Most importantly, shortens #9219 (which is quite large): lets it delete the duplicated `HostRequestThread` message and its separate builder, replacing them with the real `HostRequest` plus one new `unseen_message_count` field.

## Testing
- `make protos`
- Run backend locally with no errors
- Run app locally and click around app messages, dashboard, etc. Everything should work as before

**Backend checklist**

## For maintainers


app/proto/messages.proto
```diff
@@ -0,0 +1,67 @@
+syntax = "proto3";
```
aapeliv, line 1: let's call it messages.proto then?

nabramow: @aapeliv the file you mean? Even though it also has host request status in there?

tristanlabelle (dismissed):

PR #9310, opened 2026-07-22
Don't send community event invites to users already attending

nabramow:
Closes Couchers-org/bugs#99. 

When an editor approves a community invite for an event, the notification fan-out (`generate_event_create_notifications`) notified every visible community member, including people who had already RSVP'd and the event's organizers. 

Now attendees of the occurrence and organizers of the event are excluded; subscribers (followers who haven't RSVP'd) still receive the invite. Added a test covering the reported scenario (RSVP before the invite is approved).

## Testing
- Kinda hard to test as we need to trigger the however many days before notif, so mostly code review and tests passing

**Backend checklist**

## For maintainers


app/backend/src/couchers/servicers/events.py
```diff
@@ -278,7 +278,16 @@ def generate_event_create_notifications(payload: jobs_pb2.GenerateEventCreateNot
             logger.error(f"Inviting user {payload.inviting_user_id} is gone while trying to send event notification?")
             return
 
+        # people already attending or organizing the event don't need an invite to it
```
aapeliv: let's move this into `get_users_to_notify_for_new_event`; and do it in that single SQL query; instead of adding two more

aapeliv (changes requested):
i think when we made this, we were thinking it would be helpful for the organizer to see what the email looked like

nabramow:
> i think when we made this, we were thinking it would be helpful for the organizer to see what the email looked like

Does that mean you want me to remove the organizer from this PR?

app/backend/src/couchers/servicers/events.py
nabramow: done!

aapeliv (approved):
lgtm; thanks for fixing up the query!

PR #9313, opened 2026-07-23
Frontend/events: Display using local times

### LINKED ISSUE #7998: Frontend/i18n: Show event timezone

tristanlabelle:
Switches event times to display in the timezone at the event's location, as now reported by the backend after #9218 . This is more useful than seeing an event in your travel destination displayed in the middle of the night because you're still several timezones away.

⚠️ Existing events that were created with a browser timezone different from the timezone at the event location have incorrect timestamps in the database: they appeared correct in the original browser timezone (two wrongs make one right) and incorrect everywhere else. Now they'll be incorrect everywhere, so authors will need to fix the event times. There's nothing we can do about it since we've stored bad data in the database.

Follow-up question: Should we show timezones now?

Fixes #7998

## Testing
Using NEXT, created an event in Addis Ababa at 2pm. This is how it shows up on NEXT (my browser is in Eastern US timezone):

image

With the Vercel preview:

image
image
image

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -97,13 +97,16 @@ class LocalizationContext:
         self,
         value: datetime | Timestamp,
         *,
+        timezone: str | tzinfo | None = None,
         abbrev: bool = False,
         with_year: bool = True,
         with_day_of_week: bool = False,
         with_seconds: bool = False,
     ) -> str:
         return localize_datetime(
-            to_timezone(value, self.timezone),
+            # By default we display the datetime in the user's timezone.
+            # The "timezone" parameter overrides this behavior.
+            to_timezone(value, timezone or self.timezone),
             self.babel_locale,
             abbrev=abbrev,
             with_year=with_year,
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -97,7 +97,7 @@ class LocalizationContext:
         self,
         value: datetime | Timestamp,
         *,
-        timezone: str | tzinfo | None = None,
+        display_timezone: tzinfo | None = None,
         abbrev: bool = False,
         with_year: bool = True,
         with_day_of_week: bool = False,
@@ -106,7 +106,7 @@ class LocalizationContext:
         return localize_datetime(
             # By default we display the datetime in the user's timezone.
             # The "timezone" parameter overrides this behavior.
-            to_timezone(value, timezone or self.timezone),
+            to_timezone(value, display_timezone or self.timezone),
             self.babel_locale,
             abbrev=abbrev,
             with_year=with_year,
--- a/app/backend/src/couchers/i18n/context.py
+++ b/app/backend/src/couchers/i18n/context.py
@@ -103,6 +103,16 @@ class LocalizationContext:
         with_day_of_week: bool = False,
         with_seconds: bool = False,
     ) -> str:
+        """
+        Formats a date and time according to this localization context.
+
+        Params:
+          value: An instant in time to be formatted in date and time components.
+                 datetimes should be timezone-aware so they represent an unambiguous instant,
+                 but their timezones are irrelevant for formatting.
+          display_timezone: The timezone to use when formatting the datetime.
+                            Defaults to the value from this localization context.
+        """
         return localize_datetime(
             # By default we display the datetime in the user's timezone.
             # The "timezone" parameter overrides this behavior.
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -111,19 +111,15 @@ def date_to_api(date_obj: date) -> str:
     return date_obj.isoformat()
 
 
-def to_aware_datetime(ts: Timestamp, timezone: tzinfo | str = UTC) -> datetime:
+def to_aware_datetime(ts: Timestamp) -> datetime:
     """
     Turns a protobuf Timestamp object into a timezone-aware datetime
     """
-    if isinstance(timezone, str):
-        timezone = ZoneInfo(timezone)
-    return ts.ToDatetime(tzinfo=timezone)
+    return ts.ToDatetime(tzinfo=UTC)
 
 
-def to_timezone(value: Timestamp | datetime, timezone: tzinfo | str) -> datetime:
+def to_timezone(value: Timestamp | datetime, timezone: tzinfo) -> datetime:
     """Returns an instant in time as a datetime in a given timezone."""
-    if isinstance(timezone, str):
-        timezone = ZoneInfo(timezone)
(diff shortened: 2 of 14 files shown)


app/backend/src/couchers/i18n/context.py
```diff
@@ -97,13 +97,16 @@ def localize_datetime(
...
         abbrev: bool = False,
         with_year: bool = True,
         with_day_of_week: bool = False,
         with_seconds: bool = False,
     ) -> str:
         return localize_datetime(
-            to_timezone(value, self.timezone),
+            # By default we display the datetime in the user's timezone.
+            # The "timezone" parameter overrides this behavior.
+            to_timezone(value, timezone or self.timezone),
```
aapeliv: i'm not a fan of overloading with mixed semantics. do you think it would make more sense to just have an explicit function `localize_datetime_in_timezone` or whatever? i just prefer things to be as clear as possible for maintainability.

app/backend/src/couchers/i18n/context.py
```diff
@@ -97,13 +97,16 @@ def localize_datetime(
         self,
         value: datetime | Timestamp,
         *,
+        timezone: str | tzinfo | None = None,
```
aapeliv: `display_timezone`? also do we need all these overloads? does it make it easier to coerce to tzinfo at callsite?

app/backend/src/couchers/utils.py
```diff
@@ -111,15 +111,19 @@ def date_to_api(date_obj: date) -> str:
...
 
-def to_aware_datetime(ts: Timestamp) -> datetime:
+def to_aware_datetime(ts: Timestamp, timezone: tzinfo | str = UTC) -> datetime:
     """
     Turns a protobuf Timestamp object into a timezone-aware datetime
     """
-    return ts.ToDatetime(tzinfo=UTC)
+    if isinstance(timezone, str):
+        timezone = ZoneInfo(timezone)
+    return ts.ToDatetime(tzinfo=timezone)
```
aapeliv: see note in other PR

app/backend/src/couchers/utils.py
```diff
@@ -111,15 +111,19 @@ def date_to_api(date_obj: date) -> str:
...
 
 
-def to_timezone(value: Timestamp | datetime, timezone: tzinfo) -> datetime:
+def to_timezone(value: Timestamp | datetime, timezone: tzinfo | str) -> datetime:
     """Returns an instant in time as a datetime in a given timezone."""
+    if isinstance(timezone, str):
+        timezone = ZoneInfo(timezone)
```
aapeliv: ditto

aapeliv:
> ⚠️ Existing events that were created with a browser timezone different from the timezone at the event location have incorrect timestamps in the database: they appeared correct in the original browser timezone (two wrongs make one right) and incorrect everywhere else. Now they'll be incorrect everywhere, so authors will need to fix the event times. There's nothing we can do about it since we've stored bad data in the database.

Is this new behavior? Sorry, I know we discussed earlier but I keep forgetting the details on this. My understanding is that historically we simply had a datetime picker that would select an instant of time. It's probably unintuitive but iiuc nothing here is more broken than before?

> Follow-up question: Should we show timezones now?

Yes, I think that would be important.

app/backend/src/couchers/i18n/context.py
tristanlabelle: I just get annoyed at having to import ZoneInfo everywhere since we usually have the string to work with because of the db or protobufs. I quite like that python and typescript can overload through union types.

app/backend/src/couchers/i18n/context.py
tristanlabelle: To me having it in the function name or as a named parameter conveys the same, with the latter avoiding duplicating 20 lines.

aapeliv: I think in my mind; "localization" is making the platform familiar for a user in a given locale. So using their formatting rules, timezone, language, etc.

"localize_datetime" to me means "convert this time instant into a string that makes sense to the user". Mixing that in with "convert this time instant into an unrelated timezone for presentation" is somewhat different functionality (though of course very related, and they are coupled, since now you want to communicate not simply an instant of time but rather a local time in a certain timezone). I feel timezones are actually quite confusing once you start converting between different concepts of time (instants, local times, etc; even to me, and I've done a lot of dev stuff on timezones and time); so being very explicit is to me helpful and keeps the code more maintainable.

All that said, I defer to you on how you feel is the best way to do it: I'm not that strongly opinionated and I trust you on the localization stuff. Maybe adding docstring or elaborating in the comment here as to when I would want to pass in an explicit timezone may be useful?

tristanlabelle:
@aapeliv 

Removed `to_aware_datetime`'s `timezone` parameter, renamed the `localize_datetime` param to `display_timezone`, and de-overloaded a bit (grudgingly!). I'd rather not split up `localize_datetime` though. Ready for re-review.

> Is this new behavior? My understanding is that historically we simply had a datetime picker that would select an instant of time.

Correct but the datetime was converted to an instant using the browser local timezone since we didn't know the event location timezone, so all existing events that were created from a browser whose timezone mismatches that at the event location have invalid start/end timestamps stored in the db, nothing we can do about it. They rendered correctly if you were still in the same browser timezone, but that's just because two wrongs made one right -- the times would be invalid in any other timezone, including that at the event location.

> Yes, I think that would be important.

I asked in yesterday's meeting before you joined and we were all agreed that we should not show the timezone because the using the event-local timezone is what you'd expect, it's what other websites do, and timezones are a bit technical to show everywhere.

app/backend/src/couchers/i18n/context.py
tristanlabelle: I removed the overloads in the latest. Picking my fights :P

app/backend/src/couchers/utils.py
tristanlabelle: Removed timezone param in latest

app/backend/src/couchers/utils.py
tristanlabelle: I removed the overloads in the latest. Picking my fights :P

app/backend/src/couchers/i18n/context.py
tristanlabelle: Interesting I was writing pretty much the same rationale in a reply but it ended with the other conclusion. I took your `display_timezone` param name and will add some docs. I think that'll be unambiguous.

aapeliv: sounds good, thank you!

aapeliv (dismissed):
lgtm!

aapeliv (approved):

PR #9315, opened 2026-07-23
Communities: adopt new user card in shared UserSummary

### LINKED ISSUE #5840: Replace Community Members List card with card from map search
Replace the community MembersList user card with the user card we use for map search. This gives more information for the user to help with browsing.

Acceptance Criteria
- Remove UserDetails from the Community Members List and use the user card from the map instead
- Since we're using it in multiple places, consider moving it to a more general folder and rename `UserDetailsCard` instead (update import and name for map as well)
- Add pagination - for now has to be infinite loading pagination until we adjust backend

kevinortiz43:
Removes age from the shared user summary, adds a location pin before the city, and makes the
name and city truncate with a CSS ellipsis.

This updates the existing shared `UserSummary` / `UsersList` components rather than adding a
new bespoke card, so it lands consistently everywhere a user summary is rendered (friends,
community members, moderators, event attendees, event organisers, badge lists, blocked
users, references, mod page, host requests).

Fixes #5840

**What changed**
- **Age removed** from `UserSummary`, and from the bespoke mobile host-request header.
- **Location pin** added before the city, per the mockup.
- **Strong verification badge** still shown wherever it was, and newly **added** to the
  mobile host-request header, which never had it.
- **Width-based truncation** for the name and city, added with `text-overflow: ellipsis`.
- **"Open profile in new tab" removed** from user cards to match the mockup. Also removed
  the open-in-new-tab function since the button has been removed too, so profile links now
  open in the same tab on desktop.
- **Removed locale key** `profile:open_profile_new_tab`, since it is not used anymore.

**Future TODO**
- Search result cards still show age and gender (`SeachResultUserCard.tsx`) — they're a
  bespoke card outside the shared component.
- Emails still render age (`backend/templates/v2/blocks.mjml`) — separate PR with Tristan.

**Not included, per the mockup discussion on #5840**
- Join date, gender, last active and infinite-scroll pagination were dropped per the mockup
  agreed with the team in the #5840 discussion.

## Before

Members list before, desktop

Members list before

## After
image

image

image

image

## Testing

`yarn format`, `yarn tsc --noEmit` and `yarn test` pass.

- Added two `UserSummary` tests: age is not rendered in the heading, and the location pin
  renders alongside the city.
- Updated three `CommunityInfoPage` assertions that matched the old `"Name, age"` string;
  they now query the heading by role.
- Truncation and overflow can't be covered by unit tests, so I looked through every changed
  page at mobile, tablet and desktop screen sizes instead.

**Review feedback (@nabramow)**
- Removed the two UI-presence tests from `UserSummary.test.tsx`; the behavioural menu tests remain.
- `UsersList` now renders skeletons instead of a spinner while user IDs load — this applies to
  every `UsersList` surface (members, moderators, friends, event users), not just the members list.
- Avatar, location pin, and name/city text are smaller on mobile via `theme.breakpoints.up("md")`:
  avatar 72→56px, pin 20→16px, name 19.2→16px, city 16→14px. Desktop unchanged.

**Web frontend checklist**

## For maintainers


nabramow:
@kevinortiz43 Infinite Scroll was not a feature decided nor a ticket to be done, I'm wondering why this was done? Features go through a process of being discussed and prioritized.

kevinortiz43:
@nabramow understood the grid and infinite scroll was me, I will revert it back.

kevinortiz43:
@nabramow this is ready for your review

app/web/features/communities/members/CommunityMembersList.tsx
```diff
@@ -39,6 +47,14 @@ export default function CommunityMembersList({
...
 
   const currentPage = data?.pages && data.pages[pageNumber - 1];
+  const currentPageUserIds = currentPage?.memberUserIdsList ?? [];
+
+  const { data: membersById, isLoading: isLoadingMembers } =
+    useLiteUsers(currentPageUserIds);
+
+  const members = currentPageUserIds
+    .map((userId) => membersById?.get(userId))
+    .filter((user): user is NonNullable<typeof user> => !!user);
```
nabramow: Why are we adding a query here? I thought this was only an aesthetic design change?

kevinortiz43: The useLiteUsers query isn't new it was already running inside UsersList, which the old code used. UsersList does two things: hydrates the userIds via useLiteUsers, then renders each as a UserSummary

This PR needed a different card (MemberCard  new layout, drops age/gender/last-active, keeps name/location/verification). Since UsersList can only render UserSummary, we stopped using it and render MemberCard ourselves. Which means the parent now does the hydration UsersList used to do internally just lifted up out of the wrapper.

Old (query hidden inside UsersList):
```tsx
const { data: users } = useLiteUsers(userIds);
const foundUsers = userIds?.map((id) => users?.get(id)).filter((u) => !!u);
// → renders <UserSummary user={...} /> per user
```
New (same query, now in the parent so we can render MemberCard):
```tsx
const { data: membersById } = useLiteUsers(currentPageUserIds);
const members = currentPageUserIds
  .map((userId) => membersById?.get(userId))
  .filter((user): user is NonNullable<typeof user> => !!user);
// → renders <MemberCard user={member} /> per member
```

nabramow: Ah that makes more sense. I do wonder if we should just change the design in UsersList though? We can discuss in that thread but IMO this new card change probably makes sense most places. We can do a quick review where this renders first.

kevinortiz43: @nabramow no problem I can get a list of where else its rendering, that will be a new PR or want me to do it in this PR?

nabramow: Either way. If it's easier to start fresh go for that, otherwise in here. Maybe do a quick post in that thread on slack if anyone has an issue changing this everywhere with the list of where it will get replaced.

nabramow: I think it should be fine everywhere though its not a huge change.

kevinortiz43: Okay I asked in slack, waiting to hear back

kevinortiz43:
@nabramow ready for your review

app/web/components/UserSummary.test.tsx
```diff
@@ -34,4 +34,19 @@ describe("UserSummary", () => {
...
+    expect(heading).toHaveTextContent(users[0].name);
+    expect(heading).not.toHaveTextContent(`${users[0].age}`);
+  });
+
+  it("should render the location pin alongside the city", async () => {
+    render(<UserSummary user={users[0]} />, { wrapper });
+
+    expect(screen.getByTestId("LocationOnIcon")).toBeVisible();
+    expect(screen.getByText(users[0].city)).toBeVisible();
+  });
```
nabramow: You can remove these, tests just testing the UI are a bit brittle and not so helpful. Better to test behavior.

app/web/features/communities/members/CommunityMembersList.tsx
```diff
@@ -76,14 +60,12 @@ export default function CommunityMembersList({
...
-            <MemberCard key={member.userId} user={member} />
-          ))}
-        </MembersList>
-      )}
+      {isLoading && <CenteredSpinner />}
+      <Box sx={{ width: "100%", maxWidth: "450px" }}>
+        {data?.pages && data?.pages.length > 0 && (
+          <UsersList userIds={currentPage?.memberUserIdsList} titleIsLink />
+        )}
+      </Box>
```
nabramow: While you're here, let's change this to a skeleton to indicate loading instead of the spinner. We're trying to change these out as we go as it's a bit more modern.

app/web/features/communities/locales/en.json
```diff
@@ -122,9 +122,6 @@
   "load_more_organizers": "Load more organizers",
   "location": "Location",
   "location_required": "Please provide a location for the event.",
-  "member_card": {
-    "profile_link_a11y_label": "Profile of {{name}}"
-  },
```
nabramow: Why are we removing this a11y label?

nabramow:
Let's make everything a bit smaller on the mobile view, especially the avatar (I think there is already a prop for size), but also the pin icon and text can be smaller on mobile:

Screenshot 2026-08-01 at 10 45 24

app/web/features/communities/locales/en.json
kevinortiz43: The key had no references left once `MemberCard.tsx` was deleted the members list now renders the shared `UserSummary`, and the `Avatar` inside it hasthe same accessible name via [`Avatar.tsx:94`](https://github.com/Couchers-org/couchers/blob/develop/app/web/components/Avatar/Avatar.tsx#L94) → `getProfileLinkA11yLabel(user.name)` → `"Profile of {name}"`.

app/web/features/communities/members/CommunityMembersList.tsx
kevinortiz43: okay I switched out the spinner for the skeleton. let me know if i did it correctly, I delete the tests and shrank the sizes for mobile

app/web/components/UserSummary.tsx
```diff
@@ -29,41 +28,47 @@ const StyledWrapper = styled("div")({
...
+const avatarSize = (theme: Theme, isSmallAvatar: boolean) => ({
+  marginInlineEnd: theme.spacing(2),
+  height: isSmallAvatar ? "2.25rem" : "3.5rem",
+  width: isSmallAvatar ? "2.25rem" : "3.5rem",
+  [theme.breakpoints.up("md")]: {
+    height: isSmallAvatar ? "3rem" : "4.5rem",
+    width: isSmallAvatar ? "3rem" : "4.5rem",
+  },
+});
+
```
nabramow: You shouldn't need `isSmallAvatar` since if you're using `theme.breakpoints`. I think the issue is your theme.breakpoints is `up`, this means if it's bigger than medium screen and the default is also bigger screen. Switch it to be styles for `theme.breakpoints.down("md")`, this means if it's a screen smaller than medium, then just put the style directly and remove `isSmallAvatar`.

You can search the codebase for some more examples of when we use `theme.breakpoints.down{"md")` to get a feel for it.

app/web/components/UserSummary.tsx
```diff
@@ -127,20 +126,42 @@ export default function UserSummary({
         component={headlineComponentWithRef}
         variant="h2"
         noWrap={nameOnly}
-        sx={{ marginTop: "auto", fontSize: "1.2rem" }}
+        sx={(theme) => ({
+          marginTop: "auto",
+          fontSize: "1rem",
+          minWidth: 0,
+          [theme.breakpoints.up("md")]: { fontSize: "1.2rem" },
```
nabramow: I'm not sure this will be doing anything as the default is the bigger screen. You want to `theme.breakpoints.down("md")` to handle smaller screen sizes and let the default be the bigger screen. I expect to only see the down values for mobile here, otherwise we don't really want an h2?

app/web/components/UserSummary.tsx
```diff
@@ -191,13 +211,37 @@ export default function UserSummary({
...
                 >
-                  {!user ? <Skeleton /> : cityValue}
-                </Typography>
+                  {user && cityValue && (
+                    <PinIcon
+                      sx={(theme) => ({
+                        flexShrink: 0,
+                        color: "var(--mui-palette-text-secondary)",
+                        fontSize: "1rem",
+                        [theme.breakpoints.up("md")]: { fontSize: "1.25rem" },
```
nabramow: Same here this is saying "for bigger screens do this" but the default should also be for bigger screens and we want to customize for smaller screens. If you need to customize for bigger screens it means the default style you set is wrong.

app/web/components/UserSummary.tsx
```diff
@@ -191,13 +211,37 @@ export default function UserSummary({
...
+                    />
+                  )}
+                  <Typography
+                    color="textSecondary"
+                    variant="body1"
+                    noWrap
+                    sx={(theme) => ({
+                      minWidth: 0,
+                      fontSize: "0.875rem",
+                      [theme.breakpoints.up("md")]: { fontSize: "1rem" },
```
nabramow: same here

app/web/components/UserSummary.tsx
kevinortiz43: Okay I have done this  is switched to .down("md")

app/web/components/UserSummary.tsx
kevinortiz43: okay I went ahead and removed it and use .down("md") instead.

app/web/components/UserSummary.tsx
kevinortiz43: Switched to .down("md")

app/web/components/UserSummary.tsx
kevinortiz43: Switched to .down("md")

app/web/components/UserSummary.tsx
kevinortiz43: okay looked at isSmallAvatar a bit more and removed it as you have mentioned

app/web/components/UserSummary.tsx
```diff
@@ -140,15 +154,14 @@ export default function UserSummary({
             <ProfileLink
               userId={"userId" in user ? user.userId : undefined}
               username={user.username}
-              openInNewTab={!isMobile}
```
nabramow, line 143: Why remove this?

app/web/components/UserSummary.tsx
```diff
@@ -140,15 +154,14 @@ export default function UserSummary({
...
               style={{
                 display: "flex",
                 alignItems: "center",
                 justifyContent: "flex-start",
+                minWidth: 0,
               }}
             >
               {title}
-              {!isMobile && <StyledOpenInNewIcon />}
```
nabramow, line 151: Same here I don't think this was part of this ticket?

app/web/features/profile/locales/en.json
```diff
@@ -285,7 +285,6 @@
     "no_people": "There are no people with that badge!",
     "not_found": "Badge not found"
   },
-  "open_profile_new_tab": "Open profile in new tab",
```
nabramow, line 288: This should go back as well if the others do.

nabramow:
Looks okay I just didn't expect to see the open in new tab behavior removed. Could we please add that back, that's quite important for browsing on desktop.

kevinortiz43:
@nabramow the new tab behavior was removed because the new tab button was not part of the new frontend design. But I can easily add it back.

kevinortiz43:
@nabramow I added the "open profile in new tab" behavior

nabramow (approved):
Okay this is good to go now. I pushed some changes to add back more open in new tab logic that had been removed.

PR #9320, opened 2026-07-23
Backend: Remove str-based EventUpdate items

### LINKED ISSUE #9117: Backend/events: Remove str-based EventUpdated items
After deploying and letting the job queue has drain from messages referencing them.

tristanlabelle:
Removes the deprecated `updated_str_items` field from both the `EventUpdate` notification data message and the `GenerateEventUpdateNotificationsPayload` internal job payload, reserving the field numbers and names. The i18n-friendly `updated_enum_items` field is now the sole source.

(a month has passed since I migrated us over so it should be safe that there are no old queued jobs using `updated_str_items` anymore.)

Fixes #9117

### DIFF
--- a/app/proto/notification_data.proto
+++ b/app/proto/notification_data.proto
@@ -146,7 +146,6 @@ message EventUpdate {
   org.couchers.api.events.Event event = 1;
   org.couchers.api.core.User updating_user = 2;
   reserved 3; // Formerly "updated_str_items", was not i18n-friendly. Use updated_enum_items.
-  reserved "updated_str_items";
   repeated EventUpdateItem updated_enum_items = 4;
 }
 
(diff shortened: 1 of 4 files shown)


app/proto/notification_data.proto
```diff
@@ -145,8 +145,7 @@ enum EventUpdateItem {
 message EventUpdate {
   org.couchers.api.events.Event event = 1;
   org.couchers.api.core.User updating_user = 2;
-  // TODO(#9117): Remove once unused. Was not i18n-friendly. Use updated_enum_items.
-  repeated string updated_str_items = 3 [deprecated = true];
+  reserved 3; // Formerly "updated_str_items", was not i18n-friendly. Use updated_enum_items.
```
aapeliv, line 148: i believe the notif data is used in the dropdown menu?

tristanlabelle: The dropdown menu uses the textual form rendered as a push notification by the backend, so notification data is private to the backend.

https://github.com/Couchers-org/couchers/blob/46ff3ab50ae37ce0a2190cc61322bf81b8513cf5/app/proto/notifications.proto#L94:L106

aapeliv (approved):

PR #9326, opened 2026-07-24
Backend/events: Add GetEventCalendarFile API

### LINKED ISSUE #9328: Backend/events: Implement API for getting the event calendar file
Prerequisite to #9227 which will add the download UI on the frontend.

tristanlabelle:
This API will support a button on the frontend to download a calendar file for the event.

Includes some refactoring of the ics generation code and tests, the biggest visible result being that we now create a sequence number based on a timestamp to support an arbitrary number of updates.

Fixes #9328

## Testing
Added & updated tests.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -111,13 +111,11 @@ def date_to_api(date_obj: date) -> str:
     return date_obj.isoformat()
 
 
-def to_aware_datetime(ts: Timestamp, timezone: str | tzinfo = UTC) -> datetime:
+def to_aware_datetime(ts: Timestamp) -> datetime:
     """
     Turns a protobuf Timestamp object into a timezone-aware datetime
     """
-    if isinstance(timezone, str):
-        timezone = ZoneInfo(timezone)
-    return ts.ToDatetime(tzinfo=timezone)
+    return ts.ToDatetime(tzinfo=UTC)
 
 
 def to_timezone(value: Timestamp | datetime, timezone: tzinfo) -> datetime:
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -111,11 +111,13 @@ def date_to_api(date_obj: date) -> str:
     return date_obj.isoformat()
 
 
-def to_aware_datetime(ts: Timestamp) -> datetime:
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/utils.py
```diff
@@ -111,11 +111,13 @@ def date_to_api(date_obj: date) -> str:
...
 
-def to_aware_datetime(ts: Timestamp) -> datetime:
+def to_aware_datetime(ts: Timestamp, timezone: str | tzinfo = UTC) -> datetime:
     """
     Turns a protobuf Timestamp object into a timezone-aware datetime
     """
-    return ts.ToDatetime(tzinfo=UTC)
+    if isinstance(timezone, str):
+        timezone = ZoneInfo(timezone)
+    return ts.ToDatetime(tzinfo=timezone)
```
aapeliv: let's keep this function as a clean "protobuf object -> instant of time" function. we can do the conversion elsewhere or have another helper, but it is non-intuitive to many (including myself) what it means to pass in a timzone into this function; as this function is converting "naive" timestamps into "aware" ones. so let's keep it dumb and do the presentation semantics elsewhere.

app/proto/events.proto
```diff
@@ -80,7 +80,9 @@ service Events {
     // Lists all events
   }
 
-  // TODO: DownloadCalendarInvite
+  rpc GetEventCalendarFile(GetEventCalendarFileReq) returns (GetEventCalendarFileRes) {
+    // Downloads a calendar file for an event
```
aapeliv, line 86: if you want to eventually use this as a pure HTTP JSON endpoint, you can add transcoding and use the http_body stuff (see GIS servicer)

aapeliv (changes requested):
lgtm but let's keep the `to_aware_datetime` function simple

app/backend/src/couchers/utils.py
tristanlabelle: Reverted to_aware_datetime timezone param in latest

app/proto/events.proto
tristanlabelle: Switched to HTTP endpoint (returns the file directly for download) in latest

tristanlabelle:
@aapeliv ready for final stamp

aapeliv (approved):
lgtm, thanks!

PR #9329, opened 2026-07-25
Backend: fix event list pagination skipping and duplicating events

aapeliv:
All six event listing endpoints (`ListEventOccurrences`, `ListMyEvents`, `ListAllEvents`, community and group `ListEvents`, `EventSearch`) ordered occurrences by `start_time` but paginated with an `end_time`-based cursor. Whenever event durations vary — e.g. a multi-day event overlapping shorter ones — the two orders disagree: the long event was repeated on every page and shorter events that started later but ended earlier were skipped entirely.

The fix paginates on the same key the lists are sorted by:
- A shared `apply_occurrence_pagination` helper applies a fixed upcoming/past window (`end_time` vs `now()`, so ongoing events still count as upcoming) plus an exact `(start_time, id)` cursor, ordered by `(start_time, id)`. The id tiebreaker makes pagination exact when start times tie (events commonly start on the hour), removing the old ±1 second fudges.
- Page tokens now encode the cursor encrypted with the existing page token helpers (like `ListCommunities`), via new `dt_id_to_page_token`/`dt_id_from_page_token` utils.
- The `page_number` offset paths in `ListMyEvents` and `EventSearch` are unchanged.

## Testing

- New regression tests: `test_ListMyEvents_pagination_overlapping_durations` (a multi-day event overlapping four short ones — the old cursor repeated the long event and dropped the short ones) and `test_ListMyEvents_pagination_identical_start_times` (five events with the same start time — the old cursor looped). Both verified to fail on the old code.
- `test_event_search_pagination` previously asserted the token's plaintext contents and hand-built a millis token as a fake past cutoff; rewritten to walk pages via returned tokens, including past pagination over backdated occurrences.
- `pytest src/tests/test_events.py src/tests/test_search.py src/tests/test_communities.py src/tests/test_groups.py` passes (102 tests), `make mypy` and `make format` clean.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/events.py
+++ b/app/backend/src/couchers/servicers/events.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Sequence
 from datetime import datetime, timedelta
 from typing import Any, cast
 from zoneinfo import ZoneInfo
@@ -9,7 +10,7 @@ from google.protobuf import empty_pb2
 from psycopg.types.range import TimestamptzRange
 from sqlalchemy import Select, func, select
 from sqlalchemy.orm import Session
-from sqlalchemy.sql import and_, func, or_, update
+from sqlalchemy.sql import and_, func, or_, tuple_, update
 
 from couchers.context import CouchersContext, make_notification_user_context
 from couchers.db import can_moderate_node, get_parent_node_at_location, session_scope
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/servicers/events.py
```diff
@@ -288,6 +289,37 @@ def _check_occurrence_time_validity(start_time: datetime, end_time: datetime, co
         context.abort_with_error_code(grpc.StatusCode.INVALID_ARGUMENT, "event_too_far_in_future")
 
 
+def apply_occurrence_pagination(
```
tristanlabelle, line 292: next_page_token + apply_pagination is a good pattern, we should use it everywhere.

tristanlabelle (approved):

PR #9338, opened 2026-07-26
Fix native app client platform attribution and User-Agent

aapeliv:
API calls made from web views inside the native app were being recorded as `web_mobile` in `api_calls` / `user_activity`, e.g.:

```
Mozilla/5.0 (iPhone; CPU iPhone OS 18_7 like Mac OS X) ... CouchersNative/1.3.0 (ios; build 81; unknown) | web_mobile
```

Two separate bugs, both visible in that one row:

**Attribution.** 9614c4b97 dropped the `nativePlatform` WebView bridge, leaving the web bundle to pick `web_desktop`/`web_mobile` by viewport. Since the app is almost entirely WebView-driven, only the small amount of traffic from the native gRPC client ever declared `app_ios`/`app_android`. The web bundle now reads the platform off our own `CouchersNative/{version} ({os}; ...)` user agent token, which every shipped app build already sends — so this takes effect on a web deploy, with no app release needed and no bridge to keep in sync.

**User-Agent.** The last field read `Constants.expoConfig.extra.gitHash`, which `app.config.js` never sets (it sets `displayVersion` / `debugVersion`), so it has been hardcoded to `unknown` since it shipped. It now carries the running bundle's `debugVersion`, which includes the commit and is re-baked per OTA. The version and build number now both come from `expo-application`, so they describe the installed binary — `Constants.expoConfig.version` is the *running* manifest's idea of the store version, so an OTA built from a commit that bumped the version would claim a version the binary isn't.

The new shape:

```
CouchersNative/1.3.0 (ios; build 81; v1.2.18410.1156180a.20260528Z0533)
```

Note the UA change only takes effect for new bundles (OTA or store); existing installs keep sending `unknown` until they update. The attribution fix applies retroactively to all installs. Historical rows stay misattributed but are recoverable via `user_agent LIKE '%CouchersNative/%'` — no backfill here, since that rewrites analytics history.

## Testing

- Added `app/web/utils/clientPlatform.test.ts`, covering the real iOS and Android WebView user agents plus mobile and desktop browser UAs (4/4 pass).
- Mobile suite passes (155/155), `tsc --noEmit` clean, eslint and prettier clean on touched files. Updated the `expo-application` jest mock with `nativeApplicationVersion` and replaced the stale `gitHash` in the `expo-constants` mock.
- 9 web suites fail and `service/events.ts` has `tsc` errors on `develop` already (stale generated protos, `proto/messages_pb` unresolved) — confirmed pre-existing by stashing.

To verify manually: run the app against a local backend, click around a WebView screen, and check that new `api_calls` rows show `app_ios`/`app_android` rather than `web_mobile`, with the git hash and timestamp in place of `unknown`.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/mobile/fingerprint.config.js
+++ b/app/mobile/fingerprint.config.js
@@ -2,7 +2,7 @@ module.exports = {
   sourceSkips: [
     // no native-affecting build scripts, so an npm script shouldn't rebuild
     "PackageJsonScriptsAll",
-    // app.config.js injects a per-commit gitHash into `extra`; nothing native
+    // app.config.js injects per-commit versions into `extra`; nothing native
     "ExpoConfigExtraSection",
     // a version/build-number bump shouldn't break OTA continuity
     "ExpoConfigVersions",
--- a/app/mobile/jest.setup.ts
+++ b/app/mobile/jest.setup.ts
@@ -17,13 +17,15 @@ jest.mock("expo-constants", () => ({
     version: "1.0.0-test",
     extra: {
       eas: { projectId: "test-project-id" },
-      gitHash: "abc12345",
+      displayVersion: "v1.2.3456",
+      debugVersion: "v1.2.3456.abc12345.20260528Z0533",
     },
   },
 }));
 
 // Mock expo-application globally
 jest.mock("expo-application", () => ({
+  nativeApplicationVersion: "1.0.0-test",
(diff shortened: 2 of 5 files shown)


nabramow (approved):
Seems okay to me. At least it won't break anything so we can merge and see.

Issue #8854, opened 2026-05-29
Backend/tests: Support mocking the system clock


tristanlabelle:
Currently we have a lot of `today() + timedelta(days=2)` so the times are within the acceptable range. This prevents writing assertions about the stringified value of dates and so on. It would be better to mock the system clock, which `freezegun` module supports: https://pypi.org/project/freezegun/

WouldYouKindly:
@tristanlabelle in my experience, [time-machine](https://github.com/adamchainz/time-machine) is a better choice. It's faster, and covers more cases. There is a blog post by the author that explains it rather well: https://adamj.eu/tech/2020/06/03/introducing-time-machine/

tristanlabelle:
Thanks! I defer to your experience as I have none.

tristanlabelle:
Fixed in #9354

PR #9339, opened 2026-07-26
Stop duplicate host requests from repeated Send taps

aapeliv:
One user created 31 host requests to the same host in 14 seconds from the iOS webview. The form still held ≥250 characters throughout (on success it calls `reset()`), so every tap landed before the client processed a single response — it was sending fine and not receiving, so the user kept mashing Send. The server created all 31 and logged them as clean successes.

**Web** — the Send button wasn't disabled while the mutation was in flight; it now uses `loading={isPending}`, like `OfferToHostDialog` already does. `Button`'s internal `waiting` state didn't cover this: it tracks the `onClick` promise, which resolves when `mutate()` is dispatched, not when it completes. The submit handler also bails on `isPending`, covering Enter from the date fields.

**Backend** — `CreateHostRequest` rejects a second request to the same recipient within 3 hours if the stays overlap, pointing the user at their messages. This holds regardless of the client, including app builds we can't force-update. Overlap is on nights, so consecutive stays are fine. Public-trip offers are exempt; they already have a per-trip check.

## Testing

Backend suite (1223 passed), `make mypy`, `make format` clean. Web suite (912 passed), `yarn lint`, `tsc` clean.

New backend tests cover overlapping and partly overlapping (blocked), non-overlapping and back-to-back (allowed), after the window, a different host, and public-trip offers. 20 existing tests create overlapping requests between the same pair; `backdate_conversations()` in `tests/fixtures/db.py` shifts conversations past the window, preserving relative order for the listing tests.

New web test: six taps with the response withheld → one API call. It needs `pointerEventsCheck: Never` (user-event otherwise refuses the `pointer-events: none` on the disabled button, which is the state under test) and the `await act(...)` between taps, without which `Button`'s transient `waiting` state swallows taps 2-6 and it passes on `develop` too.

To reproduce: throttle the network hard in devtools and tap Send repeatedly.

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/web/features/profile/view/NewHostRequest.test.tsx
```diff
@@ -185,4 +191,45 @@ describe("NewHostRequest", () => {
...
+    const departureGroup = screen.getByRole("group", {
+      name: t("profile:request_form.departure_date"),
+    });
+    await user.click(departureGroup);
+    await user.keyboard("{Control>}a{/Control}");
+    await user.keyboard("06052026");
+
+    fireEvent.change(screen.getByLabelText(t("profile:request_form.request")), {
+      target: { value: LONG_TEXT },
+    });
```
nabramow: We use userEvent in the codebase not fireEvent. Tell claude to look at the event tests as an example or something and fix it.

app/backend/src/tests/fixtures/db.py
```diff
@@ -354,6 +355,20 @@ def make_user_invisible(user_id: int) -> None:
         session.execute(update(User).where(User.id == user_id).values(banned_at=func.now()))
 
 
+def backdate_conversations() -> None:
```
tristanlabelle, line 358: My kingdom for https://github.com/Couchers-org/couchers/issues/8854

tristanlabelle (approved):

nabramow (approved):
Stamp of approval for the frontend now that the test pattern has been updated ✅

app/backend/src/tests/fixtures/db.py
aapeliv: i will fix this after we do time machine or timewarp

PR #9340, opened 2026-07-26
Backend/notifications: Fix chat missed_messages mark-seen and reset_preference

aapeliv:
Fixes two small notification bugs.

**`chat__missed_messages` was never marked seen.** `MarkLastSeenGroupChat` marked both `chat__message` and `chat__missed_messages` seen using `key=str(group_chat_id)`, but `chat__missed_messages` is a summary across *all* chats and is created with `key=""` (`jobs/handlers.py`), so the update never matched a row. Split into two `mark_notifications_seen` calls with the right key for each. Impact is low today because both `ListNotifications` and the `Ping` unseen count filter the in-app feed to push-enabled topic actions and this one defaults to email-only — but a user who enables push for it would get a notification stuck unread forever.

For the record, the equivalent call in `requests.py` is *not* affected: `host_request__missed_messages` really is created with `key=str(conversation_id)`, so its key matches.

**`reset_preference` filtered on the wrong table.** It used `NotificationDelivery.delivery_type` where it meant `NotificationPreference.delivery_type`, giving an accidental cross join with `notification_delivery`. With that table empty the query returns no rows and the function silently deletes nothing; with multiple matching rows it would raise `MultipleResultsFound`. The function is currently uncalled, so there is no user-visible impact. Also drops the now-unused `NotificationDelivery` import.

**Unrelated drive-by:** three pre-existing mypy errors in `test_events.py` / `test_calendar_events.py` (a wrong `-> str` annotation and two unguarded `re.search(...).group(1)` calls) — `make mypy` did not pass on `develop` without these.

## Testing

Added a regression test for each bug and confirmed both fail against the unfixed code:

- `test_mark_last_seen_clears_missed_messages_notification` (`test_conversations.py`) — creates a `chat__missed_messages` notification, calls `MarkLastSeenGroupChat`, asserts it becomes seen.
- `test_reset_preference` (`test_notifications.py`) — sets push and email overrides for a topic action, resets only push, asserts the push override is gone and the email one survives.

`make format` and `make mypy` pass clean. Ran `test_conversations.py`, `test_requests.py`, `test_notifications.py`, `test_notification_settings.py`, `test_bg_jobs.py`, `test_email.py`, `test_events.py` and `test_calendar_events.py` — all pass.

**Backend checklist**
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history — n/a, no schema changes

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/conversations.py
+++ b/app/backend/src/couchers/servicers/conversations.py
@@ -612,6 +612,7 @@ class Conversations(conversations_pb2_grpc.ConversationsServicer):
             topic_actions=[NotificationTopicAction.chat__message],
         )
         # chat__missed_messages is a summary across all chats, so it's keyed with an empty string
+        # rather than a chat id: reading any chat counts as acting on it, and it gets marked seen
         mark_notifications_seen(
             session,
             user_id=context.user_id,
--- a/app/backend/src/couchers/servicers/conversations.py
+++ b/app/backend/src/couchers/servicers/conversations.py
@@ -609,10 +609,14 @@ class Conversations(conversations_pb2_grpc.ConversationsServicer):
             session,
             user_id=context.user_id,
             key=str(request.group_chat_id),
-            topic_actions=[
-                NotificationTopicAction.chat__message,
-                NotificationTopicAction.chat__missed_messages,
-            ],
+            topic_actions=[NotificationTopicAction.chat__message],
+        )
+        # chat__missed_messages is a summary across all chats, so it's keyed with an empty string
+        mark_notifications_seen(
+            session,
+            user_id=context.user_id,
+            key="",
+            topic_actions=[NotificationTopicAction.chat__missed_messages],
         )
 
         return empty_pb2.Empty()
--- a/app/backend/src/tests/test_notifications.py
+++ b/app/backend/src/tests/test_notifications.py
@@ -8,7 +8,7 @@ from urllib.parse import parse_qs, urlparse
 import grpc
 import pytest
 from google.protobuf import empty_pb2, timestamp_pb2
-from sqlalchemy import func, select, update
+from sqlalchemy import select, update
 
 from couchers.config import config
 from couchers.constants import DATETIME_INFINITY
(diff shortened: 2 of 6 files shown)


app/backend/src/couchers/servicers/conversations.py
```diff
@@ -609,10 +609,14 @@ def MarkLastSeenGroupChat(
             session,
             user_id=context.user_id,
             key=str(request.group_chat_id),
-            topic_actions=[
-                NotificationTopicAction.chat__message,
-                NotificationTopicAction.chat__missed_messages,
-            ],
+            topic_actions=[NotificationTopicAction.chat__message],
+        )
+        # chat__missed_messages is a summary across all chats, so it's keyed with an empty string
```
tristanlabelle, line 614: Maybe call in more explicitly that seeing any message is considered as having taken action on the "missed message" notification, so it can be marked as seen.

app/backend/src/tests/test_notifications.py
```diff
@@ -657,6 +663,36 @@ def test_get_topic_actions_by_delivery_type(db):
...
 
+def test_reset_preference(db):
+    user, token = generate_user()
+
+    topic_action = NotificationTopicAction.event__create_any
+    assert NotificationDeliveryType.push not in topic_action.defaults
+
+    with session_scope() as session:
+        set_preference(session, user.id, topic_action, NotificationDeliveryType.push, True)
+        set_preference(session, user.id, topic_action, NotificationDeliveryType.email, False)
```
tristanlabelle: Any reason to not implement these in terms of the API? I feel like using the DB directly weakens the integration test.

tristanlabelle (dismissed):

app/backend/src/couchers/servicers/conversations.py
aapeliv: i (read claude) elaborated on the comment

app/backend/src/tests/test_notifications.py
aapeliv: updated to use apis now

tristanlabelle (dismissed):
No concerns!

PR #9341, opened 2026-07-26
Backend: delay missed-message emails for push-capable users

### LINKED ISSUE #6234: Group message notifications within a certain time

aapeliv:
If you have the app and someone messages you, you get a push immediately and then a "you missed messages" email 5-8 minutes later. Now that the app exists that's just double-notifying people who saw the push and decided to reply later.

We can't simply turn the email off — `digest` delivery is still a stub (`notifications/background.py:218`), so `chat:missed_messages` is the only email anyone gets about a message. Instead this makes the delay depend on whether we can reach the user by push:

- no active push subscription → 5 minutes, unchanged
- active push subscription → 24 hours

`_message_unseen_long_enough()` uses the same `disabled_at > now()` predicate as `_push_to_user`, so "reachable by push" means what the push sender means by it. The `last_seen_message_id` check still applies, so anyone who opens the chat within the day gets no email at all. Applied to both `send_message_notifications` and `send_request_notifications`. Thresholds are two constants at the top of the handler.

Caveat: an active subscription isn't proof the user sees pushes — revoking notification permission leaves a token APNs still accepts. Those users get the email a day late rather than not at all.

## Testing

`make mypy` clean, `make format` applied. Three new tests in `test_bg_jobs.py` covering the 24h delay for chat and host requests, and that a disabled subscription restores the 5 minute behaviour. The full suite was green before I stripped an earlier `last_active` hunk from this branch, but hasn't been re-run since — letting CI confirm.

To reproduce by hand: register a mobile push subscription for a user, send them a chat message from another account, then run `send_message_notifications` — no email is queued until the message is over a day old.

**Backend checklist**
- [x] Added migrations if there are any database changes — n/a, no schema changes

## For maintainers

### DIFF
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -190,9 +190,9 @@ def _message_unseen_long_enough(user_id_column: InstrumentedAttribute[int]) -> C
         .where(PushNotificationSubscription.disabled_at > func.now())
         .exists()
     )
-    return or_(
-        and_(has_active_push_subscription, Message.time < now() - MISSED_MESSAGES_DELAY_WITH_PUSH),
-        and_(not_(has_active_push_subscription), Message.time < now() - MISSED_MESSAGES_DELAY),
+    return Message.time < case(
+        (has_active_push_subscription, now() - MISSED_MESSAGES_DELAY_WITH_PUSH),
+        else_=now() - MISSED_MESSAGES_DELAY,
     )
 
 
--- a/app/backend/src/couchers/jobs/handlers.py
+++ b/app/backend/src/couchers/jobs/handlers.py
@@ -12,7 +12,7 @@ from typing import Any
 import requests
 from google.protobuf import empty_pb2
 from sqlalchemy import ColumnElement, Float, Function, Integer, select
-from sqlalchemy.orm import aliased
+from sqlalchemy.orm import InstrumentedAttribute, aliased
 from sqlalchemy.sql import (
     and_,
     case,
@@ -170,6 +170,32 @@ def purge_account_deletion_tokens(payload: empty_pb2.Empty) -> None:
         )
 
 
+# how long a message must go unseen before we email the user about it
+MISSED_MESSAGES_DELAY = timedelta(minutes=5)
+# ... unless we could reach them by push, in which case they've already been told about it once
+MISSED_MESSAGES_DELAY_WITH_PUSH = timedelta(hours=24)
+
+
+def _message_unseen_long_enough(user_id_column: InstrumentedAttribute[int]) -> ColumnElement[bool]:
+    """
+    Whether `Message` has gone unseen long enough to email the given user about it.
+
+    Note that a subscription we can still push to is not proof the user sees those pushes: they may
+    have revoked notification permission in their OS settings without the token going stale. Such
+    users still get the email, just a day late.
+    """
+    has_active_push_subscription = (
+        select(PushNotificationSubscription.id)
+        .where(PushNotificationSubscription.user_id == user_id_column)
+        .where(PushNotificationSubscription.disabled_at > func.now())
+        .exists()
+    )
+    return or_(
+        and_(has_active_push_subscription, Message.time < now() - MISSED_MESSAGES_DELAY_WITH_PUSH),
+        and_(not_(has_active_push_subscription), Message.time < now() - MISSED_MESSAGES_DELAY),
+    )
+
+
 def send_message_notifications(payload: empty_pb2.Empty) -> None:
     """
     Sends out email notifications for messages that have been unseen for a long enough time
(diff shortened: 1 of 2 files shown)


tristanlabelle (dismissed):

nabramow:
We can do this in a follow up PR as it's frontend but so I don't forget I think we need to handle these things next:

* Delete app/mobile/app/(tabs)/dashboard.tsx lines 8–17
`setNotificationHandler` is a single global callback, so this overrides the one in `app/_layout.tsx:49`. That means the foreground suppression in _layout (shouldShowBanner: !isViewingTarget) is dead, and we show a banner even when the user is already reading the conversation. This is likely the main reason messages feel over-notified while actively replying.

Two further conflicts: `shouldSetBadge: false` here contradicts `shouldSetBadge: true` in `_layout`, and `shouldShowAlert` is deprecated apparently in SDK 53+ (we're on 54) in favor of shouldShowBanner/shouldShowList. Passing both is contradictory.


* Delete app/mobile/app/(tabs)/dashboard.tsx lines 19–26 
This duplicates `configureAndroidChannel()` in `features/notifications/useRegisterPushNotifications.ts:23`, and the two configs disagree. This one passes sound: null (silent), the other passes vibrationPattern and lightColor with no sound.

That matters because Android locks a channel's importance and sound at creation time; later `setNotificationChannelAsync` calls can't change them, only the user can. So whichever runs first permanently decides whether the channel makes a sound, and on a fresh install that's a race. Suggest deleting this block and keeping the one in useRegisterPushNotifications.ts as the single definition.

* Foreground check for web so no notif if actively on the site

nabramow:
If we also want to handle the web push notifications here:

* Add tag to app/web/public/service-worker.js line 4 
Mobile pushes collapse per conversation via collapseKey = `f"{topic_action}_{key}` in  `send_raw_push_notification.py:132`, but web has no equivalent, so every message in a chat stacks its own OS notification. A 10-message exchange is 10 notifications.

The payload already carries what we need: `send_raw_push_notification.py:83-91` sends topic_action and key, and this handler just ignores them. Passing tag:  + ${data.topic_action}_${data.key} to `showNotification` gives web the same per-conversation collapsing mobile already has.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -170,6 +170,32 @@ def purge_account_deletion_tokens(payload: empty_pb2.Empty) -> None:
...
+    has_active_push_subscription = (
+        select(PushNotificationSubscription.id)
+        .where(PushNotificationSubscription.user_id == user_id_column)
+        .where(PushNotificationSubscription.disabled_at > func.now())
+        .exists()
+    )
+    return or_(
+        and_(has_active_push_subscription, Message.time < now() - MISSED_MESSAGES_DELAY_WITH_PUSH),
+        and_(not_(has_active_push_subscription), Message.time < now() - MISSED_MESSAGES_DELAY),
+    )
```
nabramow, line 196: I think the delay can apply to users who were never pushed.

It's an edge case but `has_active_push_subscription` is evaluated when the email job runs, not when the message arrived? So a user with no push subscription at message time gets no push, installs the app two minutes later, and at the 5-minute mark now looks push-capable. Then their email is deferred 24 hours despite them never having been notified at all.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -170,6 +170,32 @@ def purge_account_deletion_tokens(payload: empty_pb2.Empty) -> None:
...
+    """
+    has_active_push_subscription = (
+        select(PushNotificationSubscription.id)
+        .where(PushNotificationSubscription.user_id == user_id_column)
+        .where(PushNotificationSubscription.disabled_at > func.now())
+        .exists()
+    )
+    return or_(
+        and_(has_active_push_subscription, Message.time < now() - MISSED_MESSAGES_DELAY_WITH_PUSH),
+        and_(not_(has_active_push_subscription), Message.time < now() - MISSED_MESSAGES_DELAY),
```
nabramow: `Message.time < now() - DELAY` uses `Python now()`, while line 190 uses SQL func.now(). Do the mixed clocks matter?

nabramow: Nit - since `has_active_push_subscription` is referenced twice (once positively and once negatively), I think this means the `exists()` gets planned twice per query? Would `case` make sense here?

app/backend/src/couchers/jobs/handlers.py
aapeliv: Yeah I guess that's correct, but very unlikely. Either way if you just used the app, you have seen the notification count and ignored it.

app/backend/src/couchers/jobs/handlers.py
aapeliv: Yeah that's a good point, should use `func.now()`. Thanks for noticing it.

aapeliv: i'll do it in a follow-up since testing is a pain

aapeliv: yeah good point, fixed. thank you!

PR #9342, opened 2026-07-26
Backend: record every SQL query the tests issue and diff it per PR

aapeliv:
Makes it possible to review a PR for how it changes database access patterns, before it merges.

An opt-in `--query-log` flag records every SQL query the backend suite issues, grouped both by the test that ran and by the RPC that issued it, in order. CI diffs that recording against the one `develop` last published and links a browsable report from the sticky PR comment, with the headline delta inline so the common case needs no click:

> `SQL query log` — +3/-1 query shapes · 4 tests changed · worst: ListReferences 1->14 queries/call

Each query is stored twice: fingerprinted with parameters replaced by `?` (the key used for grouping and diffing), and with the values inlined by psycopg's `ClientCursor.mogrify`, so a query reads `WHERE "references".from_user_id = 2::BIGINT AND ... = 'friend'` and can be copy-pasted into psql to explain it against real data. The report defaults to what changed, collapses consecutive repeats of one shape to `×14` as a likely N+1, and toggles between the runnable and parameterized forms.

This reuses the pipeline the schema diff already uses — per-node artifact, merge, fetch develop's baseline, render, publish, comment — so there is **no new CI job**. Overhead is ~5-8% on the test run and one ~9 MB artifact.

**This also fixes a pre-existing nondeterminism in generated SQL.** `get_moderated_models()` iterates `Base.registry.mappers`, which is a `frozenset`, so `Mapper` objects come out ordered by `id()` — addresses that vary per process. Callers build one `OR` branch per entry, so the moderation visibility predicate was emitted with its branches in a different order in every process. That made **83 of 943 tests** differ between two identical runs, and in production splits one logical query across several `pg_stat_statements` entries, defeating the aggregation you want when investigating query performance. Sorting the mappers fixes it.

Docs in `docs/query-log.md`, including the caveats: ordering within a span is deliberately not compared (background jobs tie on their queue ordering, so drain order varies), the inlined values come from a truncated test database and are a template rather than a query, and job spans are not yet split by job type.

## Testing

`uv run pytest --query-log src` locally, then verified:

- **Determinism** (the gate — if the recording is not stable, everything downstream is noise): two full-suite runs produce **0/943 canonical differences** and identical shape sets (1474 shapes). Before the moderation ordering fix this was 83/943 and 16 shapes apart. The 5 remaining strict-order differences are all background-job drain order, which the diff ignores by design.
- **No false positives**: two independent full runs diff to *"no change in DB access patterns"*.
- **Real changes are caught**: synthetic N+1, new shape, added and removed test all reported correctly, including `worst: WriteFriendReference 26->38 queries/call`.
- **3-node merge**: splitting a real recording into three node files and merging reconstructs the single-process recording exactly, including the `first_seen_in` tie-break that picks each stored example.
- Full suite green: **1228 passed**. `make format` and `make mypy` clean (464 files).

Not verified locally: the GitLab job stanzas themselves. The recorder, merge, diff, viewer and comment-item rendering are all exercised, but whether the CI wiring hangs together needs a real pipeline — hence the draft.

To see the report locally:

```bash
cd app/backend
uv run pytest --query-log src
uv run python ../scripts/query_log_report.py --input test_artifacts/queries --output test_artifacts/queries
python3 -m http.server --directory test_artifacts/queries   # the page fetches its data, so file:// won't work
```

Two drive-by mypy fixes are included (`test_events.py`, `test_calendar_events.py`) — pre-existing errors on `develop` from the `GetEventCalendarFile` commit, unrelated to this change but needed for mypy to pass. Happy to split them out.

**Backend checklist**
- [ ] Run the backend locally and it works — full test suite passes, but I did not launch the server; the changes are test infrastructure, CI, and one ordering fix in query construction
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history — no database changes

## For maintainers

### DIFF
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -796,6 +796,9 @@ preview:backend:
     - aws s3 rm s3://$AWS_PREVIEW_BUCKET/schema/$CI_COMMIT_REF_SLUG/ --recursive
     - aws s3 cp artifacts/schema_dumps s3://$AWS_PREVIEW_BUCKET/schema/$CI_COMMIT_REF_SLUG/ --recursive
     - echo "Done, schema dumps available at https://$CI_COMMIT_SHORT_SHA--schema.$PREVIEW_DOMAIN/schema.sql and https://$CI_COMMIT_REF_SLUG--schema.$PREVIEW_DOMAIN/schema.sql"
+    # The per-node query logs came in from test:backend alongside preview:backend-coverage's merged data.json.gz,
+    # and only the merged one is worth publishing
+    - find artifacts/test_artifacts/queries -name 'data.*.json.gz' ! -name 'data.json.gz' -delete || true
     # Upload postcard samples
     - aws s3 rm s3://$AWS_PREVIEW_BUCKET/test-artifacts/$CI_COMMIT_SHORT_SHA/ --recursive
     - aws s3 cp artifacts/test_artifacts s3://$AWS_PREVIEW_BUCKET/test-artifacts/$CI_COMMIT_SHORT_SHA/ --recursive 2>/dev/null || true
--- a/app/.gitlab-ci.yml
+++ b/app/.gitlab-ci.yml
@@ -342,7 +342,9 @@ test:backend:
     DATABASE_CONNECTION_STRING: "postgresql://postgres:c765064a49d18a95@postgres/testdb"
     COVERAGE_FILE: /app/.coverage.$CI_NODE_INDEX
   script:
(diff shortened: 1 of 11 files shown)


nabramow:
You tagged me for review, and the PR description is a bit of a wall of text that's hard to understand, so not sure how helpful I am as Python best practices are really not my strong suit, BUT my understanding is we're trying to better see if the tests are causing a big hit to SQL query performance?

I assume this is not super risky as it's mostly tests?

PR #9348, opened 2026-07-27
Backend/tests: clock mocking proof of concept

tristanlabelle:
Demonstrates use of the `time_machine` library to mock the clock.

We rely on comparisons with `now()` throughout the backend and tests and often offset it by some hours to ensure it won't be in the past. We also don't have a good way to test things that happen after a delay (e.g. throttled notifications). Clock mocking allows us to control time and remove such workarounds. Freezing time is especially useful since logic shouldn't depend on how fast a test executes.

One caveat: We can't use the database now() function, since its clock isn't mocked, so we must move initialization the initialization of datetimes to the client. I'm not sure if databases guarantee that two "now()" used in the same statement produce the same value, but this isn't the case on the client.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/backend/pyproject.toml
+++ b/app/backend/pyproject.toml
@@ -56,6 +56,7 @@ dev = [
     "pytest-cov>=6.2.1",
     "pre-commit>=4.2.0",
     "mypy>=1.18.2",
+    "time-machine>=3.2.0",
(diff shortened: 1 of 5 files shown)


tristanlabelle:
Abandoning in favor of #9354

PR #9335, opened 2026-07-26
Frontend/i18n: Extract i18n date.ts functions to i18n/datetimes.ts

tristanlabelle:
`utils/date.ts` had grown to include a lot, so extract the i18n stuff to `i18n/datetimes.ts`. No functional changes, just moving code around.

## Testing
Viewed dates using the Vercel preview. Didn't blow up.

**Web frontend checklist**
- [ ] There are no console warnings when running the app
- [ ] Added tests where relevant
- [x] Clicked around my changes running locally and it works
- [ ] Checked Desktop, Mobile and Tablet screen sizes

## For maintainers

- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

PR #9351, opened 2026-07-27
Frontend/i18n: Hide date year when this year

### LINKED ISSUE #9314: Frontend/i18n: Omit date year when not necessary
Our datetimes can be a mouthful. We can simplify them by omitting the year when it's the same as currently. For example in

Image

tristanlabelle:
Event datetime ranges can be a mouthful because they include so many pieces of information. We can make it more digestible by only showing the year when it's not the current year.

Next: We could do the same thing and omit minutes if they are 00 (i.e. just show "8 PM to 10 PM")

Fixes #9314

PS: Please review #9335 first!

## Testing
Added tests + ran frontend locally. See the difference between those two events:

image

Compare the verbosity on NEXT:

image

VS with this change:

image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/communities/events/EventCard.tsx
+++ b/app/web/features/communities/events/EventCard.tsx
@@ -132,6 +132,7 @@ export default function EventCard({ event, className }: EventCardProps) {
     timestampToPlainDateTime(event.endTime!, event.timezone),
     {
       locale,
+      includeYear: "auto",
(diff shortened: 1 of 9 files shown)


nabramow (approved):
LGTM and works well! Looks way better and I support the time suggestion too!

PR #9354, opened 2026-07-28
Add timewarp test fixture for shifting the clock

aapeliv:
Adds a `timewarp` test fixture that shifts the clock that both python and postgres read:

```python
def test_something(db, timewarp):
    timewarp.set_time(datetime(2024, 2, 29, 12, tzinfo=UTC))
    timewarp.add_time(timedelta(days=30))   # negative rewinds
```

Postgres reads the clock through a `mock.now()` that shadows `pg_catalog.now()` via `search_path`. It has to be installed before any DDL runs, since a column `DEFAULT now()` binds the function identity when the column is created. Python goes through a new `_real_now()` seam in `utils`, which covers the ~42 modules that do `from couchers.utils import now`.

Both clocks keep running — the offset displaces them rather than freezing them, and each side applies it to its own clock. Moving the clock while a transaction is open raises, since postgres fixes `now()` at transaction start.

Also: `servicers/bugs.py` now uses `utils.now()` instead of `datetime.now(UTC)`, and the migration drift test builds its schemas in a scratch database so both dumps stay production-shaped.

## Testing

`src/tests/test_timewarp.py` covers both clocks under set/add/rewind/reset, column defaults, query-time `now()`, the mid-transaction guard, and a `pg_depend` check that nothing still binds the real `pg_catalog.now()`.

`uv run pytest` from `app/backend` — 1243 passed, `make mypy` clean.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -138,8 +138,14 @@ def datetime_to_iso8601_local(value: datetime) -> str:
     return value.replace(tzinfo=None).isoformat()
 
 
-def now() -> datetime:
+def _real_now() -> datetime:
     return datetime.now(tz=UTC)
+
+
+def now() -> datetime:
+    # everything that reads the clock goes through this call, so tests can move it in one place
+    # by swapping _real_now; see the timewarp fixture
+    return _real_now()
 
 
 def minimum_allowed_birthdate() -> date:
@@ -162,7 +168,7 @@ def now_in_timezone(tz: str) -> datetime:
     """
     tz should be tzdata identifier, e.g. America/New_York
     """
-    return datetime.now(ZoneInfo(tz))
+    return now().astimezone(ZoneInfo(tz))
 
 
 def today_in_timezone(tz: str) -> date:
--- a/app/backend/src/couchers/utils.py
+++ b/app/backend/src/couchers/utils.py
@@ -138,14 +138,14 @@ def datetime_to_iso8601_local(value: datetime) -> str:
     return value.replace(tzinfo=None).isoformat()
 
 
-def _real_now() -> datetime:
+def _mockable_now() -> datetime:
     return datetime.now(tz=UTC)
 
 
 def now() -> datetime:
     # everything that reads the clock goes through this call, so tests can move it in one place
-    # by swapping _real_now; see the timewarp fixture
-    return _real_now()
+    # by swapping _mockable_now; see the timewarp fixture
+    return _mockable_now()
 
 
 def minimum_allowed_birthdate() -> date:
--- a/app/backend/src/tests/conftest.py
+++ b/app/backend/src/tests/conftest.py
@@ -27,11 +27,21 @@ from couchers.models import Base  # noqa: E402
 from tests.fixtures import query_log  # noqa: E402
 from tests.fixtures.db import (  # noqa: E402
     autocommit_engine,
+    create_mock_clock,
     create_schema_from_models,
     generate_user,
     populate_testing_resources,
 )
 from tests.fixtures.misc import EmailCollector, Moderator, PushCollector  # noqa: E402
+from tests.fixtures.timewarp import Timewarp, install_timewarp  # noqa: E402
+
+
+@pytest.fixture
+def timewarp() -> Generator[Timewarp]:
+    """
+    Lets a test move the clock; see Timewarp. Request `db` alongside it for the postgres side.
+    """
+    yield from install_timewarp()
 
 QUERY_LOG_DIR = Path(__file__).resolve().parents[2] / "test_artifacts" / "queries"
 
@@ -125,6 +135,13 @@ def setup_testdb(postgres_conn: Connection, testdb_engine: Engine) -> None:
     postgres_conn.execute(text("DROP DATABASE IF EXISTS testdb WITH (FORCE)"))
     postgres_conn.execute(text("CREATE DATABASE testdb"))
 
+    # A column DEFAULT resolves now() once, when the column is created, and stores the function
+    # identity forever after; later search_path changes don't reach it. So mock.now() has to
+    # already shadow pg_catalog.now() on every connection before any DDL runs, which means
+    # setting this at the database level here, ahead of the first connect. The mock schema
+    # doesn't exist yet, which postgres tolerates in a search_path.
+    postgres_conn.execute(text("ALTER DATABASE testdb SET search_path = public, mock, pg_catalog"))
+
     with testdb_engine.connect() as conn:
         conn.execute(
             text(
@@ -135,6 +152,7 @@ def setup_testdb(postgres_conn: Connection, testdb_engine: Engine) -> None:
                 "CREATE EXTENSION IF NOT EXISTS pg_stat_statements;"
             )
         )
+        create_mock_clock(conn)
 
         create_schema_from_models(testdb_engine)
         populate_testing_resources(conn)
--- a/app/backend/src/tests/conftest.py
+++ b/app/backend/src/tests/conftest.py
@@ -27,21 +27,19 @@ from couchers.models import Base  # noqa: E402
 from tests.fixtures import query_log  # noqa: E402
 from tests.fixtures.db import (  # noqa: E402
     autocommit_engine,
-    create_mock_clock,
     create_schema_from_models,
     generate_user,
     populate_testing_resources,
 )
 from tests.fixtures.misc import EmailCollector, Moderator, PushCollector  # noqa: E402
-from tests.fixtures.timewarp import Timewarp, install_timewarp  # noqa: E402
-
-
-@pytest.fixture
-def timewarp() -> Generator[Timewarp]:
-    """
-    Lets a test move the clock; see Timewarp. Request `db` alongside it for the postgres side.
-    """
-    yield from install_timewarp()
+from tests.fixtures.timewarp import (  # noqa: E402
+    FROZEN_TEST_TIME,
+    MOCK_SEARCH_PATH,
+    FrozenTimewarp,
+    Timewarp,
+    create_mock_clock,
+    install_timewarp,
+)
 
 QUERY_LOG_DIR = Path(__file__).resolve().parents[2] / "test_artifacts" / "queries"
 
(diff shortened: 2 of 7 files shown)


app/backend/src/couchers/utils.py
```diff
@@ -138,10 +138,16 @@ def datetime_to_iso8601_local(value: datetime) -> str:
     return value.replace(tzinfo=None).isoformat()
 
 
-def now() -> datetime:
+def _real_now() -> datetime:
```
tristanlabelle: Suggestion: `_mockable_now`. I think that says it all.

app/backend/src/tests/fixtures/db.py
```diff
@@ -42,6 +42,30 @@
 from tests.fixtures.sessions import _MockCouchersContext
 
 
+def create_mock_clock(conn: Connection) -> None:
```
tristanlabelle: That looks simpler than I expected!

app/backend/src/tests/fixtures/db.py
```diff
@@ -42,6 +42,30 @@
...
+    models bake the same function into their column defaults.
+    """
+    conn.execute(
+        text("""
+        CREATE SCHEMA mock;
+
+        CREATE FUNCTION mock.now() RETURNS timestamptz
+        LANGUAGE sql STABLE AS $$
+          SELECT pg_catalog.now() + coalesce(
+            nullif(current_setting('mock.offset', true), '')::interval,
```
tristanlabelle: Extract `mock.offset` as a constant since it's an implicit dependency between two files

app/backend/src/tests/fixtures/timewarp.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _as_interval(delta: timedelta) -> str:
+    # these three are exact and sum to delta, unlike total_seconds() which is a float
+    return f"{delta.days} days {delta.seconds} seconds {delta.microseconds} microseconds"
+
+
+def install_timewarp() -> Generator[Timewarp]:
+    warp = Timewarp()
+    engine = _get_base_engine()
```
tristanlabelle, line 160: What happens if the test doesn't require the db fixture?

app/backend/src/tests/test_timewarp.py
```diff
@@ -0,0 +1,199 @@
...
+import pytest
+from sqlalchemy import func, select
+from sqlalchemy.sql import text
+
+from couchers.db import session_scope
+from couchers.models import BackgroundJob
+from couchers.utils import now, now_in_timezone, today
+from tests.fixtures.timewarp import Timewarp
+
+LEAP_DAY = datetime(2024, 2, 29, 12, 0, 0, tzinfo=UTC)
```
tristanlabelle, line 15: Oh nice, good thinking!

app/backend/src/tests/fixtures/timewarp.py
```diff
@@ -0,0 +1,83 @@
...
+    Shifts the clock that both python and postgres read, for tests that need to sit at a particular
+    date or jump around in time.
+
+        timewarp.set_time(datetime(2024, 2, 29, 12, tzinfo=UTC))
+        timewarp.add_time(timedelta(days=30))
+
+    Both clocks keep running, they're just displaced by an offset. Each side applies that offset to
+    its own clock, so the two stay as (im)perfectly aligned as they normally are, and set_time lands
+    within a hair of the instant asked for rather than exactly on it. Postgres also still reports
+    transaction start time, so a long transaction sees its own start rather than a later add_time.
```
tristanlabelle: Once we have clock mocking, I can't think of a case where we'd want to keep the clock running during tests. Can you? I think we should always control it programatically.

app/backend/src/tests/conftest.py
```diff
@@ -24,11 +24,21 @@
...
 from tests.fixtures.misc import EmailCollector, Moderator, PushCollector  # noqa: E402
+from tests.fixtures.timewarp import Timewarp, install_timewarp  # noqa: E402
+
+
+@pytest.fixture
+def timewarp() -> Generator[Timewarp]:
+    """
+    Lets a test move the clock; see Timewarp. Request `db` alongside it for the postgres side.
+    """
+    yield from install_timewarp()
```
tristanlabelle: I wonder if at install time we should already pin the clock to a well-known point in time, e.g. 2000-01-01 00:00. Otherwise I we'll probably end up manually doing it in all tests.

tristanlabelle (commented):
This is really cool. I hadn't thought that our `now()` was so pervasive as to be reliable for mocking, and the DB integration is not too bad.

My biggest consideration is that I think we always want to freeze `now()` in tests that mock the clock (maybe even in all tests), and control time manually. Also a thought about standardizing on a datetime like 2000-01-01.

aapeliv:
@tristanlabelle I did think about the frozen clock vs offset. I feel that the running clocks with an offset are better. With frozen clocks, we can introduce bugs by making false assumptions. With running clocks, it guarantees that what is otherwise guaranteed (e.g. postgres now is frozen within a transaction) holds and what is not guaranteed (python and postgres will disagree) does not hold. What's the use-case where you think we really do need a frozen clock, where just setting it at a given date and allowing it to run does not suffice?

app/backend/src/tests/fixtures/timewarp.py
aapeliv: I can't think of a case where we need to have them frozen! I think we should mock/artificially change as little of the universe as we need to in these tests.

tristanlabelle: Freezing the clock allows us to:

- Assert on created/updated timestamps
- Assert on boundary conditions for minimum/maximum valid dates (e.g. that an event can start now but not now - 1ms)
- Assert about dates as formatted in text, e.g. notification should be "The event starts Tuesday, June 30"
- Decide on the precision of date (e.g. test that a request to create an event at noon + 1 second will end up creating an event at noon)
- Verify delayed events without noise (e.g. we send a notification 1 second later and want to move the clock by 1 second, without the possibility of the clock just having advanced by 1 second on its own).
- Testing leap seconds :P

I think we should treat time as an external dependency like others. I would use a frozen clock in all of my tests. Do you have a use case for non-frozen clocks?

tristanlabelle: Gemini's take ("Why Freezing Time is the Standard"): https://share.gemini.google/FpalNkJdK8P5

app/backend/src/tests/fixtures/timewarp.py
```diff
@@ -0,0 +1,176 @@
...
+from datetime import UTC, datetime, timedelta
+from unittest.mock import patch
+
+from sqlalchemy import Connection, event, text
+
+from couchers import utils
+from couchers.db import _get_base_engine
+
+# 2020-05-16 10:00 in Brisbane
+FROZEN_TEST_TIME = datetime(2020, 5, 16, 0, 0, tzinfo=UTC)
```
tristanlabelle: If we want a default value it should be as neutral as possible, I suggest 2020-01-01 or the unix epoch. Otherwise it'll be hard to understand why tests are checking for some random date in May, and if all tests have to reference offsets from FROZEN_TEST_TIME, we haven't improved the ability to assert for a formatted date within a notification using a constant, for example.

We could even call the fixture `y2k_timewarp`... though it'd get confusing when changing the time.

app/backend/src/tests/test_timewarp.py
```diff
@@ -0,0 +1,365 @@
...
+
+def db_now() -> datetime:
+    with session_scope() as session:
+        return session.execute(select(func.now())).scalar_one()
+
+
+def test_set_time_moves_both_clocks(db, timewarp: Timewarp) -> None:
+    timewarp.set_time(LEAP_DAY)
+    assert_close(now(), LEAP_DAY)
+    assert_close(db_now(), LEAP_DAY)
```
tristanlabelle, line 37: I'd be more comfortable with a 1 minute tolerance to remove the possibility of flakiness.

app/backend/src/tests/fixtures/timewarp.py
```diff
@@ -0,0 +1,176 @@
...
+        return datetime.now(tz=UTC) + self.offset
+
+    def freeze_time(self, when: datetime) -> None:
+        """Stops the clock at `when`, so it reads back exactly rather than ticking on from there."""
+        _check_aware(when)
+        self._refuse_mid_transaction()
+        # utils.now() is UTC-aware everywhere else, and callers do read the tzinfo off it
+        self.frozen_at = when.astimezone(UTC)
+
+    def set_time(self, when: datetime) -> None:
```
tristanlabelle: Suggestion: `def set_time(self, when: datetime, *, frozen: bool)` (mandatory `frozen` param and remove `freeze_time`).

Currently it's more stateful because behavior depends on whether `freeze_time` has been called. I think callsites would be more readable if they always specified whether to freeze or not.

app/backend/src/tests/test_timewarp.py
```diff
@@ -0,0 +1,365 @@
...
+    timewarp.set_time(LEAP_DAY)
+    with session_scope() as session:
+        session.add(make_job())
+    timewarp.add_time(timedelta(days=1))
+    with session_scope() as session:
+        session.add(make_job())
+    with session_scope() as session:
+        first, second = session.execute(select(BackgroundJob.queued).order_by(BackgroundJob.queued)).scalars().all()
+    assert_close(first, LEAP_DAY)
+    assert_close(second, LEAP_DAY + timedelta(days=1))
```
tristanlabelle, line 96: These have me shiver with PTSD from flaky tests. I'd freeze the clock in all tests but one that is meant to test unfrozen clocks.

tristanlabelle (dismissed):
I'm excited to put this to use. I have one API tweak suggestion. The fuzzy equality testing on `now()` screams flakiness to me.

aapeliv:
@tristanlabelle i'm still wrapping up this branch, but just want to comment on the above. I appreciate your concern for flaky tests and wanting the test environment clean. I don't think this is a permanent decision that can't be undone: if we soon decide frozen clocks are always the right thing, then it's easy to claude code refactor everything to use it. My concern is that I would rather catch issues where we assume the python/db clock is the same in tests rather than in prod, and we have had issues from making that assumption in the past. Being able to rewind clocks easily without all the custom mocking will be a huge QOL improvement, and I think a good step. We can consider the frozen clocks step next if we find this to not get us close enough.

PR #9355, opened 2026-07-28
Add editor API to list decided event community invite requests

aapeliv:
Adds a new editor-level RPC `ListDecidedEventCommunityInviteRequests` that lists event community invite requests that have already been approved or declined, complementing the existing `ListEventCommunityInviteRequests` (which only returns pending ones). This gives editors a view of past decisions.

Details:
- `ListDecidedEventCommunityInviteRequestsReq` takes `page_size`, `page_token`, and an optional `BoolValue approved` filter (unset returns both approved and declined).
- Each `DecidedEventCommunityInviteRequest` returns the request id, `user_id`, `event_url`, `community_id`, `created`, `decided`, `decided_by_user_id` and `approved`.
- Results are ordered most recently decided first, paginated with a keyset cursor on `(decided, id)` using the existing `dt_id_to_page_token`/`dt_id_from_page_token` helpers. A plain id cursor would order by creation rather than decision time, and bulk auto-declines share a `decided` timestamp, hence the id tiebreak.
- `community_id` is taken from the event's `parent_node_id` rather than `get_users_to_notify_for_new_event`, since that query is expensive and its notify count is no longer meaningful once a request is decided.

## Testing
Added `test_list_decided_community_invite_requests` in `src/tests/test_events.py`: it creates three invite requests, decides two of them (one approved, one declined), and asserts that the pending one is excluded, that ordering is newest-decision-first, that all returned fields are correct, that both values of the `approved` filter work, and that two-page pagination works.

To replicate locally, from `/app/backend`:

```
uv run pytest src/tests/test_events.py -k community_invite
```

`make format` and `make mypy` both pass.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/editor.py
+++ b/app/backend/src/couchers/servicers/editor.py
@@ -191,6 +191,7 @@ class Editor(editor_pb2_grpc.EditorServicer):
         requests = session.execute(query).scalars().all()
 
         def _request_to_pb(req: EventCommunityInviteRequest) -> editor_pb2.DecidedEventCommunityInviteRequest:
+            # these are set together, per the decided_approved check constraint
             assert req.decided is not None and req.decided_by_user_id is not None and req.approved is not None
             return editor_pb2.DecidedEventCommunityInviteRequest(
                 event_community_invite_request_id=req.id,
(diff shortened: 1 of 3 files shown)


nc225 (approved):
looks great

PR #9356, opened 2026-07-28
Fix icon size for event and reference notification settings

aapeliv:
before:
Screenshot 2026-07-27 at 22 26 38

after:

Screenshot 2026-07-27 at 22 26 58


nabramow (approved):
LGTM!

PR #9392, opened 2026-07-29
Backend: count references from deleted authors, matching the referenc…

dymbe:
Before:

image

After:

image

Bug discovered in #9162 

The reference list (ListReferences) deliberately keeps showing references written by deleted users (rendered as an anonymized deactivated account), but get_num_references filtered authors on User.is_visible, which excludes deleted users - so a profile's reference count could be lower than the list it sits next to.

Align get_num_references' author-visibility rule with ListReferences: references by banned authors are hidden, ones by shadowed authors are visible only to the shadowed author themselves, and ones by deleted or blocked authors stay counted.

Also replace GetPublicUser's own inline reference count with get_num_references: it applied neither this rule, nor the reciprocal reference hiding rule, nor moderation visibility, so public profiles could show yet another count.

Enables the previously commented-out assert from #9162 as the regression test.

## Testing

- The assert that was left commented out in #9162 is now enabled and acts as the
  regression test: on develop it fails (the count says 0 while the list shows 1),
  with this fix it passes.
- Added `test_ListReference_shadowed_users`: both the list and the count hide a
  shadowed user's reference from other users, while the shadowed author still sees
  their own reference in both (so the shadow ban isn't leaked to them).
- Added `test_GetPublicUser_num_references_visibility_rules`: the public profile
  count follows the same rules as the list (references hidden by the reciprocity
  window aren't counted, references from deleted users are, references from
  shadowed users aren't).
- Full backend test suite passes locally, `make format` and `make mypy` clean.
- Verified manually in the local dev stack: gave a user some references and set
  `deleted_at` on one of the authors. On develop the profile badge shows a lower
  count than the list next to it; with this branch they agree.

To replicate manually: create a few references to a user, set `deleted_at` on one
of the authors in the db, and compare the count badge with the references list on
their profile.

**Backend checklist**

- [x] ~~Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history~~ (no changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/public.py
+++ b/app/backend/src/couchers/servicers/public.py
@@ -18,7 +18,6 @@ from couchers.models import (
     InvoiceType,
     Node,
     ProfilePublicVisibility,
-    Reference,
     User,
     Volunteer,
 )
@@ -26,7 +25,13 @@ from couchers.models.uploads import get_avatar_upload
 from couchers.proto import api_pb2, public_pb2, public_pb2_grpc
 from couchers.proto.google.api import httpbody_pb2
 from couchers.resources import get_static_badge_dict
-from couchers.servicers.api import fluency2api, hostingstatus2api, meetupstatus2api, user_model_to_pb
+from couchers.servicers.api import (
+    fluency2api,
+    get_num_references,
+    hostingstatus2api,
+    meetupstatus2api,
+    user_model_to_pb,
+)
 from couchers.servicers.gis import _statement_to_geojson_response
 from couchers.sql import users_visible
 from couchers.utils import Timestamp_from_datetime, not_none, now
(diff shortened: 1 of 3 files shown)


aapeliv (commented):
hey @dymbe, thanks for yet another good PR. i just merged https://github.com/Couchers-org/couchers/pull/8651. i had something touching this going on there, i'm not sure if that fixed this bug or just changed it. sorry for the conflict.

nabramow:
@dymbe Are you still working on this or should it be closed? Seems from @aapeliv 's last comment there's a chance it was solved?

dymbe:
Thanks for the fix @aapeliv, no worries about the conflict, and yes @nabramow the changes made in #8651 fixed the main issue here. I see that GetPublicUser (app/backend/src/couchers/servicers/public.py) still computes its own incorrect number of references, but it also seems like it's not really in use anywhere? I pushed a commit that fixes this, as well as some tests, but if it's dead code anyway and it won't be used in the future feel free to just close this PR

aapeliv (approved):
thanks for the PR, and sorry for the delay! feel free to ping next time if we miss something

PR #9393, opened 2026-07-29
Frontend/i18n: Fix browser language detection for zh-Hant

### LINKED ISSUE #9319: Frontend/i18n: Browser locale detection ignores language variants
`getBrowserLocaleFromHeader` only considers the base language and won't distinguish `zh-Hans` from `zh-Hant` or `pt` from `pt-BR`:

https://github.com/Couchers-org/couchers/blob/69e29a177f345e290006fe69be4be55e4e612c19/app/web/utils/getBrowserLocaleFromHeader.ts#L16

tristanlabelle:
Setting the user agent to zh-Hant and navigating to Couchers would result in zh-Hans being shown, because our logic stripped any suffixes. Same with pt-BR selecting pt.

I reimplemented it according to the RFC (the best library I found didn't support suffixes). And split tests between unit tests for the acceptLanguage parsing/lookup behavior vs tests for our middleware's language choice.

Also fixes a bug where if we have "fr;es;en" and `es` is production ready while `fr` is not, we'd return `en`.

Fixes #9319

## Testing
In Vercel preview, select English, then use browser extension to override accept-language to zh-Hant, clear cookies, and reload URL. Before: would load zh-Hans, now: zh-Hant.

**Web frontend checklist**


nabramow:
I really wonder if there's a simpler way to do this with i18n? I thought you can just feed it multiple fallback languages in order in the init config file?

If not we need to test this really thoroughly with the app and logging in and out while switching between languages. That middleware.ts file is really tricky and has the potential to break a lot of things.

app/web/i18n/acceptLanguage.ts
```diff
@@ -0,0 +1,43 @@
...
+  // Consider accepted locales by descending quality.
+  const acceptLocales = parseAcceptLanguage(header)
+    .sort((a, b) => b.quality - a.quality)
+    .map((e) => e.code);
+
+  for (const acceptLocale of acceptLocales) {
+    // RFC 4647 lookup: First check pt-BR, then pt
+    let possibleLocale = acceptLocale;
+    while (true) {
+      if (supportedLocales.includes(possibleLocale)) {
```
HankAviator: RFC 4647 requires language-range matching to be case-insensitive. This includes check means a valid header such as pt-br or ZH-hant will not match the canonical pt-BR or zh-Hant locale. Could we compare case-insensitively while returning the supported locale’s canonical spelling, with a regression test?

app/web/i18n/acceptLanguage.ts
```diff
@@ -0,0 +1,43 @@
...
+    .filter((entry) => entry.code.length > 0);
+}
+
+/** Looks up a supported locale based on an Accept-Language HTTP header. */
+export function lookupAcceptLanguage(
+  header: string,
+  supportedLocales: string[],
+): string | undefined {
+  // Consider accepted locales by descending quality.
+  const acceptLocales = parseAcceptLanguage(header)
```
HankAviator, line 23: Could we filter out entries with `quality === 0` before sorting and mapping them? In an Accept-Language header, q=0 means “not acceptable,” but such an entry can currently still be selected. For example, lookupAcceptLanguage("fr;q=0", ["fr", "en"]) returns fr.

HankAviator (changes requested):
The PR makes sense to me overall. Added some edge cases which you may want to address

app/web/i18n/acceptLanguage.ts
tristanlabelle: Good eyes. Fixed in latest

app/web/i18n/acceptLanguage.ts
tristanlabelle: Fixed in latest. Thanks!

HankAviator (approved):

tristanlabelle:
> I really wonder if there's a simpler way to do this with i18n? I thought you can just feed it multiple fallback languages in order in the init config file?

It seems like i18next has something like this https://nextjs.org/docs/pages/guides/internationalization , but reading between the lines I imagine it doesn't work for us because we restrict languages at runtime based on production readiness.

> If not we need to test this really thoroughly with the app and logging in and out while switching between languages. That middleware.ts file is really tricky and has the potential to break a lot of things.

I tested various combinations of switching languages while logged in and out, clearing the cookie and mocking accept-header and I'm getting the expected behavior. I only touched the best language selection in `middleware.ts` so the blast radius should be limited.

Issue #9398, opened 2026-07-31
References do not appear


OlehDnipro:
So, after I have written the reference to my guest ( after he did same) it do not appear on my profile - and it happened second time already! The service is disfunctional without references... please look into that issue!

I've accessed reference procedure once from notifications, and once from thechat

P.S. when i try to use bug button on the app, issue is not being created...

aapeliv:
Can you tell me your username and who you believe wrote a reference for you that’s not showing up?

Also what do you mean that the issue is not being created? Does it not submit and thank you for submitting a bug?

OlehDnipro:
Sure, my guests were Miguel @mitacla96 and Gonzalo @gdejuan - i've got notification that they've written their references, so i am pretty sure they did...my username is @olehforester

And regarding bug button - it shows message that isue has been created with certain number, but if i click on it i got error 404, and its also not visible on github.

aapeliv:
Your hosting reference from June is visible. The other two people you have written a reference for have not written one for you yet. When they write their references, they will show up on your profile.

I do also see the issue you opened (https://github.com/Couchers-org/bugs/issues/331; you will not see it since it’s not public).

aapeliv:
Sorry just saw your response. Let me check why you would have gotten notifications that they wrote then given they never did.

aapeliv:
> i've got notification that they've written their references

which notifications were those? do you have timestamps or the X-Couchers-ID email header?

OlehDnipro:
I remember those were on the app, where i clicked on them and wrote a reference to guest. Now i do not see them in log. I will send a message to ask my guests if they really wrote ones, to be certain

aapeliv:
I can see we sent you push notifications for reminders to write push notifications on 2026-07-19 20:01:40.841242+00, 2026-07-22 20:26:10.314851+00, and 2026-07-29 14:14:49.870886+00; but as per logs, not for your guests having written references.

OlehDnipro:
You were right! They didnt write till now, so I perceived reminder to write one as a note that I got one. Sorry :)

aapeliv:
No worries, thanks for reporting and following up! We do sometimes have mystery bugs and it's good to check!

PR #9400, opened 2026-07-31
Remove issue link from bug report success toast

aapeliv:
Since the bugs repo is not public, this removes the link to it.

Related to https://github.com/Couchers-org/couchers/issues/9398

---

No locale changes were needed: every language's `report.bug.success_message2` keeps the same `<id/>` placeholder, which now renders as a plain `<span>` instead of a `StyledLink`. `bug_url` remains in the RPC response but is no longer used by the web frontend.

## Testing
`yarn test ReportButton` passes (11 tests). The success-toast test was updated to assert the bug ID text is visible and that the alert contains no link.

To check manually: open the report dialog from the nav bar, choose "Report a bug", submit the form, and confirm the success toast shows the bug ID as plain, unclickable text.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/Navigation/ReportButton.test.tsx
+++ b/app/web/components/Navigation/ReportButton.test.tsx
@@ -238,9 +238,8 @@ describe("ReportButton", () => {
       );
 
       const successAlert = await screen.findByRole("alert");
-      const bugLink = await within(successAlert).findByRole("link");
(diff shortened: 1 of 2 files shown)


nabramow (approved):
LGTM

PR #9401, opened 2026-07-31
Add report buttons to discussion and event pages

### LINKED ISSUE #9394: Add 'Report' icon to all Discussion and Event pages
Problem: Currently there is no way to report the content of a Discussion or Event page. Users can report comments/replies but not the pages themselves. If an Event or Discussion is violating the Community Guidelines, it isn't clear how to report that content.

Proposed solution: Events and Discussions should have report buttons to make it easier to report when necessary.

aapeliv:
Screenshot 2026-07-31 at 07 40 17
Screenshot 2026-07-31 at 07 39 59

---

Discussions and events could previously only be reported from their cards in a list — there was no way to report the content from the discussion's or event's own page. This adds a flag button to both pages.

closes #9394

- `EventPage`: flag button in the existing icon-button group, next to the ellipsis menu, using `content_ref` `event/{event_id}` — the same ref the event cards emit, so reports from the card and the page look identical to moderators.
- `DiscussionPage`: flag button in the title row, next to the ellipsis menu. Hidden while editing and for deleted discussions (mirroring how `Comment.tsx` hides it for deleted comments).
- `FlagButton` gains an optional `ariaLabel`, defaulting to the existing generic label. These pages render a flag button per comment, so without a distinct label a screen-reader user can't tell which button reports the page's own content. The page-level buttons are labelled "Report this discussion" / "Report this event".
- The `community/{id}/discussion/{id}` vs `group/{id}/discussion/{id}` ref construction that was inline in `DiscussionCard` is extracted to `getDiscussionContentRef`, now shared by the card and the page.
- `docs/content_ref.md` claimed to list all `content_ref`s in use but only had profile and chat message; the discussion, event, comment, photo and public trip refs are now documented too.

## Testing

Added tests to both page test suites:
- `EventPage`: opening the report dialog, picking a reason and submitting calls `reportContent` with `event/1` and the event's author.
- `DiscussionPage`: the same for `community/1/discussion/1`, plus an assertion that no report button shows for a deleted discussion.

`yarn test features/communities/discussions/DiscussionPage.test.tsx features/communities/events/EventPage.test.tsx` passes, as do `yarn lint` and `tsc` for the touched files.

Note that `yarn test features/communities` has 24 pre-existing failures (CreateEventPage, EventForm, CommunitySearch, EventsPage); the identical 24 fail on a clean `develop` and are unrelated to this change.

To check manually: open any discussion page and any event page, click the flag icon next to the ellipsis menu, pick a reason and submit.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/FlagButton.tsx
+++ b/app/web/features/FlagButton.tsx
@@ -19,6 +19,7 @@ import { theme } from "theme";
 interface FlagButtonProps {
   contentRef: string;
   authorUser: string | number;
+  ariaLabel?: string;
   className?: string;
   size?: IconButtonProps["size"];
   renderButton?: (onClick: (event: React.MouseEvent) => void) => React.ReactNode;
@@ -27,6 +28,7 @@ interface FlagButtonProps {
 export default function FlagButton({
   contentRef,
   authorUser,
+  ariaLabel,
   className,
   size = "large",
   renderButton,
@@ -125,7 +127,7 @@ export default function FlagButton({
         renderButton(handleButtonClick)
       ) : (
         <IconButton
-          aria-label={t("report.flag.button_aria_label")}
+          aria-label={ariaLabel ?? t("report.flag.button_aria_label")}
           className={className}
           onClick={handleButtonClick}
           color="primary"
--- a/app/web/features/communities/discussions/Comment.tsx
+++ b/app/web/features/communities/discussions/Comment.tsx
@@ -8,6 +8,7 @@ import CenteredSpinner from "components/CenteredSpinner/CenteredSpinner";
 import EllipsisMenu, { EllipsisMenuItem } from "components/EllipsisMenu";
 import Markdown from "components/Markdown";
 import MarkdownInput, { MarkdownInputProps } from "components/MarkdownInput";
+import { contentRefs } from "features/contentRefs";
 import FlagButton from "features/FlagButton";
 import CopyOnClick from "features/mod/CopyOnClick";
 import ModVisibleComponent from "features/mod/ModVisibleComponent";
@@ -228,7 +229,7 @@ export default function Comment({ topLevel = false, comment, parentThreadId, dis
         {!comment.deleted && (
           <StyledButtonsContainer>
             <StyledAvatar user={user} />
-            <FlagButton contentRef={`comment/${comment.threadId}`} authorUser={comment.authorUserId} />
+            <FlagButton contentRef={contentRefs.comment(comment)} authorUser={comment.authorUserId} />
           </StyledButtonsContainer>
         )}
         <StyledCommentContent sx={comment.deleted ? { gridColumn: "1 / -1" } : undefined}>
--- a/app/web/features/communities/discussions/DiscussionCard.tsx
+++ b/app/web/features/communities/discussions/DiscussionCard.tsx
@@ -14,6 +14,7 @@ import { routeToDiscussion } from "routes";
 import { theme } from "theme";
 
 import getContentSummary from "../getContentSummary";
+import getDiscussionContentRef from "./getDiscussionContentRef";
 
 const StyledDeletedTitle = styled(Typography)(() => ({
   color: "var(--mui-palette-text-secondary)",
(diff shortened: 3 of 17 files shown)


nabramow:
Since there's quite a few of these contentRefs now, I think it would make sense to make them in one file, like:

`features/contentRefs.ts`:

```typescript
export const contentRefs = {
  event: (id: number) => `event/${id}`,
  comment: (threadId: number) => `comment/${threadId}`,
  // …one per type
};
```

Call sites become `contentRef={contentRefs.event(event.eventId)}`. You make a custom function like this for discussions, but the rest are now scattered strings. Makes sense to make this pattern reusable for all the other ones IMO.

app/web/features/communities/discussions/getDiscussionContentRef.ts
```diff
@@ -0,0 +1,12 @@
...
+
+// see //docs/content_ref.md
+export default function getDiscussionContentRef(
+  discussion: Discussion.AsObject,
+) {
+  const owner =
+    discussion.ownerCommunityId !== 0
+      ? `community/${discussion.ownerCommunityId}`
+      : `group/${discussion.ownerGroupId}`;
+  return `${owner}/discussion/${discussion.discussionId}`;
```
nabramow: Not a blocker, just something I noticed: the proto says a discussion belongs to either a community or a group `discussions.proto:41`

```grpc
oneof owner {
  int64 owner_community_id = 5;
  int64 owner_group_id = 6;
}
```

Protobuf keeps track of which one it is, but that gets lost when we convert to a plain object. `discussions_pb.d.ts:70` just gives you two numbers with no indication of which was set. 

So checking `!== 0` really is the only option here. I think this means a discussion with no owner would quietly come out as group/0/discussion/5 rather than erroring?

aapeliv:
> Since there's quite a few of these contentRefs now, I think it would make sense to make them in one file, like:
> 
> `features/contentRefs.ts`:
> 
> ```ts
> export const contentRefs = {
>   event: (id: number) => `event/${id}`,
>   comment: (threadId: number) => `comment/${threadId}`,
>   // …one per type
> };
> ```
> 
> Call sites become `contentRef={contentRefs.event(event.eventId)}`. You make a custom function like this for discussions, but the rest are now scattered strings. Makes sense to make this pattern reusable for all the other ones IMO.

can i do that in a follow-up?

nabramow:
> > Since there's quite a few of these contentRefs now, I think it would make sense to make them in one file, like:
> > `features/contentRefs.ts`:
> > ```ts
> > export const contentRefs = {
> >   event: (id: number) => `event/${id}`,
> >   comment: (threadId: number) => `comment/${threadId}`,
> >   // …one per type
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Call sites become `contentRef={contentRefs.event(event.eventId)}`. You make a custom function like this for discussions, but the rest are now scattered strings. Makes sense to make this pattern reusable for all the other ones IMO.
> 
> can i do that in a follow-up?

Is this PR really a rush to get in?

aapeliv:
> > > Since there's quite a few of these contentRefs now, I think it would make sense to make them in one file, like:
> > > `features/contentRefs.ts`:
> > > ```ts
> > > export const contentRefs = {
> > >   event: (id: number) => `event/${id}`,
> > >   comment: (threadId: number) => `comment/${threadId}`,
> > >   // …one per type
> > > };
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Call sites become `contentRef={contentRefs.event(event.eventId)}`. You make a custom function like this for discussions, but the rest are now scattered strings. Makes sense to make this pattern reusable for all the other ones IMO.
> > 
> > 
> > can i do that in a follow-up?
> 
> Is this PR really a rush to get in?

done!

nabramow (dismissed):
You've got some merge conflicts but code looks good and I tested and all works well now.

PR #9405, opened 2026-07-31
i18n: Remove dedicated fr-CA translation

tristanlabelle:
No one has contributed strings to the fr-CA translation and I wouldn't want to dilute translators efforts by starting fr-CA. French and Canadian French translators can agree on using mutually intelligible language in a single "fr" translation.

There are two ways I could see bringing this back in the future:

1. We have lots of translator capacity and specific demand from Quebec.
2. We add fr-CA as a locale choice for regional settings (e.g. datetimes), but it still uses the "fr" translations, similar to what we want to do with "en" vs "en-US".

There should be no users today using `fr-CA` since translation progress is 0%. I'm still improving the behavior for unknown `user.ui_language_preference` in #9407.

## Testing
Ran local backend+frontend without this change, switched to fr-CA, then ran local backend+frontend with this change, and saw that it automatically switched to English.

Clicked around on Vercel preview:

image

**Backend checklist**

**Web frontend checklist**

## For maintainers


app/backend/src/couchers/email/locales/fr-CA.json
tristanlabelle, line 1: Self-TODO: Remove language from Weblate before merging

nabramow (dismissed):
As long as you tested it and the app runs without issue everything looks okay to me!

Will this automatically take it out of the TranslationProgress screen? That's the only thing I wasn't sure if I should see here in the files.

tristanlabelle:
> Will this automatically take it out of the TranslationProgress screen? That's the only thing I wasn't sure if I should see here in the files.

It's already gone in the Vercel preview and will be even more gone once I remove the language from Weblate since we get those progresses from the Weblate API.

tristanlabelle:
There's an ubuntu outage that caused the failure of `preview:protos`. Since I'm not modifying protos, I'll proceed with merging. https://gitlab.com/couchers/couchers/-/jobs/15702697434

app/backend/src/couchers/email/locales/fr-CA.json
tristanlabelle: Done as of just now

PR #9406, opened 2026-07-31
implement EndDate autofill

### LINKED ISSUE #9396: Autofill event end date on initial start date enter
It's annoying having to fill it out every time.

* When the user enters event `startDate` the first time, autofill the endDate to be one day later (only if endDate is empty).
* When user fills `startTime`, make the `endTime` initially the same as `startTime` (if it's empty)
* User needs to be able to change the date and time later with no issue (only that initial fill happens when it's blank)

For example:

Image

(The thought is, the form won't let it submit with the same endTime, but since most events are usually 2-3 hours, it's less scrolling to adjust the time this way)

nc225:
Closes https://github.com/Couchers-org/couchers/issues/9396 
Resolving [issue](https://github.com/Couchers-org/couchers/issues/9396). Note that the issue says "autofill the endDate to be one day later" but after discussion this autofill to the **same date** since that's the more common use case. 

## Testing
Tested local version and works ok. See screenshot in issue.

**Web frontend checklist**

## For maintainers


nabramow:
Noticed one thing, when I select a startDate, endDate fills fine, then when I select a startTime of 12:00pm, it fills an endTime of 12:00am instead of 12:00pm:

Screenshot 2026-07-31 at 18 54 41

Same with 2pm, it's setting 2am:

Screenshot 2026-07-31 at 18 55 59

app/web/features/communities/events/EventTimeChanger.tsx
nabramow, line 80: I think we should set `startTime` first and then do the login to set `endTime`.

nabramow:
Also just an FYI for the future if you put an x in these checkboxes it will change it to a checkmark on save :):

`[x] Some list item`
- [x] Some list item

Screenshot 2026-07-31 at 18 57 53

nabramow:
Also please write "Closes #TICKETNUMBER" in this case Closes #9396 in the description. This will automatically mark that ticket as closed when it merges.

nc225:
@nabramow thanks for catching that issue with the AM/PM- should be fixed now

nabramow:
@nc225 Looks like you've got some failing tests now!

nc225:
@nabramow  can you check the tests now? It looks like 1 check was skipped - I'm not sure why. If it's still failing, it might be this test but I'm a bit confused by the test description--what does it mean by "should make user manually fix times when the start time updates"? I suspect it means "if endTime is set and startTime changes, endTime should unset." If that's needed, let me know and I will try to implement. 
image

nabramow:
> @nabramow can you check the tests now? It looks like 1 check was skipped - I'm not sure why. If it's still failing, it might be this test but I'm a bit confused by the test description--what does it mean by "should make user manually fix times when the start time updates"? I suspect it means "if endTime is set and startTime changes, endTime should unset." If that's needed, let me know and I will try to implement. image

I think we just need to change the test to the new behavior. So change the test to check that the endDate and endTime are set when the startDate and startTime are set. This one is checking the old behavior of them manually filling it out.

As long as there is then also a test to check they are unable to submit with same startDate startTime as endDate endTime then everything is covered.

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -67,6 +74,17 @@ export default function EventTimeChanger({
       shouldDirty: true,
       shouldValidate: true,
     });
+    const endTime = getValues("endTime");
+    const userChangedMeridiem =
+      endTime && value != undefined
+        ? Math.abs(value.until(endTime).total("hours")) == 12
+        : false;
```
tristanlabelle: This bit is unexpected to me and not mentioned in the PR description. I doubt its value, but if it has some then it needs a comment.

app/web/features/communities/events/EventTimeChanger.tsx
```diff
@@ -67,6 +74,17 @@ export default function EventTimeChanger({
...
+    if (!endTime || userChangedMeridiem) {
+      setValue("endTime", value, {
```
tristanlabelle, line 74: I'm also surprised here. We default to filling in the same end date, and start time = end time, which is an invalid combination.

app/web/features/communities/events/EventTimeChanger.tsx
nc225: @tristanlabelle I've added a comment to explain. A more straightforward option IMO would be to alter Datepicker so that in addition to onPostChange, we have access to the onAccept callback. That way,  this autofill could happen just when the user OK'ed--not just changed--the startTime, and we wouldn't need userChangedMeridiem. I was just hesitant to change Datepicker as part of this PR.

app/web/features/communities/events/EventTimeChanger.tsx
nc225: Is [this](https://github.com/Couchers-org/couchers/issues/9396#issuecomment-5137360570) similar to what you're surprised by here or am I misunderstanding? We discussed not filling in endTime at all vs filling that creates an invalid combination and seemed like the preference was for invalid combination and less scrolling.

app/web/features/communities/events/EventTimeChanger.tsx
tristanlabelle: Ah I understand, we get two callbacks so this is a heuristic to only run the logic on the latter callback. Hacky but ok for now.

nabramow: I don't think it's a big deal to add an onAccept prop to the Datepicker component as long as you make it optional so it's backwards compatible.

nabramow:
@nc225 You've got a failing test, but once that's sorted this looks good!

app/web/features/communities/events/EventTimeChanger.test.tsx
```diff
@@ -422,24 +414,42 @@ describe("Event time changer", () => {
...
-      await user.keyboard("0300 PM");
+      await user.keyboard("1100 PM");
+
+      await user.click(screen.getByTestId("submit"));
+      expect(onValidSubmit).toHaveBeenCalled();
+      const submittedData = onValidSubmit.mock.calls[0][0];
+      expect(submittedData).toBeDefined();
+
+      expect(submittedData.endDate).toEqual(Temporal.PlainDate.from("2021-08-05"));
+      expect(submittedData.endTime).toEqual(Temporal.PlainTime.from("22:00"));
```
nabramow: Shouldn't this be 23:00? 11pm is 23:00 in military time.

nabramow: Also from what I read `toEqual` on Temporal objects always passes because the Temporal objects hide their data in private slots so it just compares the name. You can test this by putting something obviously wrong and checking (though that's probably why this passed). Fix by comparing strings instead, that's what the picker tests already do.

app/web/features/communities/events/EventTimeChanger.test.tsx
```diff
@@ -422,24 +414,42 @@ describe("Event time changer", () => {
...
 
-      expect(onValidSubmit).toHaveBeenCalledTimes(1);
-      const submittedData = onValidSubmit.mock.calls[0][0];
-
-      expect(submittedData.startDate).toEqual(Temporal.PlainDate.from("2021-08-05"));
-      expect(submittedData.startTime).toEqual(Temporal.PlainTime.from("14:00"));
-      expect(submittedData.endDate).toEqual(Temporal.PlainDate.from("2021-08-06"));
-      expect(submittedData.endTime).toEqual(Temporal.PlainTime.from("15:00"));
+      // Should not submit due to same date/time as start
+      expect(onValidSubmit).not.toHaveBeenCalled();
```
nabramow, line 453: I believe this test would pass even with the feature deleted. It only checks `onValidSubmit` wasn't called, but an empty form doesn't submit either. Assert the actual error text (`communities:end_time_error`) so it fails for the right reason.

app/web/features/communities/events/EventTimeChanger.test.tsx
```diff
@@ -392,26 +392,18 @@ describe("Event time changer", () => {
     });
   });
 
-  describe("when the end date/time difference from the start has been changed", () => {
-    it("should make user manually fix times when the start date/time updates", async () => {
```
nabramow, line 396: Let's add this test back, I think the title just needs to be updated, but the behavior we're asserting is still relevant:

What the branch changed: empty end fields now get filled. What didn't change: once the user has typed an end date/time, editing the start leaves them alone.

We do want to test this.

Can rename "should not overwrite an end date/time the user already set" or similar.

app/web/features/communities/events/EventTimeChanger.test.tsx
nc225: Fine with either. I used "1100 PM" to test the typing behavior as the previous test did. The previous test typed "0200 PM" (as opposed to 14:00).

nc225: Ah I didn't realize that about toEqual, thanks for the heads up and will fix

app/web/features/communities/events/EventTimeChanger.test.tsx
nc225: Sounds good. I added back one test for end date and one for end time.

nabramow (approved):
Okay I think this is good to go!

Issue #9184, opened 2026-06-26
I18n: Split international English and US English for less surprising date formats.


tristanlabelle:
The MM/DD/YYYY formats are a confusing for the rest of the world, and have caused bad birthdate entries before. We worked around it by preventing numerical date input, but English formatting still follows the US standard.

`M/DD/YYYY`: en, en-US
`DD/MM/YYYY`: en-UK, en-AU, en-NZ, en-DK, en-001 (international), en-150 (European)
`DD/M/YYYY`: en-IE, en-IN
`YYYY-MM-DD`: en-CA

# Proposal 
Offer two locales:

- "English" = `en-001`, aka English (International)
- "English (US)" = `en-US`

Keep a single set of translations in Weblate under `en`. Existing users using `en` get remapped to `en-001`.

PR #9407, opened 2026-07-31
Backend/i18n: Support en-US locale, without a dedicated translation

tristanlabelle:
Broadens the concept of supported locales to allow for a first supported locale without dedicated translations: en-US.

Couchers has a lot of international users using the English default language but getting confused by `MM/DD/YYYY` date formats. We want to split into `English (US)` and `English (International)` (default), but not duplicate translations.

The "en" locale follows US regional formats, whereas "en-001" is "English (International)" and uses `DD/MM/YYYY`. However we can't only support "en-US" and "en-001" because we need to support the "en" generic language. My solution will be to special-case "en" to resolve to the "en-001" babel locale, which we use for regional formatting, but I change the "en" behavior until the frontend offers both English variants.

As part of this it becomes more important for i18n operations to support fallbacks. Our string lookup is already setup to take a list, so `["en-US", "en"]` will match "en". I've done something similar for other regional formatting functions, though in practice it seems like Babel/the CLDR has data for all of our current locales doesn't need to fallback.

The frontend doesn't have a way to select en-US so there are no functional changes for now.

Related to #9184

## Testing
Added/updated tests. Ran backend + frontend locally and switched between English and German.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/calendar_events.py
+++ b/app/backend/src/couchers/email/calendar_events.py
@@ -155,7 +155,7 @@ def _set_datetime_with_timezone(event: Event, name: str, dt: datetime) -> None:
 
 def ics_event_to_calendar(event: Event, method: str | None, loc_context: LocalizationContext) -> Calendar:
     # PRODID is mandatory and generally follows "-//[Organization]//[Product Name]//[Language]"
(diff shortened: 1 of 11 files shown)


aapeliv (approved):
quick look through, seems appropriate

PR #9433, opened 2026-08-02
Frontend/i18n: Don't show event minutes if 00

### LINKED ISSUE #9358: Frontend/i18n: Omit datetime minutes when not necessary
For example this could have "9 - 10 AM" since all minutes are zero

Image

tristanlabelle:
Event datetime ranges can get quite verbose, but we can save some brain cells when they start at minute 0 of the hour by not showing the minutes field.

Fixes #9358

## Testing
Vercel preview:

image

Next:

image

Other examples from Vercel preview:

image

**Web frontend checklist**

## For maintainers


aapeliv:
what does this look like with other time formats? i think it looks fine with am/pm; but i wonder if it looks kind of orphaned with 24-hour time, for instance?

tristanlabelle:
@aapeliv yikes, good call! It looks very off:

image

I might be able to condition this on the presence of AM/PM but maybe I'll give up on this.

PR #9437, opened 2026-08-02
Backend: dump sample emails as a single self-contained page

aapeliv:
The backend can render a sample of every notification email in every supported locale, and CI publishes the browsable index to the preview environment. Writing an HTML and a plaintext body per email per locale put a few thousand small files into that output, all of which CI uploads to the preview bucket on every pipeline. The bodies are now gzipped and inlined into the index itself, leaving a single page plus the attachment images; inlining every locale rather than fetching payloads on demand is deliberate, as the index is opened every few weeks and its page weight matters far less than the per-pipeline S3 cost.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/email/dump_emails_index.html.jinja2
+++ b/app/backend/src/couchers/email/dump_emails_index.html.jinja2
@@ -51,12 +51,12 @@
 </summary>
 <div class="panes">
   <section>
-    <h2>HTML <a data-href="{{ r.html_filename }}" target="_blank">open</a></h2>
-    <iframe data-file="{{ r.html_filename }}" loading="lazy"></iframe>
+    <h2>HTML <a data-kind="html" target="_blank">open</a></h2>
+    <iframe data-kind="html" loading="lazy"></iframe>
   </section>
   <section>
-    <h2>Plaintext <a data-href="{{ r.plaintext_filename }}" target="_blank">open</a></h2>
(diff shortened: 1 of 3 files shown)


app/backend/src/couchers/email/dump_emails_index.html.jinja2
```diff
@@ -51,24 +51,48 @@
...
+    <h2>Plaintext <a data-kind="txt" target="_blank">open</a></h2>
+    <iframe data-kind="txt" loading="lazy"></iframe>
   </section>
 </div>
 </details>
 {% endfor %}
 </main>
 <script>
 const SUBJECTS = {{ subjects_json }};
+const PAYLOADS = {{ payloads_json }};
```
tristanlabelle, line 67: With this idea of gzipped data, we could move away from templating much of this file and just have a contract for a json data structure that includes locales, subjects and bodies, that all gets gzipped into a single javascript string, from which the page is initialized dynamically, instead of having a hybrid between 1. templated html, 2. templated subjects json and 3. templated gzipped payloads.

tristanlabelle (approved):

app/backend/src/couchers/email/dump_emails_index.html.jinja2
aapeliv: yeah i agree, that would be better. though i think locale-specific gzips is better perf wise

PR #9438, opened 2026-08-02
Web: Fix bugs#33 - Navbar Icon placement too high

papeldeorigami:

Closes Couchers-org/bugs#33 - Navbar Icon placement too high

Root cause: when the address bar is moved to the top on some iOS devices with a notch at the top, the bottom navigator shrinks because it does not have a min height defined.

## Testing

How-to | Image
-- | --
To force the address bar at the top, go to Settings --> Apps --> Safari and, in the section Tabs, select Top. _(probably only on new devices with iOS 26)_  | Simulator Screenshot - iPhone 17 - 2026-08-03 at 00 11 25

#### Before:
 https://github.com/user-attachments/assets/5dc089a7-d117-4486-b553-d6eeeef59e0a 
#### After:
 https://github.com/user-attachments/assets/c9c4cb3b-d5c3-412a-a948-eaac112be045

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/Navigation/BottomNavigation.tsx
+++ b/app/web/components/Navigation/BottomNavigation.tsx
@@ -6,6 +6,8 @@ import { useTranslation } from "next-i18next";
 import { useMemo } from "react";
 import { communitiesRoute, dashboardRoute, eventsRoute, messagesRoute, searchRoute } from "routes";
 
+import { BOTTOM_NAV_BASE_HEIGHT } from "./constants";
+
 const StyledPaper = styled(Paper)(() => ({
   position: "fixed",
   bottom: "var(--cookie-banner-height, 0px)",
@@ -16,6 +18,8 @@ const StyledPaper = styled(Paper)(() => ({
 }));
 
 const StyledBottomNavigation = styled(MuiBottomNavigation)(({ theme }) => ({
+  // the height must account for the safe area (e.g. iOS notch) + MUI base height
+  height: `calc(${BOTTOM_NAV_BASE_HEIGHT}px + env(safe-area-inset-bottom, 0px))`,
   paddingBottom: "env(safe-area-inset-bottom, 0px)",
   "& .MuiBottomNavigationAction-root": {
     minWidth: "auto",
--- a/app/web/components/Navigation/LoggedInMenu.tsx
+++ b/app/web/components/Navigation/LoggedInMenu.tsx
@@ -45,7 +45,9 @@ const StyledMenu = styled(Menu, {
 })<{ $isNativeEmbed?: boolean }>(({ theme, $isNativeEmbed }) => {
   // Native embed: full height (native tabs handle safe area)
   // Mobile web: subtract bottom nav + safe area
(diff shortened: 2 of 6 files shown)


app/web/components/Navigation/BottomNavigation.tsx
```diff
@@ -16,6 +18,8 @@ const StyledPaper = styled(Paper)(() => ({
 }));
 
 const StyledBottomNavigation = styled(MuiBottomNavigation)(({ theme }) => ({
+  // the height must account for the safe area (e.g. iOS notch) + MUI base height
+  height: `calc(${BOTTOM_NAV_BASE_HEIGHT}px + env(safe-area-inset-bottom, 0px))`,
```
papeldeorigami, line 22: Note for the reviewer: this is the actual fix. Everything else is tests and refactor for readability.

app/web/components/Navigation/LoggedInMenu.tsx
```diff
@@ -44,13 +45,15 @@ const StyledMenu = styled(Menu, {
 })<{ $isNativeEmbed?: boolean }>(({ theme, $isNativeEmbed }) => {
   // Native embed: full height (native tabs handle safe area)
   // Mobile web: subtract bottom nav + safe area
-  const menuHeight = $isNativeEmbed ? "100vh" : "calc(100vh - 56px - env(safe-area-inset-bottom, 0px))";
+  const menuHeight = $isNativeEmbed
+    ? "100vh"
+    : `calc(100vh - ${BOTTOM_NAV_BASE_HEIGHT}px - env(safe-area-inset-bottom, 0px))`;
```
nabramow, line 50: Would it make more sense to set style for small screen in general here, rather than conditionally check for isNativeEmbed? People could be checking on the mobile browser too.

Like this I mean:

```typescript
[theme.breakpoints.down("md"): {
// mobile style here
}
```

I know there is a comment about this but I still wonder if it makes sense to have two different logics?

papeldeorigami: @nabramow maybe, but let me see if I understand the idea first (thanks for sharing the example BTW): the idea is to check the breakpoint and have a single logic for the height for mobile size (no matter if running on browser or native app), right?

Well, if you look at [line 60](https://github.com/Couchers-org/couchers/blob/0a24bf03a17774adbb75afe14e1c568a54da7201/app/web/components/Navigation/LoggedInMenu.tsx#L60) below, we already check for the md breakpoint. I mean, this menuHeight variable is only applicable for mobile (either browser or native app):
```
        ...
        height: menuHeight,
        ...
```

Furthermore, I think this logic must stay here, because:
- mobile browser: renders the bottom navigator with the page, so we have to subtract it from the height
- mobile app: has its own navigation bar, so we need to fill the entire view

From that perspective, I'd recommend keeping the `isNativeEmbed` validation.

Note that this logic is prior and unrelated with the fix. My intention with touching this line was just to improve readability (by replacing the magic number 56px with the constant for the bottom navigation), because honestly I found it hard to understand at first sight.

Let me know if I missing something, I'd be happy to simplify this logic too as it is spread through many files.

nabramow (approved):
This looks good to me, thanks for looking into the breakpoints and for the great testing notes and screenshares! Makes it so much easier to review!

PR #9445, opened 2026-08-03
Allow starting over from verification page

### LINKED ISSUE #9092: Frontend/signup: Can't change a bad email address
Once you start the signup flow, https://next.couchershq.org/signup will always send you to

Image

at which point you're stuck if the email you had originally typed was incorrect.

We need a way to clear the cookie. Maybe by asking them to reenter the email address if they need a new email sent.

Alternatively we can change the url after the initial name/email page such that `/signup` always starts the process over, and `/signup-flow` or whatever resumes it.

See user bug Couchers-org/bugs#32

nc225:
Closes #9092 

A few notes:

**1. No email address shown**

The linked issue suggests that we
> Display email address on that page, 

However, here we do not display the email address, because email is not currently available via auto.proto file. If we make email available from auth_pb.js alongside flowToken, then we can eventually incorporate email address here as well.

**2. Prior verification link**
If someone clicks the link to start over, their prior verification token/email is still valid so they could end up with two distinct accounts if they somehow have access to both emails. LMK if we want to invalidate the token as well, though I don't see it posing a major issue. 

**3. Handling incomplete signups**
> new signups must not be blocked if there's an existing incomplete signup with that same email

Keeping the current behavior here: if a new signup tries to use the email address of an existing incomplete signup, there is a message to "Please check your email for a link to continue signing up" and an email is sent with a link that picks up to wherever they left off in the signup process. 

**4. Suggestion for revised SignupFlow**
Finally, I do think signup flow could benefit from a "back" button (What if you decide you want to change your Hosting Status before submitting? This is easy to do once your account is active, but still.)
## Testing
Tested locally in a few different scenarios, seems to work fine. I wasn't sure which automated tests to add so I haven't made any but happy to talk about those.

**Backend checklist**
- [X] Run the backend locally and it works

**Web frontend checklist**

## For maintainers


app/web/features/auth/locales/en.json
```diff
@@ -176,6 +176,7 @@
   "sign_up_need_verification_title": "One last thing: confirm your email.",
   "sign_up_resend_verification_email_help": "Didn't receive the email? Click <resendLink>here to resend the verification link</resendLink>.",
   "sign_up_resend_verification_done": "Done! We've sent you another email.",
+  "sign_up_restart_signup": "Mistype your email address? Click <restartSignup>here to start over<restartSignup>.",
```
papeldeorigami: Missing a closing `/` here:
```suggestion
  "sign_up_restart_signup": "Mistype your email address? Click <restartSignup>here to start over</restartSignup>.",
```
Also nit: shouldn't it be "Mistyped"? (though not a native speaker here)

app/web/service/auth.ts
```diff
@@ -144,6 +144,13 @@ export async function signupFlowResendVerificationEmail(flowToken: string) {
   return res.toObject();
 }
 
+export async function signupFlowRestartSignup(flowToken: string) {
+  const req = new SignupFlowReq();
```
papeldeorigami: I honestly couldn't figure out what this code is doing. As far as I could dig it might be just returning data about the current flow... so if that's the case, I don't see why we should name it Restart Signup - it's totally misleading. Plus - are you sure we need it? I mean, if it is not restarting the flow, then we probably need to find a different approach to this problem.

app/web/features/auth/signup/ResendVerificationEmailForm.tsx
```diff
@@ -11,43 +11,56 @@ import { service } from "service";
...
       const state = await service.auth.signupFlowResendVerificationEmail(authState.flowState!.flowToken);
       authActions.updateSignupState(state);
       setResent(true);
     },
   });
-
+  const mutationRestart = useMutation({
+    mutationFn: async () => {
+      const state = await service.auth.signupFlowRestartSignup(authState.flowState!.flowToken);
+      state.needBasic = true;
```
papeldeorigami: This looks like faking the BE response; it seems connected with the fact that the Restart Signup method above is actually not doing anything.
Sidenote: need basic means that the name/e-mail are missing.

papeldeorigami (changes requested):
Hi, @nc225! Please check the points below. I wonder if we could take a simpler approach with just deleting the flowToken while we don't have a proper BE mechanism.

Also, if we really want to show the e-mail (which I also missed when testing), then we could consider caching it locally. I don't see a big deal in storing the e-mail temporarily, as long as we delete it after the flow is completed.

Edit: In my original comment, I wasn't sure if we still want to merge the PR (because in my mind perhaps there is another approach being considered), but I just realized you have tagged me recently; sorry for the delay. Let's do it! 😊

app/web/features/auth/locales/en.json
nc225: Good catch! I'll fix that

app/web/features/auth/signup/ResendVerificationEmailForm.tsx
nc225: Yeah, I think my approach was to taken the current flow, send back to needBasic, and then a new flowToken gets reassigned when Basic Form is submitted. It sounds like you think a better approach would be to clear flowToken and start again? If so happy to try moving in that direction. Yes keeping email in localStorage sounds fine too.

app/web/service/auth.ts
nc225: The idea here was to retrieve whatever the current state is without otherwise acting on it, so that we can set its needBasic to true

nc225:
@papeldeorigami  Thanks for reviewing!
> I wonder if we could take a simpler approach with just deleting the flowToken while we don't have a proper BE mechanism.

I've changed this PR to use this approach. I added a restartSignup() function to authStore and then used that instead. 

> if we really want to show the e-mail (which I also missed when testing), then we could consider caching it locally. I don't see a big deal in storing the e-mail temporarily, as long as we delete it after the flow is completed.

Done. One issue: this isn't clearing that localStorage value on successful signup, so it sticks around. I'm not sure the best place to clear it, if necessary. Do you have any ideas about that?

I also modified Signup.test to work with the new approach. I ran into a problem with testing_library failing to match on a variable inside `<strong>` tags and solved it one way, but if you have any ideas for other solutions let me know.

app/web/features/auth/locales/en.json
```diff
@@ -176,7 +176,8 @@
   "sign_up_need_verification_title": "One last thing: confirm your email.",
   "sign_up_resend_verification_email_help": "Didn't receive the email? Click <resendLink>here to resend the verification link</resendLink>.",
   "sign_up_resend_verification_done": "Done! We've sent you another email.",
-  "sign_up_completed_prompt": "We have sent an email with a verification link to your email address. Please click the link to activate your account.",
+  "sign_up_restart_signup": "Mistyped your email address? Click <restartSignup>here to start over<restartSignup>.",
```
papeldeorigami: It looks like the `<restartSignup>` is still missing the enclosing `</restartSignup>`

app/web/features/auth/signup/BasicForm.tsx
```diff
@@ -45,6 +45,8 @@ export default function BasicForm({ submitText, successCallback, inviteCode }: B
     mutationFn: async (data) => {
       const sanitizedEmail = lowercaseAndTrimField(data.email);
       const sanitizedName = data.name.trim();
+      // locally store email for recall on verification page
+      localStorage.setItem("signupEmail", sanitizedEmail);
```
papeldeorigami: I would prefer to avoid this type of direct access to localStorage.
- Instead, we should be using [usePersistedState](https://github.com/Couchers-org/couchers/blob/cd909838111d41860f67203da203c1400dfb7e06/app/web/platform/usePersistedState.ts#L46) for consistency.
- You could put that persistence logic inside useAuthStore, maybe something like "setSignupEmail".
- The key would scoped under auth.signupEmail, which I would find much nicer 😊.

app/web/features/auth/signup/ResendVerificationEmailForm.tsx
```diff
@@ -11,24 +11,34 @@ import { service } from "service";
...
-      {mutation.error && <Alert severity="error">{mutation.error.message || ""}</Alert>}
+      {mutationResend.error && <Alert severity="error">{mutationResend.error.message || ""}</Alert>}
+      {mutationRestart.error && <Alert severity="error">{mutationRestart.error.message || ""}</Alert>}
       <Typography variant="body1" gutterBottom>
-        {t("auth:sign_up_completed_prompt")}
+        {
+          <Trans
+            i18nKey="auth:sign_up_completed_prompt"
+            values={{
+              providedEmailAddress: localStorage.getItem("signupEmail"),
```
papeldeorigami: If we go with useAuthStore, we could be getting the authState.signupEmail here.

app/web/features/auth/signup/Signup.test.tsx
```diff
@@ -446,11 +446,24 @@ describe("Signup", () => {
...
+    // Removing <strong> tags so the translated message can be compared with the element's textContent.
+    const message = t("auth:sign_up_completed_prompt", {
+      providedEmailAddress: testEmail,
+    }).replace(/<\/?strong>/g, "");
+
     render(<View />, { wrapper });
-    expect(screen.getByText(t("auth:sign_up_completed_prompt"))).toBeVisible();
+    expect(
+      screen.getByText((_, element) => {
+        return element?.textContent === `${message}`;
```
papeldeorigami: Maybe we should just do this plain text comparison, like other tests in this codebase:
```suggestion
        return element?.textContent === `We have sent an email with a verification link to your email address: ${testEmail}. Please click the link to activate your account.`:
```

app/web/features/auth/useAuthStore.ts
```diff
@@ -143,6 +143,10 @@ export default function useAuthStore() {
           return;
         }
       },
+      async restartSignup() {
+        setFlowState(null);
+        setError(null);
```
papeldeorigami: Here is one of the places where we should clear the signupEmail (with the `usePersistedState` approach).
The other is a few lines above, inside `updateSignupState`.

papeldeorigami (changes requested):
@nc225 Good progress! Here is a screenshot. See comments below for some ideas.
Image

nc225:
@papeldeorigami I made those changes, thanks for reviewing

app/web/features/auth/signup/Signup.test.tsx
nc225: I was stuck on this test for a while until I realized I had to use JSON.stringify even for a string like testEmail in `localStorage.setItem("auth.signupEmail", JSON.stringify(testEmail));` Now it works.

app/web/features/auth/useAuthStore.ts
```diff
@@ -139,10 +140,19 @@ export default function useAuthStore() {
         setFlowState(state);
         if (state.authRes) {
           setFlowState(null);
+          setSignupEmail(null);
           authActions.firstLogin(state.authRes!);
           return;
         }
       },
+      setSignupEmail(email: string) {
+        setSignupEmail(email);
```
papeldeorigami, line 149: This feels like a recursive function call due to the name clash, even though it works 😅. Perhaps we should name the persisted state setter differently, like setSignupEmailState. WDYT?

app/web/features/auth/signup/ResendVerificationEmailForm.tsx
```diff
@@ -11,24 +11,34 @@ import { service } from "service";
...
+    },
+  });
   return (
     <>
-      {mutation.error && <Alert severity="error">{mutation.error.message || ""}</Alert>}
+      {mutationResend.error && <Alert severity="error">{mutationResend.error.message || ""}</Alert>}
+      {mutationRestart.error && <Alert severity="error">{mutationRestart.error.message || ""}</Alert>}
       <Typography variant="body1" gutterBottom>
-        {t("auth:sign_up_completed_prompt")}
+        {
```
papeldeorigami: We probably don't need the `{}` here

papeldeorigami (commented):
@nc225 LGTM and we could merge as it is, but how about these last almost-nitpicking suggestions?

app/web/features/auth/useAuthStore.ts
nc225: Ah good point yeah let's change the name. Separate q: I notice useAuthStore has a mix of async and sync functions. I think the two I'm making are both async actually so I'm doing that, but LMK if that's not right!

nc225:
> @nc225 LGTM and we could merge as it is, but how about these last almost-nitpicking suggestions?

Thank you for taking a look and for the great suggestions!

app/web/features/auth/useAuthStore.ts
darrenvong: They only really need to be async functions if we need to await for things inside the function itself, or if the function returns a Promise. Otherwise it is misleading that they are written as async functions...

papeldeorigami (changes requested):
@nc225 apparently we missed something basic and critical here - we can even try to change the e-mail, but the username is already taken when we restart the flow only in the FE.

Image

I am afraid that we might need some change in the BE to cancel a signup flow first.
Is this something you checked or discussed before?

nc225:
@papeldeorigami ahhh I had not caught that, thanks for thinking of it! It's related to something I noted in the initial PR but you're right, this is a bigger issue than I realized at the time because people will want to use the same username:

> If someone clicks the link to start over, their prior verification token/email is still valid so they could end up with two distinct accounts if they somehow have access to both emails. LMK if we want to invalidate the token as well, though I don't see it posing a major issue.

Let me look at the BE a little more to see how this might work. We would need to invalidate the existing token somehow bc we don't want to allow two accounts with the same username. One solution would be twofold: invalidate the existing email token somehow, and add a BE like "signup_cancelled" object to SignupFlow so that when it checks this in servicers/auth.py:

 ```
 signup_exists = (
        session.execute(select(SignupFlow)
       .where(SignupFlow.username == username))
       ADD: .where(!SignupFlow.signup_cancelled))_**
      .scalar_one_or_none() is not None
    )
```

Will think about it a bit more -if you have thoughts please LMK. Moving to drafts for now =/

File: app/backend/src/couchers/servicers/auth.py [hunk 11]
```diff
@@ -274,10 +278,16 @@ def SignupFlow(
             else:
                 # not fresh signup
                 flow = session.execute(
-                    select(SignupFlow).where(SignupFlow.flow_token == request.flow_token)
+                    select(SignupFlow).where(SignupFlow.flow_token == request.flow_token).where(SignupFlow.signup_cancelled == False)
                 ).scalar_one_or_none()
                 if not flow:
                     context.abort_with_error_code(grpc.StatusCode.NOT_FOUND, "invalid_token")
+                if request.cancel_signup:
```
nc225: Here I am adding the `signup_cancelled` flag when a signup is cancelled. Another option would be to use session.delete() but it seems we generally prefer adding info to destroying it - happy to go with session.delete() though since adding the flag this requires adding a few lines elsewhere too.  Also, I thought it was a bit confusing to use the same name for the proto and the postgres variables so I called one `signup_cancelled` and the other `cancel_signup` but if this, too, is confusing, I will standardize.

nc225:
(I am deleting a previous comment bc I figured out the problem -- I somehow ended up with duplicate migration files and that made the backend unhappy.)

I've got this working in some form but I'd like to add more tests to the BE section before restoring from drafts. One BE question that I marked in the code too: when a signup is cancelled due to bad email, should we 
- add a `signup_cancelled` flag to Signup_Flows and check for that as needed, or 
- use session.delete() to just drop the Signup_Flow as we do after successful completion? or
- some other third option?

I've gone with the first one because it seems a bit clearer, but it also means checking for `signup_cancelled` a few different places, and I could see going either way. 

The PR is a bit complicated right now, and I may explore whether a different path (say, letting the user change their email at/after the verification screen instead of restarting) may work better.

app/backend/src/couchers/migrations/versions/0187_allow_starting_over_from_verification_.py
```diff
@@ -0,0 +1,65 @@
...
+        table_name="signup_flows",
+        postgresql_where=sa.text("username IS NOT NULL AND signup_cancelled = false"),
+    )
+    op.create_unique_constraint(
+        op.f("uq_signup_flows_username"), "signup_flows", ["username"], postgresql_nulls_not_distinct=False
+    )
+<<<<<<< HEAD
+    op.drop_column("signup_flows", "signup_cancelled")
+=======
+    op.execute("DELETE FROM signup_flows WHERE signup_cancelled")
```
nc225: Note that this will drop any `signup_cancelled` rows from the database upon downgrade so that `username` remains unique

papeldeorigami:
@nc225 I think it's fine to continue the work here as a draft, but I would recommend closing it and splitting the BE and FE work before you ask for a new round of reviews so people can take a specialized look.

nc225:
Since we have to make BE changes anyhow to prevent mistyped emails from blocking usernames in SignupFlow, I am reworking this to use option 3 from [this comment](https://github.com/Couchers-org/couchers/issues/9092#issuecomment-4985326942) instead of option 2 as I had prior. Here are a couple screenshots in development of what the front end might look like on the email verification screen - open to any design tips. (Also forgive the still-misspelled Mistype**d**) I will also split this into FE and BE PRs.

Separate convo: do we want to limit the number of times a Signup can change its email address? 

Screenshot From 2026-08-25 00-56-14
Screenshot From 2026-08-25 00-56-31

nabramow:
hey @nc225 just popping on here as these go in my email - FYI we normally try to break frontend and backend changes into separate PRs for easier review.

nc225:
Since this involves BE changes, closing in favor of these two PRs:
FE: https://github.com/Couchers-org/couchers/pull/9626
BE: https://github.com/Couchers-org/couchers/pull/9627

PR #9449, opened 2026-08-04
Frontend dev improvments

nabramow:
I got tired of the browser errors for things that don't work on dev environment. 

## Testing

**Web frontend checklist**

## For maintainers


app/web/next.config.js
```diff
@@ -15,14 +15,27 @@ const nextConfig = {
...
     }
     config.module.rules.push({
       test: /\.md$/,
       loader: "frontmatter-markdown-loader",
     });
+    if (dev) {
+      // Next defaults to "eval-source-map", which holds a full source map for every
+      // module in memory. Line-accurate mapping is enough for dev and costs much less.
+      config.devtool = "eval-cheap-module-source-map";
+    }
```
darrenvong: I think this is redundant since I'm getting this warning in my console on startup:
```
Warning: Reverting webpack devtool to 'false'.
Changing the webpack devtool in development mode will cause severe performance regressions.
Read more: https://nextjs.org/docs/messages/improper-devtool
```

The why (according to Claude, seems legit as it's digging into the source code in `node_modules` and matches the warning that gets output in my console:

> Then, after your custom `webpack()` runs and sets eval-cheap-module-source-map, Next detects the change and force-reverts it:
>
> Shows relevant source code in `webpack-config.js`:
>
> ```typescript
> if (dev && originalDevtool !== webpackConfig.devtool) {
>    webpackConfig.devtool = originalDevtool;
>    devtoolRevertWarning(originalDevtool);
>}
> ```
> That's why the warning says "Reverting webpack devtool to 'false'" — false (+ their plugin) is the real default now, and your override never takes effect in any compiler. So there's no trade-off to weigh: the deliberate choice is dead code, the memory saving it claims isn't happening, and the only observable effect is the warning noise. The official advice wins by default here.

app/web/next.config.js
```diff
@@ -15,14 +15,27 @@ const nextConfig = {
...
   // transpile these themselves rather than treating them as pre-built node_modules.
   transpilePackages: ["temporal-polyfill", "temporal-utils"],
-  webpack: (config, { isServer }) => {
+  experimental: {
+    // Trades slightly slower compiles for a lower webpack memory ceiling.
+    webpackMemoryOptimizations: true,
+    // Dev only: pages get compiled on demand anyway, so preloading every entry just
+    // inflates the footprint. In prod this is what keeps the first response per route
+    // fast, so leave it on there.
+    ...(process.env.NODE_ENV === "development" ? { preloadEntriesOnStart: false } : {}),
```
darrenvong: If pages are compiled on demand anyway, disabling preloading is basically a no-op too so we don't need this?

Claude found the relevant code in the source again in `next-server.js` L534 🧠: 

```typescript
if (!options.dev && !this.minimalMode && this.nextConfig.experimental.preloadEntriesOnStart) {
```

So that flag only gets read in non-dev environments

app/web/pages/_document.tsx
```diff
@@ -15,11 +15,14 @@ export default function MyDocument(props: DocumentProps & DocumentHeadTagsProps)
         <link rel="apple-touch-icon" href="/logo512.png" />
         <link rel="stylesheet" href="https://cdn.couchers.org/fonts/ubuntu/ubuntu.css" />
         <link rel="stylesheet" href="https://cdn.couchers.org/fonts/wordmark/wordmark.css" />
-        <script
-          dangerouslySetInnerHTML={{
-            __html: `window.addEventListener('load',function(){var s=document.createElement('script');s.async=true;s.src='https://twdl.couchers.org/script.js';document.head.appendChild(s)})`,
-          }}
-        />
+        {/* Antibot script. Its fingerprint loop leaks WebGL contexts, so we skip it in dev. */}
+        {process.env.NODE_ENV !== "development" && (
```
papeldeorigami, line 19: This part worked for me - at least the WebGL warning is not showing up anymore on hot reload.

papeldeorigami (dismissed):
Approving because the change in _document.tsx works.

before | after
-- | --
Image | Image

For the other parts, I agree with @darrenvong, they might be redundant / no-op, so consider removing them before merging.

nit - It would have been worth telling which messages or errors you were specifically trying to remove in the PR description.

app/web/next.config.js
nabramow: Oooo good catch, I shoulda looked deeper.

nabramow:
> nit - It would have been worth telling which messages or errors you were specifically trying to remove in the PR description.

🫣 Fair, I got used to there not being any other frontend devs lol.

app/web/next.config.js
nabramow: Removed now

app/web/next.config.js
nabramow: Removed!

PR #9451, opened 2026-08-04
Frontend/i18n: Don't use Trans innerHtml

tristanlabelle:
The reason `<Trans>` supports inner text is that some loc systems will use that text as the string key. We have explicit string keys so the inner text gets ignored, which is confusing and results in worse placeholder names (e.g. `<1>`)

## Testing
Spot checked most instances using Vercel preview.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/Footer/Footer.tsx
+++ b/app/web/components/Footer/Footer.tsx
@@ -262,17 +262,11 @@ export default function Footer({ bottomMargin }: { bottomMargin?: string }) {
                 version: version_text,
                 updated_ago: updated_ago_text,
               }}
-            >
-              Version{" "}
-              <VersionLink href={version_link} target="_blank" rel="noopener noreferrer">
-                {version_text}
-              </VersionLink>
-              , last updated{" "}
-              <VersionLink href={updated_ago_link} target="_blank" rel="noopener noreferrer">
-                {updated_ago_text}
-              </VersionLink>
-              .
-            </Trans>
+              components={{
+                2: <VersionLink href={version_link} target="_blank" rel="noopener noreferrer" />,
+                5: <VersionLink href={updated_ago_link} target="_blank" rel="noopener noreferrer" />,
+              }}
+            />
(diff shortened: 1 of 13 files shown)


tristanlabelle:
CI failure is due to `archive.ubuntu.com` server flakiness. Ignore.

> E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/p/python3-defaults/python3_3.8.2-0ubuntu2_amd64.deb  Connection failed [IP: 91.189.92.22 80]

app/web/components/Footer/Footer.tsx
```diff
@@ -262,17 +262,11 @@ export default function Footer({ bottomMargin }: { bottomMargin?: string }) {
...
-              </VersionLink>
-              , last updated{" "}
-              <VersionLink href={updated_ago_link} target="_blank" rel="noopener noreferrer">
-                {updated_ago_text}
-              </VersionLink>
-              .
-            </Trans>
+              components={{
+                2: <VersionLink href={version_link} target="_blank" rel="noopener noreferrer" />,
+                5: <VersionLink href={updated_ago_link} target="_blank" rel="noopener noreferrer" />,
```
nabramow, line 267: Just FYI, not sure if you did these, but on frontend and Javascript in general we always use camelCase, not snake_case. Talking about the `version_link` and the `updated_ago_link`.

tristanlabelle: I am innocent: https://github.com/Couchers-org/couchers/blob/develop/app%2Fweb%2Fcomponents%2FFooter%2FFooter.tsx#L157

aapeliv: i would guess i am to blame, let me check

aapeliv: yes it was me

aapeliv:
rebased and will merge

Issue #9446, opened 2026-08-03
"Join us" on login page should be a link


nc225:
# join us on login page should be a link
## Description
https://couchers.org/login

## Results
"Join us" in "No account yet? Join us" should be a hyperlink but it's not

## Diagnostics
**Backend version**: `develop-d1afcd17`
**Frontend version**: `develop-d1afcd17`
**User Agent**: `Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:153.0) Gecko/20100101 Firefox/153.0`
**Locale**: `en`
**Screen resolution**: 1853x1053
**Page**: https://couchers.org/dashboard
**User**: [@bknicholas](https://couchers.org/user/bknicholas) (12838) / `5pv9HY6440nt`

PR #9458, opened 2026-08-06
i18n: detect <Trans> markup that doesn't match its string

aapeliv:
When a translated string contains a tag that matches nothing in a `<Trans>` call site's `components`, react-i18next renders the tag's inner text and silently drops the wrapper — no error, no console warning. #9446 was exactly that: a reformat renumbered `Login.tsx`'s children, and "Join us" stopped being a link. The numbering is derived from the JSX source, so prettier alone is enough to break a string.

This adds two detectors, and documents the conventions that keep it from happening in the first place. It does not migrate anything yet — that's a follow-up.

- **`couchers/trans-components` ESLint rule** (`app/web/eslint-rules/`). For each `<Trans>` it resolves `i18nKey` against the real `en.json` — handling `ns:` prefixes, `ns={GLOBAL}`, the file's `useTranslation()` namespaces, ternary keys and plural suffixes — and reports tags with no matching component, dead `components` entries, missing keys, JSX children, and numbered or array `components`. Runs in the existing `test:web-linting` job and inline in editors.
- **`app/web/i18n/locales.test.ts`**, asserting that every translation carries the same markup tags as its English source. Catches translator-side drift arriving from Weblate.
- **`docs/translation-components.md`**: use named tags, prefer bare `<strong>`/`<i>`/`<p>`/`<br>` which need no `components` entry at all, never pass children to `<Trans>`.

Two baselines keep this green on today's code, each with a "nothing may be added here" header and both deleted by the migration PR:

- `i18n/legacyTransFiles.js` — 46 call sites still on the old style. Being listed waives only the style rules; a call site that disagrees with its string is an error wherever it lives.
- `i18n/knownTagMismatches.js` — 27 translations that already render wrong. A second test fails if an entry stops being a real mismatch, so the list can't rot.

Also removes two dead `components={{ bold: … }}` entries (`CouchersIntroduction.tsx`, `DidStay.tsx`) whose strings no longer contain a `<bold>` tag. Those were the rule's only findings outside the baseline.

### Follow-ups, not in this PR

- Migrate the 46 call sites to named tags, rewriting the tsx and all 17 locale files together, and delete both baseline files. Best landed while no Weblate PR is open.
- The 27 mismatches are Weblate's to fix, not the repo's. Some are worth a look: `zh-Hans`/`zh-Hant` translated the tag name itself in `what_is_cs.is_it_safe.description_2` (`<指南>` for `<guidelines>`), `pt-BR` and `tr` have unclosed tags, and `donations_box.helper_text` lost its cancellation link in five languages. Worth checking Weblate's `xml-tags` check is enabled on the component.

## Testing

- `yarn lint --max-warnings=0` and `yarn prettier --check` pass; `tsc --noEmit` reports only pre-existing `slotProps` errors in untouched files.
- `yarn test i18n/locales.test.ts eslint-rules/trans-components.test.ts` — 38 tests. The rule has 9 RuleTester cases covering each report and the baseline waiver; the locale test has unit tests for the tag comparison (dropped tag, unclosed tag, translated tag name, reordering).
- Verified the rule catches the #9446 shape by temporarily stripping the `components` prop from `Login.tsx`: "contains `<2>` but `components` has no `2` entry".
- `yarn test features/profile/view/leaveReference` passes after the `DidStay.tsx` change.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/eslint-rules/trans-components.js
+++ b/app/web/eslint-rules/trans-components.js
@@ -1,0 +1,267 @@
+/**
+ * Checks every `<Trans>` call site against the English string it renders.
+ *
+ * A tag in the string that matches nothing in `components` doesn't raise an
+ * error at runtime, it just renders the tag's inner text with the wrapper
+ * silently dropped, so this has to be caught statically.
+ *
+ * See /docs/translation-components.md.
+ */
+
+"use strict";
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const fs = require("fs");
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const path = require("path");
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const namespaces = require("../i18n/namespaces");
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { extractTags, isBasicHtmlTag, localeFilePath, resolveKey } = require("../i18n/transTags");
+
+const ROOT = path.resolve(__dirname, "..");
+const DEFAULT_NAMESPACE = "global";
+
+// `AUTH` -> `"auth"`, so that `ns={AUTH}` and `useTranslation(AUTH)` resolve.
+const NAMESPACE_BY_IDENTIFIER = Object.fromEntries(
+  Object.entries(namespaces).filter(([, value]) => typeof value === "string"),
+);
+
+const bundleCache = new Map();
+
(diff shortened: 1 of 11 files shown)


tristanlabelle (commented):
I think the drifting issue is not going to be a problem once we have all components named instead of indexed. The string writers/editors should be testing their code and catch that the component part is broken.

PR #9461, opened 2026-08-06
i18n: name every <Trans> tag instead of numbering it

aapeliv:
Part of stack #9465, on top of #9458. Base is `web/feature/trans-tag-detectors`; GitHub retargets this to `develop` when #9458 merges.

Completes what #9458 only detected. Every numbered tag is gone: `<2>` becomes `<signupLink>`, and the strings whose only markup was emphasis now carry a bare `<strong>` that needs no `components` entry at all.

- **64 strings renamed**, across `en` and all 17 other locales in the same commit — 124 locale files, 795 lines, one-for-one with no reformatting (verified the JSON round-trip is byte-identical first, so the diff is only the intended lines).
- **46 call sites rewritten**: JSX children and array `components` become named object entries; ~15 `components` props disappear entirely because `<strong>` renders itself.
- Names describe the target, not the position — `signupLink`, `tosLink`, `guideLink`, `portalLink`, `photographerLink`. Plural variants (`signup_description2_one`/`_other`) renamed together, including plural forms that only exist in other locales.
- **`i18n/legacyTransFiles.js` deleted** and the rule's allowlist option dropped, so `couchers/trans-components` now holds every call site in the repo to named tags with no waivers.
- Adds a test that the English source never numbers a tag, covering the three strings (`dashboard:landing_text`, `find_your_community_intro`, `community_missing`) that no call site renders any more and which the ESLint rule therefore can't see.

A few call sites had markup in children with no tag in the string at all (`ModNoteCard`, `LoginCard`'s login header) — those became plain `values={{ … }}`. Two used `<b>`, which isn't in react-i18next's `transKeepBasicHtmlNodesFor`; they're now `<strong>`, which renders identically.

`i18n/knownTagMismatches.js` is unchanged — the same 27 translations still disagree with their source, since the codemod renames a tag only where the translation actually had it.

## Testing

- `yarn lint --max-warnings=0` passes with the rule fully enabled and no allowlist, which is the real check here: it means all 46 rewritten call sites agree with their strings.
- `yarn prettier --check` passes; `tsc --noEmit` reports only the pre-existing `slotProps` errors in `components/Autocomplete.tsx`.
- `yarn test` — the touched areas (`i18n`, `eslint-rules`, `features/auth`, `features/donations`, `features/profile/view/leaveReference`, `components/Footer`) are 24 suites / 249 tests, all passing.
- Full suite has 9 failing suites / 41 tests. I checked these out on the base branch and they fail identically there — all `components/Autocomplete.tsx` `slotProps` breakage already on `develop`, unrelated to this PR.
- Verified zero numbered tags remain in any English source string.

## Conflicts to be aware of

- **#9451 "Don't use Trans innerHtml"** converts 13 of these same call sites from children to numbered `components`. This PR converts them to named tags directly, so the two overlap. Whichever lands second needs a rebase.
- **#9454 "Make shared translation package"** moves the `global` locale files to `app/client-shared/locales/global/`, which conflicts with the `resources/locales/*.json` changes here and will also need `localeFilePath()` in `i18n/transTags.js` updated.
- **#9455** is an open Weblate PR, so this touches locale files in the same window.

**Web frontend checklist**

## For maintainers


tristanlabelle (dismissed):
Let's merge this but sync on avoiding weblate conflicts.

PR #9460, opened 2026-08-06
Backend: extract shared message helpers, bulk mark_notifications_seen

aapeliv:
First of a stack breaking up #9219 into reviewable pieces. Two refactors with no schema or proto changes:

- **Extract shared message serialization helpers**: the conversations and requests servicers each had their own copy of `message_to_pb`, and each was missing message types the other rendered (conversations' copy didn't handle `host_request_status_changed`, requests' copy didn't handle the group chat control messages). One complete version now lives in `couchers.helpers.messages`, along with the host request status maps.
- **Bulk `mark_notifications_seen`**: takes a list of (topic actions, keys) groups and marks them seen in a single update, instead of one call per group. Needed later in the stack by `MarkAllThreadsSeen`.

## Testing
`make mypy` passes; `test_conversations.py`, `test_requests.py` and `test_notifications.py` pass locally.

**Backend checklist**
- [x] Added tests for any new code or added a regression test if fixing a bug (covered by existing tests)
- [x] Run the backend locally and it works
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no db changes)

## For maintainers
- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

_This PR was created with the Couchers PR skill._

PR #9462, opened 2026-08-06
Scope group chat unseen counts to the current subscription window

aapeliv:
Second in the stack breaking up #9219 (on top of #9460). Fixes two unread-count bugs around group chats the user was removed from:

- **ListGroupChats and GetGroupChat counted messages sent after the viewer left the chat**, which they can never open — while the Ping badge correctly excluded them, so the per-chat counts and the badge disagreed.
- **Rejoining a chat leaves the earlier subscription behind, and Ping matched on any of the viewer's subscriptions**, so the stale subscription's unread messages kept the badge up even after the viewer read everything on their current subscription (marking seen only ever advances the current one).

The window ("messages sent while the user was in the chat") and current-subscription rules now live in `couchers.helpers.group_chats`, shared by both the conversations API and the Ping badge so all the counts agree. Later in the stack, `ListMessageThreads` / `MarkAllThreadsSeen` build on the same helpers.

## Testing
Added two regression tests (`test_departed_group_chat_unseen_count_excludes_unreachable_messages`, `test_rejoined_group_chat_ping_reads_current_subscription`); both fail on the unfixed code and pass with the fix. `make mypy` passes; `test_conversations.py` and `test_api.py` pass locally.

**Backend checklist**
- [x] Added migrations if there are any database changes, rebased onto `develop` if necessary for linear migration history (no db changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/helpers/group_chats.py
+++ b/app/backend/src/couchers/helpers/group_chats.py
@@ -1,0 +1,49 @@
+"""
+Rules for scoping a user's group chat messages, shared by the conversations API and the Ping badge.
+"""
+
+from datetime import datetime
+
+from sqlalchemy import ColumnElement, SQLColumnExpression, and_, func, or_, select
+
+from couchers.models import GroupChatSubscription, Message
+
+
+def was_subscribed_at(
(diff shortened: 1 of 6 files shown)


aapeliv (commented):
lgtm

PR #9466, opened 2026-08-06
Backend: Add postal verified badge and solution for postal verification on staging

nabramow:
* Adds postal verified badge
* Solution for postal verification on non-prod. It will go through the whole flow but email the card instead, so we don't incur a cost, but can still test the flow and see the card.

## Testing

**Backend checklist**

## For maintainers


app/backend/src/couchers/config.py
```diff
@@ -68,6 +68,9 @@ class Config:
     MYPOSTCARD_PASSWORD: str
     MYPOSTCARD_PRODUCT_CODE: str
     MYPOSTCARD_CAMPAIGN_ID: str
+    # Whether to place real, billed MyPostcard orders. Off by default so that a deployment can only charge us
+    # if it deliberately opts in; non-prod runs the whole flow but emails the rendered postcard instead.
+    MYPOSTCARD_LIVE: bool = False
```
tristanlabelle, line 73: I could see a case where we might not want to turn MyPostcard on for prod, but we still wouldn't want email-based postal verification. It seems safer and more explicit to have a config for specifically enabling the email-based mocking.

app/backend/src/couchers/i18n/locales/en.json
```diff
@@ -266,6 +267,8 @@
     "donor_description": "This user has contributed funds to the project",
     "phone_verified_name": "Verified Phone",
     "phone_verified_description": "This user has verified their phone number",
+    "postal_verified_name": "Verified Address",
+    "postal_verified_description": "This user has verified their address with a code sent to them by post",
```
tristanlabelle, line 271: I wonder if we can make the description it less implementation detail-y. Some (weak) ideas:

"This user has verified their address by physical mail"

Or even:

"This user's postal address has been verified"

app/backend/src/couchers/postal/simulated.py
```diff
@@ -0,0 +1,61 @@
...
+
+from sqlalchemy.orm.session import Session
+
+from couchers.email.queuing import queue_system_email
+from couchers.postal.my_postcard import _generate_back_left_side_png
+from couchers.proto.internal import jobs_pb2
+
+logger = logging.getLogger(__name__)
+
+ATTACHMENT_FILENAME = "example-postcard.png"
```
tristanlabelle, line 19: Minor: It's not really an example. Suggest: "postcard.png"

app/backend/templates/system/simulated_postal_verification_postcard.md
```diff
@@ -0,0 +1,31 @@
...
+**************************************************************
+*                                                            *
+*   THIS IS AN EXAMPLE. NO POSTCARD WAS PRINTED OR MAILED.   *
+*                                                            *
+*   You are using a Couchers.org test/staging server, not    *
+*   the real site. Nothing has been posted to your address   *
+*   and nothing has been charged. The attached image is what *
+*   would have been printed on the real postcard.            *
+*                                                            *
+**************************************************************
```
tristanlabelle, line 14: I don't think we need this many all caps, warnings, and "again" reminders. Just "Mock postcard: your verification code is ..." and the prose you have is clear enough. What are we guarding against to need such strong warnings?

- If this gets sent by error by the real website, more warnings won't help us undo the problem.
- If it gets sent to next users, then there's no bad outcome and standard prose already calls out that this is a special path.

Also "Example" is confusing as a term because it has the real code. "Mock postcard" or "Simulated postcard" are clearer.

tristanlabelle (commented):
I'm on board with the change in general, but I think we should have the opposite config: e.g. `MOCK_MYPOSTCARD_WITH_EMAIL` set on non-prod (so it's obviously wrong if it's set on prod) and we don't need to overemphasize that the email is a mock postcard.

app/backend/src/couchers/jobs/handlers.py
```diff
@@ -827,6 +828,8 @@ def float_(stmt: Any) -> Function[float]:
             "volunteer": 3,
             "past_volunteer": 2,
             "donor": 1,
+            # Harder to fake than a phone number, easier than a biometric passport
```
aapeliv, line 831: what's this comment about?

tristanlabelle: Probably justifying the magic number chosen?

app/backend/src/couchers/i18n/locales/en.json
```diff
@@ -161,6 +161,7 @@
     "node_moderate_permission_denied": "You're not allowed to moderate that community",
     "not_a_volunteer": "You are currently not registered as a volunteer, if this is wrong, please contact us.",
     "not_donated": "Please complete donation to get phone verified.",
+    "not_donated_postal": "Please complete a donation to verify your address by post.",
```
tristanlabelle, line 164: It could read like the process of completing a donation verifies your address. Same problem with the phone verified string above. Some alternatives:

> A donation is required before using postal verification.

> A donation is required to enable postal verification.

> Postal verification is available only after donating to Couchers.org.

> Please donate to Couchers.org to enable postal verification.

app/backend/src/tests/test_postal_verification.py
```diff
@@ -682,6 +685,126 @@ def test_has_postal_verification_helper(db):
...
+
+def test_postcard_is_posted_when_bypass_is_unset(db):
+    """With the bypass unset, an order is placed and its job ID recorded."""
+    user, token = generate_user()
+    attempt_id = _confirmed_attempt_id(token)
+
+    with (
+        patch.object(config, "POSTAL_VERIFICATION_BYPASS_POST_AND_EMAIL_CODE_FOR_TESTING", False),
+        patch("couchers.jobs.handlers.send_postcard") as mock_send,
+    ):
```
tristanlabelle, line 784: Given the `testconfig` fixture, you can simply assign to `config.POSTAL_VERIFICATION_BYPASS_POST_AND_EMAIL_CODE_FOR_TESTING`. Same below

tristanlabelle (dismissed):
Looks good. Just one question about the design: I thought postal verification validated a user's address in their profile, but it seems like it validates an arbitrary address. Convo in `#postal-verification`.

app/backend/src/tests/test_postal_verification.py
aapeliv: thanks

aapeliv (approved):

PR #9469, opened 2026-08-07
Web/bugfix/deleteaccountplaceholder

### LINKED ISSUE #9452: Frontend/i18n: Delete account username placeholder overflows input control
Image

nc225:
Closes #9452. This PR addresses a problem with placeholder text overrunning the length of a textbox. I added a hidden element to the page to help us measure the length of the translated "Confirm Username" placeholder label and then adjust the textbox's width accordingly.

## Testing
Tested different screen sizes and languages. This issue could also affect other text boxes with long placeholder text labels, so I poked around some other languages and parts of the site but didn't find any other places with issues.

|  | Before | After |
| -- | ----- | ---- |
| English | Screenshot From 2026-08-07 11-16-43 | Screenshot From 2026-08-07 11-16-17 |
| Français | Screenshot From 2026-08-07 11-16-31 | Screenshot From 2026-08-07 11-16-54

Small screen behavior (width = 100%) is unchanged.

**Backend checklist**

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/auth/deletion/DeleteAccount.tsx
+++ b/app/web/features/auth/deletion/DeleteAccount.tsx
@@ -69,7 +69,7 @@ export default function DeleteAccount({ className, username }: DeleteAccountProp
             <Trans t={t} i18nKey="auth:delete_account.request.confirm_username_explanation" values={{ username }} />
           </Typography>
           <TextField
-            sx={{ width: 300, display: 'inline-flex'}}
+            sx={{ width: 360, display: 'inline-flex'}}
             id="confirmUsername"
             {...register("confirmUsername", { required: true })}
             label={t("auth:delete_account.request.confirm_username_label")}
--- a/app/web/features/auth/deletion/DeleteAccount.tsx
+++ b/app/web/features/auth/deletion/DeleteAccount.tsx
@@ -69,6 +69,7 @@ export default function DeleteAccount({ className, username }: DeleteAccountProp
             <Trans t={t} i18nKey="auth:delete_account.request.confirm_username_explanation" values={{ username }} />
           </Typography>
           <TextField
+            sx={{ width: 300, display: 'inline-flex'}}
             id="confirmUsername"
             {...register("confirmUsername", { required: true })}
             label={t("auth:delete_account.request.confirm_username_label")}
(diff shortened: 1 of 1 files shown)


app/web/features/auth/deletion/DeleteAccount.tsx
```diff
@@ -73,11 +92,15 @@ export default function DeleteAccount({ className, username }: DeleteAccountProp
             </Trans>
           </Typography>
           <TextField
+            sx={{
+              display: "inline-flex",
+              width: isMdOrWider ? usernameWidth : "100%",
```
papeldeorigami: @nc225 This is the core part of the fix, if I understand it correctly. It works because we define a width for large screens taking the translated text length into account.

However, the approach with hidden text and useLayoutEffect to get the accurate width is the part that I would like to avoid, because it would be hard to maintain in the future.

Here is one idea: we could just use the text length in `ch` units, picking up the max of that and the default width - like the 200px that you defined.

My suggestion is roughly this:
```
width: isMdOrWider ? `max(${DEFAULT_TEXFIELD_WIDTH}, ${labelUsername.length}ch)` : "100%",
```

BTW I noticed that other fields above in this screen are inside a `15.5rem` container (like the change e-mail form), so that would be the default width that I would choose instead of 200px. We could even put that into a constants file for reuse later.

The result doesn't look as accurate as the one you got (because `ch` measure can never be as accurate as measuring the actual offsetWidth), but it does fix the issue of overflowing text without the need for useLayoutEffect, useRef, etc.

From a dev perspective, I would prefer that simpler approach, despite the slightly inaccurate width.

Here is a screenshot of the result.

Image

papeldeorigami (changes requested):
PTAL at the comment below - I would prefer if we could take a simpler approach, maybe something like what I am suggesting or maybe something else. Let me know what you think.

app/web/features/auth/deletion/DeleteAccount.tsx
nc225: Yes `ch` is better here, thanks. Before I went with useState etc, I had experimented with multiplying text length by some factor but couldn't get it right. Also yes we could move the 15.5rem to a constants file but since I see it in a few other files too I'm going to leave it as a constant for the purposes of this PR

papeldeorigami (approved):

nc225:
I merged the latest `develop` into this branch today because some errors were going off. It looks like the errors have resolved now. I didn't change anything about the way this code works, but please let me know if I should have gone about today's merge in a different way,

nabramow:
> I merged the latest `develop` into this branch today because some errors were going off. It looks like the errors have resolved now. I didn't change anything about the way this code works, but please let me know if I should have gone about today's merge in a different way,

It seems okay I looked into it and it looks like there's some unrelated flaky tests on the backend.

nabramow (approved):

PR #9476, opened 2026-08-08
Add generated surfer/host columns to host requests

aapeliv:
Host requests carry two independent role axes:

* **initiator/recipient** — the conversation role: who proposed the stay, and who responds to it. The status state machine, unread tracking and archiving all key off this axis.
* **surfer/host** — the stay role: whose couch it is. References, feedback, response rates and metrics all key off this axis.

They coincide for a normal host request, but an offer on a public trip reverses them — the host initiates and the traveller responds. Since migration `0140` renamed the columns to `initiator`/`recipient`, the stay role has had nowhere to live, so every consumer that cares about it has to re-derive it from `public_trip_id IS NULL`.

This adds `surfer_user_id` and `host_user_id` as **stored generated columns**, so the stay role is always present and can't drift out of sync with the conversation role. Both get a FK to `users` and an index, and there are matching `viewonly` `surfer` / `host` relationships.

Nothing reads the new columns yet — this is inert groundwork, safe to land and deploy on its own. Switching the call sites over is a follow-up, where each one becomes a visible one-line change with the schema already in place.

Follows the `users.about_me_length` precedent from `0176`.

## Testing

* `test_db.py::test_migrations` passes — the migrated schema and the model-built schema are byte-identical under `pg_dump`, confirming the migration and model agree on columns, FK names and index names.
* `test_requests.py`, `test_public_trips.py`, `test_references.py` — 105 tests pass, confirming the new columns don't disturb the existing insert paths.
* `make mypy` and `make format` clean.
* Verified directly against PG 18.3 that a stored generated column accepts both a FK and an index, and that SQLAlchemy populates the attribute via `RETURNING` immediately after `flush()`:
  ```
  offer  -> surfer_user_id = 2  (reversed, correct)
  normal -> surfer_user_id = 1  (correct)
  ```

Note that `ALTER TABLE ... ADD COLUMN ... GENERATED ALWAYS AS ... STORED` rewrites `host_requests` under an `ACCESS EXCLUSIVE` lock. Both columns are added in a single statement so the table is only rewritten once.

**Backend checklist**
- [ ] Added tests for any new code or added a regression test if fixing a bug — no new tests; the pre-existing `test_migrations` schema comparison covers the new columns automatically

## For maintainers

### DIFF
--- a/app/backend/src/couchers/migrations/versions/0180_add_generated_host_request_role_columns.py
+++ b/app/backend/src/couchers/migrations/versions/0180_add_generated_host_request_role_columns.py
@@ -1,0 +1,43 @@
+"""Add generated surfer/host columns to host_requests
+
+Revision ID: 0180
+Revises: 0179
+Create Date: 2026-08-08 10:00:00.000000
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0180"
+down_revision = "0179"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # both columns are added in one statement so the table is only rewritten once (under an ACCESS EXCLUSIVE lock)
+    op.execute("""
+        ALTER TABLE host_requests
+        ADD COLUMN surfer_user_id bigint NOT NULL
+            GENERATED ALWAYS AS (case when public_trip_id is null then initiator_user_id else recipient_user_id end) STORED,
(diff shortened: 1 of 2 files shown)


nabramow (approved):
Probably some frontend stuff to change after this I assume? But putting these back is one solution whoopsie probably should have done that to start.

PR #9481, opened 2026-08-08
Remove the account deletion report email

aapeliv:
When a user initiates account deletion and writes a reason, we sent a copy of that reason to `REPORTS_EMAIL_RECIPIENT`. This removes that email — deletion reasons are still stored in the `account_deletion_reason` table, they're just no longer emailed out one at a time.

Changes:
- Dropped the `send_account_deletion_report_email` call from `Account.DeleteAccount`, along with the `session.flush()` that only existed to populate `reason.id` for the email template.
- Removed `send_account_deletion_report_email` from `couchers/tasks.py`.
- Deleted the `templates/system/account_deletion_report.md` template.

No behaviour changes to deletion itself: the reason row is still written, the deletion token and confirmation email/push to the user are unchanged, and `account.deletion_initiated` still logs `has_reason`.

## Testing
No new tests — this removes a side effect that nothing asserted, so there's nothing to add a test for. Existing coverage confirms deletion still behaves the same without it:

```bash
cd app/backend
uv run pytest src/tests/test_account.py src/tests/test_auth.py src/tests/test_email.py src/tests/test_dump_emails.py src/tests/test_email_localization.py
make mypy
```

All pass (40 + 125 tests), mypy clean across 472 files.

To check manually: request account deletion from Settings with a reason filled in, and confirm no mail is queued to `REPORTS_EMAIL_RECIPIENT` while a row still appears in `account_deletion_reason`.

**Backend checklist**

## For maintainers


jesseallhands (approved):
Lgtm

PR #9032, opened 2026-06-07
Improve ongoing event and stay labeling

nabramow:
- For ongoing stays and events it will say "now" instead of "today" or "yesterday"

## Testing
*Explain how you tested this PR and give clear steps so the reviewer can replicate.*

**Web frontend checklist**
- [x] There are no console warnings when running the app
- [ ] Added tests where relevant
- [x] Clicked around my changes running locally and it works
- [x] Checked Desktop, Mobile and Tablet screen sizes

## For maintainers

- [x] Maintainers can push commits to my branch
- [x] Maintainers can merge this PR for me

PR #9488, opened 2026-08-09
Fix css styles

nabramow:
I am very delayed but @darrenvong made a comment here awhile ago about incorrect CSS styles: https://github.com/Couchers-org/couchers/pull/9032#pullrequestreview-4468135186

I am finally fixing that and some similar issues I found.

## Testing
* `/dashboard` → "Your events" + community events list — check the date chips: today's/ongoing events show a shrunken "TODAY"/"NOW" label, others show month + day. Inspect a DateMonth div — no stray attribute.
* `/translate` → language cards. Progress bars fill to the right %, low-% cards are faded. Inspect a bar — should be <div class="...">, no percent="50".
* Language picker (footer/settings language selector) → same progress bars in the dropdown.
* `/messages` → Hosting and Surfing tabs. Chips should stay teal for hosting, orange for surfing. Inspect a chip — no ishost="true".

**Web frontend checklist**

## For maintainers


app/web/features/dashboard/EventListRow.tsx
```diff
@@ -187,7 +187,7 @@ export default function EventListRow({ event }: EventListRowProps) {
   return (
     <RowLink href={routeToEvent(event.eventId, event.slug)}>
       <DateChip data-today={isToday || isOngoing || undefined}>
-        <DateMonth $labelFontSize={isToday || isOngoing ? chipFontSize : undefined}>
+        <DateMonth labelFontSize={isToday || isOngoing ? chipFontSize : undefined}>
```
darrenvong: I don't think it's related to this change, but not sure why the "today" label is not showing up correctly when I literally just created events on the same day... 🤔 
Image

darrenvong (dismissed):

app/web/features/dashboard/EventListRow.tsx
nabramow: Ooo good catch it looks like this got messed up when we did the Temporal refactor.

I've fixed it and I also made all the widgets consistent:

- Use "TODAY" if there is a date and it is today
- Use "NOW" if there is a time and it's currently happening

app/web/features/dashboard/UpcomingStayCard.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import UpcomingStayCard from "./UpcomingStayCard";
+
+const { t } = i18n;
+
+const getLiteUserMock = service.user.getLiteUser as jest.Mock;
+
+function stayFrom(fromOffsetDays: number, toOffsetDays: number): HostRequest.AsObject {
+  const today = Temporal.Now.plainDateISO();
+  return {
+    ...(hostRequest as unknown as HostRequest.AsObject),
```
darrenvong: Curious why we need this assertion as casting to unknown first then to whatever is effectively disabling the type checking, which is a bit sus 🤔

Shouldn't the fixture contain all of the fields that are in the `HostRequest` object already and it should pass the type check from having the same shape/fields in the object?

app/web/features/dashboard/EventListRow.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+    expect(screen.getByText(t("dashboard:now_label"))).toBeVisible();
+    expect(screen.queryByText(t("dashboard:today_label"))).not.toBeInTheDocument();
+  });
+
+  it("shows the month and day for an event on another day", () => {
+    const now = Temporal.Now.zonedDateTimeISO();
+    const start = now.add({ days: 3 });
+    render(<EventListRow event={eventAt(start, start.add({ hours: 1 }))} />, { wrapper });
+
+    expect(screen.getByText(String(start.day))).toBeVisible();
```
darrenvong: I'm wondering if it might be worth asserting the actual date string more explicitly to make this clearer to read - it shouldn't be an issue since the current date/system time is mocked?

app/web/features/dashboard/DashboardPublicTripCard.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+import i18n from "test/i18n";
+
+import { DashboardPublicTripCard } from "./DashboardPublicTripCard";
+
+const { t } = i18n;
+
+function tripFrom(fromOffsetDays: number, toOffsetDays: number): PublicTrip {
+  const today = Temporal.Now.plainDateISO();
+  return {
+    ...(publicTrips[0] as unknown as PublicTrip),
```
darrenvong: Same comment as in the UpcomingStayCard on the assertion

app/web/features/dashboard/DashboardPublicTripCard.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+      render(<DashboardPublicTripCard trip={tripFrom(0, 5)} locale="en" {...modeProps} />, { wrapper });
+
+      expect(screen.getByText(t("dashboard:today_label"))).toBeVisible();
+      expect(screen.queryByText(t("dashboard:now_label"))).not.toBeInTheDocument();
+    });
+
+    it("shows the tomorrow label for a trip starting tomorrow", () => {
+      render(<DashboardPublicTripCard trip={tripFrom(1, 5)} locale="en" {...modeProps} />, { wrapper });
+
+      expect(screen.getByText("Tomorrow")).toBeVisible();
```
darrenvong, line 61: Is this not from `t` because it comes from the Temporal API?

darrenvong (commented):
Not properly re-tested but some more comments at the code level from a few things I spotted

app/web/features/dashboard/UpcomingStayCard.test.tsx
nabramow: Ooo you are right that was a hack and the object did contain the proper fields. Fixing this!

nabramow: I think it got confused by this unknown value in`types/json.d.ts`. I think i can remove it so incorrect things fail loudly which is probably what we want?

```typescript
declare module "*.json" {
  const value: unknown;
  export default value;
}
```

darrenvong: Yeah I don't think JSON imports need overriding like this as TS seems to infer the types correctly by default

nabramow: Looks like I silently added it in PR [#6419](https://github.com/Couchers-org/couchers/pull/6419) for translation files. I'm gonna try to remove it and see what happens ahaha.

app/web/features/dashboard/DashboardPublicTripCard.test.tsx
nabramow: Correct. It's not a translation key. It comes from `Intl.RelativeTimeFormat` with `numeric: "auto"` via `localizeRelativeTimeUnit` function, which is what turns +1 day into "tomorrow".

app/web/test/fixtures/publicTrips.ts
```diff
@@ -0,0 +1,100 @@
...
+    user: funnyDog,
+    communityId: community.communityId,
+    fromDate: "2026-04-20",
+    toDate: "2026-04-27",
+    description:
+      "Visiting NYC for the first time! I'm a photographer from Helsinki and would love to stay with a local who can show me around. I'm easygoing, tidy, and happy to cook Finnish food for my host.",
+    status: PublicTripStatus.PUBLIC_TRIP_STATUS_SEARCHING_FOR_HOST,
+    created: { seconds: 1775822400, nanos: 0 },
+    communitySlug: community.slug,
+    sameGenderOnly: true,
```
nabramow, line 20: I changed this file to be TS rather than JSON as I think that should prevent the fixtures used in tests from drifting from the real objects. It's the only one like this so far but since publicTrips is a new feature it's a good place to start!

app/web/types/json.d.ts
```diff
@@ -1,4 +0,0 @@
-declare module "*.json" {
-  const value: unknown;
-  export default value;
-}
```
nabramow, line 4: This was added in PR #6419 when doing some translation file stuff. As far as I can see it doesn't break anything removing it, but having it like this was making the types less strict which means some weird stuff could get through.

app/web/features/dashboard/EventListRow.tsx
```diff
@@ -165,16 +165,17 @@ export default function EventListRow({ event }: EventListRowProps) {
...
+  const isNow =
     endDate !== null &&
     Temporal.ZonedDateTime.compare(startDate, now) <= 0 &&
     Temporal.ZonedDateTime.compare(endDate, now) >= 0;
 
   // Define "today" as: starting on the current day in the current timezone.
-  const isToday = !isOngoing && startDate.withTimeZone(now.timeZoneId).toPlainDate() === now.toPlainDate();
-  const todayLabel = t("dashboard:events.today_label");
+  const isToday =
+    !isNow && Temporal.PlainDate.compare(startDate.withTimeZone(now.timeZoneId).toPlainDate(), now.toPlainDate()) === 0;
```
darrenvong: Minor nit: it looks like `Temporal.PlainDate` has an `.equals` method so it'd read slightly clearer than comparing the difference and checking it's 0 to mean it's equal?

```suggestion
  const isToday =
    !isNow && startDate.withTimeZone(now.timeZoneId).toPlainDate().equals(now.toPlainDate());
```

darrenvong (dismissed):

PR #9494, opened 2026-08-09
Move the tests off hand-rolled clock mocking onto timewarp

aapeliv:
Follows up on #9354, which added the `timewarp` / `frozen_timewarp` fixtures: this moves the existing tests onto them.

Before this, tests controlled time by patching `now` on whichever module they cared about (`couchers.jobs.handlers.now`, `couchers.servicers.auth.now`, `couchers.utils.now`). That only moves the clock for that one module — postgres keeps reporting the real time, and any other module involved in the same call disagrees about what time it is. All 22 such patches now go through the fixtures, which shift both clocks together.

Also:

- The missed-message tests advance by the handlers' own `MISSED_MESSAGES_DELAY` / `MISSED_MESSAGES_DELAY_WITH_PUSH` rather than the magic 5-minute and 25-hour offsets that shadowed them.
- Six request-notification tests patched the clock *inside* an open `session_scope`. Timewarp refuses to move mid-transaction (postgres reads the clock as it was at transaction start, so the move would silently do nothing), so the clock move and the job run were lifted out of the session block. The assertions that followed collapsed into the existing `_count_queued_emails()` helper, which is most of the line delta.
- Deleted the now-unused `now_5_min_in_future` / `now_1_day_in_future` / `now_30_min_in_future` helpers.
- Noted the fixtures in `CLAUDE.md` so this doesn't get reinvented.

Deliberately left alone: `backdate_conversations()` and the various `values(last_active=now() - age)` / `next_attempt_after = now() - ...` updates. Those shift *specific rows* rather than the clock — they're historical-data setup, and some (`test_job_dequeue_steps_over_other_workers_jobs`) need distinct per-row values a global clock can't express. Same for the fake `monotonic` in the scheduler test, which drives a loop counter rather than wall-clock time.

No production code changes; this is tests only.

## Testing

Full backend suite passes (1318 passed), as do `make format` and `make mypy`.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/tests/test_account.py
+++ b/app/backend/src/tests/test_account.py
@@ -33,6 +33,7 @@ from tests.fixtures.sessions import (
     real_account_session,
     requests_session,
 )
+from tests.fixtures.timewarp import Timewarp
 from tests.test_requests import valid_request_text
 
 
@@ -468,13 +469,7 @@ def test_ChangeEmailV2_wrong_token(db, fast_passwords):
         assert user_updated.email == user.email
 
 
-def test_ChangeEmailV2_tokens_two_hour_window(db):
-    def two_hours_one_minute_in_future():
-        return now() + timedelta(hours=2, minutes=1)
-
-    def one_minute_ago():
-        return now() - timedelta(minutes=1)
-
+def test_ChangeEmailV2_tokens_two_hour_window(db, timewarp: Timewarp):
     password = random_hex()
     new_email = f"{random_hex()}@couchers.org.invalid"
     user, token = generate_user(hashed_password=hash_password(password))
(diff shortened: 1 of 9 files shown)


app/backend/src/tests/test_account.py
```diff
@@ -490,37 +485,39 @@ def one_minute_ago():
...
-                auth_api.ConfirmChangeEmailV2(
-                    auth_pb2.ConfirmChangeEmailV2Req(
-                        change_email_token=new_email_token,
-                    )
+            )
+        assert e.value.code() == grpc.StatusCode.NOT_FOUND
+        assert e.value.details() == "Invalid token."
+
+    # and a minute after the two hour window closed
+    timewarp.advance(timedelta(hours=2, minutes=2))
```
tristanlabelle: nit: matching old code/comment

```suggestion
    timewarp.advance(timedelta(hours=2, minutes=1))
```

app/backend/src/tests/test_bg_jobs.py
tristanlabelle, line 1: Diff too big for github so I'll just trust you!

tristanlabelle (dismissed):
Nice improvement! I'd be in favor of also changing all the events tests that do `now() + timedelta()` to figure out a start/end time but no rush.

app/backend/src/tests/test_account.py
aapeliv: thanks, fixed!

PR #9500, opened 2026-08-09
Fix emails blowing up when a url ends in "src="

### LINKED ISSUE #9483: FileExistsError: HTML references missing relative image: /app/templates/v2/ style=
Sentry Issue: [BACKEND-NC](https://couchers.sentry.io/issues/7562588007/?referrer=github_integration)

```
FileExistsError: HTML references missing relative image: /app/templates/v2/ style=
(3 additional frame(s) were not displayed)
...
  File "couchers/notifications/background.py", line 48, in _send_email_notification
    payload = get_send_email_payload(
  File "couchers/notifications/render_email.py", line 31, in get_send_email_payload
    rendered_email = render_email(email, email_footer, loc_context)
  File "couchers/email/rendering.py", line 51, in render_email
    body_html, related_parts = embed_html_relative_images(
  File "couchers/email/smtp.py", line 47, in embed_html_relative_images
    html = re.sub(r'src="([^":]+)"', repl=process_relative_src_match, string=html)
  File "couchers/email/smtp.py", line 29, in process_relative_src_match
    raise FileExistsError(f"HTML references missing relative image: {image_path}")

HTML references missing relative image: /app/templates/v2/ style=
```

aapeliv:
Closes #9483.

`embed_html_relative_images` rewrites relative images with `re.sub(r'src="([^":]+)"', ...)` over the whole email body. The pattern isn't anchored to a tag, so it also matches across an attribute boundary: in `blocks.html` every link is `<a href="{{ url }}" style="...">`, and if a url *ends* with the literal `src=` the regex matches `src=" style="`, captures ` style=` as a relative image path, and raises `FileExistsError: HTML references missing relative image: /app/templates/v2/ style=`.

Our urls end in urlsafe base64 of 32 bytes (44 chars, always `=` padded): the signup token in `?token=...`, and the payload/signature in `/quick-link?payload=...&sig=...`. A hit needs the last three data characters to be `s`, `r`, `c`; the final data character only takes 16 values and `c` is one of them, so it's `(1/64)·(1/64)·(1/16)` ≈ **1 in 65k per token-bearing url**. Every signup email has one, every non-critical notification has two to four in the footer.

The consequences are that `SignupFlow` fails with `UNKNOWN` (the user can retry, which mints a new token), and `send_email` jobs fail and re-render with a fresh signature on retry. It also shows up as a rare backend CI flake, e.g. [this `test:backend 1/3` job](https://gitlab.com/couchers/couchers/-/jobs/15797291288), which passed on retry.

The fix is a `(?<=\s)` lookbehind, which keeps the match inside a tag: in an attribute value the `src=` is preceded by a base64 character, never whitespace.

## Testing

Two regression tests are in their own commit, ahead of the fix, and reproduce the exact production error:

```
match = <re.Match object; span=(15197, 15210), match='src=" style="'>
FileExistsError: HTML references missing relative image: .../templates/v2/ style=
```

- `test_smtp.py::test_embed_html_relative_images_ignores_url_ending_in_src` covers the unit directly, with the footer unsubscribe shape (`&sig=<base64>`) next to a genuine relative image.
- `test_email.py::test_signup_verification_email_with_token_ending_in_src` goes end to end through `send_signup_email` with the token patched, so it exercises the real `blocks.html` button markup rather than a hand-written snippet.

The full backend suite passes locally.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/tests/test_smtp.py
+++ b/app/backend/src/tests/test_smtp.py
@@ -57,3 +57,22 @@ def test_embed_html_relative_images() -> None:
     assert related_part.content_type == 'image/png; name="logo-grey.png"'
     assert related_part.content_disposition == 'inline; filename="logo-grey.png"'
     assert related_part.content_id == f"<{content_id_match.group(1)}>"
+
+
+def test_embed_html_relative_images_ignores_url_ending_in_src() -> None:
+    """
+    Our urls end in base64 tokens, whose "=" padding makes them occasionally end in "src=", which must not
+    be picked up as an image reference spanning into the next attribute.
+    """
+    sig = f"{'A' * 40}src="  # a 32 byte blake2b signature, base64 encoded
+    link = f"https://example.com/quick-link?payload=Zm9v&sig={sig}"
+    html = f"""
+        <a href="{link}" style="color: #777;">unsubscribe</a>
+        <img src="attachment_imgs/logo-grey.png" style="border:0;"/>
+    """
+
+    html, related_parts = embed_html_relative_images(html, base_dir=template_folder, content_id_domain="example.com")
+
+    assert link in html, "Link url was replaced"
+    assert len(related_parts) == 1
+    assert related_parts[0].content_disposition == 'inline; filename="logo-grey.png"'
--- a/app/backend/src/tests/test_smtp.py
+++ b/app/backend/src/tests/test_smtp.py
@@ -1,5 +1,7 @@
 import re
 
+import pytest
+
 from couchers.email.rendering import template_folder
 from couchers.email.smtp import email_proto_to_message, embed_html_relative_images
 from couchers.proto.internal.jobs_pb2 import EmailPart, SendEmailPayload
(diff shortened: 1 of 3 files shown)


app/backend/src/tests/test_smtp.py
```diff
@@ -57,3 +57,22 @@ def test_embed_html_relative_images() -> None:
...
     assert related_part.content_id == f"<{content_id_match.group(1)}>"
+
+
+def test_embed_html_relative_images_ignores_url_ending_in_src() -> None:
+    """
+    Our urls end in base64 tokens, whose "=" padding makes them occasionally end in "src=", which must not
+    be picked up as an image reference spanning into the next attribute.
+    """
+    sig = f"{'A' * 40}src="  # a 32 byte blake2b signature, base64 encoded
+    link = f"https://example.com/quick-link?payload=Zm9v&sig={sig}"
```
tristanlabelle: Would it be just as good to test for `?src=foo` while being clearer as a test than specifically replicating the slightly convoluted base64 bug?

tristanlabelle (commented):
Good catch. Fix looks good. Personally I prefer adding unit level tests since it's easier to see the condition they're validating, but I can see the argument for having both.

tristanlabelle (dismissed):
I meant to approve

app/backend/src/tests/test_smtp.py
```diff
@@ -57,3 +59,32 @@ def test_embed_html_relative_images() -> None:
     assert related_part.content_type == 'image/png; name="logo-grey.png"'
     assert related_part.content_disposition == 'inline; filename="logo-grey.png"'
     assert related_part.content_id == f"<{content_id_match.group(1)}>"
+
+
+@pytest.mark.parametrize(
```
tristanlabelle, line 64: Nice

tristanlabelle (approved):

PR #9504, opened 2026-08-10
feat: dev feature-flag override file, and feature flag docs

aapeliv:
badge on:
Screenshot 2026-07-31 at 08 06 41
Screenshot 2026-07-31 at 08 06 51

badge off:
Screenshot 2026-07-31 at 08 08 37
Screenshot 2026-07-31 at 08 08 45

It updated without restarting `yarn start`


aapeliv:
@nabramow this is ready now for review. I fixed the few small comments from before. I suspect the reason you could not run `yarn` was that the branch was super far behind. I ran again locally and works okay.

app/web/.env.development
```diff
@@ -8,5 +8,5 @@ NEXT_PUBLIC_STRIPE_KEY="pk_test_51KEzByIfR5z29g5khFE5samD8XKOGLcCrM1lhCkfOomGPUF
 NEXT_PUBLIC_GLOBAL_MESSAGE_URL="https://gm.couchershq.org/next.json"
 NEXT_PUBLIC_GROWTHBOOK_API_HOST="https://gbapi.couchershq.org"
 NEXT_PUBLIC_GROWTHBOOK_CLIENT_KEY="sdk-f8lwseEODN02p"
-# When enabled, all feature gates return true (useful for development/testing)
-NEXT_PUBLIC_EXPERIMENTATION_PASS_ALL_GATES="1"
+# When enabled, flags resolve from app/web/feature-flags.dev.json instead of GrowthBook (dev/testing)
+NEXT_PUBLIC_FEATURE_FLAGS_OVERRIDE="1"
```
nabramow, line 12: The mobile `.env` is a number and this one is a string. I think it's better if it's a number?

docs/feature-flags.md
```diff
@@ -0,0 +1,46 @@
...
+
+To set the value of a feature flag for everyone for a given environment, follow this process:
+
+1. **Click on "Add Rule"**
+  ![Screenshot showing where to find "Add Rule"](feature-flags-set1.png)
+2. **Choose "Targeted release"**
+  ![Screenshot showing how to select "Targeted release"](feature-flags-set2.png)
+3. **Select environments**: if you want this value to apply for all environments, select "all", otherwise, select the environment you want (e.g. staging/production)
+  ![Screenshot showing the rule definition page](feature-flags-set3.png)
+4. **Select the value you want to set**
```
nabramow, line 67: Do you think we could shorten this? It seems not such a complicated concept that we need so much text? My brain immediately glazed over and rebelled lol.

For example I don't think we need so much intro as most people get the concept of a feature flag, then we can jump right into a code snippet example and maybe a short succinct list format of what to do in most average case.

This makes me confused do I always need permission to add to GrowthBook? It seems only for non-default? I need GrowthBook if I want staging only?

I'd write it from the perspective of "someone has finished code, got told they need a feature flag, and now they're trying to quickly figure out how to do one (it me)".

nabramow (approved):

PR #9507, opened 2026-08-10
Make open profile more obvious on mobile

nabramow:
Closes [couchers/bugs#326 ](https://github.com/Couchers-org/bugs/issues/326)

We're getting multiple support (and instagram ) help requests that people can't tell how to open profiles from mobile map search. This PR hopefully makes it more obvious.

I decided to refactor `ProfileLink` to handle this better as I think it's relevant anywhere it's used, the profile sheet vs new tab behavior etc. That cut down on the conditional code.

I also decided to use the right chevron icon for mobile as the "open in new tab" icon has a specific meaning "opens it somewhere else" and on mobile it's not opening in the browser.

Screenshot 2026-08-10 at 19 49 12

## Testing
- Go to mobile responsive view in browser
- Search a location
- Should be able to open the profile by either clicking the avatar, the name, or the right chevron icon

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/ProfileLink/ProfileLink.test.tsx
+++ b/app/web/components/ProfileLink/ProfileLink.test.tsx
@@ -40,7 +40,7 @@ describe("ProfileLink", () => {
     expect(screen.getByRole("link")).toBeInTheDocument();
   });
 
-  it("renders a button on native when userId is provided", () => {
+  it("renders a link to the profile on native so it can be opened directly too", () => {
     (useIsNativeEmbed as jest.Mock).mockReturnValue(true);
     render(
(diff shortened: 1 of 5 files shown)


jesseallhands (approved):
This lgtm but I did find myself accidentally tapping the expand drawer arrow button which is near the carrot in the upper right corner of the card.

I know this should be in another ticket, but: I think we need to move the expand drawer arrow to the middle of the drawer and have it overlapping both the drawer and the map so people see it is there - this would solve the aforementioned problem and the other issue that people often overlook it.

PR #9513, opened 2026-08-11
Frontend/i18n: Don't use <Trans> where t suffices

tristanlabelle:
Quoting from the i18next documentation:

> As long you have no React/HTML nodes integrated into a cohesive sentence (text formatting like strong, em, link components, maybe others), you won't need it - most of the times you will be using the good old t function.

So let's not use <Trans> unless we need it.

## Testing

**Backend checklist**

**Web frontend checklist**

## For maintainers


tristanlabelle:
Ugh we need `<Trans>` because of the embedded `<strong>` tags

PR #9516, opened 2026-08-11
Frontend: Add full date on hovering relative times.

### LINKED ISSUE #9258: Frontend: Add full date tooltips to relative dates
Whenever we have "2 days ago" or similar, it'd be nice to be able to hover it and see the exact localized datetime.

tristanlabelle:
Relative times like "3 days ago" are nice for human readability, but it can be useful to also see the actual full date as hover.

I added an unstyled React component so we can do this uniformly across the website. Different pages style it as necessary. This supersedes the barely used `<TimeInterval>` component.

Note that this reveals the precise last login time which might be a little creepy, but was still known by the frontend, just not displayed fully.

Fixes #9258

## Testing
Ran frontend locally:

image
image
image
image
image
image

**Web frontend checklist**

## For maintainers


app/web/components/RelativeTime.tsx
```diff
@@ -0,0 +1,29 @@
...
+    instant = timestampToInstant(instant);
+  }
+  const {
+    t,
+    i18n: { language: locale },
+  } = useTranslation(GLOBAL);
+
+  return (
+    <Tooltip title={localizeDateTime(instantToPlainDateTime(instant), locale)} placement="top" arrow>
+      <span className={className}>{localizeRelativeTime(instant, locale, { t, ...options })}</span>
```
nabramow: If you make this a MUI `Typography` component it will automatically handle the sx prop, and also will already have bold I believe as an option rather than hand-rolling it.

app/web/features/auth/logins/LoginCard.tsx
```diff
@@ -60,19 +59,43 @@ export default function LoginsPage({ session }: { session: ActiveSession.AsObjec
...
-          text={<Trans t={t} i18nKey="auth:active_logins.expiry" values={{ expiry_datetime: expiryDisplay }} />}
+          text={
+            <Trans
+              t={t}
+              i18nKey="auth:active_logins.expiry2"
+              components={{
+                datetime: (
+                  <strong>
+                    <RelativeTime instant={session.expiry!} />
+                  </strong>
```
nabramow: The more React way to do this would be to pass a prop for the style so `sx={{ fontWeight: 600}}` (allows you to pass other styles later, or if it's just going to be bold `isBold={true}` and then handle it in the `RelativeTime` component. Unless they are all bold (it seems like they are?), then just make it bold in RelativeTime.

app/web/features/landing/SocialProof.tsx
```diff
@@ -1,17 +1,16 @@
...
 import { Box, Divider, Skeleton, Typography, useMediaQuery } from "@mui/material";
 import { useTranslation } from "i18n";
-import { localizeRelativeTime } from "i18n/datetimes";
 import { GLOBAL, LANDING } from "i18n/namespaces";
+import { Trans } from "react-i18next";
 import { Temporal } from "temporal-polyfill";
 import { theme } from "theme";
 import useSignupPageInfo from "utils/useSignupPageInfo";
 
+import RelativeTime from "../../components/RelativeTime";
```
nabramow:
```suggestion
import RelativeTime from "components/RelativeTime";
```

app/web/features/press/Facts.tsx
```diff
@@ -1,11 +1,12 @@
 import { Favorite, Language, People, Star } from "@mui/icons-material";
 import { Box, Skeleton, styled, Typography } from "@mui/material";
 import { useTranslation } from "i18n";
-import { localizeRelativeTime } from "i18n/datetimes";
 import { LANDING, PRESS } from "i18n/namespaces";
+import { Trans } from "react-i18next";
 import { Temporal } from "temporal-polyfill";
 import useSignupPageInfo from "utils/useSignupPageInfo";
 
+import RelativeTime from "../../components/RelativeTime";
```
nabramow:
```suggestion
import RelativeTime from "components/RelativeTime";
```

app/web/components/RelativeTime.tsx
tristanlabelle: The styling of relative times isn't uniformized across the app, it depends where they appear, so my goal is only to abstract the relative time + full time tooltip pattern and let the styling to the context. I probably don't need the `span className` part, claude generated it that way.

app/web/features/auth/logins/LoginCard.tsx
tristanlabelle: See https://github.com/Couchers-org/couchers/pull/9516#discussion_r3759295377 . I don't want to uniformize the styling of the text, but rather extract the logic so that every time a relative time appears in some text of the UI, the user can expect to be able to hover it. In that case it made sense for me to keep the styling outside of `<RelativeTime>`.

The `<strong>` was previously in the `en.json`, I pulled it out because it simplifies the syntax in the string. Is there a general problem with our using `<strong>` around interpolated string elements?

tristanlabelle:
@nabramow In the latest I made `<RelativeTime>` into a more typical unstyled React component and updated call sites to style it as necessary, since there is no commonality between the different stylings it uses across the app.

app/web/features/auth/logins/LoginCard.tsx
```diff
@@ -60,19 +60,39 @@ export default function LoginsPage({ session }: { session: ActiveSession.AsObjec
             <Trans
               t={t}
               i18nKey="auth:active_logins.location"
```
nabramow: I don't think there is an `active_logins.location` anymore from what I can see, only `location2`. Does this one need to be updated?

Also was that intentional to have a `location2` but no original `location(1)`?

app/web/features/landing/SocialProof.tsx
```diff
@@ -1,17 +1,15 @@
 import { Favorite, Language, Star } from "@mui/icons-material";
 import { Box, Divider, Skeleton, Typography, useMediaQuery } from "@mui/material";
+import RelativeTime from "components/RelativeTime";
 import { useTranslation } from "i18n";
-import { localizeRelativeTime } from "i18n/datetimes";
 import { GLOBAL, LANDING } from "i18n/namespaces";
+import { Trans } from "react-i18next";
```
nabramow:
```suggestion
import { Trans } from "i18n";
```

app/web/features/press/Facts.tsx
```diff
@@ -1,8 +1,9 @@
 import { Favorite, Language, People, Star } from "@mui/icons-material";
 import { Box, Skeleton, styled, Typography } from "@mui/material";
+import RelativeTime from "components/RelativeTime";
 import { useTranslation } from "i18n";
-import { localizeRelativeTime } from "i18n/datetimes";
 import { LANDING, PRESS } from "i18n/namespaces";
+import { Trans } from "react-i18next";
```
nabramow:
```suggestion
import { Trans } from "i18n";
```

app/web/features/messages/messagelist/MessageView.tsx
```diff
@@ -169,7 +171,7 @@ export default function MessageView({ className, message, onVisible, isDm = fals
           </TextBody>
         </StyledMessageBody>
         <StyledFooter>
-          <StyledTimeInterval instant={timestampToInstant(message.time!)} />
+          <StyledRelativeTime instant={timestampToInstant(message.time!)} />
```
nabramow: `MessageView` wraps in `timestampToInstant(...)` while `ControlMessageView` passes `message.time!` straight through. Pick one to keep it consistent?

app/web/components/RelativeTime.tsx
```diff
@@ -0,0 +1,37 @@
...
+    t,
+    i18n: { language: locale },
+  } = useTranslation(GLOBAL);
+
+  const plainDateTime = instantToPlainDateTime(instant);
+
+  return (
+    <Tooltip title={localizeDateTime(plainDateTime, locale)} placement="top" arrow>
+      <time dateTime={plainDateTime.toString()} {...rest}>
+        <>{localizeRelativeTime(instant, locale, { capitalize, smallestUnit, t })}</>
```
nabramow:
```suggestion
        {localizeRelativeTime(instant, locale, { capitalize, smallestUnit, t })}
```

app/web/components/RelativeTime.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Unstyled component displaying an instant in time as a relative time (e.g. "3 hours ago"),
+ * with a tooltip showing the full date and time.
+ */
+export default function RelativeTime({ instant, smallestUnit, capitalize, ...rest }: RelativeTimeProps) {
+  if (!(instant instanceof Temporal.Instant)) {
+    instant = timestampToInstant(instant);
+  }
```
nabramow, line 22: You should never update state directly in React:

https://react.dev/learn/updating-objects-in-state

app/web/components/RelativeTime.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const {
+    t,
+    i18n: { language: locale },
+  } = useTranslation(GLOBAL);
+
+  const plainDateTime = instantToPlainDateTime(instant);
+
+  return (
+    <Tooltip title={localizeDateTime(plainDateTime, locale)} placement="top" arrow>
+      <time dateTime={plainDateTime.toString()} {...rest}>
```
nabramow: Do we want this to be:

```suggestion
      <time dateTime={instant.toString()} {...rest}>
```

Would that handle the time offset better and make it less ambiguous?

If I understand correctly, the attribute describes the reader's clock, not the data, so two people viewing the same page disagree about when it happened?

app/web/features/auth/logins/LoginCard.tsx
```diff
@@ -60,19 +60,39 @@ export default function LoginsPage({ session }: { session: ActiveSession.AsObjec
...
           text={
-            <Trans t={t} i18nKey="auth:active_logins.last_activity" values={{ last_activity_ago: lastSeenDisplay }} />
+            <Trans
+              t={t}
+              i18nKey="auth:active_logins.last_activity2"
+              components={{
+                timeAgo: (
+                  <strong>
+                    <RelativeTime instant={session.lastSeen!} />
+                  </strong>
```
nabramow, line 79: Re `strong` like html like this:

The tag isn't the problem,  `<strong>` is semantically correct. The problem is that it has content inside it, which is what makes it feel like markup dumped into the component.

The repo's own convention so far is an empty element, with the text living in the translation string, i.e. `LanguagePickerSettings.tsx:30`, `WhatIsCouchSurfingPage.tsx:284`.

For plain values (location, expiry) that gives you:

```
"location2": "Near <bold>{{approximate_location}}</bold>"
```

For `Trans` components:

```
<Trans t={t} i18nKey="auth:active_logins.location2"
  values={{ approximate_location: session.approximateLocation }}
  components={{ bold: <strong /> }} />
```

Better also because the translator can see the placeholder and reposition it, and `<strong>` goes back to being purely presentational.

For `last_activity` the content is a React component, so it can't be a values interpolation, but you can still separate styling from content by nesting the tags in the string:

```
"last_activity2": "Last activity <bold><timeAgo/></bold>"
```

```
components={{ bold: <strong />, timeAgo: <RelativeTime instant={session.lastSeen!} /> }}
```

Not sure if this would come up in this case, but the pattern more generally...that also lets a translator drop the emphasis if their language reads better without it, which is impossible when the bold is welded to the value in TSX.

app/web/components/RelativeTime.tsx
tristanlabelle: This is not modifying state, it's supporting the type overloading of the property between Instant or Timestamp. Only local variables are involved and it's side-effect free.

app/web/components/RelativeTime.tsx
tristanlabelle: Done in latest (I thought I got a squiggle without it)

app/web/features/messages/messagelist/MessageView.tsx
tristanlabelle: Right, I added the overloading later. Updated.

app/web/features/auth/logins/LoginCard.tsx
tristanlabelle: Good catch, updated to `location2`. Whenever I bump a string key, it's because placeholders changed and so we don't want to reuse previous translations.

app/web/components/RelativeTime.tsx
tristanlabelle: Good catch, updated.

app/web/features/auth/logins/LoginCard.tsx
tristanlabelle: This was a conscious decision, I made the change because I needed to tweak the strings' placeholders and invalidate translations anyways. Hear me out:

Those strings have `<tag>{{placeholder}}</tag>` structure where the tag is used specifically for styling the placeholders, and not for semantics. In fact they should be using `<b>` (stylistic) rather than `<strong>` (semantic) since a screen reader wouldn't need to call out the placeholder.

Strings shouldn't dictate styling. If we wanted to show a timestamp in a pill, we wouldn't want to have every string say `<pill><timestamp/></pill>`. Moreover, since it's for styling a placeholder, the location of the `<tag>`'s won't change per language, so every translation will need to get the complex `<tag>{{placeholder}}</tag>` syntax right.

I quite favor this approach for those cases (styled placeholder).

tristanlabelle:
@nabramow I addressed or replied to all of your comments. Ready for re-review.

PR #9520, opened 2026-08-11
Return search to idle state when the location input is cleared

### LINKED ISSUE #9518: Frontend/search: Clearing the location input runs a search instead of returning to the idle state
## Summary

On the search page, clearing the location input performs a search. It should instead return to the idle state and show the "Type a location or apply a filter to search." prompt, exactly as it does on first load.

The search it runs happens to be an unfiltered global one, so it is also the slowest possible query — the page spins for a couple of seconds and then jams while rendering the results. Reproduces on production.

## Steps to reproduce

1. Go to https://couchers.org/search and switch to list view.
2. Note the prompt: **"Type a location or apply a filter to search."** — correct, the page is idle and waiting for input.
3. Search for a location (e.g. "Germany"). Results load — correct.
4. Clear the text out of the location input so the box is empty again.

## Expected

The page returns to the idle state from step 2 and shows "Type a location or apply a filter to search." again. There is no location, no keyword and no filter, so there is nothing to search for.

## Actual

The page runs a search. A spinner appears for ~2 seconds, then the UI locks up while it renders a full page of results.

## Cause

`meetsSearchCriteria` is wrongly `true` after the clear. It derives from `hasActiveFilters`, and there are **two independent defects that each make `hasActiveFilters` true when no filter is set**.

`getHasActiveFilters` (`features/search/utils/mapUtils.ts`) works by diffing `state.filters` against the module-level `initialState.filters`, so any value that doesn't match that baseline exactly reads as an active filter.

### 1. The provider seeds filters from a sparse URL query

`MapSearchProvider` (`features/search/state/mapSearchContext.tsx`) does:

```js
filters: initialFilters,
```

`initialFilters` is parsed from the URL query and is sparse — on a plain `/search` visit it is `{}`. But `initialState.filters` has `lastActive: 0`. So `state.filters.lastActive` is `undefined` while the baseline is `0`, and `undefined !== 0` makes `getHasActiveFilters` return `true`.

This also explains why the bug only appears *after* an interaction rather than on load. On first load the provider sets `hasActiveFilters` explicitly via `Object.keys(initialFilters).length > 0`, which is correctly `false`, and `getHasActiveFilters` is never consulted. It is only when `CLEAR_SEARCH_INPUT_VALUE` **recomputes** `hasActiveFilters` that the mismatch surfaces — which is exactly why the idle prompt shows on load but not after clearing.

### 2. The clear handler writes literals instead of the initial values

`CLEAR_SEARCH_INPUT_VALUE` in `features/search/state/mapSearchReducers.ts` resets the filters that a location search had defaulted on, but writes literals:

```js
hostingStatus: undefined,
showEmptyProfile: false,   // initialState.filters.showEmptyProfile is `undefined`
```

`false !== undefined`, so this trips the same comparison on its own. `CLEAR_KEYWORD_INPUT_VALUE` has the identical pattern.

This one has a second effect too: `showEmptyProfile: false` is a real filter — `service/search.ts` turns it into `setProfileCompleted(true)`. So the request that goes out is "every completed profile on Earth, sorted by recommendation score, with no bounding box", which is part of why it is so slow.

**Fixing either defect alone leaves the bug in place**, since either one is sufficient to make `hasActiveFilters` true.

## Fix

- Layer the sparse URL filters over the canonical defaults in the provider: `filters: { ...initialState.filters, ...initialFilters }`.
- Reset to `initialState.filters.*` rather than literals in both clear cases in the reducer.
- Additionally, pass `enabled: meetsSearchCriteria` to the `useInfiniteQuery` in `useUserSearch`. This is defensive rather than the fix itself: the hook already discards its own results when the criteria aren't met, so it should not be fetching in the first place.

## Related

Separately, user search is far slower than it should be even for legitimate searches — tracked in the companion backend issue. This bug made that very visible but is not its cause.

wbruntra:
The search page shows an idle prompt until a location or a filter is set. Clearing the location input did not return to that state — the page treated the emptied form as an active search and ran a global query with no bounding box, so it spun for a couple of seconds and then went briefly unresponsive rendering the results. An emptied form is now recognised as having no criteria, so the idle prompt returns and no request is made.

closes #9518

## Testing

Load `/search`, switch to list view, search a location, then clear the input.

**Web frontend checklist**

## For maintainers


papeldeorigami (changes requested):
@wbruntra sorry for the very late review.
I was able to confirm that the fix works - but while I was testing this fix I noticed that it puts the search screen in an infinite loading state if I simply press reload on the browser window.

## Confirmation that the fix works
I can see that the search endpoint is not called anymore after I press the clear button.

before | after
-- | --
UserSearchV2 endpoint was called Image |  Not called anymore Image

## Issue with infinite load
before | after
-- | ---
https://github.com/user-attachments/assets/782fddd0-7936-4c55-996c-28dd98938e77 | https://github.com/user-attachments/assets/13224086-6d7c-4056-a2b1-60193e783763

I have the impression that this could be related to the `enabled` condition that was added to the query hook, but I didn't investigate it further.

## Final comments

Although this fix might work, I think that we should rather rethink how those default filters are applied:
- Another bug remains: I can still see that the default filters suddenly override my own filters if I zoom in and out.
- Could be improved: the current logic is "verbose" (it took me a while to figure out why it was there in the first place) and it is duplicated for the two types of clear actions

PR #9522, opened 2026-08-11
Virtualize search results list to stop the UI freezing

### LINKED ISSUE #9521: Search page freezes for ~2s when results render (100 cards, no virtualization)
## Problem

Searching on `/search` feels slow, and the search input freezes while results load. This reads as a slow backend, but it isn't one — it's the frontend blocking the main thread while it renders result cards.

**The request completes in ~500–700ms. The browser then blocks for 2+ seconds rendering.** While that single long task runs, the browser can't process any input, which is the "input box freezes" symptom.

The clearest evidence: paging **backwards** to results already held in the react-query cache — zero network requests, backend not involved at all — still blocks the main thread for **2.2–2.6 seconds**.

## Cause

The results list renders all 100 results at once, with no virtualization, when only ~6–8 cards fit on screen. `MAX_PAGINATION_LENGTH = 100` in `couchers/servicers/search.py`, and the web client never calls `req.setPageSize()` for user search (`app/web/service/search.ts` — `pageSize` is in the `UserSearchFilters` interface but is only applied in `EventSearch`).

That's ~3,358 DOM nodes rebuilt on every search. Measured, holding server time constant:

| Results rendered | Server TTFB | Blocking task | Time to painted |
|---|---|---|---|
| 4 (Andorra) | ~500ms | ~172ms | ~1.1s |
| 100 (Paris) | ~500ms | ~2,100–2,350ms | ~4.0s |

Same backend cost, ~12x the freeze.

A V8 CPU profile of a production search showed two things:

1. `react-lines-ellipsis` internals as the top JS self-time entries — `calcIndexes` 503ms, `putEllipsis` 409ms, plus `reflow` 147ms. `<LinesEllipsis basedOn="letters">` in `SeachResultUserCard` clamps text by binary-searching the cut-off point while measuring the DOM, forcing a synchronous layout per iteration, per card.
2. After that, the remaining time is dominated by `(program)` — browser style recalculation, layout and paint, not JS. That's the cost that scales with DOM size.

## Measurements

Taken with headless Chromium, logged in as a real user, using `PerformanceResourceTiming` to split server think time from download and `PerformanceObserver({type:"longtask"})` for main-thread blocking. Production numbers are from couchers.org with real data (~89k visible users). These are single-client measurements and say nothing about behaviour under concurrent load.

| Cards rendered | DOM nodes in cards | Total page DOM | Blocking task |
|---|---|---|---|
| 100 | 3,358 | 4,695 | 635ms* |
| 20 | 680 | 1,941 | 275ms* |

\* local dev build; absolute numbers are inflated vs production, only the ratio is meaningful.

## Suggested fix

Virtualize the results list so only visible rows are mounted, keeping all 100 results loaded so map pins are unaffected. This is preferable to just lowering the page size, since the map legitimately wants all 100 users as pins — virtualization decouples "how many results we have" from "how many DOM nodes we build."

Separately, replacing `react-lines-ellipsis` with CSS `line-clamp` removes the forced-layout cost with no visual change.

wbruntra:
Virtualizes the search results list so only the rows actually on screen are mounted, and replaces `react-lines-ellipsis` with CSS `line-clamp` in the result card.

Searching felt like a slow backend, but the request completes in ~500–700ms and the browser then blocks for 2+ seconds rendering results. The list rendered all 100 results at once (~3,358 DOM nodes) when only 6–8 cards are ever on screen. While that single long task runs the browser can't process input, which is why the search box appears to freeze.

All 100 results stay loaded, so map pins are unaffected — this only changes how many cards exist in the DOM.

closes #9521

### Results

Measured against the **production** backend, Paris (974 matches, 100 returned):

| | Cards mounted | DOM nodes in cards | Total page DOM | Blocking task |
|---|---|---|---|---|
| Before | 100 | 3,358 | 4,695 | ~2,100–2,400ms |
| After | **4** | **134** | **1,374** | **270ms** |

Changing page on an already-open search — the interaction that produced the freeze complaint:

| | Blocking task |
|---|---|
| Before | 1,352–2,583ms |
| After | **73–84ms** |

### Notes for reviewers

- The grid was `auto-fill`, so the browser chose the column count. Virtualizing by row needs that count, so it's now measured with a `ResizeObserver`.
- `ResizeableDrawer` takes an optional `scrollRef` so the virtualizer can attach to the element that actually scrolls.
- `setSelectedUserId` scrolled via `document.getElementById(...).scrollIntoView()`. Off-screen cards are no longer in the DOM, so that moved to `rowVirtualizer.scrollToIndex()` in the list, keyed on `selectedUserId`.
- Desktop cards are fixed-height so rows use a fixed size; mobile cards are auto-height and are measured via `measureElement`.
- Adds one dependency, `@tanstack/react-virtual` (TanStack Query is already used).

## Testing

Ran the local frontend against the production backend to reproduce real result volumes (`.env.development.local` with `NEXT_PUBLIC_API_BASE_URL="https://couchers.org/api"` and `NEXT_PUBLIC_MEDIA_BASE_URL="https://usercontent.couchershq.org"`), searched Paris, and paged next/previous.

Verified with headless Chromium via `PerformanceObserver({type:"longtask"})` for main-thread blocking, plus DOM node counts and V8 CPU profiles:

- Cards recycle correctly while scrolling (mounted card ids change; total mounted stays ~4–7), and the full scroll height is correct for 100 cards.
- Narrow drawer renders 1 column; expanded drawer renders 2 columns, still virtualized.
- Mobile (390×844) and tablet (820×1180): auto-height rows measure correctly (187–224px) with **zero overlapping cards** after scrolling.
- No page errors in any viewport.

**Web frontend checklist**

## For maintainers


wbruntra:
Addressed some issues with previous commit. Detail below.

**1. `measureElement` was dead code on desktop — fixed.**

Confirmed in `virtual-core`: `setOptions` merges with `if (v !== void 0) merged[key] = v`, so passing `undefined` keeps the default `offsetHeight` measurement rather than disabling it. And since the ref is only attached on mobile, the desktop `() => 361` was never called.

Removed the option entirely rather than fixing its signature — desktop rows are fixed height, so `estimateSize` alone is correct, and mobile gets the default dynamic measurement. Fewer moving parts than keeping an override that has to be explained.

**2. Scroll position on pagination — decided: reset to top.**

Landing part-way down a fresh set of results is wrong. Keyed on `pageNumber` from the reducer rather than on `users`, because `users` is `data?.pages[...]?.resultsList || []` — the `|| []` mints a new array whenever data is undefined, so an effect keyed on it would fire every render and fight with scrolling.

Verified: scrolled to 5000px, clicked Next, scroll position is 0.

**3. Column flash — fixed.** Measurement moved to `useLayoutEffect`, so the first paint has the right column count. `useLayoutEffect` is already used elsewhere in this codebase (`AppRoute`, `InfiniteMessageLoader`, `ProfileSheet`) and no SSR warning appears.

**4. Stale size cache across breakpoints — fixed.** `rowVirtualizer.measure()` on `isMobile` change; it clears `itemSizeCache`, `laneAssignments` and bumps the cache version.

Worth noting this was observable, not just theoretical. Resizing desktop → mobile → desktop, rows now measure back to exactly 345px; without the fix the stale ~200px mobile entries persisted for un-scrolled rows.

**Re-verified after the changes:** tsc, eslint, prettier, 22 search tests; pagination blocking task still 84ms with 4 cards mounted; mobile/tablet/desktop all zero overlapping cards; no page errors, no React console warnings.

app/web/features/search/SeachResultUserCard.tsx
```diff
@@ -159,14 +168,7 @@ const generateAboutText = (user: SearchUser.AsObject, t: TFunction, isMobile: bo
...
   } else {
-    return (
-      <LinesEllipsis
-        maxLine={isMobile ? 3 : 7}
-        text={stripMarkdown(aboutText(user, t))}
-        basedOn="letters"
-        style={{ wordBreak: "break-all", overflow: "hidden" }}
-      />
-    );
+    return <ClampedAbout maxLine={isMobile ? 3 : 7}>{stripMarkdown(aboutText(user, t))}</ClampedAbout>;
```
papeldeorigami, line 171: I want to say thank you for bringing our attention to this file, because that helped us to identify a regression and address it with #9540. PTAL.

This is actually the main part of your PR that I'd be interested in getting merged. I would appreciate a concise PR focused on that improvement alone, but personally I would prefer evidence that it was tested before/after by a human (like screenshots, videos), not just headless Chrome.

app/web/features/search/SearchResultListContent.tsx
```diff
@@ -34,13 +43,22 @@ const ListContentWrapper = styled(Box, {
...
+  paddingBottom: theme.spacing(2),
+}));
+
+// A row of `columns` cards on desktop, a single card on mobile.
+const VirtualRow = styled("div", {
+  shouldForwardProp: (prop) => prop !== "columns",
+})<{ columns: number }>(({ theme, columns }) => ({
   display: "grid",
-  gridTemplateColumns: `repeat(auto-fill, minmax(${DEFAULT_DRAWER_WIDTH - 50}px, 1fr))`, // Responsive columns
+  gridTemplateColumns: `repeat(${columns}, minmax(0, 1fr))`,
```
papeldeorigami: Why are we touching the layout? I thought this was a bugfix to a performance issue.

app/web/features/search/SearchResultListContent.tsx
```diff
@@ -34,13 +43,22 @@ const ListContentWrapper = styled(Box, {
...
 
+// Full-height spacer that the virtual rows are absolutely positioned within.
 const UserCardsWrapper = styled("div")(({ theme }) => ({
+  position: "relative",
+  width: "100%",
+  paddingBottom: theme.spacing(2),
+}));
+
+// A row of `columns` cards on desktop, a single card on mobile.
+const VirtualRow = styled("div", {
```
papeldeorigami, line 54: Why are we touching the layout? This was unexpected in a bugfix to a performance issue. TBH this entire switch to a columns card, "virtualizing" rows, etc. is so unclear to me. If it's really the cause of the "freezing UI", I wonder how and when the bug was introduced.

app/web/package.json
```diff
@@ -36,13 +36,14 @@
...
     "@mui/x-date-pickers": "^9.10.0",
-    "@stripe/stripe-js": "^9.13.0",
     "@sentry/nextjs": "^10.66.0",
     "@sentry/react": "^10.62.0",
+    "@stripe/stripe-js": "^9.13.0",
     "@tanstack/query-async-storage-persister": "^5.101.4",
     "@tanstack/react-query": "^5.101.4",
     "@tanstack/react-query-devtools": "^5.101.4",
     "@tanstack/react-query-persist-client": "^5.101.4",
+    "@tanstack/react-virtual": "^3.14.9",
```
papeldeorigami, line 46: Related to my question about the virtualization approach, this is yet another side-effect - we are adding an extra dependency to resolve an issue that we cannot even tell the root cause. Perhaps it does make sense, but let's first revisit what is actually the bug, what introduced it, then we could weigh pros/cons of adding react-virtual for this particular problem. Personally, it feels like an overkill to me - if the problem is the number of results per page, let's just reduce them.

papeldeorigami (changes requested):
@wbruntra I really appreciate the fact that you took the time to try to resolve an important issue affecting the project. I added a few specific comments below, but I want to call out a few general issues that impacted me as a reviewer:
- it feels like AI was used heavily in the task description https://github.com/Couchers-org/couchers/issues/9521 - it contains cryptic text and even some measurements, instead of focusing on the fact that "search page is freezing when I do XYZ", like people usually do, sharing how to reproduce it, sometimes with screenshot (a short video here is more valuable than a thousand numbers). In fact, I'd appreciate it if you could check if the issue is still present, and rewrite it in simple terms in that case.
- the same verbose and hard-to-understand problem description leaked into the PR description, and led to an implementation that touches two completely distinct areas of the code, even adding a new dependency. It's hard to tell the cause-effect of the problem (as the problem itself is put in abstract terms).
- I see the results of the tests with Chrome headless, but I would have also appreciated some screenshots taken by you as you were testing the fix (a "before and after", as usual)

I'd be happy to support you in case the problem persists, but we'll need to decide what we are solving first.

Last, but not least, this PR helped us to find a performance regression and revert it, so PTAL at #9540 and double-check if the issue you are trying to solve is still there before investing more time. Thank you very much!

wbruntra:
Thank you for the review, @papeldeorigami . You're right that it would have been better to post a video rather than that convoluted AI-generated description. This is me writing now, not Claude.

So the problem did originate not just from measured benchmarks, but from my actual experience using the site, which is that there was a large delay when loading results. Even with your recent changes, I am still noticing the delay, so please take a look at the video capture that I made.


[Screencast from 08-15-2026 10:28:45 AM.webm](https://github.com/user-attachments/assets/97abb6f6-b26d-42d7-98d6-44f5a5997390)

You will see that the UI freezes and becomes non-responsive. At one point I am entering text in the search box and it is not appearing until after the results are fully loaded.

I should note that this video was made with Chrome DevTools open. Navigating the site normally (devtools closed) I do not see these long freezes (it is still freezing, but not nearly as long).

wbruntra:
Here is a new screencast, 

[Screencast from 08-15-2026 10:48:22 AM.webm](https://github.com/user-attachments/assets/0c9bc688-bcc4-491b-8f95-1e2e90b6e40b)

I used AI for this PR because honestly I don't have the technical chops to diagnose or fix the issue on my own. So I need to quote Claude here because I myself didn't find the issue:

> The old code truncated the About text in JavaScript. To find where to cut, react-lines-ellipsis copies the profile into a hidden element off-screen, wraps every single word or letter in its own <span>, then walks through those spans one at a time asking the browser "which line is this on?" Each of those questions forces the browser to stop and recalculate layout. For a long profile that's thousands of interruptions — per card, times 100 cards.

> CSS line-clamp asks the browser to do the same job as part of the layout it was already doing.

Performance does seem much better with just that change, so re-doing everything to use list virtualization could be overkill and introduce bugs that are not obvious right now.

app/web/features/search/SearchResultListContent.tsx
wbruntra: Specifically on the question of "when the bug was introduced" and why the recent reversion didn't fix the issue, this answer sounds correct to me. Claude:

>  #8912 (2026-06-09) set basedOn="letters", which made the clamping per-character. #9540 removed that prop — but removing it doesn't select word mode, it restores the library's default, which sniffs the text:
> 

`var basedOn = props.basedOn || (/^[\x00-\x7F]+$/.test(props.text) ? 'words' : 'letters');`
> That regex asks "is this string entirely ASCII?" It's a proxy for "does this language use spaces between words" — sensible for CJK, but it's all-or-nothing over the whole profile. One non-ASCII character anywhere puts it back on the per-character path.
> 
> That includes the curly apostrophe ' (U+2019) that iOS and Android insert automatically in "I'm", plus emoji, em dashes, and any accented name. So a plain-English profile typed on a phone is still on the slow path today.
> 
> It matters because cost scales with unit count — the library builds one <span> per unit and reads offsetTop on each. On ~2,800 characters that's ~1,140 spans in word mode versus ~3,200 in letter mode.

wbruntra: Yes, this PR really had two parts, one was changing the way text was parsed, the second part was virtualizing the list. Based on my own testing in the dev environment, it seems like the first change addresses most of the issue I was experiencing when browsing the site. The second part may theoretically improve performance but may also introduce bugs and it's *not* the cause of the performance issue I was seeing.

papeldeorigami:
Closing so we can address it with #9541

PR #9528, opened 2026-08-12
Fix cookie banner overlapping message composer

kevinortiz43:
Fixes bugs#363: the cookie-consent banner overlapped the message composer on the group-chat and host-request views, blocking users from typing/sending messages while the banner was showing.

GroupChatView and HostRequestView now subtract `var(--cookie-banner-height, 0px)` from their 100dvh-based height calc, matching the pattern already used by BottomNavigation, LandingPage, and the profile-edit forms.

AppRoute.tsx: reordered the pre-existing `isMounted`/`isPrivateRouteNotReady` declarations above the nav-height `useLayoutEffect` so that effect's dependency array can include `isPrivateRouteNotReady` to measure the height change when the header mounts. No new logic, only their declaration order moved.

Closes https://github.com/Couchers-org/bugs/issues/363

## Testing
Verified locally in a browser and android width with the cookie banner visible, opened a group chat and a host request.

Before
image

image

image

image

After

image

image

image

image

**Web frontend checklist**

## For maintainers


kevinortiz43:
GitLab failed, going to see why and fix, was working okay before I merged the latest change from develop branch

app/web/features/messages/groupchats/GroupChatView.tsx
```diff
@@ -50,11 +50,11 @@ const StyledPageWrapper = styled("div")<{
         // Use dvh (dynamic viewport height) which adjusts for mobile keyboard
         // Use CSS custom property set by Navigation component for actual height
         height: isNativeEmbed
-          ? "calc(100dvh - var(--nav-height, 3.5rem))"
-          : `calc(100dvh - var(--nav-height, 3.5rem) - ${BOTTOM_NAV_BASE_HEIGHT}px - env(safe-area-inset-bottom, 0px))`,
+          ? "calc(100dvh - var(--nav-height, 3.5rem) - var(--cookie-banner-height, 0px))"
+          : `calc(100dvh - var(--nav-height, 3.5rem) - ${BOTTOM_NAV_BASE_HEIGHT}px - env(safe-area-inset-bottom, 0px) - var(--cookie-banner-height, 0px))`,
```
papeldeorigami, line 54: We'd have to repeat this math all across the app. It's easy to miss - and it's boring to list everything that can change the view size over and over again.

papeldeorigami (commented):
@kevinortiz43 this works, but I can imagine the calculation being missed somewhere else. In fact, I just fixed a similar thing for a missing inset the other day, so it seems to be a recurrent type of problem.
How about introducing a `StyledFullHeightView` reusable component? We could do that in a separate refactor PR, but since the cookie issue doesn't seem to be urgent, it could be a good opportunity to test the approach immediately.
Let me know your preference.

kevinortiz43:
@papeldeorigami I agree, introducing a reusable component is a smarter approach. We just need to create the task then get it on the kanban board and get it on the "ToDo" list. So this PR can merge in because it solves an existing problem then the reusable component is the ultimate goal. I think we need to talk to @nabramow to see the next steps for the reusable component which I believe is just creating a task?

kevinortiz43:
I think I talked to someone and they mentioned that we should delete the cookie banner but I don't know how serious that statement was

kevinortiz43:
I mean testing the approach is simple, it's just a reusable component I think it will work easily, I just don't know if that refactoring is a priority

nabramow:
> @papeldeorigami I agree, introducing a reusable component is a smarter approach. We just need to create the task then get it on the kanban board and get it on the "ToDo" list. So this PR can merge in because it solves an existing problem then the reusable component is the ultimate goal. I think we need to talk to @nabramow to see the next steps for the reusable component which I believe is just creating a task?

I trust @papeldeorigami's judgement here. I also don't think this task is super urgent so makes sense to fix it once and for all and these issues keep popping up. I know you just got a new full-time job though so no worries at all if you need to bow out rather than start a new thing.

kevinortiz43:
@nabramow I can still help out if y'all want me to, will probably just be a bit slower throughout the week. So this PR will convert to a reusable cookies-banner component fix.

papeldeorigami (approved):
I think we have a consensus that this PR is better than the current state, so let's merge it. I can prepare that PR with the suggested refactor separately.

kevinortiz43:
@papeldeorigami I am already working on the DRY component

kevinortiz43:
@papeldeorigami but if you want it go for it, uh I just stopped coding so I can give you like my notes on where I found all the viewports where the cookiebanner code exists

PR #9530, opened 2026-08-12
Frontend/nav: Change lang picker to icon only

tristanlabelle:
As suggested by @c-kreuz , using only the standard translation icon for the language picker is less cluttered and just as recognizable if not more.

I kept the pill shape with the language name and dropdown arrow when in mobile menu view.

## Testing
Ran locally:

image

image

Logged out:
image

image

Account settings (unchanged):
image

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/features/translate/LanguagePickerSelect.tsx
+++ b/app/web/features/translate/LanguagePickerSelect.tsx
@@ -1,6 +1,6 @@
 import CheckIcon from "@mui/icons-material/Check";
 import ExpandMoreOutlinedIcon from "@mui/icons-material/ExpandMoreOutlined";
-import TranslateIcon from "@mui/icons-material/Translate";
+import LanguageIcon from "@mui/icons-material/Language";
 import {
   Box,
   FormControl,
@@ -205,7 +205,7 @@ export default function LanguagePickerSelect({ displayMode = "rounded", onNaviga
     );
   };
 
-  // Icon mode shows a generic translate icon instead of the selected language
+  // Icon mode shows a generic language icon instead of the selected language
   const renderIconValue = () => (
     <Box
       sx={{
@@ -215,7 +215,7 @@ export default function LanguagePickerSelect({ displayMode = "rounded", onNaviga
         color: "var(--mui-palette-text-primary)",
       }}
     >
-      <TranslateIcon fontSize="small" />
+      <LanguageIcon fontSize="small" />
     </Box>
   );
 
@@ -235,7 +235,7 @@ export default function LanguagePickerSelect({ displayMode = "rounded", onNaviga
               value={isLoading ? "" : locale || ""}
               displayMode={displayMode}
               onChange={handleChange}
-              // Use renderValue to display the selected language (or, in icon mode, a generic translate icon) in collapsed state
+              // Use renderValue to display the selected language (or, in icon mode, a generic language icon) in collapsed state
               renderValue={displayMode === "icon" ? renderIconValue : renderRoundedValue}
               IconComponent={ExpandMoreOutlinedIcon}
               disabled={isLoading || isChangingLanguage}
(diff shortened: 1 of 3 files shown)


tristanlabelle:
@claude update to using `import LanguageIcon from '@mui/icons-material/Language';` as the icon

app/web/features/translate/LanguagePickerSelect.tsx
```diff
@@ -54,16 +55,28 @@ const StyledSelect = styled(Select, {
...
 }));
 
 type LanguagePickerSelectProps = {
-  displayMode?: "round" | "rect";
+  displayMode?: "rounded" | "rect" | "icon";
   onNavigate?: () => void;
 };
 
-export default function LanguagePickerSelect({ displayMode = "round", onNavigate }: LanguagePickerSelectProps) {
+export default function LanguagePickerSelect({ displayMode = "rounded", onNavigate }: LanguagePickerSelectProps) {
```
nabramow, line 79: Why this change?

app/web/features/translate/LanguagePickerSelect.tsx
```diff
@@ -190,32 +203,53 @@ export default function LanguagePickerSelect({ displayMode = "round", onNavigate
...
       <Box sx={{ minWidth: 40 }}>
         <FormControl
           variant="outlined"
           sx={{
-            width: displayMode === "round" ? "fit-content" : !isMobile ? "241px" : "100%",
+            width: displayMode === "rect" ? (!isMobile ? "241px" : "100%") : "fit-content",
           }}
         >
-          {displayMode === "round" ? (
+          {displayMode !== "rect" ? (
```
nabramow, line 233: I guess we're only using this two places but will this be backwards compatible?

So far behavior is:

Desktop: Show in top nav bar
Mobile: Don't show in top nav bar, show in main menu at the bottom.

I guess for the mobile one we can keep it with the language name? Is that the idea?

app/web/features/translate/LanguagePickerSelect.tsx
tristanlabelle: The icon is now actually "round" (circular), whereas when showing the language name it's only "rounded".

app/web/features/translate/LanguagePickerSelect.tsx
tristanlabelle: If it appears in the menu then it's using the language name in a rounded button, because that menu is already all text. If it's in the top bar then it's the circular icon, because that is using icon buttons. Try it in the Vercel and let me know if it doesn't make sense.

HankAviator (dismissed):

HankAviator:
I'm seeing the language picker implies a 'focused' state, is it expected?
image

nabramow:
Hey @HankAviator just a heads up we're trying to stop non-engineers from approving engineering things. If it's a purely text translation PR that's fine, but since we had a major bug go through recently we need to get a bit stricter.

I'm about to test this myself then I can approve, but please don't approve PRs unless you can test and comment on the code, too. Better leave a comment in this case that you approve the translations then an engineer can also go approve the code.

nabramow:
Okay this is working well for me so far. Let's add the globe icon to the left of the text in the mobile view closed drop down too for consistency. Then the user will be able to recognize it by the icon in all views:

Screenshot 2026-08-18 at 13 49 58

tristanlabelle:
@HankAviator I agree the design is a bit confusing in this view. It's meant to indicate that it opens a menu. It makes more sense when logged in. We could use a design iteration to improve this.

Logged in:

image

Compare with before (same styling):

image

c-kreuz (commented):
Thanks for taking this on. I'm with @nabramow, we should add the icon next to the language on the mobile menu. Other than that I'm happy with the changes.

tristanlabelle:
@nabramow / @c-kreuz , latest adds the language icon to the mobile menu button styling:

image 

Ready for final review.

tristanlabelle:
Rebased to workaround flaky CI

nabramow (approved):
This looks good to me now!

PR #9532, opened 2026-08-13
Backend/events: Move event thread to EventOccurrence

### LINKED ISSUE #9547: [Backend] Migrate comment thread from Event to EventOccurrence
Part of #9544 (Milestone 1: Proof of concept with private beta).

tristanlabelle:
Part of supporting recurring events, every occurrence should have an independent thread rather than a shared one at the series level (occurrences mostly look like individual events to non-organizers).

All events today should have a single occurrence since we don't support recurrence yet, but for resiliency the database migration moves the thread to the latest occurrence and creates empty threads on others.

Fixes #9547

## Testing
Nuked database, then ran local backend + frontend before changes, created an event and added a comment. Synced to this branch and reran backend + frontend, observed that the event still has its comment.

Updated `test_ScheduleEvent` to verify that each occurrence has its own thread.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/migrations/versions/0181_move_event_thread_to_occurrence.py
+++ b/app/backend/src/couchers/migrations/versions/0181_move_event_thread_to_occurrence.py
@@ -1,0 +1,85 @@
+"""Move event comment thread from Event to EventOccurrence
+
+Revision ID: 0181
+Revises: 0180
+Create Date: 2026-08-13 07:21:39.779870
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0181"
+down_revision = "0180"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column("event_occurrences", sa.Column("thread_id", sa.BigInteger(), nullable=True))
+
+    # Each event's most recent occurrence keeps the event's existing thread (and its comments):
+    # this is the occurrence that comment/reply notifications were already being generated against.
+    op.execute(
+        """
+        UPDATE event_occurrences eo
+        SET thread_id = e.thread_id
+        FROM events e
+        WHERE eo.event_id = e.id
+          AND eo.id = (SELECT max(eo2.id) FROM event_occurrences eo2 WHERE eo2.event_id = e.id)
+        """
+    )
+
+    # Every other occurrence of a repeating event gets its own fresh, empty thread
+    op.execute(
+        """
+        DO $$
+        DECLARE
+            occ_id BIGINT;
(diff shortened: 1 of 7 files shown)


app/backend/src/couchers/models/events.py
tristanlabelle, line 1: Self-blocking to test backend and frontend locally

app/backend/src/couchers/migrations/versions/0187_move_event_thread_to_occurrence.py
```diff
@@ -0,0 +1,85 @@
...
+            SELECT eo.thread_id
+            FROM event_occurrences eo
+            WHERE eo.event_id = e.id
+            ORDER BY eo.id DESC
+            LIMIT 1
+        )
+        """
+    )
+
+    op.alter_column("events", "thread_id", nullable=False)
```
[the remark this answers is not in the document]
tristanlabelle: @aapeliv Is it a real concern to handle events that have no occurrences?

app/backend/src/couchers/servicers/events.py
```diff
@@ -551,6 +546,9 @@ def CreateEvent(
 
         def create_occurrence(moderation_state_id: int) -> int:
             nonlocal occurrence
+            thread = Thread()
+            session.add(thread)
+            session.flush()
```
aapeliv: we don't need to do this inside here, i think. let's keep these local functions to produce the moderatable objects as bare as possible

app/backend/src/couchers/servicers/events.py
```diff
@@ -674,6 +673,9 @@ def ScheduleEvent(
 
         def create_occurrence(moderation_state_id: int) -> int:
             nonlocal new_occurrence
+            thread = Thread()
+            session.add(thread)
+            session.flush()
```
aapeliv: ditto

aapeliv (dismissed):
two minor nits; feel free to merge after fixing

app/backend/src/couchers/migrations/versions/0181_move_event_thread_to_occurrence.py
```diff
@@ -0,0 +1,85 @@
...
+Create Date: 2026-08-13 07:21:39.779870
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "0181"
+down_revision = "0180"
```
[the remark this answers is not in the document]
tristanlabelle: Renumbered and retested in latest

app/backend/src/couchers/models/events.py
tristanlabelle: Tested

PR #9534, opened 2026-08-13
Backend: Support running the test db as arm64

tristanlabelle:
I just realized I've been running an emulated database for months which made it super slow. We had a special condition for macOS instead of testing for arm64 processors like my Windows machine + WSL.

## Testing
Ran `setup-db`

**Backend checklist**

## For maintainers


WouldYouKindly:
@tristanlabelle tested on Mac, works the same as on master. Thanks for the fix!

WouldYouKindly (approved):

Issue #9524, opened 2026-08-11
Very long lag clicking on profiles from couchers.org/search


nc225:
To replicate:
Go to https://couchers.org/search
Do a keyword search for "movies" 

Click on the **name** of any profile that comes up. For me this is lagging for 10+ seconds on Chrome and even longer on Firefox. Note: behavior seems fine when you click on the little arrow-out-of-square link button to the right.

nc225:
Resolved earlier this month

PR #9537, opened 2026-08-14
Attempt to fix the 'replay' freeze on expanding the map search results left pane

Gwened:
Investigating the freezes in the map search page.
Performed some profiling on prod.
Seems they happen within sentry/replay, which was recently enabled in https://github.com/Couchers-org/couchers/pull/9104.

> Why processMutation blocks
> Sentry Replay (rrweb) watches the DOM with a MutationObserver. On that expand, it gets a huge mutation batch and serializes it synchronously on the main thread in processMutation.

Attempts to address:
#9521
#9524

Changes by Cursor Grok 4.5.

## Testing
I tested that the map search page still behaves normally in locadev using the Next backend.
However, due to the lower number of users on that environment, I wasn't having the freeze. I only had it on prod.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/ResizeableDrawer.tsx
+++ b/app/web/components/ResizeableDrawer.tsx
@@ -7,6 +7,8 @@ import React from "react";
 import IconButton from "./IconButton";
 
 export const DEFAULT_DRAWER_WIDTH = 420;
+/** Keep in sync with CSS width transitions on the map search layout. */
+export const DRAWER_WIDTH_TRANSITION_MS = 200;
 
 interface ResizeableDrawerProps {
   children: React.ReactNode;
--- a/app/web/features/search/MapSearchContent.tsx
+++ b/app/web/features/search/MapSearchContent.tsx
@@ -1,5 +1,8 @@
 import { styled } from "@mui/material";
-import { DEFAULT_DRAWER_WIDTH } from "components/ResizeableDrawer";
+import {
+  DEFAULT_DRAWER_WIDTH,
+  DRAWER_WIDTH_TRANSITION_MS,
+} from "components/ResizeableDrawer";
 import { RpcError } from "grpc-web";
 import { SearchUser } from "proto/search_pb";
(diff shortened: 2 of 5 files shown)


wbruntra:
*This is Claude, writing at @wbruntra's request — not him.*

Your read that this is a mutation storm hitting `processMutation` looks right. But I think the mutations are coming from somewhere other than the map, and if so this PR's guards won't catch them.

`react-lines-ellipsis` re-runs its full reflow on **every re-render**, not just when the text changes:

```js
componentDidUpdate(prevProps) {
  if (this.props !== prevProps) {   // identity, not value
    this.reflow(this.props);
  }
}
```

`SeachResultUserCard` passes `style={{ wordBreak: "break-all", overflow: "hidden" }}` — a new object each render — so `PureComponent`'s shallow compare always fails and that check is always true. Nothing in the search tree is memoized, so `setDrawerWidth` on each mousemove re-renders all 100 cards, and each one rebuilds `canvas.innerHTML` twice and then removes spans one at a time in a loop.

Measured, one render of 100 cards with ~3,200-character profiles:

```
react-lines-ellipsis   nodes added: 359,300   removed: 319,500
CSS line-clamp         nodes added:     100   removed:     100
```

Per re-render, so per frame while dragging.

The reason it matters here: `initCanvas` does `document.body.appendChild(canvas)`, and it's a `<div class="LinesEllipsis-canvas">`. So these mutations are outside `SearchResultsContainer` (missing the `data-sentry-block`) and aren't matched by `block: [".maplibregl-map", "canvas"]`.

#9541 swaps that component for CSS `line-clamp`, which drops those mutations to ~200. That would also mean Replay could be turned back on after #9538, rather than staying off as a workaround.

Caveats: I haven't reproduced your exact scenario (drawer expand on prod with Replay on), so the attribution is inference from the mechanism — the counts above are measured, the connection to your profile isn't. And since #9538 removed `replayIntegration` entirely, the `instrumentation-client.ts` hunk here will conflict with current `develop`.

PR #9540, opened 2026-08-14
Frontend: fix: revert #8912 to speed up profile ellipsis truncation

papeldeorigami:
Fixes: https://github.com/Couchers-org/bugs/issues/367, https://github.com/Couchers-org/bugs/issues/373 (correction: the issues were actually solved with #9538 but this PR should contribute for a better performance)
**How**: Revert #8912 to speed up profile ellipsis truncation. This is a one-line change.
**Why**: Before that PR, the truncation was done per words instead of letters. Since most people have really long profiles, this can hurt performance.

See also: [react-line-ellipsis docs](https://www.npmjs.com/package/react-lines-ellipsis)

## Testing
*Explain how you tested this PR and give clear steps so the reviewer can replicate.*

#### Before
https://github.com/user-attachments/assets/da483ac4-44ab-452f-9fe3-ac12f378d7a8

#### After
https://github.com/user-attachments/assets/b0b6f330-5553-40fc-b369-b49325bb5923

### Mocked data

To test this I used AI to create a bunch of mocked profiles in the use search hook - I think that can help test other approaches as well:

```ts
// service/search.ts

// line 36
const MOCK_USER_COUNT = 1000;

function generateMockUsers(count: number): SearchUser.AsObject[] {
  const now = Math.floor(Date.now() / 1000);
  const cities = ["London", "Berlin", "Paris", "Madrid", "Amsterdam", "Lisbon"];
  const genders = ["Woman", "Man", "Non-binary"];
  return Array.from({ length: count }, (_, i) => ({
    userId: i + 1,
    username: `mockuser${i + 1}`,
    name: `Mock User ${i + 1}`,
    city: cities[i % cities.length],
    hasCompletedProfile: true,
    hasCompletedMyHome: false,
    lat: 51.5 + Math.floor(i / 40) * 0.01,
    lng: -0.11 + (i % 40) * 0.01,
    profileSnippet: Array.from({ length: 20 }, (_, line) => {
      return `Line ${line + 1} of mock user ${i + 1}'s profile: a fairly long sentence with plenty of words to stress the layout.`;
    }).join("\n"),
    numReferences: i % 5,
    gender: genders[i % genders.length],
    age: 18 + (i % 60),
    lastActive: { seconds: now - (i % 30) * 86400, nanos: 0 },
    hostingStatus: HostingStatus.HOSTING_STATUS_CAN_HOST,
    meetupStatus: MeetupStatus.MEETUP_STATUS_WANTS_TO_MEETUP,
[...]

PR #9541, opened 2026-08-15
Frontend: fix: clamp search card About text with CSS instead of JS

wbruntra:
Search results on the map page render up to 100 member cards at once, each showing the first few lines of that member's About text. Those lines are currently clamped in JavaScript, which measures the text against the browser's layout to find the cut-off point — work that grows with the length of every profile on the page and holds the main thread, so the search box stops accepting input while results appear. This switches the clamping to CSS, which the browser handles as part of the layout it already does. #9538 and #9540 reduced the stall without removing this cause. Split out of #9522, which also proposed virtualising the list.

The full About text now stays in the page rather than being shortened, so browser find can match text that is not visible, copying a card copies the whole snippet, and screen readers get the complete text.

## Testing

Before/after screencasts: [before](https://github.com/Couchers-org/couchers/pull/9522#issuecomment-5301394316), [after](https://github.com/Couchers-org/couchers/pull/9522#issuecomment-5301481255).

To reproduce, point a local frontend at the production API (`NEXT_PUBLIC_API_BASE_URL="https://couchers.org/api"`), search a city with 100 results, and type in the search box while they render.

**Web frontend checklist**

## For maintainers


app/web/features/search/SeachResultUserCard.tsx
```diff
@@ -135,6 +134,16 @@ const HaikuContainer = styled("div")(({ theme }) => ({
...
 }));
 
+const ClampedAbout = styled("div", {
+  shouldForwardProp: (prop) => prop !== "maxLine",
+})<{ maxLine: number }>(({ maxLine }) => ({
+  display: "-webkit-box",
+  WebkitBoxOrient: "vertical",
+  WebkitLineClamp: maxLine,
+  overflow: "hidden",
+  wordBreak: "break-all",
```
papeldeorigami:
```suggestion
  overflow-wrap: "break-word",
```

Reason: break-all used to make sense for the react-lines-ellipsis lib, but now it's superseded. It is too aggressive, ignoring word boundaries altogether, while the newer CSS overflow-wrap with break-word only breaks a word if it can't fit on a line by itself, like a long URL. Also used in EventCard.tsx:56, though there we still use the legacy wordBreak key.

app/web/features/search/SeachResultUserCard.tsx
```diff
@@ -135,6 +134,16 @@ const HaikuContainer = styled("div")(({ theme }) => ({
   flexGrow: 1,
 }));
 
+const ClampedAbout = styled("div", {
+  shouldForwardProp: (prop) => prop !== "maxLine",
+})<{ maxLine: number }>(({ maxLine }) => ({
```
papeldeorigami: With a specialized component, I feel that the parent shouldn't decide the maxLine anymore. How about using a simple theme breakpoint, like:
```
  WebkitLineClamp: 7,
    WebkitLineClamp: 3,
  },
```

Then we also get rid of the isMobile parameter, which was only used for this purpose.

papeldeorigami (changes requested):
Reviewer note: This is a follow-up from #9522.

@wbruntra thanks for creating the targeted change. PTAL at the comments below.

As a next step, we could get rid of the react-lines-ellipsis altogether. The only remaining usage is in NotificationItem.tsx. I would recommend a separate PR for that.

wbruntra:
Both suggestions make sense to me; the new commit incorporates them both.

papeldeorigami (approved):
_Approved!_ 👍 
I did further tests with the same mock code I used in #9540 and confirmed that this approach can handle profiles with thousands of lines (while the JS lib was always hanging on my notebook).

darrenvong:
> As a next step, we could get rid of the react-lines-ellipsis altogether. The only remaining usage is in NotificationItem.tsx. I would recommend a separate PR for that.

Might it be worth refactoring the `ClampedLines` component to be a more reusable utility component than inlining it? Not suggesting we have to do it in this PR, as that work feels relevant for a follow-up PR but thought I'd mention it rather than AI slopping the same components inline repeatedly?

kevinortiz43:
@darrenvong I am working on a fix now using the line-clamp, which we are using as a solution in frontend repo anyway. Just making some final touches to the pull request before review

kevinortiz43:
So looking at this pull request it was what I was trying to do as well, so I will lose my pull request and approve this one,

kevinortiz43 (approved):

Issue #9546, opened 2026-08-15
[Backend] Add Event recurrence model + scheduled job to spawn EventOccurrence


tristanlabelle:
Part of #9544 (Milestone 1: Proof of concept with private beta).

- Add new `EventRecurrence` model object, at most one per `Event`
  - Data model: RFC5545-compliant RRULE, `last_spawned_date`, `ends_on_date`
- Run scheduled job every hour
  - Create occurrences based on `get_occurrences_to_spawn`
  - After creating occurrences, update the start on date to the day after the most recent created occurrence (so we never double spawn)
- Add tests (with database manipulation)

Rationale:

- `EventRecurrence.last_spawned_date`: Avoids duplicate spawning of occurrences. If we spawn the next two occurrences, then edit their dates or cancel/delete them, the scheduled job stays idempotent because it'll never respawn an event <= `last_spawned_date`. This also allows us to later support changing the day of recurrence.
- `EventRecurrence.ends_on_date`: I think it's a good safety net to make event series expire after max 6 months. We can later support manual renewal by the organizer.

janiespalem:
Hi @tristanlabelle, I would like to take this issue if it is still available. I already have #9609 under review for the dummy profile data, and I would like to continue with a more substantial backend contribution. I checked that #9545 is complete. My plan is to follow the existing model, migration, and scheduled-job patterns, reuse `get_occurrences_to_spawn`, keep spawning idempotent through `last_spawned_date`, and cover the behavior with database tests. Could you confirm that this issue is ready to pick up and that these are the intended boundaries?

tristanlabelle:
Hi @janiespalem , work in the parent Epic is distributable and we can chat about that, but this one has a PR opened already: #9674 . You're welcome to help review it though I'll need to wait for a final review by a core dev. Are you on the team? Reach out to me on Slack and let's chat.

janiespalem:
Hi @tristanlabelle, not yet — I have been contributing through the open-source route. I would be happy to join the team and chat on Slack. What is the best way to get invited? I am also interested in taking a backend part of #9544 once we agree on a good boundary, and I can contribute regularly.

tristanlabelle:
I'm not sure if or how we accept contributions from forks, especially since we have contributor agreements, etc. Before we can chat about assigning work to you, please apply to be a volunteer: https://couchers.org/volunteer

janiespalem:
[...]

Issue #9544, opened 2026-08-15
EPIC - Recurring events


tristanlabelle:
Recurring events is a recurring (ha!) request from weekly event organizers in cities like Berlin and New York. This issue tracks the work to implement it.

For the overall implementation plan, we'll keep the `Event` / `EventOccurrence` split and add an optional `EventRecurrence` object that can be attached to an `Event` to represent its recurrence rule and serve as a template for new `EventOcurrence` objects.

## Milestone 1: Proof of concept with private beta

- #9545
- #9546
- #9547
- #9548

## Milestone 2: Minimal public feature

- #9549
- #9550
- #9551

## Milestone 3: Full public feature

- #9552
- #9553
- #9554
- #9555
- #9556

## Open questions

- How does inviting the community work with recurring events?
  - Suggestion: Only allow once per event series as to not spam people.
- How do we prevent abuse of recurring events (flooding)?
  - Suggestion: Disallow more than 2 occurrences per week (start with 1)
  - Suggestion: Only create a few events ahead of time
  - Suggestion: Require same-day events when recurring? (might be awkward for a late night event that runs from 8pm to midnight+)
- Do recurring events have an end date?
  - Suggestion: Yes, max 6 months ahead, renewable, notification to owner when renewal needed (avoids zombie recurring events).
- Do we expose Event id's in the API
  - Suggestion: Seems like we don't need it. The only awkward bit is `GetEvent(event_occurrence_id: int, recurrence_template: bool)` to query the template
- Do we allow editing the recurrence rule?
  - Suggestion: Not now, but ensure that data model will support it by storing the date of the most recent spawned occurrence and only allowing later occurrences
- Anything to do for moderation state on the series?
- Any missing moderation APIs?
- API to get events in the series?
  - Suggestion: Stretch to support navigation to previous/next occurrence
- Anything to do to facilitate users following a series?
  - Suggestion: Can treat as individual events from a user's perspective and wait for feedback.
- Any timezone trickiness?
  - Suggestion: Unclear. Pay attention as we go.
- How do we explain the event occurrence spawn rule to users?

PR #9568, opened 2026-08-16
Backend/events: Add logic for computing recurring event occurrence dates

### LINKED ISSUE #9545: [Backend] Implement pure recurrence logic
Part of #9544 (Milestone 1: Proof of concept with private beta).

- Find a library to evaluate RRULE for future dates
- Implement `get_future_ocurrences(from_date, rrule): Generator[date]`
- Implement `get_occurrences_to_spawn(last_spawn_date, rrule): list[date]`
  - Rule: return occurrences for coming 2 weeks after `now()`, or a minimum of 2 occurrences, filtered to dates after `last_spawn_date`.
- Add unit tests

Rule for creating occurrences early: create all occurrences within 2 weeks of `now()`, or a minimum of 2 occurrences (e.g. for biweekly). We can tweak the constants but this strikes a balance between seeing upcoming events whether frequent or not.

tristanlabelle:
Implements the logic to determine which event occurrences to schedule ahead of time for recurring events (#9544). Pure logic, no database. The next issue #9546 will store the rrule in the model and invoke this logic in a scheduled job to maintain the next few created events.

Closes #9545.

## Testing

**Backend checklist**

## For maintainers


aapeliv (changes requested):
dateutil.rrule already has a system for producing these, why not use that?

tristanlabelle:
> dateutil.rrule already has a system for producing these, why not use that?

@aapeliv No reason -- I didn't realize I could use that and switched over now.

aapeliv (commented):
i think this is looking good.

my one main question is: why use rrule only for date and not full datetime?

the other tiny question is: would we want the schedule occurrences interface to instead produce a certain number of events instead of fill out a certain window? i'm not sure. i feel like a policy of "schedule the next 2 occurrences" is more like what we would want.

tristanlabelle:
@aapeliv 

> why use rrule only for date and not full datetime?

Hmm my thinking is that we not only need to evaluate the rrule but also to support editing the recurrence in a user-friendly way, which gets more complicated the more we store into a single complex rrule table row in the data model (we then have to parse it out and make sense of it given what our UI can express). Since rrule is mostly about date recurrence complexity, not so much time, I figured it's better to keep out of it. It depends whether our data model is:

1. day-of-week, frequency, start-time, end-time, days-count (nothing prebaked as rrule, easily editable)
2. rrule (everything prebaked, harder to edit)

> would we want the schedule occurrences interface to instead produce a certain number of events instead of fill out a certain window? i'm not sure. i feel like a policy of "schedule the next 2 occurrences" is more like what we would want.

It supports both. The idea was that if you have an event recurring very often, like a few times a week, you probably want to generate at least two weeks' worth of it as it's useful to know what's happening next week. If the event is every other week though, then we care more about generating the two instances. We have both knobs but don't have to use both of them.

tristanlabelle:
Merging per Slack convo. We can still tweak whether to include the time and what to store in the model in the next PR which will add the model representation for recurring events.

PR #9571, opened 2026-08-16
Point VSCode at the backend venv interpreter

aapeliv:
The repo ships a `.vscode` directory with recommended extensions and workspace settings so that a checkout gives roughly the same editor experience to everyone. On the Python side that has so far only covered Ruff and the pytest configuration, leaving each person to point their editor at the backend virtualenv by hand before things like imports resolving, go-to-definition and the test runner work. This adds the Python extension to the recommendations and sets the default interpreter to the backend venv, so a fresh checkout picks it up automatically.

## For maintainers

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,6 +3,7 @@
     "dbaeumer.vscode-eslint",
     "esbenp.prettier-vscode",
     "zxh404.vscode-proto3",
-    "charliermarsh.ruff"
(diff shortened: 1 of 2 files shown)


nabramow (approved):
Seems fine to me!

I think the bscode-proto3 one is outdated though? I keep getting a warning about it and then when I add it it gives me some kinda error so I end up stuck with it always reminding me.

PR #9572, opened 2026-08-16
Give each checkout its own backend test database

aapeliv:
The backend test suite runs against a dedicated postgres instance, and at session start it drops and recreates a database called `testdb`. That name is the same for everyone, so running two suites at once — out of separate checkouts against one local test postgres — means the second run destroys the first run's database out from under it mid-suite. Each run now owns a database named after its checkout, so `DATABASE_CONNECTION_STRING` decides only which postgres and credentials to use; `TEST_DB_NAME` overrides the name for two suites out of a single checkout. Two tests that bound the Prometheus exporter and the gRPC servers to fixed ports, which collide the same way, now take ephemeral ones.

Since the database name is always chosen by the test harness, pointing `DATABASE_CONNECTION_STRING` at a real database can no longer get that database dropped.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/tests/conftest.py
+++ b/app/backend/src/tests/conftest.py
@@ -1,5 +1,5 @@
+import hashlib
 import os
-import re
 from collections.abc import Generator
 from pathlib import Path
 from tempfile import TemporaryDirectory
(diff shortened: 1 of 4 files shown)


tristanlabelle (approved):
Manipulating the environment before imports is stinky but the result is super useful!

PR #9577, opened 2026-08-16
Empty the test database with DELETE instead of TRUNCATE

aapeliv:
The backend test suite resets its database between every test, clearing all 85 non-static tables so each test starts clean. TRUNCATE does that by allocating a fresh relfilenode for every table, index, toast relation and sequence involved — roughly 600 files per call — and costs the same whether those tables hold a million rows or none, so it falls as a flat tax on all 1341 tests. The reset now clears the tables with DELETE and rewinds the sequences with setval, reaching the same state and taking the suite's total reset time from about 26s to 2s. A bare `pytest` invocation works again too, which needed `testpaths` set so the tests' conftest is loaded early enough to register its `--query-log` option.

Notifications held back behind moderation are now released in a defined order. They were queued in whatever order Postgres returned them, which TRUNCATE made chronological by handing back an untouched heap; on production tables that order was already arbitrary.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/pyproject.toml
+++ b/app/backend/pyproject.toml
@@ -111,6 +111,8 @@ extra-standard-library = ["zoneinfo"]
 # Pytest
 [tool.pytest.ini_options]
 pythonpath = ["src"]
+# Also makes src/tests/conftest.py an initial conftest, so its pytest_addoption runs
+testpaths = ["src/tests"]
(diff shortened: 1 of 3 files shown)


tristanlabelle (approved):
Nice!

PR #9580, opened 2026-08-16
Freeze the clock in the tests that bound a timestamp

aapeliv:
Several tests across pages, discussions and events check that a `created` or `last_edited` timestamp is sane by taking `now()` in python either side of the RPC and asserting the value the database wrote falls between the two. Python reads the host clock and postgres reads the clock in its container, and those drift apart by a millisecond or two, so the assertions fail at random — most recently four in `test_pages` and one in `test_discussions` in the same run. They now run on the `frozen_timewarp` fixture, which stops a clock that both python and postgres read, so the comparison is against the same instant on both sides.

Where a test is checking that an edit moves `last_edited` on from `created`, it advances the frozen clock itself between the two, which makes that assertion exact rather than dependent on how long the RPC happened to take.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/tests/test_discussions.py
+++ b/app/backend/src/tests/test_discussions.py
@@ -51,7 +51,7 @@ def test_create_discussion_errors(db):
         assert e.value.details() == "Missing discussion content."
 
 
-def test_create_and_get_discussion(db, push_collector: PushCollector, moderator: Moderator):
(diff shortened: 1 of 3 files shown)


tristanlabelle (dismissed):
Yay for frozen clocks

PR #9593, opened 2026-08-17
Only count a reply when the comment it hangs off is visible

aapeliv:
A thread's `num_responses` counted replies under comments the thread itself hides. `total_num_responses` joined `Comment` only to reach the thread, applying moderation and author visibility to the `Reply` but nothing to the `Comment`, while `GetThread` filters the comments it lists on both. So once a moderator hid a comment that already had approved replies, those replies kept counting, and the count sat next to a listing that showed nothing. The same held before any moderator acted: a reply approved under a still-shadowed comment counted for its author, who cannot see the comment it hangs off. Filter the joined comment the same way the listing does, keeping deleted comments in scope since `GetThread` renders those as stubs with their replies.

## Testing

- `test_total_num_responses_excludes_replies_under_hidden_comment`: a comment and its reply are both approved and the count is 2; hiding the comment drops both `GetThread`'s listing and the count to 0. Fails on `develop` (count stays 2 while the listing is empty).
- `test_total_num_responses_excludes_replies_under_someone_elses_shadowed_comment`: the replier's own approved reply doesn't count while the comment above it is shadowed to them, and counts once that comment is approved. Fails on `develop`.
- `src/tests/test_threads.py`, `test_discussions.py`, `test_pages.py` and `test_communities.py` pass locally; `make format` and `make mypy` clean.

**Backend checklist**
- [x] ~~Added migrations if there are any database changes~~ (no database changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/threads.py
+++ b/app/backend/src/couchers/servicers/threads.py
@@ -66,20 +66,19 @@ def total_num_responses(session: Session, context: CouchersContext, database_id:
         Comment,
         is_list_operation=True,
     )
-    replies = where_moderated_content_visible(
-        where_users_column_visible(
-            select(func.count())
-            .select_from(Reply)
-            .join(Comment, Comment.id == Reply.comment_id)
-            .where(Comment.thread_id == database_id)
-            .where(Reply.deleted == None),
-            context,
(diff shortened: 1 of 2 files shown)


aapeliv:

PR #9594, opened 2026-08-17
Only count public trip offers the owner can open

aapeliv:
A public trip's `offers_count` counted every non-cancelled host request linked to the trip, applying neither moderation nor user visibility, while the thread list the owner opens those offers from filters on both. A host request is created shadowed, so this was the ordinary case rather than an edge case: the owner saw a count of offers awaiting initial review, and clicked through to a list that didn't have them. An offer from a banned, deleted or blocked user counted the same way. Filter the count exactly as the thread list is filtered — the offer's own moderation visibility, and the offering user's visibility. The recipient of an offer on a trip is the trip owner, i.e. the viewer, so only the offering side needs a visibility check.

## Testing

- `test_list_public_trips_by_user_offers_count_owner` now pins the whole lifecycle: a fresh offer is shadowed and doesn't count, and it counts once approved. The pre-approval assertion fails on `develop`.
- `test_list_public_trips_by_user_offers_count_excludes_invisible_offers` (new): two approved offers count as 2; hiding one and banning the other's author takes the count to 0. Fails on `develop`.
- `test_list_public_trips_by_user_offers_count_excludes_cancelled` now approves its offers, so it still exercises the status rule rather than passing for the wrong reason.
- `src/tests/test_public_trips.py`, `test_requests.py` and `test_message_threads.py` pass locally; `make format` and `make mypy` clean.

**Backend checklist**
- [x] ~~Added migrations if there are any database changes~~ (no database changes)

## For maintainers

### DIFF
--- a/app/backend/src/couchers/servicers/public_trips.py
+++ b/app/backend/src/couchers/servicers/public_trips.py
@@ -77,12 +77,18 @@ def public_trip_to_pb(
         same_gender_only=public_trip.same_gender_only,
     )
     if public_trip.user_id == context.user_id:
-        pb.offers_count = session.execute(
+        # Filtered like the thread list the owner opens the offers from, since an offer they can't
+        # open shouldn't raise the count. The recipient of an offer on the trip is the owner, i.e.
+        # the viewer, so only the offering user's visibility needs checking.
+        offers = (
             select(func.count())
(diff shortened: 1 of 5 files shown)


aapeliv:

PR #9599, opened 2026-08-17
Stop notifications about hidden content reaching its author

aapeliv:
Notifications carry the moderation state of the content they are about, so that a notification is only delivered and listed once that content is visible, with an exception that lets an author see notifications about their own content while it is still shadowed. That exception was applied without any test on the visibility itself, so it also let through content a moderator had hidden: when a group chat, host request, event or any other moderated object was hidden, its author kept receiving the push notifications, emails and unseen count for it, carrying the content, while every other user correctly stopped. The exception is now restricted to shadowed content, matching the rule the equivalent helper for the content itself already applies.

**Backend checklist**

## For maintainers

### DIFF
--- a/app/backend/src/couchers/sql.py
+++ b/app/backend/src/couchers/sql.py
@@ -235,18 +235,21 @@ def moderation_state_column_visible(
                 )
             )
 
+    conditions = [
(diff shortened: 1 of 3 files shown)


aapeliv:

PR #9609, opened 2026-08-21
Complete dummy data profiles

### LINKED ISSUE #6074: Make dummy data profiles complete
Right now the dummy data profiles don’t have pictures or full bios, so when running locally, one has to upload a picture and write a bio before sending reqs.

It should instead upload some sample images and have full bios.

janiespalem:
Closes #6074

## Summary
- add complete bios and profile-gallery avatars to every dummy user
- include dedicated metadata-free sample avatars and expose them through the local media service
- verify seeded profiles are complete and every avatar upload has a corresponding file

The sample avatars are simple project-local abstract images: they contain no real people, third-party assets, text, or EXIF metadata.

## Testing
- `make docker-test file=test_dummy_data.py`
- `ruff check src/dummy_data.py src/tests/test_dummy_data.py`
- `ruff format --check src/dummy_data.py src/tests/test_dummy_data.py`
- `mypy src/dummy_data.py`
- exercised both full-size and thumbnail avatar URLs against the media container


app/docker-compose.yml
```diff
@@ -51,6 +51,8 @@ services:
     env_file: media.dev.env
     volumes:
       - "./data/media/uploads/:/uploads/"
+      - "./backend/src/data/dummy_profile_images/:/dummy-profile-images/:ro"
+    command: sh -c "mkdir -p /uploads/full && cp -n /dummy-profile-images/*.jpg /uploads/full/ && exec uwsgi --http '[::]:5001' --wsgi-file src/media/server.py --callable app"
```
aapeliv: is there a way to just mount the directory directly? for the local env, we don't need to mount it in data, i think

aapeliv (changes requested):
thanks for the diff! i think overall it looks good. one small comment, and i will write a new bio for myself at some point :P

app/docker-compose.yml
fearlesstilted: Updated in b93eb73ba: the image directory is now mounted directly at /uploads/full, with no copy step or command override. The backend test and local media mount check pass. Thanks!

janiespalem:
Hi @aapeliv, following up on this PR. I addressed the directory mount comment in b93eb73ba. Could you take another look when you have time, or let me know if anything else needs changing? Thanks!

PR #9610, opened 2026-08-21
Frontend/feature flags: Fix spinning forever due to growthbook initialization

tristanlabelle:
The frontend never got notified that `feature-flags.dev.json` was loaded and spinned forever because of the way we did growthbook initialization. It's possible it worked if things raced a different way than on my machine.

## Testing
yarn start before/after.

**Web frontend checklist**

## For maintainers


aapeliv (approved):
thanks for spotting this!

PR #9612, opened 2026-08-22
Frontend/map: Fix null reference

tristanlabelle:
Sentry reports 4.5K TypeErrors due to `getSource()` returning `null` here. https://couchers.sentry.io/issues/7682497623

It looks like `map.current.once("load"` is racing with `redrawMap`, such that sometimes we redraw a map before the circle source has been created. I presume we can no-op in redrawMap if the source hasn't been created yet, and expect the redraw function to be called again once it has. Also introduces constants for clarity.

## Testing
Did a search on the vercel preview.

**Web frontend checklist**

## For maintainers

### DIFF
--- a/app/web/components/EditLocationMap.tsx
+++ b/app/web/components/EditLocationMap.tsx
@@ -131,7 +131,7 @@ export default function EditLocationMap({
   };
 
   const redrawMap = () => {
-    const circle = map.current?.getSource(CIRCLE_SOURCE_NAME) as GeoJSONSource;
+    const circle = map.current?.getSource(CIRCLE_SOURCE_NAME) as GeoJSONSource | null;
     if (!circle) return;
     if (!exact) {
       circle.setData(circleGeoJson(extractLngLat(location.current), location.current.radius));
--- a/app/web/components/EditLocationMap.tsx
+++ b/app/web/components/EditLocationMap.tsx
@@ -13,6 +13,9 @@ import { ControllerRenderProps, FieldError } from "react-hook-form";
 
 import { GLOBAL } from "../i18n/namespaces";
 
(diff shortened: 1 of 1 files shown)


app/web/components/EditLocationMap.tsx
```diff
@@ -122,19 +125,18 @@ export default function EditLocationMap({
...
     if (!isBlank.current) {
-      map.current.setLayoutProperty("circle", "visibility", "visible");
+      map.current.setLayoutProperty(CIRCLE_LAYER_NAME, "visibility", "visible");
     }
     redrawMap();
   };
 
   const redrawMap = () => {
-    if (!map.current) return;
+    const circle = map.current?.getSource(CIRCLE_SOURCE_NAME) as GeoJSONSource | null;
```
darrenvong: The type asserting seems unnecessary here - `.getSource` already returns a proper type and with the optional chaining, TS should be able to infer it's nullable/undefined

tristanlabelle: Without the cast:
image

tristanlabelle: GeoJSONSource is probably a subtype of Source

darrenvong: Oh I see, the type cast was actually masking the bad type when I looked without deleting it first 🙃 

It looks like `getSource` accepts a generic arg when calling so might be a bit nicer to write:
```typescript
const circle = map.current.getSource<GeoJSONSource>("circle");
// rest of code
```

as `getSource` seems to return `undefined` rather than `null` so it'll be slightly more accurate.

tristanlabelle: Good catch, updated

darrenvong (approved):
LGTM

PR #9622, opened 2026-08-24
Stop buildx exporting the provenance attestation in CI

aapeliv:
CI builds every image with `docker buildx build --push` into the GitLab container registry. Since 12 August those pushes have been failing intermittently with `blob unknown to registry`, naming the image manifest rather than a layer, and have needed manual retries to get a pipeline green. buildkit 0.32 began exporting the provenance attestation as an OCI artifact whose subject is the image manifest, which turns every push into an index; GitLab's registry rejects that index while the subject manifest is still new to it (moby/buildkit#7007). We don't consume provenance anywhere, so this sets `BUILDX_NO_DEFAULT_ATTESTATIONS` and the pushes go back to being a plain manifest.

## For maintainers


tristanlabelle (approved):
If that fixes it, nice find

